PR #1, opened 2017-07-19
Create a component guide based on components in an application

fofr:
Part of:
https://trello.com/c/SD6M3IUF/12-3-create-a-component-guide-that-can-document-components-in-apps

Not all components should be in static. If they are used only in one application it is preferable to keep them close to the point of use.

Initial release:
* Allow applications to create components
* Components follow the same file structure as static components
* Component guide is generated from docs and partials living in `app/views/components/docs` and `app/views/components`
* An example of how to use this gem: https://github.com/alphagov/government-frontend/compare/components-gem
* Component guide logic, templates and styles ported from https://github.com/alphagov/govuk-component-guide and http://govuk-component-guide.herokuapp.com/
* Includes basic integration tests

See README.md for further details.

Still to do but can be separate PRs:
* Syntax highlighting
* Component guide styles should be components themselves
* Further documentation on components and principles
* Component previews that render just the component (ie no GOV.UK frame around it)

screen shot 2017-07-19 at 17 18 31
screen shot 2017-07-19 at 17 18 49


Gemfile.lock
```diff
@@ -2,7 +2,10 @@ PATH
   remote: .
```
tijmenb: The `Gemfile.lock` shouldn't be committed for gems.

fofr: Thanks, removed in 32895dd

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
...
+`govuk_publishing_components` expects the following components structure in your application:
+
+* a Partial View in `app/views/components` eg `_component-name.html.erb` - The template logic and markup, also defines the arguments expected
+* Documentation in `/app/views/components/docs` - a `.yml` per component, describing the component and containing fixture data.
+* a SCSS module in `app/assets/stylesheets/components` - The styling of the component
+* a Javascript module in `app/assets/javascripts/components` - JS behaviour of the component
+* a unit test in `/test/components` - testing the component renders correctly based on parameters passed in
+
+The component guide will include your application’s styles and scripts. It will include `application.scss` and `application.js` by default but this is [configurable](#configuration).
+
```
andysellick: Should there be a line or two in here about how to call/use a component? I guess it's already going to be in the component guide though.

README.md
```diff
@@ -1,11 +1,10 @@
...
+# GOV.UK Publishing Components
+
+A gem to document components in GOV.UK frontend applications.
 
 ## Usage
-How to use my plugin.
 
-## Installation
-Add this line to your application's Gemfile:
+Add this line to your application's Gemfile in the development and test groups:
```
andysellick: As someone new to this, I don't know what the "development and test groups" are, so I'd be stuck on this step.

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
...
+### Mount the component guide
+
+Mount the component guide in your application:
+
+```ruby
+# config/routes.rb
+unless Rails.env.production?
+  mount GovukPublishingComponents::Engine, at: "/component-guide"
+end
+```
```
andysellick: Is this a change that once done is okay to commit to that application? If so, presumably over time eventually all apps will have this already. Some clarification/expansion needed here?

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
...
+
+For example, if your application was government-frontend, the component guide would be found at:
+`government-frontend.dev.gov.uk/component-guide`
+
+### Write components
+
+`govuk_publishing_components` expects the following components structure in your application:
+
+* a Partial View in `app/views/components` eg `_component-name.html.erb` - The template logic and markup, also defines the arguments expected
```
andysellick: Having an example file name here is helpful - can example file names also be added for the following items as well?

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
...
+* Documentation in `/app/views/components/docs` - a `.yml` per component, describing the component and containing fixture data.
+* a SCSS module in `app/assets/stylesheets/components` - The styling of the component
+* a Javascript module in `app/assets/javascripts/components` - JS behaviour of the component
+* a unit test in `/test/components` - testing the component renders correctly based on parameters passed in
```
vanitabarrett: Is it worth pointing people towards an example component somewhere, so they can see exactly what you mean by the above points? Things like 'defines the arguments expected' - it may not be immediately obvious to everyone exactly how this is done?

vanitabarrett: Also, even though we're not giving people a comprehensive guide on how to write a component here, it might still be worth linking to our component principles?

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
 $ bundle
 ```
```
andysellick: Where should these commands be executed?

README.md
fofr: I think it's worth adding one or two lines.

README.md
```diff
@@ -16,13 +15,63 @@ And then execute:
...
+description: A banner that displays when content is withdrawn
+fixtures:
+  default:
+    title: 'This consultation was withdrawn on 20 April 2017'
+    explanation_html: '<p>A supporting paragraph</p>'
+```
+
+### Configuration
+
+You can configure the gem with a config block in an initializer:
```
andysellick: Remove 'you can'.

README.md
fofr: Yes, definitely. Holding off doing this until we have one to link to.

README.md
fofr: These are the standard Rails "How to install gem" instructions.

fofr:
README updated in https://github.com/alphagov/govuk_publishing_components/pull/1/commits/b929dd2aced28550096f60eebdbd6b7fe6bdf17a

README.md
```diff
@@ -16,13 +15,75 @@ And then execute:
...
+
+The `.yml` file must have:
+* a name
+* a description
+* a default fixture (this can be `{}` if the component takes no parameters)
+
+#### Example documentation
+
+```yaml
+name: Withdrawal notice
```
alecgibson: (Just a thought, but it might be nice with all of these examples to demonstrate a single component? Then you can see the process all the way through for one thing. Above for your `<%= render` example, you use descriptions; for the path examples you use a print link; here you're using a withdrawal notice)

README.md
```diff
@@ -16,13 +15,75 @@ And then execute:
...
+* test/components/print_link_test.rb
+
+The component guide will include your application’s styles and scripts. It will include `application.scss` and `application.js` by default but this is [configurable](#configuration).
+
+### Write documentation
+
+The `.yml` file must have:
+* a name
+* a description
+* a default fixture (this can be `{}` if the component takes no parameters)
```
alecgibson: It might be nice to expand on what a "default" fixture is? And what a fixture that isn't default is?

README.md
```diff
@@ -16,13 +15,75 @@ And then execute:
...
+```
+
+### Configuration
+
+Configure the gem with a config block in an initializer:
+
+```ruby
+# config/initializers/govuk_publishing_components.rb
+GovukPublishingComponents.configure do |c|
+  c.component_guide_title = "My component guide"
```
alecgibson: What's the use case for being able to configure this?

app/controllers/govuk_publishing_components/component_guide_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+                             {
+                               title: @component_doc.name
+                             }
+                           ]
+    end
+
+    def fixture
+      @component_doc = ComponentDoc.get(params[:component])
+      @component_fixture = @component_doc.fixtures.find { |f| f.id == params[:fixture] }
+      @guide_breadcrumbs = [
```
alecgibson: It might be nice to break these breadcrumbs out into their own little method.

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -0,0 +1,34 @@
+module GovukPublishingComponents
+  ComponentDoc = Struct.new(:id, :name, :description, :body, :fixtures) do
```
alecgibson: This class has a lot of collection-level functionality, which feels like it should belong in another separate singleton class that has the ability to memoize things properly, and deliver things from hashes rather than performing `find` operations.

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -0,0 +1,34 @@
...
+
+    def fixture
+      fixtures.first
+    end
+
+    def other_fixtures
+      fixtures.slice(1..-1)
+    end
+
+    def self.fetch_component_doc
```
alecgibson: fetch_component_doc**s** (plural)?

app/models/govuk_publishing_components/component_fixture.rb
```diff
@@ -0,0 +1,15 @@
+module GovukPublishingComponents
+  ComponentFixture = Struct.new(:id, :data) do
+    def name
+      id.humanize
+    end
+
+    def pretty_data
+      JSON.pretty_generate(data).gsub('\\n', "\n    ").gsub(/"(\w*)":/, '\1:').strip
```
alecgibson: When performing this sort of `gsub` wizardry, it's often nice to break it out into a one-line mutating method with a nice name so that people can understand what's going on, eg:

```ruby
def pretty_data
  prettified = JSON.pretty_generate(data)
  prepend_new_lines_with_whitespace!(prettified)
  unquote_properties!(prettified)
  prettified
end
```

On that note, just why _are_ we doing these things? To make a Ruby-looking hash? In that case maybe we should be using a gem? Or at least rename this method to something like `data_as_pretty_printed_ruby_hash`?

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -0,0 +1,34 @@
...
+    def self.build(component)
+      fixtures = component[:fixtures].map { |id, data|
+        GovukPublishingComponents::ComponentFixture.new(id.to_s, data)
+      }
+
+      self.new(component[:id], component[:name], component[:description], component[:body], fixtures)
+    end
+
+    def fixture
+      fixtures.first
```
alecgibson: This doesn't necessarily grab the "default" fixture - is that what we want?

app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -0,0 +1,32 @@
+<% content_for :title, "#{@component_doc.name} component" %>
+<%= render 'govuk_component/title', title: @component_doc.name, context: 'Component' %>
+
+<p class="component-description">
+  <%= @component_doc.description %>
+</p>
+
+<div class="component-show">
+  <div class="component-doc">
+    <h3>How to call this component</h3>
```
alecgibson: (This view and the `fixture` view share a bit of code in common. I don't know if it's possible to commonise through a partial?)

README.md
fofr: I agree. My intention is to fix this, so a component without any fixtures does not take parameters. And an "empty parameter" option can be easily illustrated. Was going to do this as a separate PR.

README.md
fofr: Good point. When we're written some I'll link directly to the files in other apps.

README.md
fofr: Unique titles for component guides in different apps.

app/models/govuk_publishing_components/component_doc.rb
fofr: This logic has largely been ported from https://github.com/alphagov/govuk-component-guide/

https://github.com/alphagov/govuk-component-guide/blob/master/app/models/component.rb

app/models/govuk_publishing_components/component_fixture.rb
fofr: Another ported model from here:
https://github.com/alphagov/govuk-component-guide/blob/master/app/models/fixture.rb#L7

It creates a hash that can be put into a `<pre>` block with correct formatting. I agree the naming should be better.

README.md
fofr: See: https://github.com/alphagov/government-frontend/pull/403/files#diff-bb6adab3eaf25dfc303b4f34455f5be9R2

app/controllers/govuk_publishing_components/component_guide_controller.rb
fofr: Added in 84537ad

app/models/govuk_publishing_components/component_doc.rb
fofr: Fixed in d723a47

README.md
alecgibson: I understand what it does, but why would we want to do this?

fofr: Because each guide will be different. Each guide will have different components.

app/models/govuk_publishing_components/component_doc.rb
fofr: This relies on convention, you’re right. I think this is ok for now. The convention of putting default first in the fixtures hasn't been an issue with static components.

The alternative would be to lookup via the `default` ID – if this was left out what should the guide page do? At the moment it'd fallback to the next fixture.

I think when there's a component generator this convention will be natural too.

fofr:
Plan regarding logic and review comments is to:
* Integrate this gem into the govuk-component-guide and remove duplicate logic: https://trello.com/c/GtIEFBfC/
* Then refactor

PR #3, opened 2017-07-24
Allow components to use application view helpers

fofr:
Previously a component that used it’s own app’s view helpers would error on the component guide with: `undefined method`

* Add example of component using a helper
* Write test against component
* Include Rails.application.helpers in ApplicationController to make those helper methods available when rendering the component through this Engine. https://stackoverflow.com/questions/9232175/#comment44856151_19453140

Note this does the opposite of a conventional gem: It needs things from the application to render views, rather than providing methods to an app.


nickcolley:
Epic 🚀

PR #5, opened 2017-07-26
Explicitly extend namespaced controller

fofr:
Depending on load ordering the value of “ApplicationController” can vary.
See: https://github.com/rails/rails/issues/12168

If you visited an application page before visiting the component guide, the guide would render within the host application’s layout rather than the gem’s.

* Be explicit about which ApplicationController we want to use
* Include what was a failing test to demonstrate fix

Hat-tip to @h-lame for pointing me in the right direction.


nickcolley:
Nice work @fofr and @h-lame 👍

PR #6, opened 2017-07-26
Render accessibility criteria and body on component page

fofr:
* Don't process all components every pageload
* Use govspeak to turn markdown into HTML
* Use govspeak (static) component to render HTML
* Style headings based on class not with default styles to avoid risk of affecting component partials and to make re-use easier

https://trello.com/c/qIwQRSai/35-1-gem-render-body-markdown-as-html
https://trello.com/c/gCTvBkD1/36-1-gem-render-accessibility-acceptance-criteria-and-update-documentation

description component - gov uk component guide 20170726

### DIFF
--- a/app/views/govuk_publishing_components/component_guide/show.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/show.html.erb
@@ -1,30 +1,48 @@
 <% content_for :title, "#{@component_doc.name} component" %>
 <%= render 'govuk_component/title', title: @component_doc.name, context: 'Component' %>
 
-<p class="component-description">
-  <%= @component_doc.description %>
-</p>
-
 <div class="component-show">
-  <div class="component-doc">
-    <h3>How to call this component</h3>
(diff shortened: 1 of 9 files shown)


app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -1,30 +1,48 @@
...
-  <%= @component_doc.description %>
-</p>
-
 <div class="component-show">
+  <div class="grid-row">
+    <div class="column-two-thirds">
+      <p class="component-description">
+        <%= @component_doc.description %>
+      </p>
+      <% if @component_doc.body.present? %>
```
nickcolley: We're referencing @component_doc.html_body below, should this conditional be the same?

fofr: No. Body is the markdown version, if it exists then we can parse and rendering the HTML version. This avoids us doing it twice.

PR #15, opened 2017-08-04
Fix height problem with code snippet examples

andysellick:
- some code examples were taller than 300px, caused overflow problem
- moving max height from container to pre fixes

Changes this:

screen shot 2017-08-04 at 15 51 21

to this:

screen shot 2017-08-04 at 15 51 35


app/assets/stylesheets/govuk_publishing_components/component_guide.scss
```diff
@@ -83,6 +82,7 @@ $border-color: #ccc;
 
   pre[class*=language-] {
     padding: $gutter;
+    max-height: 300px;
```
fofr: Was wondering whether we could remove this max-height altogether.

PR #23, opened 2017-08-08
Add a generator for components

andysellick:
- generates all the required files in the right places
- does not generate JS file as not all components need them
- will appear in the component guide without changes but test should fail

https://trello.com/c/57dzR66x/73-write-a-component-generator-in-the-gem


lib/generators/components/templates/component.yml.erb
```diff
@@ -0,0 +1,11 @@
...
+description: add description here
+body: |
+  add body here
+accessibility_criteria: |
+  The component must:
+
+  - add accessibility criteria here
+  - add accessibility criteria here
+fixtures:
+  default: {}
```
fofr: With the exception of the name and the default fixture, where we've used placeholder values, could it match this: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#example-yaml-file

lib/generators/components/templates/_component.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="app-c-<%= @public_name %>">
+  <h2>Please write this component</h2>
```
fofr: This could include a link to here:
https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#example-yaml-file

"How to build a component"

lib/generators/components/templates/component_test.rb.erb
```diff
@@ -0,0 +1,11 @@
+require 'component_test_helper'
+
+class <%= @test_name %>Test < ComponentTestCase
```
fofr: `ComponentTestCase` may not exist in the host app. 🤔

lib/components/engine.rb
```diff
@@ -0,0 +1,6 @@
+require 'rails'
```
fofr: What are these lib files needed for?

lib/generators/components/USAGE
```diff
@@ -0,0 +1,5 @@
+Description:
+    Generates a new component
+
+Example:
+    rails generate components:component [component_name]
```
fofr: I think the namespace should be `govuk_publishing_components` like the rest of the gem.

lib/generators/components/component_generator.rb
```diff
@@ -0,0 +1,37 @@
+require 'rails/generators'
+require 'fileutils'
+
+module Components
+  class ComponentGenerator < Rails::Generators::NamedBase
+    source_root File.expand_path('../templates', __FILE__)
+
+    def copy_component_files
+      @public_name = file_name.dasherize
```
fofr: Do we need to confirm that the file_name is in the format_with_underscores?

lib/generators/components/templates/component_test.rb.erb
andysellick: Fair point. So should I include them as templates and put them in place if they're not found? Or just look for them and output a warning? Seems like putting all this stuff in place could end up getting quite convoluted.

lib/components/engine.rb
andysellick: This is to get the generator to be available from the gem inside the application. Had to look it up. Do you know of another way?

lib/generators/components/templates/component_test.rb.erb
nickcolley: Do we have tests for this in the dummy application? This should be failing I'd have thought?

vanitabarrett:
Do we not use dashes between component names rather than underscores?
screen shot 2017-08-08 at 11 29 51

PR #34, opened 2017-08-17
Split ComponentDoc into ComponentDoc and ComponentDocResolver

fofr:
A small refactor to switch from structs to classes.

Specifically:
* split ComponentDoc model into ComponentDoc and ComponentDocResolver
* makes monkey patching the gem from an app easier (eg govuk-component-guide WIP: https://github.com/alphagov/govuk-component-guide/compare/component-gem#diff-bb6adab3eaf25dfc303b4f34455f5be9)

Addresses some of @alecgibson’s original review comments in https://github.com/alphagov/govuk_publishing_components/pull/1#discussion_r128566361

### DIFF
--- a/app/models/govuk_publishing_components/component_doc.rb
+++ b/app/models/govuk_publishing_components/component_doc.rb
@@ -32,6 +32,10 @@ module GovukPublishingComponents
       govspeak_to_html(accessibility_criteria) if accessibility_criteria.present?
     end
 
+    def partial_path
+      "components/#{id}"
+    end
+
   private
 
     def govspeak_to_html(govspeak)
--- a/app/models/govuk_publishing_components/component_doc.rb
+++ b/app/models/govuk_publishing_components/component_doc.rb
@@ -1,25 +1,19 @@
 module GovukPublishingComponents
-  ComponentDoc = Struct.new(:id, :name, :description, :body, :accessibility_criteria, :fixtures) do
-    def self.get(id)
-      component = fetch_component_doc(id)
-      self.build(component)
-    end
+  class ComponentDoc
+    attr_reader :id,
+                :name,
+                :description,
+                :body,
+                :accessibility_criteria,
+                :fixtures
 
-    def self.all
-      fetch_component_docs.map { |component| build(component) }
-    end
-
-    def self.build(component)
-      fixtures = component[:fixtures].map { |id, data|
-        GovukPublishingComponents::ComponentFixture.new(id.to_s, data)
-      }
-
-      self.new(component[:id],
-               component[:name],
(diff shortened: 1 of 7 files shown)


app/models/govuk_publishing_components/component_doc_resolver.rb
```diff
@@ -0,0 +1,41 @@
...
+                       component[:accessibility_criteria],
+                       fixtures)
+    end
+
+    def fetch_component_docs
+      doc_files = Rails.root.join("app", "views", "components", "docs", "*.yml")
+      Dir[doc_files].sort.map { |file| parse_documentation(file) }
+    end
+
+    def fetch_component_doc(id)
```
alecgibson: So every time we try to access documentation we have to perform multiple file operations? Feels a little bit icky;  could potentially memoise this stuff at a static level? Although given that this is all just internal code used by developers, maybe we don't have to worry about it too much.

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -1,25 +1,19 @@
...
-        GovukPublishingComponents::ComponentFixture.new(id.to_s, data)
-      }
-
-      self.new(component[:id],
-               component[:name],
-               component[:description],
-               component[:body],
-               component[:accessibility_criteria],
-               fixtures)
+  class ComponentDoc
```
alecgibson: Why have we moved away from a `Struct`? It looks like we've just wound up with more boilerplate code. To make it immutable? We could use something like [`immutable-struct`](https://github.com/stitchfix/immutable-struct)?

fofr:
> If we think this is going to happen quite a lot
I see it happening once, for static components that are different. But I like your idea of a subclass component resolver within the gem. I've been monkey patching as a proof of concept. We're also considering retiring `govuk-component-guide` and putting the gem in static. That would require something custom because static can't request from slimmer.

alecgibson:
> I see it happening once, for static components

If that's the case, then do it however you think is best - if it's just for one special case, a whole inheritance system may be overkill.

app/models/govuk_publishing_components/component_doc_resolver.rb
fofr: I want to make it faster, and memoising makes sense. As this is predominantly a tool for developing components we need to balance that with the problem of too much caching hiding local changes.

alecgibson: Makes sense to me. Not sure we'd notice much of a performance change, anyway.

fofr:
Thank you for the review @alecgibson ⭐️

PR #35, opened 2017-08-18
Update BEM docs to include flattened structure

vanitabarrett:
The official BEM docs state that classes do not need to reflect the nested nature of the DOM. We need to update our own BEM docs to include this for other devs, as it can be easily missed. http://getbem.com/faq/#css-nested-elements


nickcolley:
Epic, nice work 👍

PR #43, opened 2017-08-31
Allow static to use component gem

fofr:
Static components currently live in a different directory to app components. (“govuk_component” vs “components”) While this discrepancy exists the gem needs to know whether it’s documenting static components or app components.

Some oddities were also found concerning the way dependencies are required when Slimmer is or isn't there:
* As static doesn't use slimmer, we need to explicitly require it: 9c812fa – tests continued to pass locally but needed to be updated with ba29030 to pass on CI (this was because of different Slimmer versions locally and on CI)
* Static didn't always load dependencies from govuk_frontend_toolkit when they were referenced from another file – putting them into a single file avoided the problem but the cause wasn't identified (ab888f1)

(Gem changes tested against government-frontend and static)

Goes with https://github.com/alphagov/static/pull/1125


app/assets/stylesheets/govuk_publishing_components/application.scss
```diff
@@ -4,4 +4,176 @@
 @import "typography";
 @import "colours";
 
-@import "component_guide";
+$prism-background: #f5f2f0;
```
h-lame: It's not 100% clear to my why this change is required, but I can shed a little more light on the situation which may help others who know more about sass and sprockets.  If we add a variable to `application.scss` and try to reference it in `component_guide.scss` it doesn't work, so the failure isn't that we're not loading in the frontend toolkit variables properly; it's that we're not sharing any variables at all.  

If we move `component_guide.scss` into a subfolder (e.g. `/app/assets/stylesheets/govuk_publishing_components/whyyyy` and change the import statement accordingly (e.g. `@import "whyyyy/component_guide";`) then the stylesheet does render and can reference the variables imported from govuk_frontend_toolkit or added directly here.

This problem only surfaced when we used this gem in static, and not when we used it in government-frontend, so I suspect it's down to differences in the versions of gems used in the asset-pipeline: sass, sass-rails, sprockets, sprockets-rails, and rails itself.

In either case we have to make a large amount of git diff noise to move the file around so it's not clear which is better.

fofr:
Thanks @h-lame. Merging this to unblock https://github.com/alphagov/static

PR #50, opened 2017-09-06
Update docs to include background context

vanitabarrett:
We have added a dark-background context to the gem, so we need to update the docs.

### DIFF
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -69,9 +69,9 @@ examples:
 
 An example can have an optional description. This is a govspeak markdown block.
 
-#### Right to left examples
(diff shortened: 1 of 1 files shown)


docs/component_conventions.md
```diff
@@ -69,9 +69,9 @@ examples:
 
 An example can have an optional description. This is a govspeak markdown block.
 
-#### Right to left examples
+#### Providing context to examples
 
-A context block can be passed to examples. The only supported context is `right_to_left`:
+A context block can be passed to examples. The guide currently supported `right_to_left` and `dark_background` contexts. For example:
```
nickcolley: currently supports?

PR #52, opened 2017-09-07
Link to principles, conventions in component guide

andysellick:
- adds links to component guide homepage linking to conventions and principles

screen shot 2017-09-07 at 13 41 17

https://trello.com/c/iVZWGnZA/131-1-link-to-principles-and-conventions-from-the-component-guide

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/application.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/application.scss
@@ -12,6 +12,8 @@ $border-color: #ccc;
 }
 
 .component-list {
+  margin-left: 20px;
+
   li {
     @include core-19;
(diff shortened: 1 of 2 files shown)


fofr:
Seems worthwhile to also add a link to the list of all component guides:
https://docs.publishing.service.gov.uk/manual/components.html#component-guides

fofr:
Can we style the component list so the bullets aren't hanging off the side? They look even more odd with the nested bulleted list above them.

andysellick:
I noticed the bullet overhang but left it as outside the scope of this card. But I'll take a look and see how hard it would be.

andysellick:
Added extra link and modified bullet list indent.

PR #53, opened 2017-09-08
Set X-Frame-Options to ALLOWALL

vanitabarrett:
To enable our visual diff tool to work, we need to set X-Frame-Options to ALLOWALL so that the component guide can be embedded into an iframe.


nickcolley:
I wonder if we can add a test for this 🤔

cbaines:
I'm not quite sure what the context is here, but if govuk-puppet is somehow involved, it also sometimes configures NGinx to add the X-Frame-Options header.

1: https://github.com/alphagov/govuk-puppet/search?utf8=%E2%9C%93&q=X-Frame-Options&type=

fofr:
@cbaines We need this available when running review apps on Heroku.

It allows us to overlay a component guide on master with a PR version using iframes and css blend modes, for visual diffing.

PR #56, opened 2017-09-08
Add instructions on how to require the gem

nickcolley:
We load the gem in a specific way to ensure it's not run in production.

Thanks again to @whoojemaflip 🚀

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,29 @@ rails generate govuk_publishing_components:component [component_name]
 
 This will create a template, scss file and yml documentation file for a new component. It will not create a test file as this cannot be reliably done automatically across applications, but a test file will be necessary.
 
+## Integration with Heroku
+To make the best use of the component guide we use Heroku to serve the current `master` build and whenever a [pull request is added](https://devcenter.heroku.com/articles/github-integration-review-apps)
+
+When an app is deployed to Heroku it will be in `production` mode, so only gems that are in the main group will be installed and made available.
+To ensure that we only load the guide on Heroku and not when deployed to _production_ we need to have the gem included in the main bundle group in the Gemfile.
+For this use case we need `require: false` so that it's not loaded in _production_ but then we can manually `require` the gem in our `application.rb` based on the more complex environmental logic that we've specified.
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -29,19 +29,30 @@ The guide includes your application’s `application.scss` and `application.js`
...
 
-Add this line to your application's Gemfile in the [development and test groups](http://bundler.io/v1.12/groups.html#grouping-your-dependencies):
+Add this line to your application's Gemfile in the [development and test groups](http://bundler.io/v1.12/groups.html#grouping-your-dependencies) ([example](https://github.com/alphagov/government-frontend/blob/5110d3d33f7a6b63f218b889a5afec90e6df810f/Gemfile#L11)):
 
 ```ruby
-gem 'govuk_publishing_components'
+gem 'govuk_publishing_components', require: false
 ```
 
+Note: we make sure we do not require this by default so that we only load the guide in a development environment.
```
whoojemaflip: If the guide is only required in the `development` environment, then this gem can be safely added to the `development` group in the Gemfile. In this case there is no need to specify `require: false` as we do, in fact, wish the gem to be loaded by bundler and made available to our application.

This changes when we wish for the guide to be available on Heroku for branch deploys. When an app is deployed to Heroku it will be in `production` mode, so only gems that are in the main group will be installed and made available. To ensure that we only load the guide on Heroku and not when deployed to _production_ we need to have the gem included in the main bundle group in the Gemfile. For this use case we need `require: false` so that it's not loaded in _production_ but then we can manually `require` the gem in our `application.rb` based on the more complex environmental logic that we've specified. 

I think that's all "correct" (logically). Does it make any sense? I think that the guidance on this page possibly doesn't quite encapsulate the entire complicated story.

nickcolley: I thought the same but for the purposes of this thought it'd take up too much space to explain the whole story.

nickcolley: I've updated it slightly to mention heroku, let me know what you think.

fofr: I think the whole story is sensible.

nickcolley: Going to move this into it's own 'integrate with heroku' section thanks

nickcolley:
Kept the default installation simple and made a new section to describe heroku integrations.

fofr:
Top docs 👍

PR #58, opened 2017-09-11
[DO NOT MERGE] Add universal accessibility criteria

andysellick:
- common accessibility criteria defined in yml in the gem can be referenced in yml in components to save copying and pasting and reduce duplication
- universal_accessibility_criteria is a list
- only one defined at the moment, 'link'
- components can include accessibility_criteria, universal_accessibility_criteria, or both

screen shot 2017-09-11 at 12 18 40

screen shot 2017-09-11 at 12 20 50

https://trello.com/c/qS1pfVle/132-2-accessibility-criteria-universal-criteria-for-all-components

### DIFF
--- a/app/models/govuk_publishing_components/component_doc.rb
+++ b/app/models/govuk_publishing_components/component_doc.rb
@@ -5,14 +5,16 @@ module GovukPublishingComponents
                 :description,
                 :body,
                 :accessibility_criteria,
+                :universal_accessibility_criteria,
                 :examples
 
-    def initialize(id, name, description, body, accessibility_criteria, examples)
+    def initialize(id, name, description, body, accessibility_criteria, universal_accessibility_criteria, examples)
       @id = id
       @name = name
       @description = description
       @body = body
       @accessibility_criteria = accessibility_criteria
+      @universal_accessibility_criteria = universal_accessibility_criteria
       @examples = examples
     end
 
@@ -30,6 +32,11 @@ module GovukPublishingComponents
 
     def html_accessibility_criteria
       govspeak_to_html(accessibility_criteria) if accessibility_criteria.present?
+    end
+
+    def html_universal_accessibility_criteria
+      #govspeak_to_html(universal_accessibility_criteria) if universal_accessibility_criteria.present?
+      "universal"
     end
 
     def partial_path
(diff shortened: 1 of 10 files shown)


docs/component_conventions.md
```diff
@@ -46,12 +46,15 @@ name: Name of component
 description: Short description of component
 body: |
   Optional markdown providing further detail about the component
-acceptance_criteria: |
+accessibility_criteria: |
```
fofr: Good catch

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -41,5 +43,26 @@ def partial_path
...
+      doc_files = Rails.root.join(documentation_directory, "*.yml")
+      Dir[doc_files].sort.map { |file| parse_documentation(file) }
+    end
+
+    def parse_documentation(file)
+      { id: File.basename(file, ".yml") }.merge(YAML::load_file(file)).with_indifferent_access
+    end
+
+    def documentation_directory
+      File.join(File.dirname(File.expand_path(__FILE__)), '../../views/govuk_publishing_components/component_guide/docs')
```
fofr: Wondering if the doc file should be in `/docs/universal_accessibility_criteria/`

app/models/govuk_publishing_components/component_doc.rb
```diff
@@ -41,5 +43,26 @@ def partial_path
...
+
+    def check_for_universal_accessibility_criteria
+      fetch_component_docs.map { |criteria| match_universal_accessibility_criteria(criteria) }.join("\n") if universal_accessibility_criteria.present?
+    end
+
+    def match_universal_accessibility_criteria(criteria)
+      criteria["accessibility_criteria"] if universal_accessibility_criteria.include? criteria["id"]
+    end
+
+    def fetch_component_docs
```
fofr: I think we're mixing up concerns a little by having the ComponentDoc be responsible for fetching the shared criteria.

fofr: Seems like something that should be a different class.

fofr: It's also a bit copy-and-paste, these method names are not clear – eg not clearly different to those in the component resolver class.

PR #59, opened 2017-09-11
Add shared accessibility criteria

andysellick:
- common accessibility criteria defined in the gem can be referenced in yml in components to save copying and pasting and reduce duplication
- shared_accessibility_criteria is a list
- only one defined at the moment, 'link'
- components can include accessibility_criteria, shared_accessibility_criteria, or both

screen shot 2017-09-11 at 12 18 40

screen shot 2017-09-12 at 14 59 18

https://trello.com/c/qS1pfVle/132-2-accessibility-criteria-universal-criteria-for-all-components

### DIFF
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -46,12 +46,15 @@ name: Name of component
 description: Short description of component
 body: |
   Optional markdown providing further detail about the component
-acceptance_criteria: |
+accessibility_criteria: |
   Markdown listing what this component must do to be accessible. For example:
 
-  The link must:
+  The banner must:
 
-  * be keyboard focusable
+  - be visually distinct from other content on the page
+  - have an accessible name that describes the banner as a notice
+shared_accessibility_criteria:
+  - link
 examples:
   default:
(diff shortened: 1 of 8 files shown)


docs/component_conventions.md
```diff
@@ -65,6 +68,14 @@ examples:
       some_parameter: 'A different parameter value'
 ```
 
+#### Accessibility Criteria
+
+Markdown listing what this component must do to be accessible.
+
+[Shared accessibility criteria](https://github.com/alphagov/govuk_publishing_components/blob/master/app/models/govuk_publishing_components/shared_accessibility_criteria.rb) can be included in a list as shown. They are pre-written accessibility criteria that can apply to more than one component, created to avoid duplication. For example, links within components should all accept focus, be focusable with a keyboard, etc.
+
+A component can have accessibility criteria, shared accessibility criteria, or both.
```
fofr: Good docs 💯

andysellick: Thanks :)

PR #60, opened 2017-09-11
Add Element#matches polyfill

nickcolley:
This is not something that upsets government-frontend but static throws errors around
not having matches in CI.


fofr:
Does this get included automagically?

nickcolley:
Yeah since it's in the vendor folder.

PR #64, opened 2017-09-15
Make it clear when aXe has failed, succeeded or found errors

fofr:
Easiest to review with `?w=1` to ignore whitespace changes.

Our integration tests in government-frontend couldn't easily check that aXe was present and running on each page. If aXe itself had an error we swallowed the error and CI would report that as "no accessibility errors".

Refactoring to fix this unveiled one of these false positives: a problem with restoreScroll on pages with SVGs – a bug in aXe itself. Added a tweak to prevent this from erroring in the future. See: https://github.com/dequelabs/axe-core/issues/525

Use body class to indicate accessibility tests have run and their state.  Classes can be used in integration tests to confirm tests are running.

See: https://github.com/alphagov/government-frontend/compare/assert-axe-runs

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/accessibility-test.js
+++ b/app/assets/javascripts/govuk_publishing_components/accessibility-test.js
@@ -21,14 +21,14 @@
       }
 
       var violations = (typeof results === 'undefined') ? [] : results.violations
-      var incompleteWarnings = (typeof results === 'undefined') ? [] : results.incomplete
-
-      if (violations.length === 0 && incompleteWarnings.length === 0) {
-        return callback('No accessibility issues found')
-      }
-
-      var incompleteWarningsObj = _processIncompleteWarnings(incompleteWarnings)
       var errorText = _processViolations(violations, results.url)
+
+      var incompleteWarnings = (typeof results === 'undefined') ? [] : results.incomplete
+      var incompleteWarningsObj = _processIncompleteWarnings(incompleteWarnings)
+
+      var bodyClass = violations.length === 0 ? "test-a11y-success" : "test-a11y-failed"
+      document.body.classList.add(bodyClass);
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/accessibility-test.js
```diff
@@ -51,33 +58,36 @@
...
-
-        return {
-          'summary': help,
-          'selectors': cssSelector,
-          'url': helpUrl
-        }
-      })
-    )
+      )
+    } else {
```
nickcolley: We can return early above if we flip the conditional

app/assets/javascripts/govuk_publishing_components/accessibility-test.js
```diff
@@ -7,28 +7,35 @@
...
       }
 
-      var incompleteWarningsObj = _processIncompleteWarnings(incompleteWarnings)
-      var errorText = _processViolations(violations, results.url)
+      var errorText = _processViolations(results.violations, results.url)
+      var incompleteWarningsObj = _processIncompleteWarnings(results.incomplete)
+
+      var bodyClass = results.violations.length === 0 ? "test-a11y-success" : "test-a11y-failed"
+      document.body.classList.add(bodyClass);
+      document.body.classList.add("test-a11y-finished");
```
nickcolley: Rather than this can we assign a global maybe?

```window.__A11Y_TEST_STATUS = 'finished'```

nickcolley: Ah you want to use them in integration tests... If it's not easy to do the above seems fine. Might want to add 'js-' though.

fofr: I think globals will work too. :+1:

PR #61, opened 2017-09-12
Add visual diff JS tool

vanitabarrett:
Add a JS/CSS visual diff tool which allows developers to compare local dev pages to live for visual regressions.
screen shot 2017-09-12 at 12 20 56

The tool relies on mix-blend-mode and filters which aren't supported in IE or Edge. In these cases, the tool simply overlays the live page onto the local page:

screen shot 2017-09-18 at 16 12 46

**Note: the tool relies on X-Frame-Options being set to ALLOWALL on live pages. Therefore the tool may not be able to be run on all GOV.UK pages.**

PR #65, opened 2017-09-15
Remove cookie and survey banners from guide

vanitabarrett:
I have split this into a separate PR to https://github.com/alphagov/govuk_publishing_components/pull/61 so the Visual Diff Tool can be properly tested once this is merged.

This will allow the visual diff tool to correctly compare component guide pages without the banners adding noise to diff. We've set the path to '/component-guide' so the cookie only applies on routes on or below /component-guide.


nickcolley:
can we test this? I'm guessing not since static is not included in our tests?

fofr:
Can we test for cookie headers?

If you visited the component guide then returned to government-frontend, presumably you'd still not see the survey?

Issue #11, opened 2017-08-01
Update syntax highlighting so it passes AA level contrast


nickcolley:
We need to create a prism theme that uses https://github.com/alphagov/tech-docs-template/blob/master/template/source/stylesheets/palette/_syntax-highlighting.scss 's colour palette

PR #70, opened 2017-09-18
Fix syntax highlighting

nickcolley:
screen shot 2017-09-18 at 13 54 53
screen shot 2017-09-18 at 13 55 02
screen shot 2017-09-18 at 13 55 18


fofr:
Does this fix #11?

nickcolley:
@fofr yes, I double checked and updated one of the colours slightly which was alllmost 4.5:1 so that is passes.

PR #71, opened 2017-09-19
Fix for removal of cookie/survey banner in guide

vanitabarrett:
Cookies set in #65 to remove cookie/survey banners in the guide are not always being applied. It may be easier to hide the banners using CSS.


fofr:
Would it be simpler to hide the elements with CSS?

vanitabarrett:
@fofr Updated with CSS fix instead 🙂

PR #76, opened 2017-09-19
Override default styling of cookie banner to hide

vanitabarrett:
The cookie/survey banners have a display: none styling on which overrides the styling in the gem.
Need to add !important flag to force this styling.


nickcolley:
That old chestnut 👍

PR #80, opened 2017-09-20
Disable duplicate ID rule in axe tests

fofr:
Components sometimes contain IDs. These get repeated on the component
guide page. This repetition is not representative of actual usage,
leading to a false positive accessibility failure.

This specifically affects the document_footer component in static where
we rely on a specific component ID to link to from elsewhere in the
page – so it can’t be dynamic and unique with each use.


fofr:
Wondering if we could use markup to disable certain rules per components.

nickcolley:
Yeah if we wanted to do this more often we could have a data api

PR #82, opened 2017-09-22
Render accessibility violations in component guide

fofr:
* Refactor incomplete warnings JS to work with both incomplete and violations
* Improve typography of HTML output - large text and increased spacing between text
* Update to cope better with multiple reasons (eg an image without an alt tag can have many possible solutions)
* Add integration tests for both warnings and violations

screen shot 2017-09-22 at 11 09 06

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/accessibility-test.js
+++ b/app/assets/javascripts/govuk_publishing_components/accessibility-test.js
@@ -35,18 +35,16 @@
         return callback('aXe Error: Expected results but none returned')
       }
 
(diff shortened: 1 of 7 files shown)


vanitabarrett:
To me, the list of reasons/solutions under the violation title is a bit confusing - it makes it look like all of those attributes need to be added for the element to be accessible. Not sure what can be done about that though

fofr:
@vanitabarrett Yeah, that's the list aXe gives you. Anyone of them could be used to fix the issue. I think the top one in the list is usually the most appropriate. Not sure there's much we can do.

PR #86, opened 2017-09-26
Warn if components exist without docs

vanitabarrett:
When running bundle exec rake, if an app uses the govuk_publishing_components gem, it will warn if component files exist (html.erb) without corresponding docs (.yml)

**Example:**
screen shot 2017-09-26 at 11 30 00

### DIFF
--- a/lib/tasks/govuk_publishing_components_tasks.rake
+++ b/lib/tasks/govuk_publishing_components_tasks.rake
@@ -1,4 +1,22 @@
-# desc "Explaining what the task does"
-# task :govuk_publishing_components do
-#   # Task goes here
-# end
+desc 'warns if component view files exist without corresponding documentation'
+task :validate_component_documentation do
+  components_missing_docs = []
+  component_views = Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/**/*.html.erb"]
+
+  component_views.each do |partial|
+    expected_component_docs_file = partial.split('/')[-1].gsub('html.erb', 'yml')
+    expected_component_docs_file.sub!(/^_/, '')
+
+    expected_component_docs_path = "app/views/#{GovukPublishingComponents::Config.component_directory_name}/docs/#{expected_component_docs_file}"
+    components_missing_docs << partial unless File.exist?(expected_component_docs_path)
+  end
+
+  if components_missing_docs.any?
+    error = "You have components which are missing documentation. These components will not be displayed in the component guide:\n"
+    components_missing_docs.each { |component| error += "\t" + component + "\n" }
+    error += "\n"
+    raise NotImplementedError, error
+  end
+end
+
+Rake::Task["validate_component_documentation"].invoke


lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,28 @@
...
-#   # Task goes here
-# end
+desc 'validate component documentation'
+task :validate_component_documentation do
+  components_missing_docs = []
+
+  Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/**/*.html.erb"].each do |view_file|
+    component_doc_match = false
+
+    Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/docs/**/*.yml"].each do |doc_file|
```
fofr: This could be optimised by caching the results of `Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/docs/**/*.yml"]` to avoid looping through all the doc files for each partial.

fofr: You could probably do something like:

```ruby
partials = Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/**/*.html.erb"]

partials.each do |partial|
  expected_documentation = view_file.gsub('.html.erb', '.yml') #or something
  component_missing_docs << view_file unless File.exist?(expected_documentation_file)
end
```

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,28 @@
-# desc "Explaining what the task does"
-# task :govuk_publishing_components do
-#   # Task goes here
-# end
+desc 'validate component documentation'
```
fofr: Might want to describe what this task is doing here

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,28 @@
...
+      end
+    end
+
+    if !component_doc_match
+      components_missing_docs.push(view_file.split('/')[-1].gsub('.html.erb', ""))
+    end
+  end
+
+  unless components_missing_docs.empty?
+    puts "\n\033[33mYou have components which are missing documentation. These components will not be displayed in the component guide:\033[0m\n"
```
fofr: Not sure if we need the colour output

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,28 @@
...
+    components_missing_docs.each { |component| puts "\t \033[33m" + component + "\033[0m\n" }
```
fofr: Wonder if this should be able to fail a build

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,20 @@
...
+  components_missing_docs = []
+  component_views = Dir["app/views/#{GovukPublishingComponents::Config.component_directory_name}/**/*.html.erb"]
+
+  component_views.each do |partial|
+    expected_component_docs_file = partial.split('/')[-1].gsub('html.erb', 'yml').sub("_", "")
+    expected_component_docs_path = "app/views/#{GovukPublishingComponents::Config.component_directory_name}/docs/" + expected_component_docs_file
+    components_missing_docs << partial unless File.exist?(expected_component_docs_path)
+  end
+
+  unless components_missing_docs.empty?
```
fofr: I'd use the positive assertion here:
`if components_missing_docs.any?`

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,4 +1,20 @@
...
+    error = "\nYou have components which are missing documentation. These components will not be displayed in the component guide:\n"
+    components_missing_docs.each { |component| error += "\t" + component + "\n" }
+    error += "\n"
+    raise error
```
fofr: Maybe this could be a NotImplementedError with that message?
https://ruby-doc.org/core-2.4.0/NotImplementedError.html

eg https://github.com/alphagov/government-frontend/blob/master/test/component_test_helper.rb#L7

PR #90, opened 2017-10-12
Accessibility warning

maxgds:
Add violation warning when accessibility criteria aren't supplied (fully linted). @vanitabarrett @fofr

screen shot 2017-10-12 at 09 50 34

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/application.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/application.scss
@@ -46,6 +46,23 @@ $border-color: #ccc;
   }
 }
 
+.component-violation {
+  border: 3px solid $error-colour;
+  margin: 0 0 $gutter;
+  padding: $gutter $gutter;
+
+  .component-violation--title {
+    @include bold-24;
+  }
+
+  .component-violation--link {
+    display: block;
+    @include bold-19;
+    color: $error-colour;
+    margin: $gutter * 0.5 0;
+  }
+}
+
 .component-doc-h2 {
   @include bold-27;
   margin: ($gutter * 1.5) 0 $gutter;
--- a/app/assets/stylesheets/govuk_publishing_components/application.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/application.scss
@@ -51,15 +51,15 @@ $border-color: #ccc;
   margin: 0 0 $gutter;
   padding: $gutter $gutter;
 
-  .component-violation--title {
+  .component-violation__title {
     @include bold-24;
   }
 
-  .component-violation--link {
+  .component-violation__link {
     display: block;
     @include bold-19;
     color: $error-colour;
-    margin: $gutter * 0.5 0;
+    margin: $gutter-half 0;
   }
 }
 
--- a/app/views/govuk_publishing_components/component_guide/show.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/show.html.erb
@@ -2,20 +2,14 @@
 <%= render 'govuk_component/title', title: @component_doc.name, context: 'Component' %>
 
 <div class="component-show">
-
-  <% if !@component_doc.accessibility_criteria.present? %>
-    <div class="grid-row">
-      <div class="column-two-thirds">
-        <div class="component-violation">
-          <h2 class="component-violation--title">This component is not valid</h2>
-          <p><a class="component-violation--link" href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/accessibility_acceptance_criteria.md">Please define accessibility acceptance criteria for this component.</a></p>
-        </div>
-      </div>
-    </div>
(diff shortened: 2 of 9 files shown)


maxgds:
@fofr & @vanitabarrett - sorry about all the noise - this should finally be ready to go, pending comments from you.

app/assets/stylesheets/govuk_publishing_components/application.scss
```diff
@@ -46,6 +46,23 @@ $border-color: #ccc;
   }
 }
 
+.component-violation {
+  border: 3px solid $error-colour;
+  margin: 0 0 $gutter;
+  padding: $gutter $gutter;
+
+  .component-violation--title {
```
vanitabarrett: I think this should be using underscores instead of double dash, e.g: component-violation__title. -- is used for [modifiers](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#bem)

app/assets/stylesheets/govuk_publishing_components/application.scss
```diff
@@ -46,6 +46,23 @@ $border-color: #ccc;
...
+    @include bold-24;
+  }
+
+  .component-violation--link {
+    display: block;
+    @include bold-19;
+    color: $error-colour;
+    margin: $gutter * 0.5 0;
```
vanitabarrett: Can we use $gutter-half here?

app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -2,6 +2,18 @@
 <%= render 'govuk_component/title', title: @component_doc.name, context: 'Component' %>
 
 <div class="component-show">
+
+  <% if !@component_doc.accessibility_criteria.present? %>
+    <div class="grid-row">
+      <div class="column-two-thirds">
```
vanitabarrett: Is there a reason why we are using this class - we could just remove this and make the warning banner full width to fit with our other warning/violation styles?

lib/tasks/govuk_publishing_components_tasks.rake
```diff
@@ -1,22 +1,50 @@
...
+    puts "✔︎"
   end
 
-  if components_missing_docs.any?
-    error = "You have components which are missing documentation. These components will not be displayed in the component guide:\n"
-    components_missing_docs.each { |component| error += "\t" + component + "\n" }
-    error += "\n"
-    raise NotImplementedError, error
+  desc 'warns if component view files exist without corresponding documentation or accessibility critera'
+  task :validate_documentation do
```
vanitabarrett: Just making note of the conversation we just had - this means the tasks are no longer run automatically.

app/views/govuk_publishing_components/component_guide/show.html.erb
maxgds: @fofr - can you advise? I put this in because the following grid row (the component description) has it. When I strip it out from both the description stays at less than full width because of other styles applied to it. It looks a bit odd if the warning is wide and the description is narrow: 

screen shot 2017-10-12 at 10 47 40

app/assets/stylesheets/govuk_publishing_components/application.scss
maxgds: Fixed

app/assets/stylesheets/govuk_publishing_components/application.scss
maxgds: Yep, fixed.

lib/tasks/govuk_publishing_components_tasks.rake
maxgds: Reinstated so they'll now run as default tasks again.

app/views/govuk_publishing_components/component_guide/show.html.erb
fofr: I'd suggest moving the notice into the existing grid row and cell used by the description.

lib/tasks/govuk_publishing_components_tasks.rake
fofr: I think there's some risk running this automatically – it makes it hard for apps to opt out. eg Upgrading to this version on static will require the person upgrading to add accessibility criteria for all legacy components.

If we do opt-in we might want to use an install script to encourage more apps to use default component validation tasks, or we could add to the default Jenkinsfile:
https://github.com/alphagov/govuk_publishing_components/issues/91

vanitabarrett: Maybe for now we can revert it back to how it was (a task that needs running manually), and do the manual step to include it in the apps we want it running in. Then the install script can be done as a separate story.

maxgds:
@fofr I think I've covered everything now and it's green, can you give it a final once over?

PR #101, opened 2017-11-27
Allow excluded rules to be specified

nickcolley:
When trying to add a radio button component to government-frontend I was hitting a rule since they're not in a fieldset, I want to be able to remove this.

I've made it so you can see the id you need to use in the error/frontend, and allowed you to use that id to exclude rules.

Adds to the original test that was making sure the `duplicate-id` rule doesnt throw an error.

https://trello.com/c/yo9dWRM8/242-create-a-radio-button-component

### DIFF
--- a/README.md
+++ b/README.md
@@ -112,6 +112,15 @@ class ComponentGuideTest < ActionDispatch::IntegrationTest
 end
 ```
 
+### Exclude accessibility rules
+
+Sometimes you will have a component that will throw an error due to it being in isolation, for example radio buttons not being in a fieldset.
+
+For this case you can add `accessibility_excluded_rules` to your components' documentation yml file with the rules you want to exclude. These rules can be found in brackets in the error messages displayed.
+
+For an example of this check [test/.../docs/test-component-with-duplicate-ids.yml](test/dummy/app/views/components/docs/test-component-with-duplicate-ids.yml)
+
+
 ## Component generator
 
 The gem includes a component generator to stub the minimal files required for a component.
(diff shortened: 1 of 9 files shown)


README.md
```diff
@@ -101,6 +101,12 @@ The gem [includes integration tests](lib/govuk_publishing_components/minitest/co
 
 Automated accessibility tests use [aXe](https://github.com/dequelabs/axe-core). Using our `AccessibilityTest` wrapper the gem runs the aXe suite against each example and throws JavaScript errors for any violations. These JavaScript errors can be used to fail a build in CI.
 
+### Exclude accessibility rules
+
+Sometimes you will have a component that will through an error due to it being in isolation, for example radio buttons not being in a fieldset.
```
maxgds: that will **throw**

test/dummy/app/views/components/docs/test-component-with-duplicate-ids.yml
```diff
@@ -1,4 +1,7 @@
 name: Test invalid component with duplicate IDs
```
maxgds: Should this mention the image too? You're testing both it seems?

README.md
emmabeynon: Should it be "throw" instead of "through"?

nickcolley: brain fart cheers :)

nickcolley: brain fart cheers :)

test/dummy/app/views/components/docs/test-component-with-duplicate-ids.yml
nickcolley: Makes sense, will update

PR #105, opened 2017-12-13
Update gem to support components

andysellick:
Update the component guide gem so it can contain components. This work is being done in order to make deployment of changes to these components easier.

This PR includes three task list components copied from static. They are all identical to how they were in static (i.e. those commits shouldn't need reviewing) with the exception that their CSS namespaces have been changed from pub-c- to gem-c-.

Gem components appear in the component guide and are available to applications that have the gem installed. Example screenshot from the updated component guide:

screen shot 2017-12-13 at 15 04 53

Trello card: https://trello.com/c/ELorPDMS/393-do-all-the-task-list-gem-component-things

Related PR to remove same components: https://github.com/alphagov/static/pull/1221


nickcolley:
👍 for the namespace, be good to update the docs to explain this.

nickcolley:
This is really exciting, however we need to make sure we look forward and ensure if we move more components into the gem that we have the correct measurements, we will be duplicating a lot of files for our users to download and use.

We should 'do the hard work' to try to minimise this, but first we need to know what the impact is.

Does not block this PR of course.

nickcolley:
I think it's important that the guide communicates that these components are not in the host applications' source, this could be a label or something.

I've already had some people ask me 'where is the source for these components' and while it's not this PRs responsibility to solve that, as is it'll be very confusing to see something in the guide that you can't trace back to the application.

andysellick:
@nickcolley agreed on both points. Another thing I want to look into is a way of using a component from the gem without having to explicitly call the styles and JS for that component in the app, which is what you have to do at the moment. Got a small problem with pulling in print styles related to this, collections has no print styles so currently no way to include them for the gem components.

app/assets/javascripts/components/task-list.js
```diff
@@ -4,7 +4,7 @@
   "use strict";
   window.GOVUK = window.GOVUK || {};
 
-  Modules.Tasklist = function () {
+  Modules.Gemtasklist = function () {
```
nickcolley: Does this need to be namespaced?

app/assets/javascripts/govuk_publishing_components/application.js
```diff
@@ -10,5 +10,9 @@
 // Read Sprockets README (https://github.com/rails/sprockets#sprockets-directives) for details
 // about supported directives.
 //
+//= require ../history-support
```
sihugh: Do these requires need to be in this order for any particular reason?

fofr:
😍 :100:

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -11,4 +11,5 @@
 # Prevent PhantomJS auto install, uses PhantomJS already on your path.
 Jasmine.configure do |config|
   config.prevent_phantom_js_auto_install = true
+  config.show_console_log = false
```
fofr: Any reason for this change?

sihugh: Yeah, we added the config with a `true` to help debugging some stuff, and then disabled it (rather than removing it).

PR #108, opened 2017-12-18
Fix task list appearance

andysellick:
A few visual-related changes:

- iphone was having difficulty with the white background on the inner element of the numbers (added so numbers are still legible when text only is zoomed) where it was overflowing slightly and corrupting the outer circle. This white background has now been replaced with a similar effect achieved with text shadow.
- overflow hidden removed from overall component to prevent clipping of beginning of logic text ('or', 'and') in some steps, which is aligned to grow to the left when text only is zoomed, to prevent overlap onto the step titles.
- overflow hidden was previously added to solve a problem with iphone 4 where the final part of the line down the side of the task list 'escaped' the component (it was ignoring the calc'ed height). Now overflow hidden is removed, have added a second height declaration as a fallback for older iphones to fix this escaping line issue.
- miscellaneous print style improvements, spacing, list styling

### DIFF
--- a/app/assets/stylesheets/components/_task-list.scss
+++ b/app/assets/stylesheets/components/_task-list.scss
@@ -28,7 +28,6 @@ $top-border: solid 2px $grey-3;
 
 .gem-c-task-list {
   margin-bottom: $gutter-half;
-  overflow: hidden;
 }
 
 .gem-c-task-list__controls {
@@ -106,7 +105,10 @@ $top-border: solid 2px $grey-3;
 
   .gem-c-task-list__step:last-child {
     &:after {
+      // scss-lint:disable DuplicateProperty
+      height: -webkit-calc(100% - #{$gutter-half}); // fallback
       height: calc(100% - #{$gutter-half});
+      // scss-lint:enable DuplicateProperty
     }
 
     .gem-c-task-list__help:after {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/components/_task-list.scss
```diff
@@ -104,15 +103,18 @@ $top-border: solid 2px $grey-3;
     border-bottom: solid $stroke-width $grey-2;
   }
 
+  // scss-lint:disable DuplicateProperty
   .gem-c-task-list__step:last-child {
     &:after {
+      height: -webkit-calc(100% - #{$gutter-half}); // fix for iphone 4s
```
nickcolley: This is more than just iPhone 4s right? Could we just say 'fallback'?

Could we do the disabling only on the specific line.

andysellick: HOW DO YOU SPOT THESE PRS SO QUICKLY

andysellick: Sure, will change.

nickcolley:
Could you split the accessibility and iOS fixes into two commits and add good detail about the two so future people know when they can remove your fallbacks etc.

andysellick:
@nickcolley happy to add more detail but the two changes you mention are linked (so yeah, that probably needed more explanation). Overflow hidden was removed to fix the zoom text problem (it was cutting off the beginning of the logic words, i.e. 'and', 'or') but that was originally added to fix a problem with iphone 4, which is then fixed with the second height line. It's... confusing.

andysellick:
@nickcolley more detail added to PR.

PR #110, opened 2017-12-18
Fix task list active link

andysellick:
Fix active styles for choice lists (no before image but the big dot wasn't on the line):

screen shot 2017-12-18 at 16 11 29

Fix outline styles for Firefox, before:

screen shot 2017-12-18 at 16 12 35

After:

screen shot 2017-12-18 at 16 25 34

Tested in IE8+, Edge 14+, latest Chrome, latest Firefox.

### DIFF
--- a/app/assets/stylesheets/components/_task-list.scss
+++ b/app/assets/stylesheets/components/_task-list.scss
@@ -377,6 +377,10 @@ $top-border: solid 2px $grey-3;
     color: $black;
   }
 
+  &:focus {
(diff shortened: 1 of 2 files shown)


nickcolley:
Could this be fixed with some form of `inline-block`? magic numbers look scary to me

nickcolley:
Code looks very complicated but I believe this is due to the design, would need to push back on the design to make this meaningfully simpler.

PR #113, opened 2018-01-03
Rework get help link

andysellick:
Change how the 'get help with this step' link works, specifically:

- restyle the link to be bold so it stands out more (users missed it in lab)
- add a 'heading' element into the task list content, works like a paragraph but outputs a heading (needed for 'get help' section in main task lists)
- change get help link functionality to be a per-step option, not a global (all steps) option, necessary because only some steps need a get help link

Get help section (showing new heading element):

screen shot 2018-01-03 at 09 29 31

Get help link (now bold):

screen shot 2018-01-03 at 09 30 29

- [Trello card](https://trello.com/c/0z6EzxSr/403-add-the-get-help-link-to-the-small-task-list-component-and-the-task-list-page)

### DIFF
--- a/app/views/components/_task_list.html.erb
+++ b/app/views/components/_task_list.html.erb
@@ -69,6 +69,9 @@
                   <% if element[:type] == 'paragraph' %>
                     <p class="gem-c-task-list__paragraph"><%= element[:text] %></p>
 
+                  <% elsif element[:type] == 'heading' %>
+                    <%= content_tag("h#{heading_level + 1}", element[:text], class: 'gem-c-task-list__heading') %>
+
                   <% elsif element[:type] == 'list' %>
                     <%
                       list_style = ''
--- a/app/views/components/_task_list.html.erb
+++ b/app/views/components/_task_list.html.erb
@@ -130,7 +130,7 @@
                   </div>
                 <% end %>
 
-                <% if task_list_url %>
+                <% if task_list_url && step[:show_help_link] %>
                   <div class="gem-c-task-list__help">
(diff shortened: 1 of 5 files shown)


vanitabarrett:
@andysellick think @fofr just fixed the Uglifier issue in https://github.com/alphagov/govuk_publishing_components/pull/114#event-1407752763  - so may just need a rebase to fix?

andysellick:
@vanitabarrett brilliant, thanks. Had just spotted that line and was about to try that fix :)

app/views/components/_task_list.html.erb
```diff
@@ -127,7 +130,7 @@
                   </div>
                 <% end %>
 
-                <% if task_list_url %>
+                <% if task_list_url and step[:show_help_link] %>
```
fofr: We use `&&` rather than `and`

fofr:
We should set up the dummy app hosted on Heroku so we can see the components in the gem within a review app.

fofr:
Could you include the "Why" in your PR please?

app/views/components/_task_list.html.erb
andysellick: Argh, I knew we preferred one over the other but couldn't remember which one.

fofr: If we linted our ERB it'd tell you.

andysellick: Yeah, that'd be helpful.

andysellick:
@fofr I agree that's it's too complex. I'm going to look at that in the next quarter, including taking advantage of partials and helpers now it's in the gem. I wondered about breaking it up into smaller components but I'm not sure that would help or even be practical (internal elements rely upon their parents for styling and some would look weird out of context, so would be unlikely to be reused).

Would be interested in any suggestions anyone has.

PR #114, opened 2018-01-03
Sort components by name

fofr:
Components were being ordered by their filename but displayed their name in the component list. For some components such as “Form label”
(label.html.erb), the component wasn’t appearing where expected.

* Always sort the list of components by name

## Before

screen shot 2018-01-03 at 11 35 17

## After

screen shot 2018-01-03 at 11 35 05

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/visual-regression.js
+++ b/app/assets/javascripts/govuk_publishing_components/visual-regression.js
@@ -2,7 +2,7 @@
   window.GOVUK = window.GOVUK || {}
 
   window.GOVUK.VisualDiffTool = function (currentWindowLocation) {
-    const visualDiffSelector = 'visual-diff';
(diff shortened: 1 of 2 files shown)


fofr:
Weird error on this, unrelated to the change:

```
Uglifier::Error: Unexpected token: keyword (const). To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

Guessing it relates to:

https://github.com/alphagov/govuk_publishing_components/blame/order-by-name/app/assets/javascripts/govuk_publishing_components/visual-regression.js#L5

cc @nickcolley @vanitabarrett

nickcolley:
That is odd it wasn't raised before hand, we should be OK to change that to `var` for now.

nickcolley:
🚢 👍

Issue #98, opened 2017-11-07
Allow blocks in components


nickcolley:
https://github.com/alphagov/collections/pull/402#issuecomment-341673555

fofr:
Added in #117

PR #117, opened 2018-01-03
Support components accepting blocks

steventux:
ERB's render method accepts a block, which is useful for components which
embed multiple components for a specific purpose (eg. a content list with
a back-to-top link).
The component guide should be able to render the docs for this type of component
so this PR adds the ability to configure the component guide with a specific example data key:

```yaml
examples:
  default:
    data:
      block: |
         <div><%= some_local_thing %></div>
```

### DIFF
--- a/spec/dummy/app/views/components/docs/test-component-with-block.yml
+++ b/spec/dummy/app/views/components/docs/test-component-with-block.yml
@@ -1,0 +1,7 @@
+name: Test component with block
+description: A test component that takes a block
+examples:
+  default:
+    data:
+      block: |
+        <div class="test-block">Here's a block</div>
--- a/app/models/govuk_publishing_components/component_example.rb
+++ b/app/models/govuk_publishing_components/component_example.rb
@@ -5,13 +5,15 @@ module GovukPublishingComponents
     attr_reader :id,
                 :data,
                 :context,
-                :description
+                :description,
+                :block
 
     def initialize(id, data, context, description)
       @id = id
       @data = data || {}
       @context = context || {}
       @description = description || false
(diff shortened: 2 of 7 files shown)


spec/dummy/app/views/components/docs/test-component-with-block.yml
```diff
@@ -0,0 +1,7 @@
+name: Test component with block
+description: A test component that takes a block
+examples:
+  default:
+    data:
+      block: |
```
nickcolley: Block is a good shout, but what do you think about naming it something like 'children'?

If you understand blocks this makes sense, but children may make more sense for people who do not in the context of interfaces.

Not sure, what do you think?

fofr: I prefer `block` or `yields`.

fofr:
Could you update the documentation to mention this feature?

nickcolley:
Closes https://github.com/alphagov/govuk_publishing_components/issues/98

cc @andysellick

spec/dummy/app/views/components/docs/test-component-with-block.yml
steventux: I'm not 100% sold on `children`, we're leveraging the `ActionView::Partial#render` method's ability to accept a block and yield it. I agree though, that with little ruby experience the name `block` is not intuitive.

steventux:
@fofr https://github.com/alphagov/govuk_publishing_components/pull/117/commits/85d6a43abbe6d4cb5ad6bac5ca1517c91857463c adds some documentation of this feature

PR #128, opened 2018-01-05
Fix radio component not including welsh translation

nickcolley:
This was missed when moving this component upstream from `government-frontend`

See https://github.com/alphagov/government-frontend/pull/671#discussion_r159884811


fofr:
Reminds me of a feature I've wanted – the ability to show component fixtures against different locales

PR #136, opened 2018-01-09
Namespace components hosted in gem

fofr:
Move all components hosted by the gem to be within `govuk_publishing_components`.

Usage in app is now like:
```
<%= render "govuk_publishing_components/components/radio", … %>
```

| Old | New |
|--|--|
| `assets/components/*` | `assets/govuk_publishing_components/components/*` |
| `javascripts/components/*` | `javascripts/govuk_publishing_components/components/*` |
| `views/components/*` | `views/govuk_publishing_components/components/*` |

## The problem

Before namespacing, in an app you would see something like `render "components/blah"` which you'd expect to be in `app/views/components`. When you can't find it there, it's not clear where to find the source.

Equally it led to peculiar includes like:
```
@import 'components/*';
@import 'components/task-list';
@import 'components/task-list-header';
@import 'components/task-list-related';
```

If an app and the gem both have a component with the same name, it is not clear which one would have precedence. By setting a namespace it's explicit.

This is a breaking change


fofr:
Example of updating government-frontend after this is versioned:
https://github.com/alphagov/government-frontend/compare/example-gem-update

I found that I needed to explicitly import the css3 mixin for box sizing in 607b10b

nickcolley:
When using this namespace practically is it too verbose?

fofr:
@nickcolley I think the confusion of using a custom namespace would be worse than the problem of length with this one

fofr:
An alternative to the long namespace would be to have `govuk_publishing_components/back_link` rather than `govuk_publishing_components/components/back_link`.

I avoided that in this PR because I thought it would make it unclear what was needed for components and what was needed for the component guide.

nickcolley:
I can't block this but I think having such long verbose names will be annoying for people, since it'll make their markup much harder to read.

PR #138, opened 2018-01-10
Extract the error_summary component from government-frontend

tuzz:
We’d like to use this in email-alert-frontend

https://trello.com/c/HSre7K10/453-style-the-mvp-email-collection-page-needed-by-jan

Screenshot demonstrating the focus on page load:

screen shot 2018-01-10 at 10 45 50

### DIFF
--- a/spec/javascripts/components/error-summary-spec.js
+++ b/spec/javascripts/components/error-summary-spec.js
@@ -1,0 +1,72 @@
+/* global describe afterEach beforeEach it expect */
+
+var $ = window.jQuery
+
+function isFocused (element) {
+  return $(element)[0] === $(element)[0].ownerDocument.activeElement
+}
+
+describe('An error summary component', function () {
+  'use strict'
+
+  var GOVUK = window.GOVUK
+  var module
+  var element
+
+  beforeEach(function () {
+    element = $(
+      '<div ' +
+        'class="gem-c-error-summary" ' +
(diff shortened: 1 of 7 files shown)


spec/javascripts/components/error-summary-spec.js
```diff
@@ -0,0 +1,72 @@
...
+  'use strict'
+
+  var GOVUK = window.GOVUK
+  var module
+  var element
+
+  beforeEach(function () {
+    element = $(
+      '<div ' +
+        'class="app-c-error-summary" ' +
```
nickcolley: These tests should be failing

nickcolley: Actually nevermind, it'd be nice to be consistent but they are just JavaScript so make sense to be passing.

fofr: Everything other than the old `app-c` stuff here looks good :+1:

tuzz: Fixed.

The JavaScript tests weren't failing because the module is bound to the `js-error-summary__link` and isn't `app-` or `gem-` dependent.

We should use `gem-` here anyway for consistency.

nickcolley: Good example of not binding to presentational classes 👍

PR #146, opened 2018-01-23
Add requirements for Heroku review apps

tijmenb:
We'll need a Gemfile.lock and app.json to use review apps in this repo.


fofr:
Is it worth indicating somewhere why the Gemfile.lock is included against the best practice guidance?

tijmenb:
It's in the commit message, probably enough?

Issue #139, opened 2018-01-10
Remove hosted component styles from component guide stylesheet


fofr:
Applications that want to use components from this gem need to include the styles. When included they'll be made available to the app's component guide.

By including them in the component guide automatically we are giving a false impression that they'll work as intended in the app. As we use component guides to review PRs this feels like a false positive that could lead to broken deploys.

We should remove hosted component styles from the component guide stylesheet.

We include them here:
https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/application.scss#L7-L13

Example problem here:
https://github.com/alphagov/government-frontend/pull/672

nickcolley:
This makes sense to me.

PR #148, opened 2018-01-23
Remove doubly included stylesheets

tijmenb:
The component CSS is already included into the component guide, because the component guide pulls in the main CSS of the application (which includes the components if configured correctly).

- To make sure that new components get added to the dummy app, 2 import files are now provided that include all the styles. Applications can also use this if they don't want to import all the components individually.
- The `application.css` of the mountable component guide is renamed to `component_guide.css` so as not to confuse it with the application.css of the application it will be mounted in.

This isn't in the README yet because I'm not sure where to place it. We'll look at restructuring the installation instructions for the gem.

Closes https://github.com/alphagov/govuk_publishing_components/issues/139.


nickcolley:
We should make sure to not run this gem in production, as was removed in government-frontend

https://github.com/alphagov/government-frontend/commit/4b0e39570e6401eefcc76c6b815017a396118256 here

tijmenb:
Thanks @fofr, I've redone the PR. We now expose a single CSS file for apps to include, which is easier to keep in sync.

@nickcolley do you think it's a problem that the component guide is mounted in production?

nickcolley:
@tijmenb not immediately but we tried to stop this from happening to manage risk, so just a heads up.

tijmenb:
Turns out this doesn't actually work yet, there's some shadiness going on with the imports.

fofr:
@tijmenb Ready for review again or WIP?

tijmenb:
@fofr still trying to make this work (it doesn't at the moment)

tijmenb:
This now works! Updated PR description as well.

nickcolley:
Great work, 🙌 🚀

CHANGELOG.md
```diff
@@ -1,5 +1,19 @@
 # Unreleased
 
+* To include the CSS for all components in the gem, you can now do:
```
fofr: Should this also be in the README?

tijmenb: Yeah, but I don't know where yet because it's already so large. I have a branch to rewrite the README, would you mind if I do it there?

PR #154, opened 2018-01-31
Remove groups from component

andysellick:
# Note this PR contains breaking changes

To simplify the task list component code we are removing the concept of groups of steps. Instead, the task list data will simply be a series of steps, some of which can be configured to have the 'and' or 'or' appearance.

This means removing one of the nested arrays in the data structure, so that a task list is now defined by an array of objects, not an array of an array of objects.

i.e.

`
steps: [
   {
      title: 'A step'
   }
]
`

instead of

`
groups: [
   [
      {
         title: 'A step'
      }
   ]
]
`

**Breaking changes**

- data passed to component as 'groups' is now 'steps'
- steps that should be marked as 'and' instead of being numbered must now explicitly have the logic parameter, otherwise they'll just be given numbers
- any other references to groups such as 'highlight_group' are now steps, e.g. 'highlight_step'

Trello card: https://trello.com/c/G7kq47Db/467-remove-groups-from-component

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_task-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_task-list.scss
@@ -19,12 +19,12 @@ $top-border: solid 2px $grey-3;
 }
 
 @mixin task-list-line-position {
-  left: -($gutter + $gutter-half);
+  left: 0;
   margin-left: ($number-circle-size / 2) - ($stroke-width / 2);
 }
 
 @mixin task-list-line-position-large {
-  left: -($gutter * 2);
+  left: 0;
   margin-left: ($number-circle-size-large / 2) - ($stroke-width-large / 2);
   border-width: $stroke-width-large;
 }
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_task-list.scss
```diff
@@ -106,17 +133,15 @@ $top-border: solid 2px $grey-3;
...
-  .gem-c-task-list__step:last-child {
-    &:after {
-      // scss-lint:disable DuplicateProperty
-      height: -webkit-calc(100% - #{$gutter-half}); // fallback
-      height: calc(100% - #{$gutter-half});
-      // scss-lint:enable DuplicateProperty
-    }
+  &:after {
+    // scss-lint:disable DuplicateProperty
+    height: -webkit-calc(100% - #{$gutter-half}); // fallback
```
vanitabarrett: Is this required? We use `calc` in other places without also including `-webkit-calc`?

app/views/govuk_publishing_components/components/_task_list.html.erb
```diff
@@ -2,102 +2,97 @@
...
+  >
+    <ol class="gem-c-task-list__steps">
+      <% steps.each_with_index do |step, step_index| %>
         <%
-          group_is_active = group_index + 1 == highlight_group
+          step_is_active = step_index + 1 == highlight_step
+          step_count += 1
+          id = TaskListHelper.new.generate_task_list_id(step[:title])
+
+          logic = false
```
vanitabarrett: This may need renaming - it's not obvious what `logic` is referring to, and it's difficult to get any kind of context when it's used, e.g:

```
<% if logic %>
    <%= logic %>
```

app/views/govuk_publishing_components/components/docs/task_list.yml
```diff
@@ -1,7 +1,7 @@
 name: Task list
-description: Numbered groups containing expanding/collapsing steps
+description: A series of expanding/collaping steps designed to navigate a user through a process
```
vanitabarrett: Typo, should be "collapsing"

app/assets/stylesheets/govuk_publishing_components/components/_task-list.scss
nickcolley: https://caniuse.com/#feat=calc doesn't look like we need it.

See also https://github.com/postcss/autoprefixer which uses caniuse data with a browser list to automatically figure this out for you.

andysellick: I should have put a comment in, will add. This is to correct a bug on iphone 4, which doesn't seem to obey calc properly. @nickcolley is there an alternative way of writing this out with autoprefixer? This was the only thing that would work.

app/views/govuk_publishing_components/components/_task_list.html.erb
andysellick: I'll make a note to look at this in a separate PR, but this is outside the scope of this change. Good point though.

sihugh: I'm OK with `logic` for the moment, as we've been trying to think of something else that describes it better and haven't quite managed it yet.  

Our schema (https://github.com/alphagov/govuk-content-schemas/blob/25c0bc646225dfde0389861d1f62bb82bfb498ac/formats/shared/definitions/_step_by_step_nav.jsonnet#L29-L35) has it limited to `and` or `or` and it's used to interrupt the usual numbering sequence to indicate a concurrent step.

vanitabarrett: @sihugh Sure, I don't think it needs to block this PR, but I think it's worth changing in future as it's not very easily understandable if other people come to work on this.

PR #155, opened 2018-01-31
Fix RSpec tests in CI

sihugh:
It appears that RSpec is not available until it's been required.  This was causing rspec tests to not be run as part of `bundle exec rake` or `bundle exec rake spec`.

(I'm not sure whether this fixes the problem it was originally changed to resolve.)

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,6 +1,8 @@
-if defined?(RSpec)
+begin
   require 'rspec/core/rake_task'
   RSpec::Core::RakeTask.new(:spec)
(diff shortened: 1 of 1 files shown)


sihugh:
@andysellick spotted it, I got a chance to wield `git bisect` for about 30sec. :-)

tijmenb:
Nice one.

For the record, this is solving the same problem intended. When running on Heroku, the `rspec` gem isn't loaded because it's a development dependency in the gemfile.

PR #156, opened 2018-02-01
Rename task list

andysellick:
# Note this PR contains breaking changes

This PR depends upon https://github.com/alphagov/govuk_publishing_components/pull/154 being merged first, as it has been branched from that branch.

Rename 'task list' to 'step by step navigation'. Includes:

- renaming the files for each of the three task list components (separate PRs will be created in each app to account for the changes to the render call for these components)
- updating class names and JS
- updating documentation

See Trello card for specific naming conventions: https://trello.com/c/7eVDky11/460-rename-task-list-components


nickcolley:
👀 interested to know if the component system means this'll be much easier to roll out to the apps

andysellick:
@nickcolley well in this case probably not, as this change means we'll have to update all the calls to the components in all the apps that use them, so I'll be raising a ton of other PRs to follow this one...

sihugh:
@nickcolley had we done this in static, we'd have had to write code in all the apps to pick up either version, or duplicated it all in static with the new name, and then deprecated it.  So this way is not easy, but probably _easier_ than the alternative. :-/

sihugh:
@andysellick looks like we've got some merge conflicts

andysellick:
@sihugh on it

PR #160, opened 2018-02-08
Step nav accessibility changes

andysellick:
A variety of changes relating to accessibility following feedback from our trip to DAC.

Also includes a small change to the analytics, removing a redundant tracking call.

Trello cards: 

- https://trello.com/c/t2UnqDX0/461-fix-agreed-accessibility-issues-with-step-by-step-navigation
- https://trello.com/c/tc1P7cuE/450-minor-change-to-tracking-on-step-by-step-remove-tasklistlinkclicked

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -387,18 +387,6 @@
       function trackClick() {
         var tracking_options = {label: trackingLabel(), dimension28: stepView.numberOfContentItems().toString()}
         stepNavTracker.track('pageElementInteraction', trackingAction(), tracking_options);
-
-        if (!stepView.isHidden()) {
-          stepNavTracker.track(
-            'stepNavLinkClicked',
-            String(stepIndex()),
-            {
-              label: stepView.href,
-              dimension28: String(stepView.numberOfContentItems()),
-              dimension29: stepView.title
-            }
-          )
-        }
       }
 
       function trackingLabel() {
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -95,7 +95,7 @@
             linkText = actions.hideLinkText;
           }
           if (!$(this).find('.js-toggle-link').length) {
-            $(this).append('<span class="gem-c-step-nav__toggle-link js-toggle-link" aria-hidden="true">' + linkText + '</span>');
+            $(this).find('.js-step-title-button').append('<span class="gem-c-step-nav__toggle-link js-toggle-link" aria-hidden="true">' + linkText + '</span>');
           }
         });
       }
@@ -154,6 +154,10 @@
           var $step = $(this);
           var $title = $step.find('.js-step-title');
           var contentId = $step.find('.js-panel').first().attr('id');
+
+          $title.wrapInner(
+            '<span class="js-step-title-text"></span>'
+            );
 
           $title.wrapInner(
             '<button ' +
@@ -308,8 +312,7 @@
       var $stepContent = $stepElement.find('.js-panel');
       var shouldUpdateHash = rememberShownStep;
 
-      this.title = $stepElement.find('.js-step-title').text();
-      this.href = $titleLink.attr('href');
+      this.title = $stepElement.find('.js-step-title-text').text().trim();
       this.element = $stepElement;
 
       this.show = show;
@@ -426,7 +429,7 @@
       }
 
       function clickedOnHeading() {
-        return $target.hasClass('js-step-title-button');
+        return $target.hasClass('js-step-title-text');
       }
 
       function iconType() {
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -157,13 +157,14 @@
 
           $title.wrapInner(
             '<span class="js-step-title-text"></span>'
-            );
+          );
 
           $title.wrapInner(
             '<button ' +
             'class="gem-c-step-nav__button gem-c-step-nav__button--title js-step-title-button" ' +
             'aria-expanded="false" aria-controls="' + contentId + '">' +
-            '</button>' );
+            '</button>'
+          );
         });
       }
 
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -95,7 +95,7 @@
             linkText = actions.hideLinkText;
           }
           if (!$(this).find('.js-toggle-link').length) {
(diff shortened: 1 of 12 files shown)


spec/javascripts/components/step-by-step-nav-spec.js
```diff
@@ -54,13 +54,13 @@ describe('A stepnav module', function () {
             </div>\
           </div>\
         </li>\
-        <li class="gem-c-step-nav__step gem-c-step-nav__step--active js-step" id="topic-step-three" data-track-count="stepnavStep">\
```
selfthinker: Can `data-optional` be without a parameter?

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -34,12 +34,12 @@
           circle_class = "gem-c-step-nav__circle--logic" if logic
         %>
         <li class="gem-c-step-nav__step js-step
-          <%= "gem-c-step-nav__step--optional" if step[:optional] %>
           <%= "gem-c-step-nav__step--active" if step_is_active %>"
           <% if step_is_active %>aria-current="step"<% end %>
           <%= "data-show" if step_count == show_step %>
           id="<%= id %>"
           data-track-count="stepNavSection"
+          <%= "data-optional" if step[:optional] %>
```
selfthinker: Can `data-optional` be without a parameter?

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -249,7 +249,8 @@ $top-border: solid 2px $grey-3;
   }
 
   &:hover {
-    .gem-c-step-nav__button {
+    .gem-c-step-nav__button,
+    .gem-c-step-nav__circle {
```
selfthinker: Although this is a bit clearer and more consistent with how other links look and behave, it is hardly noticeable as the text that is changed is usually just a single tiny character/number. I wonder if it should be changed (to e.g. changing the background colour or the colour of the circle border). Although that means you'd have to decide what's more important: noticeability and consistency? Or is there a way both could be achieved?

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -41,23 +41,27 @@
...
-                  <span class="visuallyhidden">Step</span> <%= step_number %>
-                <% end %>
-              </span>
-            </span>
-          </div>
-
           <div class="gem-c-step-nav__header js-toggle-panel" data-position="<%= "#{step_index + 1}" %>">
-            <%= content_tag("h#{heading_level}", step[:title], class: 'gem-c-step-nav__title js-step-title') %>
+            <%= content_tag("h#{heading_level}", class: 'gem-c-step-nav__title') do %>
+              <div class="gem-c-step-nav__circle <%= circle_class %>">
```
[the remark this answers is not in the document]
andysellick: Good spot, will change.

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -155,6 +155,10 @@
           var $title = $step.find('.js-step-title');
           var contentId = $step.find('.js-panel').first().attr('id');
 
+          $title.wrapInner(
+            '<span class="js-step-title-text"></span>'
+            );
```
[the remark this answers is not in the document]
andysellick: Good spot, thanks.

spec/javascripts/components/step-by-step-nav-spec.js
andysellick: Yes, I believe so. It's used as a true/false value: true if present, false if not.

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
andysellick: Yes, I believe so. It's used as a true/false value: true if present, false if not.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
andysellick: Yeah, good question. I struggled with this one for a bit - we had no good design ideas. Eventually I decided that while only a small change, it was at least a small improvement. We may have to revisit this in future.

andysellick:
@selfthinker thanks for the feedback. Regarding your questions, links that are actually links should be now consistent throughout the component. The show/hide button has been made to more resemble a link, but since it's a button I can't give it a visited style. As for the step headings, they're also buttons, and they're supposed to look different in order to visually break up the component... yeah, it's probably not ideal.

Regarding documentation referring to how things were rather than are, the only instance I can think of is the 'optional' parameter, and I left in that description so that no one in the future suddenly decides to try styling the line differently, after we went through all the effort to get it tested and deciding to remove it. I'm trying to record decisions made. Were there other instances of this kind of documentation?

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
selfthinker: Interesting, I assumed it would be invalid without a parameter. But I just checked this in a validator and you're right, it's fine.

app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
```diff
@@ -5,9 +5,7 @@ body: |
 accessibility_criteria: |
   The component is designed to go into the top of an existing content page and should not interfere with the heading structure of the page, so therefore should not contain a heading tag.
 
-  The component must:
-
-  - provide a skip link to the step by step navigation on the page, if one is present
+  An early version of the component contained a hidden skip link for keyboard and screen reader users, that jumped to the step by step navigation component in the sidebar (similiar to the 'skip to content' link at the top of all GOV.UK pages). User testing suggested that rather than helping users it confused them, so this has been removed.
```
selfthinker: This is the other occurrence of referencing in the doc how the code used to be.

nickcolley: One of the most important parts of the criteria is that it stops future people from making the same mistake

selfthinker: I would then rather say it in a form of an acceptance criterion without solutionising it, e.g.
> The component must:
> - not provide a skip link to jump to the navigation component

None of the other criteria does anything like this.
The step by step navigation also has this criterion which [should not mention any technical implementation](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/accessibility_acceptance_criteria.md#dont-define-the-solution):

> be readable when only the text of the page is zoomed - achieved for the numbers and logic elements by their text being wrapped in several elements that give them a white background and ensure the text when zoomed expands to the left, not to the right, where it would obscure the step titles

nickcolley: Sounds good to rewrite it, but may be good to include context of the research so people know why it did not work.

selfthinker:
I would still rewrite those acceptance criteria, but that's not a blocker. :+1:

PR #163, opened 2018-02-14
Add feedback component

andysellick:
Build the new feedback form to go at the foot of all GOV.UK pages. This is based on the code from the Service Manual but with some changes, as this version requires additional functionality.

Ultimately this component will be moved to the publishing components gem, but I started building it in static so getting it reviewed here because the commit history might prove useful to that.

Initial state:

screen shot 2018-02-06 at 14 58 57

Users have three options:

1. click 'yes' (page is useful) records a GA event and shows a 'thanks for your feedback' message
2. click 'no' (page is not useful) opens a form to take an email address to send a link to a survey to. This is a rework of the form that currently appears underneath the header of GOV.UK for some users at some times (you can trigger this by visiting GOV.UK and running GOVUK.userSurveys.displaySurvey(GOVUK.userSurveys.defaultSurvey) in the browser console)
3. click 'is there anything wrong with this page?' opens a feedback form. This is a rework of the existing feedback form at the foot of GOV.UK

**Result of option 1 (and the others, if submitted successfully):**

screen shot 2018-02-06 at 15 02 41

**'Page is not useful' form:**

screen shot 2018-02-15 at 08 49 19

**'Something wrong with page' form:**

screen shot 2018-02-15 at 08 48 27

If javascript is disabled, this form should be the only element that appears, and should still function.

- this code was originally in static, where it received some feedback, [original PR is here](https://github.com/alphagov/static/pull/1235)
- [Trello card](https://trello.com/c/oRLXhaTU/155-make-the-service-manual-feedback-pattern-a-component)
- [Component guide](https://govuk-static-pr-1235.herokuapp.com/component-guide/feedback)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -1,0 +1,152 @@
+window.GOVUK = window.GOVUK || {};
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  "use strict";
+  window.GOVUK = window.GOVUK || {};
+
+  Modules.Feedback = function () {
+
+    this.start = function ($element) {
+      this.$prompt = $element.find('.js-prompt');
+      this.$fields = $element.find('.gem-c-feedback__form-field');
+      this.$forms = $element.find('.js-feedback-form');
+      this.$toggleForms = $element.find('.js-toggle-form');
+      this.$closeForms = $element.find('.js-close-form');
+      this.$activeForm = false;
+      this.$pageIsUsefulButton = $element.find('.js-page-is-useful');
+      this.$pageIsNotUsefulButton = $element.find('.js-page-is-not-useful');
+      this.$somethingIsWrongButton = $element.find('.js-something-is-wrong');
+      this.$promptQuestions = $element.find('.js-prompt-questions');
+      this.$promptSuccessMessage = $element.find('.js-prompt-success');
+
+      var that = this;
+      var jshiddenClass = 'js-hidden';
+
+      setInitialAriaAttributes();
+
+      this.$toggleForms.on('click', function(e) {
+        e.preventDefault();
+        toggleForm($(e.target).attr('aria-controls'));
+        trackEvent(getTrackEventParams($(this)));
+        updateAriaAttributes($(this));
+      });
+
+      this.$closeForms.on('click', function(e) {
+        e.preventDefault();
+        toggleForm($(e.target).attr('aria-controls'));
+        setInitialAriaAttributes();
+      });
+
+      this.$pageIsUsefulButton.on('click', function(e) {
+        e.preventDefault();
+        trackEvent(getTrackEventParams(that.$pageIsUsefulButton));
+        showFormSuccess();
+        revealInitialPrompt();
+      });
+
+      $element.find('form').on('submit', function(e) {
+        e.preventDefault();
+        var $form = $(this);
+        $.ajax({
+          type: "POST",
+          url: $form.attr('action'),
+          dataType: "json",
+          data: $form.serialize(),
+          beforeSend: disableSubmitFormButton($form),
+          timeout: 6000
+        }).done(function (xhr) {
+          trackEvent(getTrackEventParams($form));
+          showFormSuccess(xhr.message);
+          revealInitialPrompt();
+          setInitialAriaAttributes();
+          that.$activeForm.toggleClass(jshiddenClass);
+        }).fail(function (xhr) {
+          showError(xhr);
+          enableSubmitFormButton($form);
+        });
+      });
+
+      function disableSubmitFormButton ($form) {
+        $form.find('input[type="submit"]').prop('disabled', true);
+      }
+
+      function enableSubmitFormButton ($form) {
+        $form.find('input[type="submit"]').removeAttr('disabled');
+      }
+
+      function setInitialAriaAttributes () {
+        that.$forms.attr('aria-hidden', true);
+        that.$pageIsNotUsefulButton.attr('aria-expanded', false);
+        that.$somethingIsWrongButton.attr('aria-expanded', false);
+      }
+
+      function updateAriaAttributes (linkClicked) {
+        linkClicked.attr('aria-expanded', true);
+        $('#' + linkClicked.attr('aria-controls')).attr('aria-hidden', false);
+      }
+
+      function toggleForm (formId) {
+        that.$activeForm = $element.find('#' + formId);
+        that.$activeForm.toggleClass(jshiddenClass);
+        that.$prompt.toggleClass(jshiddenClass);
+
+        var formIsVisible = !that.$activeForm.hasClass(jshiddenClass);
+
+        if (formIsVisible) {
+          that.$activeForm.find('.gem-c-input').first().focus();
+        } else {
+          that.$activeForm = false;
+        }
+      }
+
+      function getTrackEventParams ($node) {
+        return {
+          category: $node.data('track-category'),
+          action: $node.data('track-action')
+        }
+      }
+
+      function trackEvent (trackEventParams) {
+        if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+          GOVUK.analytics.trackEvent(trackEventParams.category, trackEventParams.action);
+        }
+      }
+
+      function clearMessages () {
+        $element.find('.js-errors').html('').addClass(jshiddenClass);
+      }
+
+      function showError (error) {
+        var error = [
+          '<h2 class="gem-c-feedback__heading">',
+          '  Sorry, we’re unable to receive your message right now. ',
+          '</h2>',
+          '<p>If the problem persists, we have other ways for you to provide',
+          ' feedback on the <a href="/contact/govuk">contact page</a>.</p>'
+        ].join('');
+
+        if (typeof(error.responseJSON) !== 'undefined') {
+          error = typeof(error.responseJSON.message) == 'undefined' ? error : error.responseJSON.message;
+        }
+        var $errors = that.$activeForm.find('.js-errors');
+        $errors.html(error).removeClass(jshiddenClass).focus();
+      }
+
+      function clearAllInputs () {
+        that.$fields.val('');
+      }
+
+      function showFormSuccess () {
+        that.$promptQuestions.addClass(jshiddenClass);
+        that.$promptSuccessMessage.removeClass(jshiddenClass).focus();
+      }
+
+      function revealInitialPrompt () {
+        that.$prompt.removeClass(jshiddenClass);
+        that.$prompt.focus();
+      }
+    }
+
+  };
+})(window.GOVUK.Modules);
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -1,0 +1,136 @@
+// govuk_frontend_toolkit
+@import 'grid_layout';
+@import 'design-patterns/buttons';
+
+.gem-c-feedback {
+  max-width: 960px;
+  margin: 0 auto;
+}
+
+// hide without js
+// show with js, unless also has the js-hidden class
+.gem-c-feedback__js-show {
+  display: none;
+
+  .js-enabled & {
+    display: block;
+
+    &.js-hidden {
+      display: none;
+    }
+  }
+}
+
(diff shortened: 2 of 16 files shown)


app/views/govuk_publishing_components/components/_feedback.html.erb
```diff
@@ -0,0 +1,112 @@
...
+        } %>
+
+        <%= render "govuk_publishing_components/components/input", {
+          label: {
+            text: "What went wrong?"
+          },
+          name: "what_wrong"
+        } %>
+
+        <input class="pub-c-feedback__submit" type="submit" value="Send">
```
tijmenb: Maybe this needs a `TODO` comment saying that we'll use the button component once it's moved to the gem.

app/views/govuk_publishing_components/components/_feedback.html.erb
```diff
@@ -0,0 +1,112 @@
...
+        <input name="email_survey_signup[ga_client_id]" type="hidden" value="1627485790.1515403243">
+
+        <h2 class="pub-c-feedback__form-heading">Help us improve GOV.UK</h2>
+        <p id="survey_explanation" class="pub-c-feedback__form-paragraph">To help us improve GOV.UK, we’d like to know more about your visit today. We’ll send you a link to a feedback form. It will take only 2 minutes to fill in. Don’t worry we won’t send you spam or share your email address with anyone.</p>
+
+        <%= render "govuk_publishing_components/components/input", {
+          label: {
+            text: "Email address"
+          },
+          name: "email_survey_signup[email_address]"
```
tijmenb: We can use `email` type now: https://github.com/alphagov/govuk_publishing_components/pull/165

(Thanks @kevindew)

app/views/govuk_publishing_components/components/_feedback.html.erb
andysellick: There's one in the description in the YML file, but I'll update it to be more explicit.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+  outline: 0;
+}
+
+.pub-c-feedback__prompt-question,
+.pub-c-feedback__prompt-success {
+  @include bold-16;
+  display: inline;
+
+  &:focus {
+    outline: solid $yellow 3px;
```
nickcolley: Could we use a semantic variable here? https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss#L63

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -0,0 +1,153 @@
...
+
+      function clearMessages () {
+        $element.find('.js-errors').html('').addClass('js-hidden');
+      }
+
+      function showError (error) {
+        var error = [
+          '<h2 class="pub-c-feedback__heading">',
+          '  Sorry, we’re unable to receive your message right now. ',
+          '</h1>',
```
nickcolley: H2 is not closed with a H2 tag here

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -0,0 +1,153 @@
...
+        trackEvent(getTrackEventParams(that.$pageIsUsefulButton));
+        showFormSuccess();
+        revealInitialPrompt();
+      });
+
+      $element.find('form').on('submit', function(e) {
+        e.preventDefault();
+        var $form = $(this);
+        console.log($form.serialize())
+        $.ajax({
```
nickcolley: What happens if this ajax call never resolves? Does it timeout? Or does the user sit waiting forever?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+  &:focus {
+    outline: solid 3px $yellow;
+  }
+
+  @include media(tablet) {
+    border-width: 5px;
+  }
+
+  // this comes from the backend so we can't put a class on it
+  h2,
```
nickcolley: Do we not have control over the backend code?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+}
+
+.pub-c-feedback__form-label {
+  @include core-16;
+  display: block;
+  padding-bottom: $gutter-half;
+}
+
+// nesting this in order to override other page styles for form fields
+.pub-c-feedback {
```
nickcolley: Is there a plan to use the input component in the future for these?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -0,0 +1,153 @@
...
+          dataType: "json",
+          data: $form.serialize(),
+          beforeSend: disableSubmitFormButton($form)
+        }).done(function (xhr) {
+          console.log(xhr)
+          trackEvent(getTrackEventParams($form));
+          showFormSuccess(xhr.message);
+          revealInitialPrompt();
+          setInitialAriaAttributes();
+          that.$activeForm.toggleClass('js-hidden');
```
vanitabarrett: Can we also put `js-hidden` in a variable?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+  .js-enabled & {
+    display: block;
+
+    &.js-hidden {
+      display: none;
+    }
+  }
+}
+
+.pub-c-feedback__grid-row {
```
vanitabarrett: Is there a reason why we're not able to just use the `grid-row` and `column-two-thirds` classes directly on the HTML elements?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+
+.pub-c-feedback__column-two-thirds {
+  @include grid-column( 2 / 3 );
+}
+
+.pub-c-feedback__prompt {
+  @extend %contain-floats;
+  background-color: $govuk-blue;
+  color: $white;
+  padding: 10px $gutter-half 0;
```
vanitabarrett: Can use $gutter-one-third instead of 10px

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+}
+
+.pub-c-feedback__prompt-link--wrong {
+  float: right;
+}
+
+.pub-c-feedback__error-summary {
+  margin-bottom: $gutter-half;
+  padding: $gutter-half;
+  border: solid 4px $red;
```
vanitabarrett: We can use `$error-colour`

app/assets/javascripts/govuk_publishing_components/components/feedback.js
andysellick: That's an excellent question. Ideally I guess it times out... how best to implement that?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -0,0 +1,148 @@
...
+    @include core-16;
+    margin: 10px 0;
+  }
+}
+
+.pub-c-feedback__error-message {
+  @include bold-19;
+  display: block;
+  padding: 4px 0 0;
+  color: $red;
```
vanitabarrett: Again: we can use `$error-colour`

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: That's a good point, I haven't looked into it. I don't really know where to begin, but it would be good to change...

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: We're using it already, but I failed to remove these styles! Thanks, have removed.

app/assets/stylesheets/govuk_publishing_components/components/print/_feedback.scss
```diff
@@ -0,0 +1,3 @@
+.pub-c-feedback {
+  display: none;
```
vanitabarrett: I'm not sure how print styles work... But if I print the component guide page, I think the component still shows (just without styling)

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: Yes. I don't think they're included in all of our frontend apps - I initially tried this component in feedback (I think it was) and it didn't have the grid.

vanitabarrett: That makes sense, thanks for explaining! I had assumed importing the grid_layout scss here would mean it was available whenever the component is used, but I just remembered these components work a bit differently.

app/assets/stylesheets/govuk_publishing_components/components/print/_feedback.scss
andysellick: Yeah, it does, which is weird. Looks like the print stylesheet is working (works for the step nav component) but not this one. Been staring at this for ages and can't figure it out. Will keep going.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
tijmenb: We should take care of that in a follow up PR, maybe we can make it return better HTML.

app/assets/stylesheets/govuk_publishing_components/components/print/_feedback.scss
andysellick: Ah, figured it out. It was in the print stylesheet for the components, but not the print stylesheet for the component guide. Obviously.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: Also I guess there's an ongoing question over the isolation of these components and how much we should assume is available to them. But that's a story for another day. Happy to approve?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -0,0 +1,151 @@
...
+        e.preventDefault();
+        trackEvent(getTrackEventParams(that.$pageIsUsefulButton));
+        showFormSuccess();
+        revealInitialPrompt();
+      });
+
+      $element.find('form').on('submit', function(e) {
+        e.preventDefault();
+        var $form = $(this);
+        $.ajax({
```
nickcolley: I swear I added this in a review so maybe I'm just going mad but:

What happens if this ajax request never resolves? Does a user get no response?

andysellick: You did add this, but it got squashed because I removed a console log near the line you commented on. I've added in a timeout (good shout) and a test for it. I've not implemented a timeout like this before, I've set it to 6 seconds, is that a good number?

selfthinker:
I've done some automated, manual and AT accessibility testing. Here is what I've found (roughly in severity order):

* Changing colours in Firefox makes the buttons invisible: 


* After expanding one of the two forms, the focus is moved directly to the first input field. That means that screen reader users would miss the introduction (heading and paragraph before). (Tested with VO on macOS and iOS and NVDA in Firefox.)

* When closing the forms again with VoiceOver on macOS (not iOS), the focus gets lost and moves to the top of the page.

* I reckon that it might be confusing to some people that clicking on something makes it disappear and re-appear when closing the revealed content. I would advise to test with users with cognitive issues to make sure this is understood.

* When increasing the font size (not zooming in), hardly any text is increased, only some. We currently have that issue all over gov.uk but it would be good to fix it in new stuff. Here are some screenshots from Firefox (although some but not all browsers will be similar):



* When the form gets submitted or the Yes link gets clicked, NVDA doesn't read the feedback but the focus is kept in the area (straight after).

* The "Is there anything wrong with this page?" link might be missed by screen magnifier users as it's so far to the right and it's not apparent that there might be more content.

* When disabling CSS there are no spaces between the links, making them appear as one single link. That might be an issue for people who use their own user styles.

PR #164, opened 2018-02-14
Add related navigation component

emmabeynon:
For https://trello.com/c/FxxChPUj/257-add-navigation-sidebar-component-to-govuk-publishing-components-gem
Component guide: https://govuk-publishing-compon-pr-164.herokuapp.com/component-guide/related_navigation

In this PR we move the `related_navigation` component from `government-frontend` to this app so that `frontend` can make use of it.  In order for the component to render correctly we have also moved the `RelatedNavigationSidebarHelper` from `govuk_navigation_helpers` and `RelatedNavigationHelper` from `government-frontend` to this app and combined them.


Gemfile.lock
```diff
@@ -271,11 +291,12 @@ DEPENDENCIES
...
+  govuk_schemas (~> 1.0)
   jasmine (~> 2.4.0)
   poltergeist (~> 1.16.0)
   rspec-rails (~> 3.6)
   uglifier (>= 1.3.0)
   webmock (~> 3.0.1)
 
 BUNDLED WITH
-   1.15.1
+   1.16.1
```
vanitabarrett: Do we normally avoid changing this in commits?

app/views/govuk_publishing_components/components/_related_navigation.html.erb
```diff
@@ -0,0 +1,76 @@
+<% related_nav_helper = RelatedNavigationHelper.new %>
+<% related_content = related_nav_helper.build_related_content(local_assigns) %>
+<% if related_content.map(&:values).flatten.any? || related_nav_helper.other.flatten.any? %>
+<% random = Random.new.rand(100) %>
```
vanitabarrett: This does seem hacky, and I noticed it doesn't always work (I guess because the same random number is sometimes generated more than once on the same page). Perhaps this shouldn't block this PR, but a separate issue should be raised around looking into if the ID errors can be ignored in component guide pages?

app/views/govuk_publishing_components/components/_related_navigation.html.erb
```diff
@@ -0,0 +1,76 @@
...
+  <aside class="gem-c-related-navigation" role="complementary">
+    <h2 id="related-nav-related_items-<%= random %>"class="gem-c-related-navigation__main-heading"
```
vanitabarrett: Space missing between the id and class

lib/govuk_publishing_components/components/related_navigation_sidebar_helper.rb
```diff
@@ -0,0 +1,140 @@
+require 'govuk_navigation_helpers/content_item'
+require 'govuk_navigation_helpers/grouped_related_links'
+
+class RelatedNavigationSidebarHelper
```
vanitabarrett: Now `related_navigation_sidebar_helper` and `related_navigation_helper` are in the same place, it might be worth renaming these to make it clear what the difference is between them. Same goes for the related test files.

Gemfile.lock
emmabeynon: Good spot, will change back.

app/views/govuk_publishing_components/components/_related_navigation.html.erb
emmabeynon: Yes definitely an issue to add to the backlog, as it's not an ideal solution.

lib/govuk_publishing_components/components/related_navigation_sidebar_helper.rb
emmabeynon: Just having a think about this.  From what I understand `RelatedNavigationSidebarHelper` helps construct the links that are passed in to the related-navigation component, so we could rename to something like `RelatedNavigationLinksHelper`?  What do you think?

vanitabarrett: Kind of, I think it's more about retrieving the data and putting it in a format that can be passed to the related navigation component

app/views/govuk_publishing_components/components/_related_navigation.html.erb
nickcolley: Consider using `SecureRandom` which should guarantee randomness 

See https://github.com/alphagov/govuk_publishing_components/blob/a323d988e691d4e44c8e91db3ad92ab33d869a63/app/views/govuk_publishing_components/components/_input.html.erb#L2

emmabeynon: Good shout, thanks @nickcolley

lib/govuk_publishing_components/components/related_navigation_sidebar_helper.rb
steventux: I think we should lose the word `Sidebar` from however this is named. It infers position and this varies across devices.

Gemfile.lock
```diff
@@ -89,35 +98,45 @@ GEM
...
+    govuk-lint (3.6.0)
       rubocop (~> 0.51.0)
       rubocop-rspec (~> 1.19.0)
       scss_lint
-    govuk_frontend_toolkit (7.2.0)
+    govuk_ab_testing (2.4.1)
+    govuk_frontend_toolkit (7.3.0)
       railties (>= 3.1.0)
       sass (>= 3.2.0)
+    govuk_navigation_helpers (9.0.0)
```
steventux: How much code duplication/headache would we have by not making this a dependency?

emmabeynon: We'd need to copy over these two files:
https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/content_item.rb

https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/grouped_related_links.rb

Gemfile.lock
```diff
@@ -89,35 +98,45 @@ GEM
       money (~> 6.7)
       nokogiri (~> 1.5)
       sanitize (~> 2.1.0)
-    govuk-content-schema-test-helpers (1.6.0)
-      json-schema (~> 2.8.0)
-    govuk-lint (3.4.0)
+    govuk-content-schema-test-helpers (1.5.0)
```
maxgds: downgrading? is that deliberate?

lib/govuk_publishing_components/components/related_navigation_helper.rb
```diff
@@ -0,0 +1,245 @@
...
+        text: link["title"]
+      }.merge(additional_attr)
+    end
+  end
+
+  def world_location_base_path(title)
+    "/world/#{parameterise(title)}/news"
+  end
+
+  def related_items
```
steventux: The [finder links are currently appended to the `related_items` collection of links](https://github.com/alphagov/government-frontend/blob/master/app/presenters/navigation/finders.rb#L14). I don't think this should be a concern for this PR though as we are pretty doubtful of their efficacy and there's a good likelihood of their removal soon. If they are likely to be iterated into the related navigation component we can move the relevant presenter logic later.

lib/govuk_publishing_components/components/related_navigation_helper.rb
```diff
@@ -0,0 +1,245 @@
+class RelatedNavigationHelper
```
steventux: :+1: This is a great refactor. Having all the helper logic in one place is much clearer.

Gemfile.lock
emmabeynon: I've had a look into this and it looks like the `govuk_schemas` gem is the reason this is being downgraded, as we're using an early version of it which depends on an older version of `govuk-content-schema-test-helpers`.  I've had a quick look into what it would take to upgrade to a newer version of `govuk_schemas` and it looks a bit fiddly, so I'll create a separate card for this.

emmabeynon:
Thanks @steventux - I missed out updating the data for that world locations example, that's now fixed.

PR #165, opened 2018-02-14
Support different input types for Input component

kevindew:
Hello :wave:

This allows the rendered HTML to be `<input type="email" />` or whatever
other incarnation may be needed. By default this will be text.

There's potential here that you could hit some rendering issues by
putting in something wacky here, but it feels too restrictive to try
prevent that here.

See the [rendered component](https://govuk-publishing-compon-pr-165.herokuapp.com/component-guide/input)

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/input.yml
+++ b/app/views/govuk_publishing_components/components/docs/input.yml
@@ -12,11 +12,17 @@ examples:
   default:
     data:
       label:
-        text: "What is your email address?"
-      name: "address"
-  with_error:
+        text: "What is your name?"
+      name: "name"
+  email_type:
     data:
       label:
         text: "What is your email address?"
       name: "address"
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -10,13 +10,19 @@ accessibility_criteria: |
...
   default:
+    data:
+      label:
+        text: "What is your name?"
+      name: "name"
+  email_type:
     data:
       label:
         text: "What is your email address?"
+        type: "email"
```
nickcolley: Looks like the type might be in the wrong place here?

kevindew: Doh! good catch

kevindew: @nickcolley should be good now.

PR #167, opened 2018-02-15
Fix print styles

tijmenb:
This adds the print stylesheet from the application to the component guide. That stylesheet includes the print styles for the components. They're removed from the main stylesheet of the component guide because that should not contain component code.


fofr:
Print styles added in this PR:
https://github.com/alphagov/govuk_publishing_components/pull/19

Not sure how/why they got removed from application.html.erb.

fofr:
Looks like they got removed here:
https://github.com/alphagov/govuk_publishing_components/commit/754fab47222e05e6d0d05d78d454a0a551ced6f3#diff-56a57a6d4d7d7cd5c2d441170c392562

PR #173, opened 2018-02-19
Fix feedback component accessibility issues

andysellick:
Changes in response to [accessibility review of the feedback component](https://github.com/alphagov/govuk_publishing_components/pull/163#issuecomment-366267858)

Issues addressed:

- 2, expanded the input component used by the feedback component to include an aria-describedby attribute, then set to the id of the describing paragraph. Have only tested this in VO on Safari.
- 3, have set focus back to the initial 'is something wrong' prompt on form close.
- 5, can't do much against the GOV.UK font mixins, but I've removed the explicit definition on the initial 'is something wrong' text so it's at least consistent compared with the links next to it (i.e. all of them are now resizable)

I don't currently have fixes for the remaining issues, @selfthinker happy to discuss further.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -36,6 +36,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         e.preventDefault();
         toggleForm($(e.target).attr('aria-controls'));
         setInitialAriaAttributes();
+        revealInitialPrompt();
       });
 
       this.$pageIsUsefulButton.on('click', function(e) {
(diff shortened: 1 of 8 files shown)


selfthinker:
The commit message for 90f93c4820ebf6a6a28796b8df98ddc332acdd50 should say `aria-describedby`, not `aria-labelledby`.

andysellick:
@selfthinker commit message updated, good spot, thanks.

Regarding shifting the focus on clicks, unfortunately we haven't done any user testing on this component. I have raised this, and work will be ongoing. It would be really good if someone could push for this while I'm away.

selfthinker:
I also wanted to add that I agree changing the mixins is out of scope for this, but it's good that you have investigated and found what's easily possible.
Although it would generally be good to fix the mixins, as the new frontend and design system are under way, it's probably not feasible to do it from that perspective either.

PR #174, opened 2018-02-19
Make legend optional on a fieldset

kevindew:
Legend is an optional child of a fieldset element however this component previously required one which meant if you couldn't use the legend for some reason you would have to abandon the component.

On email-alert-frontend we were specifically caught out by wanting to style headers and copy via govspeak (which means it can't go into the legend as the govspeak component uses a `<div>`) and so we've gone without a fieldset which has left us missing a vertical margin:

screen shot 2018-02-19 at 18 14 33

Other options to fix our immediate problem would be:

- Keeping the text in the legend but finding a way to style it and then:
  - Having a govspeak class on the legend itself (I believe we're allowed `h*` / `p` elements in a `legend` but not `div`) or
  - Using something else to replicate the govspeak styling on the legend
- Having a margin-bottom modifier for radio buttons and not using the fieldset (which I mocked up in a [different branch](https://github.com/alphagov/govuk_publishing_components/tree/radio-button-margin-bottom) but then decided it was probably better to try maintain a fieldset)

/cc @andysellick

### DIFF
--- a/app/views/govuk_publishing_components/components/_fieldset.html.erb
+++ b/app/views/govuk_publishing_components/components/_fieldset.html.erb
@@ -1,7 +1,12 @@
 <% text = text || yield %>
-<fieldset class="gem-c-fieldset">
-  <legend class="gem-c-fieldset__legend">
-    <%= legend_text %>
-  </legend>
-  <%= text %>
-</fieldset>
+<% legend_text = legend_text || nil %>
+<% if text.present? || legend_text.present? %>
+  <fieldset class="gem-c-fieldset">
+    <% if legend_text %>
+      <legend class="gem-c-fieldset__legend">
+        <%= legend_text %>
+      </legend>
+    <% end %>
+    <%= text %>
+  </fieldset>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/fieldset.yml
+++ b/app/views/govuk_publishing_components/components/docs/fieldset.yml
@@ -32,3 +32,12 @@ examples:
 
         <input type="radio" id="html-legend-no" name="html-legend">
         <label for="html-legend-no">No</label>
+  without_legend:
+    data:
+      text: |
+        <!-- Use the radio component, this is hardcoded only for this example -->
+        <input type="radio" id="html-legend-yes" name="html-legend">
+        <label for="html-legend-yes">Yes</label>
+
+        <input type="radio" id="html-legend-no" name="html-legend">
+        <label for="html-legend-no">No</label>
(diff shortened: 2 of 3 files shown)


app/views/govuk_publishing_components/components/_fieldset.html.erb
```diff
@@ -1,7 +1,12 @@
 <% text = text || yield %>
-<fieldset class="gem-c-fieldset">
-  <legend class="gem-c-fieldset__legend">
-    <%= legend_text %>
-  </legend>
-  <%= text %>
-</fieldset>
+<% legend_text = legend_text || nil %>
+<% if text.present? || legend_text.present? %>
```
andysellick: Is it necessary to check for the existence of the legend_text here? Is there a situation where you'd want the component to render with only the legend_text?

app/views/govuk_publishing_components/components/_fieldset.html.erb
```diff
@@ -1,7 +1,12 @@
 <% text = text || yield %>
-<fieldset class="gem-c-fieldset">
-  <legend class="gem-c-fieldset__legend">
-    <%= legend_text %>
-  </legend>
-  <%= text %>
-</fieldset>
+<% legend_text = legend_text || nil %>
```
andysellick: I think this can be written as just `legend_text ||= nil` ?

app/views/govuk_publishing_components/components/docs/fieldset.yml
```diff
@@ -30,5 +30,14 @@ examples:
         <input type="radio" id="html-legend-yes" name="html-legend">
         <label for="html-legend-yes">Yes</label>
 
+        <input type="radio" id="html-legend-no" name="html-legend">
+        <label for="html-legend-no">No</label>
+  without_legend:
+    data:
+      text: |
+        <!-- Use the radio component, this is hardcoded only for this example -->
+        <input type="radio" id="html-legend-yes" name="html-legend">
```
andysellick: The ids used here and below - html-legend-yes and html-legend-no - are already used in another example in the same file, so the component guide is reporting this as an error, when it isn't really. Best to use unique ids in data for each example.

andysellick:
@bevanloon that's a good point and one I'd not thought about yet. I guess in the example above, why isn't the text 'how often do you want to get updates' the legend?

Maybe the solution isn't to remove the legend but to allow further styling options on it. @kevindew what do you think? Happy to help with that if it works?

kevindew:
@bevanloon @andysellick yup, if we can get more styling in the legend this won't be needed. Thanks for offer of help, would be most appreciated 👍 

A small caveat is that this work is becoming pressing though so if we can't get that approach in relatively quick we might need to get something else in quick though.

kevindew:
After a chat with @andysellick we're going to try spin this a different way - PR coming soon.

PR #182, opened 2018-02-20
Improve accessibility of related navigation

maxgds:
Based on feedback from the accessibility team this
- Moves the navigation header inside the navigation element
- Removes the aside/complementary semantics

The accessibility team have agreed that the navigation component should be marked up as navigation and not as tangential content (see [Trello card](https://trello.com/c/S96QAn30/274-improve-accessibilty-context-of-related-navigation-link-grouping)).

### DIFF
--- a/app/views/govuk_publishing_components/components/_related_navigation.html.erb
+++ b/app/views/govuk_publishing_components/components/_related_navigation.html.erb
@@ -2,7 +2,7 @@
 <% related_content = related_nav_helper.related_navigation %>
 <% if related_content.map(&:values).flatten.any? || related_nav_helper.other.flatten.any? %>
 <% random = SecureRandom.hex(4) %>
-  <aside class="gem-c-related-navigation" role="complementary">
+  <div class="gem-c-related-navigation">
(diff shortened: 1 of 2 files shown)


app/views/govuk_publishing_components/components/_related_navigation.html.erb
```diff
@@ -2,7 +2,7 @@
 <% related_content = related_nav_helper.related_navigation %>
 <% if related_content.map(&:values).flatten.any? || related_nav_helper.other.flatten.any? %>
 <% random = SecureRandom.hex(4) %>
-  <aside class="gem-c-related-navigation" role="complementary">
+  <div class="gem-c-related-navigation">
```
nickcolley: If there's anything interesting could we put it in the accessibility acceptance criteria so that someone trying to be useful (me) later does not add aside back?

emmabeynon: Good shout Nick, I've added a note about it in the documentation.

nickcolley: Thanks Emma 👍

PR #190, opened 2018-02-26
Set feedback component font sizes for initial state

andysellick:
- set font size for 'is this page useful?' and links beside it
- various apps have styles that were affecting the component, defining them explicitly like this should help
- does prevent an earlier accessibility fix to allow text resizing in the browser settings to work, but since that doesn't work in the rest of the component where we already defined font sizes, or even in the rest of GOV.UK, this doesn't seem like a huge concern

No visual changes.

Trello card: https://trello.com/c/vKfhGFtn/4-font-sizing-on-frontend-whitehall

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -39,19 +39,26 @@
 
 .gem-c-feedback__prompt-question,
 .gem-c-feedback__prompt-success {
-  display: inline;
-  font-weight: bold;
+  @include bold-19;
+  display: inline-block;
 
   &:focus {
     outline: 0;
   }
+
+  @include media(tablet) {
+    @include bold-16;
+    float: left;
+  }
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -39,19 +39,26 @@
 
 .gem-c-feedback__prompt-question,
 .gem-c-feedback__prompt-success {
-  display: inline;
-  font-weight: bold;
+  @include bold-19;
```
selfthinker: This was `bold-16` before it was changed in 12c1b283186b639f3b729264f0f17ff2b2121f63. Is it intentional that it's 19px now?

andysellick: I think this is to ensure that the font size is 16px on both desktop and mobile - our type mixins automatically shrink the font for mobile, which is a bit unreadable if you start at 16px on desktop.

PR #191, opened 2018-02-26
Add spacing above feedback component

andysellick:
Trello card: https://trello.com/c/T3ez6Xxk/12-often-no-space-between-content-and-component


tijmenb:
Should we have something in the component docs about the desired spacing between the component and the elements above? Should it just be what is defined in this PR, or do we expect there to be other spacing? 

For example, there's already a `30px` space between the [content and component on a browse page](https://www.gov.uk/browse).

PR #192, opened 2018-02-26
Release 5.2.2

tijmenb:


Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (5.2.1)
+    govuk_publishing_components (5.2.2)
```
andysellick: I didn't think this file needed to be changed to do a release? I thought it was just the version and the changelog.

PR #194, opened 2018-02-27
Feedback component: Fix styling on pages with older layouts

tijmenb:
I suspect that this was caused by https://github.com/alphagov/govuk_publishing_components/pull/190, but can't be sure.

## Before

screen shot 2018-02-27 at 15 31 11

## After

screen shot 2018-02-27 at 15 31 03


nickcolley:
Looks like there's a `margin-bottom` too...

screen shot 2018-02-27 at 16 56 45

PR #196, opened 2018-03-01
Add step by step navigation component helpers

sihugh:
There is a version of this in govuk_navigation_helpers which provided similar assistance when we were using hard coded step navigation configuration held in the gem.  We are retiring that approach for two reasons:

- govuk_navigation helpers is too closely coupled to govuk_publishing_components so we need to move [those coupled things](https://github.com/alphagov/govuk_navigation_helpers/blob/master/lib/govuk_navigation_helpers/step_nav_content.rb) out of that gem and into this one.
- we can now use linked `part_of_step_navs` content to power the step nav configuration so we can get move away from hardcoded files.

`step_nav` is fairly barebones, but I intend to expand this in future to replace [the current equivalent in collections](https://github.com/alphagov/collections/blob/086758a7ec50e33f94f137b4606299b5fe7bbf50/app/services/step_nav_content.rb).

`step_nav_helper` does most of the heavy lifting for transforming a linked step by step navigation content item into something that can be easily used to show the navigation components.

There's some specimen usage for this in `spec/dummy/app/views/step_nav/show.html.erb`
which is based on the default rules we've settled on.  These are:

- if a piece of content is in one step by step navigation, show the header,
  a related links item (it's the title) and the sidebar component
- if it's in 2, 3 or 4 step navs, just show the related step navigation components
  for each of them
- if it's in 0 or > 4, don't show any of the step nav elements (probably fall back to topics)

We're patching links to the learn-to-drive-a-car step by step navigation here: https://github.com/alphagov/collections-publisher/pull/330

Once these two things are merged, we can start to migrate the frontend apps off the hardcoded config in govuk_navigation_helpers.

https://trello.com/c/nhjWvdTN/535-update-govukpublishingcomponents-to-read-sidebar-nav-data-from-links-data

### DIFF
--- a/spec/features/step_nav_helper_spec.rb
+++ b/spec/features/step_nav_helper_spec.rb
@@ -1,0 +1,185 @@
+require "rails_helper"
+require "gds_api/test_helpers/content_store"
+
+describe 'Specimen usage of step by step navigation helpers' do
+  include GdsApi::TestHelpers::ContentStore
+
+  context 'no related step by step navigation journeys' do
+    before do
+      content_store_has_random_item(base_path: '/vomit-comet', schema: 'transaction', part_of_step_navs: nil)
+
+      visit '/step-nav/vomit-comet'
+    end
+
+    it 'does not show step nav related components' do
+      expect(page).not_to have_selector('.gem-c-step-nav-header')
+      expect(page).not_to have_selector('.gem-c-step-nav-related')
+      expect(page).not_to have_selector('.gem-c-step-nav')
+    end
+  end
+
+  context 'one related step by step navigation journey' do
+    before do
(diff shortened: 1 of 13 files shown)


spec/features/step_nav_helper_spec.rb
```diff
@@ -0,0 +1,185 @@
...
+    end
+
+    it "doesn't show the full step nav sidebar or header" do
+      expect(page).not_to have_selector('.gem-c-step-nav-header')
+      expect(page).not_to have_selector('.gem-c-step-nav')
+    end
+  end
+
+
+  def content_store_has_random_item(base_path:, schema: 'placeholder', part_of_step_navs: [])
```
Davidslv: Maybe we should add a note to extract this into `gds-api-adapters` as `frontend` also has this method (just not the same exact code)

sihugh: 👍

PR #201, opened 2018-03-08
Namespace and reorganise components location

Davidslv:
This PR is to properly namespace both StepNavHelper and RelatedNavigationHelper to be under GovukPublishingComponents namespace.

It also corrects the spec locations.

cc\ @emmabeynon @sihugh


sihugh:
I'm happy if @emmabeynon 's happy

CHANGELOG.md
```diff
@@ -1,3 +1,8 @@
+** Unreleased **
+
+Properly namespaces StepNavHelper and RelatedNavigationHelper to be under GovukPublishingComponents namespace.
+This is a breaking change to all the apps using this two components.
```
tijmenb: Why is this breaking? Aren't these classes only used from inside the component?

emmabeynon:
This seems fine to me.  I think @tijmenb is correct that it shouldn't be breaking.

Davidslv:
Thank you, I got confused with the classes that were exposed. I've removed the line in the CHANGELOG.md

lib/govuk_publishing_components/components/step_by_step_nav_helper.rb
```diff
@@ -1,107 +1,111 @@
-class StepNavHelper
-  include ActionView::Helpers
-  include ActionView::Context
+module GovukPublishingComponents
+  module Components
+    class StepNavHelper
```
Davidslv: this class name is wrong according to the file name (I've just noticed this)

tijmenb:
This is superseded by https://github.com/alphagov/govuk_publishing_components/pull/206. Thanks @Davidslv!

PR #202, opened 2018-03-08
Update document list with description

vanitabarrett:
We want to make use of the document list component on topic pages. For mainstream services, we want to show a description of the document rather than metadata.

screen shot 2018-03-08 at 11 47 03

In the future, we may also want to supply additional metadata for each document, such as organisation and collection the document belongs to. This PR refactors the component to stop it being limited to only updated date and document type

Component Guide Link: https://govuk-publishing-compon-pr-202.herokuapp.com/component-guide/document_list

### DIFF
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -14,16 +14,24 @@
             )
           %>
         </h3>
-        <ul>
-          <li class="gem-c-document-list__attribute">
-            <time datetime="<%= item[:metadata][:public_updated_at].iso8601 %>">
-              <%= l(item[:metadata][:public_updated_at], format: '%e %B %Y') %>
-            </time>
-          </li>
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -14,16 +14,22 @@
...
-          <li class="gem-c-document-list__attribute">
-            <time datetime="<%= item[:metadata][:public_updated_at].iso8601 %>">
-              <%= l(item[:metadata][:public_updated_at], format: '%e %B %Y') %>
-            </time>
-          </li>
-          <li class="gem-c-document-list__attribute gem-c-document-list__attribute--document-type">
-            <%= item[:metadata][:document_type] %>
-          </li>
-        </ul>
+        <p class="gem-c-document-list__item-description" ><%= item[:link][:description] %></p>
```
maxgds: Since the description is optional we should check for it before writing paragraph tags.

PR #203, opened 2018-03-08
Add inverse header component

maxgds:
This component is a wrapper header to be used on topic and list pages:
https://trello.com/c/g2MMWMQr/110-blue-header

screen shot 2018-03-08 at 14 52 18

Component Guide Link: https://govuk-publishing-compon-pr-203.herokuapp.com/component-guide/inverse_header

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/inverse_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/inverse_header.yml
@@ -1,0 +1,41 @@
+name: Inverse header
+description: A wrapper to contain header content in white text
+body: |
+  This component can be passed a block of template code and will wrap it in a blue header. This is as light-touch as possible and doesn't attempt to deal with the margins and paddings of its content, neither does it enforce the white text it requires. Implemented to accomodate topic and list page headings and breadcrumbs but unopinionated about its contents.
+
+accessibility_criteria: |
+  The component must:
+
+  * be used in conjunction with content that renders a text contrast ratio higher than 4.5:1
+  against the header colour to meet WCAG AA.
+
+examples:
+  default:
+    data:
+      block: |
+        <div class="pub-c-title pub-c-title--inverse">
+          <h1 class="pub-c-title__text ">
+            Education, Training and Skills
+          </h1>
+        </div>
+  with_breadcrumbs_and_paragraph:
+    data:
+      block: |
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/docs/inverse_header.yml
```diff
@@ -0,0 +1,41 @@
...
+  * be used in conjunction with content that renders a text contrast ratio higher than 4.5:1
+  against the header colour to meet WCAG AA.
+
+examples:
+  default:
+    data:
+      block: |
+        <div class="pub-c-title pub-c-title--inverse">
+          <h1 class="pub-c-title__text ">
+            HTML publication page title
```
vanitabarrett: Can we change the examples to Topic/List page examples to reflect how we think this component will be used at the moment

maxgds: Do you mean the content text? Or the (rendered) component html?

steventux:
@vanitabarrett I think the publication header component could eventually be replaced with this.
It seems very specific to HTML publications though.
The use case for this header will be in both topic and list pages, they have differing contents, so being able to wrap any content is going to be very useful.

app/views/govuk_publishing_components/components/docs/inverse_header.yml
vanitabarrett: Just changing 'HTML publication page title' to (for example) 'Education, training and skills' to show how it looks on a Topic page. When topic pages are live, we can link to an example in context, but for now we can do this as a quick win to give an example of what it would look like.

maxgds:
The component as written will go to 100% of whatever it's put inside - I don't know how the topic pages will be structured but if it's at the same level as the global banner it would go full width. If it's put inside a constrained panel I'm not sure how we'd break back out of it without getting a bit hacky.

As for the other component, it might be similar enough to share them, but maybe consolidation is something to look at later? It would have to come into the gem to be used everywhere anyway.

nickcolley:
What's the difference between this and http://govuk-static.herokuapp.com/component-guide/title/in_html_publication ?

vanitabarrett:
@maxgds @steventux I've added a story in the backlog here to remind us to revisit Inverse Header vs Publication Headers https://trello.com/c/0pbli2e8/37-replace-publication-header-component-with-inverse-header-component

PR #206, opened 2018-03-09
Restructure gem and add documentation

tijmenb:
This builds on https://github.com/alphagov/govuk_publishing_components/pull/201.

This PR is an attempt to structure the app in such a way that we separate internal classes and helpers from publicly exposed classes.

This hopefully will solve a dilemma we've been having with [govuk_navigation_helpers](https://github.com/alphagov/govuk_navigation_helpers). That gem was introduced to generate payloads for components. This was a suboptimal solution to the problem that we couldn't reasonably add more complex logic to the components in Static ([see the analytics component a cautionary example](https://github.com/alphagov/static/blob/e5603f0596ef324c077f988b81a1cc9ab874409c/app/views/govuk_component/analytics_meta_tags.raw.html.erb)).

Now that we are moving all components to this gem, we are better able to use helpers and there's no more need to separate the gems that create payloads and the components. I plan to merge in govuk_navigation_helpers in https://github.com/alphagov/govuk_publishing_components/pull/207. 

https://trello.com/c/roHUGIpK


tijmenb:
@nickcolley I've renamed the `ComponentHelpers` to `Presenters` to make the distinction clearer.

PR #210, opened 2018-03-12
Handle related items not tagged to mainstream browse pages

emmabeynon:
For https://trello.com/c/nGtzTPCu/283-update-government-frontend-to-use-relatednavigation-component

Some work unrelated to this PR has flagged that there are occasions where an
ordered related link may not be tagged to a mainstream browse pages, and the
current implementation does not handle this gracefully, so we update it to
better handle this edge case.


emmabeynon:
@tijmenb good point - test now added 👍

PR #212, opened 2018-03-12
Add option to removed side padding if full width

fmlharrison:
In this commit I have added the option to pass a flag to the component
to indicate if the header component will cover the full width of the
page. If so, the left-right padding is removed for the header so it’s 
content can line up with the rest of the content on the page.

I also added a bottom margin of 30px to the component.

https://trello.com/c/xfcxY63h/122-correct-page-header-component-padding

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
@@ -1,5 +1,11 @@
 .gem-c-inverse-header {
   width: 100%;
   background-color: $govuk-blue;
+  margin-bottom: $gutter;
   padding: $gutter-half;
+}
+
+.gem-c-inverse-header--full-width {
+  padding-left: 0;
+  padding-right: 0;
 }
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
```diff
@@ -1,5 +1,15 @@
-.gem-c-inverse-header {
+%inverse-header {
```
nickcolley: If you follow BEM you can avoid generating additional CSS.

```css
.inverse-header { ... } /* default styles */
.gem-c-inverse-header--full-width {
  padding-left: 0;
  padding-right: 0;
}
```

```erb
<%
  full_width ||= false
  header_class = full_width ? "gem-c-inverse-header--full-width" : ""
%>
<% block = yield %>
<% unless block.empty? %>
  <header class="gem-c-inverse-header <%= header_class %>">
    <%= block %>
  </header>
<% end %>
```

fmlharrison: Thanks @nickcolley! That is really useful.

PR #214, opened 2018-03-13
Enforce white text and links for inverse header component

maxgds:
Tests for accessibility in other apps failed where the inverse header component in isolation didn't have the required context to make its example content render in white. Since the point of an inverse state is to have light on dark, I have updated the component to be more opinionated about white text. This would require overrides if white weren't desired but all planned implementations currently use white anyway. Unfortunately, this has required me to colour the "a" tag as the component is unaware of what CSS classes might be applied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+Enforce white content in inverse component (PR #214).
+
 # 5.5.2
 
 * Add `full-width` flag to remove left and right padding when using a full width page header (PR #212).
--- a/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
@@ -1,8 +1,13 @@
 .gem-c-inverse-header {
   width: 100%;
   background-color: $govuk-blue;
+  color: $white;
   margin-bottom: $gutter;
   padding: $gutter-half;
+}
+
+.gem-c-inverse-header a {
+  color: $white;
 }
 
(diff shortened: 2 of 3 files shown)


maxgds:
@nickcolley 
> In other components we've allowed the 'inverse' option to be passed to the components that need to be inverse, is this possible for this?

It's a weird one - this component takes content and sticks a wrapper around it, it's really lightweight. It's intended to be used with other components in inverse mode (they've been updated to have this where wanted). But tests are failing in collections and government-frontend because the context in which the tests are run doesn't import enough css, so it tests unstyled content within the wrapper, which defaults to black and fails contrast. So the options are we either force white for contents in the component (which they should be anyway) or we put example content with inline styling to make it white. As @steventux points out, an inverse component that doesn't set a foreground colour, only a background, isn't really inverting anything, so this seemed like the better fix.

nickcolley:
If tests are failing you could consider excluding a rule: https://github.com/alphagov/govuk_publishing_components/blob/7fd453675617dc54137743cc5252b9f45686b411/spec/dummy/app/views/components/docs/test-component-with-duplicate-ids.yml#L2

The strength of a good component system relies on making sure the components do not leak and break other things unintentionally, if it's possible to do this without breaking out of BEM that'd be good.

CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# 5.5.3
```
steventux: You can leave this as `# Unreleased`. We usually bump the gem in a separate, subsequent PR. We usually do this to prevent immediate publishing of the gem, in case there's some integration testing necessary.

PR #216, opened 2018-03-13
[WIP] White text

maxgds:


maxgds:
@nickcolley This is the alternative fix. The other option was my preferred choice because it actually makes the inverse a real inverse by forcing links white. Nobody will ever accidentally wrap content they don't want to be pale, so the non-bem styling issue only comes in if they REALLY want to force a different pale colour for links. Thoughts on which is the lesser evil?

nickcolley:
@maxgds if you're confident that no one will want to nest content that would not be white in the future you should go with your gut 👍

maxgds:
Cool, will go with the other one, then. Thanks for your comments.

PR #222, opened 2018-03-15
Add margin_top flag to feedback component

vanitabarrett:
Topic pages need the feedback component to be flush with the bottom of the taxon grid. We therefore need to be able to add a flag to remove the top margin in these cases

Example use on a topic page:

screen shot 2018-03-15 at 10 27 52

### DIFF
--- a/app/views/govuk_publishing_components/components/_feedback.html.erb
+++ b/app/views/govuk_publishing_components/components/_feedback.html.erb
@@ -1,8 +1,10 @@
 <%
   contact_govuk_path = "/contact/govuk"
+  margin_top ||= 1
+  margin_top_class = "gem-c-feedback--top-margin" if margin_top == 1
 %>
 
-<div class="gem-c-feedback" data-module="feedback">
+<div class="gem-c-feedback <%= margin_top_class %>" data-module="feedback">
   <div class="gem-c-feedback__prompt gem-c-feedback__js-show js-prompt" tabindex="-1">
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_feedback.html.erb
```diff
@@ -1,8 +1,10 @@
 <%
   contact_govuk_path = "/contact/govuk"
+  margin_top||= 1
```
maxgds: missing space before the ||

app/views/govuk_publishing_components/components/_feedback.html.erb
```diff
@@ -1,8 +1,10 @@
...
+  margin_top_class = "gem-c-feedback--top-margin" if margin_top == 1
```
maxgds: not sure you need the 
== 1

vanitabarrett:
@maxgds Added the missing space 👍 For the second comment,  I think the `== 1` is needed otherwise the margin top class is always applied (because margin_top param always exists)

maxgds:
I thought if it was set to false it would be the same as not existing in the evaluation, but happy to take your word for it.

vanitabarrett:
@maxgds Ah okay, I was following a pattern for another component where we pass `margin_top: 0` rather than false. Perhaps we need to add a story in the frontend backlog to go and standardise how we do this across all components.

PR #226, opened 2018-03-16
First version of the contextual sidebar component

tijmenb:
This is the only sidebar you'll ever need. Its purpose is to capture all the logic for sidebars in a single place so frontend apps become really simple.

This sidebar component is extracted from government-frontend - the corresponding PR that uses this will be up soon.

Note that I've changed as little as possible to make it work. We'll follow this up with a full refactor of the system. I hope to have something like a single class that decides what to show in the sidebar, and then different classes that generate the proper payloads for the HTML. Also, once this is complete we probably won't need the concept of `AppHelpers` anymore, making everything even simpler.

Todo:
- [ ] Super high level integration test with random content


tijmenb:
Superseded by https://github.com/alphagov/govuk_publishing_components/pull/228, which also does the sidebar and more.

PR #227, opened 2018-03-17
Prefill step nav tracking_id

selfthinker:
This prefills the `tracking_id` for step nav components introduced in #162 with the <del>slug</del> <ins>content_id</ins> of each step nav.
It also corrects the use of that option in the `step_by_step_nav_related` component to be set per step nav and not per the whole component.

We need this to be able to differentiate between different step navs as soon as a page is shared between two or more step navs.

[Trello card](https://trello.com/c/xzXqnQgf)

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
@@ -1,7 +1,6 @@
 <%
   links ||= false
   pretitle ||= t("govuk_component.step_by_step_nav_related.part_of", default: "Part of")
-  tracking_id ||= false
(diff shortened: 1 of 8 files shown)


andysellick:
Would be good to update the documentation for each of the three components to be explicit about this tracking and the nature of it.

selfthinker:
@andysellick, I've made the changes to the documentation. Is that clearer?

I've also managed (with a bit of help from David) to actually use the **content_id** and not the slug, so that is much more future-proof in case the slug changes. I rebased my original changes.
I think @sihugh mentioned that the content_id might not always be there. Was that an issue which is fixed now or do we need to take precautions because of that?

PR #231, opened 2018-03-20
Improve step nav substep creation

andysellick:
- modify data structure so substeps themselves have contents, rather than occurring in a sequence of elements and relying on the template to close the substep div
- move substep generation into the helper rather than the template

**WARNING**: this is a breaking change. No live step nav currently uses substeps, but the publishing tool is currently using the old data structure.

Old data structure:

```
{
    type: 'substep',
    optional: true
},
(elements inside the substep)
```

New:

```
{
    type: 'substep',
    optional: true,
    contents: [
        (elements inside the substep)
    ]
},
```

Trello card: https://trello.com/c/PdxP84XT/410-refactor-task-list-component-code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Unreleased
+
+* Improve substep creation (PR #231)
+
 # 5.6.0
 
 * Restore 'referer' field to feedback component form submission (PR #232)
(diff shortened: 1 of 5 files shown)


andysellick:
@sihugh here's the breaking change I mentioned. Won't merge until you're happy, shout if you want me to go through it in any more detail.

sihugh:
I don't see any particular reason to not merge this.  We'd need to do some work in the publishing tool and to the content schema of step_by_step_nav to allow us to actually use this, but as you say, it's unused at present so seems fine to me.

We're just removing the Get Help section from the publishing tool so this shouldn't be particularly pressing.  (https://github.com/alphagov/collections-publisher/pull/352#pullrequestreview-105345592) 

It obviously needs to build first, but it's fine by me

andysellick:
It's failing due to an intermittent problem that @tijmenb is fixing, I'll kickstart it to see if I can get it to pass this time...

PR #241, opened 2018-03-23
Fix error styles on input component

fofr:
Spotted on https://github.com/alphagov/email-alert-frontend/pull/183 where the error message wasn't showing in red as expected.

* Rearrange styling classes so that correct red text and borders apply to field
* Update documentation to clarify why hint text is not inside a label

See: https://govuk-publishing-compon-pr-241.herokuapp.com/component-guide/input

## Before
screen shot 2018-03-23 at 15 44 17

## After
screen shot 2018-03-23 at 15 43 59

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
@@ -25,12 +25,6 @@ input[type=text].gem-c-input {
 }
 // scss-lint:enable QualifyingElement
 
-@media (min-width: 40.0625em) {
-  .govuk-c-input {
-    margin-bottom: 30px;
-  }
-}
-
 .gem-c-input:focus {
   outline: $gem-focus-width solid $gem-focus-colour;
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
@@ -22,20 +22,13 @@ input[type=text].gem-c-input {
 
   // Disable inner shadow and remove rounded corners
   appearance: none;
+
+  &.gem-c-input:focus {
+    outline: $gem-focus-width solid $gem-focus-colour;
+  }
+
+  &.gem-c-input--error {
+    border: $gem-border-width-error solid $gem-error-colour;
+  }
 }
 // scss-lint:enable QualifyingElement
-
-.gem-c-input:focus {
-  outline: $gem-focus-width solid $gem-focus-colour;
-}
-
-.gem-c-input--error {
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -22,26 +22,19 @@ input[type=text].gem-c-input {
...
+
+  &.gem-c-input--error {
+    border: $gem-border-width-error solid $gem-error-colour;
+  }
 }
 // scss-lint:enable QualifyingElement
 
 @media (min-width: 40.0625em) {
-  .govuk-c-input {
+  .gem-c-input {
```
fofr: I wasn't sure whether to delete this or update it. cc @nickcolley 
I still need to check the specificity of it too: see https://github.com/alphagov/govuk_publishing_components/pull/241/files#diff-57a28dc54e1ce322564d3c1e15811c5fR7

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -22,26 +22,19 @@ input[type=text].gem-c-input {
 
   // Disable inner shadow and remove rounded corners
   appearance: none;
+
```
fofr: Moving these into main block after specificity was updated in https://github.com/alphagov/govuk_publishing_components/commit/19035c7df55db19d11373da00727e83801619aa6

app/views/govuk_publishing_components/components/_label.html.erb
```diff
@@ -24,16 +24,16 @@
     <% end %>
   >
     <%= text %>
+    <% if hint_text.present? %>
```
nickcolley: I did a bunch of work to move the hint outside the label, tested it extensively.

But if it's easier might be better to have parity with GOV.UK Frontend, even if it's somewhat a regression.

If we do move it back in, we should remove the code and tests for this specific behaviour.

nickcolley: https://github.com/alphagov/govuk_elements/issues/574

https://github.com/alphagov/government-frontend/commit/786ab1df467373f4b2e26e83102a52e26fd03218

fofr: Thanks Nick. As you were the author I figured it was intentional but couldn't see anything in the commit history and then referred to GOV.UK Frontend as canonical source. I'll back that change out and refactor the fixes, but also make sure the accessibility context is in the docs :+1:

nickcolley: The commit history was somewhat lost by moving the work into this repository, sorry.

PR #242, opened 2018-03-26
Add shared helper for taxon breadcrumbs

tijmenb:
These breadcrumbs are shown on the new finder pages (finder-frontend) and the new topic pages (collections).

cc @steventux @leenagupte 

https://trello.com/c/WJlqy3VD

### DIFF
--- a/spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
+++ b/spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
@@ -1,0 +1,94 @@
+require 'spec_helper'
+
+RSpec.describe GovukPublishingComponents::AppHelpers::TaxonBreadcrumbs do
+  describe "Taxon breadcrumbs" do
+    it "can handle any valid content item" do
+      payload = GovukSchemas::RandomExample.for_schema(
+        frontend_schema: "taxon",
+      )
+
+      expect {
+        described_class.new(payload).breadcrumbs
+      }.to_not raise_error
+    end
+
+    it "returns the root when taxon is not specified" do
+      content_item = {
+        "title" => "Some Content",
+        "document_type" => "answer",
+        "links" => {},
+      }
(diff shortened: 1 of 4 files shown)


spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+            { title: "Home", url: "/", is_page_parent: false },
+            { title: "Another-parent", url: "/another-parent", is_page_parent: false },
+            { title: "A-parent", url: "/a-parent", is_page_parent: true },
+          ]
+        )
+      end
+    end
+
+    context 'with a taxon content item with parent taxons' do
+      it "includes parents and grandparents when available" do
```
leenagupte: I'd rename this, as grandparent taxons aren't being tested.

spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
```diff
@@ -0,0 +1,154 @@
...
+          [
+            { title: "Home", url: "/", is_page_parent: false },
+            { title: "A-parent", url: "/a-parent", is_page_parent: true },
+          ]
+        )
+      end
+    end
+
+    context 'with multiple parents' do
+      it "selects the first parent taxon in alphabetical order by title" do
```
leenagupte: I'm not sure if this test is needed. Have we decided how we're are going to handle polyhierarchy yet?

spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
tijmenb: Removed!

spec/lib/app_helpers/taxon_breadcrumbs_spec.rb
tijmenb: Good point, that was logic that was specific to the normal breadcrumbs!

PR #244, opened 2018-03-26
Merge in govuk_navigation_helpers

tijmenb:
This commit merges in the govuk_navigation_helpers gem so that we can drop the dependency.

In subsequent PRs we'll merge in the classes properly and make everything part of the same namespace, but I think this is worthwhile as a first PR.

https://trello.com/c/WJlqy3VD

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -4,7 +4,6 @@ PATH
     govuk_publishing_components (5.7.0)
       govspeak (>= 5.0.3)
       govuk_frontend_toolkit
-      govuk_navigation_helpers (~> 9.2.1)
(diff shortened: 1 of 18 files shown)


Gemfile.lock
```diff
@@ -105,14 +104,9 @@ GEM
       rubocop (~> 0.51.0)
       rubocop-rspec (~> 1.19.0)
       scss_lint
-    govuk_ab_testing (2.4.1)
```
sihugh: I guess we don't need this any more (it was probably us that added it)

tijmenb: No, there's nothing using it at the moment.

PR #247, opened 2018-03-26
Use NavigationHelper classes directly

tijmenb:
This removes some indirection by using `TaxonomySidebar`, `TaxonBreadcrumbs`, `Breadcrumbs` directly. It will make future refactoring easier.

Follows the merging in of govuk_navigation_helpers (https://github.com/alphagov/govuk_publishing_components/pull/244).

### DIFF
--- a/lib/govuk_publishing_components/presenters/contextual_navigation.rb
+++ b/lib/govuk_publishing_components/presenters/contextual_navigation.rb
@@ -14,11 +14,11 @@ module GovukPublishingComponents
       end
 
       def taxonomy_sidebar
-        nav_helper.taxonomy_sidebar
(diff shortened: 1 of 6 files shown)


lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -37,7 +37,7 @@ def breadcrumbs
             }
           ]
         else
-          nav_helper.breadcrumbs[:breadcrumbs]
+          GovukNavigationHelpers::Breadcrumbs.new(content_item).breadcrumbs[:breadcrumbs]
```
sihugh: Any reason why we don't memoize breadcrumbs?  We do everything else in this class.  Doesn't stop a merge, just a question

PR #253, opened 2018-03-28
Add link sections to taxonomy sidebar

steventux:
https://trello.com/c/Ar1mXL7Y/159-add-missing-related-links-to-taxonomy-sidebar
and
https://trello.com/c/lOtOV4wU/160-add-taxonomy-sidebar-to-all-pages-tagged-to-live-taxons

* Adds a new taxonomy-navigation component (ported from taxonomy-sidebar in alphagov/static)
* Adds policies, topical events, statistical data sets and world locations sections to the taxonomy navigation.
* Removes the constraint that taxonomy navigation only shows for guidance formats.

screenshot from 2018-03-28 09-45-20


app/views/govuk_publishing_components/components/docs/taxonomy_navigation.yml
```diff
@@ -0,0 +1,102 @@
+name: "Taxonomy navigation"
+description: "Sidebar navigation for displaying on pages tagged to the GOV.UK taxonomy."
+body: |
+  Accepts an array of items. These items are typically the list of taxons the current
```
tijmenb: Worth noting that this component is never used directly, only via the contextual sidebar component.

app/views/govuk_publishing_components/components/docs/taxonomy_navigation.yml
```diff
@@ -0,0 +1,102 @@
...
+  content page is tagged to.
+
+  Each item is a hash with a title, url, description, and list of related content
+  associated with that item.
```
tijmenb: Also worth noting it's showing policies, collections, etc.

lib/govuk_navigation_helpers.rb
```diff
@@ -0,0 +1,11 @@
+require 'active_support'
```
tijmenb: I think this file can be removed, it's probably a holdover from a rebase?

lib/govuk_publishing_components/presenters/taxonomy_navigation.rb
```diff
@@ -1,20 +1,36 @@
...
+      RELATED_SECTIONS = %w(
+        policies
+        topical_events
+        world_locations
+        statistical_data_sets
+      ).freeze
+
       def initialize(content_item)
         @content_item = ContentItem.new content_item
+        @related_navigation = GovukPublishingComponents::Presenters::RelatedNavigationHelper.new(content_item)
```
tijmenb: Maybe only use `RelatedNavigationHelper` here for brevity.

spec/lib/presenters/taxonomy_navigation_spec.rb
```diff
@@ -1,10 +1,11 @@
 require 'spec_helper'
 require 'webmock/rspec'
 require 'gds_api/test_helpers/rummager'
+require 'govuk_publishing_components/presenters/taxonomy_navigation'
```
tijmenb: This shouldn't be necessary, the requiring should happen in `govuk_publishing_components.rb`.

spec/features/contextual_navigation_spec.rb
```diff
@@ -145,15 +145,9 @@ def then_i_see_the_taxonomy_sidebar_and_collection
     # The taxonomy sidebar is currently render via a shared slimmer component.
```
vanitabarrett: We can remove these comments now the sidebar has moved and the tests have been updated?

app/views/govuk_publishing_components/components/docs/taxonomy_navigation.yml
tijmenb: And maybe that it should be merged with the related_navigation component eventually.

app/views/govuk_publishing_components/components/_taxonomy_navigation.html.erb
```diff
@@ -0,0 +1,84 @@
...
+      </div>
+    <% end %>
+
+    <% local_assigns.except(:items).each do |section_name, section| %>
+      <% if section && !section.blank? %>
+        <div>
+          <h3 class='gem-c-taxonomy-navigation__section-heading'>
+            <%= t("components.taxonomy_navigation.#{section_name}") %>
+          </h3>
+          <nav class='gem-c-taxonomy-navigation__navigation' role='navigation'>
```
vanitabarrett: We're note handling having a long list of values (this normally applies to world locations), like the related navigation component which puts them behind a show more toggle

steventux: Added here https://github.com/alphagov/govuk_publishing_components/pull/253/commits/9e5d8fb0b08e72481dd8002ee796fa9af17227d7, I need to write a feature covering this.

vanitabarrett: @steventux Worth adding an example to the documentation 👍 Also, the indentation in the view doesn't look quite right now?

steventux:
Thanks @vanitabarrett, @tijmenb I think I've covered your review comments now.

app/views/govuk_publishing_components/components/docs/taxonomy_navigation.yml
steventux: https://github.com/alphagov/govuk_publishing_components/pull/253/files#diff-da8a7099ce702fe7841db94b111e1489R14

app/views/govuk_publishing_components/components/_taxonomy_navigation.html.erb
```diff
@@ -0,0 +1,115 @@
...
+          %>
+        </h3>
+
+        <p class='gem-c-taxonomy-navigation__taxon-description'>
+          <%= item[:description] %>
+        </p>
+
+        <% item[:related_content] ||= [] %>
+        <% if item[:related_content].any? %>
+          <nav class='gem-c-taxonomy-navigation__navigation' role='navigation'>
```
vanitabarrett: Just noticed this: you need to add `data-module="toggle"` to the nav, as otherwise the toggle button will append '#' to the end of the URL instead of actually toggling the content. A toggle example should probably be added to the component guide so this can be checked when someone makes changes to the component in the future.

steventux: Good catch! I'll fix and add an example to the guide. :+1:

steventux: @vanitabarrett added and example https://github.com/alphagov/govuk_publishing_components/pull/253/commits/9776414b1cf26538219685a40bd33272d437d40c

PR #254, opened 2018-03-28
Component for a success alert

kevindew:
Trello: https://trello.com/c/QJ4DK2YQ/740-add-flashes-to-subscription-management

This adds in a component that can be used to report back to users when a
successful situation has occured - such as them submitting a form and we
informing that this was successful.

@fofr here's a draft of the component. I wasn't too sure if I wanted to specify a data-module attribute or tabindex. I'm assuming I don't need the aria labelled by since it's only a single paragraph.

### DIFF
--- a/app/views/govuk_publishing_components/components/_success_alert.html.erb
+++ b/app/views/govuk_publishing_components/components/_success_alert.html.erb
@@ -1,0 +1,8 @@
+<div
+  class="gem-c-success-alert"
+  data-module="success-alert"
+  role="alert"
+  tabindex="-1"
+>
+  <p class="gem-c-success-alert__message"><%= message %></p>
+</div>
--- a/app/views/govuk_publishing_components/components/docs/success_alert.yml
+++ b/app/views/govuk_publishing_components/components/docs/success_alert.yml
@@ -1,0 +1,19 @@
+name: Success Alert
+description: Used at the top of the page, to summarise a successful user action.
+accessibility_criteria: |
+  - should be focused on page load, to ensure the message is noticed by
+    assistive tech
+  - Should have a role of ‘alert’ to communicate that is a important and
+    time sensitive message
+examples:
+  default:
+    data:
+      message: Message to alert the user to a successful action goes here
+  long_example:
+    data:
+      message: |
+        Lorem ipsum dolor sit amet, consectetur adipiscing elit. Ut aliquet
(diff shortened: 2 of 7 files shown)


app/views/govuk_publishing_components/components/docs/success_alert.yml
```diff
@@ -0,0 +1,16 @@
+name: Success Alert
+description: Used at the top of the page, to summarise a successful user action.
+accessibility_criteria: |
+  - should be focused on page load, to ensure this error is noticed by assistive tech
```
fofr: This isn't right, it's not an error. We should probably still include `tabindex=-1` so that the alert is focusable.

kevindew: @fofr doh! missed that when copying.

I think I have three options here.

1) give it a tabindex of -1 and set up a module as per https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/error-summary.js so it's auto focused on page load
2) give it a tabindex of 0 so it is tabbable (which would be inconsistent with error-summary) 
3) Not give it one, treat it as ordinary flow

I'm tempted to go down the route of 3 as that seems the path of least risk of controversy. What do you think? I'll set this up now.

kevindew:
Component can be seen here: https://govuk-publishing-compon-pr-254.herokuapp.com/component-guide/success_alert

@fofr depending on what you think we should do about tabindex this might be good to go 🤞

app/views/govuk_publishing_components/components/_success_alert.html.erb
```diff
@@ -0,0 +1,6 @@
+<div
+  class="gem-c-success-alert"
```
fofr:
```html
<div class="gem-c-success-alert" role="alert" tabindex="-1" autofocus>
```

app/views/govuk_publishing_components/components/docs/success_alert.yml
fofr: Option 1, but without the JS:
https://github.com/alphagov/govuk_publishing_components/pull/254/files#r178077033

Docs should also indicate the autofocus behaviour.

kevindew:
@fofr cool thanks - I've made those changes.

I'm not sure the autofocus works on a div, at least not for me on chrome?

PR #255, opened 2018-03-28
Move dependency imports into _all_components.scss

kevindew:
This is to resolve an issue I encountered where some components are
relying on imports made by different ones (in my example it was creating
an alert component that was similar to error-summary, which wouldn't
work as error_summary relied on variables included by _back-link).

Rather than having individual components require the files they need the
shared pool of dependencies is imported in the all components.

The advantages to this are:

- Components aren't vulnerable to break if component ordering is changed
  or some components are removed
- Imports are included once for the collection - if any of these were to
  output CSS it'd be included multiple times

The disadvantages are:

- It's no longer explicit that a particular component has a particular
  dependency

Ideally we'd want a way to import into just a particular scope rather
than global but I'm not aware of this functionality existing in SASS.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -2,21 +2,27 @@
 // the components.
 
 // `govuk_frontend_toolkit`
-@import "measurements";
-@import "grid_layout";
-@import "typography";
 @import "colours";
+@import "css3";
+@import "design-patterns/buttons";
+@import "grid_layout";
+@import "measurements";
+@import "typography";
+
+@import "components/helpers/clearfix";
+@import "components/helpers/px-to-em";
+@import "components/helpers/variables";
 
 @import "components/back-link";
 @import "components/document-list";
 @import "components/error-summary";
+@import "components/feedback";
 @import "components/fieldset";
 @import "components/input";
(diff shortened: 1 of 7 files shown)


nickcolley:
Would a better fix to make sure each component has the right imports?

One principle I like with UI code is 'optimise for deleting code'.

Centralising imports makes it harder to know when to delete some code in this case.

Disclaimer: This is the approach we're taking with GOV.UK Frontend

kevindew:
@nickcolley that's the approach I took first - doing a rather tedious approach of commenting out all components and uncommenting one by one to see which breaks - the problem with that is there isn't really a good approach to maintain that going forward. 

I figured we're also already half way there with global imports at the start: https://github.com/alphagov/govuk_publishing_components/blob/ad17f7ad9434a44bc1412476cc5d1a0aaea22fcf/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L5-L8 

I do think ideally they'd be fully independent - but we're a bit caught in a worst of both worlds scenario now.

nickcolley:
This is how we test individual components compile (JavaScript but might be useful) https://github.com/alphagov/govuk-frontend/blob/master/tasks/gulp/__tests__/check-individual-components-compile.test.js

kevindew:
I'm going to close this as it's unlikely I'm going to find the time to look into the different approaches here.

PR #258, opened 2018-03-28
remove top padding from .gem-c-inverse-header

markhurrell:
to make the whitespace above the breadcrumb consistent with other page layouts

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
@@ -3,7 +3,7 @@
   background-color: $govuk-blue;
   color: $white;
   margin-bottom: $gutter;
-  padding: $gutter-half $gutter $gutter;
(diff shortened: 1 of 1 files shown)


vanitabarrett:
This change works on topic pages, but it's slightly more complicated because this component is also used on publications as a HTML publication header, which breaks without the padding:

screen shot 2018-03-28 at 15 26 44

We can probably add a flag to the inverse header, e.g: allow `padding-top: true` to be passed to the component which sets a `gem-c-inverse-header__padding-top` class, which sets the 15px padding-top. We do something similar with margin-bottom in [this component](https://github.com/alphagov/static/blob/master/app/views/govuk_component/title.raw.html.erb)

markhurrell:
replaced with this pull request
https://github.com/alphagov/govuk_publishing_components/pull/259

vanitabarrett:
Replaced with https://github.com/alphagov/govuk_publishing_components/pull/260

PR #260, opened 2018-03-28
Add padding-top flag to inverse header

vanitabarrett:
This will be set on topic pages so the breadcrumb renders in a consistent place with the rest of GOV.UK.

**Before:**
screen shot 2018-03-29 at 10 03 18

**After:**
screen shot 2018-03-29 at 10 02 36

Component Guide: https://govuk-publishing-compon-pr-260.herokuapp.com/component-guide/inverse-header

Topic Page: https://www.gov.uk/education
Other GOV.UK page with breadcrumbs for comparison: https://www.gov.uk/student-finance

### DIFF
--- a/app/views/govuk_publishing_components/components/_inverse_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_inverse_header.html.erb
@@ -1,10 +1,11 @@
 <%
-  full_width ||= false
-  header_class = full_width ? "gem-c-inverse-header--full-width" : ""
+  css_classes = []
+  css_classes << " gem-c-inverse-header--full-width" if local_assigns[:full_width]
+  css_classes << " gem-c-inverse-header--padding-top" unless local_assigns[:padding_top].eql?(false)
 %>
 <% block = yield %>
 <% unless block.empty? %>
-  <header class="gem-c-inverse-header <%= header_class %>">
+  <header class="gem-c-inverse-header <%= css_classes.join(' ') %> ">
     <%= block %>
   </header>
 <% end %>
--- a/spec/components/inverse_header_spec.rb
+++ b/spec/components/inverse_header_spec.rb
@@ -36,4 +36,10 @@ describe "Inverse header", type: :view do
 
     assert_select ".gem-c-inverse-header--full-width"
   end
+
+  it "renders correct css class when padding_top flag is set to false" do
+    render(component_path, padding_top: false) { block }
+
+    assert_select ".gem-c-inverse-header--padding-top", false
+  end
 end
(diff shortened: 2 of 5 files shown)


app/views/govuk_publishing_components/components/_inverse_header.html.erb
```diff
@@ -1,10 +1,12 @@
 <%
   full_width ||= false
   header_class = full_width ? "gem-c-inverse-header--full-width" : ""
+  no_padding_top ||= false
+  padding_top_class = no_padding_top ? "gem-c-inverse-header--no-padding-top" : ""
```
leenagupte: This reads a little strangely to me. If you pass `padding_top` == true, then you render the component without padding. Is that right?

steventux: An alternative to these ternary conditions would be to have an array of css classes...
```ruby
css_classes = []
css_classes << "gem-c-inverse-header--full-width" if full_width
css_classes << "gem-c-inverse-header--no-padding-top" if no_padding_top
```
then used like...
```ruby
<header class="gem-c-inverse-header<%= css_classes.join(' ') %>">
```

app/views/govuk_publishing_components/components/_inverse_header.html.erb
```diff
@@ -1,10 +1,12 @@
 <%
   full_width ||= false
   header_class = full_width ? "gem-c-inverse-header--full-width" : ""
+  no_padding_top ||= false
```
andysellick: I think this is fractionally more confusing than it could be. Could the variable be just 'padding_top', and adjust the logic accordingly?

spec/components/inverse_header_spec.rb
```diff
@@ -36,4 +36,10 @@ def component_name
 
     assert_select ".gem-c-inverse-header--full-width"
   end
+
+  it "renders correct css class when padding-top flag is set to false" do
+    render(component_path, no_padding_top: true) { block }
```
andysellick: Again, this feels slightly more confusing than my sleep-deprived mind can handle. The test name says the flag is set to false, but the value given is true?

app/views/govuk_publishing_components/components/_inverse_header.html.erb
vanitabarrett: Agreed - I put this up for review then realised it was a bit backwards. Will change :+1:

vanitabarrett:
@andysellick @steventux Thanks for your comments, I think I've addressed them all now. Please re-review when you get time

PR #263, opened 2018-03-29
Move text for JS appended elements into step nav template

andysellick:
Step by step navigation component has JS appended elements 'show' and 'hide' on all steps, plus 'show/hide all' button. Currently all that text is hard coded into the JS, which means it can't be translated in Rails.

- move text from JS to data attributes within the component template
- add translation stuff for the text (although it's still all in English)
- also simplify how the text is stored in JS and remove unnecessary distinction between 'Show all' in FE and 'Show All' in analytics - now both 'Show all'

No visual changes to the component.

Trello card: https://trello.com/c/PdxP84XT/410-refactor-task-list-component-code

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
@@ -22,6 +22,10 @@
     class="gem-c-step-nav js-hidden <% unless small %>gem-c-step-nav--large<% end %>"
     <%= "data-remember" if remember_last_step %>
     <%= "data-id=#{tracking_id}" if tracking_id %>
+    data-show-text="<%= t("govuk_component.step_by_step_nav.show", default: "Show") %>"
+    data-hide-text="<%= t("govuk_component.step_by_step_nav.hide", default: "Hide") %>"
+    data-show-all-text="<%= t("govuk_component.step_by_step_nav.show_all", default: "Show all") %>"
+    data-hide-all-text="<%= t("govuk_component.step_by_step_nav.hide_all", default: "Hide all") %>"
   >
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -22,6 +22,10 @@
     class="gem-c-step-nav js-hidden <% unless small %>gem-c-step-nav--large<% end %>"
     <%= "data-remember" if remember_last_step %>
     <%= "data-id=#{tracking_id}" if tracking_id %>
+    data-show-text="<%= t("govuk_component.step_by_step_nav.show", default: "Show") %>"
```
tijmenb: I wonder if this is the right way to push translations into the views. Not sure what the alternative is though.

I think I've seen this pattern before somewhere:

```html
<script>
var GOVUK.StepByStepTranslations = <%= t("govuk_component.step_by_step_nav").to_json %>;
</script>
```

Then in the actual JS you use `GOVUK.StepByStepTranslations.show` to get the text.

andysellick: I'm not sure I see the benefit of this option. The advantage of the existing approach is that it's tied to the component itself, and we're not adding any new variables outside of the scope of the component JS. 

Also would this work okay in the component guide - wouldn't we be recreating these variables for each instance of the component shown?

tijmenb: Yeah I'm not sure as well, I guess the `data-` thing just looks very messy in the code. Perhaps we can refactor that a bit later. Leaving this for now.

PR #268, opened 2018-04-05
Reorganise assets

tijmenb:
This PR:

- Separates the component guide JS/CSS more clearly from the JS/CSS that we include in application
- Creates a `/lib` directory for non-specific javascripts
- Fixes some double includes between the application and gem

There are no changes for consumers of this gem.

Paired with @andysellick


andysellick:
Need to update the CHANGELOG.

tijmenb:
@andysellick there shouldn't be any interface changes, so I think there's no need to write something to the changelog.

PR #271, opened 2018-04-11
Add button component from static

andysellick:
- only change is to the class name from pub-c-button to gem-c-button
- opening test has been slightly rewritten as assert_nothing_raised wasn't working
- includes an uncomfortable but necessary amount of new mixin files

Trello card: https://trello.com/c/pU7YINt1/45-move-all-components-from-static-into-the-gem


app/assets/stylesheets/govuk_publishing_components/components/helpers/_buttons.scss
```diff
@@ -0,0 +1,119 @@
+// global button styles
```
nickcolley: I think most of the file if not all of it should be removed, as it only existed in static to style buttons that are not the component.

app/assets/stylesheets/govuk_publishing_components/components/_button.scss
```diff
@@ -0,0 +1,58 @@
...
+    width: 100%;
+    text-align: center;
+  }
+
+  &:focus {
+    outline: 3px solid $focus-colour;
+  }
+}
+
+// this will be moved and extended into a model for general component spacing
```
nickcolley: Is this comment still correct?

app/views/govuk_publishing_components/components/docs/button.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  [GOV.UK Elements has more information](https://govuk-elements.herokuapp.com/buttons/) on how buttons should be used.
+
+  Note: We do not consume GOV.UK Elements directly due to the naming conventions being leaky,
+  in time this component will be a wrapper for the [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) project's button component.
+
+  Global button styles seen at `app/assets/stylesheets/helpers/_buttons.scss`
+  should be considered deprecated with this component replacing any instances.
+
+  govuk-* namespace replaced with pub-c-* otherwise we will conflict with [GOV.UK Frontend's namespace](https://github.com/alphagov/govuk-frontend/blob/master/docs/components.md)
```
nickcolley: This detailed description seems like it may make more sense to be removed as I suspect it's detailed in this repository generally?

app/assets/stylesheets/govuk_publishing_components/components/_button.scss
andysellick: It is. Funnily enough we were just talking about this earlier today. The current situation is that there's no urgency to implement this model and we'll probably look at it properly when the design system stuff starts to be integrated. In the meantime there's probably some work to be done to ensure that those components that do implement options for spacing do so in a consistent way (they might already, but we've yet to look).

nickcolley: Could change the comment to reflect your comments here?

PR #275, opened 2018-04-12
Show "normal" navigation in addition to step by step, make sidebars consistent

tijmenb:
Currently, when the page is tagged to 2,3 or 4 step by step lists, the sidebar will only show links to those steps by steps and hide the navigation that would otherwise be there.

I think it's a good change to make because this way we don't lose the normal navigation once a page is added to a step by step. It also makes the [logic for the sidebar](https://docs.google.com/document/d/1DQRTt0AYuPXSHkYvpcwPEEUDc3MHN2VubgYJA-hzeOM/edit) simpler. In the future we can integrate the code as well, making the components simpler.

## Impact

There is 1 page that will be affected: https://www.gov.uk/book-driving-test

### Before

screen shot 2018-04-13 at 12 41 31

https://govuk-publishing-components.herokuapp.com/contextual-navigation/book-driving-test

### After

https://govuk-publishing-compon-pr-275.herokuapp.com/contextual-navigation/book-driving-test

https://trello.com/c/rWqoHcOu/26-simplify-logic-around-related-links-on-content-pages

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_related-navigation.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_related-navigation.scss
@@ -9,7 +9,7 @@
 }
 
 .gem-c-related-navigation__sub-heading {
-  @include core-19;
+  @include core-16;
   border-top: 1px solid $border-colour;
   padding-top: $gutter-half;
 }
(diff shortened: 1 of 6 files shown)


tijmenb:
I've updated the screenshot in the PR.

sihugh:
This is pretty cool - we should make sure Tim LB knows about this in addition to Sam.  Is it worth an AB test?

tijmenb:
This has a 👍 from Mark/Alex, so somebody can review this.

PR #276, opened 2018-04-13
[Don't merge] Allow all types of documents to appear in taxonomy sidebar

tijmenb:
Currently we filter the related links in the taxonomy sidebar to pages that are "guidance". The definition of guidance we use was created when we had the old new education navigation pages. This supertype is being removed (https://github.com/alphagov/rummager/pull/1222, https://github.com/alphagov/govuk_document_types/pull/43), so we need to update the sidebar logic as well.

An alternative to this is to use another supertype (like [content purpose](https://docs.publishing.service.gov.uk/document-types/content_purpose_supergroup.html)), but I suspect no filtering will yield better results.

We'll have to do some before/after analysis before merging this.

https://trello.com/c/6wRHXCyT


tijmenb:
This is a pretty big change, putting on hold until the team can look at it more. It's not urgent because we haven't removed the document types.

PR #277, opened 2018-04-13
Refactor related navigation sidebar code

tijmenb:
This PR starts to refactor some of the Related navigation sidebar code. No changes in layout / behaviour are intended.

The commits have more details, but the general goal is to split up the HTML a bit and use fewer conditionals.

For review, I'm especially interested in if I kept the correct behaviour for the "other" things in the sidebar. The current implementation seems to imply that these sections are somehow special, but I can't find how they were displayed differently:

https://github.com/alphagov/govuk_publishing_components/blob/21fac6fa641339578ab6c3a85ad541592bfb97be/lib/govuk_publishing_components/presenters/related_navigation_helper.rb#L33-L40

I suspect this was a holdover from previous iterations of the sidebar. Maybe @vanitabarrett knows?


vanitabarrett:
@tijmenb I think the other section was simply put in place for styling certain sections differently

E.g: 'Elsewhere on the Web' section  should look different to 'Policy' section
screenshot_20180413-083945 01

PR #284, opened 2018-04-24
Add alpha & beta banners from static

tijmenb:
This ports over the alpha & beta banners from static. No visual changes have been made. I've refactored the two components to use the same HTML and CSS because they're identical.

I'm not a 100% sure about the sharing of CSS/HTML, comments welcome. 

https://trello.com/c/qtXCDGQI

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
@@ -1,0 +1,7 @@
+@import "mixins/touch-friendly-links";
+@import "design-patterns/alpha-beta";
+
+.gem-c-phase-banner {
+  @include phase-banner;
+  @include touch-friendly-links;
+}
--- a/app/views/govuk_publishing_components/shared/_phase_banner.html.erb
+++ b/app/views/govuk_publishing_components/shared/_phase_banner.html.erb
@@ -1,0 +1,6 @@
+<div class="gem-c-phase-banner">
+  <p>
+    <strong class="phase-tag"><%= phase %></strong>
+    <span><%= raw message %></span>
+  </p>
+</div>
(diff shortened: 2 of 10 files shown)


app/views/govuk_publishing_components/shared/_phase_banner.html.erb
```diff
@@ -0,0 +1,6 @@
+<div class="gem-c-phase-banner">
+  <p>
```
andysellick: Ideally wherever styles need to be applied we should be using an explicit class, but I appreciate that this CSS is being imported wholesale from somewhere else. It's probably okay, because the component's CSS is still namespaced to gem-c, but I think this should at least be recorded as technical debt.

app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
```diff
@@ -0,0 +1,7 @@
+@import "mixins/touch-friendly-links";
+@import "design-patterns/alpha-beta";
```
andysellick: We might want to look into this in a bit more depth as this file seems to be pulling in a lot of stuff - https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/design-patterns/_alpha-beta.scss#L1-L4. Will that cause duplication in the CSS? Is all of this needed?

tijmenb: Commenting here because it's still relevant in https://github.com/alphagov/govuk_publishing_components/pull/333. GOV.UK Frontend toolkit doesn't include concrete classes, only mixins. This means that there won't be any extra CSS added because of these imports.

app/views/govuk_publishing_components/shared/_phase_banner.html.erb
tijmenb: Yeah agree. Basically GOV.UK Frontend toolkit is future tech debt, once we introduce GOV.UK Frontend.

PR #285, opened 2018-04-25
Show breadcrumbs on non live taxons

cbaines:
When looking at a taxon, it's useful to see the breadcrumbs regardless
of the phase it's in as without the breadcrumbs it's hard to navigate
up the taxonomy.

Therefore, apply the restriction regarding the phase only if the
content item isn't a taxon.

## Before
before

## After
after


cbaines:
> The conditional isn't necessary, you can just remove the phase_is_live? stuff. This class is only used on topic pages and taxonomy-based finders, where we'd want this behaviour.

Ok, I think I follow. I've updated the changes to remove the `phase` stuff.

PR #287, opened 2018-04-27
RFC 1: Add government department colours to components

andysellick:
👉 [Rendered version](https://github.com/alphagov/govuk_publishing_components/blob/91ebfd55b3d553aedb14410623dc925e7224846c/docs/rfc-001.md)

### DIFF
--- a/docs/rfc-001.md
+++ b/docs/rfc-001.md
@@ -1,0 +1,124 @@
+# Add government department brand colours to gem for use by components
+
+## Problem and background
+
+Organisation pages ([example](https://www.gov.uk/government/organisations/attorney-generals-office)) have their own brand, which includes specific colours applied to links and borders. These colours are pulled in from [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_organisation.scss) using a [mixin](https://github.com/alphagov/whitehall/blob/master/app/assets/stylesheets/frontend/views/_organisations.scss#L855-L858) to generate a range of classes.
+
+Most organisations have one colour for borders and one for links, even if they are broadly the same colour. These are applied to the existing structure of the frontend of whitehall.
+
+The structure of the new organisation pages (to be built in collections) uses components, which are individual and isolated. To reproduce this approach in the same way would mean adding all of the CSS for all of the colour options to all of the components, dramatically increasing the size of the CSS involved and introducing much duplication.
+
+## Proposed solution
+
+### In the gem
+
+All colour classes are included in a helper Sass file in the gem.
+
+Sass in the gem looks like this:
+
+```
+@import "colours/organisation";
+
+@mixin organisation-brand-colour($selector, $property, $websafe: false) {
+  @each $organisation in $all-organisation-brand-colours {
+    .color-#{$selector}--#{nth($organisation, 1)} .color-#{$selector} {
+      @if $websafe {
+        #{$property}: nth($organisation, 3);
+      }
+      @else {
+        #{$property}: nth($organisation, 2);
+      }
+    }
+  }
+}
+
+@include organisation-brand-colour("link", color, true);
+@include organisation-brand-colour("border", border-color);
+```
+
+This is included in the `_all_components.scss` file.
+
+CSS in the gem looks like this:
+
+```
+.color-link--department-for-work-and-pensions .color-link {
+  border-color: #8f1111;
+}
+
+.color-border--department-for-work-and-pensions .color-border {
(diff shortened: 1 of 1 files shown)


Davidslv:
There's one thing that comes to my mind, we are using the name of the departments to set their colours, when/if a department changes their name we have to come back and rename the css rules to a new name.

Question: wouldn't be better to give them the option to choose a colour that we already have rules for? E.g: Dropdown with green,brown, red, we would then have `.green, .brown, .red` css classes. This would require an extra field on the schema (which you mentioned already wasn't ideal) but wouldn't it be better than having to have a ticket to rename or change a department colour, or add a new one?

andysellick:
@Davidslv that's a good point, if we had generic colour classes not tied to department names that would be simpler, and avoid duplication (some of the current colours are duplicated already).

We would need to change the mechanism by which the colours are delivered, which could be tricky - I think it's in use in more places than whitehall, so it would mean creating a new mechanism alongside the current one and having them running in parallel for a while, potentially. I'd be slightly concerned about the creation of tech debt there.

vanitabarrett:
@andysellick I agree with @Davidslv - I think it makes a lot of sense to be using generic colour variables rather than tying them in with department names. 

I'm not 100% sure of how this currently works and what this change would mean in terms of tech debt/how complicated it is to implement, but I think if it's a good solution then running 2 mechanisms in parallel for a while to get this working is a good trade-off, we just need to make sure we make a note of any tech debt to revisit.

Davidslv:
Btw, I just want to give some heads up that this has been kinda of hardcoded in the backend, there's a class for the Organisations Brand Colour which is based on their name... 

I hope we can try to take this entropy away if we can, please :)

https://github.com/alphagov/whitehall/commit/d5cf054b5ce8ea78afd1972d0ec08167932253cb

andysellick:
This is good feedback. Some brief thoughts.

- Doing it this way would definitely be more work, both for frontend and backend. That isn't necessarily a reason to not do it though.
- We'd need to have all or part of the class name in the content item, so that it could be passed to  components. How much effort would it be to add this @Davidslv ?
- We'd need a naming convention for classes. `.colour-FF0000` obviously isn't going to work, but I worry about getting into shades of different colours e.g. `.colour-blue`, `.colour-light-blue`, `.colour-lighter-blue`, particularly if other colours get introduced later. Any other alternatives?
- the basis for these colours currently is in the toolkit and pulled in by `government-frontend`, `whitehall`, `govuk_prototype_kit`, `govuk_developer_docs` and possibly by `govuk_frontend_toolkit` itself, so we'd need a plan for those at the end of such a change.

In terms of the RFC, if this change broadly fits in with the rest of the plan, does anyone have any other comments on any other part of it?

docs/rfc-001.md
```diff
@@ -0,0 +1,124 @@
...
+.color-border--department-for-work-and-pensions .color-border {
+  border-color: #9f1888;
+}
+```
+
+etc.
+
+### In the component
+
+Pass component parameters for link_colour and border_color as required, as strings that match the pre-defined list e.g. 'department-for-work-and-pensions'.
```
tijmenb: Do we need to be specific about the intention of the passed in variables? 

So instead of:

```rb
<%= render 'govuk_publishing_components/components/title', link_colour: 'department-for-education', border_color: 'department-for-education' %>
```

Do this:

```rb
<%= render 'govuk_publishing_components/components/title', brand: 'department-for-education' %>

Comment (andysellick) [reply docs/rfc-001.md]: Yeah, that's simpler.

**Comment by nickcolley** [comment]
Could the designs of these pages be reviewed?

Is there a user need?

Is there a way to avoid these odd pages while also retaining the department branding?

**Comment by tijmenb** [comment]
@nickcolley those questions are out of scope for this RFC.

File: docs/rfc-001.md [hunk 2]
```diff
@@ -0,0 +1,124 @@
...
+@include organisation-brand-colour("link", color, true);
+@include organisation-brand-colour("border", border-color);
+```
+
+This is included in the `_all_components.scss` file.
+
+CSS in the gem looks like this:
+
+```
+.color-link--department-for-work-and-pensions .color-link {
```
nickcolley: Could consider `theme` or `brand`, for the namespace, to be consistent with @tijmenb 's suggestion.

andysellick:
Thanks for all the feedback everyone. I've made a decision and have opened a new PR with the changes. I've not implemented all the suggestions made here: if you disagree with the approach I'm happy to discuss further but please have a specific suggestion as to how your proposal could be implemented as I've got a lot to do this mission. Thanks!

PR #289, opened 2018-04-30
Add translation nav component

andysellick:
Moving from government-frontend to the gem. A separate PR will be raised to remove it from government-frontend. No changes to the component other than changing the CSS namespace from app-c to gem-c.

Trello cards:

- https://trello.com/c/ehg2nkE9/53-modify-component-translation-nav
- https://trello.com/c/0zt60phw/20-move-the-translation-nav-component-to-the-gem

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_translation-nav.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_translation-nav.scss
@@ -1,0 +1,35 @@
+.gem-c-translation-nav {
+  @include responsive-top-margin;
+  @include core-16;
+  margin-bottom: $gutter;
+  border-bottom: 1px solid $border-colour;
+}
+
+.gem-c-translation-nav__list {
+  @extend %contain-floats;
+  list-style: none;
+  margin-left: -$gutter-one-third;
+  margin-right: -$gutter-one-third;
+}
+
+.gem-c-translation-nav__list-item {
+  float: left;
+  padding-left: $gutter-one-third;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_translation-nav.scss
```diff
@@ -0,0 +1,35 @@
+.gem-c-translation-nav {
+  margin-bottom: $gutter;
+  border-bottom: 1px solid $border-colour;
+  @include responsive-top-margin;
+  @include core-16;
```
vanitabarrett: I know this is just a straight copy from government-frontend, but makes sense to tidy up while we're here - in this case, could use your suggestion of always putting `@includes` at the top of a CSS block

spec/components/translation_nav_test_spec.rb
```diff
@@ -0,0 +1,72 @@
+require "rails_helper"
+
+describe "Search", type: :view do
```
vanitabarrett: Is this a copy paste error?

app/assets/stylesheets/govuk_publishing_components/components/_translation-nav.scss
andysellick: Good shout.

spec/components/translation_nav_test_spec.rb
andysellick: D'oooooooh..

PR #290, opened 2018-04-30
Add subscription links component

andysellick:
Moving from government-frontend to the gem. A separate PR will be raised to remove it from government-frontend. No changes to the component other than changing the CSS namespace from app-c to gem-c.

Trello cards:

- https://trello.com/c/NjgS3vym/55-modify-component-get-updates
- https://trello.com/c/ezIOl79E/21-move-the-subscription-links-component-to-the-gem


app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
```diff
@@ -0,0 +1,40 @@
...
+    background-repeat: no-repeat;
+    background-position: 0 20%;
+
+    @include media(tablet) {
+      background-position: 0 35%;
+    }
+  }
+
+  .gem-c-subscription-links__link--feed {
+    background-image: image-url("feed-icon-black.png");
```
vanitabarrett: We need to copy over these icons into the gem as well

app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
```diff
@@ -0,0 +1,40 @@
...
+    list-style: none;
+    margin-left: -$gutter-half / 2;
+    margin-right: -$gutter-half / 2;
+  }
+
+  .gem-c-subscription-links__list-item {
+    display: inline-block;
+    margin-left: $gutter-half / 2;
+    margin-right: $gutter-half / 2;
+    margin-bottom: $gutter / 2;
```
vanitabarrett: Isn't this just `$gutter-half` ?

app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
andysellick: Ah, can't believe I missed that. Actually on second thought I can, worth noting... I think I missed this because I was moving the gem while using the component guide in government-frontend for reference, so it was still picking up the images from government-frontend. Fixed.

app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
andysellick: Yeah, that's weird. Have changed.

PR #292, opened 2018-05-01
RFC 2: Add spacing (margins) model to components

andysellick:
👉 [Rendered version](https://github.com/alphagov/govuk_publishing_components/blob/97e81931ed339830a2fc75ccdaad78b206f8d8e7/docs/rfc-002.md)

### DIFF
--- a/docs/rfc-002.md
+++ b/docs/rfc-002.md
@@ -1,0 +1,95 @@
+# Create a universal component model for spacing
+
+## Problem and background
+
+We need a model for applying different spacing to components depending on their context, for example adding or removing bottom margin. This issue has been [extensively documented](https://trello.com/c/KEkNsxG3/60-implement-customisable-spacing-for-components) prior to now, but was put on hold as there was no urgent need for it.
+
+Organisation pages are being migrated out of whitehall and this has prompted the reopening of this issue. Specifically it is intended that the heading component be extended for use throughout these pages, which will require varying spacing being applied to it in different places.
+
+In addition, the recent [RFC relating to organisation pages colours](https://github.com/alphagov/govuk_publishing_components/pull/287) may provide a similar solution to this issue.
+
+## Proposed solution
+
+A helper sass file in the gem that provides shared classes for any component that wants them. The spacing model itself would be based on the `$gutter` variables and would be reduced by a third on mobile.
+
+Spacing model:
+
+| Class                          | Value   | Desktop             | Mobile                         |
+| ------------------------------ | ------- | ------------------- | ------------------------------ |
+| margin-top--gutter             | 1       | $gutter             | $gutter-two-thirds             |
+| margin-top--gutter-two-thirds  | 2       | $gutter-two-thirds  | ($gutter-two-thirds / 3) * 2   |
+| margin-top--gutter-one-half    | 3       | $gutter-one-half    | ($gutter-one-half / 3) * 2     |
+| margin-top--gutter-one-third   | 4       | $gutter-one-third   | ($gutter-one-third / 3) * 2    |
+| margin-top--gutter-one-quarter | 5       | $gutter-one-quarter | ($gutter-one-quarter / 3) * 2  |
+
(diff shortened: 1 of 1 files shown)


vanitabarrett:
I'm not sure about the using a "value" to set the margins - the 1-5 system seems fairly arbitrary and is confusing to understand. I'd find it confusing to supply '3' to get a half margin, for example. 

Along the same lines, I'm also not sure on the readability of `margins: [1, 0, 2, 0]`. Someone new to GOV.UK/our frontend wouldn't be able to understand that without cross-referencing with some docs somewhere.

**A couple of questions, more than comments:**

- What happens if someone supplies `margins: [2]`? How do we interpret this? Do we follow CSS patterns or only process the margin values we have? (so `margin: 2, 2, 2, 2` vs `margin: 2, 0, 0, 0`)
- Will this pattern work for components which, for some reason, have unusual spacing needs? E.g: values outside our spacing grid or negative margins. Or do we just not allow this?

maxgds:
I agree with Vanita on the arbitrary values mapping to two-thirds etc - I saw that you'd raised it as a possible concern already, but I do think it's somewhat opaque in meaning and there would be a fair bit of digging around to find the relevant mapping when building or implementing a new component. It would be good to tidy up the differences between components (false vs 0 vs "" for flags, for example) but I think it needs to be more understandable at a glance than this proposition.

andysellick:
I agree with all of these points. Does anyone have any suggestions?

andysellick:
I think any implementation we come up is likely to be complicated enough that it will need to be documented for developers to be able to add it to any new or existing components.

I guess an alternative is to take the smallest value (say, gutter-one-quarter) and do the scale as multiples of that. So 1 would be one quarter, 2 would be two quarters, etc. It still doesn't match exactly but it makes a little bit more sense.

tijmenb:
Related:

https://medium.com/fed-or-dead/handling-spacing-in-a-ui-component-library-70f3b22ec89

nickcolley:
Awesome RFC, great to see something so well thought out. 👍 

---

Alternatives to the proposed kind of scale include 't-shirt' sizing, but that can get confusing with 'xxxl' or 'xxxs' etc.

---

In terms of the interface for the spacing, you could consider a more boring approach.

For the most part, vertical rhythm is best set via margin on the bottom of components.

With this in mind, the majority of times you'll only want to set `margin-bottom`.

So with this in mind you could have `margin_bottom: N` where N is a number from your scale  (and so on for the other sides).

---

Your questions:

> we already have individual ways in which components handle margin, so this is just adding another one that may or may not be used by more than one component

This proposal could deprecate those methods, it may be easy to replace them if this RFC is accepted.

> aligning the classes to a scale like this (1 to 5) is a little confusing as it doesn't correspond to the numbers involved

It'd be good to have you folks see the spacing scale in the design system, it aligns well with your proposal.

> this model only adds margin, it provides no option to have no margin i.e. it assumes no margin as the default. Components that already have margins by default may require more work to adapt to using this system.

Since most components can have the same default margin, we're setting the same margin by default.

This means for the most part you don't have to worry about spacing for vertical rhythm, and add spacing in for the cases where you need something different.

This is the same way the current components with spacing work, where they can have their spacing set to 0.

---

Finally, since the new GOV.UK Design System has a spacing scale that aims to solve this same issue, would be good to align on this if possible.

I think it would provide a similar solution as described in this RFC.

andysellick:
Thanks for all the input everyone. I'm going to close this RFC for now. I think that while a system like this to apply margin to all components is possible, due to the complexity of the existing margins on components and how closely that's tied to the design of the site this issue is larger than just FE.

My recommendation is that we plan to take a look at this again later, perhaps as part of a mission team, and include a proper audit of the existing components and the design context in which they're used. My hope is that as GOV.UK becomes more componentised over time this will be easier to solve.

For now I'm going to try a model similar to the one proposed on one component (heading) and see how well it works.

PR #294, opened 2018-05-02
Modify subscription links component

andysellick:
The subscription links component is going to be used on the new organisation pages, but needs a few extra features adding.

- add option to change link text
- add option to turn feed link into a toggle to display an input box containing a link
- includes toggle.js from static, included without changes (apart from the namespace) as a module outside of the component (so it can be used by other components if required)

Component guide link: https://govuk-publishing-compon-pr-294.herokuapp.com/component-guide/subscription-links

Trello card: https://trello.com/c/NjgS3vym/55-modify-component-subscription-links

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
@@ -27,6 +27,15 @@
 
   .gem-c-subscription-links__link--feed {
     background-image: image-url("govuk_publishing_components/feed-icon-black.png");
+
+    // if this is a toggle, only show if js is enabled
+    &[data-controls] {
+      display: none;
+
+      .js-enabled & {
+        display: block;
+      }
+    }
   }
 
   .gem-c-subscription-links__link--email-alerts {
@@ -37,4 +46,15 @@
       background-size: 20px 14px;
     }
   }
+
+  .gem-c-subscription-links__feed-box {
+    padding: $gutter-half;
+    margin-bottom: $gutter-half;
+    background: $grey-3;
+  }
+
+  .gem-c-subscription-links__feed-description {
+    font-weight: bold;
+  }
 }
+
--- a/app/views/govuk_publishing_components/components/_subscription-links.html.erb
+++ b/app/views/govuk_publishing_components/components/_subscription-links.html.erb
@@ -1,21 +1,50 @@
 <%
   email_signup_link ||= false
+  email_signup_link_text ||= t("govuk_component.subscription_links.email_signup_link_text", default: "Get email alerts")
   feed_link ||= false
+  feed_link_text ||= t("govuk_component.subscription_links.feed_link_text", default: "Subscribe to feed")
+  feed_link_box_value ||= false
+  feed_link_data = false
+
+  if feed_link_box_value
+    feed_link = "#"
+    feed_link_data = {
+      controls: "feed-reader",
+      expanded: "false"
+    }
+  end
 %>
-<% if email_signup_link || feed_link %>
-  <section class="gem-c-subscription-links">
-    <h2 class="visuallyhidden">Subscriptions</h2>
+<% if email_signup_link || feed_link || feed_link_box_value %>
+  <section class="gem-c-subscription-links" data-module="gem-toggle">
+    <h2 class="visuallyhidden"><%= t("govuk_component.subscription_links.subscriptions", default: "Subscriptions") %></h2>
     <ul class="gem-c-subscription-links__list">
       <% if email_signup_link.present? %>
         <li class="gem-c-subscription-links__list-item">
-          <%= link_to "Get email alerts", email_signup_link, class: "gem-c-subscription-links__link gem-c-subscription-links__link--email-alerts" %>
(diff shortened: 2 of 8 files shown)


app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -1,21 +1,50 @@
...
+              class: "gem-c-subscription-links__link gem-c-subscription-links__link--feed",
+              data: feed_link_data
+          %>
         </li>
       <% end %>
     </ul>
+
+    <% if feed_link_box_value %>
+      <div class="gem-c-subscription-links__feed-box js-hidden" id="feed-reader">
+        <p class="gem-c-subscription-links__feed-description js-hidden"><%= feed_link_text %></p>
```
vanitabarrett: It looks like the `js-hidden` means that this is never shown? 

See this example in the component guide: https://govuk-publishing-compon-pr-294.herokuapp.com/component-guide/subscription-links/with_copyable_feed_link/preview

app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
```diff
@@ -37,4 +46,15 @@
...
   }
+
+  .gem-c-subscription-links__feed-box {
+    padding: $gutter-half;
+    margin-bottom: $gutter-half;
+    background: $grey-3;
+  }
+
+  .gem-c-subscription-links__feed-description {
+    font-weight: bold;
```
vanitabarrett: Should we be using our font mixins here?

app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -1,21 +1,50 @@
 <%
   email_signup_link ||= false
+  email_signup_link_text ||= "Get email alerts"
```
vanitabarrett: Should this and the default feed link text be in locale files for translation?

app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -1,21 +1,50 @@
...
+      controls: "feed-reader",
+      expanded: "false"
+    }
+  end
 %>
-<% if email_signup_link || feed_link %>
-  <section class="gem-c-subscription-links">
+<% if email_signup_link || feed_link || feed_link_box_value %>
+  <section class="gem-c-subscription-links" data-module="gem-toggle">
     <h2 class="visuallyhidden">Subscriptions</h2>
```
vanitabarrett: Same here for using translated string

app/views/govuk_publishing_components/components/docs/subscription-links.yml
```diff
@@ -15,3 +23,17 @@ examples:
...
       feed_link: '/foreign-travel-advice/singapore.atom'
+  with_custom_text:
+    data:
+      email_signup_link: '/foreign-travel-advice/singapore/email-signup'
+      email_signup_link_text: 'Get notifications'
+      feed_link: '/foreign-travel-advice/singapore.atom'
+      feed_link_text: 'View feed'
+  with_copyable_feed_link:
+    description: |
+      This option changes the feed link to a toggle control, which opens a hidden element containing an input containing the value passed to the component, usually a URL to an atom feed. This uses the [form input](/component-guide/input) component.
```
vanitabarrett: Nitpick: this reads a bit weirdly "..which opens a hidden element **containing** an input **containing** the value...". 

Could we say "..which opens a hidden element **containing** an input **prepopulated with** the value...

app/views/govuk_publishing_components/components/_subscription-links.html.erb
andysellick: If Javascript is disabled, it gets shown and the main feed link gets hidden. Seemed like a sensible way to keep the text but remove the link/button.

app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
andysellick: I think the only way to do that is to specify a font size as well? I deliberately didn't, because I didn't want to set a font size, only the weight.

app/views/govuk_publishing_components/components/_subscription-links.html.erb
andysellick: Not a bad idea, will add.

app/views/govuk_publishing_components/components/docs/subscription-links.yml
andysellick: Much better, thanks.

app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -1,21 +1,50 @@
 <%
   email_signup_link ||= false
+  email_signup_link_text ||= t("govuk_component.subscription_links.email_signup_link_text", default: "Get email alerts")
   feed_link ||= false
+  feed_link_text ||= t("govuk_component.subscription_links.feed_link_text", default: "Subscribe to feed")
+  feed_link_box_value ||= false
```
tijmenb: This is way too late (again), but will `feed_link_box_value` ever be different from `feed_link`? Perhaps a boolean param to show the box or not would make it easier to use.

PR #296, opened 2018-05-03
Add department colours to components

andysellick:
This PR follows the discussion over [this RFC](https://github.com/alphagov/govuk_publishing_components/pull/287) to add department colour branding to components. 

The main point of discussion in the RFC centred around using colour specific classes instead of department ones, to avoid duplication and make the classnames more meaningful. I have decided not to implement this for the following reasons:

- coming up with classnames based on colours is hard i.e. `colour--light-red`, `colour--lighter-red` doesn't scale well if new colours are needed later, and neither does `colour--red-1`, `colour--red-2` etc.
- changing the backend so all the organisations could have this different approach to branding would have involved a lot of work that we don't have time for this mission

I've written the code to be as flexible as possible (classes are all output from a helper method) so if a new approach is needed in future it shouldn't be too difficult to adapt. This change therefore broadly follows the originally proposed idea, with some specific changes:

- the mixin generates a single parent class containing child classes for font colour (using the websafe colour) and border colour (using the secondary colour)
- branding classes have been improved to be more BEM-like

In total this adds around 4KB to the CSS in the gem. 

This PR includes two commits; one to add the colours and one to show [how the model works on a component](https://govuk-publishing-compon-pr-296.herokuapp.com/component-guide/document_list) (note that the link to documentation in it won't work yet).

Trello card: https://trello.com/c/r4PLGbH5/58-create-a-branding-model-for-components

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_organisation-colours.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_organisation-colours.scss
@@ -1,0 +1,20 @@
+@import "colours/organisation";
+
+// this uses the colours from govuk_frontend_toolkit
+// https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_organisation.scss
+
+@mixin organisation-brand-colour() {
+  @each $organisation in $all-organisation-brand-colours {
+    .brand--#{nth($organisation, 1)} {
+      .brand__color {
+        color: nth($organisation, 3);
+      }
+
+      .brand__border-color {
+        border-color: nth($organisation, 2);
+      }
+    }
+  }
+}
+
+@include organisation-brand-colour;
--- a/docs/component_branding.md
+++ b/docs/component_branding.md
@@ -1,0 +1,47 @@
+# Component branding
+
+## Overview
+
+Organisation pages in whitehall ([example](https://www.gov.uk/government/organisations/attorney-generals-office)) have department-specific branding on them. This includes specific colours for links and borders. These pages are being migrated out of whitehall and will use components for their frontend, which means some components need a sensible way of displaying these colours.
+
+This work follows [this RFC](https://github.com/alphagov/govuk_publishing_components/pull/287) to discuss the approach taken.
+
+## Adding to a component
+
+To add colours to a component, modify the component to follow the example below.
+
+```
+<%
+  brand ||= false
+  brand_helper = GovukPublishingComponents::Presenters::BrandHelper.new(brand)
+%>
+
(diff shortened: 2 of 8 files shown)


app/assets/stylesheets/govuk_publishing_components/components/helpers/_organisation-colours.scss
```diff
@@ -0,0 +1,20 @@
...
+      }
+
+      .brand__border-color {
+        border-color: nth($organisation, 2);
+      }
+    }
+  }
+}
+
+@include organisation-brand-colour();
```
vanitabarrett: Not sure we need the ()?

vanitabarrett:
I like this approach, I think it makes sense to use the logic in `govuk_frontend_toolkit`. Perhaps in the future we can revisit this and see how we can make it more specific to colour rather than department, but this seems a good solution for now to prevent blocking your work.

Happy to re-review and 👍 once you've added some documentation for the document list component using the department colours

docs/component_branding.md
```diff
@@ -0,0 +1,47 @@
...
+
+To add colours to a component, modify the component to follow the example below.
+
+```
+<%
+  brand ||= false
+  brand_helper = GovukPublishingComponents::Presenters::BrandHelper.new(brand)
+%>
+
+<div class="gem-c-component <%= brand_helper.get_brand %>">
```
tijmenb: We mostly [avoid prefixing getters with `get_`](https://github.com/bbatsov/ruby-style-guide#accessor_mutator_method_names) (I'm not sure why Rubocop hasn't picked that up, I'll investigate).

To me `get_brand` also implies that it'll return the `brand` passed in. Perhaps `brand_class` would reveal the intention more?

docs/component_branding.md
```diff
@@ -0,0 +1,47 @@
...
+  <div class="gem-c-component__title <%= brand_helper.get_brand_element("border-color") %>">
```
tijmenb: I think it's worth it creating separate method for this and `color` to avoid typos.

We mostly [avoid prefixing getters with `get_`](https://github.com/bbatsov/ruby-style-guide#accessor_mutator_method_names) (I'm not sure why Rubocop hasn't picked that up, I'll investigate).

Perhaps something like:

```diff
- <%= brand_helper.get_brand_element("border-color") %>">
+ <%= brand_helper.border_color_class %>
```

```diff
- <%= brand_helper.get_brand_element("color") %>">
+ <%= brand_helper.color_class %>
```

docs/component_branding.md
andysellick: Great suggestions, thanks. See this PR -> https://github.com/alphagov/govuk_publishing_components/pull/300

PR #299, opened 2018-05-08
Subscription links add colour and tracking

andysellick:
Adds colour branding options and the ability to add tracking to the subscription links component.

Trello card: https://trello.com/c/NjgS3vym/55-modify-component-subscription-links

PR #298, opened 2018-05-04
Translation nav add brand and tracking

andysellick:
Adds the following options to the translation nav component:

- branding (to change the colour of the links given a parameter matching a department name, ahead of using this component on organisation pages)
- tracking (to add any passed GA tracking attributes to each link, if passed to the component)

Preview: https://govuk-publishing-compon-pr-298.herokuapp.com/component-guide/translation-nav
Trello card: https://trello.com/c/ehg2nkE9/53-modify-component-translation-nav

PR #300, opened 2018-05-09
Iterate branding model

andysellick:
Improve the colour branding model following @tijmenb 's comments on the [PR to add it](https://github.com/alphagov/govuk_publishing_components/pull/296).

Note that there are a few components that have started using this model already, I'll need to rebase and update these PRs once this change is approved.

- https://github.com/alphagov/govuk_publishing_components/pull/298
- https://github.com/alphagov/govuk_publishing_components/pull/299
- https://github.com/alphagov/collections/pull/651

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,7 @@
 # Unreleased
 
-* Add an optional `canonical` meta tag.
+* Add an optional `canonical` meta tag (PR #302)
+* Iterate branding model (PR #300)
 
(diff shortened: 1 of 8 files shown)


andysellick:
@vanitabarrett I'd originally planned to make it component specific but couldn't think of a way of doing it that didn't result in duplicating all of the branding styles for every component. This way it can be shared between more than one component without a hit to the size of the CSS.

PR #302, opened 2018-05-10
Add a canonical URL meta tag

thomasleese:
This is built from the base_path of the content item which means it strips out any query parameters or fragments.

[Trello Card](https://trello.com/c/J67JSy8r/148-investigate-how-to-implement-canonical-tags-for-government-frontend-to-help-remove-duplicate-urls-2-hour-timebox)

### DIFF
--- a/lib/govuk_publishing_components/presenters/meta_tags.rb
+++ b/lib/govuk_publishing_components/presenters/meta_tags.rb
@@ -20,6 +20,7 @@ module GovukPublishingComponents
         meta_tags = add_political_tags(meta_tags)
         meta_tags = add_taxonomy_tags(meta_tags)
         meta_tags = add_step_by_step_tags(meta_tags)
+        meta_tags = add_canonical_tag(meta_tags)
         meta_tags
       end
 
@@ -103,6 +104,20 @@ module GovukPublishingComponents
         stepnavs = links[:part_of_step_navs] || []
         stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
         meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content.present?
+        meta_tags
+      end
+
+      def add_canonical_tag(meta_tags)
+        if local_assigns.key?(:canonical_path)
+          canonical_path = if local_assigns[:canonical_path] == true
+                             content_item[:base_path]
+                           else
+                             local_assigns[:canonical_path]
+                           end
+
+          meta_tags["canonical"] = Plek.new.website_root + canonical_path
+        end
+
         meta_tags
       end
 
--- a/spec/components/meta_tags_spec.rb
+++ b/spec/components/meta_tags_spec.rb
@@ -31,6 +31,30 @@ describe "Meta tags", type: :view do
     assert_meta_tag('govuk:schema-name', 'publication')
   end
 
+  it "renders the canonical URL in a meta tag if requested" do
+    render_component(
+      content_item: { base_path: '/test' },
+      canonical_path: true,
+    )
+
+    assert_meta_tag('canonical', 'http://www.dev.gov.uk/test')
+  end
+
+  it "renders the canonical URL in a meta tag if overridden" do
+    render_component(
+      content_item: { base_path: '/test' },
+      canonical_path: '/test2',
+    )
+
+    assert_meta_tag('canonical', 'http://www.dev.gov.uk/test2')
+  end
+
+  it "doesn't renders the canonical URL in a meta tag if not requested" do
+    assert_empty render_component(content_item: {
+      base_path: '/test'
+    }).strip
+  end
+
   it "renders organisations in a meta tag with angle brackets" do
     content_item = {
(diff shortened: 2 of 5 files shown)


kevindew:
Would this be for every page? I don't know a huge amount about govuk_publishing_components so I may be wrong but off the top of my head pages like https://www.gov.uk/foreign-travel-advice/thailand/safety-and-security have different path to base_path (basically any with a prefix base_path)

thomasleese:
@kevindew Yeah, I think it should be for every page. I did have a think about whether to use `base_path` or take the request page. I stuck with `base_path` as I think that's actually a better definition of the canonical URL, in the case of foreign travel advice I think you could argue that it makes sense for `/foreign-travel-advice/thailand` to be the canonical URL for `/foreign-travel-advice/thailand/safety-and-security` but of course we have a problem when the canonical URL ends up being `/government`. Do you think it would be better to use the request path?

sihugh:
Might be worth talking to a content designer about this?  Mainstream guides have a bunch of parts with the same base path but different slugs.  Some are more coherent as a single entity than others.  I'd recommend Jon Sanger

thomasleese:
@kevindew In terms of query strings, we've decided that content where the user can filter shouldn't be index and that's covered in https://trello.com/c/PHjO8hP2/139-noindex-pagination-from-external-search-2. Smart answers uses paths not query strings to indicate content so that shouldn't be a problem.

I'm more in favour of using the request path now as the canonical URL, but it's true that applying this to all pages there are inevitably cases where either option doesn't work and could cause harm. We could have logic in the component that understands how to build the canonical URL of different document types, but that does feel better suited in the individual frontend apps.

Based on this and our earlier discussion on Slack @bevanloon and @tijmenb what do you think?

thomasleese:
(For context, @kevindew, my initial idea was to have a blacklist of query parameters we know that we want to remove from the request URL to build a canonical URL (i.e. `utm_`) and then leave the rest untouched, meaning we get a slight improvement on the situation now and we can add to that blacklist as we see fit.)

kevindew:
okey doke, cool re query strings.

If we use request path it feels like it's sufficiently dumb it might not provide much value to me. As it stands google uses sitemap to work out what it thinks is the canonical path so I can't imagine it's getting caught mistaking analytics parameters often and then that means if we have instances of rendering same content on multiple pages they'll be flagged as having different canonical paths.

Sorry Tom, I know I'm getting more caught up in the value of the task than the implementation which isn't particularly helpful.

lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -106,6 +107,14 @@ def add_step_by_step_tags(meta_tags)
         meta_tags
       end
 
+      def add_canonical_tag(meta_tags)
```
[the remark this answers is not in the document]
thomasleese: I've pushed up a change. It makes the attribute configurable but if you leave it as `true` it sticks with the default base path (to avoid having to add too much logic to the frontend apps), what do you think?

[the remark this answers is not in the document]
kevindew: That change looks good to me 👍 - although as it's a path I think we should still prefix with Plek website root when it's configured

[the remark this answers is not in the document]
thomasleese: Good point, I've amended that now!

spec/components/meta_tags_spec.rb
```diff
@@ -31,6 +31,30 @@ def example_document_for(schema_name, example_name)
...
+      canonical_path: true,
+    )
+
+    assert_meta_tag('canonical', 'http://www.dev.gov.uk/test')
+  end
+
+  it "renders the canonical URL in a meta tag if overridden" do
+    render_component(
+      content_item: { base_path: "/test" },
+      canonical_path: "http://www.dev.gov.uk/test2",
```
kevindew: just this to go 😉

thomasleese: you spotted...

spec/components/meta_tags_spec.rb
```diff
@@ -31,6 +31,30 @@ def example_document_for(schema_name, example_name)
...
+    )
+
+    assert_meta_tag('canonical', 'http://www.dev.gov.uk/test2')
+  end
+
+  it "doesn't renders the canonical URL in a meta tag if not requested" do
+    assert_empty render_component(content_item: {
+      base_path: "/test"
+    }).strip
+  end
```
kevindew: I'm sorry I was all set to give the ✅ but there's some quote inconsistencies

thomasleese: Yeah... I've made them all single quotes I think... I like to think they were consistent with the inconsistent style of the file...

tijmenb:
I'm away today but let's chat tomorrow how this would work. If it's an optional thing it might be better to implement per-app anyway, it's more explicit. I'm not sure apps would know whether or not to add the tag currently.

PR #305, opened 2018-05-11
Avoid usage of html_safe

tijmenb:
This is a follow up to an [incident last month][repo] where we found a XSS vulnerability caused by `html_safe` being called on untrusted user input (see https://github.com/alphagov/finder-frontend/pull/480 and follow ups). 

To avoid recurrence, this PR makes sure that 1) we're never calling `html_safe` so that everything is escaped by default, 2) we can't use it anymore. This makes escaping and calling `html_safe` the exclusive responsibility of the application. 

See the individual commits for details. 

[repo]: https://docs.google.com/document/d/1HDpDuEtmR0PVt2Te5OgMGb6MnLlPU1s6Rhlt9Qhva1Y/edit#heading=h.v4arcxojubbv

### DIFF
--- a/spec/components/all_components_spec.rb
+++ b/spec/components/all_components_spec.rb
@@ -1,0 +1,9 @@
+require 'rails_helper'
+
+describe "All components" do
+  it "doesn't use `html_safe`" do
+    files_with_html_safe = `grep -rni "html_safe" app/views/govuk_publishing_components/components`.lines
+
+    expect(files_with_html_safe).to be_empty
+  end
+end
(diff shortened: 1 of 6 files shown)


spec/components/all_components_spec.rb
```diff
@@ -0,0 +1,9 @@
+require 'rails_helper'
+
+describe "All components" do
+  it "doesn't use `html_safe`" do
+    files_with_html_safe = `grep -rni "html_safe" app/views/govuk_publishing_components/components`.lines
+
+    expect(files_with_html_safe).to be_empty
```
sihugh: 🙏

PR #307, opened 2018-05-11
Iterate heading component

andysellick:
Expand the heading component for general use across GOV.UK. Specifically:

- add option to control font size
- add option to give padding
- add option to control margin
- add options for border
- add branding

**Note this will break** headings in government-frontend as by default it removes margins and padding. Will need to update government-frontend with appropriate options to restore the original layout (there's a specific option included in the component now to do just that, which hopefully we'll eventually deprecate in favour of a more universal approach to margins).

Component guide link: https://govuk-publishing-compon-pr-307.herokuapp.com/component-guide/heading
Trello card: https://trello.com/c/iehFbdw0/54-modify-component-heading


docs/component_branding.md
```diff
@@ -29,9 +29,16 @@ To add colours to a component, modify the component to follow the example below.
 
 Note that the helper must be called for each element that needs a border or link colour applying. This allows for flexibility if one is required but not the other.
 
+The `brand_border_color` method can also be applied to the parent element of a component, in the situation where the component consists of only one element. This is currently not possible with `brand_color` as there is no requirement for this.
```
vanitabarrett: What do you mean here by "parent element"?

app/views/govuk_publishing_components/components/docs/heading.yml
```diff
@@ -21,6 +21,12 @@ examples:
     data:
       text: 'Original consultation'
       heading_level: 3
+  different_font_sizes:
+    description: Choose a font size appropriate to the chosen heading level. Valid options are 2 (24px) and 3 (19px), with the component defaulting to 27px, appropriate for a H1. This option is not tied to the heading_level option in order to give flexibility.
```
vanitabarrett: I'm not sure I understand the logic behind using the numbering system here. It seems to suggest that all h2s and h3s ideally should be 24px and 19px respectively (I wasn't aware that was the case)... but also suggests they're not enforced so we don't have to do that?

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -1,5 +1,10 @@
 .gem-c-heading {
   @include bold-27;
+}
+
+// special case for publications and consultations pages
+// separated to allow normalisation of the component for wider use
+.gem-c-heading--mobile-top-margin {
```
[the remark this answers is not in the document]
andysellick: Yeah, I really struggled to name this well. How about something more related to its use, like gem-c-heading--publications?

docs/component_branding.md
andysellick: The highest level element of the component - i.e. .gem-c-component-name. This was a bit tricky because I'd already come up with the branding model where I'd assumed parent element with children, but of course the heading component is only one element, so I had to redo the branding a little to be able to apply the branding classes to the parent element as well, not just the parent and child.

Does this need to be written more clearly? Was 'parent element' in this context confusing?

app/views/govuk_publishing_components/components/docs/heading.yml
andysellick: They're not, I just chose some semi-sensible numbers for the different heading sizes. We don't have a convention (our heading sizes are all over the place).

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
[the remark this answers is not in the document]
vanitabarrett: Just taking a look at examples of where this is used - do we know why publications and consultations are special cases? Personally I think it's clearer without the extra styling we give the heading, see below:

As-Is (special case styling being applied)
screen shot 2018-05-14 at 09 04 51

Removed special case styling
screen shot 2018-05-14 at 09 04 56

Maybe we should try and make these headings consistent? Either way, I'd be wary of naming it `--publications` as we may find another use case for a different document type, which would leave the name feeling misleading.

docs/component_branding.md
vanitabarrett: Ah ok, when I read "parent element" I think of an element immediately before another element. So in this case I thought you meant something that might wrap the `gem-c-` element..

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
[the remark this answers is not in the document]
andysellick: The reason for keeping this special case is that on mobile it adds a top margin which is removed on the element above it. I could start digging around and fixing the CSS on that non-component element above and not have this option, but this seemed like the quickest way forward.

Without it we'd get things like this...

screen shot 2018-05-14 at 09 46 44

Actually now I come to say this maybe fixing government-frontend is the better option. Okay, I'll look into that.

app/views/govuk_publishing_components/components/docs/heading.yml
andysellick: I like 24 and 19, much better, will change, thanks.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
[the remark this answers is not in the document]
andysellick: @vanitabarrett I've had a think, and I'm going to leave this option in here for now as I'm a bit pressed for time. I'm going to have to update government-frontend when the gem is bumped anyway, so I'll look into it then. Are you happy to approve on that basis?

docs/component_branding.md
andysellick: I've rewritten this a little to hopefully make it clearer - I can see why it was confusing.

PR #308, opened 2018-05-14
Add share links component

andysellick:
Adds the share links component from government-frontend. A separate PR will be raised to remove it from that app.

Only change in this component is the namespace, in CSS from `app-c` to `gem-c` and in JS to `GemTrackShareButtonClicks`.

Component guide link: https://govuk-publishing-compon-pr-308.herokuapp.com/component-guide/share_links
Trello card: https://trello.com/c/NZwDW9xV/89-modify-component-share-links

### DIFF
--- a/app/views/govuk_publishing_components/components/_share_links.html.erb
+++ b/app/views/govuk_publishing_components/components/_share_links.html.erb
@@ -1,0 +1,36 @@
+<%
+  title ||= 'Share this page'
+  facebook_href ||= false
+  twitter_href ||= false
+%>
+<% if facebook_href || twitter_href %>
+  <div class="gem-c-share-links dont-print" data-module="gem-track-share-button-clicks">
+    <h2 class="gem-c-share-links__title"><%= title %></h2>
+
+    <% if facebook_href %>
+      <%= link_to facebook_href,
+        target: "_blank",
+        rel: "noopener noreferrer",
+        class: "gem-c-share-links__link js-share-facebook" do %>
+        <span class="gem-c-share-links__link__icon">
+          <svg xmlns="http://www.w3.org/2000/svg" width="32" height="32">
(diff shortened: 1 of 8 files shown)


app/views/govuk_publishing_components/components/_share_links.html.erb
```diff
@@ -0,0 +1,36 @@
+<%
+  title ||= 'Share this page'
+  facebook_href ||= false
+  twitter_href ||= false
+%>
+<% if facebook_href || twitter_href %>
+  <div class="gem-c-share-links dont-print" data-module="track-share-button-clicks">
```
vanitabarrett: The [corresponding Javascript](https://github.com/alphagov/government-frontend/blob/591807f8b0f477fa9e852b69fdce6ab736f32c81/app/assets/javascripts/modules/track-share-button-clicks.js) needs to be copied over for this tracking to work

andysellick: Ah, thanks, good spot. Is there a reason this isn't just using the standard tracking?

vanitabarrett: I'm not sure, it looks like it was copied over when the links were moved from Whitehall, so it might just be it wasn't considered then? https://github.com/alphagov/government-frontend/pull/207

andysellick: Okay, I'll see if I can reproduce it using the standard module. Also if we pass the tracking, that makes it more flexible for the changes I'm planning.

andysellick: On second thought, I'll do that when I iterate the component, as it's a fairly major change and it'll block this PR. Have added the necessary JS - happy to approve?

PR #309, opened 2018-05-15
Add breadcrumbs component from static

thomasleese:
This moves over [the breadcrumbs component from static](https://govuk-static.herokuapp.com/component-guide/breadcrumbs).

[Trello Card](https://trello.com/c/t11eJLxC/138-implement-breadcrumb-schema-3)

https://trello.com/c/XLuzv0pB/25-move-the-breadcrumbs-component-to-the-gem

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
@@ -1,0 +1,43 @@
+@import "design-patterns/breadcrumbs";
+@import "mixins/back-arrow";
+@import "mixins/media-down";
+@import "mixins/touch-friendly-links";
+
+.govuk-breadcrumbs {
+  // reset the default browser styles
+  ol {
+    padding: 0;
+    margin: 0;
+  }
+
+  @include breadcrumbs;
+  @include touch-friendly-links();
+
+  .breadcrumb-for-current-page {
+    color: $secondary-text-colour;
+    text-decoration: none;
+  }
+
+  @include media-down(mobile) {
+    &.collapse-on-mobile li {
+      display: none;
+
+      &.parent-breadcrumb {
+        background-image: none;
+        display: block;
+        margin-left: 0;
+        padding-left: 14px;
+        position: relative;
+
+        &:before {
+          @include back-arrow;
+        }
+      }
+    }
+  }
+}
+
+.breadcrumb-for-current-page.govuk-breadcrumbs--inverse,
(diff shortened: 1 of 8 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -0,0 +1,43 @@
+@import "design-patterns/breadcrumbs";
+@import "mixins/back-arrow";
+@import "mixins/media-down";
+@import "mixins/touch-friendly-links";
+
+.govuk-breadcrumbs {
```
vanitabarrett: As we're moving the component into the gem,  we should change it to meet our component principles. One of these is we need to rename the namespaced CSS classes, e.g: this would be `gem-c-breadcrumbs`. More info here: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#namespacing

vanitabarrett: All the classes need to match the `gem-c-` pattern

thomasleese:
@vanitabarrett thanks, I've pushed up a commit to change the names and use BEM, is that better now?

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,16 +13,16 @@
...
 
-  .breadcrumb-for-current-page {
+  .gem-c-breadcrumbs--current {
     color: $secondary-text-colour;
     text-decoration: none;
   }
 
   @include media-down(mobile) {
-    &.collapse-on-mobile li {
+    &.gem-c-breadcrumbs--collapse-on-mobile li {
```
vanitabarrett: We tend to avoid selecting by element type - I'd suggest replacing `li` with a class, e.g: `gem-c-breadcrumbs__item` or `gem-c-breadcrumbs__list-item`

vanitabarrett: You'd then probably be able to make [some of the test selectors](https://github.com/alphagov/govuk_publishing_components/pull/309/commits/0082e890413d19d2077a58b9016715d07293f411#diff-79d1370df22d4cef0c4e9a10d443b147R32) more specific

thomasleese:
Thanks @vanitabarrett!

tijmenb:
Don't forget the changelog!

PR #310, opened 2018-05-15
Schema data breadcrumbs

thomasleese:
This implements schema data as JSON/LD.

[Trello Card](https://trello.com/c/t11eJLxC/138-implement-breadcrumb-schema-3)

PR #311, opened 2018-05-15
Version 8.0.0

andysellick:
Includes the following. Changes to the heading component are breaking (about to create a PR on government-frontend to account for that)

- Use new breadcrumbs component in contextual breadcrumbs (PR #313)
- Add share links component (PR #308)
- The Button component no longer accepts unescaped HTML in the info_text, you'll have to call html_safe on it yourself. Probably the only affected application is frontend (#305)
- Remove optional canonical meta tag (applications can add this tag explicitly if they need it)
- Translation nav add brand and tracking (PR #298)
- Subscription links add colour and tracking (PR #299)
- Iterate heading component (PR #307)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,6 @@
-# Unreleased
+# 8.0.0
 
+* Use new breadcrumbs component in contextual breadcrumbs (PR #313)
 * Add share links component (PR #308)
 * The Button component no longer accepts unescaped HTML in the `info_text`,
   you'll have to call `html_safe` on it yourself. Probably the only affected
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (7.3.0)
+    govuk_publishing_components (8.0.0)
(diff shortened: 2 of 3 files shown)


thomasleese:
Is there a chance we can wait until #310 is in?

andysellick:
@thomasleese probably, how long do you think it'll be before it's ready?

thomasleese:
@andysellick should be pretty soon, it's ready for review now.

andysellick:
@thomasleese @tijmenb has approved it, but apparently the contextual breadcrumb component is still using the breadcrumb component from static (or something like that) so we should look into that for this release as well.

andysellick:
D'oh! Also I think this is the thing @tijmenb is referring to... https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb

thomasleese:
Does https://github.com/alphagov/govuk_publishing_components/pull/313 cover that?

andysellick:
@vanitabarrett @thomasleese I'm fixing the stray `%>`

PR #320, opened 2018-05-16
Add option to display the generated HTML for components

tijmenb:
In some cases it's useful to be able to see the resulting HTML of the component. This adds an option to do this and enables it for the meta tags component.

## Preview

https://govuk-publishing-compon-pr-320.herokuapp.com/component-guide/machine_readable_metadata
https://govuk-publishing-compon-pr-320.herokuapp.com/component-guide/breadcrumbs
https://govuk-publishing-compon-pr-320.herokuapp.com/component-guide/meta_tags

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/meta_tags.yml
+++ b/app/views/govuk_publishing_components/components/docs/meta_tags.yml
@@ -7,6 +7,7 @@ body: |
   The code which reads the meta tags can be found <a href="https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/static-analytics.js#L76-L96">in static-analytics.js</a>.
 accessibility_criteria: |
   The analytics meta tags component should not be visible to any users.
+display_html: true
 examples:
   with_organisations:
     data:
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -106,6 +106,12 @@ $border-color: #ccc;
   }
 }
 
+.component-output {
+  > pre {
+    max-height: 200px;
+  }
+}
+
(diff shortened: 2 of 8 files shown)


app/views/govuk_publishing_components/components/docs/meta_tags.yml
```diff
@@ -7,6 +7,7 @@ body: |
   The code which reads the meta tags can be found <a href="https://github.com/alphagov/static/blob/master/app/assets/javascripts/analytics/static-analytics.js#L76-L96">in static-analytics.js</a>.
 accessibility_criteria: |
   The analytics meta tags component should not be visible to any users.
+display_html: true
```
nickcolley: 👍 for not having this on by default

nickcolley:
Non blocking: Is there any way to remove the indentation from the output?

screen shot 2018-05-16 at 13 59 36

Non blocking: Could this be done in a way where we say 'example_html' that overrides the example, since you can't see anything.

tijmenb:
I've rejiggled things a bit to get rid of the whitespace and reduce duplication. Also added the HTML output to breadcrumbs, because they output JSON-LD (schema.org data). I don't want to replace the actual output, because it's also important for people to see that a component outputs nothing.

nickcolley:
Non-blocking: Is it worth some kind of 'do not copy me' for the HTML? Or a readonly view.

Maybe I'm paranoid but one good thing about the current setup is that you have to go out your way to copy paste markup.

andysellick:
Is it enough that it's turned off by default? I don't imagine we'd enable this for most of the components.

tijmenb:
Let's not make this any more complicated than it needs to be, it's fine like this.

PR #322, opened 2018-05-17
Add image card component

andysellick:
Adds a new component, image card, intended for use for news articles and people entries on organisation pages. See the [component guide](https://govuk-publishing-compon-pr-322.herokuapp.com/component-guide/image_card) for a full explanation.

Examples in context:

screen shot 2018-05-29 at 10 01 46

Mobile:

screen shot 2018-05-18 at 15 29 05

Tested in IE8+, latest Chrome, Firefox, iphone 6, iphone 5, Google Nexus.

Component guide: https://govuk-publishing-compon-pr-322.herokuapp.com/component-guide/image_card
Trello card: https://trello.com/c/v8vHh54c/102-new-component-image-link

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -53,13 +53,13 @@
   }
 }
 
-.gem-c-image-card__metadata,
+.gem-c-image-card__context,
 .gem-c-image-card__description {
   position: relative;
   z-index: 2;
 }
 
-.gem-c-image-card__metadata {
+.gem-c-image-card__context {
   @include core-14;
   color: $grey-1;
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -38,10 +38,12 @@
   color: $grey-1;
 }
 
+.gem-c-image-card__description {
+  padding-top: $gutter-one-quarter;
+}
+
 .gem-c-image-card__title {
   @include bold-19;
-  margin-bottom: $gutter-one-quarter;
-  text-decoration: none;
 }
 
 .gem-c-image-card__title-link {
@@ -50,9 +52,9 @@
     position: absolute;
     z-index: 1;
     top: 0;
-    left: 0;
+    left: $gutter-half;
+    right: $gutter-half;
     height: 100%;
-    width: 100%;
   }
 }
 
@@ -81,10 +83,33 @@
 
   .gem-c-image-card__title {
     @include bold-24;
-    margin-bottom: $gutter-one-third;
+    padding-bottom: $gutter-one-third;
   }
 
   .gem-c-image-card__description {
     @include core-19;
+  }
+}
+
+.gem-c-image-card__list {
+  position: relative;
+  z-index: 2;
+  padding: $gutter-one-quarter 0 0 0;
+  margin: 0;
+  list-style: none;
+
+  &.gem-c-image-card__list--indented {
+    padding-left: $gutter-half;
+
+    .gem-c-image-card__list-item {
+      position: relative;
+
+      &:before {
+        content: "-";
+        position: absolute;
+        top: 0;
+        left: -$gutter-half;
+      }
+    }
   }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -1,0 +1,24 @@
+.gem-c-image-card {
+  position: relative;
+}
+
+.gem-c-image-card__image {
+  max-width: 100%;
+}
+
+.gem-c-image-card__title {
+  @include bold-19;
+  text-decoration: none;
+}
+
+.gem-c-image-card__title-link {
+  &:after {
+    content: "";
+    position: absolute;
+    z-index: 1;
+    top: 0;
+    left: 0;
+    height: 100%;
+    width: 100%;
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -55,3 +55,36 @@
     width: 100%;
   }
 }
+
+.gem-c-image-card--large {
+  .gem-c-image-card__image-wrapper {
+    @include grid-column( 2 / 3, tablet );
+  }
+
+  .gem-c-image-card__text-wrapper {
+    @include grid-column( 1 / 3, tablet );
+  }
+
+  .gem-c-image-card__image-wrapper {
+    @include media(mobile) {
+      margin-bottom: $gutter-one-third;
+    }
+  }
+
+  .gem-c-image-card__image-wrapper,
+  .gem-c-image-card__text-wrapper {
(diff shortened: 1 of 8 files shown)


app/views/govuk_publishing_components/components/docs/image_card.yml
```diff
@@ -0,0 +1,75 @@
...
+          href: "/3"
+        },
+      ]
+  extra_links_without_indent:
+    description: Don't indent the extra links. Used for links to people pages.
+    data:
+      href: "/government/people/jeremy-wright"
+      image_src: "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/person/image/354/s216_TME_3860TME_3860_960_Jeremy_Wright.JPG"
+      metadata: "The Rt Hon"
+      heading_text: "Jeremy Wright QC MP"
```
tijmenb: Perhaps use a fictitious person here, these pages might accidentally be indexed by Google / found by members of the public/parliament.

app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -0,0 +1,56 @@
+<%
+  href ||= false
+  image_src ||= false
```
tijmenb: Is `image_src` something we use elsewhere, or could this be `image_url`?

app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  image_alt ||= ""
+  heading_text ||= false
+  metadata ||= false
```
tijmenb: Is this the same concept as `context` in the [title component](https://github.com/alphagov/static/blob/master/app/views/govuk_component/title.raw.html.erb#L3)?

app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  description ||= false
+  large ||= false
+  heading_level ||= 2
+  heading_tag = "h#{heading_level}" if [1, 2, 3, 4, 5, 6].include? heading_level
+  extra_links ||= []
+  extra_links_no_indent ||= false
```
tijmenb: Out of interest, where is `extra_links_no_indent` used/not used?

app/views/govuk_publishing_components/components/docs/image_card.yml
andysellick: 👍

app/views/govuk_publishing_components/components/_image_card.html.erb
andysellick: I was following the convention of the attribute name, i.e. `src=`.

app/views/govuk_publishing_components/components/_image_card.html.erb
andysellick: I think so, will change.

app/views/govuk_publishing_components/components/_image_card.html.erb
andysellick: It's used [here](https://govuk-publishing-compon-pr-322.herokuapp.com/component-guide/image_card/with_extra_links/preview) and not used [here](https://govuk-publishing-compon-pr-322.herokuapp.com/component-guide/image_card/extra_links_without_indent/preview).

tijmenb: Sorry, I meant where on the site will we use https://govuk-publishing-compon-pr-322.herokuapp.com/component-guide/image_card/with_extra_links/preview? I don't see it on the current org page.

tijmenb: Okido. As an alternative, would it work to use a block to populate the description, and then use a list component inside it?

```erb
<%= render '.../image_card', href: "/foo" do %>
  <%= render '.../list', items: [{ text: "About the Number 10 guest historian series", href: "/bar"}] %>
<% end %>
```

andysellick: That could work, although I'm not sure we have list components that visually match what I need. Also do we have a mechanism for documenting that in the yml file?

maxgds:
Some at-a-glance critique: I'm not convinced by your use of "figure" here - the point of the figure tag is to present images with accompanying text, marked up as a figcaption. You're using accompanying text but not applying the figcaption to it - I can see why, given the markup you've ended up with, but it feels semantically broken to me.

I strongly approve of the link wrapper solution that means you're not duplicating links, which is a horrible experience for tabbing through a page.

It seems strange to call this an image card when images are optional - especially since the intro to the component states it's all about the image. Should we be trying to use the same card for both purposes? If so I'd recommend a rename (maybe link-card?) and rewording.

alextea:
The images in [this example](https://user-images.githubusercontent.com/861310/40240372-57c0a156-5ab0-11e8-8d92-e07d8f868c62.png) don't align to the grid. I don't know where the mark up for this is, so I can't see what's causing the issue.

andysellick:
@alextea my bad on the grid alignment, I chose an image that wasn't wide enough for the space (I've updated the screenshot in the description). I assume that authors should upload the right sized image (might need some instructions in the management interface?) - if not, I could put in a CSS rule to make the image always 100% wide, but I'd prefer not to do that as it can lead to poor image quality, better that the right size is uploaded.

andysellick:
@maxgds using a figure without a figcaption is valid, and as you noted, I need to wrap the image in an element anyway. As for the name of the component, I'll have a go at some rewording.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
```diff
@@ -0,0 +1,125 @@
...
+}
+
+.gem-c-image-card__image-wrapper {
+  @include grid-column( 1 / 3, mobile );
+
+  @include media(tablet) {
+    margin-bottom: $gutter-one-third;
+  }
+
+  + .gem-c-image-card__text-wrapper {
```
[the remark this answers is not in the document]
andysellick: It's not, no. It's to tighten up the spacing on mobile. Without it, you get this...

screen shot 2018-05-29 at 10 10 55

...and with this rule:

screen shot 2018-05-29 at 10 11 01

vanitabarrett:
Commit history changes look better, I have a few suggestions on how it could be cleaned up further:

- Pull the YML quote changes out of https://github.com/alphagov/govuk_publishing_components/pull/322/commits/6c9d69036164bcf9afd00af682b5f4dc637c652f and into each relevant commit above where that YML was introduced.
- Pull the quote changes in the spec file out of https://github.com/alphagov/govuk_publishing_components/pull/322/commits/6c9d69036164bcf9afd00af682b5f4dc637c652f and fixup to https://github.com/alphagov/govuk_publishing_components/pull/322/commits/b09a28aaf3358f7a5703c8360c09a4e7988d5260
- That will leave https://github.com/alphagov/govuk_publishing_components/pull/322/commits/6c9d69036164bcf9afd00af682b5f4dc637c652f as being mostly SASS changes, so the commit could be renamed to "Styling tweaks for browser compatibility" (or something)
- Could probably split and fixup commit https://github.com/alphagov/govuk_publishing_components/pull/322/commits/e66807439f81fb0e1aba409eb090c109c0fdec2b into the commit where the change was introduced.
- Could perhaps split this but that might be more complicated https://github.com/alphagov/govuk_publishing_components/pull/322/commits/a5c97ed75a73c978487c34052ac4012e54287b58

PR #324, opened 2018-05-18
Move title component from Static

tijmenb:
This moves the [title component](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e/app/views/govuk_component/title.raw.html.erb) from static into this gem.

The files were copied at [commit 600b93](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e).

## Preview

(...once review app is deployed)

https://trello.com/c/EruCrkqO


app/assets/stylesheets/govuk_publishing_components/_all_components.scss
```diff
@@ -14,20 +14,21 @@
 @import "components/button";
 @import "components/document-list";
 @import "components/error-summary";
+@import "components/title";
```
andysellick: +1 for sorting, but this is in the wrong place...

app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
```diff
@@ -10,3 +10,4 @@
 @import "components/print/step-by-step-nav";
 @import "components/print/step-by-step-nav-header";
 @import "components/print/feedback";
+@import "components/print/title";
```
andysellick: Maybe sort this one too?

app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -0,0 +1,63 @@
...
+    description: |
+      It’s unclear if links in the context of a title are useful and are being clicked by users. Data attributes are included to track this behaviour.
+
+      Context links are used on [topic pages](https://www.gov.uk/topic/business-tax/vat) where there is also a breadcrumb.
+    data:
+      context:
+        text: Publication
+        href: '/link'
+        data:
+          some_tracking_parameter: 'tracking-param'
```
andysellick: We've got a few components like this that include the option for tracking but don't call the tracking module on the parent element. I think this is something we should address long term.

spec/components/title_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+  it "title context link appears" do
+    render_component(title: "Hello World", context: { text: "Format", href: "/format", data: { tracking: true } })
+    assert_select ".gem-c-title__context-link[href='/format'][data-tracking]", text: "Format"
+  end
+
+  it "applies title length if supplied" do
+    render_component(title: "Hello World", context: "format", average_title_length: 'long')
+    assert_select ".gem-c-title .gem-c-title__text--long", text: "Hello World"
+  end
+end
```
andysellick: There doesn't seem to be a test for the inverse header option, there probably should be. I appreciate that this isn't your fault, you're just copying and pasting :)

tijmenb:
@andysellick updated!

PR #325, opened 2018-05-18
Move lead paragraph component from Static

tijmenb:
This moves the [lead paragraph component](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e/app/views/govuk_component/lead_paragraph.raw.html.erb) from static into this gem. The files were copied at [commit 600b93](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e).

Note that there's some clunkiness related to the use of `html_safe` and the [functionality to avoid "windows" in text](https://github.com/alphagov/government-frontend/pull/88). See the commits for more details. 

## Preview

https://govuk-publishing-compon-pr-325.herokuapp.com/component-guide/lead_paragraph

https://trello.com/c/7Fwv498s


spec/component_guide/component_example_spec.rb
```diff
@@ -26,7 +26,7 @@
 
   it 'passes params in examples to component example' do
     visit '/component-guide/test-component-with-params'
-    expect(body).to include('A test component that takes a required parameter')
+    expect(body).to include('A test component that takes a required')
```
andysellick: Not sure what this means? It made more sense before the removal of the word 'parameter'.

tijmenb: Fair enough, I went with a different solution.

tijmenb: And added some more context to the component in https://github.com/alphagov/govuk_publishing_components/pull/325/commits/d831255e5acb0e11932e02bee3e535389c1ed46a

PR #327, opened 2018-05-18
Move govspeak component from Static

tijmenb:
The govspeak component is the last static-based component that's used by the component guide, which means once this is merged we're really close to be able to render the component guide without static and slimmer. This would open up the possibility of using the components for admin applications as well.

## Preview

https://govuk-publishing-compon-pr-327.herokuapp.com/component-guide/govspeak

---

https://trello.com/c/ZMVwIyj5


nickcolley:
While out of scope for a lift and shift, wanted to share what I thought govspeak component could become.

The govspeak rendering for GOV.UK is done server side, not part of the typical frontend stack.

This has meant that there is duplication in the govspeak component, a good example is buttons.

http://govuk-static.herokuapp.com/component-guide/govspeak/button
http://govuk-static.herokuapp.com/component-guide/button

It would be unreasonable for the govspeak rendering to depend on static, since it's a service.

However with this gem, you could conceive of an approach where the govspeak gem uses this gem to call on the same button that is used in authored templates.

This would then make the govspeak component itself largely redundant, maybe only leaving the typography styles.

PR #332, opened 2018-05-21
Move the Previous/Next component from Static

tijmenb:
This moves the [Previous/Next component](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e/app/views/govuk_component/previous_and_next_navigation.raw.html.erb) from Static into this gem.

The files were copied at [commit 600b93](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e).

https://trello.com/c/NWSyhT8W


app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
```diff
@@ -0,0 +1,83 @@
+.gem-c-pagination {
+  display: block;
+  margin-top: $gutter;
+  margin-bottom: $gutter;
+  margin-left: -$gutter-half;
+  margin-right: -$gutter-half;
+}
+
+// scss-lint:disable SelectorFormat
```
andysellick: Do we still need to disable SelectorFormat here?

app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
```diff
@@ -0,0 +1,83 @@
...
+  }
+}
+
+.gem-c-pagination__link-title {
+  @include core-27($line-height: (33.75 / 27));
+  display: block;
+}
+
+.gem-c-pagination__item--previous {
+  @include media-down(mobile) {
```
andysellick: I've not seen this media query on GOV.UK before. What's it doing?

andysellick: Thanks Tom, I wondered if it was something like that. I'll have a crack at rewriting these if @tijmenb is happy.

tijmenb: Done in https://github.com/alphagov/govuk_publishing_components/pull/332/commits/7149545984b16e8b505ac77729c9e7911bf15f50.

## Before

before-mob
before-wide

## After

after-mob
after-wide

PR #334, opened 2018-05-21
Move the Government navigation component from Static

tijmenb:
This moves the [Government navigation](https://github.com/alphagov/static/blob/600b93807bd169b6a607f77ddc98878769c08c2e/app/views/govuk_component/government_navigation.raw.html.erb) component from Static into this gem.

The files were copied at [commit 600b93](https://github.com/alphagov/static/commit/600b93807bd169b6a607f77ddc98878769c08c2e).

https://trello.com/c/uwcWXXNp


sihugh:
These haven't changed in static since [the referenced commit](https://github.com/alphagov/static/commit/600b93807bd169b6a607f77ddc98878769c08c2e). 👍 

(I updated the link to the commit in the PR as it didn't work when I clicked it)

PR #338, opened 2018-05-29
Add start button icons

sihugh:
We should include these with the component that uses them.  I copied these from [static](https://github.com/alphagov/static/tree/bba61d34edd0a1f05d93c35b302ebec3d6571116/app/assets/images).  (I think this was an oversight when the component was originally migrated)

I've checked that the other components here are using images included with this
gem.


tijmenb:
This seems OK, but I'm completely confused why it works on master:

https://govuk-publishing-components.herokuapp.com/component-guide/button/start_now_button

Which actually references this one:

https://govuk-publishing-components.herokuapp.com/assets/icon-pointer-2x-3994c32525c697dfd3e09a2d53247a72d6e6f7810f5694993968e8144cb971f8.png

tijmenb:
Mystery solved, the image currently comes from https://github.com/alphagov/govuk_frontend_toolkit/blob/master/images/icon-pointer-2x.png

tijmenb:
I guess that the image is made available by https://github.com/alphagov/govuk_publishing_components/blob/1fcf6548584b053008426d6ceb1dd7376798c7cf/spec/dummy/app/assets/config/manifest.js#L2 or https://github.com/alphagov/govuk_publishing_components/blob/0ffeab9ebe1fd53300c80bea63572ed39b6bf18a/lib/govuk_publishing_components/engine.rb#L4

sihugh:
Rebased changelog

PR #341, opened 2018-05-30
Add Publisher logo for GOV.UK

tijmenb:
This is required by Google, and might be displayed on Google News.

https://trello.com/c/gKHoUGuA


lib/govuk_publishing_components/presenters/machine_readable/article_schema.rb
```diff
@@ -24,6 +24,10 @@ def structured_data
             "@type" => "Organization",
             "name" => "GOV.UK",
             "url" => "https://www.gov.uk",
+            "logo" => {
+              "@type" => "ImageObject",
+              "url" => page.logo_url,
```
sihugh: The example in http://schema.org/ImageObject implies that we should use `contentUrl` for this, though it's not very clear, as `url` and `contentUrl` are both valid properties.

`url` is listed as  _URL of the item_ and `contentUrl` as _Actual bytes of the media object, for example the image file or video file_.  

Having checked it in https://search.google.com/structured-data/testing-tool/u/0/ it complains if we don't have a `url`, but not if we are missing `contentUrl`.  So I guess I'm happy!

tijmenb: 👍 thanks for the digging!

PR #343, opened 2018-05-30
Add Person schema for use on people pages

tijmenb:
This implements http://schema.org/Person. It's to be used on pages like https://www.gov.uk/government/people/theresa-may.

https://trello.com/c/gKHoUGuA

### DIFF
--- a/lib/govuk_publishing_components/presenters/machine_readable/person_schema.rb
+++ b/lib/govuk_publishing_components/presenters/machine_readable/person_schema.rb
@@ -1,0 +1,39 @@
+module GovukPublishingComponents
+  module Presenters
+    class PersonSchema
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/machine_readable/person_schema.rb
```diff
@@ -0,0 +1,39 @@
...
+      def initialize(page)
+        @page = page
+      end
+
+      def structured_data
+        # http://schema.org/Person
+        {
+          "@context" => "http://schema.org",
+          "@type" => "Person",
+          "mainEntityOfPage" => {
```
sihugh: 👌 I hadn't spotted this before

PR #346, opened 2018-05-31
Add notice component

andysellick:
- from government-frontend
- changed CSS namespace from app-c to gem-c
- test file fairly heavily rewritten
- govspeak template included in dummy app to allow tests to work
- otherwise should be the same as the original component

PR to remove this component from government-frontend is here: https://github.com/alphagov/government-frontend/pull/916

Trello card: https://trello.com/c/vdIdfjHn/121-modify-component-notice

### DIFF
--- a/spec/components/notice_spec.rb
+++ b/spec/components/notice_spec.rb
@@ -1,0 +1,55 @@
+require 'rails_helper'
+
+describe "Notice", type: :view do
+  def component_name
+    "notice"
+  end
+
+  it "renders nothing when no params provided" do
+    assert_empty render_component({})
+  end
+
+  it "renders a notice with only a title and no description" do
+    render_component(title: "Statistics release cancelled")
+    assert_select ".gem-c-notice__title", text: "Statistics release cancelled"
+    assert_select ".gem-c-notice__description", false
+  end
+
+  it "renders a notice with an aria label and region" do
+    render_component(title: "Statistics release cancelled")
+    assert_select ".gem-c-notice[aria-label=Notice][role=region]"
+  end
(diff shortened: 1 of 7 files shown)


spec/components/notice_spec.rb
```diff
@@ -0,0 +1,56 @@
+require 'rails_helper'
+
+describe "Notice", type: :view do
+  def component_name
+    "notice"
+  end
+
+  it "fails to render when no parameters are given" do
```
vanitabarrett: I'd change the name of this test to something like "renders nothing when no params provided". The current test name suggests we should get an error or failure.

spec/components/notice_spec.rb
```diff
@@ -0,0 +1,56 @@
...
+  # rendering the govspeak component doesn't work here, so we visit a component guide page where it's already rendered and use that instead
+  # this is only a workaround until the govspeak component is also part of the gem
+  # note also that this test will fail if run individually because the govspeak component is rendered differently if run individually
+
+  it "renders a notice with a title and description govspeak" do
+    visit '/component-guide/notice/with_description_govspeak'
+
+    within '.component-guide-preview', match: :first do
+      assert page.has_selector?(".gem-c-notice__title", text: "Statistics release update")
+      assert page.has_selector?(".govuk-govspeak")
```
vanitabarrett: Not sure this assert is needed and the below ones will fail if `.govuk-govspeak` aren't present anyway?

andysellick: Ah yes, that was debug, thanks well spotted.

PR #347, opened 2018-05-31
Stop inserting &nbsp; between last 2 words

tijmenb:
This feature was added when the component lived in Static. It replaces the space between the last 2 words with a non-breaking space, so we'll never have a dangling word in a paragraph.

This PR removes the feature because:

1. This component was recently moved into the gem. While converting applications to use it, I find that it turns out to be very hard to maintain this feature without jumping through a lot of hoops. For example, a [simple test like this][1] doesn't work, because `@content_item["description"] ` doesn't have the `nbsp`, but the actual page does. This is *very* surprising, even to me -- and I ported this feature so I should've known.

2. The Sanitisation code is pretty inscrutable, and introduces potential security vulnerabilities.

[1]: https://github.com/alphagov/government-frontend/blob/23ab4380b09e1e76bfbab70079530900a1f3e7dc/test/integration/specialist_document_test.rb#L34


andysellick:
@tijmenb have you got some examples of pages where the nbsp is relevant so we can see before/after?

andysellick:
I've found one. Before:

screen shot 2018-05-31 at 12 42 05

After:

screen shot 2018-05-31 at 12 41 58

Do we have a similar feature for headings somewhere? I thought that was what this PR was about when I first glanced at it.

tijmenb:
@andysellick the same thing doesn't exist for heading.

PR #349, opened 2018-05-31
Version 9.0.0

andysellick:
Contains breaking changes.

* Import images for the button component from static (PR #338)
* Add contents list component (PR #342)
* BREAKING: Iterate share links component (PR #316)
* Add image card component (PR #322)
* Add notice component (PR #346)
* Add "person" schema to the Machine readable metadata component (PR #343)


andysellick:
Holding off on this for the moment as I have some other changes in the works that might as well go in. But if there's an urgency for a new version of the gem let me know and we can go ahead.

PR #353, opened 2018-06-04
Remove Slimmer::GovukComponents

tijmenb:
We no longer need this because we're just using gem-based components here now.

https://trello.com/c/pU7YINt1


nickcolley:
This is huge! 👍 💯

PR #356, opened 2018-06-05
Deprecate passing in unsafe HTML into Govspeak

tijmenb:
Currently, when passing in HTML into the Govspeak component the component will mark the input as "HTML safe", meaning it will be rendered on the page without sanitisation.

A simple situation like this, where `@bar` is untrusted user input, will result in an XSS vulnerability.

```
<%= render 'govuk_publishing_components/components/govspeak', content:
"<p>Foo #{@bar}</p>" %>
```

We currently have 1 such [vulnerability in email-alert-frontend][xss], but it's not exploitable.

[xss]: https://github.com/alphagov/email-alert-frontend/blob/73dc8d4db0174b215fc61c42fa8ad992f2d8f6d2/app/views/authentication/request_sign_in_token.html.erb#L17

A better way to pass in input is to use a block, which will use sanitisation by default:

```erb
<%= render 'govuk_publishing_components/components/govspeak' do %>
  <p>Foo <%= @bar %></p>
<% end %>
```

This PR deprecates (but doesn't remove) the functionality to pass in untrusted input. See commits for details.


spec/components/all_components_spec.rb
```diff
@@ -2,7 +2,7 @@
 
 describe "All components" do
   it "doesn't use `html_safe`" do
-    files_with_html_safe = `grep -rni "html_safe" app/views/govuk_publishing_components/components`.lines
+    files_with_html_safe = `grep -rni "html_safe " app/views/govuk_publishing_components/components`.lines
```
andysellick: Why this change?

app/views/govuk_publishing_components/components/_govspeak.html.erb
```diff
@@ -14,6 +14,21 @@
     <% if content.html_safe? %>
       <%= content %>
     <% else %>
+      <% puts "
```
andysellick: I've not seen puts used like this in a template before. Will this render the text inside it as written, or will it all collapse into a single chunk?

tijmenb: It looks like this in the output:

```
        You've passed in unsanitised HTML into the govspeak component as the
        `content` param.

        Passing in unsafe HTML is deprecated and will be removed in a future
        version. You need to pass in a block instead or use the `capture` helper.

        See the component guide for examples.

        If you're 100% sure there's no unsanitised user input in the string you
        could also call `.html_safe` on the string or use the `raw` helper before
        passing it in.

        Called from /var/lib/jenkins/workspace/components_govspeak-content-LXCJ5KGV3W4GAGTNRFJXBGJGVEBXPWFEIRVZXHMIVRN7HGSF5OKA/app/views/govuk_publishing_components/component_guide/show.html.erb:16:in `__var_lib_jenkins_workspace_components_govspeak_content______________________________________________________app_views_govuk_publishing_components_component_guide_show_html_erb___1824013196706400841_47344039424500'
```

https://ci.integration.publishing.service.gov.uk/job/govuk_publishing_components/job/govspeak-content/2/console

spec/components/all_components_spec.rb
tijmenb: Lol I hoped you wouldn't notice. It's to avoid this test triggering on the deprecation message in the component.

tijmenb:
Updated this PR to add tests, fix a problem with govspeak-htmlpub and fix the deprecated usage in the component guide.

PR #359, opened 2018-06-06
Add placeholders for pages that don't have an image

tijmenb:
Most pages on GOV.UK do not have an image, only things like news articles and case studies have. But Google requires an image for its news and article schema. This PR adds placeholders (in 1x1, 4x3 and 16x9 as [recommended by the Google guidelines][1]).

Caveat: I don't agree with this change. The purpose of the `image` field is to supply machines with a relevant image - which this is not. By having a placeholder we subvert the meaning of the schema and make it harder for machines to interpret our content. The image seems to be needed only by Google, so we're making the machine readability worse to appease a stricter interpretation of the schema. See [this StackOverflow answer](https://stackoverflow.com/a/33070334/421664) for more context. The counterargument to my view is that we should make sure Google accepts our stuff, which will make search results better and be better for users.

## Before

screen shot 2018-06-06 at 12 03 14

https://search.google.com/structured-data/testing-tool#url=https%3A%2F%2Fwww.gov.uk%2Fstudent-finance-register-login

## After

screen shot 2018-06-06 at 12 26 59

https://search.google.com/structured-data/testing-tool#url=https%3A%2F%2Fgovuk-frontend-app-pr-1526.herokuapp.com%2Fstudent-finance-register-login

Trello: https://trello.com/c/i15awEUU

[1]: https://developers.google.com/search/docs/data-types/article


tijmenb:
@sihugh good points, I'll ask Sam.

MatMoore:
@joelanman was doing some work over the firebreak to auto generate images from page titles, so maybe we could swap out the images at some point. I think something like this is more appropriate than the generic gov.uk logo for anything that does link unfurling.

PR #370, opened 2018-06-08
Update document list documentation

vanitabarrett:
The document list component was going to be used on the new /government/organisations pages in Collections. However, we've decided that using this component will introduce more limitations.
We therefore need to update this documentation in case it misleads people in the future.

Component guide for this PR:
https://govuk-publishing-compon-pr-370.herokuapp.com/component-guide/document_list

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/document_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/document_list.yml
@@ -104,7 +104,7 @@ examples:
           text: 'School exclusion'
           path: '/government/publications/school-exclusion'
   with_context:
-    description: Context can be provided to render next to the item title. This will be used on /government/organisations to indicate if an organisation is hosted on a separate website or being moved to GOV.UK
+    description: Context can be provided to render next to the item title.
     data:
       items:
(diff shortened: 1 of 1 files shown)


app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -104,7 +104,7 @@ examples:
           text: 'School exclusion'
           path: '/government/publications/school-exclusion'
   with_context:
-    description: Context can be provided to render next to the item title. This will be used on /government/organisations to indicate if an organisation is hosted on a separate website or being moved to GOV.UK
+    description: Context can be provided to render next to the item title.
```
andysellick: Probably not for this PR, but was this added just for the org pages index? In which case we could probably remove the whole feature.

vanitabarrett: We could but I'm not sure if it'll come in useful for any of the other org pages work we have left to do so I'm going to leave it for now. We can revisit it later if we think it will never be used

PR #372, opened 2018-06-12
Add no10 to the branding model

andysellick:
- prime ministers office brand colours are not included in the list of organisations pulled from frontend toolkit, added manually
- also add hover effect

---

Component guide for this PR:
https://govuk-publishing-compon-pr-372.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
@@ -8,6 +8,10 @@
     .brand--#{nth($organisation, 1)} {
       .brand__color {
         color: nth($organisation, 3);
+
+        &:hover {
+          color: darken( nth($organisation, 3), 10% );
+        }
       }
 
       // the & declaration allows border-color to also be applied to the parent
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
```diff
@@ -8,6 +8,10 @@
     .brand--#{nth($organisation, 1)} {
       .brand__color {
         color: nth($organisation, 3);
+
+        &:hover {
+          color: lighten( nth($organisation, 3), 10% );
```
vanitabarrett: For some colours, this will mean they no longer pass colour contrast checks.

**Top link hovered; bottom link normal:**
screen shot 2018-06-12 at 13 16 43

screen shot 2018-06-12 at 13 17 37

andysellick: The default link hover colour on GOV.UK fails that test too.

andysellick: Actually we can fix this easily if I go with 'darken' instead of 'lighten'. I'll do that.

PR #374, opened 2018-06-12
Move Metadata component from Static

tijmenb:
https://trello.com/c/HPvqjaYP


app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -0,0 +1,70 @@
...
+  @include core-16;
+  @include responsive-bottom-margin;
+  @extend %contain-floats;
+
+  &.direction-rtl {
+    direction: rtl;
+    text-align: start;
+  }
+
+  dt {
```
andysellick: We should be using a class name here, not a tag. Plenty of other examples of this throughout this file as well.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -0,0 +1,70 @@
...
+    margin-top: 0.5em;
+    line-height: normal;
+
+    @include media(tablet) {
+      float: left;
+      clear: left;
+      width: 30%;
+      box-sizing: border-box;
```
andysellick: This should use the box sizing mixin - https://github.com/alphagov/govuk_publishing_components/blob/9a338ca0bade6af4271e9e5d8d0b441225f9f900/app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss#L9

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -0,0 +1,70 @@
...
+    a:hover {
+      text-decoration: underline;
+    }
+  }
+
+  &.direction-rtl dd {
+    float: right;
+  }
+
+  .toggle-wrap {
```
andysellick: gem-c namespace?

app/views/govuk_publishing_components/components/_metadata.html.erb
```diff
@@ -0,0 +1,62 @@
...
+
+  part_of ||= []
+  part_of = Array(part_of)
+
+  other ||= nil
+
+  direction_class = ""
+  direction_class = " direction-#{direction}" if local_assigns.include?(:direction)
+%>
+<div class="gem-c-metadata<%= direction_class %>" data-module="toggle">
```
andysellick: We've got a copy of the toggle module in the components gem now, just replace `toggle` here with `gem-toggle`.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: Where does this mixin come from?

andysellick: https://github.com/alphagov/govuk_frontend_toolkit/blob/452f82e7538d7f5b8f73ee7d81abea9bc857af13/stylesheets/_css3.scss#L68

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -0,0 +1,70 @@
...
+  @include core-16;
+  @include responsive-bottom-margin;
+  @extend %contain-floats;
+
+  &.direction-rtl {
+    direction: rtl;
+    text-align: start;
+  }
+
+  .gem-c-metadata__term {
```
andysellick: This doesn't need to be nested since we're using BEM - will make it more readable.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -0,0 +1,70 @@
...
+  }
+
+  &.direction-rtl .gem-c-metadata__definition {
+    float: right;
+  }
+
+  .gem-c-metadata__toggle-wrap {
+    display: none;
+
+    a {
```
andysellick: Is it possible to replace this with an explicit class?

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: Ok. I get an error for `.js-enabled & .gem-c-metadata__toggle-wrap {` though:

```
Sass::SyntaxError: Base-level rules cannot contain the parent-selector-referencing character '&'.
```

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: Is it necessary? I've found 8 instances of `box-sizing: border-box` in other CSS files.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: Removed the class.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
andysellick: I think we should be using the mixin if only to ensure consistency. We should probably get around to updating those other 8 instances as well at some point.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
andysellick: I don't think you need the `&` in there anymore? It should just be `js-enabled .gem-c-metadata__toggle-wrap {` I think?

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: There are more instances of `box-sizing: border-box` than `@include box-sizing`, so for more consistency we would keep `box-sizing: border-box`. I guess I'm not sure why we need the browser prefixes if we've been doing fine without? (I've done it anyway to move it along).

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
tijmenb: 👍 updated

PR #376, opened 2018-06-14
Modify share links component

andysellick:
Fix some small issues with the component ahead of using it in the new organisation pages. Specifically:

- added rel="external" to the links for tracking purposes
- removed the `dont-print` style and added specific print styles for this component
- removed the visually hidden 'share on' text as it no longer works in the context of providing custom text to the component, and where it's currently in use the links have a heading above them saying 'share on'
- stop long link names from wrapping beneath the icons

Trello card: https://trello.com/c/wPvE62vw/185-modify-share-links-component

---

Component guide for this PR:
https://govuk-publishing-compon-pr-276.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
@@ -11,6 +11,7 @@
 @import "components/print/feedback";
 @import "components/print/govspeak-html-publication";
 @import "components/print/govspeak";
+@import "components/print/share-links";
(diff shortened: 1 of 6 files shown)


vanitabarrett:
The line height is quite large when it wraps:

screen shot 2018-06-14 at 13 31 51

Looks like we're setting the line-height to be the height of the image. I wonder if we can replace this and introduce the spacing another way so we don't affect the line-height of the text

andysellick:
@vanitabarrett thanks, have addressed.

PR #379, opened 2018-06-15
Remove special casing for Static

tijmenb:
We've moved all the components to the gem, so we can start removing some of the special cases that were needed to render static's components in the component guide in static. See commits for the details.

https://trello.com/c/pU7YINt1/45-move-all-components-from-static-into-the-gem

### DIFF
--- a/app/views/layouts/govuk_publishing_components/application.html.erb
+++ b/app/views/layouts/govuk_publishing_components/application.html.erb
@@ -13,11 +13,6 @@
   <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
   <%= stylesheet_link_tag "#{GovukPublishingComponents::Config.application_stylesheet}" %>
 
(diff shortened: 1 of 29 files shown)


app/views/layouts/govuk_publishing_components/application.html.erb
```diff
@@ -13,11 +13,6 @@
   <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
   <%= stylesheet_link_tag "#{GovukPublishingComponents::Config.application_stylesheet}" %>
 
-  <% if GovukPublishingComponents::Config.static %>
-    <%= stylesheet_link_tag "component_guide/all_components", media: "screen" %>
-    <%= stylesheet_link_tag "component_guide/all_components_print", media: "print" %>
```
andysellick: If we're removing these references we should remove them from the component gem as well.

PR #384, opened 2018-06-22
Add isPartOf to articles that are in a step by step

sihugh:
- Adds a `HowTo` schema to more accurately (than using `Article`) represent the semantics of this relationship.
- Links content that should show things in a `step-by-step-related` component to those pages by using the [`isPartOf` property](https://schema.org/isPartOf).

The `HowTo` schema implemented here is minimal because we link using [`sameAs`](https://schema.org/sameAs) to the step by step page which has a [fuller schema implemented in collections](https://github.com/alphagov/collections/blob/7e0134d150916d37cef679980a7d13400c583b71/app/models/step_nav.rb#L31-L65).

It seems valid (at least the Google structured data testing tool doesn't object) to be "isPartOf" many things. This fits quite well with what we want to do with step by steps.

We could easily add the step by steps that are in the `related_to_step_nav` links array, but this seems a step too far at this point as it's not exactly certain how this will affect search rankings, snippets, etc.

## Structured data test

This was tested by pasting the json generated for https://www.gov.uk/employer-preventing-discrimination (which is part of two step by steps) into https://search.google.com/structured-data/testing-tool/u/0/

screen shot 2018-08-02 at 15 58 24

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Step by step component, Google snippet improvement (PR #461)
 * Add tabs component (PR #455)
 * Adds styling for singular related step by step navs (PR #458)
+* Add schema.org `isPartOf` links from articles to step by steps (PR #384)
 
 ## 9.8.0
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (9.7.0)
+    govuk_publishing_components (9.8.0)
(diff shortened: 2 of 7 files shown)


sihugh:
I've added a do not merge to this as I'd like to chat with people about this approach before doing anything futher

tijmenb:
Did you look into linking to the step by step instead of including the schema in this one? Perhaps something with `sameAs` would work?

sihugh:
I've updated this to use `sameAs` and link to the step by step, rather than implementing a partial HowTo schema.

PR #397, opened 2018-07-09
Add organisation meta tag if current page is org

vanitabarrett:
Trello: https://trello.com/c/cbRSHmA9/30-add-organisation-id-to-custom-dimensions-in-google-analytics

Adds the govuk:analytics:organisations meta tag if the current page is an org (has an analytics_identifier value in the content item)

Component guide for this PR:
https://govuk-publishing-compon-pr-397.herokuapp.com/component-guide

### DIFF
--- a/lib/govuk_publishing_components/presenters/meta_tags.rb
+++ b/lib/govuk_publishing_components/presenters/meta_tags.rb
@@ -46,6 +46,7 @@ module GovukPublishingComponents
 
       def add_organisation_tags(meta_tags)
         organisations = []
+        organisations += [content_item] if content_item[:document_type] == "organisation"
         organisations += links[:organisations] || []
         organisations += links[:worldwide_organisations] || []
         organisations_content = organisations.map { |link| "<#{link[:analytics_identifier]}>" }.uniq.join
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@sihugh I think this will fix https://trello.com/c/cbRSHmA9/30-add-organisation-id-to-custom-dimensions-in-google-analytics

vanitabarrett:
@tijmenb Potentially, although from what I could see only organisations actually have a value for `analytics_identifier` - the rest seem to be `null`. I can add a check against the `document_type` though to make sure we only do this for organisations?

vanitabarrett:
@tijmenb I've adjusted the logic slightly so we only set `govuk:analytics:organisations` to be equal to the high level content_item[:analytics_identifier] if the document is an organisation.

lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -45,15 +45,19 @@ def add_core_tags(meta_tags)
...
-
-        world_locations = links[:world_locations] || []
-        world_locations_content = world_locations.map { |link| "<#{link[:analytics_identifier]}>" }.join
-        meta_tags["govuk:analytics:world-locations"] = world_locations_content if world_locations.any?
+        if content_item[:document_type].eql?("organisation")
+          meta_tags["govuk:analytics:organisations"] = "<#{content_item[:analytics_identifier]}>" if content_item[:analytics_identifier]
+        else
+          organisations = []
+          organisations += links[:organisations] || []
+          organisations += links[:worldwide_organisations] || []
```
tijmenb: I wonder if you could replace the conditional by adding this line?

```rb
organisations += [content_item] if content_item[:document_type] == "organisation"
```

PR #371, opened 2018-06-11
[WIP] Create layout component for admin apps

tijmenb:
This PR creates a template for admin apps using GOV.UK Frontend, split up in 4 components.

This layout will be used by the Release app in https://github.com/alphagov/release/pull/228, and is intended to be used after that by the new publisher workflow app, organisations publisher and an app to manage content.

This follows on earlier spikes https://github.com/alphagov/govuk_publishing_components/pull/328 and https://github.com/alphagov/govuk_publishing_components/pull/331, which were used to identify how we might consume GOV.UK Frontend.

## Preview

- [Component guide](https://govuk-publishing-compon-pr-371.herokuapp.com/component-guide/layout_for_admin)
- [Special preview page](https://govuk-publishing-compon-pr-371.herokuapp.com/admin)

screen shot 2018-06-11 at 16 38 30

https://trello.com/c/09fVPrRN

PR #398, opened 2018-07-10
Add GOV.UK Frontend

tijmenb:
This commit adds GOV.UK Frontend (https://design-system.service.gov.uk/) to the gem.

When we run `npm install` npm will install the library into `node_modules`. This is ignored in git to avoid us having to commit the entire 15.000 lines of code in govuk_frontend. To include it in the distributed version of the gem, the node_modules directory is added to `s.files` in the gemspec. To make the files available during build, we run `npm install` before the tests (see Jenkinsfile).

PRs that actually make use of GOV.UK Frontend will come separately (see https://github.com/alphagov/govuk_publishing_components/pull/371 for how we intend to use it).

https://trello.com/c/R7Ija8dR/16-create-layout-system-for-admin-applications

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,19 @@ There are 2 types of helper classes in this app:
 - [AppHelpers](lib/govuk_publishing_components/app_helpers). Are exposed to the applications using this gem. They should be documented using RDoc.
 - [Component Presenters](lib/govuk_publishing_components/presenters). Anything in these classes is only for use within the components. They should be marked `@private`.
 
+## GOV.UK Frontend
+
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -30,6 +30,17 @@ There are 2 types of helper classes in this app:
...
 
+## GOV.UK Frontend
+
+This gem is the way GOV.UK consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), part of the GOV.UK Design System.
+
+We use NPM to pull in the GOV.UK Frontend styles and Javascript, but we don't commit
+the `node_modules` directory. To install GOV.UK Frontend on your machine, run:
+
+```
+npm install
```
nickcolley: Ruby users may not have used npm before and need to install Node.js, should we include a step for that?

PR #404, opened 2018-07-10
Add footer component

tijmenb:
This implements the footer component from GOV.UK Frontend.

Things of note:

- Even though we don't have custom CSS, we still create a CSS file for this component. This is mainly so it's easy to discover for a developer where the styles come from.
- When the component is `part_of_admin_layout`, we'll show a warning banner. This hopefully makes it clear that the component can't be used on the public frontend (because GOV.UK Frontend isn't available).

[View in component guide](https://govuk-publishing-compon-pr-404.herokuapp.com/component-guide/layout_footer)

https://trello.com/c/k5agy6jw

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
@@ -1,0 +1,1 @@
+// This component relies on styles from GOV.UK Frontend
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -50,6 +50,16 @@ $border-color: #ccc;
   }
 }
 
+.component-warning {
+  border: 3px solid $orange;
+  margin: 0 0 $gutter;
+  padding: $gutter $gutter 0 $gutter;
+
+  .component-warning__title {
+    @include bold-24;
+  }
+}
+
 .component-doc-h2 {
   @include bold-27;
   margin: ($gutter * 1.5) 0 $gutter;
--- a/app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
@@ -9,3 +9,5 @@ $govuk-font-url-function: 'font-url';
 // Include our GOV.UK components
 @import "govuk_publishing_components/all_components";
 
+// Components that are only available inside the admin layout
+@import "govuk_publishing_components/components/layout-footer";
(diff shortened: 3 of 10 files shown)


alex-ju:
I have a few suggestions for this PR, but since it's already merged I guess I can raise a separate PR to bring them in:
- we can use the notice component instead of adding css and markup to the review app
```
<%= render "govuk_publishing_components/components/notice", {
  title: "Admin layout only",
  description_govspeak: "<p>This component uses the Design System, so only works inside the <a href=\"/component-guide/layout_for_admin\">admin layout</a></p>"
} %>
```
- we could name the component simply "Footer" instead of "Layout footer" as it's not tightly coupled with the layout
- I would try to avoid tying this component with the `admin_layout` even if the immediate purpose of it is to serve it. we instead mention that it needs/relies on govuk-frontend - this will help long-term, if we'll create a new layout
- It is a minimal implementation of the footer component - no logic was added to the component - which means consumers can not add links in the footer for example; again, this may not be an immediate need, but is limiting the reuse of the component.

@tijmenb I'm happy to have a chat about these when you have time.

tijmenb:
> we can use the notice component instead of adding css and markup to the review app

Good idea! I'll do that.

> we could name the component simply "Footer" instead of "Layout footer" as it's not tightly coupled with the layout

I figure "Header" and "Footer" are relative terms - if we we're just to use "Header" this will confuse with the ["Inverse header" component](https://govuk-publishing-components.herokuapp.com/component-guide/inverse_header), which is a header to the main content.  

Also, this is a bit of a hack to make sure the layout, footer, header and main components appear next to each other in the component guide, since we don't have a grouping system.

> I would try to avoid tying this component with the admin_layout even if the immediate purpose of it is to serve it. we instead mention that it needs/relies on govuk-frontend - this will help long-term, if we'll create a new layout

That's definitely the long term plan. The reason for not including the header component CSS in the main component stylesheet at this time is because it will be bundled into the CSS by all of the applications. It seems better if we can avoid that now. 

> It is a minimal implementation of the footer component - no logic was added to the component - which means consumers can not add links in the footer for example; again, this may not be an immediate need, but is limiting the reuse of the component.

Yeah, the intention is to extend it as needed.

app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
```diff
@@ -0,0 +1 @@
+// This component relies on styles from GOV.UK Frontend
```
nickcolley: This is a great idea 👍

PR #408, opened 2018-07-11
Add header component

alex-ju:
This implements the header component from GOV.UK Frontend exemplifying with environment, product name and navigation items.

## Notes
- Contains [a CSS fix](https://github.com/alphagov/govuk_publishing_components/pull/408/files#diff-e9749171eb74b1f0e663c2b759893bf5R16) that should be removed after [updating to the next govuk-frontend release](https://github.com/alphagov/govuk-frontend/pull/884).
- JavaScript needed for the navigation on mobile devices not added yet, will be part of the admin layout. (l.e. added in https://github.com/alphagov/govuk_publishing_components/pull/415)

[View in component guide](https://govuk-publishing-compon-pr-408.herokuapp.com/component-guide/layout_header)

[Trello card](https://trello.com/c/faHdZDk2)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -1,11 +1,11 @@
 // This component relies on styles from GOV.UK Frontend
 
-.gem-c-header--integration .govuk-header__container,
-.gem-c-header--staging .govuk-header__container {
+.gem-c-layout-header--integration .govuk-header__container,
+.gem-c-layout-header--staging .govuk-header__container {
   border-bottom-color: govuk-colour("yellow");
 }
 
-.gem-c-header--development .govuk-header__container {
+.gem-c-layout-header--development .govuk-header__container {
   border-bottom-color: govuk-colour("grey-1");
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -1,0 +1,30 @@
+// This component relies on styles from GOV.UK Frontend
+
+.gem-c-header--integration .govuk-header__container,
+.gem-c-header--staging .govuk-header__container {
+  border-bottom-color: govuk-colour("yellow");
+}
+
+.gem-c-header--development .govuk-header__container {
+  border-bottom-color: govuk-colour("grey-1");
+}
+
+// Fix header component's reliance on markup whitespace
+// https://github.com/alphagov/govuk-frontend/pull/884/files
+//
+// Remove this fix after updating to GOV.UK Frontend 1.1.0
+.govuk-header__container {
+  @include govuk-clearfix;
+}
+
+.govuk-header__logo {
+  @include mq ($from: desktop) {
+    float: left;
+  }
+}
+
+.govuk-header__content {
+  @include mq ($from: desktop) {
+    float: left;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_layout_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_header.html.erb
@@ -1,0 +1,60 @@
+<%
+navigation_classes ||= ''
+  navigation_items ||= []
+  css_classes ||= ''
+  container_classes ||= ''
+  homepage_url ||= '/'
+  product_name ||= nil
+  service_name ||= nil
+  service_url ||= nil
+%>
+<header class="gem-c-header govuk-header gem-c-header--<%= environment %>" role="banner" data-module="header">
+  <div class="govuk-header__container <%= container_classes.present? || 'govuk-width-container' %>">
+
+    <div class="govuk-header__logo">
+      <a href="<%= homepage_url %>" class="govuk-header__link govuk-header__link--homepage">
+        <span class="govuk-header__logotype">
+          <svg role="presentation" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 132 97" height="32" width="36">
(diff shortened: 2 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -0,0 +1,34 @@
+// This component relies on styles from GOV.UK Frontend
+
+.govuk-header--production .govuk-header__container {
```
nickcolley: Should this use the 'gem-' namespace?

alex-ju: Good point raised here @nickcolley. I would argue that we should keep the prefix of the component when extending it with modifiers in order to identify the source of the component. It can be indeed more difficult to figure out where the modifier class comes from though.

nickcolley: Since the root component still uses the upstream namespace, your modifier could show that it belongs in this gem, while still having the indication that most of the code is upstream?

nickcolley: Otherwise people may find it confusing that these modifiers are not in GOV.UK Frontend.

It also can mean that if GOV.UK Frontend introduces a modifier that you do not conflict (not so likely in this case, but as a general rule it could be sensible)

alex-ju: Sounds reasonable, will update.

app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -2,12 +2,10 @@
 <%= render 'govuk_publishing_components/components/title', title: @component_doc.name, context: "Component" %>
 
 <% if @component_doc.part_of_admin_layout? %>
-<div class="component-warning">
-  <h2 class="component-warning__title">Admin layout only</h2>
-  <p>
-    This component uses the Design System, so only works inside the <%= link_to "admin layout", "/component-guide/layout_for_admin" %>
-  </p>
-</div>
+<%= render "govuk_publishing_components/components/notice", {
```
tijmenb: Done in https://github.com/alphagov/govuk_publishing_components/pull/411

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -0,0 +1,61 @@
+<%
+navigation_classes ||= ''
+  navigation_items ||= []
+  css_classes ||= ''
+  container_classes ||= ''
+  homepage_url ||= '/'
+  assets_path ||= '/assets/images'
+  product_name ||= nil
```
tijmenb: What's the difference between `product_name` and `service_name`?

app/views/govuk_publishing_components/component_guide/show.html.erb
alex-ju: Rebased after #411 was merged.

app/views/govuk_publishing_components/components/_layout_header.html.erb
alex-ju: Product name is the one displayed next to GOV.UK on the lest side (e.g. https://www.payments.service.gov.uk/), the product name is mainly used for services and is displayed on the two-thirds container (e.g. https://www.registertovote.service.gov.uk/register-to-vote/country-of-residence)

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -0,0 +1,61 @@
...
+            ></path>
+            <image src="<%= assets_path %>/govuk-logotype-crown.png" class="govuk-header__logotype-crown-fallback-image"></image>
```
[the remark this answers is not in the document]
alex-ju: Sounds great. Will update!

app/views/govuk_publishing_components/components/_layout_header.html.erb
tijmenb: Ah ok thanks! Maybe we should start just with supporting product name? I think we'll go with that for the admin apps.

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
```diff
@@ -14,6 +14,7 @@ body: |
 
   Typically you'll use this together with:
 
+  - the [layout header component](/component-guide/layout_header)
```
tijmenb: Can you use this component in the admin example layout (/admin) too?

alex-ju: Yes, we can. Wasn't sure if I should add this as part of this PR. I can raise a new PR to use both the header and the footer.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -0,0 +1,60 @@
...
+  homepage_url ||= '/'
+  product_name ||= nil
+  service_name ||= nil
+  service_url ||= nil
+%>
+<header class="gem-c-header govuk-header gem-c-header--<%= environment %>" role="banner" data-module="header">
+  <div class="govuk-header__container <%= container_classes.present? || 'govuk-width-container' %>">
+
+    <div class="govuk-header__logo">
+      <a href="<%= homepage_url %>" class="govuk-header__link govuk-header__link--homepage">
```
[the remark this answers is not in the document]
alex-ju: Tried but I get undefined local variable or method `root_path`

alex-ju:
@tijmenb updated:
- stripped the header component from optional parameters and make sure there are tests are covering all parameters
- added the header and the footer to the [admin layout](https://govuk-publishing-compon-pr-408.herokuapp.com/admin)

PR #414, opened 2018-07-11
Create the "main" component for admin layout

tijmenb:
This component adds the "Layout main" component. It currently only makes the content use the grid (a single column). In the future this component will also contain the flash messages and other page essentials.

https://trello.com/c/4Z5AAIPI/15-create-main-component

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
@@ -11,4 +11,5 @@ $govuk-font-url-function: "font-url";
 
 // Components that are only available inside the admin layout
 @import "govuk_publishing_components/components/layout-footer";
+@import "govuk_publishing_components/components/layout-main";
 @import "govuk_publishing_components/components/layout-header";
--- a/app/views/govuk_publishing_components/components/_layout_main.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_main.html.erb
@@ -1,0 +1,3 @@
+<div class="gem-c-layout-main govuk-width-container">
+  <%= yield %>
+</div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * The Notice component now accepts blocks (PR #407)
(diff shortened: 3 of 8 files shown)


app/views/govuk_publishing_components/components/_layout_main.html.erb
```diff
@@ -0,0 +1,3 @@
+<div class="gem-c-layout-main govuk-width-container">
```
nickcolley: Does this intend to be the semantic area that the skip-link links too?

In the regular page template we have:

```html
  <div class="govuk-width-container">
    <!-- breadcrumbs, backlink, phasebanner goes here-->
    <main class="govuk-main-wrapper " id="main-content" role="main">
      <h1 class="govuk-heading-xl">Default page template</h1>
    </main>
  </div>
```

https://design-system.service.gov.uk/styles/page-template/

alex-ju: I have the main container in the skip link PR coming up. But can be added here too.

app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
```diff
@@ -11,4 +11,5 @@ $govuk-font-url-function: "font-url";
 
 // Components that are only available inside the admin layout
 @import "govuk_publishing_components/components/layout-footer";
-@import "govuk_publishing_components/components/layout-header";
+@import "govuk_publishing_components/components/layout-main";
+@import "govuk_publishing_components/components/layout-header"
```
alex-ju: Missing semicolon fails linting

app/views/govuk_publishing_components/components/_layout_main.html.erb
tijmenb: Hmm, I'm not sure. Let's make this the responsibility of the app at first, after that we'll see how to abstract it into this component.

tijmenb:
Talked about this offline with @alex-ju after https://github.com/alphagov/govuk_publishing_components/pull/414#discussion_r201998780. Closing this PR, we're going to require the applications to set the width container and associated areas themselves. Once we know how that works we might extract it into a component, but we might not need it.

PR #417, opened 2018-07-12
Add skip link component and main container to admin layout

alex-ju:
This PR:
- adds skip link component based on GOVUK Frontend
- fix: moves the header and footer from the admin layout published in gem to the dummy admin layout in spec
- adds container wrapper and main element to the admin layout
- updates changelog

---

Component guide for this PR:
https://govuk-publishing-compon-pr-417.herokuapp.com/admin

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/skip_link.yml
+++ b/app/views/govuk_publishing_components/components/docs/skip_link.yml
@@ -1,0 +1,16 @@
+name: Skip link (experimental)
+description: Skip link component helps keyboard-only users skip to the main content on a page
+part_of_admin_layout: true
+govuk_frontend_components:
+  - skip-link
+accessibility_criteria: |
+  - accept focus
+  - be focusable with a keyboard
+  - be usable with a keyboard
+  - indicate when they have focus
+  - becomes visible when pressing tab
+  - is the first items that screen readers hear and that keyboard users tab to
+examples:
+  default:
+    data:
+      href: '#content'
--- a/spec/dummy/app/views/welcome/admin_example.html.erb
+++ b/spec/dummy/app/views/welcome/admin_example.html.erb
@@ -1,4 +1,4 @@
-<%= render 'govuk_publishing_components/components/title', title: "Hello! I am an example admin page" %>
+<h1 class="govuk-heading-xl">Hello! I am an example admin page</h1>
 
 <p class="govuk-body">
   Pages with the admin layout can use GOV.UK styles directly.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,17 @@
 
 ## Unreleased
 
-* The Notice component now accepts blocks (PR #407)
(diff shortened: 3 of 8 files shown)


app/views/govuk_publishing_components/components/docs/skip_link.yml
```diff
@@ -0,0 +1,11 @@
+name: Skip link
+description: Skip link component helps keyboard-only users skip to the main content on a page
+accessibility_criteria: |
+  - the link is the first items that screen readers hear and that keyboard users tab to
+  - is becomes visible when pressing tab
```
maxgds: typo, remove "is"

app/views/govuk_publishing_components/components/docs/skip_link.yml
```diff
@@ -0,0 +1,11 @@
...
+shared_accessibility_criteria:
+- link
```
maxgds: Not all link criteria are relevant, (the link doesn't have visible text, for example, until it's in focus) and the component IS a link, rather than a container for links. I'd pick and choose the relevant ones and put them in criteria and leave out the shared ones.

spec/dummy/app/views/welcome/admin_example.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render 'govuk_publishing_components/components/title', title: "Hello! I am an example admin page" %>
+<h1 class="govuk-heading-xl">Hello! I am an example admin page</h1>
```
maxgds: Why this change? Isn't this is what the title component is for? If you need to add a class and the component doesn't allow, we should probably extend it rather than not use it. Or was it rendering a h2 or something previously?

app/views/govuk_publishing_components/components/docs/skip_link.yml
alex-ju: Updated as suggested (dropped the shared criteria an rewrite explicitly

spec/dummy/app/views/welcome/admin_example.html.erb
alex-ju: Since we're trying to use the GOVUK Frontend components (which, for example, only have margins at the bottom) and layout for the admin interface I switched to using a simple heading instead of a title component

maxgds: Ok, I think that makes sense. I'm used to using the existing gem components everywhere so maybe need to get out of that mindset.

alex-ju: I agree, I intend to do the same (use gem components everywhere possible), but in this case the title component brings more than I need to the example.

nickcolley: Perhaps the 'core' (https://github.com/alphagov/govuk-frontend/tree/master/src/core) parts of GOV.UK Frontend do not need to be wrapped but any 'components' should.

One of the best bits of having everything wrapped as components is if we need to change any markup (for example the class names change) it can be done in one place.

PR #418, opened 2018-07-12
Add anti-forgery meta tags

maxgds:
This has added the cross site forge request headers

However I can't work out how to test it. The required meta tags appear as expected in the example pages, but they aren't there in the context of the rspec test. I suspect this is related to forgery protection being off by default in a test environment, but can't work out how to switch it back on. Input welcome!

---

Component guide for this PR:
https://govuk-publishing-compon-pr-418.herokuapp.com/component-guide/


tijmenb:
I'd be fine with not testing this here. If the tag isn't present, Rails applications just won't work, so perhaps that's enough testing.

PR #424, opened 2018-07-16
Attempt to improve Google snippet display

andysellick:
Google is currently displaying snippets for some of our step by step pages like this:

screen shot 2018-07-12 at 13 27 02

This PR attempts to improve the snippet display by:

- wrapping the 'show' elements in brackets (hidden to users, both visually and with aria)
- lowercasing the 'show' elements (using CSS to capitalize, don't know if this will actually work)

This hopefully will produce a snippet more like this:

"Step 2 Update your vehicle's log book (V5C) (show). ... "

---

Trello card: https://trello.com/c/xTkNr64l/712-fix-google-snippet-display
Component guide for this PR:
https://govuk-publishing-compon-pr-424.herokuapp.com/component-guide/

### DIFF
--- a/spec/javascripts/components/step-by-step-nav-spec.js
+++ b/spec/javascripts/components/step-by-step-nav-spec.js
@@ -448,7 +448,7 @@ describe('A stepnav module', function () {
 
     it("sets the show/hide link text to 'hide'", function () {
       var $step1 = $element.find('#topic-step-one');
-      expect($step1.find('.js-toggle-link')).toHaveText("Hide");
+      expect($step1.find('.js-toggle-link')).toHaveText("(Hide)");
     });
 
     it("sets the show all/hide all button text correctly", function () {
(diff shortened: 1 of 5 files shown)


spec/javascripts/components/step-by-step-nav-spec.js
```diff
@@ -448,7 +448,7 @@ describe('A stepnav module', function () {
 
     it("sets the show/hide link text to 'hide'", function () {
       var $step1 = $element.find('#topic-step-one');
-      expect($step1.find('.js-toggle-link')).toHaveText("Hide");
+      expect($step1.find('.js-toggle-link')).toHaveText("(Hide)");
```
DilwoarH: do we need another test to check for "(Show)"

andysellick: This test is to check that the text for an open step is set correctly. There should be other tests unrelated to this change that check that the text is correct when opening and closing steps.

Turns out those tests don't exist, so I've added a card to get them implemented, but that's outside the scope of this PR (https://trello.com/c/2MXu4wYS/525-extend-tests-for-show-hide-text)

PR #429, opened 2018-07-17
Improve metadata positioning in doc list for right to left layouts

vanitabarrett:
Move the metadata to the right when on a right to left layout. Otherwise it's difficult to see how the metadata is related to the document list item heading

## Before
screen shot 2018-07-17 at 13 33 52

## After
screen shot 2018-07-17 at 13 39 13

Component guide: https://govuk-publishing-compon-pr-429.herokuapp.com/component-guide/document_list

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -17,6 +17,11 @@
 
 .gem-c-document-list__item-title--context {
   margin-right: $gutter-one-third;
+
+  .direction-rtl & {
+    margin-right: 0;
+    margin-left: $gutter-one-third;
+  }
 }
 
 .gem-c-document-list__item-context {
@@ -29,9 +34,14 @@
 
 .gem-c-document-list__attribute {
   @include core-14;
-  float: left;
+  display: inline-block;
   list-style: none;
   padding-right: $gutter-two-thirds;
+
+  .direction-rtl & {
+    padding-right: 0;
+    padding-left: $gutter-two-thirds;
+  }
 }
 
 .gem-c-document-list--bottom-margin {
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
```diff
@@ -32,6 +32,10 @@
   float: left;
   list-style: none;
   padding-right: $gutter-two-thirds;
+
+  .direction-rtl & {
+    float: right;
+  }
```
andysellick: Does this also need to have the padding-right changed to padding-left? The indenting of the date doesn't look quite right...

screen shot 2018-07-17 at 12 29 19

app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -116,3 +116,21 @@ examples:
...
+          description: "Statutory guidance for schools, local authorities and the police on dealing with poor attendance and behaviour in school"
+        metadata:
+          public_updated_at: 2017-01-05 14:50:33
+          document_type: 'Statutory guidance'
+      - link:
+          text: 'School exclusion'
+          path: '/government/publications/school-exclusion'
+        metadata:
+          public_updated_at: 2017-07-19 15:01:48
+          document_type: 'Statutory guidance'
```
andysellick: Would be good to test a rtl example with some `gem-c-document-list__item-context` elements included.

app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
```diff
@@ -32,6 +32,10 @@
   float: left;
   list-style: none;
   padding-right: $gutter-two-thirds;
+
+  .direction-rtl & {
+    float: right;
```
andysellick: Instead of floating this right you could simply take the float left off the original element style and replace it with display inline-block - it's then influenced by the text direction but uses less CSS (of course, I haven't tested this cross browser but I think it should work).

vanitabarrett: That sounds good, I wonder why we used float to start with.. I'll have a go 👍

PR #432, opened 2018-07-17
Extend the double dot problem solution

andysellick:
The double dot problem occurs where a page occurs more than once in a single step by step navigation. The backend doesn't know which link was clicked, so we use JS to store the clicked link in sessionStorage, then work out which link and step to highlight when the page loads again based on that. The aims are:

- to ensure that no more than one link and one step are highlighted
- to ensure that the last link the user clicked on (assuming it is still in a step by step navigation) is the one that is highlighted

screen shot 2018-07-17 at 16 24 41

This solution was previously implemented only to handle highlighting of the links themselves, but has been expanded to include the highlighting of the steps as well.

- now handles highlighting of both the links and the steps
- uses the step nav id to create a unique session store attribute, so cannot conflict with another step nav

---

- Trello card: https://trello.com/c/U3bgGMWy/722-incorrect-step-opened-by-default-when-same-link-appear-multiple-times-in-single-step-by-step
- Component guide for this PR:
https://govuk-publishing-compon-pr-432.herokuapp.com/component-guide/step_by_step_nav/solve_the_double_dot_problem

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -9,6 +9,7 @@
     var stepNavSize;
     var sessionStoreLink = 'govuk-step-nav-active-link';
     var activeLinkClass = 'gem-c-step-nav__list-item--active';
+    var activeStepClass = 'gem-c-step-nav__step--active';
     var activeLinkHref = '#content';
     var uniqueId;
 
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -29,6 +30,7 @@
       var $showOrHideAllButton;
 
       uniqueId = $element.data('id') || false;
+      sessionStoreLink = sessionStoreLink + "_" + uniqueId;
```
leenagupte: What happens here if `uniqueId` is `false`?

andysellick: @leenagupte that's an excellent question - I hadn't considered that possibility. The functionality depends upon that, so I'll have a think about how this should be handled.

andysellick: I've put a check in so if `uniqueId` is false `sessionStoreLink` is just 'govuk-step-nav-active-link', and added a note to the documentation.

PR #433, opened 2018-07-17
Update radio component

alex-ju:
This PR updates the radio component to:
 - use govuk-frontend styling which requires:
   - updating class names 
   - update radio divider (or) implementation
   - update bold option implementation
   - use Rails helpers to generate markup
 - ~~adds conditional reveal support~~ (will be part of a separate PR)

This PR does not address the label, hint and fieldset, which will be approached in separate PRs.

Recommend reviewing-by-commit.

[Trello card](https://trello.com/c/g3BbaJhv)

---

Component guide for this PR:
https://govuk-publishing-compon-pr-433.herokuapp.com/component-guide/radio


app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -117,3 +120,14 @@ examples:
         hint_text: "Lorem ipsum dolor sit amet, consectetur adipiscing elit. Phasellus sapien justo, lobortis elementum tortor in, luctus interdum turpis. Nam sit amet nulla nec arcu condimentum dapibus quis varius metus. Suspendisse cursus tristique diam et vestibulum. Proin nec lacinia tortor. Morbi at nisi id lorem aliquam ullamcorper. Pellentesque laoreet sit amet leo sodales ultricies. Suspendisse maximus efficitur odio in tristique."
         text: "Quisque tincidunt venenatis bibendum. Morbi volutpat magna euismod ipsum consequat cursus. Etiam bibendum interdum ultricies."
         bold: true
+  conditional:
```
tijmenb: This is a nice approach, but I wonder if we need to change something to support revealing other elements than radio buttons. I'm specifically thinking about the  [review reminder](https://whitehall-prototype.herokuapp.com/content-settings), which reveals a block with other form things:

screen shot 2018-07-18 at 08 03 14

Or on the public site, the [contact GOV.UK form](https://www.gov.uk/contact/govuk), which would reveals a string input:

screen shot 2018-07-18 at 08 02 05

Perhaps we can pass in the conditional HTML to the option?

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -1,26 +1,36 @@
 <%
   id_prefix ||= "radio-#{SecureRandom.hex(4)}"
   items ||= []
+  is_conditional = false
 %>
+<% if items.any? %>
+<% items.each do |item| %>
```
andysellick: I think the indentation of this file could be improved, and some lines spaces could help break it up a bit.

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -29,17 +39,29 @@
...
       } %>
     </div>
+    <% if item[:conditional] %>
+    <div class="govuk-radios__conditional" id="<%= conditional_id %>">
+      <%= item[:conditional] %>
+    </div>
+    <% end %>
   <% end %>
 <% end %>
+<% if items.any? %>
```
andysellick: I'm not sure that having opening and closing divs in separate IF statements is a good approach. Have you considered using content_tag and a helper method to wrap some of this code?

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -1,26 +1,36 @@
...
     <%
       item_next = items[index + 1] unless index === items.size - 1
       label_id = item[:id] ? item[:id] : "#{id_prefix}-#{index}"
       label_hint_id = "label-hint-#{SecureRandom.hex(4)}" if item[:hint_text].present?
+      conditional_id = "conditional-#{SecureRandom.hex(4)}" if item[:conditional].present?
     %>
     <div
       class="
         gem-c-radio
+        govuk-radios__item
```
andysellick: This stacking approach to individual items (in this case classes) leads to more readable code but puts extra whitespace in the outputted markup, which over time is going to (fractionally) increase our HTML file size. An alternative approach could be to assemble the classes needed outside of the markup and then insert them, e.g.

```
<%
  classes = "gem-c-radio govuk-radios__item"
  classes << " gem-c-radio--margin-bottom-0" if item_next === :or && item[:hint_text].present?
%>

<input class="<%= classes %>">
```

app/assets/stylesheets/govuk_publishing_components/components/_radio.scss
```diff
@@ -1 +1,20 @@
-@import "../../../node_modules/govuk-frontend/components/radios/radios";
+// This component relies on styles from GOV.UK Frontend
+
+// Specify the functions used to resolve assets paths in SCSS
+$govuk-image-url-function: "image-url";
```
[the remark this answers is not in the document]
alex-ju: I was thinking about that, the only downside is that by doing so you're not able to inport a single component

app/assets/stylesheets/govuk_publishing_components/components/_label.scss
```diff
@@ -6,7 +6,7 @@
   @include core-19;
 }
 
-.gem-c-label--bold {
```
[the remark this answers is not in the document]
alex-ju: Yes, it's being used for making a label bold and has a test for it too. When/if we're updating the label component to use govuk-frontend we can replace it too.

app/views/govuk_publishing_components/components/_radio.html.erb
alex-ju: I can use `content_tag` if that makes it easier to read

app/views/govuk_publishing_components/components/_radio.html.erb
alex-ju: Fixed indentation

app/views/govuk_publishing_components/components/docs/radio.yml
tijmenb: I think I completely misread the PR at first, it already does what I suggested.. Never mind me!

app/views/govuk_publishing_components/components/_radio.html.erb
alex-ju: Updated as suggested.

app/views/govuk_publishing_components/components/_radio.html.erb
alex-ju: @andysellick what's the advantage of using the `content_tag` with a helper?

andysellick: I assumed that as soon as you started using content_tag and passing in chunks of code the view would start to get overly complicated, at which point it would be good to move some of that code into a helper. The [step by step component](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb) does some nesting of elements like this, if that helps.

nickcolley: This is solved in GOV.UK Frontend by making the fieldset component an additional component then using the nesting interface 'caller'.

In ERB this can be achieved with 'yield'

something like

```
fieldset do
  radios
end
```

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -3,7 +3,9 @@ description: A radio button is a GOV.UK element that allows users to answer a qu
 body: |
   Forked from the upcoming [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), when GOV.UK Frontend release we can replace these source files.
```
nickcolley: I believe this line can be removed now?

app/views/govuk_publishing_components/components/_radio.html.erb
alex-ju: @nickcolley this is what I want to achieve, but I'm trying to do the minimum in this PR to update the radios and make them use Frontend then I'm planning to address the labels, hints and fieltset in separate PRs

app/assets/stylesheets/govuk_publishing_components/components/_label.scss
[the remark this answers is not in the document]
tijmenb: 👍

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -1,26 +1,31 @@
 <%
   id_prefix ||= "radio-#{SecureRandom.hex(4)}"
   items ||= []
+
+  is_conditional = false
```
tijmenb: You can use:

```rb
has_conditional = items.any? { |item| item["conditional"] }
```

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -1,26 +1,31 @@
...
+  items.each do |item|
+    is_conditional = true if item["conditional"].present?
+  end
 %>
+<% if items.any? %>
```
tijmenb: As discussed, this can go because if there are no items it's fine to show something.

PR #434, opened 2018-07-18
Add select dropdown component based on design system select

maxgds:
Only works in admin layout
https://trello.com/c/zW7BT3w9/28-add-select-dropdown-component

---

Component guide for this PR:
https://govuk-publishing-compon-pr-434.herokuapp.com/component-guide/select

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/select.yml
+++ b/app/views/govuk_publishing_components/components/docs/select.yml
@@ -1,0 +1,35 @@
+name: Select (experimental)
+description: A dropdown select
+part_of_admin_layout: true
+govuk_frontend_components:
+  - select
+accessibility_criteria: |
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/select.yml
```diff
@@ -0,0 +1,34 @@
+name: Select (experimental)
+description: A dropdown select
+part_of_admin_layout: true
+govuk_frontend_components:
+  - select
+accessibility_criteria: |
+  - accept focus
```
andysellick: To make this a little more readable, I'd add in something like "The component must:"

maxgds: Fixed, thanks.

PR #436, opened 2018-07-19
Remove redundant styles from contents list

andysellick:
These styles (white background, box shadow and z index) were added prior to universal layout to ensure position fixed back to top link disappeared beneath the contents list in the left hand column. Both the layout and the contents list have now changed, so this CSS is no longer needed

---

Component guide for this PR:
https://govuk-publishing-compon-pr-436.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


andysellick:
@vanitabarrett drat. I didn't know that. Thanks for pointing it out!

PR #435, opened 2018-07-19
Change breadcrumb class name

andysellick:
- from 'gem-c-breadcrumbs--item' to 'gem-c-breadcrumbs__item'
- was not following BEM convention, 'item' is an element, not a modifier

This should also fix the problem with this component in the 'highlight components' option of the GOV.UK browser extension, see below.

screen shot 2018-07-19 at 09 57 25

---

Component guide for this PR:
https://govuk-publishing-compon-pr-435.herokuapp.com/component-guide/

PR #437, opened 2018-07-20
Version 9.5.4

andysellick:
* Change breadcrumb class name (PR #435)
* Add admin component for select (PR #434)
* Extend the double dot problem solution (PR #432)

---

Component guide for this PR:
https://govuk-publishing-compon-pr-437.herokuapp.com/component-guide/


vanitabarrett:
Not sure if this should be a minor version rather than a patch as it includes #434 ?

PR #439, opened 2018-07-20
Update button component

alex-ju:
This PR updates the button component to use GOV.UK Frontend styles:
 - updates class names
 - updates govspeak buttons to extend govuk-frontend classes
 - removes the implicit responsive bottom margin set by default for each component by govukfrontend and continue to rely on setting a `margin_bottom` attribute to apply the `gem-c-button--bottom-margin` class until we make a decision about applying spacing consistently across components.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-439.herokuapp.com/component-guide/button

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Update radio component to use GOV.UK Frontend styles (PR #433)
+* Update button component to use GOV.UK Frontend styles (PR #439)
 
 ## 9.6.0
 


app/assets/stylesheets/govuk_publishing_components/components/_button.scss
```diff
@@ -1,20 +1,18 @@
...
+@import "mixins/margins";
 
-  &:focus {
-    outline: 3px solid $focus-colour;
-  }
+// Because govuk-frontend adds a responsive bottom margin by default for each component
+// we reset it to zero so we can set it separately using `gem-c-button--bottom-margin`
+// If we decide to use responsive margins consistently across components we can remove this
+.gem-c-button {
+  margin-bottom: 0;
```
nickcolley: Could this instead be replaced with `.govuk-!-margin-bottom-0` in the template?

alex-ju: yes, I can use the overwrite classes. good point.

alex-ju: ah, I need to import the overwrite helpers for that, not sure if it's worth doing that for each component. Ideally we would settle for consistent spacing across component and we will remove these resets.

nickcolley: Fair, not blocking, you do you. 👍

PR #440, opened 2018-07-25
Update back-link component

alex-ju:
The font style for this component is currently coming from govuk_template. This PR updates the component to use GOV.UK Frontend styles which also extends component's compatibility (the left chevron) down to IE8.

### Before
&nbsp;&nbsp;screen shot 2018-07-25 at 12 35 39

### After
screen shot 2018-07-25 at 12 36 28

---

Component guide for this PR:
https://govuk-publishing-compon-pr-440.herokuapp.com/component-guide/back-link

### DIFF
--- a/spec/components/back_link_spec.rb
+++ b/spec/components/back_link_spec.rb
@@ -13,11 +13,11 @@ describe "Back Link", type: :view do
 
   it "renders a back link correctly" do
     render_component(href: '/back-me')
-    assert_select ".gem-c-back-link[href=\"/back-me\"]", text: "Back"
+    assert_select ".govuk-back-link[href=\"/back-me\"]", text: "Back"
   end
 
   it "can render in welsh" do
     I18n.with_locale(:cy) { render_component(href: '/back-me') }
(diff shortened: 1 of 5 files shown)


spec/components/back_link_spec.rb
```diff
@@ -13,11 +13,11 @@ def component_name
 
   it "renders a back link correctly" do
     render_component(href: '/back-me')
-    assert_select ".gem-c-back-link[href=\"/back-me\"]", text: "Back"
+    assert_select ".govuk-back-link[href=\"/back-me\"]", text: "Back"
```
tijmenb: I think the tests will pass without this change - might be better?

alex-ju: Yes, they still pass without the change. I thought would be good to rely on `.govuk` selectors for consistency (we're currently querying elements using `.govuk` classes) since we're aiming to keep `.gem-c-back-link` for highlighting components, not I can roll that back if you think we should continue to use `.gem-c` selectors where possible.

PR #441, opened 2018-07-25
Add conditional reveal support for radios

alex-ju:
This PR adds conditional reveal support for radios using GOV.UK Frontend scripts.

### Integration
Because GOV.UK Frontend components are not compatible (they are being initialised differently) with the `GOVUK.Modules` script, I attached them to a `GOVUK.FrontendModules` global.

For example the Radios script can be accessed via `GOVUK.FrontendModules.Radios`.

### Cross-browser support
This PR also contains a `nodeListForEach` utility function to iterate through each module instance on the page (if multiple). This could potentially be replaced by a Polyfill and placed at a higher lever to be shared between components.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-441.herokuapp.com/component-guide/radio/conditional/preview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Update radio component to use GOV.UK Frontend styles (PR #433)
 * Update button component to use GOV.UK Frontend styles (PR #439)
 * Update back-link component to use GOV.UK Frontend styles (PR #440)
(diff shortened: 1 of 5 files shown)


alex-ju:
@kevindew thanks for the heads-up on how we attach GOV.UK Frontend modules to globals.
I updated the code to integrate the new scripts while follow the existing way of attaching functions and added more details about it to this PR description.

PR #446, opened 2018-07-30
Add error message and hint

alex-ju:
This PR adds error message and hint as components and updates the two components using them (input and radio).

The reason for having these as separate components instead of being part of the label component is that they should also be able to be used for fieldsets in legends so decoupling them allows them to be used in other circumstances.

---
[Trello card](https://trello.com/c/4iwUWRwz)
Component guide for this PR:
https://govuk-publishing-compon-pr-446.herokuapp.com/component-guide/

### DIFF
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -1,32 +1,52 @@
 <%
   id ||= "input-#{SecureRandom.hex(4)}"
-  hint_id ||= "hint-#{SecureRandom.hex(4)}"
   value ||= nil
-  error_message ||= false
-  label ||= {}
   type ||= "text"
   describedby ||= false
-  ariadescribedby ||= nil
+
+  label ||= nil
+  hint ||= nil
+  error_message ||= nil
+  hint_id = "hint-#{SecureRandom.hex(4)}" if hint
+  error_message_id = "error-message-#{SecureRandom.hex(4)}" if error_message
+
   css_classes = %w(gem-c-input govuk-input)
   css_classes << "govuk-input--error" if error_message
-  hint_text_css_classes = "govuk-error-message" if error_message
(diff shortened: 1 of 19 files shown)


app/views/govuk_publishing_components/components/docs/error_message.yml
```diff
@@ -0,0 +1,14 @@
+name: Form error message
+description: Use error messages for any form fields.
+accessibility_criteria: |
```
tijmenb: can you add the `govuk_frontend_components` to this, so that it'll link to the design system?

alex-ju: Updated.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -1,32 +1,52 @@
 <%
   id ||= "input-#{SecureRandom.hex(4)}"
-  hint_id ||= "hint-#{SecureRandom.hex(4)}"
   value ||= nil
-  error_message ||= false
-  label ||= {}
   type ||= "text"
   describedby ||= false
-  ariadescribedby ||= nil
```
tijmenb: Is this a breaking change?

alex-ju: no, `ariadescribedby` was introduced to a few commit ago (29ef8a8) to concatenate the ids of the items than need to be read out by assistive tech.
At that point it should've been `ariadescribedby = nil` without `||`, so no breaking change. Now we use `aria_described_by` for that purpose, to be consistent with the other variables names.

tijmenb: Sorry, I meant is any app using `ariadescribedby` instead of `aria_described_by`? Would be good to accept the old param with a deprecation warning if so.

tijmenb: 👍

PR #447, opened 2018-07-30
Update feedback component

alex-ju:
This PR:
 - updates the feedback component to use the button components
 - applies a stronger selector for inputs; static overwrites the input styles coming from GOV.UK Frontend using `input[type="text"]` as selector, so we need to apply `.govuk-input` styles with a stronger selector until the styles from static will be removed.

Below is an example of how this component is rendered in gov.uk and how it's rendered after the update. While the styles overwrite could be done in the applications I consider it's better to be done here since there are strong chances that this component will be loaded with static styles in other projects.

### Before
screen shot 2018-07-30 at 09 54 29

### After
screen shot 2018-07-30 at 09 55 28

---

Component guide for this PR:
https://govuk-publishing-compon-pr-447.herokuapp.com/component-guide/feedback

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -1,6 +1,5 @@
 // govuk_frontend_toolkit
-@import 'grid_layout';
-@import 'design-patterns/buttons';
+@import "grid_layout";
 
 .gem-c-feedback {
   max-width: 960px;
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -178,6 +177,11 @@
   }
 }
 
-.gem-c-feedback__submit {
-  @include button;
+// static.css on GOV.UK overwrites the component styles using input[type="text"]
+// so we need to apply  govuk-input styles using a stronger selector
+.gem-c-feedback input[type="text"] {
```
andysellick: It might be good to fix this in the input component itself rather than here, but I guess we're working towards killing off static.css anyway so maybe not worth the effort.

alex-ju: Thanks @andysellick! Good point about having it in input. For the moment this is the only place I've seen the input being with static, I'll scout more potentially move it if necessary.

PR #451, opened 2018-07-30
Add resets to document list

alex-ju:
This PR adds resets paddings and margins and adds font styles without having the component rely on the reset CSS.

Here is an example of how the component looks like without reset CSS.

### Before
screen shot 2018-07-30 at 15 47 42

### After
screen shot 2018-07-30 at 15 47 01

---

Component guide for this PR:
https://govuk-publishing-compon-pr-451.herokuapp.com/component-guide/document_list


alex-ju:
thanks @andysellick! good catch. added a class on `ul` to reset the padding

PR #453, opened 2018-07-31
Add details component

tijmenb:
The details component is useful to reveal more detailed information.

https://govuk-publishing-compon-pr-453.herokuapp.com/component-guide/details

https://trello.com/c/d1jqgiXk/54-add-details-component

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_details.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_details.scss
@@ -1,0 +1,1 @@
+// This component relies on styles from GOV.UK Frontend
--- a/app/views/govuk_publishing_components/components/docs/details.yml
+++ b/app/views/govuk_publishing_components/components/docs/details.yml
@@ -1,0 +1,24 @@
+name: Details
+description: Make a page easier to scan by letting users reveal more detailed information only if they need it
+shared_accessibility_criteria:
+  - link
+accessibility_criteria: |
+  The component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * be usable with touch
+  * indicate when it has focus
+  * toggle the visibility of the details element's content when interacted with
+  * indicate the expanded state when details' content is visible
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_details.scss
```diff
@@ -0,0 +1 @@
+// Use the Details component from GOV.UK Frontend
```
andysellick: Is this file necessary? I get what it's here to explain, but is that important?

app/views/govuk_publishing_components/components/docs/details.yml
```diff
@@ -0,0 +1,21 @@
...
+shared_accessibility_criteria:
+  - link
+accessibility_criteria: |
+  The component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * be usable with touch
+  * indicate when it has focus
```
alex-ju: I would add:
* toggle the visibility of the details element's content when interacted with
* indicate the expanded state when details' content is visible
* indicate the collapsed state when details' content is hidden

app/assets/stylesheets/govuk_publishing_components/components/_details.scss
tijmenb: Yeah, we've done this for every component that uses GOV.UK Frontend. I think it's important so that devs can go to any component's CSS file and discover where the styling comes from. If we didn't have it, it would be up to the developer to find out where the styles come from.

app/views/govuk_publishing_components/components/docs/details.yml
tijmenb: Thanks, updated.

alex-ju: Can't see them in yet. Happy to approve after they're added.

tijmenb: They're there 👇

PR #454, opened 2018-07-31
Refactor button component

tijmenb:
Move all of the variables into the class and use better view helpers to create the tags.


tijmenb:
@andysellick not necessarily, if the component is tested sufficiently in the view tests. View tests are preferable over unit tests in this case anyway, since it allows refactoring without changing tests.

PR #458, opened 2018-07-31
Changes required when there are only one "part of" step by step navs

DilwoarH:
## Changes
- Increase size of font to be the same size as the step by step headings
- Increase the margin-top for the "part of" text to $gutter-two-third
- Decrease the margin-bottom for the "part of" text to $gutter-one-quarter
- Decrease the margin-bottom of step by step title to $gutter-one-third + 3

## Notes:
Change number 4 is being done as there is a padding of 7px on the show all button which makes the margin-bottom seem much larger.

This should not affect cases where there are multiple step by steps.

Ticket: https://trello.com/c/zlQ0Go3P/700-fix-the-step-by-step-title-sizing-and-spacing-on-mobile-when-it-appears-as-a-footer-s

## Screenshot
screen shot 2018-07-31 at 14 16 11

---

Component guide for this PR:
https://govuk-publishing-compon-pr-458.herokuapp.com/component-guide/step_by_step_nav_related

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-related.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-related.scss
@@ -15,6 +15,19 @@
   padding: 0;
 }
 
+.gem-c-step-nav-related--singular {
+  margin-bottom: $gutter-one-third + 3;
+
+  .gem-c-step-nav-related__heading {
+    @include _core-font-generator(19px, 19px, 19px, 1.4, 1.4, false, bold);
+    margin-top: $gutter-two-thirds;
+  }
+
+  .gem-c-step-nav-related__pretitle {
+    margin-bottom: $gutter-one-quarter;
+  }
+}
+
 .gem-c-step-nav-related__pretitle {
   display: block;
   margin-bottom: $gutter-half;
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
@@ -3,7 +3,9 @@
   pretitle ||= t("govuk_component.step_by_step_nav_related.part_of", default: "Part of")
 %>
 <% if links.any? %>
-  <div class="gem-c-step-nav-related" data-module="track-click">
+  <div 
+    class="gem-c-step-nav-related <%= "gem-c-step-nav-related--singular" if links.length == 1 %>" 
+    data-module="track-click">
     <h2 class="gem-c-step-nav-related__heading">
       <span class="gem-c-step-nav-related__pretitle"><%= pretitle %></span>
       <% if links.length == 1 %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,7 +11,7 @@
 
 * Step by step component, Google snippet improvement (PR #461)
 * Add tabs component (PR #455)
-
+* Adds styling for singular related step by step navs (PR #458)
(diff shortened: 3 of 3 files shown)


app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
```diff
@@ -3,7 +3,9 @@
   pretitle ||= t("govuk_component.step_by_step_nav_related.part_of", default: "Part of")
 %>
 <% if links.any? %>
-  <div class="gem-c-step-nav-related" data-module="track-click">
+  <div 
+    class="gem-c-step-nav-related <% if links.length == 1 %> gem-c-step-nav-related-singular <% end %>"
```
andysellick: Could be written more cleanly as

```
class="gem-c-step-nav-related <%= "gem-c-step-nav-related-singular" if links.length == 1 %>" 
```

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-related.scss
```diff
@@ -15,6 +15,22 @@
   padding: 0;
 }
 
+.gem-c-step-nav-related-singular {
```
andysellick: This class name doesn't follow our [BEM convention](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#styles) - should be `gem-c-step-nav-related--singular`.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-related.scss
```diff
@@ -15,6 +15,22 @@
...
+  margin-bottom: $gutter-one-third + 3;
+}
+
+.gem-c-step-nav-related-singular .gem-c-step-nav-related__heading {
```
andysellick: I think this whole section would be more readable if the classes were nested.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-related.scss
```diff
@@ -15,6 +15,22 @@
...
+  @include _core-font-generator(19px, 19px, 19px, 1.4, 1.4, false, bold);
```
andysellick: I've noticed you use the core-font-generator mixin a lot - are you aware you can use `@include core-19` or `@include bold-19` instead? ([see here](https://github.com/alphagov/govuk_frontend_toolkit/blob/452f82e7538d7f5b8f73ee7d81abea9bc857af13/stylesheets/_typography.scss#L92)). Obviously if you want to use custom line heights etc. then that's fine, although we should  use the standard mixins where reasonable, just for consistency and ease of maintenance.

DilwoarH: The reason why I used the core-font-generator is because we needed the related size to be consistent with the step title. https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss#L62

If I use bold-19 or core-19, the font size variable between mobile and desktop.

app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
DilwoarH: Thanks I will update this, we'll need one as we only want it to match if length is 1

DilwoarH:
@andysellick - updated

PR #459, opened 2018-07-31
Add guidance for how to pass HTML to components

nickcolley:
Feel free to close this, but I hope this is helpful.

This is documentation explaining a preferred way to construct components that accept HTML based on:

- [an incident involving multiple components](https://github.com/alphagov/govuk_publishing_components/pull/305)
- [a potential vulnerability with the govspeak component](https://github.com/alphagov/govuk_publishing_components/pull/356).

### DIFF
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -201,3 +201,75 @@ Code can be called and referred to in the template as follows:
 
 <%= card_helper.heading_tag %>
 ```
+
+## Passing HTML to a component
+Avoid marking HTML as safe within components, this means avoiding use of `raw` or `html_safe` within a component's view.
+
+By doing this we can avoid the risk of a Cross Site Scripting (XSS) attack.
+
+Only sanitise HTML when the component is used in the application.
+
+There are a few methods to achieve this:
+
+### Yielding blocks with single slots for nested children
+
+Similar to HTML, there may be a clear slot where nested children should appear in a component.
+
+For a panel component, you may expect anything nested within it to appear
+at the bottom of the component.
+
+Do not:
+
+```erb
+<%= render 'panel', content: "Your reference number<br><strong>HDJ2123F</strong>" %>
+```
+
+Do:
+
+```erb
+<%= render 'panel' do %>
+  Your reference number
+  <br>
+  <strong>HDJ2123F</strong>
+<% end %>
+```
+
+### Parameters with HTML for multiple slots
+
+If you have multiple slots where HTML may go, you could consider passing them as parameters.
+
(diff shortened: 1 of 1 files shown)


docs/component_conventions.md
```diff
@@ -201,3 +201,73 @@ Code can be called and referred to in the template as follows:
...
+  Your reference number
+  <br>
+  <strong>HDJ2123F</strong>
+<% end %>
+```
+
+Do:
+
+```erb
+<% panelTitle = capture do %>
```
tijmenb: Can we use `panel_title` for consistency with ruby?

nickcolley: Oops, sure thing, got my JavaScript hat on :P

docs/component_conventions.md
```diff
@@ -201,3 +201,73 @@ Code can be called and referred to in the template as follows:
 
 <%= card_helper.heading_tag %>
 ```
+
+## Passing HTML to a component
+To avoid the risk of Cross Site Scripting (XSS) attack, we only mark HTML as safe  only sanitise HTML when the component is used, rather than doing so on the application's behalf.
+
+This means avoiding use of 'raw' or 'html_safe' within a component's view.
```
tijmenb: Backticks for `raw` and `html_safe` will make 'm look nicer.

docs/component_conventions.md
```diff
@@ -201,3 +201,73 @@ Code can be called and referred to in the template as follows:
...
+
+There are a few methods to achieve this:
+
+### Single slots for nested children
```
tijmenb: Maybe instead of slots for children we can talk about parameters with HTML? Would that be clearer?

nickcolley: I'm using slots as it's the word used in Web Components, but I think your suggestion is much clearer.

tijmenb: This is the first time I've heard of Web Components :)

docs/component_conventions.md
```diff
@@ -201,3 +201,73 @@ Code can be called and referred to in the template as follows:
 
 <%= card_helper.heading_tag %>
 ```
+
+## Passing HTML to a component
+To avoid the risk of Cross Site Scripting (XSS) attack, we only mark HTML as safe  only sanitise HTML when the component is used, rather than doing so on the application's behalf.
```
andysellick: I think this is missing a word or some punctuation? `we only mark HTML as safe  only sanitise HTML`

nickcolley: Ah yeah, that's not right, thanks Andy.

nickcolley:
@andysellick updated that wording so it makes sense

PR #469, opened 2018-08-07
Add data-attribute support to input

alex-ju:
This PR adds the following to the input component:
-  data attributes (for JavaScript module initialisation - e.g. `data-module`; analytics tracking - e.g. `data-track`)

---

Component guide for this PR:
https://govuk-publishing-compon-pr-469.herokuapp.com/component-guide/input


alex-ju:
@davbo: seem the Heroku preview is not working. I would check to see why, but I don't have access.

PR #470, opened 2018-08-07
Add publishing_app to metatag

binaryberry:
At the moment smart-answers are being categorised as having a format of transaction in GA, because their start pages are being rendered as transactions.

Adding the publishing_app metatag would allow to differentiate between smart-answers and transactions in GA, the latter being published by Publisher.

https://trello.com/c/vhfMGngX/239-investigate-and-correct-smart-answer-google-analytics-tracking


lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -27,6 +27,7 @@ def meta_tags
 
       def add_core_tags(meta_tags)
         meta_tags["govuk:format"] = content_item[:document_type] if content_item[:document_type]
+        meta_tags["govuk:publishing-app"] = content_item[:publishing_app] if content_item[:publishing_app]
```
thomasleese: I wonder if it’s worth renaming this to “publishing-application” before this becomes too permanent  for consistency with the “rendering-application” meta tag.

PR #473, opened 2018-08-07
Improve step by step component double dot problem solving code

andysellick:
- shorten code
- if no sessionStorage value (i.e. user is arriving at the page cold) highlight the first active link and active step, not the last
- expand documentation to clarify how this behaves

See the documentation for full details. Expands the solution to this problem:

screen shot 2018-08-07 at 14 47 31

---

Trello card: https://trello.com/c/LTzIkmOX/776-optimise-default-open-step
Component guide for this PR:
https://govuk-publishing-compon-pr-473.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -232,32 +232,22 @@
           return;
         }
 
-        var lastClicked = loadFromSessionStorage(sessionStoreLink);
+        var lastClicked = loadFromSessionStorage(sessionStoreLink) || $element.find('.' + activeLinkClass).first().attr('data-position');
 
-        if (lastClicked) {
-          // it's possible for the saved link position value to not match any of the currently duplicate highlighted links
-          // so check this otherwise it'll take the highlighting off all of them
-          if (!$element.find('.js-link[data-position="' + lastClicked + '"]').parent().hasClass(activeLinkClass)) {
-            lastClicked = $element.find('.' + activeLinkClass).first().find('.js-link').attr('data-position');
-          }
-          removeActiveStateFromAllButCurrent($activeLinks, lastClicked);
-          setActiveStepClass();
-        } else {
-          var activeLinkInActiveStep = $element.find('.' + activeStepClass).find('.' + activeLinkClass).first();
-
-          if (activeLinkInActiveStep.length) {
-            $activeLinks.removeClass(activeLinkClass);
-            activeLinkInActiveStep.addClass(activeLinkClass);
-          } else {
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/step_by_step_nav.yml
```diff
@@ -340,22 +334,24 @@ examples:
...
+      JavaScript is included in the component to solve this. It uses sessionStorage to capture the data-position attribute of non-external links when clicked, and then uses this value to decide which link to highlight after the new page loads. It uses the value of the tracking_id option to uniquely identify the current step nav in the session (if this is not passed to the component this may result in other step navs having the wrong link highlighted).
 
-      The current page in the step by step navigation is an anchor link to the top of the page. If there are more than one of these and the user clicks one that is not currently highlighted, that one will be highlighted.
+      If a user has not clicked a link (i.e. has visited the page without first clicking on a step by step navigation) only the first duplicate link will be highlighted. If that link is not in the active step, the JS makes the highlighted link's parent the active step. If there is no active step, the first active link will be highlighted (but there should always be an active step).
```
leenagupte: > If that link is not in the active step, the JS makes the highlighted link’s parent the active step

How would this be possible?

andysellick: The code basically goes...

```
if we've got the same link twice in the component
AND
there's nothing in the session to tell us which link to highlight
AND
the first highlighted link is not in the active step
THEN
change the active step to be the one that the first highlighted link is in
```

andysellick: It's the thing I was talking about where the frontend code no longer trusts the backend to be right about which step should be highlighted and expanded, and takes matters into its own hands.

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -232,32 +232,22 @@
           return;
         }
 
-        var lastClicked = loadFromSessionStorage(sessionStoreLink);
+        var lastClicked = loadFromSessionStorage(sessionStoreLink) || $element.find('.' + activeLinkClass).first().attr('data-position');
 
-        if (lastClicked) {
```
DilwoarH: should you keep this in? What would happen if the active class couldn't be found either?

DilwoarH:
Other than that one comment, it looks good 👍

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
andysellick: If the active class isn't found then this entire function has already exited on line 231. The rest only gets called if there's more than one active class found.

PR #474, opened 2018-08-07
Version 9.12.0

binaryberry:
* Adds publishing-app metatag (PR #470)

https://trello.com/c/vhfMGngX/239-investigate-and-correct-smart-answer-google-analytics-tracking


andysellick:
@binaryberry is this a minor change? Just wondering if this should actually be version 9.11.1.

binaryberry:
@andysellick I think it is! I'm adding functionality, not removing a bug: https://semver.org/

PR #484, opened 2018-08-15
Add secondary and secondary quiet modifiers to button component

alex-ju:
This PR adds secondary and secondary quiet modifiers to button component needed for the admin interfaces.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-484.herokuapp.com/component-guide/button

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Updates the tabs component to make the heading inside the panel optional and to add a modifier for panel without border (PR #485)
+* Add secondary and secondary quiet modifiers to button component (PR #484)
 
 ## 9.14.0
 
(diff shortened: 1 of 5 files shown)


andysellick:
@alex-ju why is it called a 'secondary quite' button?

andysellick:
Spotted a problem with the focus state for these new buttons...

screen shot 2018-08-15 at 12 15 31

alex-ju:
Good catch @andysellick! For some reason the specific `:link:focus` selector that was meant to ensure compatibility with govuk_template is overwriting the `color` attribute implicitly (without turning compatibility with govuk_template on). I added an overwrite in the component to fix it and I'll raise this to the Design System team.

Regarding the naming, we have two secondary buttons as per design one of them is meant to be quiet (as in less, but I spell that wrongly in the initial PR. Fix the spelling error now.

PR #485, opened 2018-08-15
Update tabs component

alex-ju:
Makes the redundant heading inside the panel optional and adds panel without border modifier

---

Component guide for this PR:
https://govuk-publishing-compon-pr-485.herokuapp.com/component-guide/tabs

### DIFF
--- a/app/views/govuk_publishing_components/components/_tabs.html.erb
+++ b/app/views/govuk_publishing_components/components/_tabs.html.erb
@@ -1,5 +1,9 @@
 <%
   tabs ||= []
+  panel_border ||= true unless panel_border == false
+  panel_css_classes = %w(govuk-tabs__panel)
+  panel_css_classes << "gem-c-tabs__panel--no-border" if panel_border == false
+  panel_css_classes = panel_css_classes.join(" ")
 %>
 <% if tabs.count > 1 %>
   <div class="govuk-tabs gem-c-tabs" data-module="tabs">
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/_tabs.html.erb
```diff
@@ -1,5 +1,9 @@
 <%
   tabs ||= []
+  no_panel_border ||= false
```
maxgds: I think it would be better to use "panel_border" and set it to true by default - the double negative here is a bit confusing.

app/views/govuk_publishing_components/components/_tabs.html.erb
```diff
@@ -1,5 +1,9 @@
...
+  panel_css_classes = %w(govuk-tabs__panel)
+  panel_css_classes << "gem-tabs__panel--no-border" if no_panel_border
```
maxgds: why gem-tabs rather than gem-c-tabs ?

alex-ju: To show the styles are being introduced at the gem level

maxgds: I think I need you to explain this to me in person, I thought that's what gem-c-tabs (or gem-c-*) did.

alex-ju: Ah, you're right, I'm missing the `-c-`. My bad, reading between the lines. Google spot, will update!

maxgds: 👍

alex-ju:
@maxgds updated with the fixes

app/views/govuk_publishing_components/components/_tabs.html.erb
```diff
@@ -1,5 +1,9 @@
 <%
   tabs ||= []
+  panel_border ||= true unless panel_border == false
```
maxgds: I don't think you need the "unless" clause. 
If you want the border false, you pass in panel_border = false and it takes that, the ||= just means "if nothing is passed in, set this value.

alex-ju: "if nothing is passed in" is the same with false since the boolean operator is looking for any falsy value

maxgds: Ah, right. Ok as is then.

PR #489, opened 2018-08-23
Add Google Tag Manager script as component

alex-ju:
Component to generate the Google Tag Manager script for a given `gtm_id` (mandatory), `gtm_auth` and `gtm_preview` (optional depending on the GTM setup).

This also contains [a cross browser implementation of `doNotTrack`](https://github.com/alphagov/govuk-design-system/pull/296), so users don't have to rely on obscure plugins to block tracking.

[Trello card](https://trello.com/c/wh7K3ul2)

### DIFF
--- a/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
+++ b/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
@@ -1,0 +1,14 @@
+<%
+  gtm_auth ||= nil
+  gtm_preview ||= nil
+%>
+<script>
+var doNotTrack = ( navigator.doNotTrack === '1' || navigator.doNotTrack === 'yes' || navigator.msDoNotTrack === '1' || window.doNotTrack === '1' )
+if (!doNotTrack) {
+  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
+  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
+  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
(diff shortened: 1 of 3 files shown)


alex-ju:
> Do we need anything to anonymise IP address. As per our usage with analytics: https://github.com/alphagov/govuk_frontend_toolkit_npm/blob/bd4bb837acf9a1e92698feb10d41fd2b2d813c45/javascripts/govuk/analytics/google-analytics-universal-tracker.js#L15

IP anonymise can be done via GTM without additional script added here. If this is something we want to enforce I can try to add it as part of this script.

> Secondly. I gather that this component has a required attribute of gtm_id and optional ones of gtm_auth and gtm_preview. This feels like the sort of component that people would want in some environments (prod) and not others (dev). Would they just not call the component in a dev environment or should the component just spit nothing out for missing key variables? (I'm not sure if we have any precedent for this)

I would leave the decision for each project; for example wrap the component like this, so if the ENVs are not set we don't render the component where the ENV is not defined.
```
<% if ENV["GOOGLE_TAG_MANAGER_ID"] && ENV["GOOGLE_TAG_MANAGER_AUTH"] && ENV["GOOGLE_TAG_MANAGER_PREVIEW"] %>
<%= render "govuk_publishing_components/components/google_tag_manager_script", {
      gtm_id: ENV["GOOGLE_TAG_MANAGER_ID"],
      gtm_auth: ENV["GOOGLE_TAG_MANAGER_AUTH"],
      gtm_preview: ENV["GOOGLE_TAG_MANAGER_PREVIEW"]
    } %>
<% end %>
```

> Also should we explain anywhere in here about the what the different gtm_* variables do and why you'd set some and not others? Or is that something to be explained elsewhere?

I can try and document the attributes here too.

alex-ju:
Update describing what each variable means.
Will introduce IP anonymise separately after discussing with performance analysts to see if this should be enforced.

PR #491, opened 2018-08-28
Active step by step

DilwoarH:
Spike on showing relevant step by step nav based on previous action

Part of this epic: https://trello.com/c/sPtieY54

## What
There are pages that are in more than one step by step journey. 
e.g. https://www.gov.uk/book-driving-test

For users arriving cold from Google we should show users the multiple journeys this content is part of.

But for users navigating using a specific step by step navigation e.g. Learn to drive a car. We should show that navigation expanded on the page. We should though give them a route into the alternative journey using a 'Part of' link.

## Why
We should show the relevant step by steps to users wherever possible.

## Before
screenshot-frontend dev gov uk-2018 09 03-10-42-52

## After
screenshot-frontend dev gov uk-2018 09 03-10-39-37

---

Component guide for this PR:
https://govuk-publishing-compon-pr-491.herokuapp.com/component-guide/

### DIFF
--- a/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
@@ -3,13 +3,7 @@
 <div class="gem-c-contextual-sidebar">
   <% if navigation.content_tagged_to_a_reasonable_number_of_step_by_steps? %>
     <!-- Rendering step by step related items because there are a few but not too many of them -->
-    <%
-      related_links_override = false
-      if navigation.step_nav_helper.get_active_step_nav 
-        related_links_override = [ navigation.step_nav_helper.get_active_step_nav ]
-      end
-    %>
-    <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links(related_links_override) %>
+    <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links %>
   <% end %>
   
   <% if navigation.content_tagged_to_single_step_by_step? or navigation.step_nav_helper.get_active_step_nav %>
--- a/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
@@ -9,7 +9,7 @@
     <%
       related_links_override = false
       if navigation.step_nav_helper.get_active_step_nav 
-        related_links_override = [navigation.step_nav_helper.get_active_step_nav]
+        related_links_override = [ navigation.step_nav_helper.get_active_step_nav ]
       end
     %>
     <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links(related_links_override) %>
@@ -35,7 +35,7 @@
   <% if navigation.step_nav_helper.get_active_step_nav %>
     <%= render "govuk_publishing_components/components/step_by_step_nav_related", {
       pretitle: "Also part of",
-      links: navigation.step_nav_helper.related_links
+      links: navigation.step_nav_helper.also_part_of_step_nav
     } %>
   <% end %>
 </div>
--- a/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
@@ -9,7 +9,7 @@
     <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links %>
   <% end %>
   
-  <% if navigation.content_tagged_to_single_step_by_step? or active_step_nav_content_id %>
+  <% if navigation.content_tagged_to_single_step_by_step? or navigation.step_nav_helper.get_active_step_nav %>
     <!-- Rendering step by step sidebar because there's 1 step by step list -->
     <%= render 'govuk_publishing_components/components/step_by_step_nav', navigation.step_nav_helper.sidebar %>
   <% elsif navigation.content_tagged_to_mainstream_browse_pages? %>
--- a/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
@@ -1,12 +1,15 @@
-<% navigation = GovukPublishingComponents::Presenters::ContextualNavigation.new(content_item, request.path) %>
+<%
+  active_step_nav_content_id = request.query_parameters['step-by-step-nav']
+  navigation = GovukPublishingComponents::Presenters::ContextualNavigation.new(content_item, request.path, active_step_nav_content_id) 
+%>
 
 <div class="gem-c-contextual-sidebar">
   <% if navigation.content_tagged_to_a_reasonable_number_of_step_by_steps? %>
     <!-- Rendering step by step related items because there are a few but not too many of them -->
     <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links %>
   <% end %>
-
-  <% if navigation.content_tagged_to_single_step_by_step? %>
+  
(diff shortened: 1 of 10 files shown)


lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,7 +92,12 @@ def get_list_element(style)
         style == "choice" ? "ul" : "ol"
       end
 
-      def link_href(active, href)
+      def link_href(active, href, queryString = false)
+        
+        if queryString
+          href += "?#{queryString}"
```
DilwoarH: probably need to check if there is already a query string or if it has a hash

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -57,7 +57,7 @@ def list(element)
       def create_list_item_content(link)
         if link[:href]
           @link_index += 1
-          href = link_href(link[:active], link[:href])
+          href = link_href(link[:active], link[:href], "step-by-step-nav=#{@options[:content_id]}")
```
DilwoarH: probably should change this into a object/array so that you can pass in multiple query parameters

app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
```diff
@@ -1,12 +1,15 @@
...
 
 <div class="gem-c-contextual-sidebar">
   <% if navigation.content_tagged_to_a_reasonable_number_of_step_by_steps? %>
     <!-- Rendering step by step related items because there are a few but not too many of them -->
     <%= render 'govuk_publishing_components/components/step_by_step_nav_related', links: navigation.step_nav_helper.related_links %>
   <% end %>
-
-  <% if navigation.content_tagged_to_single_step_by_step? %>
+  
+  <% if navigation.content_tagged_to_single_step_by_step? or active_step_nav_content_id %>
```
leenagupte: I think that in the "real" world, we would need to split up this condition so that if there is an `active_step_nav_content_id`, both the `step_by_step_nav_related` and `step_by_step_nav` will be rendered?

app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
```diff
@@ -1,12 +1,15 @@
-<% navigation = GovukPublishingComponents::Presenters::ContextualNavigation.new(content_item, request.path) %>
+<%
+  active_step_nav_content_id = request.query_parameters['step-by-step-nav']
+  navigation = GovukPublishingComponents::Presenters::ContextualNavigation.new(content_item, request.path, active_step_nav_content_id)
```
leenagupte: Perhaps change `GovukPublishingComponents::Presenters::ContextualNavigation` to accept the whole `request` object as a parameter, rather than adding `active_step_nav_content_id` as a new one.

app/views/govuk_publishing_components/components/_contextual_sidebar.html.erb
leenagupte: Consider adding a new method to `ContextualNavigation` called `show_expanded_step_by_step_navigation?` that does this check.  If you do that, there will be less code in the view, and the new condition will be easier to test.

lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -6,9 +6,10 @@ class ContextualNavigation
 
       # @param content_item A content item hash with strings as keys
       # @param request_path `request.path`
-      def initialize(content_item, request_path)
+      def initialize(content_item, request_path, active_step_nav_content_id = false)
         @content_item = content_item
         @request_path = simple_smart_answer? ? content_item['base_path'] : request_path
+        @active_step_nav_content_id = active_step_nav_content_id
```
leenagupte: As mentioned in the view, If you changed the definition of the constructor to:
`def initialize(content_item, request)`

Then you don't need to pass in the `active_step_nav_content_id` with a default value. Also, `false` is misleading, the default value should be `nil` as an actual value (not `true`) is being passed in.

You could use `request` to assign the rest of the values:
```
@request_path = simple_smart_answer? ? content_item['base_path'] : request.path
@active_step_nav_content_id =  request.query_parameters ? request.query_parameters['step-by-step-nav'] : nil
```

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -3,9 +3,10 @@ module Presenters
     # @private
     # Only used by the step by step component
     class PageWithStepByStepNavigation
-      def initialize(content_store_response, current_path)
+      def initialize(content_store_response, current_path, active_step_nav_content_id = false)
```
leenagupte: Again, the default value for `active_step_nav_content_id` should be `nil`

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -19,7 +20,7 @@ def show_sidebar?
       end
 
       def show_header?
-        step_navs.count == 1
+        step_navs.count == 1 or @active_step_nav_content_id
```
leenagupte: I would split this condition out, so that it's clear what's being checked and why. 

e.g. rather than `or @active_step_nav_content_id` maybe use, `or active_step_by_step_journey?`

```
def active_step_by_step_journey?
  active_step_nav_content_id.present?
end
```

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -3,9 +3,10 @@ module Presenters
     # @private
     # Only used by the step by step component
     class PageWithStepByStepNavigation
-      def initialize(content_store_response, current_path)
+      def initialize(content_store_response, current_path, active_step_nav_content_id = false)
         @content_item = content_store_response.to_h
         @current_path = current_path
+        @active_step_nav_content_id = active_step_nav_content_id
```
leenagupte: You can make @active_step_nav_id accessible by setting the `attr_reader` property.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -20,7 +20,7 @@ def show_sidebar?
       end
 
       def show_header?
-        step_navs.count == 1 or @active_step_nav_content_id
+        step_navs.count == 1 or get_active_step_nav
```
leenagupte: Consider renaming this, you don't need the `get`.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -58,16 +58,22 @@ def header
         end
       end
 
-    private
-
-      attr_reader :content_item, :current_path
-
-      def first_step_nav
+      def get_active_step_nav
```
leenagupte: This should really be two functions, one that returns the active step nav, and the other that checks if there is an active step nav.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -58,16 +58,22 @@ def header
...
         if @active_step_nav_content_id
           active_step_nav = step_navs.select { |step_nav| step_nav.content_id == @active_step_nav_content_id }
           return active_step_nav.first
+        else
+          return false
```
leenagupte: You shouldn't have to return `false` here. If there isn't an active step nav this method should will return `nil` be default, and `nil` evaluates to `false`.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -62,6 +63,11 @@ def header
       attr_reader :content_item, :current_path
 
       def first_step_nav
+        if @active_step_nav_content_id
```
leenagupte: This should `@active_step_nav_content_id.present?` as you are checking if the value exists.

leenagupte:
Superseded by #501

PR #494, opened 2018-08-31
Create "copy to clipboard" component

tijmenb:
This component will initially be used in the page preview in content publisher to allow users to copy a URL. Once we have it we can then use it to improve the [Subscription links component](https://govuk-publishing-components.herokuapp.com/component-guide/subscription-links/with_copyable_feed_link), which would benefit from the same functionality.

screen shot 2018-08-31 at 17 30 39

Preview: https://govuk-publishing-compon-pr-494.herokuapp.com/component-guide/copy_to_clipboard

https://trello.com/c/hAfZNnxt/183-add-copy-to-clipboard-component

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/copy-to-clipboard.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/copy-to-clipboard.js
@@ -11,17 +11,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       copyButton.addEventListener('click', function (event) {
         event.preventDefault()
         var input = element[0].querySelector('.gem-c-input')
-        copyToClipboard(input.value)
+        input.select()
+        document.execCommand('copy')
       })
-    }
-
-    function copyToClipboard (text) {
-      var temp = document.createElement('input')
-      temp.value = text
-      document.body.appendChild(temp)
-      temp.select()
-      document.execCommand('copy')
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/copyable-content.js
```diff
@@ -0,0 +1,25 @@
...
+        copyToClipboard(element.find("input").val())
+      })
+    }
+
+    function copyToClipboard(text) {
+      var $temp = $("<input>");
+      $("body").append($temp);
+      $temp.val(text).select();
+      document.execCommand("copy");
+      $temp.remove();
```
kevindew: I'd imagine you could achieve this with the same amount of lines without having the dependency on jQuery

tijmenb: Show me!

kevindew:
@tijmenb Here's an example without jQuery: https://gist.github.com/Chalarangelo/65320a1953388d0042829ac1fb989e33#file-copytoclipboard-js

alex-ju:
Updated the code to use plain JavaScript instead of jQuery and since we use an input component already I removed the addition and removal of an input element.

Overall looks good, the only thing that I see being different from how it was designed in the content publisher app is that an input field is being used instead of an inset text for outputting the url.

PR #502, opened 2018-09-05
Adds step by step content ID to links inside steps

DilwoarH:
# User need
As a someone trying to do a thing
I want to see the steps in the step by step journey I am on
So that I know what I need to do next

# What
If a content page is in more than one step nav, we only show a link to all of the step navs. This makes it harder for the user to navigate to the next step in their journey.

Add a parameter to the querystring when the user clicks on a link in a step nav. Then add something to the component that reads this value, so when then the content page is loaded the sidebar component expands the step by step that the user is following.

# Why
To make it easier for users to complete the journey they are on. 

# Measurement
I think we should see more linear journeys, and less clicking of the back button.

# Impact for users or other teams
There are limitations to this solution:
1. If the user comes to a content page with multiple step navs "cold" (i.e. directly from Google),  only the link to the step navs will be displayed.

2. If the user "breaks" the journey by going to a service or completing a smartanswer, then the querystring parameter will be lost, and if the done page is on muliple step navs, only the link to the step navs will be displayed.

# Applications involved

govuk_publishing_components 
government_frontend, frontend and collections to bump the gem version

---

Component guide for this PR:

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
@@ -9,6 +9,7 @@
   step_nav_url ||= false
   highlight_step ||= false
   tracking_id ||= false
+  step_nav_content_id ||= tracking_id
 
   step_count = 0
   step_number = 0
@@ -70,6 +71,7 @@
           <div class="gem-c-step-nav__panel js-panel" id="step-panel-<%= id %>-<%= step_index + 1 %>">
             <%
               in_substep = false
+              options[:step_nav_content_id] = step_nav_content_id
               options[:step_index] = step_index
               options[:link_index] = 0
             %>
--- a/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
+++ b/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
@@ -93,7 +93,23 @@ module GovukPublishingComponents
       end
 
       def link_href(active, href)
-        active ? "#content" : href
+        return "#content" if active
+        return href if external_url?(href)
+        link_with_step_nav_query_parameter(href)
+      end
+
+      def external_url?(href)
+        href.start_with?('http')
+      end
+
+      def link_with_step_nav_query_parameter(href)
+        step_nav_content_id = @options[:step_nav_content_id]
+        return href if step_nav_content_id.blank?
+        uri = URI.parse(href)
+        exisiting_query_params = uri.query.present? ? CGI.parse(uri.query) : {}
+        new_query_params = exisiting_query_params.merge("step-by-step-nav" => step_nav_content_id)
+        uri.query = new_query_params.to_query
+        uri.to_s
       end
 
       def link_text(active, text)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Show relevant step by step nav based on user journey (PR #501)
 * Add Error alert component (PR #503)
 * Add experimental panel component based on GOV.UK Frontend (PR #507)
+* Adds step by step content ID to links inside steps (PR #502)
 
 ## 9.17.1
 
--- a/spec/components/step_by_step_nav_spec.rb
+++ b/spec/components/step_by_step_nav_spec.rb
@@ -176,6 +176,18 @@ describe "step nav", type: :view do
   end
 
   it "renders links and link attributes correctly" do
+    render_component(steps: stepnav, tracking_id: "this-is-the-step-by-step-content-id")
+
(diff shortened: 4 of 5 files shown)


app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -70,6 +71,7 @@
           <div class="gem-c-step-nav__panel js-panel" id="step-panel-<%= id %>-<%= step_index + 1 %>">
             <%
               in_substep = false
+              options[:step_nav_content_id] = step_nav_content_id
```
leenagupte: Is `tracking_id` always populated? What happens if `step_nav_content_id` is false?

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,10 +92,24 @@ def get_list_element(style)
         style == "choice" ? "ul" : "ol"
       end
 
-      def link_href(active, href)
+      def link_href(active, href, parameters = {})
+        external_url = href.start_with?('http')
```
leenagupte: You could extract this out to a private function. The benefit of doing that is that you won't need the conditional check below.

e.g. the first line could be: `return href if external_url?(href)`

I'd also consider modifying the check for `active` as well.

The function could then just be three simple lines:

```
return href if external_url?(href)
return  "#content" if active
update_href_with_parameters(href, parameters)
```

`update_href_with_parameters` will only be called if the other two conditions evaluate to false.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,10 +92,24 @@ def get_list_element(style)
...
+        href = external_url ? href : update_href_with_parameters(href, parameters)
         active ? "#content" : href
       end
 
+      def update_href_with_parameters(href, parameters)
```
leenagupte: I'd consider renaming this to something a bit more descriptive for readability. Something like `add_step_nav_id_to_query_parameters` or `link_with_step_nav_query_parameter`

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,10 +92,24 @@ def get_list_element(style)
...
+        if parameters.any?
```
leenagupte: You can remove this `if/else` but just doing `return href if parameters.blank?`.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,10 +92,24 @@ def get_list_element(style)
...
+          uri = URI.parse(href)
+          query_params = uri.query.present? ? CGI.parse(uri.query) : {}
```
leenagupte: I'd rename this variable to `existing_query_params` to make it clear that these are the ones that are already in the URL.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -57,7 +57,7 @@ def list(element)
       def create_list_item_content(link)
         if link[:href]
           @link_index += 1
-          href = link_href(link[:active], link[:href])
+          href = link_href(link[:active], link[:href], "step-by-step-nav" => @options[:step_nav_content_id])
```
leenagupte: I don't think you need to add pass the query parameters in here as the step nav id is always going to added to the URL. You can do it directly in `update_href_with_parameters`.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
leenagupte: As stated above, you don't need to pass parameters to `update_href_with_parameters` as for now they are not going to change per step_nav.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -92,10 +92,24 @@ def get_list_element(style)
...
+        external_url = href.start_with?('http')
+        href = external_url ? href : update_href_with_parameters(href, parameters)
         active ? "#content" : href
       end
 
+      def update_href_with_parameters(href, parameters)
+        if parameters.any?
+          uri = URI.parse(href)
+          query_params = uri.query.present? ? CGI.parse(uri.query) : {}
+          query_params = query_params.merge(parameters)
```
leenagupte: Following on from some of the comments above, you could just merge the step_nav id here.
e.g. `query_params.merge("step-by-step-nav" => @options[:step_nav_content_id])`

DilwoarH:
hey @leenagupte - updated.

PR #507, opened 2018-09-06
Add panel component

alex-ju:
This PR adds panel component based on GOV.UK Frontend

[Trello card](https://trello.com/c/1fOAUyRc)

---

Component guide for this PR:
https://govuk-publishing-compon-pr-507.herokuapp.com/component-guide/panel

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_panel.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_panel.scss
@@ -1,1 +1,6 @@
 // This component relies on styles from GOV.UK Frontend
+
+// Specify the functions used to resolve assets paths in SCSS
+$govuk-font-url-function: "font-url";
+
+@import "../../../../node_modules/govuk-frontend/components/panel/panel";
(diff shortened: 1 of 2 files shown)


tijmenb:
Is there reason to make it admin-layout only? It would be good to use this component in email-alert-frontend, I'm worried we'll forget about that otherwise.

alex-ju:
It's dependent on `govuk-frontend` styles, that's why I added the admin layout to flag it as "needs govuk-frontend".

I would consider adding `govuk-frontend` to `email-alert-frontend` and import the needed css bit rather then duplicating the code from `govuk-frontend` to `govuk_publishing_components`

tijmenb:
Can we only add the CSS for the panel component, like we do for https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_button.scss#L14?

alex-ju:
Yes, we can. We do this for a couple of components that are needed in projects that don't have govuk-frontend yet. I will do it this way then, but long term they would probably consume it directly.

PR #512, opened 2018-09-10
Pass through label params for input

benthorner:
https://trello.com/c/l8MAts3s/205-build-the-index-page-with-filtering

This changes the input component so that any params for the label are passed through directly.


benthorner:
@maxgds good point. I was going from this PR: https://github.com/alphagov/govuk_publishing_components/pull/508.

tijmenb:
I think as long as there's a test that confirms the current behaviour (which I hope there is), it's fine to not add any for these. Needs an updated PR description though.

benthorner:
@tijmenb I've checked the tests and they cover the previous (text) behaviour.

benthorner:
@maxgds are you happy for us to continue?

maxgds:
Sure, go for it.

PR #515, opened 2018-09-11
Add file upload component

alex-ju:
This PR adds file upload component based on GOV.UK Frontend

[Trello card](https://trello.com/c/CoEtwAl4)

---

Component guide for this PR:
https://govuk-publishing-compon-pr-515.herokuapp.com/component-guide/file-upload

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_file-upload.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_file-upload.scss
@@ -1,0 +1,7 @@
+// This component relies on styles from GOV.UK Frontend
+
+// Specify the functions used to resolve assets paths in SCSS
+$govuk-font-url-function: "font-url";
+
+@import "../../../../node_modules/govuk-frontend/components/file-upload/file-upload";
+@import "../../../../node_modules/govuk-frontend/objects/form-group";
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_file-upload.scss
```diff
@@ -0,0 +1,7 @@
+// This component relies on styles from GOV.UK Frontend
+
+// Specify the functions used to resolve assets paths in SCSS
+$govuk-font-url-function: "font-url";
+
+@import "../../../../node_modules/govuk-frontend/components/file-upload/file-upload";
+@import "../../../../node_modules/govuk-frontend/objects/form-group";
```
kevindew: As discussed IRL:

I don't think we need to do imports to this degree of path traversal. As we have https://github.com/alphagov/govuk_publishing_components/blob/917d55cf0462f0f189286e34feff1eafbb6932bd/config/initializers/assets.rb#L35 I think these would work with:

```
@import "components/file-upload/file-upload";
@import "objects/form-group
```

What rails does by default is include node_modules in the assets path so it might make sense that we copy this and add `#{__dir__}/../../node_modules/` to the assets path and then we can include these files with `govuk-frontend/components/file-upload/file-upload`

alex-ju: Will raise a separate PR for this. Thanks!

PR #518, opened 2018-09-13
Show active step by step on pages where step by step is disabled

DilwoarH:
## What 
We want change how 'Hide navigation' option works.

At the moment when navigation is hidden its hidden for all users. We want to change this so that if a user uses the step by step navigation to get to the page with hidden navigation, the navigation shows.

## Why
Lab research has shown the value of showing step by step navigation throughout the journey.

Ticket: https://trello.com/c/SsoemgHH/860-related-to-step-by-step

# Before
screen shot 2018-09-13 at 15 31 57

# After
screen shot 2018-09-13 at 15 31 38

---

Component guide for this PR:
https://govuk-publishing-compon-pr-518.herokuapp.com/component-guide/

### DIFF
--- a/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
+++ b/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
@@ -15,6 +15,12 @@ module GovukPublishingComponents
         end
       end
 
+      def related_to_step_navs
+        @related_to_step_navs ||= parsed_related_to_step_navs.map do |step_nav|
+          StepByStepModel.new(step_nav)
+        end
+      end
+
       def show_sidebar?
         show_header? && current_step_nav.steps.present?
       end
@@ -24,7 +30,8 @@ module GovukPublishingComponents
       end
 
       def show_related_links?
-        step_navs.any? && (step_navs.count < 5 || active_step_by_step?)
+        return true if active_step_by_step?
+        step_navs.any? && step_navs.count < 5
       end
 
       def show_also_part_of_step_nav?
@@ -37,7 +44,8 @@ module GovukPublishingComponents
       end
 
       def also_part_of_step_nav
-        step_by_step_navs = step_navs.delete_if { |step_nav| step_nav.content_id == active_step_by_step.content_id }
+        step_navs_list = step_navs_combined_list
+        step_by_step_navs = step_navs_list.delete_if { |step_nav| step_nav.content_id == active_step_by_step.content_id }
         format_related_links(step_by_step_navs)
       end
 
@@ -67,7 +75,8 @@ module GovukPublishingComponents
       end
 
       def active_step_by_step
-        @active_step_navs ||= step_navs.select { |step_nav| step_nav.content_id == active_step_nav_content_id }
+        step_navs_list = step_navs_combined_list
+        @active_step_navs ||= step_navs_list.select { |step_nav| step_nav.content_id == active_step_nav_content_id }
         @active_step_navs.first
       end
 
@@ -88,8 +97,19 @@ module GovukPublishingComponents
         @steps ||= step_nav[:steps]
       end
 
+      def step_navs_combined_list
+        step_nav_list = []
+        step_nav_list += step_navs if step_navs.any?
+        step_nav_list += related_to_step_navs if related_to_step_navs.any?
+        step_nav_list
+      end
+
       def parsed_step_navs
         content_item.dig("links", "part_of_step_navs").to_a
+      end
+
+      def parsed_related_to_step_navs
+        content_item.dig("links", "related_to_step_navs").to_a
       end
 
       def configure_for_sidebar(step_nav_content)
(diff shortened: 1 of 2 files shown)


lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -24,7 +30,8 @@ def show_header?
       end
 
       def show_related_links?
-        step_navs.any? && (step_navs.count < 5 || active_step_by_step?)
+        return true if active_step_by_step? && (step_navs.any? || related_to_step_navs.any?)
```
leenagupte: You should only need to check for `active_step_by_step?` here. `active_step_by_step?` is already implicitly doing the `step_navs.any? || related_to_step_navs.any?` check

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -67,7 +74,8 @@ def active_step_by_step?
       end
 
       def active_step_by_step
-        @active_step_navs ||= step_navs.select { |step_nav| step_nav.content_id == active_step_nav_content_id }
+        step_navs_list = step_navs.any? ? step_navs : related_to_step_navs
```
leenagupte: You could also just add the two step_nav lists together here:
e.g. `step_navs_list = step_navs + related_to_step_navs`

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -92,6 +100,10 @@ def parsed_step_navs
         content_item.dig("links", "part_of_step_navs").to_a
       end
 
+      def parsed_related_to_step_navs
+        content_item.dig("links", "related_to_step_navs").to_a
```
leenagupte: `related_to_step_navs` is an array. You shouldn't need to convert it to an array.

leenagupte:
@DilwoarH It looks good. 👍 You definitely need tests around it though. Not just for the side bar, but for the header components too. e.g. the "also part of" is not displayed if.... etc.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: What would happen if one of them is undefined? @leenagupte

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: I copied the function above (line 99-101), should I update the above function too?

leenagupte: Yes, to keep it consistent. Thanks.

DilwoarH: Changing this causes a lot of tests to fail...

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -88,10 +97,21 @@ def steps
         @steps ||= step_nav[:steps]
       end
 
+      def step_navs_combined_list
```
leenagupte: Perhaps you could split this refactoring out into a separate commit and prove that the existing tests still pass. Then it will be easier to focus on the changes that directly implement showing step navs on pages where they've been disabled.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -24,7 +30,8 @@ def show_header?
       end
 
       def show_related_links?
-        step_navs.any? && (step_navs.count < 5 || active_step_by_step?)
+        return true if active_step_by_step?
+        step_navs.any? && step_navs.count < 5
```
leenagupte: Did method need to change? How is the logic different?

DilwoarH: this will check `active_step_by_step?` which checks related and step navs. 

if we have it the way it was before, then we'd only check if there is any step navs and not check for related to step navs

leenagupte: @DilwoarH We've talked about this in person, but at it's core the logic being used here hasn't changed. If you think it is worth the refactor though, can you separate it out from the rest of the changes? Thanks

leenagupte: Can you split this code up a bit so that it's easier to follow, and add the comment above to the commit message. Cheers.

DilwoarH:
Closing - superseded by https://github.com/alphagov/govuk_publishing_components/pull/535

PR #520, opened 2018-09-17
Add half_width flag to highlight boxes

vanitabarrett:
screen shot 2018-09-17 at 15 55 46
screen shot 2018-09-17 at 15 55 51

https://govuk-publishing-compon-pr-520.herokuapp.com/component-guide/highlight_boxes/half_width

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
@@ -26,6 +26,13 @@
   }
 }
 
+.gem-c-highlight-boxes__item-wrapper--half-width {
+  @include media(desktop) {
+    width: 50%;
+    padding: ($gutter / 6) ($gutter-half + $gutter-one-third) ($gutter-half + $gutter-one-third) 0;
+  }
+}
+
 .gem-c-highlight-boxes__item {
   @include box-sizing(border-box);
   border: 1px solid $grey-2;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
```diff
@@ -26,6 +26,13 @@
   }
 }
 
+.gem-c-highlight-boxes__item-wrapper--half-width {
+  @include media(desktop) {
+    width: 50%;
+    padding: ($gutter / 6) ($gutter-half + $gutter-one-third) ($gutter-half + $gutter-one-third) 0;
```
leenagupte: Is there a way we can test for this as I can't tell what the value for the padding is supposed to be.

vanitabarrett: Do you mean adding a test? We normally just use tests to check the presence of a class - I've added a test to check `gem-c-highlight-boxes__item-wrapper--half-width` is applied when the `half_width: true` is passed to the component.

vanitabarrett: If you mean visual check, the boxes should be 50% width and padding should be `5px 25px 25px 0`

leenagupte: If you think that you'll understand what this line does in 6 months, I'm happy to approve.

PR #523, opened 2018-09-18
Add destructive modifier to button component

benthorner:
https://govuk-publishing-compon-pr-523.herokuapp.com/component-guide/button

As this introduced another class selector for the button, in addition to
the existing 'secondary' and 'secondary-quiet' classes, I thought it
would be cleaner to refactor the decision logic to explicitly choose
which class to apply, rather than implying you can have multiple.

[Trello card](https://trello.com/c/QAEhhJdf/244-build-workflow-for-delete-draft-remove-and-retire-document-m)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add table component (PR #531)
+* Add destructive modifier to button component (PR #523)
 
 ## 9.25.0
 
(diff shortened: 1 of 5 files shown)


benthorner:
@alex-ju good points - sorry I missed the start one. And I changed 'warning' to 'danger', which I hope will do the trick ;-)

The variable is just an idea - I think it makes it clearer how the button works, but can change it back if you want.

alex-ju:
Having a variable for modifiers sounds like a good idea, it's just that I would use it more widely. Something to look at when consolidating these templates to be more consistent.

benthorner:
Thanks for spotting that @nickcolley - our design has it as a black background, so I've set that.

@alex-ju I didn't realise the start button uses a different class prefix, so I've gone back to the original behaviour to avoid having a half-refactored helper.

nickcolley:
That shadow is implemented using `box-shadow` so you might want to consider something like:

`box-shadow: 0 2px 0 $colour-here`

Instead of setting the border.

benthorner:
Thanks again @nickcolley - you may be able to tell I'm not a frontend dev ;-).

PR #524, opened 2018-09-18
[SPIKE] Remove slimmer

alex-ju:
This PR is a spike in removing slimmer dependency and replace it with a header and a footer component from the library.

While the overall outcome looks good, it reveals the dependency a lot of existing components have on external styling and resets (presumably coming from `govuk_template` via `slimmer`), which means they're not entirely isolated.

This results in some components looking slightly different than how they look when used in an application which contains `govuk_template`, but this is how they look if someone tries to use them in a new project (e.g. content-publisher) and makes it easier to correct these style leakages.

Also helps having a styled component guide for the the projects that don't use `govuk_template`.

---

Component guide for this PR:
👉https://govuk-publishing-compon-pr-524.herokuapp.com/component-guide/


kevindew:
D'oh I was experimenting with the same sort of spike as you! but looks like we want to achieve the same things :tada:

I actually had a chat with @tijmenb about this yesterday and he thought we'd need to keep slimmer because of the effects of the reset stylesheet for apps that use it.

The approach I took was to allow [disabling slimmer](https://github.com/alphagov/govuk_publishing_components/commit/dd42aeeccdbe6f9a1c5c335a3f85e6d078103d4c) and then had a [conditional in the layout to change the view](https://github.com/alphagov/govuk_publishing_components/commit/00179c6c44adb7e97d033e6bf172522895fe2042)

Hopefully we can have a chat to work out a good way forward.

kevindew:
Closing this as I've picked it up in https://github.com/alphagov/govuk_publishing_components/pull/558

PR #531, opened 2018-09-21
Add table component

alex-ju:
This PR adds a table component and makes the table helper available for projects that consume the gem.
---

Component guide for this PR:
https://govuk-publishing-compon-pr-531.herokuapp.com/component-guide/table

Example of helper usage:
https://govuk-publishing-compon-pr-531.herokuapp.com/table


spec/dummy/app/views/welcome/table.html.erb
```diff
@@ -0,0 +1,30 @@
+<%= render 'govuk_publishing_components/components/title', title: "Table helper" %>
+
+<%= GovukPublishingComponents::AppHelpers::TableHelper.helper(self, "Income tax rates") do |t| %>
+
+  <%= t.head do %>
```
nickcolley: 👀 🔍 🕵️

spec/components/all_components_spec.rb
```diff
@@ -23,7 +23,7 @@
 
       it "has the correct class in the ERB template",
         skip: component_name.in?(%w[step_by_step_nav_related step_by_step_nav_header step_by_step_nav previous_and_next_navigation]),
-        not_applicable: component_name.in?(%w[meta_tags machine_readable_metadata google_tag_manager_script layout_for_admin]) do
+        not_applicable: component_name.in?(%w[meta_tags machine_readable_metadata google_tag_manager_script layout_for_admin table]) do
```
tijmenb: Can we still add the class? It's useful for the browser extension.

alex-ju: Added the `gem-c-table` class for the extension, but it's failing because is not in the `.erb`, is in the helper

PR #540, opened 2018-09-26
Remove policies related code

AlanGabbianelli:
Remove code related to Policies

Trello card: 
https://trello.com/c/P9Fmi7wX/213-remove-now-unnecessary-code-relating-to-policies-and-policy-areas

As part of the work for the new Single Taxonomy, Policies are being
unpublished, therefore code related to Policies is not needed anymore.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,4 +1,5 @@
 {
+  "requires": true,
   "lockfileVersion": 1,
   "dependencies": {
     "govuk-frontend": {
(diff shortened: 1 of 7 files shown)


package-lock.json
```diff
@@ -1,4 +1,5 @@
 {
+  "requires": true,
```
cbaines: What does this change relate to?

nickcolley: Is likely since your team members are running different versions of Node.js and therefore npm.

We have solved this by encouraging the use of 'nvm': https://github.com/alphagov/govuk-frontend/pull/462

In this way you have a .nvmrc file in your root that indicates the version everyone should use, and then you have the same npm version.

AlanGabbianelli: Add `requires` field in `package-lock.json`

This is automatically added when running `npm istall` with newer
versions of npm (since 5.1.0).

More info: stackoverflow.com/a/45180638
Related npm PR: npm/npm#17508
Related npm Changelog: npm/npm@v5.1.0 (release)

cbaines: Ok, I've just noticed that this is described by the first commit as well, maybe I missed that when looking at the diff yesterday...

PR #541, opened 2018-09-27
Use time element in image card component

vanitabarrett:
Trello: https://trello.com/c/8UGqW1fT/184-use-time-elements-in-the-image-card-component

**This is a BREAKING CHANGE.**

This PR introduces the following changes:

- Context is now passed to the image component as a hash with a `date` and `text`, rather than a String. These are both optional.
- If a date is passed, it is rendered within a `<time>` element
- If a date and text is passed, the `<time>` element is rendered within a `<p>`, and an `aria-hidden` dash is introduced between the two.

This now matches how we render dates in other components, e.g: [document list](https://govuk-publishing-components.herokuapp.com/component-guide/document_list)

## Before
screen shot 2018-09-27 at 11 31 07

## After
screen shot 2018-09-27 at 11 31 11

Component guide for this PR:
https://govuk-publishing-compon-pr-541.herokuapp.com/component-guide/image_card

### DIFF
--- a/lib/govuk_publishing_components/presenters/image_card_helper.rb
+++ b/lib/govuk_publishing_components/presenters/image_card_helper.rb
@@ -40,7 +40,22 @@ module GovukPublishingComponents
       end
 
       def context
-        content_tag(:p, @context, class: "gem-c-image-card__context") if @context
+        return unless @context
+
+        content_tag(:p, class: 'gem-c-image-card__context') do
+          if @context[:date]
+            date = content_tag(:time, l(@context[:date], format: '%e %B %Y'), datetime: @context[:date].iso8601)
+            dash = content_tag(:span, ' — ', 'aria-hidden': true)
+
+            if @context[:text]
+              date.concat(dash).concat(@context[:text])
+            else
+              date
+            end
+          else
+            @context[:text]
+          end
+        end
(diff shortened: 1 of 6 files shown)


lib/govuk_publishing_components/presenters/image_card_helper.rb
```diff
@@ -40,7 +40,20 @@ def image
       end
 
       def context
-        content_tag(:p, @context, class: "gem-c-image-card__context") if @context
```
andrewgarner: If we supply the content as a block we can remove the need to duplicate the content tag markup.

How about something like this:

```rb
def context
  return unless @context

  content_tag(:p, class: 'gem-c-image-card__context') do
    if @context[:date]
      date = content_tag(:time, l(@context[:date], format: '%e %B %Y'), datetime: @context[:date].iso8601)
      dash = content_tag(:span, ' — ', 'aria-hidden': true)

      if @context[:text]
        date.concat(dash).concat(@context[:text])
      else
        date
      end
    else
      @context[:text]
    end
  end
end
```

vanitabarrett: @andrewgarner done 🙂

PR #543, opened 2018-09-27
Bump version to 10.0.0

AlanGabbianelli:
There are breaking changes. Please, read the CHANGELOG.md file.

---

Component guide for this PR:


tijmenb:
For future reference, this wasn't a breaking change - it just changes the behaviour of the component, which is fine as a minor version bump. `related_policies` is an internal method and not used by clients.

PR #545, opened 2018-09-27
Update to govuk-frontend 2.1.0

alex-ju:
Update govuk-frontend from 1.2.0 to 2.1.0, which means we'll have to deal with [a few breaking changes](https://github.com/alphagov/govuk-frontend/releases). Most notable changes that are affecting this component library are:
- turn off compatibility mode by default for GOV.UK Elements, GOV.UK Template, GOV.UK Frontend Toolkit
- turn on relative typography (rem) by default
- use text colour on focus for better contrast

This PR does the followings:
- creates a `govuk-frontend-settings` helper that stores the variables we need to set in order to get the appropriate SCSS from `govuk-frontend`. This means setting up SCSS functions and flags.
```
$govuk-image-url-function: "image-url";
$govuk-font-url-function: "font-url";
$govuk-compatibility-govuktemplate: true;
$govuk-compatibility-govukelements: true;
$govuk-typography-use-rem: false;
```
As you can see above we turn on compatibility mode with govuktemplate and elements and we turn of rem-based typography as it currently interferes with the existing CSS. I'd like to do a spike in the future to figure out what it takes to be able to enable this.
- updates the table component to use the new `govuk-link` styles

In addition it:
 - expose the npm package when importing JavaScript and SCSS (to make it consistent and easier to follow where the code comes from)
- adds the missing components in `all_components.scss`
- adds import for all the components that use govuk-frontend styles and keep as experimental (admin-layout-only) only the components that rely on JavaScript from `govuk-frontend` (skip link, tabs, header).

---

Component guide for this PR:
https://govuk-publishing-compon-pr-545.herokuapp.com/component-guide/


config/initializers/assets.rb
```diff
@@ -32,5 +32,5 @@
 Rails.application.config.assets.paths += %W(
   #{__dir__}/../../node_modules/govuk-frontend/assets/images
   #{__dir__}/../../node_modules/govuk-frontend/assets/fonts
```
kevindew: Do we know if we use the above two includes at all? I wonder if it'd cause pain to change them so those files are included via `govuk-frontend/assets/images/file`

app/assets/stylesheets/govuk_publishing_components/admin_styles.scss
```diff
@@ -1,8 +1,6 @@
 // Include all of the GOV.UK Frontend styles. This includes fonts, and individual components.
-$govuk-image-url-function: "image-url";
-$govuk-font-url-function: "font-url";
-
-@import "../../../node_modules/govuk-frontend/all";
```
kevindew: Yay 👍

config/initializers/assets.rb
alex-ju: > Only thing I'd suggest is updating the changelog and noting this is a breaking change for anyone including govuk-frontend through this gem

Updated the changelog, but I don't see why it would be a breaking change. It still consumes `govuk-frontend` via `govuk_publishing_components` gem it's just that now is doing it without using relative paths.

PR #552, opened 2018-10-01
Allow autofocus on inputs

alex-ju:
This attribute will be used for the login page to ease the authentication process. It shouldn't be used without user research to back this behaviour.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-552.herokuapp.com/component-guide/input/autofocused/preview

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/input.yml
+++ b/app/views/govuk_publishing_components/components/docs/input.yml
@@ -13,6 +13,8 @@ accessibility_criteria: |
   * be of the appropriate type for their use, e.g. password inputs should be of type 'password'
 
   Labels use the [label component](/component-guide/label).
+
+  Avoid using autofocus and tabindex unless you have user research to support this behaviour.
 govuk_frontend_components:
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -56,3 +56,10 @@ examples:
         text: "Search for"
       name: "name"
       value: "moose"
+  autofocused:
+    data:
```
nickcolley: totally non-blocking: you could add a 'description' to this to make a note about user research like you have done in this pull request summary

alex-ju: I'm using this to replicate the current setting of the Signon login page behaviour, which I've been told have been tested with users. I'd like to add more details, but I don't have 😞 .

nickcolley: I meant just that the description says be careful with autofocus

PR #553, opened 2018-10-01
Fix label in component guide

tijmenb:
The `label` attribute should be a hash with keys, but the component guide passes in strings.


app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -25,13 +25,15 @@
     aria_described_by = aria_described_by.join(" ")
   end
 
+
+puts label
```
nickcolley: is this intentional?

alex-ju: No, it was forgotten there. @tijmenb could you remove this please?

PR #557, opened 2018-10-02
Add the environment indicator to header

tijmenb:
This adds the environment indicator (staging, integration, development) to the header. From user research we've seen that it might prevent confusion, especially because we currently show the tag in all interfaces.

Note that we're not adding this to production, which won't have any overridden styling.

https://trello.com/c/05wMNlxg


app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -18,12 +18,19 @@
...
         </span>
+
         <% if product_name %>
         <span class="govuk-header__product-name">
           <%= product_name %>
         </span>
         <% end %>
       </a>
+
+      <% unless environment == "production" %>
```
alex-ju: Can we move this inside the anchor so Assistive Tech users know what the environment is when they get to the landmark?

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -10,3 +10,22 @@
 .gem-c-layout-header--test .govuk-header__container {
   border-bottom-color: govuk-colour("grey-1");
 }
+
+.govuk-header__logo {
+  white-space: nowrap;
+}
+
+.gem-c-environment-tag {
```
alex-ju: After this sits inside the anchor, can we have:
```
padding: 2px 5px 0;
vertical-align: middle;
```
This will make it consistent with the other GDS products

tijmenb: @alex-ju is there a way to hide this on mobile? It's too wide otherwise.

kevindew:
Aw I like the blue too.

Could we maintain the customisability of the name and whether the environment label is there in this heading, but default to "Publishing"? Reason I ask is that we've been trying to use this header for the component guide itself.

PR #560, opened 2018-10-04
Version 11.2.0

vanitabarrett:
## Version 11.2.0
- Fix toggle behaviour in related navigation and taxonomy navigation when JS disabled (PR #551)
- Make "GOV.UK Publishing" the default product name (PR #557 #561)
- Change the colour of production to red (PR #557)
- Add environment tag in the header for all environments (PR #557)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 11.2.0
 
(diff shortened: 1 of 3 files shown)


alex-ju:
@vanitabarrett is it ok if we wait for #561 to get in and then rebase? It adds missing entries to the changelog - which means this is actually 11.2.0

vanitabarrett:
@alex-ju That's fine! 🙂

PR #566, opened 2018-10-08
Update production & example environments

tijmenb:
This updates the production icon to match the red production colour. Also adds a purple "example" environment which is to be used on Heroku.


kevindew:
Might want to add a suggestion in changelog to add the heroku env var?

PR #568, opened 2018-10-08
Remove a few more Slimmer remnants

tijmenb:


tijmenb:
🙄

app/controllers/govuk_publishing_components/application_controller.rb
```diff
@@ -3,16 +3,11 @@ class ApplicationController < ActionController::Base
     helper ::Rails.application.helpers
     protect_from_forgery with: :exception
     before_action :set_x_frame_options_header
-    before_action :set_disable_slimmer_header
```
kevindew: We need this to turn off slimmer in apps that use the Gem

tijmenb: Yes, that's why I closed this PR :) I just saw "slimmer" and started pressing backspace furiously 🤕

PR #558, opened 2018-10-03
Remove slimmer

kevindew:
This PR continues the work from https://github.com/alphagov/govuk_publishing_components/pull/524 to remove Slimmer from this gem.

Removing slimmer means we no longer have [core-layout](https://github.com/alphagov/static/blob/8aaad41bde3c3380be34da8f4defcb56ed19abc1/app/assets/stylesheets/core-layout.scss) which provides various resets and defaults. This PR therefore goes through and modifies each of the components so that they no longer require this and present identically in and outside of a slimmer context.

👉 Check this out: https://govuk-publishing-compon-pr-558.herokuapp.com/component-guide/

## How to Review

I'm sorry about the size of this PR, despite the scary number of commits it's not actually got that many lines changed. That said these are some useful aspects to review.

1. Are the right GOV.UK Frontend patterns used? When I'm using `@extend` should they be `@include` instead?
2. Browse through the [component guide](https://govuk-publishing-compon-pr-558.herokuapp.com/component-guide/) and look for differences between that and that [master branch](https://govuk-publishing-components.herokuapp.com/component-guide) - ideally in a non-chrome browser and checking non-static styles like focus/hover/active
3. If you want to check whether any thing is broken using this with a slimmer enabled app you can change a Gemfile to have `gem 'govuk_publishing_components', path: "../govuk_publishing_components"` (note you can't use a github branch)

## Consequences of merging/releasing this

- Apps that don't use Slimmer, will no longer need to [remove Slimmer middleware](https://github.com/alphagov/content-publisher/blob/049a2c9847b7b7bd98275e31b63cc47dc34df990/config/application.rb#L35)
- Component guide will be usable in apps that don't have Slimmer, and lots of components will now be more consistently GOV.UK
- The default running of this Component Guide app will be without Slimmer
- Apps that do use Slimmer ~~will continue to render this guide with Slimmer~~ will no longer have the component guide rendered with Slimmer
- Numerous components have had small tweaks to fix margins/font-sizes/etc that previously depended on cascading rules into components.
---

## Components 

Links to visual diffs are done from the preview page e.g. https://govuk-publishing-components.herokuapp.com/component-guide/breadcrumbs/preview and use the [visual-regression](https://github.com/alphagov/govuk_publishing_components/blob/5514d0df735417b398763292a154f64aa3395a9a/app/assets/javascripts/component_guide/visual-regression.js) script. Less you can see in these the better and most of these are pretty much empty. They're mostly added to show they were checked.

[...]

PR #569, opened 2018-10-08
Don't include govuk-frontend/all in public frontend

tijmenb:
https://github.com/alphagov/govuk_publishing_components/pull/558 accidentally included `govuk-frontend/all` in the components library. This means that frontends include the entirety of govuk-frontend including the font (which is huge).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix an issue that included govuk-frontend with the components (#569)
+
 ## 12.0.0
 
 * Append the product name to the browser title (PR #563)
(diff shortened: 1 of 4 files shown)


kevindew:
@tijmenb Be good to grab you for a few to chat about this and https://github.com/alphagov/govuk_publishing_components/pull/570 I've got some problems using this guide consistently in all contexts.

If I understand correctly you'll always have the font output if we have any govuk-frontend components included but the things that change is whether you have the govuk-frontend helper classes output (which we need for both component guide itself and admin layout). 

I think our options are:

- Include those classes in all components. Larger all components file, but no risk that classes used don't exist and components aren't included twice
- Have a separate stylesheet for all components that is used in admin and component guide that spits out all govuk frontend components but does mean a bunch of CSS is included twice.

tijmenb:
Update: I was partially mistaken about this. I assumed that GOV.UK Frontend still used the base64 encoding and so would pull in the font entirely. 

However it uses `@font-face` now, so it only declares the `nta` font. We now think that this causes the `nta` from Static to be overwritten, so accidentally the browser will only download a single copy of the font.

alex-ju:
I think this can be closed in favour of #571

tijmenb:
Superseded by https://github.com/alphagov/govuk_publishing_components/pull/571.

PR #570, opened 2018-10-09
[DO NOT MERGE] Consistent installation approach

kevindew:
Marked DNM as it will conflict with https://github.com/alphagov/govuk_publishing_components/pull/569

To resolve the conflict with that we need to come to a resolution as to
where should inject GOV.UK Frontend helper classes and how they can
be applied consistently.

---

This commit aims to resolve some installation/usage problems that can
occur because of different ways to set-up this app.

Previously:

- Usage for an app including slimmer was to include
  govuk_publishing_components/all in application JS and SCSS
- Usage in an app using admin template was to not need to include
  govuk_publishing_components in application JS and SCSS, and start them
  automatically
- Usage of the component guide did not include
  govuk_publishing_components automatically

This meant that it was easy for an app using admin layout to have styles
and scripts but for the component guide to not.

The way I've approached solving this is to stop the automatic
installation and rely on application JS and SCSS to include
govuk_publishing_components in all cases. Hopefully this will be simpler
to understand and debug, at the cost of increased setup for someone
using admin layout.

As parts of this, I've moved the no_slimmer.js file out of
component_guide and into govuk_publishing_components since that felt a
more logical place for it. This is file auto included on admin layout
and component guide to provide a slimmer like experience. I've removed
this from the component_guide/application.js so it is distinct from that
set-up process.

I've updated the dummy files so they match the installation guide for
this gem.

I've removed the admin_styles and admin_scripts files since they no
longer serve a special usage.

---

Component guide for this PR:


tijmenb:
Superseded by https://github.com/alphagov/govuk_publishing_components/pull/571.

PR #571, opened 2018-10-09
Rework the way we include CSS & JS

tijmenb:
This is another attempt at unfuzzing the way we include CSS & Javascript. This PR reduces it down to the following:

- All applications should include the CSS themselves. Public frontend applications [already do this](https://github.com/alphagov/collections/blob/10914fb0c310a74cc53424a0e5d20c4c48df3d12/app/assets/stylesheets/application.scss#L24), but the admin apps (Signon, Content Publisher and Content Data Admin) should be updated to do this too. The admin layout component doesn't include CSS automatically anymore, instead it uses the application's stylesheet (`application.css` by convention). This means public and admin apps use the exact same mechanism to load the CSS.
- For Javascript, the same applies. Except that Slimmer provides some core dependencies that admin apps have to include for themselves.

See [this pull request on how to convert the application](https://github.com/alphagov/content-publisher/pull/373).


app/assets/javascripts/component_guide/application.js
```diff
@@ -1,3 +1,4 @@
-//= require ./no_slimmer
+//= require govuk_publishing_components/dependencies
+//= require govuk_publishing_components/all_components
```
kevindew: I don't think these includes are needed as they're expected to be covered by the javascript installation process so may be included twice.

app/assets/javascripts/govuk_publishing_components/dependencies.js
```diff
@@ -1,5 +1,5 @@
-// This adds in javascript that initialiases components and dependencies
-// that would normally be provided by slimmer
+// This adds in javascript that initialises components and dependencies
+// that are provided by Slimmer in public frontend applications.
```
kevindew: 👍

app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -1,12 +1,6 @@
 <% content_for :title, "#{@component_doc.name} component" %>
 <%= render 'govuk_publishing_components/components/title', title: @component_doc.name, context: "Component" %>
 
-<% if @component_doc.part_of_admin_layout? %>
-  <%= render "govuk_publishing_components/components/notice", title: "Admin layout only" do %>
-    <p>This component only works inside the <%= link_to "layout for admin component", "/component-guide/layout_for_admin" %>.</p>
-  <% end %>
-<% end %>
-
```
kevindew: yay 🙂

app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
```diff
@@ -9,15 +9,15 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <%= csrf_meta_tags %>
     <%= favicon_link_tag "govuk_publishing_components/favicon-#{environment}.png" %>
-    <%= stylesheet_link_tag "govuk_publishing_components/admin_styles" %>
+    <%= stylesheet_link_tag "application" %>
```
kevindew: Rather than using this by convention we could use the variables from the config: https://github.com/alphagov/govuk_publishing_components/blob/d8a98757dcacccdf95d9de48feac22d0b497cf5f/app/views/layouts/govuk_publishing_components/application.html.erb#L20-L24

Which will make it less painful if someone doesn't use the convention.

app/assets/javascripts/component_guide/application.js
alex-ju: Indeed [we mention that they should add these in their application.js](https://github.com/alphagov/govuk_publishing_components/pull/571/files#diff-e0b066a7db963c9c680eded09ad7a928R73); so these will be imported at the application/consumer level, as we do with the scss.

app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
alex-ju: We can stick to the config as @kevindew suggested.

tijmenb: I'd like to go with the convention of using `application.js`. It's one of the things we track in [the consistency spreadsheet](https://docs.google.com/spreadsheets/d/1J-0IPyr2ycPuw1IWOIoX0o9Tk9nF35D0DJdsj_ylLQE/edit#gid=0), so apps should not diverge from this intentionally.

PR #574, opened 2018-10-11
Use a normal dash for the browser title

tijmenb:
The emdash is hysterically long, which looks weird.

## Before

screen shot 2018-10-11 at 11 28 05

## After

screen shot 2018-10-11 at 11 27 56


nickcolley:
How about an endash ;) `–`

PR #577, opened 2018-10-11
Send the current user's organisation to GA

tijmenb:
This is currently done in Signon, and I'm assuming that this is fine to happen in all admin applications.

### DIFF
--- a/app/views/govuk_publishing_components/components/_admin_analytics.html.erb
+++ b/app/views/govuk_publishing_components/components/_admin_analytics.html.erb
@@ -1,3 +1,5 @@
+<% user_organisation ||= nil %>
+
 <script class="analytics">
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_admin_analytics.html.erb
```diff
@@ -6,5 +8,10 @@
 
   ga('create', 'UA-26179049-6', '<%= ENV['GOVUK_APP_DOMAIN'] %>');
   ga('set', 'anonymizeIp', true);
+
+  <% if user_organisation %>
+    ga('set', 'dimension8', "<%= user_organisation.presence || '(not set)' %>");
```
alex-ju: This is great as it's needed for https://trello.com/c/32e3xgj4. But we need to make `dimension8` customisable, as every GA dashboard will return a different dimension depending on how many are added. In the case of content-publsiher may be the only one, hence `dimension1`

PR #593, opened 2018-10-29
Updates current breadcrumbs to use design systems

DilwoarH:
Part of:

https://trello.com/c/UrJYOEVF

The breadcrumbs component in govuk_publishing_components need to be updated to use govuk-frontend code

Ticket: https://trello.com/c/PXQahpcq/357-update-breadcrumb-component

## IMPORTANT
The only difference between the current component and this update is that previously there was no underline for items which had a link to `#content` on the current page. The new design systems standards show all links with an underline styling.

Also important to notice that I've had to include media-down mixin globally as it was being used by a lot of govspeak helpers but not actually required in these files.

### Before
screenshot-govuk-publishing-components herokuapp com-2018 10 29-10-19-58

### After
screenshot-meet google com-2018 10 29-10-22-08 1

---

Component guide for this PR:
https://govuk-publishing-compon-pr-593.herokuapp.com/component-guide/

### DIFF
--- a/spec/components/breadcrumbs_spec.rb
+++ b/spec/components/breadcrumbs_spec.rb
@@ -17,7 +17,7 @@ describe "Breadcrumbs", type: :view do
   it "renders a single breadcrumb" do
     render_component(breadcrumbs: [{ title: 'Section', url: '/section' }])
 
-    assert_link_with_text_in('.gem-c-breadcrumbs__item:first-child', '/section', 'Section')
+    assert_link_with_text_in('.govuk-breadcrumbs__list-item:first-child', '/section', 'Section')
   end
 
   it "renders schema data" do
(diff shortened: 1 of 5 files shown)


spec/components/breadcrumbs_spec.rb
```diff
@@ -17,27 +17,7 @@ def assert_link_with_text_in(selector, link, text)
   it "renders a single breadcrumb" do
     render_component(breadcrumbs: [{ title: 'Section', url: '/section' }])
 
-    assert_link_with_text_in('.gem-c-breadcrumbs__item:first-child', '/section', 'Section')
-  end
-
-  it "renders schema data" do
```
alex-ju: We may need to keep this test. cc @vanitabarrett 
Is it broken in any way by the updated code?

vanitabarrett: I think it's broken by removing lines 10-12 here: https://github.com/alphagov/govuk_publishing_components/pull/593/files/1cd3097e003479de96abf524900ff8c8ab7d0aa8#diff-0fdb3d47a2d7dd6a6eeb3693cbbf865cL10

alex-ju: Thanks @vanitabarrett!
@DilwoarH I think we should keep the script; even though I agree is not good practice, but until we'll be able to chase it and find how can we improve it I would keep it to be backwards compatible.

DilwoarH: Hey @alex-ju - maybe I'm missing something but as far as I'm aware, this script was only used by the breadcrumbs, as we have deleted the code related to breadcrumbs - how would this make it backwards compatible?

DilwoarH: @alex-ju - I have added back the schema data

PR #595, opened 2018-10-30
Skip full validation of an org schema

sihugh:
This was causing downstream failures because govuk_schemas [validates random schemas that have a block applied](https://github.com/alphagov/govuk_schemas/blob/f709a10cd2af63b29c22288fc7e0ad6af426926a/lib/govuk_schemas/random_example.rb#L88).  We're only overriding simple things to assert against here, and the schema tests to validate examples are run elsewhere.

No changelog necessary for this as it affects tests only.

### DIFF
--- a/spec/lib/presenters/schema_org_spec.rb
+++ b/spec/lib/presenters/schema_org_spec.rb
@@ -407,8 +407,16 @@ RSpec.describe GovukPublishingComponents::Presenters::SchemaOrg do
     end
 
     def generate_org(details)
-      GovukSchemas::RandomExample.for_schema(frontend_schema: "organisation") do |org|
-        org.merge(details)
+      # We're skipping validation of the example schema for now because content
+      # schemas expect a lot of linked items to have base paths.  This is not the
+      # case.  It appears that the full fix is not trivial, partly because lots
+      # of tests in other apps rely on generated examples having base paths in
+      # linked items.
+      # We're making this change to unblock development, but it should be addressed
+      # soon, particularly as we can no longer rely on the below example reflecting
+      # a real content item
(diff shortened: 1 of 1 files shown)


spec/lib/presenters/schema_org_spec.rb
```diff
@@ -407,8 +407,8 @@ def generate_structured_data(args)
     end
 
     def generate_org(details)
-      GovukSchemas::RandomExample.for_schema(frontend_schema: "organisation") do |org|
-        org.merge(details)
+      GovukSchemas::RandomExample.for_schema(frontend_schema: "organisation").tap do |org|
```
tijmenb: What kind of failure did this cause? The only reason why this would fail is that the `details` are invalid, so it would be more reliable to fix that then to disable the test.

sihugh: It was for this: https://ci.integration.publishing.service.gov.uk/job/govuk_publishing_components/job/breadcrums-updated/5/console

tijmenb: It looks like it's generating invalid content here. Maybe https://github.com/alphagov/govuk-content-schemas/pull/818 didn't make the base path optional for links nested within links?

`ordered_parent_organisations` is defined with `#/definitions/frontend_links_with_base_path`, which has a `links` field with `#/definitions/frontend_links_with_base_path` again:

You could loosen up this definition to make it all valid:

https://github.com/alphagov/govuk-content-schemas/blob/50cd03e7219787963799faffd2bbd5f0f5475ec2/dist/formats/organisation/frontend/schema.json#L831-L834

sihugh:
Big old comment added 👍

PR #608, opened 2018-11-02
Add a maybe button to the feedback component

thomasleese:
We have a problem where bots are filling in the feedback component and skewing the results. The theory here is that they are clicking on the first link, so we're introducing a new hidden first link called 'Maybe'.

I haven't added support for JavaScript so if a user were to click it, it would always make a normal browser request. I figure this is fine since we're targeting bots.

Component guide for this PR:
https://govuk-publishing-compon-pr-608.herokuapp.com/component-guide/

[Trello Card](https://trello.com/c/4Yo6hC5c/592-set-up-a-honeypot-to-catch-link-clicking-bots)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+* Add a hidden 'Maybe' option to the feedback component. (PR #608)
+
 ## 12.4.0
 
 * Updates current breadcrumbs components to be based on GOV.UK Frontend (PR #593)
--- a/app/views/govuk_publishing_components/components/_feedback.html.erb
+++ b/app/views/govuk_publishing_components/components/_feedback.html.erb
@@ -13,6 +13,20 @@
       <h2 class="gem-c-feedback__prompt-question">Is this page useful?</h2>
 
       <%= link_to contact_govuk_path, {
+        class: 'gem-c-feedback__prompt-link',
+        data: {
+          'track-category' => 'yesNoFeedbackForm',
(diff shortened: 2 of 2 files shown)


stevenjmesser:
Suggest adding 

```
          'aria-controls': 'page-is-useful',
          'aria-expanded': false,
          'role': 'button'
```

to the `js-page-is-useful` class of the [Feedback component](https://github.com/alphagov/govuk_publishing_components/blob/8b575b247e1eeaa0b3bf8544492fe4303de050d2/app/views/govuk_publishing_components/components/_feedback.html.erb). @bevanloon noticed it was missing! Probably worth adding `'role': 'button'` to the Maybe option too.

thomasleese:
@stevenjmesser @bevanloon I've pushed up a commit to add those suggestions, not sure how to solve the Windows Narrator and NVDA problem (I'm surprised NVDA read it out).

thomasleese:
@bevanloon If you can that would be great, @stevenjmesser suggested that we just go ahead and merge this and fix the accessibility issues later though.

stevenjmesser:
The accessibility issues are surplus to the original card, imo, and the changes will allow the data scientists to continue with their experiment. It's possible that fixing the accessibility issues may change their results later though!

bevanloon:
I've tested this again - JAWS and NVDA were ok. Windows Narrator was the same as before. There is an accessibility issue though.
screen shot 2018-11-08 at 13 25 49

thomasleese:
Looks like we don't need `aria-controls` because the Yes/Maybe buttons don't reference a related element, in this case the form for more information.

PR #610, opened 2018-11-08
Add GA tracking to breadcrumbs

oscarwyatt:
Adds separate GA tracking data to links to the homepage for breadcrumbs

Trello: https://trello.com/c/9QaZlJC9/117-make-home-govuk-logo-or-breadcrumbs-root-and-breadcrumbs-into-2-separate-ga-events

---

Component guide for this PR:
https://govuk-publishing-compon-pr-610.herokuapp.com/component-guide/

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_header.html.erb
@@ -6,7 +6,7 @@
   <div class="govuk-header__container govuk-width-container">
 
     <div class="govuk-header__logo">
-      <a href="/" class="govuk-header__link govuk-header__link--homepage">
+      <a href="/" class="govuk-header__link govuk-header__link--homepage" data-module="track-click" data-tracking-category="homeLinkClicked">
         <span class="govuk-header__logotype">
           <svg role="presentation" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 132 97" height="32" width="36">
             <path fill="currentColor" fill-rule="evenodd"
--- a/app/views/govuk_publishing_components/components/_layout_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_header.html.erb
@@ -6,7 +6,7 @@
   <div class="govuk-header__container govuk-width-container">
 
     <div class="govuk-header__logo">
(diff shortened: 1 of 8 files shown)


app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -6,7 +6,7 @@
   <div class="govuk-header__container govuk-width-container">
 
     <div class="govuk-header__logo">
-      <a href="/" class="govuk-header__link govuk-header__link--homepage">
+      <a href="/" class="govuk-header__link govuk-header__link--homepage" data-module="track-click" data-tracking-category="homeLinkClicked">
```
karlbaker02: I think `data-tracking-category` should be `data-track-category`.

karlbaker02: Also, should we have `data-track-action="homeBreadcrumb"`?

lib/govuk_publishing_components/app_helpers/taxon_breadcrumbs.rb
```diff
@@ -15,7 +15,7 @@ def breadcrumbs
           {
             title: parent.title,
             url: parent.base_path,
-            is_page_parent: index.zero?
+            is_page_parent: index.zero?,
```
karlbaker02: Probably not necessary here but I doubt it'll break anything (unless it breaks any linter tests)

lib/govuk_publishing_components/presenters/breadcrumbs.rb
```diff
@@ -20,15 +20,68 @@ def structured_data
...
+              dimension29: crumb[:title]
           }
+        }
+
+        is_homepage = crumb[:url] == "/"
+
+        if is_homepage
+          data[:track_category] = 'homeLinkClicked'
+          data[:track_action] = 'homeBreadcrumb'
+          data[:track_label] = ''
```
karlbaker02: It's perhaps worth having this for clarity - it'll add an empty attribute of `data-track-label` to the element but this shouldn't break anything.

oscarwyatt: Haur wanted it to be a blank string so it does need to be changed...

lib/govuk_publishing_components/app_helpers/taxon_breadcrumbs.rb
oscarwyatt: Yes, it's a linting thing

PR #615, opened 2018-11-13
Support multiple errors in form fields

benthorner:
https://trello.com/c/4iKDzhBY/417-separate-saving-a-change-from-the-creation-of-a-preview

Currently we support an 'error_message' on a form field, but sometimes a
field can have multiple errors. This adds an alternative 'error_items'
option to the input, textarea, radio and file upload components to
support multiple errors as items of the error summary component.

---

Component guide for this PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add 'error_items' option for input, radio, textarea and file upload components (PR #615)
 * Update layout header to allow a full width option (PR #616)
 
 ## 12.7.1
--- a/app/views/govuk_publishing_components/components/_error_summary.html.erb
+++ b/app/views/govuk_publishing_components/components/_error_summary.html.erb
@@ -1,4 +1,5 @@
 <%
+  id ||= "error-summary-#{SecureRandom.hex(4)}"
   title ||= false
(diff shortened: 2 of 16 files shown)


.gitignore
```diff
@@ -6,6 +6,7 @@ spec/dummy/tmp/
 spec/dummy/public/assets/
 doc
 .yardoc
+yarn.lock
```
DilwoarH: do we want to ignore this file?
ref: https://github.com/yarnpkg/yarn/issues/1583

nickcolley: Our applications use the official npm client, so if you're using yarn as a developer it doesnt make sense for this to be commited

DilwoarH:
Hey @benthorner  - everything looks fine in terms of code, however, I believe the version bump usually happens in a separate PR. (https://github.com/alphagov/govuk_publishing_components/pull/614/files). I'm not sure if this is a requirement or just something that has become a convention.

benthorner:
@DilwoarH this is what I thought originally, but I've seen it done recently. Unless someone's really keen on it, can we let this one through and I'll remember to do it separately next time (and in general).

benthorner:
@DilwoarH I learnt my lesson with the version :-). How does it look now?

PR #628, opened 2018-11-20
Update back link and panel components

MuriloDalRi:
We need to display custom text on the back link component and this
changes that to allow text to be injected.
The panel component should not render a div for the description if there
isn't one, this fixes that.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-628.herokuapp.com/component-guide/

### DIFF
--- a/app/views/govuk_publishing_components/components/_back_link.html.erb
+++ b/app/views/govuk_publishing_components/components/_back_link.html.erb
@@ -1,4 +1,7 @@
-<%= link_to t('components.back_link.back'),
+<%
+  text ||= t('components.back_link.back')
+%>
+<%= link_to text,
   href,
   class: %w(
     gem-c-back-link
--- a/app/views/govuk_publishing_components/components/_panel.html.erb
+++ b/app/views/govuk_publishing_components/components/_panel.html.erb
@@ -5,9 +5,9 @@
   <h2 class="govuk-panel__title">
     <%= title %>
   </h2>
-  <div class="govuk-panel__body">
-    <% if description %>
-      <%= description %>
-    <% end %>
-  </div>
+  <% if description %>
+    <div class="govuk-panel__body">
+      <%= description %> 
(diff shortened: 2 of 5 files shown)


app/views/govuk_publishing_components/components/_back_link.html.erb
```diff
@@ -1,4 +1,7 @@
-<%= link_to t('components.back_link.back'),
+<%
+  text ||= t('components.back_link.back')
+%>
+<%= link_to text,
```
kevindew: This looks good - but it'd be worth having this as an example for the component page on https://govuk-publishing-compon-pr-628.herokuapp.com/component-guide/back_link

Here's an example of a component with different examples: https://govuk-publishing-compon-pr-628.herokuapp.com/component-guide/breadcrumbs

app/views/govuk_publishing_components/components/_panel.html.erb
```diff
@@ -5,9 +5,9 @@
...
-  <div class="govuk-panel__body">
-    <% if description %>
-      <%= description %>
-    <% end %>
-  </div>
+  <% if description %>
+    <div class="govuk-panel__body">
+      <%= description %> 
+    </div>
+  <% end %>
```
kevindew: This would also be useful to have an example.

app/views/govuk_publishing_components/components/_back_link.html.erb
MuriloDalRi: I've added an example 👍

app/views/govuk_publishing_components/components/_panel.html.erb
MuriloDalRi: Also added an example here 🌮 🎉 👍

MuriloDalRi:
@kevindew I've added examples and tidied up the commits a little bit

PR #634, opened 2018-11-21
Bundle Update

kevindew:
This updates the gems and fixes the linting issues. 

This was part of https://github.com/alphagov/govuk_publishing_components/pull/633 but I've separated it out as it got a bit big with the linting fixes required.


lib/govuk_publishing_components/presenters/taxonomy_list_helper.rb
```diff
@@ -14,7 +14,7 @@ def image_card_data
             image_src: image_card[:image][:url],
             image_alt: image_card[:image][:alt],
             heading_level: image_card[:link][:heading_level],
-            href_data_attributes: (image_card[:link][:data_attributes] if image_card[:link][:data_attributes])
+            href_data_attributes: (image_card[:link][:data_attributes])
```
cbaines: Not a big issue, but looks like the `(` `)` brackets are redundant here.

kevindew: Thanks - had missed that 👍

kevindew: Fixed in 40eeedc

kevindew:
Thanks @cbaines

PR #635, opened 2018-11-21
Add heading options to radio component

andysellick:
Adds the following options to the radio component:

- heading text (adds a legend that looks like a heading)
- is page heading (makes the heading into a H1)

With heading:

screen shot 2018-11-21 at 13 17 59

With heading as page heading:

screen shot 2018-11-21 at 13 18 05

---

Trello card: https://trello.com/c/8Z3BCpTU/79-let-radio-component-have-its-legend-as-the-page-heading-and-have-hint-text
Component guide for this PR:
https://govuk-publishing-compon-pr-635.herokuapp.com/component-guide/radio

### DIFF
--- a/app/views/govuk_publishing_components/components/_radio.html.erb
+++ b/app/views/govuk_publishing_components/components/_radio.html.erb
@@ -3,91 +3,109 @@
   items ||= []
 
   label ||= nil
+  heading ||= nil
+  is_page_heading ||= false
   hint ||= nil
   error_message ||= nil
   error_items ||= nil
+
   has_error = error_message || error_items&.any?
-  hint_id = "hint-#{SecureRandom.hex(4)}"
+  hint_id = "hint-#{SecureRandom.hex(4)}" if hint
   error_id = "error-#{SecureRandom.hex(4)}"
 
   form_group_css_classes = %w(govuk-form-group)
   form_group_css_classes << "govuk-form-group--error" if has_error
 
+  aria = "#{hint_id} #{"#{error_id}" if has_error}".strip if hint or has_error
+
   # check if any item is set as being conditional
   has_conditional = items.any? { |item| item.is_a?(Hash) && item[:conditional] }
 %>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -57,6 +57,60 @@ examples:
...
+      name: "radio-group-heading"
+      heading: "Is it snowing?"
+      is_page_heading: true
+      hint: "Sleet or hail doesn’t count."
+      items:
+      - value: "yes"
+        text: "Yes"
+      - value: "no"
+        text: "No"
   with_hint_text:
```
tombye: Would be worth renaming this as something like `with_hint_text_on_radios` to match the form of `with_hint_on_form_group`

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -3,91 +3,109 @@
...
 <%= content_tag :div, class: form_group_css_classes do %>
+  <%= tag.fieldset class: "govuk-fieldset", "aria-describedby": aria do %>
 
-  <% if hint %>
-    <%= render "govuk_publishing_components/components/hint", {
-      id: hint_id,
-      text: hint
-    } %>
-  <% end %>
+    <% if heading.present? %>
```
tombye: Bit pedantic but can the indenting match what it was? It seems to have gained two spaces which is making this diff confusing because blocks without changes are showing up as different.

andysellick: Yeah, appreciate that it's confusing. I had to add a fieldset element to wrap the radios so I think the indenting is correct, unless I've missed something? I think you can set the PR to ignore whitespace changes when reviewing changes?

tombye: Ah, yep, it's the fieldset that's doing it. Sorry, missed that. Ignore this comment.

PR #642, opened 2018-11-23
[WIP] Put nested checkboxes into lists to add support for screen readers

DilwoarH:

---

Component guide for this PR:


app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -34,30 +34,34 @@
       <%= tag.span error, id: "#{id}-error", class: "govuk-error-message" %>
     <% end %>
 
-    <%= tag.div class: "govuk-checkboxes", data: {
+    <%= tag.ul class: "govuk-checkboxes gem-c-checkboxes__list", data: {
```
andysellick: Is the class `govuk-checkboxes` still in use? Can't see it in the sass and turning it off doesn't seem to make any difference.

app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -34,30 +34,34 @@
...
-        } %>
-        <% if item[:conditional] %>
-          <%= tag.div item[:conditional], id: "#{id}-conditional-#{index}", class: "govuk-checkboxes__conditional govuk-checkboxes__conditional--hidden" %>
-        <% end %>
+        <%= tag.li class: "gem-c-checkboxes__list-item" do %>
+          <%= render 'govuk_publishing_components/components/checkbox', {
+            id: id,
+            name: name,
+            item: item,
+            index: index,
```
andysellick: Is it possible to pass data attributes to these checkboxes? And same for the nested ones.

(I realise this was already in so maybe something to consider separately)

app/views/govuk_publishing_components/components/_checkboxes.html.erb
DilwoarH: hey yeah - it's needed by the design systems https://design-system.service.gov.uk/components/checkboxes/

app/views/govuk_publishing_components/components/_checkboxes.html.erb
DilwoarH: I think it's best to consider separately, do you have a need right now?

andysellick: Not urgently, no. We can add it in when we need it.

app/views/govuk_publishing_components/components/_checkboxes.html.erb
andysellick: Interesting - might be an unused class in the Design system, same thing happens there (the class appears to have no styles associated with it and removing it has no effect). Fair enough though - best leave it.

app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
```diff
@@ -1,10 +1,28 @@
 .govuk-checkboxes--nested {
   margin-left: (govuk-spacing(3) + 3px);
-  padding-left: govuk-spacing(4);
   box-sizing: border-box;
   border-left-style: solid;
   border-left-width: 4px;
   border-color: govuk-colour('grey-2');
+  margin-top: govuk-spacing(2);
```
andysellick: Might be good for maintainability to group some of these together i.e. put all the margin declarations together.

app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -34,30 +34,34 @@
...
-        <%= render 'govuk_publishing_components/components/checkbox', {
-          id: id,
-          name: name,
-          item: item,
-          index: index,
-        } %>
-        <% if item[:conditional] %>
-          <%= tag.div item[:conditional], id: "#{id}-#{index}-conditional-#{index}", class: "govuk-checkboxes__conditional govuk-checkboxes__conditional--hidden" %>
-        <% end %>
+        <%= tag.li class: "gem-c-checkboxes__list-item" do %>
```
andysellick: Annoyingly, you could save yourself some markup and some CSS here by having the checkbox div (`govuk-checkboxes__item`) simply be the list element, but of course you can't because that's part of the component. I think this could be another thing to add to the 'things we could do to make this component slightly better if we have time' list.

app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -34,30 +34,34 @@
...
-          <%= tag.div id: "#{id}-nested-#{index}", class: "govuk-checkboxes govuk-checkboxes--nested", data: { parent: "#{id}-#{index}" } do %>
-            <% item[:items].each_with_index do |nested_item, nested_index| %>
-              <%= render 'govuk_publishing_components/components/checkbox', {
-                id: id,
-                name: name,
-                item: nested_item,
-                index: "#{index}-#{nested_index}",
-              } %>
+          <% if item[:items].present? %>
+            <%= tag.ul id: "#{id}-nested-#{index}", class: "govuk-checkboxes govuk-checkboxes--nested gem-c-checkboxes__list", data: { parent: "#{id}-#{index}" } do %>
```
andysellick: Is the ID on this UL needed? I can see the `data-parent` attribute is used by the JS, but I can't see this anywhere?

PR #643, opened 2018-11-23
Update components for finders

andysellick:
We want to use components for form fields in finders, which requires some changes to components to support this. Specifically:

- input component needs `aria-controls` attribute option
- input component needs `id` option (turns out this was already present, but not documented)
- change how `id` option works in checkbox
- add `data-attributes` option to checkbox
- add `aria-controls` option to checkbox using Javascript to apply
- add `margin-bottom` option to checkbox

---

Trello card: https://trello.com/c/fhhTwfLo/178-use-form-element-components-in-finders
Component guide for this PR:
https://govuk-publishing-compon-pr-643.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
@@ -1,3 +1,6 @@
+@import "helpers/govuk-frontend-settings";
+@import "govuk-frontend/components/checkboxes/checkboxes";
+
 .gem-c-checkbox.gem-c-checkbox--margin-bottom:last-child,
 .gem-c-checkbox.gem-c-checkbox--margin-bottom:last-of-type {
   margin-bottom: $gutter-two-thirds;
--- a/app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
@@ -1,2 +1,8 @@
-@import "helpers/govuk-frontend-settings";
-@import "govuk-frontend/components/checkboxes/checkboxes";
+.gem-c-checkbox.gem-c-checkbox--margin-bottom:last-child,
+.gem-c-checkbox.gem-c-checkbox--margin-bottom:last-of-type {
+  margin-bottom: $gutter-two-thirds;
+
+  @include media(tablet) {
+    margin-bottom: $gutter;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_checkbox.html.erb
+++ b/app/views/govuk_publishing_components/components/_checkbox.html.erb
@@ -6,7 +6,7 @@
   name = item[:name].present? ? item[:name] : name
 %>
 
-<%= tag.div class: "gem-c-checkbox govuk-checkboxes__item" do %>
+<%= tag.div class: "gem-c-checkbox govuk-checkboxes__item", data: { module: "checkbox", 'aria-controls': controls }  do %>
   <%= tag.input id: id,
     name: name,
     type: "checkbox",
@@ -15,8 +15,7 @@
     checked: item[:checked].present? ? "checked" : nil,
     data: item[:data_attributes],
     aria: {
-      describedby: item[:hint].present? ? "#{id}-item-hint" : nil,
-      controls: controls
+      describedby: item[:hint].present? ? "#{id}-item-hint" : nil
     }
   %>
   <%= tag.label item[:label], class: "govuk-label govuk-checkboxes__label", for: id %>
--- a/app/views/govuk_publishing_components/components/_checkbox.html.erb
+++ b/app/views/govuk_publishing_components/components/_checkbox.html.erb
@@ -1,22 +1,27 @@
 <%
+  index ||= false
   id ||= "checkbox-#{SecureRandom.hex(4)}"
-  index ||= rand(1..100)
+  id = "#{id}-#{index}" if index
   name = item[:name].present? ? item[:name] : name
 %>
 
 <%= tag.div class: "gem-c-checkbox govuk-checkboxes__item" do %>
-  <%= tag.input id: "#{id}-#{index}",
+  <%= tag.input id: id,
     name: name,
     type: "checkbox",
     value: item[:value],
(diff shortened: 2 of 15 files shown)


app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -30,6 +30,13 @@ examples:
         text: "What is your email address?"
       name: "address"
       type: "email"
+  with_an_identifier:
+    description: Give the input a specific ID.
+    data:
+      label:
+        text: "Has an id"
+      name: "hasid"
+      id: "hasid"
```
vanitabarrett: Should we add a test for this too?

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -3,6 +3,7 @@
   value ||= nil
   type ||= "text"
   describedby ||= false
+  controls ||= nil
```
vanitabarrett: Should we be consistent with `controls` and `describedby`? (set either to `false` or `nil`)

app/views/govuk_publishing_components/components/docs/checkbox.yml
```diff
@@ -35,3 +35,30 @@ examples:
...
+      If an index is given, this is appended onto the end of the ID (given or generated). This is used in the checkboxes component where multiple checkboxes are output.
+    data:
+      name: "with_id"
+      id: "moose"
+      index: 10
+      item:
+        label: "With an id of 'moose-10'"
+        value: "with_id"
+  checkbox_with_data_attributes:
+    description: Data attributes such as tracking can be applied if required. Note that for a checkbox this will not fire tracking events automatically, additional work will be required when the component is used.
```
vanitabarrett: Can you expand on what additional work is required?

app/views/govuk_publishing_components/components/docs/checkbox.yml
```diff
@@ -35,3 +35,30 @@ examples:
...
+    data:
+      name: "With tracking"
+      item:
+        label: "Tracked"
+        value: "tracked"
+        data_attributes: {
+          track_category: "checkboxClicked",
```
vanitabarrett: Should we be including `track-action` as well?

app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
```diff
@@ -1,2 +1,8 @@
-@import "helpers/govuk-frontend-settings";
-@import "govuk-frontend/components/checkboxes/checkboxes";
```
DilwoarH: This probably needs to be added in as the component might be stand alone

andysellick: Will that duplicate anything @DilwoarH ?

app/views/govuk_publishing_components/components/_checkbox.html.erb
```diff
@@ -16,7 +16,8 @@
     aria: {
       describedby: item[:hint].present? ? "#{id}-item-hint" : nil,
       controls: item[:conditional].present? ? "#{id}-conditional-#{index}" : nil
-  } %>
+    }
+  %>
```
andysellick: This was a more impressive change before someone fixed it ahead of me and I had to rebase.

app/views/govuk_publishing_components/components/_checkbox.html.erb
```diff
@@ -1,22 +1,26 @@
 <%
+  index ||= false
   id ||= "checkbox-#{SecureRandom.hex(4)}"
-  index ||= rand(1..100)
+  id = "#{id}-#{index}" if index
```
vanitabarrett: I'm not sure where the checkbox component is currently used - is this change (passed id's don't have a random number assigned to them any more) likely to break anything anywhere?

app/views/govuk_publishing_components/components/_checkbox.html.erb
```diff
@@ -1,22 +1,26 @@
...
   } %>
-  <%= tag.label item[:label], class: "govuk-label govuk-checkboxes__label", for: "#{id}-#{index}" %>
+  <%= tag.label item[:label], class: "govuk-label govuk-checkboxes__label", for: id %>
   <% if item[:hint].present? %>
-    <%= tag.span item[:hint], id: "#{id}-#{index}-item-hint", class: "govuk-hint govuk-checkboxes__hint" %>
+    <%= tag.span item[:hint], id: "#{id}-item-hint", class: "govuk-hint govuk-checkboxes__hint" %>
   <% end %>
 <% end %>
+
+
```
vanitabarrett: Nitpick: Stray new line

app/views/govuk_publishing_components/components/_checkbox.html.erb
vanitabarrett: Might be worth moving this id change out of the commit which adds the data attributes in case, for any reason, we need to revert one but not both of these changes in the future.

andysellick: I checked with @DilwoarH and this should be okay. The index is added for when the component is used in the checkboxes component, which passes both id and index to the checkbox component.

app/assets/stylesheets/govuk_publishing_components/components/_checkbox.scss
andysellick: My understanding is that when you use the gem in an app it pulls in all of the CSS from the gem, not just from the components that are used. Therefore as long as these imports are included somewhere, they'll get in. I think that perhaps (unless duplication doesn't occur) they should be moved from the `checkboxes` component into the `checkbox` component, instead.

andysellick: @DilwoarH have checked with @alex-ju and apparently if we include these lines twice there _will_ be duplication, which I want to avoid. I would suggest therefore that these lines go in only the checkbox component, since it's the 'original' component and checkboxes is using it. I'll add a separate commit to do this.

andysellick:
@vanitabarrett are you happy to approve this PR?

PR #650, opened 2018-11-29
Add data attributes to select component

andysellick:
Adds the ability to pass arbitrary data attributes to the select component, that will be applied to the individual options.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-650.herokuapp.com/component-guide/select
Trello card: https://trello.com/c/CF6DmPXk/137-add-sorting-to-the-news-and-communications-finder

### DIFF
--- a/app/views/govuk_publishing_components/components/_select.html.erb
+++ b/app/views/govuk_publishing_components/components/_select.html.erb
@@ -2,6 +2,8 @@
   options ||= []
   id ||= false
   label ||= false
+
+  select_helper = GovukPublishingComponents::Presenters::SelectHelper.new(options)
 %>
 <% if options.any? && id && label %>
   <div class="govuk-form-group gem-c-select">
@@ -9,9 +11,7 @@
       <%= label %>
     </label>
     <select class="govuk-select" id="<%= id %>" name="<%= id %>">
-      <% options.each do |option| %>
-        <option value="<%= option[:value] %>" <% if option[:selected] %>selected<% end %>><%= option[:text] %></option>
-      <% end %>
+      <%= options_for_select(select_helper.option_markup, select_helper.selected_option) %>
     </select>
   </div>
 <% end %>
--- a/lib/govuk_publishing_components/presenters/select.rb
+++ b/lib/govuk_publishing_components/presenters/select.rb
@@ -1,0 +1,38 @@
+module GovukPublishingComponents
+  module Presenters
+    class SelectHelper
+      attr_reader :options, :option_markup, :selected_option
+
+      def initialize(options)
+        @options = options
+        @option_markup = get_options
+      end
+
+    private
+
+      def get_options
+        return if options.nil?
+
+        options.map { |option|
+          @selected_option = option[:value] if option[:selected]
+          [
+            option[:text],
+            option[:value],
+            options_data_attribute(option[:data_attributes])
+          ]
+        }
+      end
+
+      def options_data_attribute(attributes)
+        return {} if attributes.nil?
+
+        attrs = {}
+        attributes.each do |key, value|
+          key_name = "data-#{key.to_s.split('_').join('-')}"
+          attrs[key_name] = value.is_a?(Hash) ? value.to_json : value
+        end
+        attrs
+      end
+    end
+  end
+end
--- a/app/views/govuk_publishing_components/components/docs/select.yml
+++ b/app/views/govuk_publishing_components/components/docs/select.yml
@@ -22,7 +22,7 @@ examples:
   with_preselect:
     data:
       id: 'dropdown2'
-      label: 'My Dropdown'
+      label: 'Option 2 preselected'
       options:
       - text: 'Option one'
         value: 'option1'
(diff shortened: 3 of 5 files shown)


lib/govuk_publishing_components/presenters/select.rb
```diff
@@ -0,0 +1,42 @@
+module GovukPublishingComponents
+  module Presenters
+    class SelectHelper
+      attr_reader :options, :option_markup, :selected_option
+
+      def initialize(local_assigns)
+        @options = local_assigns[:options] if local_assigns[:options]
```
bilbof: `if local_assigns[:options]` isn't required here. It's better that `options` is nil rather than being undefined.

lib/govuk_publishing_components/presenters/select.rb
```diff
@@ -0,0 +1,42 @@
...
+        @option_markup = get_options
+      end
+
+    private
+
+      def get_options
+        return {} if @options.nil?
```
bilbof: We probably want to return an array here if that's what `formatted_options` is returned as.

We don't need to put `@` here actually as the `attr_reader` is used above.

lib/govuk_publishing_components/presenters/select.rb
```diff
@@ -0,0 +1,42 @@
...
+        options.each do |option|
+          formatted_options << [
+            option[:text],
+            option[:value],
+            options_data_attribute(option[:data_attributes])
+          ]
+          @selected_option = option[:value] if option[:selected]
+        end
+
+        formatted_options
```
bilbof: If `options` is an array as [in the spec](https://github.com/alphagov/govuk_publishing_components/pull/650/files#diff-4d49ce1d3dc073aa8f98fb53fd8368dcR111), then we can use `map` and have a second method for the `find`. It's two operations but a little clearer.


```
def get_options
    return [] unless options.present?

    options.map { |option| {
        [ option[:text], option[:value], options_data_attribute(option[:data_attributes]) ]
    }
end

def selected_option
    return unless options.present?

    @selected_option ||= options.find { |option| option[:selected] }
end
```

lib/govuk_publishing_components/presenters/select.rb
```diff
@@ -0,0 +1,42 @@
...
+      end
+
+      def options_data_attribute(attributes)
+        return {} if attributes.nil?
+
+        attrs = {}
+        attributes.each do |key, value|
+          key_name = "data-#{key.to_s.split('_').join('-')}"
```
vanitabarrett: Is this needed? This method looks like it could be simplified into returning `{} if attributes.nil?` or just `attributes` if not.

andysellick: If you're referring to line 35's `to_s.split('_')` etc. then the reason for this is because data attributes are passed from the YML as keys with underscores, but data attributes need hyphens. Does that make sense or have I misunderstood your comment?

lib/govuk_publishing_components/presenters/select.rb
andysellick: My Ruby knowledge isn't up to knowing which is better, but I'm slightly hesitant to do the loop twice if we don't need to. Let's catch up about this.

lib/govuk_publishing_components/presenters/select.rb
andysellick: I think the difference is on those other components we're using the rails tag generation stuff, which takes a data object and does the conversion. In this case we're using [options_for_select](https://apidock.com/rails/ActionView/Helpers/FormOptionsHelper/options_for_select), which appears to work slightly differently. Here's an example from the page:

```
options_for_select([["Dollar", "$", {class: "bold"}], ["Kroner", "DKK", {onclick: "alert('HI');"}]])
```

Firstly, if we just returned the attributes it doesn't put `data-` on the front of them, so we need to append that. And then unless we do the string replace `data_something` comes out as an attribute on the element called `data-something` - but if I pass through `data_something_else` it would come out as `data-something_else` which I don't think is right. 

Kind of frustrating that this rails helper doesn't seem to quite work the same as others, but I guess it's a bit unusual to want to put data attributes on an `option` element?

app/views/govuk_publishing_components/components/_select.html.erb
```diff
@@ -2,16 +2,16 @@
   options ||= []
   id ||= false
   label ||= false
+
+  select_helper = GovukPublishingComponents::Presenters::SelectHelper.new(local_assigns)
```
vanitabarrett: Are we able to just pass options? I don't think anything else (id and label) is needed.

lib/govuk_publishing_components/presenters/select.rb
bilbof: In that case, you could add `@selected_option = option if option[:selected]` as the first line in the map block rather than using find.

andysellick: Updated, thanks.

bilbof:
:shipit:  🎉

PR #655, opened 2018-12-03
Modify warning text component

andysellick:
Modify the warning text component so that it:

- doesn't have to have an icon
- has an option to have larger text
- has an option to have highlighted (i.e. red) text

---

Component guide for this PR:
https://govuk-publishing-compon-pr-655.herokuapp.com/component-guide/warning_text

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/warning_text.yml
+++ b/app/views/govuk_publishing_components/components/docs/warning_text.yml
@@ -21,3 +21,7 @@ examples:
     data:
       text_icon: ""
       text: "You can be fined up to £320 if you don't register."
+  with_larger_font:
+    data:
+      text: "This content was updated on the 24th of November 2018"
+      large_font: true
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/warning_text.yml
```diff
@@ -17,3 +17,7 @@ examples:
     data:
       text_icon: "£"
       text: "You can be fined up to £5,000 if you don’t register."
+  without_icon:
+    data:
+      text_icon: ""
```
vanitabarrett: I wonder if it would make more sense to be passing `text_icon: false`?

andysellick: I tried that originally, but I wanted to leave the default behaviour as it is, which is if nothing is passed the icon is shown as `!`. So if you passed false, `text_icon ||= '!'` defaults back to `!`. This seemed the cleanest way of doing it without having to update any apps.

vanitabarrett: Chatted on Slack and decided this may be the safest option in case people have been passing in `text_icon: false` and expecting to get the default `!`

PR #664, opened 2018-12-12
Make layout footer support spawning new windows for links

kgarwood:
https://trello.com/c/TiZgaPxw/537-open-support-links-in-a-new-tab

Enhancing the layout footer component so that it can make pressing a link open a new window
or change the page of the current window.
- Making use of link_to in _layout_footer.html.erb to handle the target attribute and
  explicitly setting target to '' if no target attribute is passed to the component.
- Adding a new example to layout_footer.yml so users can see examples of using the
  layout footer component to open links in either the same or a new window.

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_footer.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_footer.html.erb
@@ -18,15 +18,11 @@
                 <% item[:items].each do |item| %>
                   <% if item[:href] && item[:text] %>
                     <li class="govuk-footer__list-item">
-                      <% if item[:target].nil? %>
-                        <%= link_to item[:text], item[:href], target: '',
-                          class: "govuk-footer__link"
-                        %>
-                      <% else %>
-                        <%= link_to item[:text], item[:href], target: item[:target],
-                          class: "govuk-footer__link"
-                        %>
-                      <% end %>
+                      <%
+                        attributes = { class: "govuk-footer__link" }.merge(item.fetch(:attributes, {}))
+                        attributes[:rel] = "noopener" if attributes[:target] == "_blank" && !attributes[:rel]
+                      %>
+                      <%= link_to item[:text], item[:href], attributes %>
                     </li>
                   <% end %>
                 <% end %>
--- a/app/views/govuk_publishing_components/components/_layout_footer.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_footer.html.erb
@@ -18,9 +18,15 @@
                 <% item[:items].each do |item| %>
                   <% if item[:href] && item[:text] %>
                     <li class="govuk-footer__list-item">
-                      <a class="govuk-footer__link" href="<%= item[:href] %>">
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -18,9 +18,15 @@
...
                     <li class="govuk-footer__list-item">
-                      <a class="govuk-footer__link" href="<%= item[:href] %>">
-                        <%= item[:text] %>
-                      </a>
+                      <% if item[:target].nil? %>
+                        <%= link_to item[:text], item[:href], target: '',
+                          class: "govuk-footer__link"
+                        %>
+                      <% else %>
+                        <%= link_to item[:text], item[:href], target: item[:target],
```
nickcolley: There are some security considerations to make with opening links in new windows: https://developers.google.com/web/tools/lighthouse/audits/noopener

Additionally we often avoid opening links in new windows because of accessibility issues associated with doing so, but I'm not sure what the user need is.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -18,9 +18,15 @@
                 <% item[:items].each do |item| %>
                   <% if item[:href] && item[:text] %>
                     <li class="govuk-footer__list-item">
-                      <a class="govuk-footer__link" href="<%= item[:href] %>">
-                        <%= item[:text] %>
-                      </a>
+                      <% if item[:target].nil? %>
```
alex-ju: I think these 9 lines can be simplified as:
`<%= link_to item[:text], item[:href], target: item[:target], class: "govuk-footer__link" %>`

app/views/govuk_publishing_components/components/docs/layout_footer.yml
```diff
@@ -123,3 +123,15 @@ examples:
             text: Rhestr o Wasanaethau Cymraeg
           - href: '/government/organisations/government-digital-service'
             text: Government Digital Service
+
+  with_new_windows:
```
alex-ju: can we name this `whith_target_on_links` to make it more clear?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
alex-ju: I agree with @nickcolley, if we're going to allow target `_blank`s, we should at least add `rel="noopener"`.

app/views/govuk_publishing_components/components/docs/layout_footer.yml
kgarwood: Done

kgarwood:
I've generalised the component to accept a collection of anchor attributes rather than just focus on target.  I've also ensured that if target="_blank" then it'll set rel="noopener".

kevindew:
@kgarwood these commits seem to be associated with your VM's email address

PR #670, opened 2018-12-17
Add optional maxlength to input fields

leenagupte:
Trello: https://trello.com/c/hGmJ5BfV

## Motivation

Content Publisher allows publishers to add an image credit. The intention is to limit the number of characters publishers can use in the image credit.

Sets a maxlength attribute in a similar way to textarea.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-670.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 - BREAKING: Merge checkbox components (PR #659)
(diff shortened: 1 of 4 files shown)


leenagupte:
@andysellick Good to know that I'm on the right track though!

PR #671, opened 2018-12-18
Fix autocomplete styles

andysellick:
Arrow was not appearing on autocomplete, even though we're using the `showAllValues` option - see [autocomplete examples](https://alphagov.github.io/accessible-autocomplete/examples/).

This change fixes this by increasing the z-index on both the arrow svg and the input, so the arrow is visible but does not overlap the click area of the input.

Fixes this:

screen shot 2018-12-18 at 12 23 51

...to this:

screen shot 2018-12-18 at 12 23 57

---

Component guide for this PR:
https://govuk-publishing-compon-pr-671.herokuapp.com/component-guide/accessible_autocomplete

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,10 +9,12 @@
 
 ## Unreleased
 
-- BREAKING: Merge checkbox components (PR #659)
+* Fix autocomplete styles (PR #671)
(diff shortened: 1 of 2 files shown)


andysellick:
@alex-ju good question. I'll look into it.

PR #680, opened 2018-12-19
Put margin bottom on autocomplete

andysellick:
Wrap the accessible autocomplete in a `govuk-form-group` in order to give it some default bottom margin consistent with the other form components (for example input, which it visually closely resembles).

---

Component guide for this PR:


vanitabarrett:
Have we got some examples of how this change will look on GOV.UK pages so we can see it in context? Like the finders (or perhaps somewhere else where the autocomplete is used so we know it looks ok).

andysellick:
@vanitabarrett so far this component is only used in the finders, and this change was motivated by the spacing not being right there - we had some custom CSS in finder-frontend to provide margin. It occurred to me that since this component had no bottom margin, it will likely always need some custom CSS to provide that, which is against the whole point of components.

I considered leaving no margin as the default and adding an option, but we're only using it in finders so far and all the rest of the form components have margin bottom by default. This just brings this one in line.

PR #683, opened 2018-12-20
Move all commonly used imports into _all_components.scss

andysellick:
This is an experimental PR to move duplicate import lines such as `@import "components/helpers/govuk-frontend-settings";` out of individual SASS files and into the main application.scss file.

Advantages:

- no duplication
- easier to see in one place all of the things we import
- might compile fractionally quicker?

Disadvantages:

- doesn't impact overall compiled CSS size (actually it seems to have reduced it by 1KB, but that's probably immaterial as that saving is likely already achieved through gzipping)
- can't easily see which files use which imports, so if we stop using an import we might not know to remove it from application.scss

Would be good to get some feedback on whether people think this worth doing or not.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-683.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Move all commonly used imports into all_components.scss (PR #683)
+
 ## 13.5.0
 
 * Allow heading on single checkbox (PR #686)
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -1,18 +1,27 @@
-// Include all of the GOV.UK Frontend styles. This includes fonts, and individual components.
-@import "components/helpers/govuk-frontend-settings";
-
 // This is the file that the application needs to include in order to use
 // the components.
+
+// Include all of the GOV.UK Frontend styles. This includes fonts, and individual components.
+
+@import "components/helpers/govuk-frontend-settings";
+
+// Include common imports used by many components
+@import "govuk-frontend/all";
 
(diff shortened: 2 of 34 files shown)


app/assets/stylesheets/govuk_publishing_components/_all_components.scss
```diff
@@ -1,18 +1,34 @@
...
+// the components.
+
 // Include all of the GOV.UK Frontend styles. This includes fonts, and individual components.
+
 @import "components/helpers/govuk-frontend-settings";
 
-// This is the file that the application needs to include in order to use
-// the components.
+// Include common imports used by many components
+@import "govuk-frontend/all";
```
alex-ju: `/all` will suffice, as it include all the other imports listed below

alex-ju: If we remove the redundant `govuk-frontend` imports below I'm happy to get this merged as we're not able to consume the components individually anyway at this point.

andysellick: @alex-ju have done so, can you check? Seems to be alright but I want to be sure (breaking _all_ of the components just before Christmas wouldn't be good).

alex-ju: sure, will test it now

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Move all commonly used imports into application.scss (PR #683)
```
alex-ju: Shouldn't it be _... into`_all_components.scss`_?

andysellick: Oh yeah, good point. Will fix.

PR #684, opened 2018-12-20
Change heading/legend/hint options on checkboxes component

andysellick:
This PR changes the checkboxes component as follows:

- makes the heading (legend) text option mandatory, without which the component will fail to render, as this is an accessibility violation
- provides an option to make the heading (legend) visually hidden if not required
- provides an option to make hint text optional

---

Component guide for this PR:
https://govuk-publishing-compon-pr-684.herokuapp.com/component-guide/checkboxes

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -71,6 +71,17 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  without_hint_text:
+    description: Hint text defaults to 'Select all that apply'. This can be removed entirely with this option.
+    data:
+      name: "favourite_skittle"
+      heading: "What is your favourite skittle?"
+      no_hint_text: true
+      items:
+        - label: "Mauve"
+          value: "mauve"
+        - label: "Sunset orange"
+          value: "sunsetorange"
   with_legend_as_page_heading:
     description: Since the legend/heading is required, if the checkboxes are alone on a page it makes sense to use this element as the H1 on the page rather than duplicate text.
     data:
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -35,6 +35,19 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  with_a_hidden_heading:
+    description: If the heading/legend on the checkboxes is not required, it can be visually hidden using this option. It will still be visible to screen readers.
+    data:
+      name: "favourite_colour"
+      heading: "What is your favourite colour?"
+      visually_hide_heading: true
+      items:
+        - label: "Red"
+          value: "red"
+        - label: "Green"
+          value: "green"
+        - label: "Blue"
+          value: "blue"
   with_a_custom_id_attribute:
     description: Note that if an id is not given one is generated automatically. In either case, the id is applied to the parent element of the checkboxes, and each checkbox is given the same id with an incremented number at the end, e.g. the checkboxes below have ids of potatoes-0 and potatoes-1.
     data:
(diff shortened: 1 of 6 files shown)


lib/govuk_publishing_components/presenters/checkboxes_helper.rb
```diff
@@ -20,10 +20,13 @@ def initialize(options)
         @id = options[:id] || "checkboxes-#{SecureRandom.hex(4)}"
         @heading = options[:heading]
         @is_page_heading = options[:is_page_heading]
-        @hint_text = options[:hint_text] || "Select all that apply."
+        @no_hint_text = options[:no_hint_text]
+        @hint_text = options[:hint_text] || "Select all that apply." unless @no_hint_text
       end
 
       def fieldset_describedby
+        return if @no_hint_text
```
vanitabarrett: I think this could be rewritten as:

```
unless @no_hint_text
   text = %w()
   text << "#{id}-hint" if @hint_text
   text << "#{id}-error" if @error
end
```

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
```diff
@@ -4,7 +4,7 @@ class CheckboxesHelper
       include ActionView::Helpers
       include ActionView::Context
 
-      attr_reader :items, :name, :css_classes, :error, :has_conditional, :has_nested, :id, :hint_text
+      attr_reader :items, :name, :css_classes, :error, :has_conditional, :has_nested, :id, :hint_text, :no_hint_text
```
vanitabarrett: Is `no_hint_text` used outside this class? If not, I don't think this is needed.

spec/components/checkboxes_spec.rb
```diff
@@ -88,6 +88,19 @@ def component_name
...
+    render_component(
+      name: "favourite_colour",
+      no_hint_text: true,
+      items: [
+        { label: "Red", value: "red" },
+        { label: "Green", value: "green" },
+      ]
+    )
+    assert_select ".govuk-hint", false
+    assert_select ".govuk-fieldset[aria-describedby]", false
```
vanitabarrett: This should possibly be a separate test, but as long as there's a test covering this situation it probably doesn't matter too much!

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -72,10 +85,11 @@ examples:
         - label: "Blue"
           value: "blue"
   without_hint_text:
-    description: Hint text defaults to 'Select all that apply'. This can be removed entirely with this option.
+    description: Hint text defaults to 'Select all that apply'. This can be removed entirely with this option. Note that this option can be combined with the visually_hide_heading option, as shown.
```
vanitabarrett: The first sentence (`Hint text defaults to 'Select all that apply'.`) initially confused me - I thought it meant that in this example, the hint text defaulted to "Select all that apply". We could move this to the `with_custom_hint_text` example instead.

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -71,6 +71,17 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  without_hint_text:
+    description: Hint text defaults to 'Select all that apply'. This can be removed entirely with this option.
```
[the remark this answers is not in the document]
andysellick: We don't mention that it's added in the first place, do we? If not I don't think it's worth explaining.

spec/components/checkboxes_spec.rb
andysellick: I think the two are linked... actually we should probably test from the other side as well i.e. that describedby is set correctly when hint text isn't hidden.

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
vanitabarrett: @andysellick Sorry, my mistake for forgetting this when I copy/paste-d: you still need to return `text` before the `end`.

PR #686, opened 2018-12-20
Allow heading on single checkbox

andysellick:
- there is a requirement to have a heading/legend on a single checkbox, so if a heading is passed, render it

---

Component guide for this PR:
https://govuk-publishing-compon-pr-686.herokuapp.com/component-guide/checkboxes

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -11,11 +11,8 @@ accessibility_criteria: |
   - be usable with touch
   - indicate when they have focus
   - have correctly associated labels
-  - have a legend (see below)
 
   Labels use the [label component](/component-guide/label).
-
-  Checkboxes in a fieldset require a legend. The text of the legend is passed using the 'heading' option. If no heading is passed, no legend is shown, which causes an accessibility error. You must pass a heading when using this component, or provide other measures to ensure accessibility.
 examples:
   default:
     data:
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/checkboxes_helper.rb
```diff
@@ -25,6 +25,10 @@ def initialize(options)
         @visually_hide_heading = options[:visually_hide_heading]
       end
 
+      def should_have_fieldset
+        true if @items.length > 1 || @heading
```
vanitabarrett: I think we can remove `true if`

spec/components/checkboxes_spec.rb
```diff
@@ -134,14 +145,13 @@ def component_name
...
       name: "favourite_colour",
-      heading: "What is your favourite colour?",
-      hint_text: "Choose carefully",
       items: [
         { label: "Red", value: "red" },
       ]
     )
     assert_select ".govuk-hint", false
     assert_select ".govuk-fieldset__legend", false
+    assert_select ".govuk-hint", false
```
vanitabarrett: Isn't this duplicating line 152?

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -35,6 +35,38 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  with_a_heading_on_one_checkbox:
+    description: One checkbox does not require a fieldset and therefore does not require a legend. If a heading is supplied, this will be shown with one checkbox.
```
karlbaker02: When I check on the component guide, this particular example is still rendered with a fieldset. Should this be removed in order to make it consistent with the description?

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
karlbaker02: In relation to my other comment, perhaps this should be:

`true if @items.length > 1`

vanitabarrett: I don't think that would work as we don't want items with, for example, 5 checkboxes but no heading to fall through to the `else` condition?

app/views/govuk_publishing_components/components/docs/checkboxes.yml
andysellick: Sorry, I think this is to do with the description here rather than the behaviour of the component. Will re-word.

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -35,6 +35,38 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  with_a_heading_on_one_checkbox:
+    description: One checkbox does not require a fieldset and therefore does not require a legend. However, if a heading is supplied, a fieldset will be included in the component and the heading used as the legend, as shown.
```
karlbaker02: This is still a bit contradictory to the usage of the checkbox from above: _You must pass a heading when using this component_

If that's the case then the case of one checkbox will always be rendered with a fieldset and legend?

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
andysellick: Yeah, this is really complex. The logic is:

- if one checkbox, don't render a fieldset and a legend
- if one checkbox and a heading is supplied, render a fieldset and a legend (using the heading)
- if more than one checkbox, and no heading, fail

Tried a few options for this but so far nothing works. I've updated the helper method with a little comment to explain. If I think of a better way of doing this in future I'll definitely change it but for now I think it's okay.

app/views/govuk_publishing_components/components/docs/checkboxes.yml
andysellick: Good point. I've updated the documentation by removing this bit - I think the examples explain themselves better now anyway.

PR #688, opened 2018-12-20
Org schemas: prefer description over body

sihugh:
They contain the same text, but the body contains markup which is possibly best to avoid for schema text.

The description is not always populated at present, so we still allow fallback.

Related to https://github.com/alphagov/whitehall/pull/4569


sihugh:
This can be deployed whenever

PR #689, opened 2018-12-20
Tweak "about" in Article schema

sihugh:
A topic is more of a "Thing" than it is a "CreativeWork". I'm not sure what difference this will actually make, but it seems conceptually better to me.

I think this is small enough to not need an entry in the change log


sihugh:
This can be merged whenever anyone fancies giving it a 👍

PR #691, opened 2018-12-31
allow navigation items to only show in collapsed menu

DilwoarH:
Needed for https://github.com/alphagov/content-publisher/pull/647

## Why this is needed

On the content publisher app there are two navigations, primary (layout_header component) and secondary (phase banner component).

On mobile view the phase banner hides the navigation items, however, these still need to be accessible for mobile users. Therefore, we need to add them to the primary navigation bar and hide them unless it's on mobile.

### Desktop
screen shot 2019-01-02 at 16 46 55

### Mobile
screen shot 2019-01-02 at 16 39 52

---

Component guide for this PR:
https://govuk-publishing-compon-pr-691.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* allow navigation items to only show in collapsed menu (PR #691)
+
 ## 13.5.3
 
 * Update error-summary component to use govuk-frontend styles (PR #692)
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -54,3 +54,9 @@
   background-color: govuk-colour("yellow");
   color: govuk-colour("black");
 }
+
+.govuk-header__navigation-item--collapsed-menu-only {
+  @include govuk-media-query($from: desktop) {
+    display: none;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_layout_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_header.html.erb
@@ -40,7 +40,8 @@
     <nav>
       <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
         <% navigation_items.each_with_index do |item, index| %>
-          <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>">
+          <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
+            <%= "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu] %>">
             <%= link_to(
               item[:text],
               item[:href],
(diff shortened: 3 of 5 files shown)


DilwoarH:
@andysellick - the description has been updated, please let me know if this makes sense now.

andysellick:
@DilwoarH thanks that's a lot clearer. However I still don't see the need to pass arbitrary classes - surely we know in advance the behaviour that's needed and could build that into the component. So instead of passing an arbitrary class, we pass a flag to say 'add this extra class to this menu item' then we can include that CSS in the component and avoid any glue code?

alex-ju:
While my understanding of the isolation principle seems to be different, I do get that we don't allow custom classes to any component yet (as far as I know at least) but instead we use a flag/attribute at the component level to add a modifier class. I think this is what Andy is suggesting too.

alex-ju:
For me isolated means that a component/element doesn't depend and is not affected by styles applied to another component/elements anywhere else in the document. So my view is that this is tightly couple with the "single responsibility principle" not across projects, but I can see how other people could see it differently, that's why is valuable to have a discussion about this in one of our next workshops.

DilwoarH:
@andysellick @alex-ju - I have changed it so that it doesn't add custom classes anymore but allows users to say which items should show in collapsed menu only

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -54,3 +54,11 @@
   background-color: govuk-colour("yellow");
   color: govuk-colour("black");
 }
+
+.gem-c-layout-header .govuk-header__navigation {
```
alex-ju: We shouldn't need this nesting, the navigation item modifier will suffice

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Move all commonly used imports into all_components.scss (PR #683)
```
andysellick: This has already been released - I think you need a rebase.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -40,7 +40,8 @@
     <nav>
       <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
         <% navigation_items.each_with_index do |item, index| %>
-          <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>">
+          <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
```
andysellick: This is much better. Might be cleaner to do it in a helper method, like https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/heading_helper.rb

spec/components/layout_header_spec.rb
```diff
@@ -41,11 +41,13 @@ def component_name
     navigation_items = [
       { text: "Foo", href: "/foo", active: true },
       { text: "Bar", href: "/bar" },
+      { text: "Hello", href: "/hello", show_only_in_collapsed_menu: true },
```
andysellick: Wondering about this option name... would it be better as `show_only_in_mobile_menu`? I'm not sure.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
DilwoarH: @alex-ju - updated

CHANGELOG.md
DilwoarH: @andysellick - rebased

PR #702, opened 2019-01-04
Refactor feedback component

tijmenb:
See commits for the deets.


app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
```diff
@@ -0,0 +1,66 @@
+<%
+contact_govuk_path = "/contact/govuk"
```
andysellick: Nitpick: would indent.

tijmenb:
```suggestion
  contact_govuk_path = "/contact/govuk"
```

PR #706, opened 2019-01-07
Add search filter to component guide

vanitabarrett:
Adds a search box to the component guide which allows users to filter the list of components underneath.

## How it looks
screen shot 2019-01-08 at 10 13 09

## How it looks with search term
screen shot 2019-01-08 at 11 00 51

Component guide for this PR:
https://govuk-publishing-compon-pr-706.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/javascripts/component_guide/filter-components.js
+++ b/app/assets/javascripts/component_guide/filter-components.js
@@ -1,0 +1,32 @@
+(function() {
+  window.GOVUK = window.GOVUK || {};
+
+  window.GOVUK.FilterComponents = function filterList(searchTerm) {
+    var itemsToFilter = document.querySelectorAll('.component-list li');
+
+    for (var i = 0; i < itemsToFilter.length; i++ ) {
+      var currentComponent = itemsToFilter[i];
+      var componentText = currentComponent.innerText.toLowerCase();
+
+      if (componentText.includes(searchTerm.toLowerCase())) {
+        currentComponent.classList.remove('component-guide-hidden');
+      } else {
+        currentComponent.classList.add('component-guide-hidden');
+      }
+    }
+  };
+
+  var formElement = document.querySelector('[data-module=filter-components]');
+
+  if (formElement) {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/component_guide/filter-components.js
```diff
@@ -0,0 +1,36 @@
+(function() {
+  window.GOVUK = window.GOVUK || {};
+
+  window.GOVUK.FilterComponents = function filterList(searchTerm) {
+    var itemsToFilter = document.querySelectorAll('.component-list li');
+
+    // Scroll so results are more visible
+    var position = document.querySelector('[data-module="filter-components"]').offsetTop;
+    window.scrollTo({top: position});
```
alex-ju: While I see the intent this results in a bit of a weird case where, while you type (e.g. `Details`) the input moves constantly around the page making it harder to follow. It may be only me – or we can get a user researcher involved and see how others feel about it 😁.

app/assets/stylesheets/component_guide/application.scss
```diff
@@ -24,10 +24,24 @@
...
+  }
+}
+
 .component-body {
   margin-bottom: $govuk-gutter * 1.5;
 }
 
+.component-guide-hidden {
+  display: none;
+  visibility: hidden;
```
alex-ju: I may be missing something, but why do we need both these attributes?

vanitabarrett:
@andysellick I'd used the search component because I think it gives more of a visual indicator about what the input does. You're right that it's unlikely people will hit the search button if things start filtering out when they're typing, but there's a possibility users could type in a search term which matches so many components that the filtering out is off the screen (to them, visually nothing changes). In those cases, they may hit search if they're unsure the search has done anything.

andysellick:
@vanitabarrett that's true, but in the case you outlined clicking the search button won't actually do anything, will it? Having said that, I think it looks okay.

tijmenb:
Just used this, very helpful! 👍 🥇

PR #711, opened 2019-01-15
Add type option to button component

andysellick:
Add an option to the button component to specify the button `type`. Currently defaults to submit (unless it's a link).

---

Component guide for this PR:
https://govuk-publishing-compon-pr-711.herokuapp.com/component-guide/button

### DIFF
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -4,7 +4,7 @@ module GovukPublishingComponents
   module Presenters
     class ButtonHelper
       attr_reader :href, :text, :title, :info_text, :rel, :data_attributes,
-        :margin_bottom, :target, :start, :secondary, :secondary_quiet, :destructive
+        :margin_bottom, :target, :type, :start, :secondary, :secondary_quiet, :destructive
 
       def initialize(local_assigns)
         @href = local_assigns[:href]
@@ -15,6 +15,7 @@ module GovukPublishingComponents
         @data_attributes = local_assigns[:data_attributes]
         @margin_bottom = local_assigns[:margin_bottom]
         @target = local_assigns[:target]
+        @type = local_assigns[:type]
         @start = local_assigns[:start]
         @secondary = local_assigns[:secondary]
         @secondary_quiet = local_assigns[:secondary_quiet]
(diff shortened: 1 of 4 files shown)


lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -28,14 +29,22 @@ def link?
...
-        options[:type] = "submit" unless link?
+        options[:type] = button_type
         options[:rel] = rel if rel
         options[:data] = data_attributes if data_attributes
         options[:title] = title if title
         options[:target] = target if target
         options
       end
 
+      def button_type
```
vanitabarrett: Could this be simplified to something like:

```
type || "submit" unless link?

```

spec/components/button_spec.rb
```diff
@@ -22,6 +22,18 @@ def component_name
     assert_select ".govuk-button", text: "Submit"
   end
 
+  it "renders with the correct types" do
+    render_component(text: "Submit")
+    assert_select ".govuk-button[type=submit]", text: "Submit"
+
+    render_component(text: "Link", href: "#")
+    assert_select "a.govuk-button", text: "Link"
+    assert_select "a.govuk-button[type=submit]", false
```
vanitabarrett: We may also want a test that checks if you pass a `href` and a `type`, that the type isn't applied (to confirm that type isn't applied if the button is actually a link).

tijmenb:
@andysellick what kind of button types are we expecting?

spec/components/button_spec.rb
andysellick: Good point, will add.

lib/govuk_publishing_components/presenters/button_helper.rb
andysellick: Ooh that's better, thanks.

andysellick:
@tijmenb I was going to use a button as a toggle, but since it was inside a form it wasn't possible to use the component as it is because clicking it would just submit the form.

Except there's ways around that that don't require this change, and it turns out I'm not going to use a button anyway, so it's possible I don't need this PR at all now. Unless it might be useful for anything?

alex-ju:
@andysellick it seems like we have [`html_options` which allows type attribute to be passed](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/button_helper.rb#L31) for this component.

lib/govuk_publishing_components/presenters/button_helper.rb
andysellick: @vanitabarrett have made those changes.

alex-ju:
Any blockers in getting this merged?

andysellick:
@alex-ju just need an approval, I think, I've made the changes @vanitabarrett suggested (and then I'll sort out the conflicts, obv).

spec/components/button_spec.rb
vanitabarrett: I'd be tempted to split this into multiple tests so if one part fails it's clearer exactly where it's gone wrong. But not a blocker

PR #712, opened 2019-01-18
Extend toggle module

andysellick:
Adds a new option to the `gem-toggle` Javascript module, allowing a class to be passed that will be used in the toggle, in place of the default `js-hidden` class. 

Needed for [finders](https://trello.com/c/HeKk9mpK/192-add-progressive-disclosure-to-facets-on-mobile), where I need to toggle the display of facets, but only on mobile. If we use the default `js-hidden` class, the toggle functionality is there on desktop as well, and we don't want that.

Also linted some files, recommend this be reviewed commit by commit.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-712.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Extend toggle module (PR #712)
+
 ## 13.6.0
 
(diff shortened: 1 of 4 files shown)


andysellick:
@vanitabarrett fair point, have updated the description.

PR #714, opened 2019-01-21
Add accordion component

injms:
Adds the Accordion component based on GOV.UK Frontend.

[Card](https://trello.com/c/9KTI0qjV/577-add-accordion-component-as-a-component-and-use-it-for-markdown-styleguide).

This component is required by Content Publisher to hide and show the Govspeak guidance.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-714.herokuapp.com/component-guide/accordion/

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
@@ -1,0 +1,15 @@
+@import "govuk-frontend/components/accordion/accordion";
+
+.govuk-accordion--condensed {
+  .govuk-accordion__section-button {
+    @include govuk-media-query($from: tablet) {
+      @include govuk-font($size: 19, $weight: bold);
+    }
+  }
+
+  .govuk-accordion__section-summary {
+    @include govuk-media-query($from: tablet) {
+      @include govuk-font($size: 16);
+    }
+  }
+}
--- a/app/views/govuk_publishing_components/components/_accordion.html.erb
+++ b/app/views/govuk_publishing_components/components/_accordion.html.erb
@@ -1,0 +1,45 @@
+<%
+  heading_helper = GovukPublishingComponents::Presenters::AccordionHelper.new(local_assigns)
+
+  id ||= "default-id-#{SecureRandom.hex(4)}"
+  items ||= []
+  condensed ||= false
+  accordion_classes = %w(gem-c-accordion govuk-accordion)
+  accordion_classes << 'govuk-accordion--condensed' if condensed
+
+  data_attributes ||= {}
+  data_attributes[:module] = 'accordion'
+%>
+<% if items.any? %>
+  <%= tag.div(class: accordion_classes, id: id, data: data_attributes) do %>
+    <% items.each_with_index do |item, i| %>
+      <%
+        # Nunjucks starts a loop on 1 and the client side JavaScript also
+        # adopts this behaviour. To prevent things from breaking, the index
+        # here also need to be increase by one. (Nunjucks is used by GOV.UK
+        # Frontend, which this component is based on.)
+        index = i + 1
+
+        item[:data_attributes] ||= nil
+
+        section_classes = %w(govuk-accordion__section)
+        section_classes << 'govuk-accordion__section--expanded' if item[:expanded]
+
+        summary_classes = %w(govuk-accordion__section-summary govuk-body)
+      %>
+      <%= tag.div(class: section_classes) do %>
+        <div class="govuk-accordion__section-header">
+          <%=
+            content_tag(
+              heading_helper.heading_tag,
+              content_tag('span', item[:heading][:text], class: "govuk-accordion__section-button", id: "#{id}-heading-#{index}", data: item[:data_attributes]),
+              class: 'govuk-accordion__section-heading'
+            )
+          %>
+          <%= tag.div(item[:summary][:text], id: "#{id}-summary-#{index}", class: summary_classes) if item[:summary].present? %>
+        </div>
+        <%= tag.div(item[:content][:html], id: "#{id}-content-#{index}", class: "govuk-accordion__section-content", 'aria-labelledby': "#{id}-heading-#{index}") %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/accordion.yml
+++ b/app/views/govuk_publishing_components/components/docs/accordion.yml
@@ -1,0 +1,351 @@
+name: Accordion (experimental)
+description: The accordion component lets users show and hide sections of related content on a page.
+govuk_frontend_components:
+  - accordion
+body: |
+  This component is based on the [design system accordion component](https://design-system.service.gov.uk/components/accordion/) and is currently experimental. If using this component, please feed back any research findings to the design team.
+
+accessibility_criteria: |
+  The accordion must:
+
+    * accept focus
+    * be usable with a keyboard
+    * indicate when they have focus
+    * be usable with touch
+    * be usable with voice commands
+    * have visible text
+    * indicate to users that each section can be expanded and collapsed
+    * inform the user when a step has been expanded or collapsed
+    * be readable when only the text of the page is zoomed in
+
+  Section headings must use a button element:
+
+    * so that sections can be toggled with the space and enter keys
+    * so that sections can't be opened in a new tab or window
+
+  When JavaScript is unavailable the component must:
+
+  * be fully expanded
+  * not be marked as expandable
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      items:
+        - heading:
+            text: "Writing well for the web"
+          content:
+            html: "<p class='govuk-body'>This is the content for Writing well for the web.</p>"
+        - heading:
+            text: "Writing well for specialists"
+          content:
+            html: "<p class='govuk-body'>This is the content for Writing well for specialists.</p>"
+        - heading:
+            text: "Know your audience"
+          content:
+            html: "<p class='govuk-body'>This is the content for Know your audience.</p>"
+        - heading:
+            text: "How people read"
+          content:
+            html: "<p class='govuk-body'>This is the content for How people read.</p>"
+  with_supplied_identification:
+    description: |
+      An `id` is optional as it's automatically generated, but it can be supplied if a specific `id` is required.
+
+      The `id` must be **unique** across the domain of your service - this is because as the open / closed state of individual instances of the accordion persists across page loads using `localStorage`.
+
+      Used as an `id` in the HTML for the accordion as a whole, and also as a prefix for the `id`s of the section contents and the buttons that open them, so that those `id`s can be the target of `aria-labelledby` and `aria-control` attributes.
+    data:
+      id: with-supplied-id-thats-unique-across-the-domain
+      items:
+        - heading:
+            text: "Writing well for the web"
+          content:
+            html: "<p class='govuk-body'>This is the content for Writing well for the web.</p>"
+        - heading:
+            text: "Writing well for specialists"
+          content:
+            html: "<p class='govuk-body'>This is the content for Writing well for specialists.</p>"
+        - heading:
+            text: "Know your audience"
+          content:
+            html: "<p class='govuk-body'>This is the content for Know your audience.</p>"
+        - heading:
+            text: "How people read"
+          content:
+            html: "<p class='govuk-body'>This is the content for How people read.</p>"
+  with_summary:
+    description: Adds a subheading below each section heading.
+    data:
+      items:
+        - heading:
+            text: "Understanding agile project management"
+          summary:
+            text: "Introductions, methods, core features."
+          content:
+            html:
+              '<ul class="govuk-list">
+                  <li>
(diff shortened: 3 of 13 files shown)


app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,34 @@
+<%
+  id ||= "default-id-#{SecureRandom.hex(4)}"
+  additional_classes ||= []
```
andysellick: What's the requirement for this? I'm looking into some other components that do this and I believe it's against our component model - it could encourage developers to add glue code to apps, which we want to avoid.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <% if items.present? %>
+    <div class="<%= wrapper_classes %>" data-module="accordion" id="<%= id %>">
+      <% items.each_with_index do |item, i| %>
+        <%
+          index = i + 1
+          item[:data_attributes] ||= nil
+        %>
+        <div class="govuk-accordion__section">
+          <div class="govuk-accordion__section-header">
+            <h2 class="govuk-accordion__section-heading">
```
andysellick: Might be worth considering a parameter to change the heading level here.

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: What do you mean by 'this'? The `id` or `additional_classes`?

andysellick: The `additional_classes` line.

injms: The accordion works in a nice and lovely fashion when given a container that's wide - but is inflexible when not in the ideal layout.

For example - in my particular use case this is the Govspeak guidance is next to the text editor in the content publisher; this is the set of `<details>` that is going to be replaced with an accordion:

> Partial screenshot of the content publisher interface, showing a narrow column containing the Govspeak guidance.

This is why I'd like to be able to extend the styling of the accordion to make it work when narrower. I think it's better to make it part of the component rather than pushing developers to wrap components in an extra `div` to override the styles.

A modifier could be used instead of adding extra classes - so you'd then have `govuk-accordion govuk-accordion--smaller-width-please` for example. 

Prior art for passing through classes exists in a couple of other components - for example, warning text and error message- though this is not documented for these components.

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: Making the heading level configurable was discussed in the  [GOV.UK Frontend accordion pull request](https://github.com/alphagov/govuk-frontend/pull/958#issuecomment-432990753), but hasn't been implemented. 

I'll take a look and see if any work I do could be fed back into the accordion in Frontend - but I'd prefer it to be part of a separate piece of work so it doesn't block this pull request.

andysellick: https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/heading_helper.rb

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: Thanks for the discussion on this yesterday @andysellick - I feel I understand more about the wider context now. 

As we talked about, I'm adding a parameter / flag which will trigger the modifier class, and I'll see about the appetite for feeding this modifier back to the component in GOV.UK Frontend.

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: Thanks! I've added this in, so a `heading_level` parameter can now be used.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,48 @@
...
+
+        item[:data_attributes] ||= nil
+
+        section_classes = 'govuk-accordion__section'
+        section_classes << ' govuk-accordion__section--expanded' if item[:expanded]
+
+        summary_classes = 'govuk-accordion__section-summary'
+        summary_classes << (condensed ? ' govuk-body-s' : ' govuk-body')
+      %>
+      <div class="<%= section_classes %>">
```
DilwoarH: ```<% tag.div class: section_classes do %>```

This might be better for consistency

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  condensed ||= false
+  accordion_classes = 'gem-c-accordion govuk-accordion'
+  accordion_classes << ' govuk-accordion--condensed' if condensed
+
+%>
+<% if items.any? %>
+  <%= tag.div(class: accordion_classes, id: id, data: data_attributes) do %>
+    <% items.each_with_index do |item, i| %>
+      <%
+        index = i + 1
```
DilwoarH: is there a reason why you're incrementing the index?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,48 @@
...
+              class: 'govuk-accordion__section-heading'
+            )
+          %>
+          <% if item[:summary].present? %>
+            <div class="<%= summary_classes %>" id="<%= id %>-summary-<%= index %>">
+              <%= item[:summary][:text] %>
+            </div>
+          <% end %>
+        </div>
+        <div id="<%= id %>-content-<%= index %>" class="govuk-accordion__section-content" aria-labelledby="<%= id %>-heading-<%= index %>">
```
DilwoarH: same as above

```
<% tag.div ... do %>
```

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,48 @@
...
+        <div class="govuk-accordion__section-header">
+          <%=
+            content_tag(
+              heading_helper.heading_tag,
+              content_tag('span', item[:heading][:text], class: "govuk-accordion__section-button", id: "#{id}-heading-#{index}", data: item[:data_attributes]),
+              class: 'govuk-accordion__section-heading'
+            )
+          %>
+          <% if item[:summary].present? %>
+            <div class="<%= summary_classes %>" id="<%= id %>-summary-<%= index %>">
```
DilwoarH: this can be put on one line by doing the following:

```
<% tag.div item[:summary][:text], class: summary_classes, id: "#{id}-summary-{index}" %>
```

DilwoarH:
Few minor things but looks good otherwise 👍

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: ✅

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: ✅

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,45 @@
...
+              class: 'govuk-accordion__section-heading'
+            )
+          %>
+          <%= tag.div(item[:summary][:text], id: "#{id}-summary-#{index}", class: summary_classes) if item[:summary].present? %>
+        </div>
+        <%= tag.div(item[:content][:html], id: "#{id}-content-#{index}", class: "govuk-accordion__section-content", 'aria-labelledby': "#{id}-heading-#{index}") %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
```
DilwoarH: needs an extra line :)

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -0,0 +1,316 @@
+name: Accordion (experimental)
+description: The accordion component lets users show and hide sections of related content on a page.
+govuk_frontend_components:
+  - accordion
+body: |
+  This component is based on the [design system accordion component](https://design-system.service.gov.uk/components/accordion/) and is currently experimental. If using this component, please feed back any research findings to the design team.
+
+accessibility_criteria: |
+  - The accordion must:
```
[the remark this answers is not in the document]
injms: ✅

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -0,0 +1,316 @@
...
+    * be usable with a keyboard
+    * indicate when they have focus
+    * be usable with touch
+    * be usable with voice commands
+    * have visible text
+
+examples:
+  default:
+    data:
+      id: example-accordion
```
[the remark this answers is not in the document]
injms: ✅

spec/components/accordion_spec.rb
```diff
@@ -0,0 +1,228 @@
...
+              },
+              {
+                heading: { text: 'Heading 3' },
+                content: { html: '<p>Content 3.</p>' }
+              }]
+    }
+
+    render_component(test_data)
+
+    assert_select '.govuk-accordion__section-button', text: /Heading [1-3]/, count: 3
```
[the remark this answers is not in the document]
injms: Yeah, the Venn diagram of the two sets of tests is so close to being a single circle that it's not worth running them both. I've removed the first set of tests.

spec/components/accordion_spec.rb
```diff
@@ -0,0 +1,228 @@
+require 'rails_helper'
+
+describe 'Accordion', type: :view do
+  def component_name
+    'accordion'
+  end
+
```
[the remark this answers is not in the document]
injms: ✅

andysellick:
@injms one other thought - does this component need a print stylesheet?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,44 @@
+<%
+  id ||= "default-id-#{SecureRandom.hex(4)}"
+  heading_helper = GovukPublishingComponents::Presenters::HeadingHelper.new(local_assigns)
```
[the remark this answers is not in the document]
injms: As discussed, I've added an accordion helper that copies the functionality that's needed from the heading helper. It seems a simpler solution for the moment and we can add create a separate card to reduce duplication if we think it's needed.

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -0,0 +1,316 @@
...
+  This component is based on the [design system accordion component](https://design-system.service.gov.uk/components/accordion/) and is currently experimental. If using this component, please feed back any research findings to the design team.
+
+accessibility_criteria: |
+  - The accordion must:
+    * accept focus
+    * be usable with a keyboard
+    * indicate when they have focus
+    * be usable with touch
+    * be usable with voice commands
+    * have visible text
```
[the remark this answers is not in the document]
injms: I've drawn a large amount of inspiration from the step-by-step criteria and added them to the accordion criteria. ✅

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -0,0 +1,11 @@
+@import "govuk-frontend/components/accordion/accordion";
+
+.govuk-accordion--condensed {
+  .govuk-accordion__section-button {
+    @include govuk-font($size: 19, $weight: bold);
+  }
+
+  .govuk-accordion__section-summary {
+    @include govuk-font($size: 16);
```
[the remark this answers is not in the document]
injms: Hmm, it is intentional - but now not so sure that it's a good idea. 

I've changed this to not condense the accordion on smaller screens - since that's when the grid stacks columns on top of each other, and the default accordion works well in that context.

injms:
> @injms one other thought - does this component need a print stylesheet?

Yes, yes it does need a print stylesheet - this is because the component in GOV.UK Frontend doesn't have one. I'll raise an issue on GOV.UK Frontend and put in some sensible defaults (open, remove appearance of buttons etc).

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -0,0 +1,48 @@
+<%
+  heading_helper = GovukPublishingComponents::Presenters::AccordionHelper.new(local_assigns)
+
+  id ||= "default-id-#{SecureRandom.hex(4)}"
+
+  items ||= []
+
```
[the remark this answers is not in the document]
alex-ju: Will do

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -0,0 +1,13 @@
+@import "govuk-frontend/components/accordion/accordion";
+
+@include govuk-media-query($from: mobile) {
+  .govuk-accordion--condensed {
+    .govuk-accordion__section-button {
+      @include govuk-font($size: 19, $weight: bold);
```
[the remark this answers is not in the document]
alex-ju: I can look into making them the same size. We're using responsive font sizes and that's why this is happening.

PR #720, opened 2019-01-23
Improve component conventions documentation

andysellick:
Trying to improve the component documentation by providing a list of dos and don'ts.

This is a bit of a work in progress, would appreciate some comments.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-720.herokuapp.com/component-guide/

### DIFF
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -19,13 +19,46 @@ The namespace indicates where a component lives. A single page on GOV.UK could r
 
 | Type | Location | Example | Description |
 | -- | -- | -- | -- |
-| Template | `app/views/components` | `_my-comp.html.erb` | The template logic and markup. The template defines the component’s API. Filename must begin with an underscore. |
+| Template | `app/views/components` | `_my-comp.html.erb` | [Template logic and markup](#template) |
 | Documentation | `app/views/components/docs` | `my-comp.yml` | [Describes the component](#write-documentation) |
 | Styles | `app/assets/stylesheets/components` | `_my-comp.scss` | [Component styles](#styles) |
+| Print styles | `app/assets/stylesheets/components/print` | `_my-comp.scss` | [Component styles](#styles) |
 | Images | `app/assets/images/govuk_publishing_components` | `my-comp.png` | [Images](#images) |
 | Scripts | `app/assets/javascripts/components` | `my-comp.js` | [Javascript enhancements](#javascript) |
-| Tests | `test/components` | `my_comp_test.rb` | Unit tests |
+| Tests | `test/components` | `my_comp_test.rb` | [Unit tests](#tests) |
+| Javascript tests | `spec/components` | `my_comp_spec.rb` | [Unit tests](#tests) |
 | Helpers | `lib/govuk_publishing_components/presenters` | `my_comp_helper.rb` | [Helpers](#helpers) |
+
+## Template
+
+The template logic and markup. The template defines the component’s API. Filename must begin with an underscore.
+
+If complex logic is required this should be handled using a [helper](#helpers).
+
+Example:
+
+```ruby
+<%
+  options ||= []
+  id ||= false
+  helper = GovukPublishingComponents::Presenters::MyComponentHelper.new(options)
+%>
+
+<% if options.any? %>
+  <div class="govuk-something" id="<%= id %>">
+    <h2>An example component</h2>
+    <%= helper.someContent %>
+  </div>
+<% endif %>
+```
+
+If a component includes a heading, consider including an option to control the heading level (see the [heading component](https://govuk-publishing-components.herokuapp.com/component-guide/heading/specific_heading_level) for example).
+
+Components can use other components within their template, if required (see the [input component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_input.html.erb#L37) for example).
+
+Complex components can be split into smaller partials to make them easier to understand and maintain (see the [feedback component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_feedback.html.erb) for example).
+
+Components should not have an option to include arbitrary classes as this could violate the principle of isolation. If there is a requirement for a styling variation on a component this should be included as an option e.g. `small: true`.
 
 ## Write documentation
 
@@ -145,7 +178,7 @@ With the exception of namespaces, follow the [GOV.UK Frontend CSS conventions](h
 
 All CSS selectors should follow the BEM naming convention shown above, explained in [more detail here](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md#block-element-modifier-bem).
 
-Note: to avoid long and complicated class names, we follow the [BEM guidance](http://getbem.com/faq/#css-nested-elements) that classes do not have to reflect the nested nature of the DOM.
+Note: to avoid long and complicated class names, we follow the [BEM guidance](http://getbem.com/faq/#css-nested-elements) that classes do not have to reflect the nested nature of the DOM. We also try to avoid nesting classes too deeply, so that styles can be overridden more easily if needed.
 
 ```scss
   // Avoid this:
@@ -163,6 +196,12 @@ Visit the links below for more information:
 
 * [Official BEM Documentation](https://en.bem.info/methodology/naming-convention/#css-selector-naming-convention)
 * [Guide on BEM naming conventions](https://webdesign.tutsplus.com/articles/an-introduction-to-the-bem-methodology--cms-19403)
+
+### Layout
+
+New components should be built with a bottom margin and no top margin.
+
+A standard for options to control this spacing has [not been decided upon yet](https://github.com/alphagov/govuk_publishing_components/pull/292), although it is likely we will adopt something using the [Design System spacing](https://design-system.service.gov.uk/styles/spacing/).
 
 ### Linting
 All stylesheets must be linted according to [the style rules](https://github.com/alphagov/govuk-lint/blob/master/configs/scss_lint/gds-sass-styleguide.yml) in [govuk-lint](https://github.com/alphagov/govuk-lint).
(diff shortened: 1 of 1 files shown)


docs/develop-component-checklist.md
```diff
@@ -0,0 +1,23 @@
+# Developing a component - things to remember
+
+##Don't forget..
```
vanitabarrett: I feel like this is already covered in the [component conventions](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#structure)?

Maybe we just need to expand on things there.

docs/develop-component-checklist.md
```diff
@@ -0,0 +1,23 @@
...
+
+- test file, and a JS spec file if there is any JS involved
+- put complex logic into a helper
+- print stylesheet
+
+##A component should..
```
vanitabarrett: This implies all components should have a bottom margin and tracking, which isn't always needed? Perhaps it's ok to implement a bottom margin without it being used anywhere yet, but not tracking.

docs/develop-component-checklist.md
```diff
@@ -0,0 +1,23 @@
...
+
+- only have bottom margin
+- include tracking (see the [select component](https://govuk-publishing-components.herokuapp.com/component-guide/select/with_tracking) for [example code](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/select.js)) and allow other data attributes to be passed
+- have an option to control the heading level if the component contains one (see the [heading component](https://govuk-publishing-components.herokuapp.com/component-guide/heading/specific_heading_level) for example)
+- test that the component doesn't fail if no data is passed
+
+##A component should not..
+
+- have an option to pass arbitrary classes. If there is a requirement for a styling variation on a component this should be included as an option e.g. `small: true`
```
vanitabarrett: This feels like something which should live in the component conventions?

vanitabarrett:
There's definitely some good things here but to me at the moment it seems like we could just add this into the existing component conventions doc? I'm not sure on the difference between the two.

andysellick:
@vanitabarrett thanks, good feedback. Yeah, I was wondering about where to put this - my aim was to provide a quick reference to help when putting a component together, but maybe you're right and it should just be added where appropriate to existing documentation. Or maybe we could restructure that to make it easier to navigate - at the moment it feels like it branches quite a lot rather than providing a series of steps to follow...

Ha, maybe we should do a 'building a component' step by step... :)

docs/component_conventions.md
```diff
@@ -19,14 +19,45 @@ The namespace indicates where a component lives. A single page on GOV.UK could r
...
-| Tests | `test/components` | `my_comp_test.rb` | Unit tests |
+| Tests | `test/components` | `my_comp_test.rb` | [Unit tests](#tests) |
+| Javascript tests | `spec/components` | `my_comp_spec.rb` | [Unit tests](#tests) |
 | Helpers | `lib/govuk_publishing_components/presenters` | `my_comp_helper.rb` | [Helpers](#helpers) |
 
+## Template
+
+The template logic and markup. The template defines the component’s API. Filename must begin with an underscore.
+
+If complex logic is required this should be handed using a [helper](#helpers).
```
vanitabarrett: Typo: `handled` not `handed`

docs/component_conventions.md
```diff
@@ -19,14 +19,45 @@ The namespace indicates where a component lives. A single page on GOV.UK could r
...
+
+<% if options.any? %>
+  <div class="govuk-something" id="<%= id %>">
+    <h2>An example component</h2>
+    <%= helper.someContent %>
+  </div>
+<% endif %>
+```
+
+If a component includes a heading, there should be an option to control the heading level (see the [heading component](https://govuk-publishing-components.herokuapp.com/component-guide/heading/specific_heading_level) for example).
```
vanitabarrett: I wonder if we need to reword this to make it more flexible. Not all components with headings require this control (e.g: page title), but I agree it's something to consider.

docs/component_conventions.md
```diff
@@ -19,14 +19,45 @@ The namespace indicates where a component lives. A single page on GOV.UK could r
...
+
+Components can use other components within their template, if required.
```
vanitabarrett: Can we link to an example? It might also be worth mentioning splitting complex components into partials and including them in the component, as we have examples of those.

docs/component_conventions.md
```diff
@@ -143,7 +174,7 @@ With the exception of namespaces, follow the [GOV.UK Frontend CSS conventions](h
 
 `.block__element--modifier {}`
 
-All CSS selectors should follow the BEM naming convention shown above, explained in [more detail here](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md#block-element-modifier-bem).
+All CSS selectors should follow the BEM naming convention shown above, explained in [more detail here](https://github.com/alphagov/govuk-frontend/blob/master/docs/contributing/coding-standards/css.md#block-element-modifier-bem). Try to avoid nesting classes if possible.
```
vanitabarrett: I think the addition here about nesting repeats what is said in the paragraph immediately below?

docs/component_conventions.md
```diff
@@ -188,6 +226,16 @@ Follow the [GOV.UK Frontend JS conventions](https://github.com/alphagov/govuk-fr
 
 Scripts should use the [module pattern provided by govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules) and be linted using [StandardJS](https://standardjs.com/).
 
+Most components should include tracking (see the [select component](https://govuk-publishing-components.herokuapp.com/component-guide/select/with_tracking) for [example code](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/select.js)).
```
vanitabarrett: 'Most components should include tracking': maybe include something like 'if there are interactive elements to it'?

nickcolley: Tracking can have a negative impact on users (privacy, performance).

We already have a problem with things being tracked without a real reason for doing so, I think you should reconsider making tracking a default in components.

Could we make this clear that tracking is only added to components that can be tied to something a team needs to know, and have a process to review and remove that?

docs/component_conventions.md
andysellick: Not exactly. I'm trying to say we don't need to have long chains of selectors, like `.class1 .class2 .class3 .class4 { ... }` whereas the para below is talking about not naming individual classes to reflect the dom structure, e.g. `.block__elem1__elem2__elem3`. Trying to think of a way to make this clearer.

andysellick:
@vanitabarrett @nickcolley thanks for the good feedback both. I've reworded everything, hopefully better now.

docs/component_conventions.md
```diff
@@ -164,6 +197,13 @@ Visit the links below for more information:
 * [Official BEM Documentation](https://en.bem.info/methodology/naming-convention/#css-selector-naming-convention)
 * [Guide on BEM naming conventions](https://webdesign.tutsplus.com/articles/an-introduction-to-the-bem-methodology--cms-19403)
 
+### Layout
+
+New components should be built with a bottom margin and no top margin.
+
+It may be desirable for components to have an option to control the size of the bottom margin but there is no standard for this right now. This is a [complex issue](https://github.com/alphagov/govuk_publishing_components/pull/292), a suggested model has been implemented in the [heading component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss#L31).
```
nickcolley: Non-blocking comment: Is the standard the design system spacing scale?

I imagined being able to pass numbers from the spacing scale and it working on any component but using the helper classes.

https://design-system.service.gov.uk/styles/spacing/

`bottom-spacing: 2`

`<div class="govuk-!-margin-bottom-2"></div>`

Probably a discussion for another PR

andysellick: Ah, good point. That RFC was written before the design system was adopted, I'll have a think about how to reword this bit.

andysellick: Have reworded. Hopefully we'll be able to update this section soon!

docs/component_conventions.md
```diff
@@ -164,6 +197,12 @@ Visit the links below for more information:
 * [Official BEM Documentation](https://en.bem.info/methodology/naming-convention/#css-selector-naming-convention)
 * [Guide on BEM naming conventions](https://webdesign.tutsplus.com/articles/an-introduction-to-the-bem-methodology--cms-19403)
 
+### Layout
+
+New components should be built with a bottom margin and no top margin.
```
alex-ju: 💯

docs/component_conventions.md
alex-ju: @nickcolley I'm a bit concerned about giving the option of defining a custom `margin-bottom` to any component (even based on the spacing scale) as it will introduce spacing inconsistencies. I think this is very much the same with allowing custom classes at the component level - the only difference would be that this would affect only the margin property.

docs/component_conventions.md
```diff
@@ -164,6 +197,12 @@ Visit the links below for more information:
...
+
+A standard for options to control this spacing has [not been decided upon yet](https://github.com/alphagov/govuk_publishing_components/pull/292), although it is likely we will adopt something using the [Design System spacing](https://design-system.service.gov.uk/styles/spacing/).
```
DilwoarH: +1 for design system spacing

docs/component_conventions.md
```diff
@@ -188,6 +227,16 @@ Follow the [GOV.UK Frontend JS conventions](https://github.com/alphagov/govuk-fr
...
 
+Most components should have an option to include arbitrary data attributes (see the [checkboxes component](https://govuk-publishing-components.herokuapp.com/component-guide/checkboxes/checkboxes_with_data_attributes) for example). These can be used for many purposes including tracking (see the [select component](https://govuk-publishing-components.herokuapp.com/component-guide/select/with_tracking) for [example code](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/select.js)) but specific tracking should only be added to a component where there is a real need for it.
+
+Some [common Javascript modules](https://github.com/alphagov/govuk_publishing_components/tree/master/app/assets/javascripts/govuk_publishing_components/lib) are available. If new functionality is required, consider adding it as a common module.
+
+## Tests
+
+Component tests should include a check that the component doesn't fail if no data is passed.
```
DilwoarH: 💯

andysellick:
Seems like everyone's okay with this change? Are you happy to approve?

PR #722, opened 2019-01-24
Move select.js into /lib as it's used by other elements.

peteglondon:
# What

Move select.js from `components/` to `lib/`

# Why

Because it's in use by other elements than the components here.

Component guide for this PR:
https://govuk-publishing-compon-pr-722.herokuapp.com/component-guide/


andysellick:
@vanitabarrett what other components use this code? Is it worth documenting this somewhere?

vanitabarrett:
@andysellick It's used by the new sorting "button" on finders (actually a select but not the gem select component, I think because it needs custom Javascript for the sorting). This moves the tracking JS for selects into a common place so it can be used by other apps, the same way the GemToggle can.

PR #724, opened 2019-01-28
Add govuk-frontend spacing scale to hint

andysellick:
**This PR cannot be merged until finder-frontend has been updated to use this new option, [coming soon](https://github.com/alphagov/finder-frontend/pull/833)**

This PR removes the option to pass an arbitrary class to the hint component and replaces it with an option to pass a number between 0 and 9 to provide variable bottom margin using the [govuk-frontend spacing classes](http://govuk-frontend-review.herokuapp.com/docs/#settings/spacing-variable-govuk-spacing-points).

screen shot 2019-01-29 at 08 50 17

This removes the potential loss of isolation for the component and replaces it with an option that provides the functionality for which the class option was intended. 

Places where hint is used:

- finder-frontend (where it is given a a govuk-frontend spacing class)
- file upload component, using the default margin
- textarea component, using default
- radio component, using default
- input component, using default

---

Component guide for this PR:
https://govuk-publishing-compon-pr-724.herokuapp.com/component-guide/hint

### DIFF
--- a/app/views/govuk_publishing_components/components/_hint.html.erb
+++ b/app/views/govuk_publishing_components/components/_hint.html.erb
@@ -1,8 +1,9 @@
 <%
   id ||= "hint-#{SecureRandom.hex(4)}"
-  classes ||= ''
+  margin_bottom ||= 3
+
   css_classes = %w( gem-c-hint govuk-hint )
-  css_classes << classes if classes
+  css_classes << ([*0..9].include?(margin_bottom) ? "govuk-!-margin-bottom-#{margin_bottom}" : "govuk-!-margin-bottom-3")
 %>
 
 <%= tag.span id: id, class: css_classes do %>
--- a/spec/components/hint_spec.rb
+++ b/spec/components/hint_spec.rb
@@ -8,6 +8,24 @@ describe "Hint", type: :view do
   it "renders hint" do
     render_component(text: "For example, ‘QQ 12 34 56 C’.")
 
-    assert_select(".govuk-hint", text: "For example, ‘QQ 12 34 56 C’.")
+    assert_select '.govuk-hint.govuk-\!-margin-bottom-3', text: "For example, ‘QQ 12 34 56 C’."
+  end
+
+  it "applies a specified bottom margin" do
+    render_component(text: "For example, ‘QQ 12 34 56 C’.", margin_bottom: 7)
+
+    assert_select '.govuk-hint.govuk-\!-margin-bottom-7'
+  end
+
+  it "applies zero bottom margin" do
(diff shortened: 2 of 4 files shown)


spec/components/hint_spec.rb
```diff
@@ -8,6 +8,18 @@ def component_name
   it "renders hint" do
     render_component(text: "For example, ‘QQ 12 34 56 C’.")
 
-    assert_select(".govuk-hint", text: "For example, ‘QQ 12 34 56 C’.")
+    assert_select '.govuk-hint.govuk-\!-margin-bottom-3', text: "For example, ‘QQ 12 34 56 C’."
+  end
+
+  it "applies a specified bottom margin" do
```
vanitabarrett: Should we add a test which makes sure the class is not applied if you pass a number that's not 0-9?

andysellick: Yes, good point. Had to rewrite the template a bit, but I think it's better now.

app/views/govuk_publishing_components/components/_hint.html.erb
```diff
@@ -1,8 +1,9 @@
 <%
   id ||= "hint-#{SecureRandom.hex(4)}"
-  classes ||= ''
+  margin_bottom ||= 3
+
   css_classes = %w( gem-c-hint govuk-hint )
-  css_classes << classes if classes
+  css_classes << ([*0..9].include?(margin_bottom) ? "govuk-!-margin-bottom-#{margin_bottom}" : "govuk-!-margin-bottom-3")
```
nickcolley: If/when this moves into other components it could be a helper?

spec/components/hint_spec.rb
```diff
@@ -8,6 +8,24 @@ def component_name
...
+    render_component(text: "For example, ‘QQ 12 34 56 C’.", margin_bottom: 7)
+
+    assert_select '.govuk-hint.govuk-\!-margin-bottom-7'
+  end
+
+  it "applies zero bottom margin" do
+    render_component(text: "For example, ‘QQ 12 34 56 C’.", margin_bottom: 0)
```
nickcolley: I was wondering if the API could avoid CSS and be 'spacing_bottom' but this feels as explicit as possible so 👍

app/views/govuk_publishing_components/components/_hint.html.erb
andysellick: Agreed. I've got some more candidates but wanted to see if the initial model would meet with approval first.

spec/components/hint_spec.rb
andysellick: I don't know of an example of it, but at some point there might be a component where we might want to control `padding_bottom` for some reason, being this explicit seemed like a good approach.

PR #725, opened 2019-01-31
Show finder breadcrumb for specialist documents

karlbaker02:
This PR stops the contextual breadcrumbs component from showing taxonomy breadcrumbs when the content being viewed is a specialist document. This allows the content item's finder to be displayed instead, for which the logic already exists in the `breadcrumbs` method of `contextual_navigation.rb`. By making this change, we allow users to more easily navigate to the correct specialist finder if they come from somewhere other than a specialist finder.

Note that if `prioritise_taxon_breadcrumbs` is specified and set to `true` as part of instantiating the contextual breadcrumbs component, it is possible to override the above behaviour.

Solo: @karlbaker02

Example: for a specialist doc at https://www.gov.uk/cma-cases/funerals-market-study

Before:

screen shot 2019-01-31 at 15 22 09

After:

screen shot 2019-01-31 at 15 22 25

---

Component guide for this PR:
https://govuk-publishing-compon-pr-725.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Update contextual breadcrumbs to not show taxonomy breadcrumbs when showing specialist documents (PR #725)  
+
 ## 13.7.0
 
 * Extend accessible autocomplete onConfirm function (PR #718)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Update contextual breadcrumbs to not show taxonomy breadcrumbs when showing specialist documents
```
andysellick: Please add the PR number to this line (see other lines for examples).

vanitabarrett:
@karlbaker02 Have you got any examples of what this will look like on specialist documents? (e.g: from running it locally)

CHANGELOG.md
karlbaker02: Sorry, PR wasn't created yet when I pushed this.

karlbaker02:
Also a good point @vanitabarrett, I meant to do this. Will do so now :)

PR #728, opened 2019-02-05
Fix accessible autocomplete so it can handle HTML entities

maxgds:
Fix autocomplete js:
This fixes an issue where ampersands (and other characters) could not be
matched in the select box, resulting in an empty selection. This failed
silently in the background so the user was submitting blank choices they
were not aware of.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-728.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 13.8.1
+
+* Fix autocomplete to allow it to handle HTML entities in the select list (PR #728)
+
 ## 13.8.0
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 13.8.1
```
andysellick: Shouldn' this be 'unreleased'? (sorry, I'm a bit late to this one)

maxgds: I messed it up a bit - I meant to release it but forgot to bump the gem files, have done that separately now.

PR #729, opened 2019-02-05
Adds custom event category for checkbox change events. Any app with it's own tracking logic for checkboxes will need to remove it

oscarwyatt:
As checkboxes fire their own tracking event it will be necessary to
remove any custom logic in apps that listens for change events
otherwise it will fire twice.

Trello: https://trello.com/c/l6kGfBhh/230-add-missing-facet-tag-email-alert-tracking-and-checkbox-issue

Component guide for this PR:
https://govuk-publishing-compon-pr-729.herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* `BREAKING:` Adds custom event category for checkbox change events. Any app with it's own tracking logic for checkboxes will need to remove it (PR #729)
+
 ## 14.0.0
 
 * BREAKING: Add govuk-frontend spacing scale to hint (PR #724)
--- a/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
@@ -23,6 +23,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       });
 
+      $(scope).find('input[type=checkbox]').on('change', function(e) {
+        if (GOVUK.analytics && GOVUK.analytics.trackEvent) {
+          var $checkbox = $(e.target);
+          var category = $checkbox.data("track-category");
+          if (typeof category !== "undefined") {
+            var isChecked = $checkbox.is(":checked");
+            var uncheckTrackCategory = $checkbox.data("uncheck-track-category");
+            if (!isChecked && typeof uncheckTrackCategory !== "undefined") {
+              category = uncheckTrackCategory;
+            }
+            var action = $checkbox.data("track-action");
+            var options = $checkbox.data("track-options");
+            if (typeof options !== 'object' || options === null) {
+              options = {};
+            }
+            options['value'] = $checkbox.data("track-value");
+            options['label'] = $checkbox.data("track-label");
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Adds custom event category for checkbox change events. Any app with it's own tracking logic for checkboxes will need to remove it (PR #729)
```
vanitabarrett: Should we mark this as a breaking change? Like https://github.com/alphagov/govuk_publishing_components/pull/729/files#diff-4ac32a78649ca5bdd8e0ba38b7006a1eR82

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
```diff
@@ -23,6 +23,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        var category = $checkbox.data("track-category");
+        if (typeof category !== "undefined") {
+          var isChecked = $checkbox.is(":checked");
+          var uncheckTrackCategory = $checkbox.data("uncheck-track-category");
+          if (!isChecked && typeof uncheckTrackCategory !== "undefined") {
+            category = uncheckTrackCategory;
+          }
+          var action = $checkbox.data("track-action");
+          var label = $checkbox.data("track-label");
+          var value = $checkbox.data("track-value");
```
vanitabarrett: What is `track-value`? And do we need to also be fetching `track_options` if they are passed into the component?

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -114,7 +114,7 @@ examples:
         - label: "Blue"
           value: "blue"
   checkboxes_with_data_attributes:
-    description: Data attributes such as tracking can be applied if required. Note that for a checkbox this will not fire tracking events automatically, so if you want your tracking to work you'll need to write some tracking Javascript in the app around the component to detect it.
+    description: Data attributes such as tracking can be applied if required. This will fire tracking events.
```
vanitabarrett: We should probably make it clear that this will fire an identical event for checking and unchecking, and then point people to the example below if they need to be able to distinguish between the 2.

CHANGELOG.md
andysellick: This should just be the title of the PR.

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
```diff
@@ -23,6 +23,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            category = uncheckTrackCategory;
+          }
+          var action = $checkbox.data("track-action");
+          var options = $checkbox.data("track-options");
+          if(typeof options !== 'object' || options === null){
+            options = {};
+          }
+          options['value'] = $checkbox.data("track-value");
+          options['label'] = $checkbox.data("track-label");
+          GOVUK.analytics.trackEvent(category, action, options);
```
vanitabarrett: We probably want to wrap this in `if (GOVUK.analytics && GOVUK.analytics.trackEvent) {` like we do for other tracking JS, as otherwise it errors when GOVUK.analytics is not available.

PR #731, opened 2019-02-08
Increase focus contrast of feedback links

selfthinker:
The contrast of the focused feedback links doesn't meet WCAG criterion 1.4.3 (minimum contrast).
This changes the contrast to use the same <del>slightly darker colour that is also used for other links, copied from [govuk_template](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_accessibility.scss#L62-L65)</del> <ins>black-ish colour that is also used for in the Design System and GOV.UK Frontend</ins>.

feedback-focus-colour-contrast-issue

Component guide for this PR:
https://govuk-publishing-compon-pr-731.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -84,7 +84,7 @@
   color: $white;
 
   &:focus {
-    color: $link-colour;
+    color: $govuk-focus-text-colour;
   }
 }
 


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -84,7 +84,7 @@
   color: $white;
 
   &:focus {
-    color: $link-colour;
+    color: darken($link-colour, 2.5%);
```
tijmenb: I think we can use `$govuk-focus-text-colour` from GOV.UK Frontend here?

selfthinker: That would be inconsistent with the rest of GOV.UK. I'm not sure if this could be an intentional tradeoff for moving towards using more styles from the Design System?

tijmenb: I'd be fine with a slight inconsistency within GOV.UK, since the Design System style is better and will replace the rest anyway. We're already using it here:

https://github.com/alphagov/govuk_publishing_components/blob/4a9a3267ae6b5a70af7a127bfae633bcec0ebf23/app/assets/stylesheets/govuk_publishing_components/components/_table.scss#L76-L78

tijmenb: Found that we're even using the Design System style in content body (like https://www.gov.uk/vehicle-log-book), since [we use this GOV.UK Frontend mixin for govspeak](https://github.com/alphagov/govuk_publishing_components/blob/4a9a3267ae6b5a70af7a127bfae633bcec0ebf23/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss#L80).

selfthinker: Oh, interesting. I will update this then.

selfthinker: Done.

alex-ju: Thanks for catching this @tijmenb – indeed it's better to use this variable especially because colours (including highlight may change soon, so by doing this we're sure we're using the latest colours which have been tested for contrast), and thanks @selfthinker for this update – there are still a couple of links on gov.uk that don't rely on `govuk-frontend` hence don't have enough contrast.

alex-ju:
@selfthinker do you mind adding this update to the [changelog file](https://github.com/alphagov/govuk_publishing_components/blob/master/CHANGELOG.md) under an "Unreleased" title?

selfthinker:
Yup, I updated the changelog in #732.

PR #733, opened 2019-02-08
Add `autocomplete` attribute to feedback form and use it on feedback component's survey signup form

selfthinker:
* Add `autocomplete` attribute option to input component
* Use `autocomplete=email` on feedback component's survey signup form to conform with WCAG criterion 1.3.5 (Identify Input Purpose)

I'm not sure if it also makes sense to add the `autocomplete` attribute option to other form components? Checkboxes and radio buttons probably don't make sense, but selects and textareas might, not sure about file upload.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-[733].herokuapp.com/component-guide/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add `autocomplete` attribute option to input component and use it on feedback component's survey signup form (PR #733)
+
 ## 16.3.0
 
 * Fix issue embedding videos in govspeak after YouTube API is loaded (PR #775)
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -5,6 +5,7 @@
   describedby ||= nil
   controls ||= nil
   data ||= nil
+  autocomplete ||= nil
 
   label ||= nil
   hint ||= nil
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add `autocomplete` attribute option to input component
+* Use `autocomplete=email` on feedback component's survey signup form
```
andysellick: This should be a list of the PRs, not a list of the changes. So this entry should be:

`- Add autocomplete attribute to feedback form (PR #733)`

selfthinker:
> one test to check it isn't added when it shouldn't be

None of the other tests does anything like that. I would generally assume that something like that wouldn't be required? Then you would otherwise also need a test that checks that e.g. no `data` attribute is added or no `aria-describedby` etc.

selfthinker:
I have written one test, adjusted the changelog and fixed the conflict.

CHANGELOG.md
```diff
@@ -7,7 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
 * Add description_html option to notice component (PR #740)
+* Add `autocomplete` attribute option to input component and use it on feedback component's survey signup form (PR #733)
```
andysellick: This should literally be the name of the PR, see my previous comment.

selfthinker: I've rebased but instead of adjusting the changelog entry, I changed the name of the PR. ;-)

Issue #736, opened 2019-02-12
Possible injection if axe warning has HTML


nickcolley:
If you wrap the tag component's text with <marquee> it'll also apply to the axe warning since the output has HTML in the markup.

nickcolley:
We could solve this by constructing this DOM with `document.createElement` and `textNode.textContent` which'll stop raw HTML from being rendered.

https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/component_guide/accessibility-test.js#L136

barrucadu:
Here's a reproduction:

```
<%= tag.div class: "gem-c-tag" do %>
  <marquee><%= text %></marquee>
<% end %>
```

Maybe the most surprising lesson here is that Chrome still supports `<marquee>`.

PR #737, opened 2019-02-12
Reformat output accessibility error to avoid injections

nickcolley:
Some axe summaries include HTML, so by constructing the DOM manually we can avoid any issues.

This also adds some typography improvements and headings

## Before

screen shot 2019-02-12 at 13 05 27
screen shot 2019-02-12 at 13 05 22

## After

screen shot 2019-02-12 at 13 04 31
screen shot 2019-02-12 at 13 04 26

Fixes https://github.com/alphagov/govuk_publishing_components/issues/736

---

Component guide for this PR:
https://govuk-publishing-compon-pr-737.herokuapp.com/component-guide/

### DIFF
--- a/app/assets/javascripts/component_guide/accessibility-test.js
+++ b/app/assets/javascripts/component_guide/accessibility-test.js
@@ -132,15 +132,53 @@
         var activeElParent = _findParent(activeEl, '[data-module="test-a11y"]')
         var wrapper = activeElParent.querySelector(resultContainerSelector)
 
-        if (wrapper) {
-          var resultHTML = '<h3>(' + result.id + ') ' + result.summary + ' <a href="' + result.url + '">(see guidance)</a></h3>' +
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/component_guide/accessibility-test.js
```diff
@@ -132,15 +132,53 @@
...
-          selectorObj.selector +
-          '</span></p>'
-
-          wrapper.insertAdjacentHTML('beforeend', resultHTML)
+        if (!wrapper) {
+          return
         }
+
+        // Section to announce the overall problem.
+        var headerNodeLink = document.createElement('a')
```
nickcolley: Might be better to use `<template>` rather than constructing the DOM like this here...

alex-ju: Indeed, but does the job for now. Thanks for doing this work!

PR #740, opened 2019-02-12
Support rendering raw HTML in notice component

benthorner:
https://trello.com/c/HcfYSpjX/634-allow-notice-component-to-accept-html

Previously only paragraph text or Govspeak could be passed to this
component. This adds a 'description_text' option as an alternative,
which is not nested in any other tag.

As with the Govspeak component, it seems better to require the consumer
of this component to specify that the HTML they're passing-in is safe,
so it was necessary to specify this in the tests.

---

Component guide for this PR:

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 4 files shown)


benthorner:
@alex-ju the breaking change might be coming next ;-).

PR #745, opened 2019-02-13
Bump version to 15.1.0

benthorner:
https://trello.com/c/HcfYSpjX/634-allow-notice-component-to-accept-html

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 15.1.0
+
 * Add description option to notice component (PR #740, PR #742)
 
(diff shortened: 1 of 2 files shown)


andysellick:
@benthorner I think this PR should have included a change to `Gemfile.lock` (as per this PR: https://github.com/alphagov/govuk_publishing_components/pull/739/files) but it seems to have created a new version of the gem anyway. Might be worth checking that the new version contains everything it should and that it works okay? If it does maybe we don't need to update the lock file after all...

benthorner:
@andysellick you're right - I do keep getting caught-out by this. Here's a fix: https://github.com/alphagov/govuk_publishing_components/pull/746.

PR #747, opened 2019-02-14
Allow embed option in component docs

kevindew:
Trello: https://trello.com/c/mHrLEKSK/635-allow-govukpublishingcomponents-to-show-contextual-code-around-an-example

The embed option allows for component examples to be illustrated within
contextual HTML. It is intended for components that do not have a visual
presence without some other HTML. The Content Publisher application has
a number of these, such as Contextual Guidance and Input Length
Suggestor, and is likely to have another for modals. These cannot be
viewed inside the component-guide and just appear as empty boxes.

This resolves this by allowing the ability to include an option of
`embed` in a component document file which contains the HTML that is
needed in the vicinity of the component eg:

```
name: Component
embed: |
  <button data-trigger-component=true>Click me</button>
  <%= component %>
examples:
  default:
```

This can also be applied on individual examples eg:

```
name: Component
examples:
  default:
  inside_div:
    description: This component behaves differently inside a div
    embed: |
      <div><%= component %></div>
```

## Examples

There's an [example component in the dummy app](https://govuk-publishing-compon-pr-747.herokuapp.com/component-guide/test-component-with-embed)

More useful though is looking at a couple of cases in Content Publisher:

Before:
screen shot 2019-02-14 at 11 43 03

After:
screen shot 2019-02-14 at 11 46 03

Before:
screen shot 2019-02-14 at 11 43 14

After:
screen shot 2019-02-14 at 11 45 48

There's currently a workaround for this in this PR: https://github.com/alphagov/content-publisher/pull/735

### DIFF
--- a/app/models/govuk_publishing_components/component_example.rb
+++ b/app/models/govuk_publishing_components/component_example.rb
@@ -6,7 +6,8 @@ module GovukPublishingComponents
                 :data,
                 :context,
                 :description,
-                :block
+                :block,
+                :embed
 
     def initialize(id, example)
       @id = id
@@ -14,6 +15,7 @@ module GovukPublishingComponents
       @context = example["context"] || {}
       @description = example["description"] || false
       @block = @data.delete(:block) || false
+      @embed = example["embed"]
     end
 
     def name
@@ -71,6 +73,10 @@ module GovukPublishingComponents
 
     def has_block?
       !!block
+    end
+
+    def has_embed?
+      !!embed
     end
 
   private
--- a/app/views/govuk_publishing_components/component_guide/component_doc/_call.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/component_doc/_call.html.erb
@@ -1,12 +1,20 @@
+<%
+component_code =
+  if example.has_block?
+    "<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} do %\>\n" +
+    "  #{example.block.html_safe}" +
+    "<% end %\>"
+  else
+    "<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} %\>"
+  end
+
+code =
+  if example.has_embed?
+    example.embed.sub("<%= component %\>", component_code)
+  else
+    component_code
+  end
+%>
 <div class="component-call component-highlight" contenteditable>
-  <% if example.has_block? %>
-    <%
-      code = "<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} do %\>\n"
-      code += "  #{example.block.html_safe}"
-      code += "<% end %\>"
-    -%>
-    <pre><code><%= example.highlight_code(code) %></code></pre>
-  <% else %>
-    <pre><code><%= example.highlight_code("\<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} %\>") %></code></pre>
-  <% end %>
+  <pre><code><%= example.highlight_code(code) %></code></pre>
 </div>
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -127,6 +127,32 @@ examples:
         <span>Some text</span>
 ```
 
+#### Yaml configuration for components that need contextual HTML
+
+If a component is only visible, or behaves differently, in a certain context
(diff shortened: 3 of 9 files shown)


docs/component_conventions.md
```diff
@@ -127,6 +127,32 @@ examples:
...
+
+To configure a HTML embed in the component yaml file you can specify `embed` at
+the root or individual examples:
+
+```yaml
+embed: |
+  <button class="trigger-for-component">Click me<button>
+  <%= component %>
+examples:
+  default:
```
benthorner: Really minor, but should we lose the `default:`?

app/views/govuk_publishing_components/component_guide/component_doc/_call.html.erb
```diff
@@ -1,12 +1,20 @@
...
+    "<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} do %\>\n" +
+    "  #{example.block.html_safe}" +
+    "<% end %\>"
+  else
+    "<%= render \"#{@component_doc.partial_path}\", #{example.pretty_data} %\>"
+  end
+
+code =
+  if example.has_embed?
+    example.embed.sub('<%= component %' + '>', component_code)
```
benthorner: Why do we have a '+' here?

app/models/govuk_publishing_components/component_example.rb
```diff
@@ -6,14 +6,17 @@ class ComponentExample
...
+                :block,
+                :embed
 
     def initialize(id, example)
       @id = id
       @data = example["data"] || {}
       @context = example["context"] || {}
       @description = example["description"] || false
       @block = @data.delete(:block) || false
+      @description = example["description"] || false
```
benthorner: We seem to have this line twice :-|.

spec/component_guide/component_example_spec.rb
```diff
@@ -43,6 +43,34 @@
...
+
+    example = <<~EXAMPLE
+      <input class="input-control-for-embedded-component">
+      <%= render "components/test-component-with-embed", {
+      } %>
+    EXAMPLE
+    expect(page).to have_content(example)
+  end
+
+  it 'allows specific examples to embedd different HTML' do
```
benthorner: 'dd' -> 'd'

docs/component_conventions.md
kevindew: We need it here if we want to illustrate both the default one and a specific one. I know it looks a bit funky 😞

app/views/govuk_publishing_components/component_guide/component_doc/_call.html.erb
kevindew: Having `%>` together terminates the embedded ruby block. I couldn't find a better way to escape. Do you know one?

I'll update commit message to mention it.

kevindew: Oh darn there's one in the lines above 😞. Massive d'oh!

spec/component_guide/component_example_spec.rb
kevindew: Thanks 👍

app/models/govuk_publishing_components/component_example.rb
kevindew: oh dear! thanks

kevindew:
Thanks @benthorner I've made those amends

docs/component_conventions.md
benthorner: Ah. Could we just put something small in here, then?

app/views/govuk_publishing_components/component_guide/component_doc/_call.html.erb
benthorner: That looks better :-).

docs/component_conventions.md
kevindew: I don't think there's anything we can really put since it doesn't need anything. It's just an aspect (or failing) of the config file design.

benthorner: No worries - it was only minor.

kevindew:
@benthorner got the tests fixed.

PR #752, opened 2019-02-15
Change Header link from announcements to news and communications

bilbof:
This will change Announcements to News and communications
throughout govuk. This has already been done in whitehall
with https://github.com/alphagov/whitehall/pull/4636,
so this will make things consistent.

Example page that will change once the version has been
bumped and government-frontend deployed:
https://www.gov.uk/government/news/assets-liquidated-an-unhappy-valentine-for-teesside-crook

https://trello.com/c/7NS3f7I8/378-investigate-why-announcements-link-is-persisting-in-whitehall-header-in-some-places

### DIFF
--- a/app/views/govuk_publishing_components/components/_government_navigation.html.erb
+++ b/app/views/govuk_publishing_components/components/_government_navigation.html.erb
@@ -40,8 +40,8 @@
       </a>
     </li>
     <li>
-      <a class="<%= 'active' if active == 'announcements' %>" href="/government/announcements">
-        <%= t("govuk_component.government_navigation.announcements", default: "Announcements") %>
+      <a class="<%= 'active' if active == 'announcements' %>" href="/news-and-communications">
+        <%= t("govuk_component.government_navigation.news_and_communications", default: "News and communications") %>
       </a>
     </li>
   </ul>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Support an embed option on examples to allow embedding a component in a particular HTML context (PR #747)
+* Change header link from Announcements to News and communications (PR #752)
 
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_government_navigation.html.erb
```diff
@@ -40,8 +40,8 @@
       </a>
     </li>
     <li>
-      <a class="<%= 'active' if active == 'announcements' %>" href="/government/announcements">
-        <%= t("govuk_component.government_navigation.announcements", default: "Announcements") %>
+      <a class="<%= 'active' if active == 'announcements' %>" href="/news-and-communications">
+        <%= t("govuk_component.government_navigation.news_and_communications", default: "News and communications") %>
```
andysellick: Where do these translations come from? I thought it was necessary to update `/config/locales/en.yml` with the text, but it doesn't seem to have the translation this replaces in it either.

bilbof: I don't know where they come from. The default seems to do the trick, and I don't have the welsh translation of this text (yet!! https://trello.com/c/ibFA5dnk/360) so didn't investigate further.

andysellick: Ah, of course, the default. Yeah, that's probably fine then. I've been putting translations into the `en.yml` file on the basis of habit, even though I've never had another language to add. I genuinely don't know what's best in this situation... ?

andysellick: I think we should add text into the language files, even if we only have the english. That way we have a central record of all required translations, so if we have the services of a translator in the future they don't need to hunt through the whole codebase trying to find all the text that needs translating, it's in one place.

PR #753, opened 2019-02-15
[DO NOT MERGE] Add an option count to the autocomplete when no facets

andysellick:
When the autocomplete is in multiple mode it appends facet tags beneath itself to show what's been selected (in single selection mode the chosen option remains in the input). 

We have a requirement to use this in finders but facet tags are already present on the page so we're using an option in the component to hide the component-generated facet tags. The result of this is a potential accessibility problem - users with assistive text may become confused when selecting an option only for their option to disappear, and this can also be a problem on mobile where the facet tags collapse beneath the facets and may not be visible on screen.

Based on the previous option select component, this PR adds a 'x selected' hint to the component when it is in this mode, as shown.

screen shot 2019-02-15 at 15 50 50

This text updates when the user makes a selection and also is set to the correct value on page load if options are pre-selected.

https://govuk-publishing-compon-pr-753.herokuapp.com/component-guide/accessible_autocomplete

Trello card: https://trello.com/c/iZxdvyMI/380-make-autocomplete-more-accessible

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Support an embed option on examples to allow embedding a component in a particular HTML context (PR #747)
+* Add an option count to the autocomplete when no facets (PR #753)
 
 ## 15.2.0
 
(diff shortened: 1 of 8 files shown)


andysellick:
@vanitabarrett thanks for the review, but we've decided not to use this anymore (sorry). Do you think it's worth keeping this functionality? I'm inclined to just close the PR.

vanitabarrett:
No problem, I'd close it 👍

PR #755, opened 2019-02-18
Update skip-link to allow custom link text

aliuk2012:
This commit ensures that the default behaviour for this component creates
a skip link to the main page content. It also allows the developer to
specify the link text should they need to have multiple skip links.

By doing this it also brings this component more inline with
GOV.UK Design System https://design-system.service.gov.uk/components/skip-link/

### DIFF
--- a/app/views/govuk_publishing_components/components/_skip_link.html.erb
+++ b/app/views/govuk_publishing_components/components/_skip_link.html.erb
@@ -1,4 +1,5 @@
 <%
   href ||= '#main-content'
+  text ||= t('components.skip_link.text')
 %>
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_skip_link.html.erb
```diff
@@ -1,4 +1,5 @@
 <%
   href ||= '#main-content'
+  text ||= 'Skip to main content'
```
andysellick: This text should be translated.

aliuk2012: Good spot 👍

PR #757, opened 2019-02-18
Add a margin_bottom option to textarea component

alex-ju:
This PR replaces a custom overwrite for the `magin-bottom` with a custom attribute to set a value from the spacing scale which seems to became a common approach in dealing with spacing for components (see #724, #727).

The reason behind this update is that textarea components should, by default, as the rest of the input form fields, have a `govuk-spacing(3)` set to `margin-bottom`.

~~Tests and changelog to be added.~~

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove classes option from components (PR #727)
+* Add margin option to textarea component (PR #757)
 
 ## 15.3.0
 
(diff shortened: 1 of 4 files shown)


alex-ju:
@andysellick tests and changelog entry added. can you give it a second review?

alex-ju:
@andysellick: I consider this PR a fix and not a breaking change as [the reason this overwrite was updated](https://github.com/alphagov/govuk_publishing_components/commit/06f585d484d99ea7edf380f3f8a4dba82418f2dc) is not aligned with the component principles and rolling back to the default space will fix how it visually looks across applications.

andysellick:
@alex-ju great, just wanted to check.

Issue #124, opened 2018-01-04
Update SCSS linting to be stricter


nickcolley:
This should align with GOV.UK Frontend project if possible.

https://github.com/alphagov/govuk_publishing_components/pull/116/files/6e58bf10b1586388a22c76d47e82d939600d0889..f5da3377ca30ddab628e72ae7d7b61876f5cbbbf#r159670908

fofr:
Should be linting by default:
https://github.com/alphagov/govuk_publishing_components/blob/master/Jenkinsfile

🤔

nickcolley:
Maybe the linting doesnt care about new lines and we could make it stricter?

alex-ju:
@nickcolley this repo is now using the same [.sass-lint](https://github.com/alphagov/govuk_publishing_components/blob/master/.sass-lint.yml) rules with GOV.UK Frontend. Are you fine with me to close this?

nickcolley:
Sure thing

PR #758, opened 2019-02-19
Add JS and SCSS lint

alex-ju:
### Changes
- set up npm-based linting using standardjs and sass-lint conform reflecting the [component conventions](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#javascript)
- update JavaScript files to conform with standardjs linting rules
- update SCSS files to conform with `sass-lint` linting rules
- update Jenkins tasks to run linting

This will align linting tools with the [govuk-frontend repo](https://github.com/alphagov/govuk-frontend).

**Note**
`scss-` rules are still preserved until `govuk-lint` is retired otherwise CI will fail

### IDE Integration
For `standardjs` there is a list on the project's website: https://standardjs.com/#are-there-text-editor-plugins

For `sass-lint` there are a couple of extensions, I'll list a couple below:
 - Atom: https://atom.io/packages/linter-sass-lint
 - Sublime: https://sublimelinter.readthedocs.io/en/stable/
 - Visual Studio Code: https://marketplace.visualstudio.com/items?itemName=glen-84.sass-lint
 - Other IDEs: https://github.com/sasstools/sass-lint#ide-integration

### Related
https://github.com/alphagov/govuk_publishing_components/issues/124

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -6,7 +6,12 @@ node {
   govuk.buildProject(
     rubyLintDiff: false,
     beforeTest: {
-      sh("npm install")
+      stage("Install npm dependencies") {
+        sh("npm install")
+      }
+      stage("Lint Javascript and SCSS") {
+        sh("npm run lint")
+      }
     }
   )
 }
--- a/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
@@ -1,78 +1,78 @@
-// This component relies on JavaScript from GOV.UK Frontend
-//= require govuk-frontend/components/checkboxes/checkboxes.js
-window.GOVUK = window.GOVUK || {};
+/* eslint-env jquery */
+// = require govuk-frontend/components/checkboxes/checkboxes.js
+window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  'use strict';
+  'use strict'
 
-   Modules.Checkboxes = function () {
+  Modules.Checkboxes = function () {
     this.start = function (scope) {
-      var _this = this;
-      this.applyAriaControlsAttributes(scope);
+      var _this = this
+      this.applyAriaControlsAttributes(scope)
 
-      $(scope).on('change', '[data-nested=true] input[type=checkbox]', function(e) {
-        var checkbox = e.target;
-        var isNested = $(checkbox).closest('.govuk-checkboxes--nested');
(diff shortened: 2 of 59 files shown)


app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
```diff
@@ -1,75 +1,74 @@
-// This component relies on JavaScript from GOV.UK Frontend
-//= require govuk-frontend/components/checkboxes/checkboxes.js
-window.GOVUK = window.GOVUK || {};
+/* global $ */
```
nickcolley: I believe you can do something like `/* eslint-env jquery */` which may be more explicit

andysellick:
Any chance you can provide some details of how to set this up locally? My editor currently has different linting rules, obviously would be good to align.

alex-ju:
@andysellick:
Standardjs: https://standardjs.com/#are-there-text-editor-plugins

Sass-lint: there are a couple of extensions, I'll list two below:
 - Atom: https://atom.io/packages/linter-sass-lint
 - Visual Studio Code: https://marketplace.visualstudio.com/items?itemName=glen-84.sass-lint

tijmenb:
This is awesome. I've raised https://github.com/alphagov/govuk-rfcs/pull/100 to talk about the entire linting system.

Jenkinsfile
```diff
@@ -6,7 +6,12 @@ node {
   govuk.buildProject(
     rubyLintDiff: false,
     beforeTest: {
-      sh("npm install")
+      stage("Install npm dependencies") {
+        sh("npm install")
+      }
+      stage("Lint Javascript and SCSS") {
+        sh("npm run lint")
```
kevindew: It looks like @tijmenb changed the linting approach yesterday on govuk-jenkinslib [yesterday](https://github.com/alphagov/govuk-jenkinslib/pull/46), so this will end up slightly weird where we have ruby and css linting reported as separate things but JS linting as part of the main build (it's an oddity we'll have on Content Publisher too). So might want to change this to run in a similar way otherwise JS linting will be set-up differently to CSS and Ruby. Might be that there is value in setting up a function in govuk-jenkinslib for linting so there's less boilerplate - I'll drop a comment on Tijmen's PR.

alex-ju: Thanks for flagging this @kevindew. I waited for an answer on [your comment](https://github.com/alphagov/govuk-jenkinslib/pull/46#issuecomment-465931313), but since nothing changed unfortunately.

I'm planning to keep the oddity of double linting and merge this PR as is because the longer we wait the harder it is to maintain it as new code comes in without following the linting rules.

Jenkinsfile and tasks can be refactored once an approach is agreed regarding linting.

Let me know if you have major concerns about it.

PR #761, opened 2019-02-22
Remove autocomplete hide facets option

andysellick:
There was an option in the autocomplete component where when it was in multiple mode you could hide the facet tags that it automatically created. This was done because in finders we already had facet tags of our own, so we didn't need duplicates.

Having investigated this in more detail, we've decided that we don't want this option anymore, as it created an accessibility issue, so we're removing it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove autocomplete hide facets option (PR #761)
 * Remove classes option from components (PR #727)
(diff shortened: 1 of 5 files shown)


andysellick:
@maxgds I don't think this represents a breaking change because I'm fairly sure it was only me using this option, but please take a look and let me know if otherwise.

maxgds:
Shouldn't have any effect on us as we aren't using multiselect at all at the moment.

PR #766, opened 2019-02-25
Add ability to add data attributes to a single radio button

1pretz1:
This change is needed for a PR soon to be opened for content-publisher
in which we need to assign some data attributes to a single radio item
instead of the whole element itself (could contain multile items).

Trello:
https://trello.com/c/pcbHvl7B/636-change-note-guidance

### DIFF
--- a/app/views/govuk_publishing_components/components/_radio.html.erb
+++ b/app/views/govuk_publishing_components/components/_radio.html.erb
@@ -71,6 +71,10 @@
 
             data_attrs = { "aria-controls": conditional_id }
             data_attrs["tracking-url"] =  item[:url] if item.key?(:url)
+
+            if item.fetch(:data_attributes, {}).any?
+              data_attrs = data_attrs.merge(item[:data_attributes])
+            end
           %>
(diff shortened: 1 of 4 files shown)


1pretz1:
@andysellick Yep sure, have done!

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -71,6 +71,10 @@
 
             data_attrs = { "aria-controls": conditional_id }
             data_attrs["tracking-url"] =  item[:url] if item.key?(:url)
+
+            if item.fetch(:data_attributes, {}).any?
```
alex-ju: Not a rails _connoisseur_, but can we use `local_assigns` instead? cc: @kevindew
I haven't seen this type of checking, it may be alright.

alex-ju: It seems like we're using this approach in the govspeak component. Seems to be working, so I'll get over it.

1pretz1: 😆👍

PR #768, opened 2019-02-25
Add a margin_bottom option to details component

alex-ju:
This PR adds custom margin to the details component.


alex-ju:
@andysellick yep, that would be great. we need it this sprint so there's not much rush with it. I'll add a DNM and wait to rebase.

andysellick:
@alex-ju have merged the shared helper so you can use it here now.

alex-ju:
@andysellick updated to use the shared helper

PR #770, opened 2019-02-26
Use delegated event handlers for checkboxes

36degrees:
Rather than setting up an event handler on every individual checkbox, set up delegated event handlers on the wrapper.

This will significantly reduce the number of event handlers being bound on pages with large numbers of checkboxes (like finder-frontend).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Enable ability to add data attributes to a single radio button (PR #766)
+* Use delegated event handlers for checkbox events (PR #770)
 
 ## 16.0.0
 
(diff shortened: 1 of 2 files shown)


36degrees:
I can't work out how to test the analytics side of this manually, but [the spec seems to cover it reasonably comprehensively](https://github.com/alphagov/govuk_publishing_components/blob/00c81f7a7b23d883e61838cec1c458726bc94c43/spec/javascripts/components/checkboxes-spec.js#L105-L135) – does this feel like a safe change?

alex-ju:
@36degrees Don't know much about how GA is configured. I think those specs are covering the change.

PR #778, opened 2019-03-05
Add modal dialogue component from content-publisher

alex-ju:
This PR moves the [modal dialogue component](https://govuk-publishing-compon-pr-778.herokuapp.com/component-guide/modal_dialogue) from [content-publisher](https://github.com/alphagov/content-publisher) to this repo so it can be shared with [content-data-admin](https://github.com/alphagov/content-data-admin).

More details on the component in the [original PR in content-publisher](https://github.com/alphagov/content-publisher/pull/735). The code has been iterated a bit since then, but the criteria remained the same.

This component is marked as experimental as more user research is needed.


andysellick:
@alex-ju would it be possible to highlight what's changed since this was approved in content publisher? Might make reviewing quicker.

alex-ju:
@andysellick: not that much it seems; after a diff:
- the JavaScript was updated to be attached to GOV.UK.Modules
- the template was updated to not require and render the title, but only the main container (which can contain the title
- no updates on the styling

app/views/govuk_publishing_components/components/docs/modal_dialogue.yml
```diff
@@ -0,0 +1,71 @@
+name: Modal dialogue
+description: Shows only one section, with no other navigation options, until the user finishes the immediate task
+body: |
+  Use where the application has gotten into a state from which it shouldn’t or can’t proceed without input from the user or the state of the current page needs to be preserved.
```
maxgds: Could we change this to:

It can be used to highlight supplemental information without taking the user off the page, or used where the application has gotten into a state from which it shouldn’t or can’t proceed without input...

Just because this fits how I want to use it and I think it's an absolutely valid use case for it.

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
```diff
@@ -0,0 +1,92 @@
...
+      '[data-toggle="modal"][data-target="' + this.$module.id + '"]'
+    )
+
+    if ($triggerElement) {
+      $triggerElement.addEventListener('click', this.$module.open)
+    }
+
+    if (this.$closeButton) {
+      this.$closeButton.addEventListener('click', this.$module.close)
+    }
```
andysellick: Much of this seems to have been written intentionally without making use of jquery, is there a reason for that?  It seems like lines 18 to 28 could be simplified if we didn't.

Also do we need to check for the existence of `$triggerElement` and `$closeButton`? Is there a scenario in which they don't exist? And even if they didn't, if we tried to attach an eventListener to them and the code can't find them, would that cause an error, or would it fail silently? (I'm genuinely not sure of the answer to that last question, so this is me wanting to find out)

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
```diff
@@ -0,0 +1,92 @@
...
+  }
+
+  ModalDialogue.prototype.handleOpen = function (event) {
+    if (event) {
+      event.preventDefault()
+    }
+
+    this.$body.classList.add('app-o-template__body--modal')
+    this.$body.classList.add('app-o-template__body--blur')
```
andysellick: Again, these two lines could be combined using jquery.

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
```diff
@@ -0,0 +1,92 @@
...
+        break
+      case KEY_ESC:
+        this.$module.close()
+        break
+      default:
+        break
+    }
+  }
+
+  Modules.ModalDialogue = ModalDialogue
```
andysellick: It looks like this code has been modified slightly to fit with the GOVUK.Modules structure, but it's still internally using the previous conventions (prototype, for example). What's the benefit of this?

My concern is that this component's JS is structured slightly differently from all the others unnecessarily (I like consistency). Obviously rewriting it at this point is a bit of a pain, but we should have a conversation about the best way of writing these modules at some point.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -0,0 +1,138 @@
...
+    width: auto;
+    max-width: $govuk-page-width * 2 / 3;
+    height: auto;
+    margin: $govuk-modal-margin auto;
+    border: $govuk-border-width-form-element solid $govuk-input-border-colour;
+  }
+
+  @include govuk-font($size: 19);
+  @include govuk-text-colour;
+  @include govuk-focusable;
```
andysellick: Personally I'd put these three `@include` lines above the `@include govuk-media-query($from: tablet) {` line, because they're effectively properties that belong with the others. But this is non-blocking.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -0,0 +1,138 @@
+$govuk-modal-margin: govuk-spacing(6);
+$govuk-modal-close-button-size: 40px;
```
andysellick: Is this big enough for touch? I vaguely remember WCAG saying it should be 44px.

maxgds: Oh good spot there, yes, it should be 44x44

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
alex-ju: Short-term I'd would like not to enforce jQuery as a dependency for projects that are do not need it. Long term I'd like to question the need of it in GOV.UK.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Oh.. the button used to have a 2px border which is now gone, that's how I ended up with this. Will update.

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
alex-ju: The main difference between using prototype vs defining methods straight in the constructor is in performance. Assigning methods within the constructor creates a new function object for every class object, using more memory and taking more processing time when calling the class constructor.

[This repository's conventions](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md#javascript) regarding JavaScript aim to use the GOV.UK Frontend conventions (I like consistency too), which means that the legacy JavaScript code should be updated. Meanwhile the JavaScript code we introduce should stick to the conventions we aim for.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Updated addressing the close button size too. This should be ready for a final review.
As a reminder, the component is marked as experimental as it's likely to be updated based on user research, but it's a good way of sharing it between projects.

PR #760, opened 2019-02-21
Remove vendored axe-core with latest axe-core via npm

nickcolley:
We have been building some pages that are in GOV.UK.

The bank holidays page throws errors for contrast so we updated the sizing of the text in our version.

I thought there might be an older version of axe-core which was not throwing this up in the component guide.

Turned out the error is only shown in Google Lighthouse (which uses axe-core) but not the official axe-core chrome extension.

So... here's a free upgrade :)

Updates axe-core v3.0.0-alpha.1 to v3.1.2

You do not have a documented (that I can see) which Node.js version you use, so I have updated the package-lock.json based on my current version: v8.9.4

PR #779, opened 2019-03-06
Update axe core

alex-ju:
This is a continuation of #760 which updates axe-core to the latest version.

### Update 1

This PR fails on Ruby 2.4.0  presumably because capybara-2.14.4 used with it doesn't support the following JavaScript features.

```
Capybara::Poltergeist::JavascriptError:
One or more errors were raised in the Javascript code on the page. If you don't care about these errors, you can ignore them by setting js_errors: false in your Poltergeist configuration (see documentation for details).

- TypeError: 'undefined' is not a function (evaluating 'window.btoa.bind(window)')
- TypeError: 'undefined' is not a function (evaluating 'Promise.resolve().then.bind(Promise.resolve())')
```

### Update 2

The PR failed because [poltergeist doesn’t support the JavaScript `bind()` function](https://github.com/teampoltergeist/poltergeist/issues/292). I added a `bind()` ‘polyfill’ in a similar fashion we're doing for `matches()` as a workaround to make the tests pass.

---

Thanks a lot to @nickcolley for the initial work on this PR.


alex-ju:
@andysellick I'm trying to dig through the documentation and see how can I get access to Heroku

alex-ju:
@andysellick heroku still fails complaining about precompiled assets although I made sure they're being specified in the same way we do for the govuk-frontend package. Will have to look more into it., I'll mark it with DNM meanwhile.

andysellick:
@alex-ju I could just check it locally :)

alex-ju:
@andysellick I'm afraid this means following PRs may fail for the same reason.

alex-ju:
managed to make it work- annoyingly heroku doesn't install the devDependencies by default

nickcolley:
Is there a way to tell heroku to install devDependencies rather than slow down the production builds for all the apps?

alex-ju:
@nickcolley currently looking into it. not familiar with how the govuk apps are being deployed on heroku, but will figure out. also I'm looking to replace poltergeist and remove the unnecessary polyfills

alex-ju:
Replaced poltergeist with govuk_test (headless_chrome) so we don't need the bind polyfill anymore. Not the problem seems to be that the way Selenium is setup in govuk_test doesn't allow for JavaScript errors (`Selenium::WebDriver::Error::JavascriptError`) to be thrown.

So now we're failing on:
```
Failure/Error:
       expect { visit '/component-guide/test-component-with-a11y-issue' }
         .to raise_error(Selenium::WebDriver::Error::JavascriptError)
     
       expected Selenium::WebDriver::Error::JavascriptError but nothing was raised
     # ./spec/component_guide/component_example_accessibility_testing_spec.rb:15:in `block (2 levels) in <top (required)>'

  2) Component example with automated testing shows accessibility violations on the page
     Failure/Error:
       expect { visit '/component-guide/test-component-with-a11y-issue' }
         .to raise_error(Selenium::WebDriver::Error::JavascriptError)
     
       expected Selenium::WebDriver::Error::JavascriptError but nothing was raised
```

If someone has ideas please leave a comment.

nickcolley:
@alex-ju it might be that https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/component_guide/accessibility-test.js#L186 now doesnt work as intended.

We are relying on poltergeist catching thrown errors so they'll fail builds.

alex-ju:
@nickcolley good hint. I've [tried to capture those using the Capybara config](https://github.com/alphagov/govuk_publishing_components/pull/779/files#diff-93830fa29d616f7c87903d08b5b1b29aR67) but I must me missing something along the way.

alex-ju:
Because of the need to raise JavaScript errors to Capybara, replacing poltergeist with govuk_test (headless_chrome) proves to be a lot harder than I imagine. I've dropped those commits and sticked with only replacing axe-core in this PR.

Regarding the `axe-core` being a devDependency, it is possible to achieve that, but I don't have permissions to reconfigure the application setup unfortunately. But this shouldn't affect other projects, as the npm installation only happens in this repo.

With these two things in mind, @andysellick @nickcolley do you think it's valuable to merge this as is, or should I close and re-address it when I have a better understanding of govuk_test?

andysellick:
@alex-ju feels like we should keep this as it's definitely a step in the right direction, even if you feel there's more to be done.

nickcolley:
I would go with putting axe-core in 'dependencies' given that it's not actually used in production builds...

alex-ju:
I'm getting to merge this as @aliuk2012 made good progress in replacing Poltergeist in a separate branch.

PR #780, opened 2019-03-08
Add shared helper info to conventions doc

andysellick:
Documenting the [shared helper code](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/shared_helper.rb).

### DIFF
--- a/docs/component_conventions.md
+++ b/docs/component_conventions.md
@@ -277,6 +277,14 @@ Code can be called and referred to in the template as follows:
 <%= card_helper.heading_tag %>
 ```
 
+### Shared helper
+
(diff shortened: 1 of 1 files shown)


docs/component_conventions.md
```diff
@@ -277,6 +277,14 @@ Code can be called and referred to in the template as follows:
 <%= card_helper.heading_tag %>
 ```
 
+### Shared helper
+
+There is also a [shared helper](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/shared_helper.rb) that can provide common functionality to all components. This includes a margin bottom option and a heading level option. See components that use the shared helper for examples of usage.
```
vanitabarrett: "See components that use the shared helper for examples of usage." - can we link to a component that uses a shared helper here?

PR #781, opened 2019-03-11
[WIP] Add GTM tracking to the Accessible autocomplete component

cbaines:
This commit extends the Accessible autocomplete component to
optionally send user input to Google via the Google Tag Manager data
layer.

This is motivated by tracking what queries users are getting no
results for, in the context of the Content Data Admin.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accessible-autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accessible-autocomplete.js
@@ -6,6 +6,59 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   "use strict";
 
+  var TRACK_QUERIES_WAIT_TIME = 500
+
+  function sourceForResults(select, trackQuery) {
+    var options = $.map(select.options, function (option) {
+      return option.textContent || option.innerText
+    })
+
+    var inputDebounceTimer = null
+
+    return function (query, populateResults) {
+      var results = options.filter(function (option) {
+        return option.toLowerCase().indexOf(query.toLowerCase()) !== -1
+      })
+
+      populateResults(results)
+
+      if (trackQuery) {
+        clearTimeout(inputDebounceTimer)
+        inputDebounceTimer = setTimeout(
+          function () {
+            trackQuery(query, results)
+          },
+          TRACK_QUERIES_WAIT_TIME
+        )
+      }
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/govuk_publishing_components/components/accessible-autocomplete.js
```diff
@@ -6,6 +6,59 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            trackQuery(query, results)
+          },
+          TRACK_QUERIES_WAIT_TIME
+        )
+      }
+    }
+  }
+
+  function trackQueriesWithGoogleTagManager (query, results) {
+    dataLayer.push({
```
andysellick: This is causing a console error if the component is used outside of the app where GTM is included.

```
Uncaught ReferenceError: dataLayer is not defined
    at trackQueriesWithGoogleTagManager
```

If we were going to go with this approach, it should be an option on the component, not the default behaviour. That would mean adding an option into the template (perhaps so a `data` attribute was added to the parent element and then having the JS check for that and behaving appropriately).

nickcolley: Agree with Andy here, I would consider using Google Analytics events directly as we have done in our existing frontend code. Whenever we implement that method we also guard the functionality to check if Google Analytics has loaded, since many people will block or not load it.

nickcolley: The Design System website loads Google Analytics via GTM, it's not something I'd personally recommend and the decision was made before I was working on the team but now that we have it setup we have not prioritised moving away from it.

On GOV.UK there is already a working Google Analytics setup, so I would avoid adding GTM in this case.

nickcolley: Also worth nothing that the Design System website has a much much smaller impact given our immediate users are designers and developers. We need to hold end-user services to a higher standard, and if the Design System website were to have the same demographic I would be pushing to remove GTM.

cbaines:
I should have closed this a little while ago, but got distracted. Anyway, I think the way forward is to make the logic here more generic, and remove the direct use of GTM, so that this is pushed back in to the applications using the component.

I'm not looking at this right now though, so I'll close the Pull Request.

PR #786, opened 2019-03-18
Suppress search components submit button 

aliuk2012:
We are using this component in finders-frontend and need to be able to suppress the submit button otherwise there will be two submit buttons in the form.

ticket: https://trello.com/c/LznDjL1r/505-all-content-update-page-heading-to-search?menu=filter&filter=label:Finders

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## UNRELEASED
 
 * Allow custom name attribute on search input (PR #787)
+* Suppress search components submit button (PR #786)
 
 ## 16.5.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -152,6 +152,10 @@ $large-input-size: 50px;
       border-right: 0;
     }
 
+    .gem-c-search__input[type="search"].gem-c-search__suppressed-button {
+      border-right: solid 1px $grey-2;
+    }
+
     .js-enabled & {
       .gem-c-search__label {
         color: $secondary-text-colour;
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -9,6 +9,7 @@
   id ||= "search-main-" + SecureRandom.hex(4)
   label_text ||= "Search GOV.UK"
   name ||= 'q'
+  hide_search_button ||= false
 %>
 
 <div class="gem-c-search <%= class_name %>" data-module="gem-toggle-input-class-on-focus">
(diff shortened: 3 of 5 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNPUBLISHED
+* Adds an option to search component to suppress the search submit button
```
andysellick: Can you change this to match the PR name and include the PR number please?

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -152,6 +152,10 @@ $large-input-size: 50px;
       border-right: 0;
     }
 
+    .gem-c-search__input[type="search"].gem-c-search__suppressed-button{
```
andysellick: I'm not sure the `[type="search"]` is needed on this line, can you check?

Also, nitpick, should be a space before the `{`.

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -8,16 +8,21 @@
   value ||= ""
   id ||= "search-main-" + SecureRandom.hex(4)
   label_text ||= "Search GOV.UK"
+  show_search_button = true if local_assigns.exclude?(:show_search_button)
```
andysellick: I'd be tempted to rewrite this as

```
hide_search_button ||= false
...

<% unless hide_search_button %>
...
```

Maybe makes it a bit cleaner and more readable?

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -8,16 +8,21 @@
...
     <%= label_text %>
   </label>
   <div class="gem-c-search__item-wrapper">
-    <input type="search" value="<%= value %>" id="<%= id %>" name="q" title="Search" class="gem-c-search__item gem-c-search__input js-class-toggle">
-    <div class="gem-c-search__item gem-c-search__submit-wrapper">
-      <button type="submit" class="gem-c-search__submit">Search</button>
-    </div>
+    <input type="search" value="<%= value %>"
+      id="<%= id %>" name="q" title="Search"
+      class="gem-c-search__item gem-c-search__input js-class-toggle<%= ' gem-c-search__suppressed-button' if show_search_button == false %>">
```
andysellick: If we changed the variable to `hide_search_button` then this line could be

```
class="gem-c-search__item gem-c-search__input js-class-toggle<%= ' gem-c-search__suppressed-button' if hide_search_button %>">
```

app/views/govuk_publishing_components/components/docs/search.yml
```diff
@@ -38,3 +38,7 @@ examples:
   large_version:
     data:
       size: "large"
+  remove_search_submit_button:
+      description: Sometimes this component maybe used inside an existing form which has another submit button. A form should only have one submit button. The search button will be displayed by default.
```
andysellick: Really tiny nitpick: I think this should be `this component may be used`.

app/views/govuk_publishing_components/components/_search.html.erb
aliuk2012: Yeah, I wasn't entirely happy with my approach because I tried `show_search_button ||= true` and this failed when I tried to set the variable to false.

I prefer your example.

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
aliuk2012: I was trying to be consistent with the selector based on this comment: https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_search.scss#L18

Any thoughts?

PR #787, opened 2019-03-18
Allow custom name attribute on search input

aliuk2012:
Not all use cases will need to use 'q' but to ensure backwards compatiblity
I have made it the default.

In finder-frontend we use 'keyword' as the field name.

Ticket: https://trello.com/c/LznDjL1r/505-all-content-update-page-heading-to-search

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNRELEASED
+
+* Allow custom name attribute on search input (PR #787)
+
 ## 16.5.0
 
 * Change consultations header link to link to new finders (PR #783)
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNRELEASED
+
+* Adds option to search component to allow custom name attributes on the input field
```
andysellick: Can you update this to match the PR name and include the PR number please? i.e. `Allow custom name attribute on search input (PR #787)`

spec/components/search_spec.rb
```diff
@@ -42,4 +42,14 @@ def component_name
     render_component(size: "large")
     assert_select ".gem-c-search.gem-c-search--large"
   end
+
+  it "renders a search box with a default name" do
+    render_component(id: "my-unique-id")
```
andysellick: I was briefly confused by this, although it makes sense. Maybe (if possible) have this test pass no parameters, to make it clear?

PR #790, opened 2019-03-19
Add custom width to inputs

alex-ju:
This PR adds a `width` option to the input component, which allows to set a specific width for an text input based on the number characters that are expected to be inserted. This feature is added primarily as a preparation for the date-input component which will use the input component within.

Currently, the supported values (defined in govuk-frontend) are 2, 3, 4, 5, 10, 20, 30 which implies that the bigger value should be used in case the expected length is between these values.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add width option to input component (PR #790)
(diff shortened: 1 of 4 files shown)


alex-ju:
@andysellick extra test added for unsupported width value and changelog updated. I was thinking about adding it to the shared helper, but currently it's the only component that supports custom width, so I'll wait until we'll have a new one with this need.

PR #791, opened 2019-03-19
Update fieldset component to use styles from govuk-frontend

alex-ju:
This update is done primarily as a preparation for the date-input component which will use the fieldset component within it, but also removes duplicated code between repositories.

The code we're now importing from govuk-frontend contains the same code that this PR removes from this repo (presumably it originated from Elements): https://github.com/alphagov/govuk-frontend/blob/master/src/components/fieldset/_fieldset.scss#L13

### Impact
Applications using this component:
- email-alert-frontend (2 occurrences: [new_frequency](https://github.com/alphagov/email-alert-frontend/blob/b2592b94dde594c4d9ef1385bdffa55ac975113a/app/views/subscriptions/new_frequency.html.erb#L17) and [update_frequency](https://github.com/alphagov/email-alert-frontend/blob/b2592b94dde594c4d9ef1385bdffa55ac975113a/app/views/subscriptions_management/update_frequency.html.erb#L15))
- government-frontend (1 occurrence: [service_sign_in](https://github.com/alphagov/government-frontend/blob/5e91bdca894ebe732cda3d543295803229f723a9/app/views/content_items/service_sign_in/_choose_sign_in.html.erb#L26))

### Before
Screen Shot 2019-03-19 at 13 50 41
Screen Shot 2019-03-19 at 13 50 50

### After
Screen Shot 2019-03-19 at 13 50 14
Screen Shot 2019-03-19 at 13 50 27


app/views/govuk_publishing_components/components/_fieldset.html.erb
```diff
@@ -1,6 +1,6 @@
 <% text = text || yield %>
-<fieldset class="gem-c-fieldset">
-  <legend class="gem-c-fieldset__legend">
+<fieldset class="gem-c-fieldset govuk-fieldset">
+  <legend class="gem-c-fieldset__legend govuk-fieldset__legend">
```
andysellick: Probably don't need the `gem-c-fieldset__legend` class here anymore?

app/views/govuk_publishing_components/components/_fieldset.html.erb
```diff
@@ -1,6 +1,6 @@
 <% text = text || yield %>
-<fieldset class="gem-c-fieldset">
-  <legend class="gem-c-fieldset__legend">
+<fieldset class="gem-c-fieldset govuk-fieldset">
```
andysellick: At some point we should talk about the merit of keeping `gem-c-` classes on components if they don't actually style anything. Although I guess it's needed for the browser extension to recognise them?

alex-ju: Yep, we currently only keep them for the browser extension. We could think about removing them in-sync with an update on the extension.

PR #792, opened 2019-03-19
Add date input component

alex-ju:
This PR
- adds the [date input component](https://govuk-publishing-compon-pr-792.herokuapp.com/component-guide/date_input) from `govuk-frontend` and
  - updates the input component to allow `error_style` without an `error_message` (so that inputs can be styles with a thick red box without requiring an error message attached to them)
  - updates the fieldset component to support `describedby` and `role` attributes required for accessibility enhancements of the date-input component.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
@@ -1,1 +1,15 @@
 @import "govuk-frontend/components/date-input/date-input";
+
+// Add spacing between input items on narrow containers when they collapse under each other
+.govuk-date-input {
+  margin-top: - govuk-spacing(2);
+}
+
+.govuk-date-input__item {
+  margin-top: govuk-spacing(2);
+}
+
+// Remove spacing for the last input item to make the most of space when in narrow containers
+.govuk-date-input__item:last-child {
+  margin-right: 0;
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
@@ -1,0 +1,1 @@
+@import "govuk-frontend/components/date-input/date-input";
--- a/app/views/govuk_publishing_components/components/_date_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_date_input.html.erb
@@ -54,7 +54,8 @@
             label: {
               text: item[:label] || item[:name].capitalize
             },
-            error_style: error_message,
+            grouped: true,
+            has_error: error_message,
             name: name ? (name + "[" + item[:name] + "]") : item[:name],
             value: item[:value],
             width: item[:width],
--- a/app/views/govuk_publishing_components/components/_date_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_date_input.html.erb
@@ -44,7 +44,7 @@
           label: {
             text: item[:name].capitalize
           },
-          has_error: has_error,
+          error_style: has_error,
           name: item[:name],
           value: item[:value],
           width: item[:width],
--- a/app/views/govuk_publishing_components/components/_date_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_date_input.html.erb
@@ -12,20 +12,18 @@
   error_message ||= nil
   describedby ||= nil
 
-  has_error = error_message
-
   css_classes = %w(gem-c-date-input govuk-date-input)
   form_group_css_classes = %w(govuk-form-group)
-  form_group_css_classes << "govuk-form-group--error" if has_error
+  form_group_css_classes << "govuk-form-group--error" if error_message
 
   hint_id = "hint-#{SecureRandom.hex(4)}"
   error_id = "error-#{SecureRandom.hex(4)}"
 
   aria_described_by ||= nil
-  if hint || has_error || describedby
+  if hint || error_message || describedby
     aria_described_by = []
     aria_described_by << hint_id if hint
-    aria_described_by << error_id if has_error
+    aria_described_by << error_id if error_message
     aria_described_by << describedby if describedby
     aria_described_by = aria_described_by.join(" ")
   end
@@ -56,7 +54,7 @@
             label: {
               text: item[:label] || item[:name].capitalize
             },
-            error_style: has_error,
+            error_style: error_message,
             name: name ? (name + "[" + item[:name] + "]") : item[:name],
             value: item[:value],
             width: item[:width],
--- a/app/views/govuk_publishing_components/components/_date_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_date_input.html.erb
@@ -4,13 +4,26 @@
   legend_text ||= nil
   hint ||= nil
   error_message ||= nil
+  describedby ||= nil
 
   has_error = error_message
 
   css_classes = %w(gem-c-date-input govuk-date-input)
-  css_classes << "govuk-input--error" if has_error
   form_group_css_classes = %w(govuk-form-group)
   form_group_css_classes << "govuk-form-group--error" if has_error
+
+  hint_id = "hint-#{SecureRandom.hex(4)}"
+  error_id = "error-#{SecureRandom.hex(4)}"
+
+  aria_described_by ||= nil
+  if hint || has_error || describedby
+    aria_described_by = []
+    aria_described_by << hint_id if hint
+    aria_described_by << error_id if has_error
+    aria_described_by << describedby if describedby
+    aria_described_by = aria_described_by.join(" ")
+  end
+
 %>
 
 <% date_input_items = [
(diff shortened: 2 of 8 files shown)


andysellick:
@alex-ju I haven't done a proper review of this yet but I've raised an issue I noticed that relates to this component in the design system - https://github.com/alphagov/govuk-frontend/issues/1250.

alex-ju:
@andysellick: I'll try a fix. Thanks for raising it, I wasn't aware of this behaviour.

alex-ju:
@andysellick pushed a fix that removes the margin-right on the last element and deal with the margin when they collapse while preserving the height of the component (and implicitly the border-right when we have an error state)

Kapture 2019-03-20 at 16 48 27

andysellick:
@alex-ju very nice. That's exactly how I'd have fixed it 👍

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,85 @@
...
+    aria_described_by = []
+    aria_described_by << hint_id if hint
+    aria_described_by << error_id if has_error
+    aria_described_by << describedby if describedby
+    aria_described_by = aria_described_by.join(" ")
+  end
+
+%>
+
+<% date_input_items = items ? items : date_input_items = [
```
benthorner: This is tricky to understand, as contains two assignments. Could this just be the default for `items`?

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,85 @@
...
+        text: error_message
+      } %>
+    <% end %>
+
+    <%= tag.div class: css_classes, id: id do %>
+      <% date_input_items.each do |item| %>
+        <%= tag.div class: "govuk-date-input__item" do %>
+          <%= render "govuk_publishing_components/components/input", {
+            label: {
+              text: item[:label] || item[:name].capitalize
```
benthorner: What's the use case for re-using the name of the field?

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,85 @@
...
+            },
+            error_style: has_error,
+            name: name ? (name + "[" + item[:name] + "]") : item[:name],
```
benthorner: Could we do this automatically for the defaults? Since overriding the items requires you to specify a name anyway, I think we can express this helpful behaviour less magically :-).

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,85 @@
...
+            value: item[:value],
+            width: item[:width],
+            id: item[:id],
+            type: "number",
+            data: item[:data],
```
benthorner: Minor, but I remember you saying we prefer `data_attributes`...unless we don't?

alex-ju:
@andysellick are you happy to review this one when you have time? since you're planning to use it too in one of the finders

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -21,7 +22,7 @@
   error_id = "error-#{SecureRandom.hex(4)}"
 
   css_classes = %w(gem-c-input govuk-input)
-  css_classes << "govuk-input--error" if has_error
+  css_classes << "govuk-input--error" if has_error || error_style
```
benthorner: It looks like we can already override the `has_error` attribute on the input component, so do we need the extra attribute? Minor, but if we keep it, I was expecting it to contain some classes or something :-|.

app/views/govuk_publishing_components/components/_date_input.html.erb
alex-ju: We do, but input component has `data`.. so we're inconsistent anyway. Happy to make it `data_attributes `

app/views/govuk_publishing_components/components/_input.html.erb
alex-ju: How can we override the `has_error`? happy to change the name with a better proposal.

benthorner: I noticed it's an undocumented parameter of the input component :-).

alex-ju: Yes, but isn't it overwritten on [L20](https://github.com/alphagov/govuk_publishing_components/blob/337e3f02ff232fc3d2682438932826388a863d34/app/views/govuk_publishing_components/components/_input.html.erb#L20)?

app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
```diff
@@ -0,0 +1,13 @@
+@import "govuk-frontend/components/date-input/date-input";
+
+.govuk-date-input {
```
nickcolley: Worth a comment explaining why you're overwriting the upstream component?

app/views/govuk_publishing_components/components/_input.html.erb
benthorner: Oh yeah, sorry. But instead of adding a new param, you could just add a `||`, since `has_error` is a pretty good name :-).

app/assets/stylesheets/govuk_publishing_components/components/_date-input.scss
alex-ju: Yep, will add. I'm not happy that I'm overwriting using the original classes, but there's not way to pass classes in this lib, so I don't see another way.

app/views/govuk_publishing_components/components/_input.html.erb
alex-ju: ok, will re-use it then.

alex-ju: Gave it a try, but we're using `has_error` for more than styling an input (which is what we want in this case); we use it to generate the red border on form group and attach an ID to the aria describedby attribute. So not really an option to recycle it.

app/views/govuk_publishing_components/components/_date_input.html.erb
alex-ju: The name and label are usually the same so we're doing this to avoid having to define both (for our default values or for custom values)

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -0,0 +1,34 @@
...
+  Inputs in the component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * be usable with touch
+  * indicate when they have focus
+  * be recognisable as form input elements
+  * have correctly associated labels
+  * be of the appropriate type for their use, e.g. password inputs should be of type 'password'
```
andysellick: This looks like it was copied from somewhere else, can we be more specific? e.g. `be of the appropriate type for their use, in this case 'number'`

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -0,0 +1,54 @@
...
+  Labels use the [label component](/component-guide/label).
+
+  Avoid using autofocus and tabindex unless you have user research to support this behaviour.
+govuk_frontend_components:
+  - date-input
+examples:
+  default:
+    data: {}
+  with_name:
+    data:
```
andysellick: Is it worth adding a short description here to explain that the name is actually set to `name[field name]`?

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,64 @@
+<%
+  id ||= "input-#{SecureRandom.hex(4)}"
+
+  legend_text ||= nil
+  hint ||= nil
+  error_message ||= nil
+
+  has_error = error_message
+
+  css_classes = %w(gem-c-date-input govuk-date-input)
```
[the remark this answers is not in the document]
alex-ju: All form components (input, textarea, etc) don't have the component class (`gem-c-`) at the wrapper level (form group level), but at the component level. The reason behind this is that the markup for these components are a composite of a hint component, and error message component and the component itself wrapped within a group (which can arguably be a component on its own).

app/views/govuk_publishing_components/components/docs/date_input.yml
alex-ju: Yep, copied from input component, will update with number as you stated above.

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -0,0 +1,64 @@
+<%
+  id ||= "input-#{SecureRandom.hex(4)}"
+
+  legend_text ||= nil
+  hint ||= nil
+  error_message ||= nil
+
+  has_error = error_message
```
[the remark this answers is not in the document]
alex-ju: Fixed

app/views/govuk_publishing_components/components/docs/date_input.yml
alex-ju: Good point. Description added explaining how we use this attribute.

alex-ju:
@andysellick thanks for the review. updated as per comments. this can be reviewed again when you have time.

PR #795, opened 2019-03-21
Add cookie banner component

alex-ju:
## Need
We need to update the cookie banner which is currently part of govuk_template. There was [an attempt to do this update in govuk_template](https://github.com/alphagov/govuk_template/pull/368), but since that's a deprecated repository we shouldn't do any new work or releases on it.

Instead, this PR moves [the remaining `govuk-template.js` scripts](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js) into `app/assets/javascripts/govuk_publishing_components/lib/` and extracts the cookie banner into its own component.

## Complementary work
A follow up PR will come after this, updating static to remove `govuk-template.js` and the associated markup for the cookie banner and replace it with this component.

Update: PR in static https://github.com/alphagov/static/pull/1662

[Trello card](https://trello.com/c/y8JyxGL9)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
@@ -1,0 +1,65 @@
+// used by the cookie banner component
+
+(function () {
+  'use strict'
+  var root = this
+  if (typeof root.GOVUK === 'undefined') { root.GOVUK = {} }
+
+  /*
+    Cookie methods
+    ==============
+
+    Usage:
+
+      Setting a cookie:
+      GOVUK.cookie('hobnob', 'tasty', { days: 30 });
+
+      Reading a cookie:
+      GOVUK.cookie('hobnob');
+
+      Deleting a cookie:
+      GOVUK.cookie('hobnob', null);
+  */
+  window.GOVUK.cookie = function (name, value, options) {
+    if (typeof value !== 'undefined') {
+      if (value === false || value === null) {
+        return window.GOVUK.setCookie(name, '', { days: -1 })
+      } else {
+        return window.GOVUK.setCookie(name, value, options)
+      }
+    } else {
+      return window.GOVUK.getCookie(name)
+    }
+  }
+
+  window.GOVUK.setCookie = function (name, value, options) {
+    if (typeof options === 'undefined') {
+      options = {}
+    }
+    var cookieString = name + '=' + value + '; path=/'
+    if (options.days) {
+      var date = new Date()
+      date.setTime(date.getTime() + (options.days * 24 * 60 * 60 * 1000))
+      cookieString = cookieString + '; expires=' + date.toGMTString()
+    }
+    if (document.location.protocol === 'https:') {
+      cookieString = cookieString + '; Secure'
+    }
+    document.cookie = cookieString
+  }
+
+  window.GOVUK.getCookie = function (name) {
+    var nameEQ = name + '='
+    var cookies = document.cookie.split(';')
+    for (var i = 0, len = cookies.length; i < len; i++) {
+      var cookie = cookies[i]
+      while (cookie.charAt(0) === ' ') {
+        cookie = cookie.substring(1, cookie.length)
+      }
+      if (cookie.indexOf(nameEQ) === 0) {
+        return decodeURIComponent(cookie.substring(nameEQ.length))
+      }
+    }
+    return null
+  }
+}).call(this)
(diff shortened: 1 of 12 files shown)


app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -0,0 +1,65 @@
+// used by the cookie banner component
```
andysellick: What's the reason for keeping this separate from the component JS? Is something else using it?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -0,0 +1,19 @@
+$govuk-cookie-banner-background: #d5e8f3;
```
andysellick: Should we use a variable for this from GOVUK Frontend? Or is it a unique colour?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -0,0 +1,19 @@
...
+
+.js-enabled {
+  .gem-c-cookie-banner {
+    display: none; // shown with JS, always on for non-JS
+  }
+}
+
+.gem-c-cookie-banner {
+  @include govuk-font($size: 16);
+  width: 100%;
```
andysellick: Setting width seems unnecessary for a block level element, or does it behave differently when in position?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -0,0 +1,8 @@
+<% message ||= capture do %>
+  GOV.UK uses cookies to make the site simpler. <a class="govuk-link" href="https://www.gov.uk/help/cookies" data-module="track-click" data-track-category="cookieBanner" data-track-action="Cookie banner clicked">Find out more about cookies</a> or <a class="govuk-link" href="#" data-hide-cookie-banner="true" data-module="track-click" data-track-category="cookieBanner" data-track-action="Cookie banner hidden">hide this message</a>
+<% end %>
+<% id ||= 'global-cookie-message' %>
```
andysellick: Could this line of code be included with the lines above? e.g.

```
<% 
  message ||= capture do
    GOV.UK uses cookies...
  end

%>
```

app/views/govuk_publishing_components/components/docs/cookie_banner.yml
```diff
@@ -0,0 +1,16 @@
+name: Cookie banner
+description: Help users manage their personal data by telling them when you store cookies on their device.
+body: |
+  Setting `data-hide-cookie-banner="true"` on any link inside the banner will overwrite the default action and when clicked will dismiss the cookie banner for a period of 365 days (approx. 1 year).
+accessibility_criteria: |
+  - accept focus
```
andysellick: I think this is missing `Links in the component must...`. 

Also is there anything we can add about the visibility and readability of the cookie banner? I've seen plenty of confusing ones and badly worded language can be a barrier. Something like `text in the banner must be clear and unambiguous`?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
alex-ju: Yep, carried over from the other repo. Good spot.

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
alex-ju: Unique colour unfortunately..

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
alex-ju: I don't think it works this day as the message includes html tags. Will swap them and use a single block.

andysellick: Oh, of course. Never mind then.

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
alex-ju: I made the assumption that it makes more sense to preserve this as a common lib than add it to the banner and then extract it later on when needed. The code looks very generic to be embedded in the cookie banner and there are other potential usages although we avoid storing cookies as much as possible.

alex-ju:
Thanks for the review @andysellick! I addressed all the comments including tests for the cookie banner visibility and the cookie state.

fofr:
Can I ask why a "hide this message" link has been added?

alex-ju:
This update is part of a GOV.UK Cookie & Tracking Policy mission in an effort to align it with the spirit of the EU cookie and tracking policy legislation. To play back team's PMs reply to this question: "it’s a temporary change to increase visibility of the message while we consider other options in the next few months. that work will include looking at how this could be expanded across government."

PR #796, opened 2019-03-22
Add tracking to toggle.js

karlbaker02:
This PR adds tracking to toggle.js, which enables analytics to be fired the `$toggle` element if it is set up with `data-track-category` and `data-track-action` attributes.

### DIFF
--- a/spec/javascripts/components/toggle-spec.js
+++ b/spec/javascripts/components/toggle-spec.js
@@ -4,6 +4,12 @@ describe('A toggle module', function () {
   var toggle,
     element;
 
+  var GOVUK = window.GOVUK || {};
+
+  GOVUK.analytics = {
+    trackEvent: function () {}
+  };
+
   beforeEach(function () {
     toggle = new GOVUK.Modules.GemToggle();
   });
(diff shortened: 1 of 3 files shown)


spec/javascripts/components/toggle-spec.js
```diff
@@ -59,6 +65,39 @@ describe('A toggle module', function () {
...
     });
+
+    it('does not track when element is not trackable', function() {
+      spyOn(GOVUK.analytics, 'trackEvent');
+
+      element = $('\
+        <div>\
+          <a href="#" class="my-toggle" data-expanded="false" data-controls="target" data-toggled-text="Show fewer">Toggle</a>\
+          <div id="target" class="js-hidden">Target</div>\
+        </div>');
```
andysellick: This seems to be creating an element similar to the one already created in the `beforeEach` function (which has also been modified in this PR). Do we have confidence that this test only has one toggle element?

karlbaker02: Yes, because we're overriding `element` and creating a ` new GOVUK.Modules.GemToggle()` that is overriding `toggle`. Happy to add another clause in there to prove this if you feel it's required.

karlbaker02:
Updating documentation now, thanks for pointing that out @andysellick

spec/javascripts/components/toggle-spec.js
andysellick: Oh I see now. Thanks.

PR #802, opened 2019-03-26
Release v16.8.0

alex-ju:
## 16.8.0
* Add date input component based on GOV.UK Frontend (PR #792)
* Add cookie banner component and migrate the remaining [govuk-template.js](https://github.com/alphagov/govuk_template/blob/master/source/assets/javascripts/govuk-template.js) scripts from [govuk_template](https://github.com/alphagov/govuk_template/) (PR #795)


CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 16.8.0
 * Add date input component based on GOV.UK Frontend (PR #792)
```
andysellick: Nitpick: we seem to have lost the newline after the release heading for this and the last release.

alex-ju: Good catch, must have got lost in rebase. Updated.

alex-ju:
@andysellick we usually run `bundle` when we do a release and (as far as I understand) one of the desired effects of it is to make sure we're updating the dependency tree - and that's what I've got :)

andysellick:
@alex-ju sorry, I wasn't aware that was the procedure. No worries then.

alex-ju:
@andysellick: it seems like we don't have the release process documented anywhere, which is a common question about this repo, so I'll make a note to add one.

PR #803, opened 2019-03-27
Adds small form option to subscription component

aliuk2012:
In finder-frontend, we needed the subscriptions component to be less "shouty" and be less prominent on the page. The component is used in three other repositories. This PR creates a 'small form' version by making the font smaller and not bold.

- updates css to use govuk-frontend mixins
- add small form option and styling
- position icons closer to the text

Component: https://govuk-publishing-compon-pr-803.herokuapp.com/component-guide/subscription-links

Demo: https://govuk-publishing-compon-pr-803.herokuapp.com/component-guide/subscription-links/as_small_form

Ticket: https://trello.com/c/nMB9EUx0/581-update-top-section-of-results-layout

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 ## Unreleased
 
 * Control when checkboxes change event sends Google Analytics tracking info (PR #801)
+* Updates govuk-frontend from 2.5.1 to 2.9.0 (PR #794)
+* Adds small form option to subscription component (PR #803)
 
 ## 16.8.0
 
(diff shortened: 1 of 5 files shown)


andysellick:
I'm a bit late to this party, but wondered if you'd considered underline for these smaller, non-bold links?

aliuk2012:
@andysellick thanks for your comment, I haven't considered it but I think they look actionable enough without the underlines (icons and text). I ran the idea passed Joe and he felt the same.

I think we should leave it as is and if need be we can easily add it.

image

andysellick:
That looks way better with the underline.

PR #804, opened 2019-03-27
Suppress subscription components heading

aliuk2012:
For some unknown reason this component includes a h2 heading by default.
This means that if a service uses it unexpectedly it will include the heading.
It also stops a team from using this component anywhere else on a page.

By using the 'hide_heading' option, it will suppress the heading when
the component is being rendered.

Component: https://govuk-publishing-compon-pr-804.herokuapp.com/component-guide/subscription-links

Demo: https://govuk-publishing-compon-pr-804.herokuapp.com/component-guide/subscription-links/without_heading

Ticket: https://trello.com/c/nMB9EUx0/581-update-top-section-of-results-layout

### DIFF
--- a/app/views/govuk_publishing_components/components/_subscription-links.html.erb
+++ b/app/views/govuk_publishing_components/components/_subscription-links.html.erb
@@ -11,10 +11,13 @@
   css_classes << (shared_helper.get_margin_bottom) unless local_assigns[:margin_bottom] == 0
   css_classes << brand_helper.brand_class
   data = {"module": "gem-toggle"} if sl_helper.feed_link_box_value
+  hide_heading ||= false
 %>
 <% if sl_helper.component_data_is_valid? %>
   <%= tag.section class: css_classes, data: data do %>
-    <h2 class="gem-c-subscription-links__hidden-header visuallyhidden"><%= t("govuk_component.subscription_links.subscriptions", default: "Subscriptions") %></h2>
+    <% unless hide_heading %>
+      <h2 class="gem-c-subscription-links__hidden-header visuallyhidden"><%= t("govuk_component.subscription_links.subscriptions", default: "Subscriptions") %></h2>
+    <% end %>
     <ul
       class="gem-c-subscription-links__list<%= ' gem-c-subscription-links__list--small' if local_assigns[:small_form] == true %>"
       <%= "data-module=track-click" if sl_helper.tracking_is_present? %>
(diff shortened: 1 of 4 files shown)


aliuk2012:
I plan to do a release once this has been approved & merged

app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -11,10 +11,13 @@
   css_classes << (shared_helper.get_margin_bottom) unless local_assigns[:margin_bottom] == 0
   css_classes << brand_helper.brand_class
   data = {"module": "gem-toggle"} if sl_helper.feed_link_box_value
+  hide_heading ||= false
```
alex-ju: I was wondering if we could turn this into a positive flag `show_heading ||= true` which can be turned off. It's a matter of flavour, but I always find confusing setting things to true to disable something. Especially because we don't know why we set this `<h2>` implicitly on this component we may consider dropping it at some point.

aliuk2012:
@andysellick Good idea, I've added a some copy. I'll merge it now.

alex-ju:
@aliuk2012 would be great if we can get #805 in the same release

app/views/govuk_publishing_components/components/_subscription-links.html.erb
aliuk2012: I've tried using positive flags before but it doesn't work with the way we set the defaults using `||=` operator.  

TBH, I would have liked to have simply remove the h2 but I just don't have time to update and test the other repos.

andysellick: It could be a positive flag, but we'd have to go and change every existing use of the component.

alex-ju: I'm not very opinionated, whatever does the job 👍

andysellick: @alex-ju that's the problem Al was talking about. If you set the default to true, like this: `show_heading ||= true` when you pass through 'false' it defaults to true.

alex-ju: yep, familiar with this situation, but we dealt with it in [other components with a conditional statement](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_tabs.html.erb#L3).

PR #805, opened 2019-03-27
 Update the modal component to prevent the page from scrolling behind it when open

alex-ju:
govuk-frontend v2.8.0 introduces an `overflow-y: scroll` for `.govuk-template` that needs to be overwritten when the modal opens otherwise the content will scroll behind the modal. https://github.com/alphagov/govuk-frontend/pull/1230

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
@@ -72,11 +72,17 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
   touch-action: none;
 }
 
-.app-o-template__body--modal {
+.gem-o-template--modal {
+  @include govuk-media-query($media-type: screen) {
+    overflow-y: inherit;
+  }
+}
+
+.gem-o-template__body--modal {
   overflow: hidden;
 }
 
-.app-o-template__body--blur {
+.gem-o-template__body--blur {
   .govuk-skip-link,
   .govuk-header,
   .govuk-phase-banner,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Updates govuk-frontend from 2.5.1 to 2.9.0 (PR #794)
 * Adds small form option to subscription component (PR #803)
 * Suppress subscription components heading (PR #804)
+* Update the modal component to prevent the page from scrolling behind it when open (PR #805)
 
 ## 16.8.0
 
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -72,11 +72,17 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
   touch-action: none;
 }
 
-.app-o-template__body--modal {
+.gem-o-template--modal {
+  @include govuk-media-query($media-type: screen) {
+    overflow-y: inherit;
```
andysellick: I'm not sure I understand this change. Why not just put `overflow: hidden` here? From playing around with it, it seems that you wouldn't then need to set hidden on the body element. Or am I missing something?

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -72,11 +72,17 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
...
+  }
+}
+
+.gem-o-template__body--modal {
```
andysellick: I'm sure this used to be documented somewhere but I can't find it. Apologies - remind me what `gem-o` relates to, and why it isn't `gem-c`?

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -72,11 +72,17 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
   touch-action: none;
 }
 
-.app-o-template__body--modal {
+.gem-o-template--modal {
+  @include govuk-media-query($media-type: screen) {
```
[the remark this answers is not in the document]
alex-ju: As mentioned in the PR details, this is undoing [the change added to the govuk-frontend in #1230](https://github.com/alphagov/govuk-frontend/pull/1230) for when the modal is open, hence we use the same media query and the and an overflow reset.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Indeed the overflow: hidden property could be in theory set to either `<html>` or `<body>`. I personally prefer not to add styles to the html element unless absolutely necessary because it is a special element, that's why I decided in the initial work to set it to `<body>` not knowing the overflow of the `<html>` element will be changed.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Following the ITCSS model, I consider the `template` component being an object, not a component that's why I used a different prefix `gem-o`. Since we generally only use `gem-c` across our shared components, it would be good to question the need of type `-c-` prefix.

alex-ju:
@andysellick are you happy with me to merge this or shall we have a chat about it beforehand?

andysellick:
@alex-ju thanks for asking. I think I'm happy with this change, although I'm still nervous about the non-isolation aspects of the component as a whole. But, this is something we need, so that's fine.

alex-ju:
@andysellick I share your worry about isolation. It's the only solution I've seen at the moment to deal with the blur effect. I'm looking into removing it if I find a good alternative that doesn't involve 'communicating' with the template.

PR #801, opened 2019-03-26
Control when checkboxes change event sends tracking info

aliuk2012:
There are some cases where you might not want to track the change to the
checkboxes. For example, an event being triggered from another component
which unticks/ticks a checkbox like finder-frontend has. Facet tags
when clicked untickes the relevant checkbox and tracks the change.

https://trello.com/c/7vQb8Yww/434-bug-fix-issue-where-use-of-brexit-checkbox-fires-multiple-ga-events

Will be needed for https://github.com/alphagov/finder-frontend/pull/984

PR #806, opened 2019-03-28
Release v16.8.0

aliuk2012:
## Changelog

* Control when checkboxes change event sends Google Analytics tracking info (PR #801)
* Updates govuk-frontend from 2.5.1 to 2.9.0 (PR #794)
* Adds small form option to subscription component (PR #803)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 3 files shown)


aliuk2012:
@andysellick I spoke with Alex earlier and it shouldn't be a problem as only one or two services are using the modal. They are going to wait until modal fix before upgrading their gem version.

PR #808, opened 2019-03-29
fix margin spacing for small form subscription links

aliuk2012:
there no need for the far left/right margin to be applied to small form subscription links. If space is needed it can be done by wrapping an element around this component and applying a maring left/right to it


app/assets/stylesheets/govuk_publishing_components/components/_subscription-links.scss
```diff
@@ -16,13 +16,31 @@
     @include govuk-font(16);
   }
 
+
```
vanitabarrett: Stray new lines

PR #811, opened 2019-04-04
Bump patch version 16.9.2

sihugh:


alex-ju:
@sihugh may I ask for a correction in the changelog file before releasing. The entry below should be part of this release -it was merged recently without rebasing which resulted in it being added to 16.9.0 in the changelog file. It's also a fix, so it fits with this patch.

Remove
> Update the modal component to prevent the page from scrolling behind it when open (PR #805)

Add
> Fix the modal component to prevent the page from scrolling behind it when open (PR #805)

PR #817, opened 2019-04-12
Upgrade to govuk-frontend 2.10.0

andysellick:
This PR:

- updates this repo to govuk-frontend version 2.10.0
- updates the checkboxes component to have a 'small' option
- updates the radio component to have a 'small' option

[See detailed changelog](https://github.com/alphagov/govuk-frontend/releases/tag/v2.10.0)

Links:

- [small checkboxes](https://govuk-publishing-compon-pr-817.herokuapp.com/component-guide/checkboxes/with_small_checkboxes)
- [small radios](https://govuk-publishing-compon-pr-817.herokuapp.com/component-guide/radio/with_small_radios)

Trello card: https://trello.com/c/wp21yw0C/635-use-small-radios-and-checkboxes-in-finder-frontend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+- Upgrade to govuk-frontend 2.10.0 (PR #817)
+
 ## 16.10.1
 
 - Enforce compatibility with deprecated packages for media print stylesheet (PR #815)
--- a/app/views/govuk_publishing_components/components/_checkboxes.html.erb
+++ b/app/views/govuk_publishing_components/components/_checkboxes.html.erb
@@ -1,4 +1,5 @@
 <%
+  # cb_helper.css_classes generates "gem-c-checkboxes"
   cb_helper = GovukPublishingComponents::Presenters::CheckboxesHelper.new(local_assigns)
   id = cb_helper.id
 %>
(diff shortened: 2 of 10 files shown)


spec/components/all_components_spec.rb
```diff
@@ -22,7 +22,7 @@
       end
 
       it "has the correct class in the ERB template",
-        skip: component_name.in?(%w[step_by_step_nav_related step_by_step_nav_header step_by_step_nav previous_and_next_navigation]),
+        skip: component_name.in?(%w[step_by_step_nav_related step_by_step_nav_header step_by_step_nav previous_and_next_navigation checkboxes]),
```
alex-ju: We shouldn't need to skip this one.

andysellick: Yes, how do we do this? I wasn't familiar with this test. It looks like it's checking to see there's a correctly formatted class name for the component in the template, but in this case we're generating lists of classes and passing them into the template from the helper. I could move some of that code into the template, but it seems unnecessary? I'm not sure why this test is useful - it looks like I could put the class name in a comment and it would pass.

alex-ju: Well.. I wasn't familiar either, but I guess it's our way of making sure we're not missing these 'tag'/'selector' classes when we create a new component or update classes on another. Ideally we would check the generated markup instead of the ERB file, but until we update the tests to do that could we add `# cb_helper.css_classes generates "gem-c-checkboxes"`? instead of listing this as an exception. It's not great either way, so up to you.

What bothers me is that we're branching more and more the checkboxes from radios, when in theory they should be very similar.

PR #820, opened 2019-04-16
[DO NOT MERGE] Add expander component

andysellick:
Adds an expander component. This is a simple component that accepts a block of any content, then hides it behind a toggle button. This is intended for use in finder-frontend, to wrap search facets with.

Screen Shot 2019-04-16 at 15 19 29

The accessibility of this component has been modelled after finder-frontend's option select component, and does the following:

- content is expanded by default if JS disabled
- toggle control is turned into a button if JS enabled
- aria-expanded and aria-controls are set accordingly

See it: https://govuk-publishing-compon-pr-820.herokuapp.com/component-guide/expander

Trello card: https://trello.com/c/bDJmf7Sq/609-prototype-the-new-facet-accordion

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/expander.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/expander.js
@@ -1,0 +1,53 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}; // if this ; is omitted, none of this JS runs :(
+
+(function (Modules) {
+  function Expander () {}
+
+  Expander.prototype.start = function ($module) {
+    this.$module = $module[0] // this is the expander element
+    this.$toggle = this.$module.querySelector('.js-toggle')
+    this.$content = this.$module.querySelector('.js-content')
+
+    var openOnLoad = this.$module.getAttribute('data-open-on-load') === 'true' ? true : false
+
+    if (!openOnLoad) {
+      this.collapseContent()
+    }
+    this.replaceTitleWithButton(openOnLoad)
+
+    this.$module.toggleContent = this.toggleContent.bind(this)
+    this.$toggleButton = this.$module.querySelector('.js-button')
+    this.$toggleButton.addEventListener('click', this.$module.toggleContent)
+  }
+
+  Expander.prototype.collapseContent = function () {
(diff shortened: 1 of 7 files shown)


alex-ju:
Didn't manage to review yet, but I was wondering if using a `<details>` element would simplify this component and remove some of the JavaScript.

aliuk2012:
@alex-ju I thought about suggesting that but then you have to rely on polyfills for IE 11 etc

alex-ju:
@aliuk2012 we already have the polyfill in for the details component.

andysellick:
@alex-ju this might be clearer if I give you some context - we're going to use this component to wrap our facets in finder-frontend, like this:

Screen Shot 2019-04-18 at 08 43 31

Would a `<details>` element would be appropriate here?

alex-ju:
In terms of functionality, I think this is exactly what a `<details>` element is meant to be – a disclosure widget from which the user can obtain additional information or controls.

I'm also not sure how this is different from the [accordion component](https://govuk-publishing-compon-pr-820.herokuapp.com/component-guide/accordion). The description and acceptance criteria are nicely overlapping. If we add two options to the accordion component to use chevron icons (instead of +/-) and to hide the "Open all sections" link we achieve the same thing.

The code looks good, but I'm reserved about duplicating functionality and creating confusion about which component to be used where. I may be missing something

andysellick:
@alex-ju these are valid concerns. I had the same thought about the accordion duplication myself, but for two things. Firstly, that they're visually different (although as you said we could have options to restyle the accordion). Secondly, I've built this as a single 'accordion' element (rather than allowing it to have multiple expanding sections) because of the very specific use case here.

We want to wrap facets in these expandable sections (as shown in the screenshot in my previous comment) but the template structure of the code makes it very difficult to use the accordion, as all of the contents of the accordion are in different templates. In addition, three of the expanding sections in that screenshot do not use this component, they use the option select component (in finder-frontend), which is being restyled to look like this component (the reasons for this are complicated).

Having thought about this some more, I've decided that the gem is not the place for this component, so I've opened https://github.com/alphagov/finder-frontend/pull/1056 to put the code into finder-frontend and not here.

@aliuk2012 the code is virtually identical, are you happy to approve? @alex-ju I'll wait for your thoughts before proceeding.

PR #821, opened 2019-04-18
Track when the cookie banner isn't shown

vanitabarrett:
Trello: https://trello.com/c/tOKIPmZ3/99-improve-cookie-banner-tracking

Because the cookie banner is now persistent, we'd rather track when it's not shown rather than when it is.


vanitabarrett:

PR #823, opened 2019-04-26
Remove link to Whitehall publications

karlbaker02:
This PR removes the link to the Whitehall publications finder from the _Government navigation_ component, as it does not map directly to any of the new finders.

Trello: https://trello.com/c/3BuGZK2A

### DIFF
--- a/app/views/govuk_publishing_components/components/_government_navigation.html.erb
+++ b/app/views/govuk_publishing_components/components/_government_navigation.html.erb
@@ -25,11 +25,6 @@
       </a>
     </li>
     <li class="clear-child">
-      <a class="<%= 'active' if active == 'publications' %>" href="/government/publications">
-        <%= t("govuk_component.government_navigation.publications", default: "Publications") %>
-      </a>
-    </li>
-    <li>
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_government_navigation.html.erb
```diff
@@ -25,11 +25,6 @@
       </a>
     </li>
     <li class="clear-child">
-      <a class="<%= 'active' if active == 'publications' %>" href="/government/publications">
-        <%= t("govuk_component.government_navigation.publications", default: "Publications") %>
```
bilbof: Is there a translation we can remove as well?

alex-ju: This repo doesn't seem to have a translation for this.

alex-ju:
@karlbaker02 this change looks good. is there any blocker from getting it merged?

sihugh:
@alex-ju we had to pause for a while, but this will be picked up again in the next couple of days.

PR #825, opened 2019-04-30
Version 6.12.0

vanitabarrett:
Bump gem to include the following change:

- Fire GA event when cookie banner isn't shown, instead of when it is (PR #821) ([Trello](https://trello.com/c/tOKIPmZ3/99-improve-cookie-banner-tracking))


vanitabarrett:

PR #827, opened 2019-04-30
Remove govspeak dependency

kevindew:

I've marked this as a draft PR while I experiment with govspeak using it.

This removes the dependency this gem has on govspeak. This is in
preparation of for govspeak to have the ability to call this gem as a
means to render components within govspeak. This avoids a cyclic dependency.

This gem previously used govspeak as a means to render markdown, there
wasn't actually any specific reason this was govspeak - probably
convenience more than anything. So I've stripped this back to use the
Kramdown gem directly.

I did debate whether to use commonmarker [1] instead of kramdown, but
felt that since there's a high chance govspeak will be
installed in places of this gem they could share the same markdown
library.

For styling I created a SASS mixin that could be used for markdown
typography and included in govspeak. This allows the styles to be
shared. I feel in time these may diverge slightly to allow features
govspeak disables like more headers, but this didn't seem necessary to
implement now.

For the wrapper for markdown HTML I went for a class of
`component-markdown` on the basis that it is intended to have the
results of markdown HTML inside it. I did consider using the term
"prose" instead of "markdown", as that is govuk-frontend terminology for
HTML without classes. However govuk-frontend have since removed this [2]
so I felt using the name "markdown" would be easier to understand.

As I put this together I actually think there's value to be unlocked
from having means to render markdown in apps and govspeak be two
separate things. We have to disable govspeak security features to
allow generic app rendering [3] so a divergence of these would likely
allow both govspeak security and in app HTML flexibility to increase.
Without the baggage of govspeak options we could also just allow
passing raw markdown direct to a component rather than having an app
convert it as is done with govspeak.

I've not taken these thoughts any further here though, I'll see how this
pans out and if these are real needs.

[1]: https://github.com/gjtorikian/commonmarker
[2]: https://github.com/alphagov/govuk-frontend/pull/778
[3]: https://github.com/alphagov/content-publisher/blob/654bf1a257e9a0de607e2c16b61754d55f5f8a72/app/helpers/application_helper.rb#L9


nickcolley:
> This is in
preparation of for govspeak to have the ability to call this gem as a
means to render components within govspeak.

💌 😍 ❤️ 💘 😻

nickcolley:
I think this is a good first step.

Not something that should block this but something to consider:

When/if you have govspeak depending on this gem you can consider adding custom rendering for the common outputs that call the components from this gem.

This would then allow you to avoid duplicating the CSS, as the markdown itself will have classes.

kevindew:
@nickcolley Thanks

I was a bit unsure what you mean with your second comment. Do you mean parsing the markdown generated HTML and adding classes to each element? e.g. input of `<h1>Test</h1>` output of `<h1 class="govuk-heading-xl">Test</h1>` or am I miles off?

nickcolley:
@kevindew yeah exactly.

And for components that exist in this gem we can remove the custom implementations: https://github.com/alphagov/govspeak/blob/c3a0742c87537a371108d097cea23688efa90a14/lib/govspeak.rb#L166

kevindew:
> @kevindew yeah exactly.

Ah ha! that'd be rather clever. That feels like quite a into the future idea as we've got those ~400,000 documents with plain HTML. 

I'm not really sure where would be best in the stack that conversion would be (probably a good argument for frontend apps rendering markdown), would be a nice way to reduce CSS.

> And for components that exist in this gem we can remove the custom implementations: https://github.com/alphagov/govspeak/blob/c3a0742c87537a371108d097cea23688efa90a14/lib/govspeak.rb#L166
> 
> You'd then end up with a markdown output that feels like a developer had authored it, you ship less CSS to users and it's more maintainable with a single source of truth :)

Yup it'd be really cool to replace those bits with component calls.

Fingers crossed this first experiment works out. Thanks for your comments 👍

PR #828, opened 2019-04-30
Permit Document List metadata fields to be blank

bilbof:
This will enable us to provide nil as a value in document list
metadata. E.g. public_updated_at: nil. This is useful, as
sometimes we can have nil fields and dont check for them,
which causes errors. We could fix this when using the component
but it is nicer to fix it once here.

### DIFF
--- a/spec/components/document_list_spec.rb
+++ b/spec/components/document_list_spec.rb
@@ -74,6 +74,27 @@ describe "Document list", type: :view do
     assert_select "#{attribute} time[datetime='2017-07-19T15:01:48Z']"
   end
 
+  it "renders a document list item even when public_updated_at is nil" do
+    render_component(
+      items: [
+        {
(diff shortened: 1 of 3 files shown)


spec/components/document_list_spec.rb
```diff
@@ -74,6 +74,26 @@ def component_name
...
+  it "renders a document list item even when public_updated_at is nil" do
+    render_component(
+      items: [
+        {
+          link: {
+            text: "Some news stories are timeless",
+            path: "/timeless-news",
+          },
+          metadata: {
+            document_type: "News Story"
```
barrucadu: Should this also have `public_updated_at: nil` in the hash?

bilbof: Ha, yes it should. Thanks for catching that!

PR #830, opened 2019-04-30
Add gds-api-adapters and Plek as dependencies

kevindew:

GDS Api Adapters was a listed as a development dependency for this gem,
however it's required outside dev environment to pull in data from
Rummager [1]. Plek is also used here, so this is flagged as a
dependency.

Arguably both of these stretch beyond what this gem should actually be
doing, so shouldn't be encouraged patterns.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/c1c602af10a3ee1dea24c77357c34b5afc843561/lib/govuk_publishing_components/presenters/services.rb

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+- Add gds-adpi-adapters and plek as dependencies (PR #830)
+
 ## 16.13.0
 
  - Add an option to display search icon in input element (PR #824)
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -2,9 +2,11 @@ PATH
   remote: .
   specs:
     govuk_publishing_components (16.13.0)
+      gds-api-adapters
       govspeak (>= 5.0.3)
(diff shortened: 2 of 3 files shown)


benthorner:
Plek seems to be used in a few places, but I can't see where that rummager service is used - could it be redundant?

kevindew:
> Plek seems to be used in a few places, but I can't see where that rummager service is used - could it be redundant?

A link is hidden away in the commit message: https://github.com/alphagov/govuk_publishing_components/blob/c1c602af10a3ee1dea24c77357c34b5afc843561/lib/govuk_publishing_components/presenters/services.rb

benthorner:
@kevindew yes, I saw that, but I can't see where it's used in this repo. I searched a couple of times, but it (and therefore gds-api-adapters) seems to be redundant?

kevindew:
> @kevindew yes, I saw that, but I can't see where it's used in this repo. I searched a couple of times, but it (and therefore gds-api-adapters) seems to be redundant?

Oh I see what you mean, yes doesn't look like anything calls it anymore. That's great - good catch :+1: I'll raise a PR once I finish what I'm doing.

PR #831, opened 2019-05-01
Adds support to allow parent finder to show as breadcrumbs

DilwoarH:
Previously only content items that were a specialist document showed it's parent finder as breadcrumb links. This functionality has now been changed to support all documents which have a finder linked.

This work was done as part of the following ticket:
https://trello.com/c/uR7v1Rri/71-update-the-breadcrumbs-of-business-finder-content-to-link-to-the-business-finder

### DIFF
--- a/spec/components/contextual_breadcrumbs_spec.rb
+++ b/spec/components/contextual_breadcrumbs_spec.rb
@@ -104,6 +104,14 @@ describe "ContextualBreadcrumbs", type: :view do
     assert_select "a", text: "Competition and Markets Authority cases"
   end
 
+  it "renders parent finder breadcrumb if content has a finder linked" do
+    content_item = example_document_for("guide", "guide-with-facet-groups")
+    render_component(content_item: content_item)
+
+    assert_select "a", text: "Home"
+    assert_select "a", text: "EU Withdrawal Act 2018 statutory instruments"
+  end
+
   it "renders no breadcrumbs if there aren't any" do
     content_item = example_document_for("guide", "guide")
     content_item = remove_mainstream_browse(content_item)
--- a/spec/components/contextual_breadcrumbs_spec.rb
+++ b/spec/components/contextual_breadcrumbs_spec.rb
@@ -112,6 +112,14 @@ describe "ContextualBreadcrumbs", type: :view do
     assert_select "a", text: "EU Withdrawal Act 2018 statutory instruments"
   end
 
+  it "renders parent finder breadcrumb if content has a finder linked and taxon is prioritised" do
+    content_item = example_document_for("guide", "guide-with-facet-groups")
+    render_component(content_item: content_item, prioritise_taxon_breadcrumbs: true)
+
+    assert_select "a", text: "Home"
+    assert_select "a", text: "EU Withdrawal Act 2018 statutory instruments"
+  end
+
   it "renders no breadcrumbs if there aren't any" do
     content_item = example_document_for("guide", "guide")
     content_item = remove_mainstream_browse(content_item)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,60 +7,66 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased changes
+
+* Adds support to allow parent finder to show as breadcrumbs (PR #831)
+
 ## 16.18.0
 
-- Add inline radio button option (PR #860)
-- Add new version of (experimental) cookie banner (PR #845)
-- Add inline variation for button component (PR #845)
-- Fix govspeak overriding attachment styling (PR #856)
+* Add inline radio button option (PR #860)
+* Add new version of (experimental) cookie banner (PR #845)
+* Add inline variation for button component (PR #845)
+* Fix govspeak overriding attachment styling (PR #856)
+
 
 ## 16.17.0
 
-- Replace gems version of warning button with GOV.UK Frontend version (PR #848)
-- Prevent double click by default for submit buttons (PR #849)
-- Add inline SVG icons to attachment component (PR #850)
-- Update contextual navigation feature tests to use examples from govuk-content-schemas (PR #847)
+* Replace gems version of warning button with GOV.UK Frontend version (PR #848)
+* Prevent double click by default for submit buttons (PR #849)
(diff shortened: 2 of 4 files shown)


DilwoarH:
> Typo in the commit message: "breadcrumb" has a "b". Can you fix it?
> 
> The bigger question is, what happens if the content is tagged to a priority taxon, like "Brexit", i.e. `prioritise_taxon_breadcrumbs` is true? The finder should take precedence. Can you add a test to make sure that's true?
> I think that for the test to pass a conditional needs to be added above this line: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L9

So because the [breadcrumbs](https://github.com/alphagov/govuk_publishing_components/pull/831/files#diff-4dc9c663c65088d93e00d218ef9228faR23) function is the [last function](https://github.com/alphagov/govuk_publishing_components/blob/928efa7fb82cf78a465b8eb6a4c555e9d54b8069/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L26) that gets called, all the other rules will apply first then the breadcrumb function's rules will be applied. Hope that makes sense.

DilwoarH:
> Typo in the commit message: "breadcrumb" has a "b". Can you fix it?
> 
> The bigger question is, what happens if the content is tagged to a priority taxon, like "Brexit", i.e. `prioritise_taxon_breadcrumbs` is true? The finder should take precedence. Can you add a test to make sure that's true?
> I think that for the test to pass a conditional needs to be added above this line: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L9

If a content item is tagged to a finder then it will display the finder if it passes in the [breadcrumbs function](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L17). However for the example above the taxon will take precedence as the [taxon breadcrumbs are passed in](https://github.com/alphagov/govuk_publishing_components/blob/928efa7fb82cf78a465b8eb6a4c555e9d54b8069/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L13).

Regarding the tests, there are already tests for making sure taxon breadcrumbs are prioritised correctly: https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/contextual_breadcrumbs_spec.rb#L116

leenagupte:
> If a content item is tagged to a finder then it will display the finder if it passes in the [breadcrumbs function](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L17). However for the example above the taxon will take precedence as the [taxon breadcrumbs are passed in](https://github.com/alphagov/govuk_publishing_components/blob/928efa7fb82cf78a465b8eb6a4c555e9d54b8069/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L13).
> 
> Regarding the tests, there are already tests for making sure taxon breadcrumbs are prioritised correctly: https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/contextual_breadcrumbs_spec.rb#L116

@DilwoarH Finders should take priority over taxons in the breadcrumb, even if the taxon has been "prioritised". That's because finders are supposed to show more specialised content than a taxon. So, if content is tagged to the business finder and to Brexit, we want to show the finder in the breadcrumb.

We need to leave in the rules for priority taxons though, because if the content is not tagged to a finder, but is tagged to a priority taxon, and to mainstream browse, the breadcrumb should link to the taxon.

spec/components/contextual_breadcrumbs_spec.rb
```diff
@@ -104,6 +104,14 @@ def set_live_taxons(content_item)
     assert_select "a", text: "Competition and Markets Authority cases"
   end
 
+  it "renders parent finder breadcrumb if content has a finder linked" do
+    content_item = example_document_for("guide", "guide-with-facet-groups")
+    render_component(content_item: content_item)
```
leenagupte: We should add another test that sets `prioritise_taxon_breadcrumbs` to true.

DilwoarH:
> > If a content item is tagged to a finder then it will display the finder if it passes in the [breadcrumbs function](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L17). However for the example above the taxon will take precedence as the [taxon breadcrumbs are passed in](https://github.com/alphagov/govuk_publishing_components/blob/928efa7fb82cf78a465b8eb6a4c555e9d54b8069/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L13).
> > Regarding the tests, there are already tests for making sure taxon breadcrumbs are prioritised correctly: https://github.com/alphagov/govuk_publishing_components/blob/master/spec/components/contextual_breadcrumbs_spec.rb#L116
> 
> @DilwoarH Finders should take priority over taxons in the breadcrumb, even if the taxon has been "prioritised". That's because finders are supposed to show more specialised content than a taxon. So, if content is tagged to the business finder and to Brexit, we want to show the finder in the breadcrumb.
> 
> We need to leave in the rules for priority taxons though, because if the content is not tagged to a finder, but is tagged to a priority taxon, and to mainstream browse, the breadcrumb should link to the taxon.

So this will be the new order:
1. **Content tagged to current step by step** - this will render the step by step breadcrumb
2. **Content tagged to a finder** - this will render the finder breadcrumb
3. **Content tagged to a live taxon and is prioritised** - this will render the taxons breadcrumbs
4. **Content tagged to a mainstream browse page** - this will render parent-based breadcrumbs because the page is tagged to mainstream browse
5. **Content has curated related items** - this will render parent-based breadcrumbs because the page has curated related links
6. **Content tagged to a live taxon** - this will render taxonomy breadcrumbs because the page is tagged to live taxons
7. **Any other breadcrumbs** - this will render parent-based breadcrumbs

spec/components/contextual_breadcrumbs_spec.rb
DilwoarH: Done 👍

PR #833, opened 2019-05-01
Attachment link (experimental) component

kevindew:

👉 [View component](https://govuk-publishing-compon-pr-833.herokuapp.com/component-guide/attachment_link)

This adds a component for an attachment link which has the intention to be used by Govspeak and Content Publisher application.

It's intended as a start of moving the attachment rendering to a shared place that all GOV.UK Publishing apps can use ([relevant XKCD](https://xkcd.com/927/)) and ties into Content Publisher work on adding attachments and overhauling some of the formats that are actually used.

For more details see the commits.


app/views/govuk_publishing_components/components/docs/attachment_link.yml
```diff
@@ -0,0 +1,45 @@
...
+  This component is marked as experimental as it is part of a drive to provide
+  a consistent place for attachment presentation logic (currently this logic is
+  within [Govspeak][] and [Whitehall][]). The API for this
+  component may change during this process.
+
+  It provides a means to show a link to an attachment with some
+  relevant file data.
+
+  It is expected to be embedded inside an element that provides text styles
+  (such as `.govuk-body`) so does not provide it's own text styling.
```
andysellick: Nitpick: should be `its`.

spec/components/attachment_link_spec.rb
```diff
@@ -0,0 +1,42 @@
...
+  end
+
+  it "can include attribute metadata" do
+    render_component(
+      attachment: {
+        title: "Attachment",
+        url: "attachment",
+        content_type: "text/xml",
+        file_size: 2048,
+        number_of_pages: 2,
```
andysellick: Should there be a test that 1 page outputs the text `page` as opposed to `pages`?

app/views/govuk_publishing_components/components/docs/attachment_link.yml
kevindew: Thanks - sorted

spec/components/attachment_link_spec.rb
kevindew: I don't think that's necessary as that is testing pluralize that ActionView provides. If we were to test that it'd beg the question as to whether to test all the content_type inputs and file_size variations.

kevindew:
Thanks for the review @andysellick 

I've just made a small change of taking filename as an argument too as a source of extension data as I realised not all urls will have that data.

PR #839, opened 2019-05-03
Revert cookie banner tracking

vanitabarrett:
Removes the cookie banner tracking which fires when the cookie banner is hidden. This would fire an additional event on every user page view, which will build up our analytics events to hit our limit. Instead, we will be adding a custom dimension to the existing pageview event.

This essentially reverts #821.


vanitabarrett:

PR #842, opened 2019-05-03
Add Attachment component

emmabeynon:
For: https://trello.com/c/zMYmFTPx/801-show-metadata-for-attachments

🎉 View in the [component guide](http://govuk-publishing-compon-pr-842.herokuapp.com/component-guide/attachment) 🎉 

This adds a component for an attachment, which is intended to be used by Govspeak and Content Publisher.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -1,0 +1,44 @@
+$thumbnail-width: 99px;
+$thumbnail-height: 140px;
+$thumbnail-border: 5px;
+$thumbnail-border-colour: rgba(11, 12, 12, .1);
+$thumbnail-shadow-colour: rgba(11, 12, 12, .4);
+$thumbnail-shadow-width: 0 2px 2px;
+
+.gem-c-attachment {
+  @include govuk-font(19);
+  @include govuk-clearfix;
+}
+
+.gem-c-attachment__thumbnail {
+  float: left;
+  margin-bottom: govuk-spacing(3);
+  margin-right: govuk-spacing(5);
+  padding: $thumbnail-border;
+  position: relative;
+}
+
+.gem-c-attachment__thumbnail-image {
+  display: block;
+  width: $thumbnail-width;
+  max-width: 100%;
+  height: $thumbnail-height;
+  outline: $thumbnail-border solid $thumbnail-border-colour;
+  background: govuk-colour("white");
+  box-shadow: $thumbnail-shadow-width $thumbnail-shadow-colour;
+}
+
+.gem-c-attachment__title {
+  @include govuk-font($size: 27);
+  margin: 0 0 govuk-spacing(3);
+}
+
+.gem-c-attachment__metadata {
+  @include govuk-font($size: 14);
+  margin: 0 0 govuk-spacing(3);
+}
+
+.gem-c-attachment__abbr {
+  cursor: help;
+  text-decoration: none;
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -24,6 +24,7 @@ $thumbnail-shadow-width: 0 2px 2px;
   max-width: $thumbnail-width;
   height: $thumbnail-height;
   outline: $thumbnail-border solid $thumbnail-border-colour;
+  border: $thumbnail-border-colour; // for IE9 & IE10
   background: govuk-colour("white");
   box-shadow: $thumbnail-shadow-width $thumbnail-shadow-colour;
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -20,8 +20,8 @@ $thumbnail-shadow-width: 0 2px 2px;
 
 .gem-c-attachment__thumbnail-image {
   display: block;
-  width: $thumbnail-width;
-  max-width: 100%;
+  width: auto; // for IE8
+  max-width: $thumbnail-width;
   height: $thumbnail-height;
   outline: $thumbnail-border solid $thumbnail-border-colour;
   background: govuk-colour("white");
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -8,15 +8,15 @@
     content = tag.abbr(attachment.content_type_abbr,
                        class: "gem-c-attachment__abbr",
                        title: attachment.readable_content_type)
-    attributes << tag.span(content)
+    attributes << tag.span(content, class: "gem-c-attachment__attribute")
   end
 
   if attachment.readable_file_size
-    attributes << tag.span(attachment.readable_file_size)
+    attributes << tag.span(attachment.readable_file_size, class: "gem-c-attachment__attribute")
   end
 
   if attachment.readable_number_of_pages
-    attributes << tag.span(attachment.readable_number_of_pages)
+    attributes << tag.span(attachment.readable_number_of_pages, class: "gem-c-attachment__attribute")
   end
 %>
 <%= tag.section class: "gem-c-attachment" do %>
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -1,0 +1,47 @@
+<%
+  attachment = GovukPublishingComponents::Presenters::Attachment.new(attachment)
+  target ||= "_self"
+  hide_help_text ||= false
+  attributes = []
+
+  if attachment.content_type_abbr
+    content = tag.abbr(attachment.content_type_abbr,
+                       class: "gem-c-attachment__abbr",
+                       title: attachment.readable_content_type)
+    attributes << tag.span(content)
+  end
+
+  if attachment.readable_file_size
+    attributes << tag.span(attachment.readable_file_size)
+  end
+
+  if attachment.readable_number_of_pages
+    attributes << tag.span(attachment.readable_number_of_pages)
+  end
+%>
+<%= tag.section class: "gem-c-attachment" do %>
+  <%= tag.div class: "gem-c-attachment__thumbnail" do %>
+    <%= link_to attachment.url, target: target, tabindex: "-1", "aria-hidden": true do %>
+      <%= tag.img class: "gem-c-attachment__thumbnail-image",
+                  src: attachment.thumbnail
+      %>
+    <% end %>
+  <% end %>
+
+  <%= tag.h2 class: "gem-c-attachment__title" do %>
+    <%= link_to attachment.title, attachment.url,
+          class: "govuk-link",
+          target: target %>
+  <% end %>
+  <%= tag.p class: "gem-c-attachment__metadata" do %>
+    <%= raw(attributes.join(', ')) if attributes.any? %>
+  <% end %>
+
+  <% unless hide_help_text %>
+    <% if attachment.opendocument? %>
+      <%= tag.p class: "gem-c-attachment__metadata" do %>
+        <%= t("components.attachment.opendocument_html", target: target) %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/attachment.yml
+++ b/app/views/govuk_publishing_components/components/docs/attachment.yml
@@ -1,0 +1,57 @@
+name: Attachment (experimental)
+description: Displays a link to download an attachment and metadata about the file
+body: |
+  This component is marked as experimental as it is part of a drive to provide
+  a consistent place for attachment presentation logic (currently this logic is
+  within [Govspeak][] and [Whitehall][]). The API for this
+  component may change during this process.
+
+  It shows a link to a document that is attached to GOV.UK content along with a
+  thumbnail and relevant file data.
+
+  It is intended to be rendered in Govspeak and as an attachment 'preview' in
+  Content Publisher.
+
+  [Govspeak]: https://github.com/alphagov/govspeak/blob/c3a0742c87537a371108d097cea23688efa90a14/lib/govspeak/presenters/attachment_presenter.rb
+  [Whitehall]: https://github.com/alphagov/whitehall/blob/5c760eea912320c5a80ef59791df47161d889f2f/app/helpers/document_helper.rb
+shared_accessibility_criteria:
+- link
+accessibility_criteria: |
+  The thumbnail image has no `alt` attribute, which causes the link which wraps
+  the thumbnail to contain no content. Since screen readers will have nothing to
+  read in this case, we set `aria-hidden=true` on the thumbail so that it's hidden.
+examples:
+  default:
+    data:
+      attachment:
+        title: "Department for Transport information asset register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/747661/department-for-transport-information-asset-register.csv
+        filename: department-for-transport-information-asset-register.csv
+        content_type: text/csv
+        file_size: 20000
+  with_number_of_pages:
+    data:
+      attachment:
+        title: "Temporary snow ploughs: guidance note"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/259634/temporary-snow-ploughs.pdf
+        filename: temporary-snow-ploughs.pdf
+        content_type: application/pdf
+        file_size: 20000
+        number_of_pages: 7
+  opendocument:
+    data:
+      attachment:
+        title: "BEIS Information Asset Register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/744083/BEIS_Information_Asset_Register_.ods
+        filename: BEIS_Information_Asset_Register_.ods
+        content_type: application/vnd.oasis.opendocument.spreadsheet
+        file_size: 20000
+  help_text_disabled:
+    data:
+      attachment:
+        title: "BEIS Information Asset Register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/744083/BEIS_Information_Asset_Register_.ods
+        filename: BEIS_Information_Asset_Register_.ods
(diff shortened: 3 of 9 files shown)


app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+  <%= tag.div class: "gem-c-attachment__thumbnail" do %>
+    <%= link_to attachment.url, target: target do %>
+      <%= tag.img class: "gem-c-attachment__thumbnail-image",
+                  src: attachment.thumbnail,
+                  alt: attachment.title
+      %>
+    <% end %>
+  <% end %>
+
+  <%= tag.div do %>
```
andysellick: Is this `div` necessary? It's not styled and doesn't seem to serve any purpose.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+            class: "govuk-link",
+            target: target %>
+    <% end %>
+    <%= tag.p class: "gem-c-attachment__metadata" do %>
+      <%= raw(attributes.join(', ')) if attributes.any? %>
+    <% end %>
+  <% end %>
+
+  <% unless hide_help_text %>
+    <%= tag.div do %>
```
andysellick: Is this `div` necessary? It's not styled and doesn't seem to serve any purpose. Also it's outside of the logic below so it can appear empty (see first example in component guide).

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <% if attachment.opendocument? %>
+        <%= tag.p class: "gem-c-attachment__metadata" do %>
+            This file is in an
```
andysellick: We should ideally have this text be translatable.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+            <%= link_to "OpenDocument",
+              "https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software",
+              target: "_blank" %>
+            format
```
andysellick: Also should be translatable.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+    attributes << tag.span(attachment.readable_file_size)
+  end
+
+  if attachment.readable_number_of_pages
+    attributes << tag.span(attachment.readable_number_of_pages)
+  end
+%>
+<%= tag.section class: "gem-c-attachment" do %>
+  <%= tag.div class: "gem-c-attachment__thumbnail" do %>
+    <%= link_to attachment.url, target: target do %>
```
andysellick: This link is focusable but doesn't have the correct focus styles set, it just has the browser default.

Screen Shot 2019-05-07 at 09 25 53

Having thought about it, it's probably not worth letting this link be focusable - it's the same as the text link right? Maybe just give it a `tabindex=-1` attribute, people don't need to tab through two identical links.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,56 @@
...
+      <%= raw(attributes.join(', ')) if attributes.any? %>
+    <% end %>
+  <% end %>
+
+  <% unless hide_help_text %>
+    <%= tag.div do %>
+      <% if attachment.opendocument? %>
+        <%= tag.p class: "gem-c-attachment__metadata" do %>
+            This file is in an
+            <%= link_to "OpenDocument",
```
andysellick: This link should be styled - it currently looks like it doesn't have the correct link colour or focus styles.

Screen Shot 2019-05-07 at 09 31 51

app/views/govuk_publishing_components/components/docs/attachment.yml
```diff
@@ -0,0 +1,49 @@
...
+  This component is marked as experimental as it is part of a drive to provide
+  a consistent place for attachment presentation logic (currently this logic is
+  within [Govspeak][] and [Whitehall][]). The API for this
+  component may change during this process.
+
+  It is used to show data about an attachment document to GOV.UK content.
+
+  [Govspeak]: https://github.com/alphagov/govspeak/blob/c3a0742c87537a371108d097cea23688efa90a14/lib/govspeak/presenters/attachment_presenter.rb
+  [Whitehall]: https://github.com/alphagov/whitehall/blob/5c760eea912320c5a80ef59791df47161d889f2f/app/helpers/document_helper.rb
+shared_accessibility_criteria:
```
andysellick: I think we might be able to define some accessibility criteria for this component, maybe images in the component should have meaningful alt text?

app/views/govuk_publishing_components/components/docs/attachment.yml
```diff
@@ -0,0 +1,49 @@
...
+- link
+examples:
+  default:
+    data:
+      attachment:
+        title: "Department for Transport information asset register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/747661/department-for-transport-information-asset-register.csv
+        filename: department-for-transport-information-asset-register.csv
```
andysellick: What's `filename` for? It doesn't seem to be used anywhere?

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -0,0 +1,43 @@
...
+}
+
+.gem-c-attachment__thumbnail-image {
+  display: block;
+  width: $thumbnail-width;
+  max-width: 100%;
+  height: $thumbnail-height;
+  outline: $thumbnail-border solid $thumbnail-border-colour;
+  background: govuk-colour("white");
+  box-shadow: 0 2px 2px $thumbnail-shadow-colour;
```
alex-ju: Stylistic comment: can we move the box shadow width into its own variable at the top.
I'm aware I introduces this line in Content Publisher, but it's good practice to have every value in a variable and we don't have yet the linting setup properly.

app/views/govuk_publishing_components/components/docs/attachment.yml
emmabeynon: Thanks, I've added that in 👍

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: Now removed

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: Now removed

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've moved this in to en.yml

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've added `tabindex=1`

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've added `govuk-link` as the class - is that correct?

app/views/govuk_publishing_components/components/docs/attachment.yml
emmabeynon: We can use the filename to determine the content type of the attachment: https://github.com/alphagov/govuk_publishing_components/blob/dc8595d2a7d6a2404f85666a387b8f466ae960c3/lib/govuk_publishing_components/presenters/attachment.rb#L27

config/locales/en.yml
```diff
@@ -23,6 +23,8 @@ en:
   common:
     translations: "Translations"
   components:
+    attachment:
+      opendocument: "This file is in an <a href='https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software' target='_blank' class='govuk-link'>OpenDocument</a> format"
```
kevindew: If you suffix this with _html you won't need to run raw on it: https://guides.rubyonrails.org/i18n.html#using-safe-html-translations

lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -48,6 +50,14 @@ def readable_number_of_pages
         pluralize(attachment_data[:number_of_pages], "page")
       end
 
+      def thumbnail
+        "https://www.gov.uk/government/assets/pub-cover-doc-afe3b0a8a9511beeca890340170aee8b5649413f948e512c9b8ce432d8513d32.png"
+      end
+
+      def opendocument?
+        OPENDOCUMENT_ABBREVIATIONS.include?(content_type_abbr)
```
kevindew: I don't think we should couple this to abbreviation as that is a frontend rendering concern. 

What I'd suggest we do is set constants for the mime types and then replace their string usage in https://github.com/alphagov/govuk_publishing_components/blob/57ff9157c9c8be3902acb8fc154265ed35464c29/lib/govuk_publishing_components/presenters/attachment.rb#L73-L75

and then check the content_type is one of those.

config/locales/en.yml
kevindew: Should this always be target blank by the way? doesn't look like Whitehall one is: https://github.com/alphagov/whitehall/blob/7b5c5a086b89cb62ffba62b152a0a8dcfc10c8e6/config/locales/en.yml#L447

Probably should use the target attribute to component to decide this, maybe this could be pushed into an argument of this but default to _self?

app/views/govuk_publishing_components/components/docs/attachment.yml
```diff
@@ -0,0 +1,53 @@
+name: Attachment (experimental)
+description: Displays a link to download an attachment and metadata about the file
+body: |
+  This component is marked as experimental as it is part of a drive to provide
+  a consistent place for attachment presentation logic (currently this logic is
+  within [Govspeak][] and [Whitehall][]). The API for this
+  component may change during this process.
+
+  It is used to show data about an attachment document to GOV.UK content.
```
kevindew: Not sure if I wrote this or not but feel like this sentence could be improved with a bit more information on what this is and it's usages.

app/views/govuk_publishing_components/components/docs/attachment.yml
```diff
@@ -0,0 +1,53 @@
...
+
+  [Govspeak]: https://github.com/alphagov/govspeak/blob/c3a0742c87537a371108d097cea23688efa90a14/lib/govspeak/presenters/attachment_presenter.rb
+  [Whitehall]: https://github.com/alphagov/whitehall/blob/5c760eea912320c5a80ef59791df47161d889f2f/app/helpers/document_helper.rb
+accessibility_criteria: |
+  The image must:
+
+  * have alt text
```
kevindew: Whitehall has a precedent of these not having alt text and instead using ARIA hidden to not show that to users. So I think here we'd say that we don't allow targetting an image and mark them as hidden from screen readers.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  attachment = GovukPublishingComponents::Presenters::Attachment.new(attachment)
+  target ||= nil
+  hide_help_text ||= false
+  attributes = []
+
+  if attachment.content_type_abbr
+    content = tag.abbr(attachment.content_type_abbr,
+                       class: "gem-c-attachment__abbr",
+                       title: attachment.readable_content_type)
+    attributes << tag.span(content)
```
kevindew: I think we should add similar classes to the ones we have for attachment link: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_attachment_link.html.erb#L18

Mainly because it'd be such a pain to do any styling without the classes once these components are embedded in content HTML.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,48 @@
...
+  if attachment.readable_number_of_pages
+    attributes << tag.span(attachment.readable_number_of_pages)
+  end
+%>
+<%= tag.section class: "gem-c-attachment" do %>
+  <%= tag.div class: "gem-c-attachment__thumbnail" do %>
+    <%= link_to attachment.url, target: target, tabindex: "-1" do %>
+      <%= tag.img class: "gem-c-attachment__thumbnail-image",
+                  src: attachment.thumbnail,
+                  alt: attachment.title
```
kevindew: This should be blank based on Whitehall precedent.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -0,0 +1,48 @@
...
+    attributes << tag.span(attachment.readable_file_size)
+  end
+
+  if attachment.readable_number_of_pages
+    attributes << tag.span(attachment.readable_number_of_pages)
+  end
+%>
+<%= tag.section class: "gem-c-attachment" do %>
+  <%= tag.div class: "gem-c-attachment__thumbnail" do %>
+    <%= link_to attachment.url, target: target, tabindex: "-1" do %>
```
kevindew: Are we marking this as aria-hidden as per Whitehall: https://github.com/alphagov/whitehall/blob/3cee9551c6247c7edded9204b4f5f4ed0d71c514/app/views/documents/_attachment.html.erb#L5

app/views/govuk_publishing_components/components/_attachment.html.erb
alex-ju: yep, that correct 👍

app/views/govuk_publishing_components/components/docs/attachment.yml
andysellick: Ah, fair enough, thanks.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -0,0 +1,43 @@
+$thumbnail-width: 99px;
+$thumbnail-height: 140px;
+$thumbnail-border: 5px;
+$thumbnail-border-colour: rgba(11, 12, 12, .1);
```
andysellick: Very minor comment - this is using an RGBA value where it isn't really necessary - since I think this will always be displayed on a white background? It could be changed to `rgb(230, 230, 230);` and appear the same, but older browsers (IE8) would be able to display it.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -0,0 +1,43 @@
...
+  box-shadow: 0 2px 2px $thumbnail-shadow-colour;
+}
+
+.gem-c-attachment__title {
+  @include govuk-font($size: 27);
+  margin: 0 0 govuk-spacing(3);
+}
+
+.gem-c-attachment__metadata {
+  @include govuk-font($size: 14);
```
andysellick: We probably don't want to mess with the design too much but on mobile this shrinks to 12px which is pretty small, and there's loads of space available. Maybe increase this line to size 16?

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
emmabeynon: I changed this to the rgb value and for some reason the shadow disappeared 😱

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I noticed that `gem-c-attachment-link__attribute` isn't in attachment-link.scss.  Should we be adding classes with no styling to the markup?

kevindew: I'd suggest so, since we can't add the classes later once this is embedded in a documents HTML.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
alex-ju: Checked this with our designer and probably the best option here is to use 14px for both desktop and mobile. What do you think about this @andysellick?

We can achieve this by explicitly declaring the font-size for mobile.
```
.gem-c-attachment__metadata {

      font-size: 14px;
    }
}
```

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
emmabeynon: I've called this `$thumbnail-shadow-width`.

config/locales/en.yml
emmabeynon: I've changed the target attributes to default to`_self` and the translation is now suffixed with `-html`.

lib/govuk_publishing_components/presenters/attachment.rb
emmabeynon: As discussed, `#opendocument?` now delegates to SupportedContentType to check if it's an OpenDocument or not.

app/views/govuk_publishing_components/components/docs/attachment.yml
emmabeynon: I've added a bit more to this documentation - let me know if that's ok or if there's anything else you think needs to be added.

app/views/govuk_publishing_components/components/docs/attachment.yml
emmabeynon: Now removed

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've added these classes in a separate commit just to be clear about why they don't have any styling yet.

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've removed the alt attribute.

app/views/govuk_publishing_components/components/_attachment.html.erb
emmabeynon: I've added `aria-hidden: true` to this link

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
emmabeynon: Spoke with @alex-ju about this and he suggested we stick with using RGBA here; it turns out we'll be rendering this component in Content Publisher on a light grey background.

PR #843, opened 2019-05-07
Update cookies banner behaviour without Javascript

chao-xian:
# Changes

The banner was supposed to not show without Javascript. It did not
display in the component guide, but showed up on the frontend. However
we want the banner to still display, as there are other cookies that
may still be set by servers, and not Javascript.

We also don't show the "Hide this message" link without Javascript.

## Screenshots
All are running without Javascript:

### Before
Screenshot at May 07 3-53-57 pm

### After
Screenshot at May 07 3-54-10 pm

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
@@ -16,3 +16,11 @@ $govuk-cookie-banner-background: #d5e8f3;
   margin-top: 0;
   margin-bottom: 0;
 }
+
+.gem-c-cookie-banner__hide-link {
+  display: none;
+
+  .js-enabled & {
+    display: inline-block;
+  }
+}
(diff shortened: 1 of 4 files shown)


chao-xian:
["Hide this message" should not display without Javascript.](https://trello.com/c/pzRDyjcl/122-hide-this-message-should-not-display-without-javascript)

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -6,6 +6,14 @@ $govuk-cookie-banner-background: #d5e8f3;
   }
 }
 
+.gem-c-cookie-banner__hide-link {
```
DilwoarH: Is it possible this below the `.gem-c-cookie-banner` class so that the component CSS is at the top

chao-xian: Thanks @DilwoarH - [done](https://github.com/alphagov/govuk_publishing_components/pull/843/commits/7fa4cc5d49bd9771bebd8bb9669c7386a1a660c3)!

PR #847, opened 2019-05-08
Update feature tests to use examples from govuk-content-schemas

leenagupte:
Trello: https://trello.com/c/uR7v1Rri

Related to:
https://github.com/alphagov/govuk-content-schemas/pull/892
https://github.com/alphagov/govuk-content-schemas/pull/891

Content item types, `facet_groups`, `facets` and `facet_values` were changed to be "base_path-less" in https://github.com/alphagov/govuk-content-schemas/pull/891. 

We did this because, that's how they were already being treated in publishing-api and content-store, but we couldn't add re-usable test examples until this was explicitly stated in `govuk-content-schemas`.

However, this started causing the [feature tests in this app](https://ci.integration.publishing.service.gov.uk/job/govuk_publishing_components/job/master/731/consoleFull) to fail with errors like:

```
 Failure/Error:
       GovukSchemas::RandomExample.for_schema(frontend_schema: schema_name) do |random_item|
         random_item.merge(merge_with)
       end

     GovukSchemas::InvalidContentGenerated:
       The content item you are trying to generate is invalid against the schema.
       The item was valid before being customised.
```
and
```
         {
           "schema": "046a2c23-3e82-578d-86d3-8d23a5e24eb2#",
           "fragment": "#/links/taxons/1/links/facet_groups/2",
           "message": "The property '#/links/taxons/1/links/facet_groups/2' did not contain a required property of 'base_path' in schema 046a2c23-3e82-578d-86d3-8d23a5e24eb2#",
           "failed_attribute": "Required"
         },
```

The feature tests rely on `govuk_schemas` to create a random content item based on a schema, and then modify a part of it, e.g. the content item title. `govuk_schemas` validates the content item it creates twice. Once when the base content item is created, and again when the modifications are added, i.e. when the content item is "customised".

It's not entirely clear why the validation was failing. The content item was being created "correctly", but the links for `facet_groups` and `facet_values` were being validated against the wrong schema definition. 

Also, the links being created were unnecessary. `taxons` and `parent` links etc, don't need to be tagged to a `facet_group` and it's extremely unlikely that they ever will be.

Rather than spending a lot of time trying to fix random content item generation here, the feature tests have been updated to read in already validated examples to generate the links hash.

## Note about tests
The tests won't pass until https://github.com/alphagov/govuk-content-schemas/pull/892 is deployed to production and the new [taxon_in_alpha_phase](https://github.com/alphagov/govuk-content-schemas/blob/master/examples/taxon/frontend/taxon_in_alpha_phase.json) example is available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 - Replace gems version of warning button with GOV.UK Frontend version (PR #848)
 - Prevent double click by default for submit buttons (PR #849)
+- Update contextual navigation feature tests to use examples from govuk-content-schemas (PR #847)
 
 ## 16.16.0
 - Add attachment (experimental) component (PR #842)
--- a/spec/features/contextual_navigation_spec.rb
+++ b/spec/features/contextual_navigation_spec.rb
@@ -91,7 +91,7 @@ describe "Contextual navigation" do
     content_store_has_random_item(
       links: {
         "parent" => [random_item("mainstream_browse_page", "title" => "A parent")],
-        "ordered_related_items" => [random_item("guide", "title" => "A related link curated in Publisher")]
+        "ordered_related_items" => [example_item("guide", "guide")]
       }
     )
   end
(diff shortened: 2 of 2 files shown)


steventux:
I suspect the problem arises from the facets relating to their facet group via the parent link https://github.com/alphagov/govuk-content-schemas/blob/master/formats/facet.jsonnet#L91 which is a recursive link type.

PR #849, opened 2019-05-09
Prevent double click for buttons

alex-ju:
### What
Set `prevent-double-click` data attribute to `true` by default for the button component.

See [original commit](https://github.com/alphagov/govuk-frontend/commit/884e7dd73ad943d71fb701da15626e6a7ad0b933) for more details about how this works.

### Why
We discovered in research that users with low digital skills sometimes use double click to trigger an action which may result in unexpected system failures (for example when trying to remove an entity the second attempt fails) or double additions. For this reason we think the best measure is to introduce this preventive measure for all instances of buttons instead of trying to figure out when a system may fail because of it.

[Trello card](https://trello.com/c/JLkfkNxQ)

### DIFF
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -12,7 +12,8 @@ module GovukPublishingComponents
         @title = local_assigns[:title]
         @info_text = local_assigns[:info_text]
         @rel = local_assigns[:rel]
-        @data_attributes = local_assigns[:data_attributes]
+        @default_data_attributes = { "prevent-double-click" => true }
+        @data_attributes = local_assigns[:data_attributes] ? local_assigns[:data_attributes].merge(@default_data_attributes) : @default_data_attributes
         @margin_bottom = local_assigns[:margin_bottom]
         @target = local_assigns[:target]
         @type = local_assigns[:type]
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -32,8 +32,7 @@ module GovukPublishingComponents
         options[:role] = "button" if link?
         options[:type] = button_type
         options[:rel] = rel if rel
-        options[:data] = data_attributes ? data_attributes.merge(@default_data_attributes) : @default_data_attributes
-        options[:data] = data_attributes if button_type != "submit"
+        options[:data] = button_type == "submit" ? data_attributes.to_h.merge(@default_data_attributes) : data_attributes
         options[:title] = title if title
         options[:target] = target if target
         options
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -13,7 +13,7 @@ module GovukPublishingComponents
         @info_text = local_assigns[:info_text]
         @rel = local_assigns[:rel]
         @default_data_attributes = { "prevent-double-click" => true }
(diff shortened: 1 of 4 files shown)


lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -12,7 +12,8 @@ def initialize(local_assigns)
         @title = local_assigns[:title]
         @info_text = local_assigns[:info_text]
         @rel = local_assigns[:rel]
-        @data_attributes = local_assigns[:data_attributes]
+        @default_data_attributes = { "prevent-double-click" => true }
```
nickcolley: I think this should only be true if the type is unset or `submit` otherwise you could break any instances of this component where you'd want users to be able to click a button quickly.

benthorner: Good point. I'm interested to know more about the scenarios where we expect users will click two buttons quickly in sequence - that sounds more like an issue with the feature, where we would want to make a deliberate decision to allow for fast clicking in sequence.

alex-ju: Fair, because of the debounce timeout..

alex-ju: @benthorner: _Add item > Submit_ or _Save within a modal > Submit_

nickcolley: When buttons have a type of 'button' then you can expect them to be clicked quickly.

Submit buttons already are debounced by browsers such as Internet Explorer and Edge, they're safe to debounce.

Perhaps we should add something to the guidance about this?

alex-ju: @nickcolley now it's only enabled for buttons with type submit (explicitly defined or not) and added tried to document it in the yml.

alex-ju: @nickcolley for the record, are you ok with how it's being applied now (submit only)?

nickcolley: Can't say on the code specifics, but in theory yes.

lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -31,7 +32,8 @@ def html_options
         options[:role] = "button" if link?
         options[:type] = button_type
         options[:rel] = rel if rel
-        options[:data] = data_attributes if data_attributes
+        options[:data] = data_attributes ? data_attributes.merge(@default_data_attributes) : @default_data_attributes
```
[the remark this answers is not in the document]
alex-ju: Refactored using `to_h`

alex-ju:
@benthorner I consider it [covered by the default case](https://github.com/alphagov/govuk_publishing_components/pull/849/files#diff-f93e2443ed956fd60bf90da71d82d156R102) and the merge is [tested with the rest of data attributes](https://github.com/alphagov/govuk_publishing_components/pull/849/files#diff-f93e2443ed956fd60bf90da71d82d156R110) - happy to add a new test for when explicitly defined as "submit", but it seemed to me like an overkill.

benthorner:
@alex-ju I think that would be a good replacement for https://github.com/alphagov/govuk_publishing_components/pull/849/files#diff-f93e2443ed956fd60bf90da71d82d156R123 - I just noticed we don't have any test for when the double-click attribute isn't present.

alex-ju:
@benthorner updated tests to check for existence (default, type="submit") and non-existence (type="button") of `data-prevent-double-click` attribute.

PR #850, opened 2019-05-09
Add SVG thumbnail icons to attachment component

kevindew:
Trello: https://trello.com/c/HO5MZZEG/822-show-thumbnails-for-attachment-file-formats

This adds in inline SVGs for the different file types that can be shown as an attachment: document, spreadsheet and generic. It does not include HTML attachments as this component does not support these yet. It also does not allow the concept of passing an image in for a PDF attachment as a [product decision](https://gov-uk.atlassian.net/wiki/spaces/GOVUK/pages/1399422981/Design+documentation+Content+Publisher+Q1+19+20#Designdocumentation%E2%80%93ContentPublisherQ119/20-RemovepreviewfromPDFattachments) has been made to remove that functionality.

This also fixes two small issues: content type wasn't showing when it lacked an abbreviation, and we were still getting the browser default outline on the thumbnail link.

Examples of the different icons (which some are rendering incorrectly currently)
Screen Shot 2019-05-09 at 12 17 07
Screen Shot 2019-05-09 at 12 17 26
Screen Shot 2019-05-09 at 12 17 31

This PR is marked as a draft as I'm hoping @alex-ju can do some more styling with the svgs and fix the missing grey fills

### DIFF
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -25,7 +25,11 @@
 %>
 <%= tag.section class: "gem-c-attachment" do %>
   <%= tag.div class: "gem-c-attachment__thumbnail" do %>
-    <%= link_to attachment.url, target: target, tabindex: "-1", "aria-hidden": true do %>
+    <%= link_to attachment.url,
+                class: "govuk-link",
+                target: target,
+                tabindex: "-1",
+                "aria-hidden": true do %>
       <% if attachment.document? %>
         <%= render "govuk_publishing_components/components/attachment/thumbnail_document.svg" %>
       <% elsif attachment.spreadsheet? %>
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -4,10 +4,14 @@
   hide_help_text ||= false
   attributes = []
 
-  if attachment.content_type_abbr
-    content = tag.abbr(attachment.content_type_abbr,
-                       class: "gem-c-attachment__abbr",
-                       title: attachment.readable_content_type)
+  if attachment.readable_content_type
+    content = if attachment.content_type_abbr
+                raw tag.abbr(attachment.content_type.abbr,
+                             title: attachment.readable_content_type,
+                             class: "gem-c-attachment__abbr")
+              else
+                attachment.readable_content_type
+              end
     attributes << tag.span(content, class: "gem-c-attachment__attribute")
   end
 
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -22,9 +22,13 @@
 <%= tag.section class: "gem-c-attachment" do %>
   <%= tag.div class: "gem-c-attachment__thumbnail" do %>
     <%= link_to attachment.url, target: target, tabindex: "-1", "aria-hidden": true do %>
-      <%= tag.img class: "gem-c-attachment__thumbnail-image",
-                  src: attachment.thumbnail
-      %>
+      <% if attachment.document? %>
+        <%= render "govuk_publishing_components/components/attachment/thumbnail_document.svg" %>
+      <% elsif attachment.spreadsheet? %>
+        <%= render "govuk_publishing_components/components/attachment/thumbnail_spreadsheet.svg" %>
+      <% else %>
+        <%= render "govuk_publishing_components/components/attachment/thumbnail_generic.svg" %>
+      <% end %>
     <% end %>
   <% end %>
 
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -45,8 +45,9 @@
           class: "govuk-link",
           target: target %>
   <% end %>
-  <%= tag.p class: "gem-c-attachment__metadata" do %>
-    <%= raw(attributes.join(', ')) if attributes.any? %>
+
+  <% if attributes.any? %>
+    <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
   <% end %>
 
   <% unless hide_help_text %>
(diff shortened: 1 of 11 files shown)


app/views/govuk_publishing_components/components/attachment/_thumbnail_document.svg
```diff
@@ -0,0 +1,4 @@
+<svg class="gem-c-attachment__thumbnail-image gem-c-attachment__thumbnail-image--svg" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 99 140">
```
nickcolley: I think you could squeeze some bytes out of these using a tool like https://jakearchibald.github.io/svgomg/ :)

nickcolley: I would also recommend having fallback `height` and `width`, since if CSS fails you don't want these images filling the page.

You could also consider putting `fill=currentColor` onto the path for similar reasons.

kevindew: Thanks! I already gave the a spin before they went up (well I used the node package for it).

nickcolley: Fair, I can see some stuff that doesnt need to be there so worth going over it again I think.

kevindew: Hmm you're right. I did save some more again. Kinda weird the tool doesn't get it to the lowest the first time 🤔 

based on an original:

```
➜  tmp npx svgo pub-cover-spreadsheet.svg

pub-cover-spreadsheet.svg:
Done in 34 ms!
0.924 KiB - 42.7% = 0.529 KiB
➜  tmp npx svgo pub-cover-spreadsheet.svg

pub-cover-spreadsheet.svg:
Done in 33 ms!
0.529 KiB - 3.9% = 0.509 KiB
➜  tmp npx svgo pub-cover-spreadsheet.svg

pub-cover-spreadsheet.svg:
Done in 29 ms!
0.509 KiB - 0% = 0.509 KiB
➜  tmp npx svgo pub-cover-spreadsheet.svg
```

kevindew: I'm mostly hoping @alex-ju will take over on SVG front though to get it all nice

nickcolley: Janitor dancing

alex-ju: Updated SVGs to clean up markup and apply styling via CSS

kevindew: Thanks Alex - they look good!

I notice you removed the svg modifier class, so I just wanted to check that these would still display ok if this class was put on an img element? Just for the scenario that we continued to support pdf images? (mostly asking since it's hard to add classes later once embedded in documents)

alex-ju: Yes, they will be fine regardless of using an image or an svg - the SVG-specific attributes (`fill` and `stroke`) won't affect a regular bitmap image.

kevindew: Awesome 👍

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -4,10 +4,14 @@
   hide_help_text ||= false
   attributes = []
 
-  if attachment.content_type_abbr
-    content = tag.abbr(attachment.content_type_abbr,
-                       class: "gem-c-attachment__abbr",
-                       title: attachment.readable_content_type)
+  if attachment.readable_content_type
```
benthorner: Appreciate it's not part of this part, but 'readable' confused me for a while until I realised it means 'human_readable' :-|.

spec/components/attachment_spec.rb
```diff
@@ -41,6 +40,16 @@ def component_name
...
+  it "can show file type that doesn't have an abbreviation" do
+    render_component(
+      attachment: {
+        title: "Attachment",
+        url: "attachment",
+        content_type: "text/plain",
+      },
+    )
+    expect(rendered).to match(/Plain Text/)
+  end
```
benthorner: Missing a new line here :-).

benthorner: Also, what do you think about a test for an unknown content type?

app/views/govuk_publishing_components/components/_attachment.html.erb
kevindew: I'm not particularly keen on the readable method names here. I don't really think human_readable helps either though as it gets a bit whacky when it's human_readable_number_of_pages 🤔 It might be that it's best to just do the view helper transformations in the view rather than presenter.

spec/components/attachment_spec.rb
kevindew: Thanks re new line.

I agree it'd be good to test unknown but as there's nothing to output there's not a clear assertion to make. I'll see if I can test that with nothing to show it doesn't show a metadata paragraph (which it currently shows and probably shouldn't)

app/views/govuk_publishing_components/components/_attachment.html.erb
benthorner: Ah, I see what you mean. Looking at it again, it seems the 'readable' prefix is somewhat unnecessary - there's no (unreadable ;-) number_of_pages, for example. Perhaps it could more usefully be 'content_type_name', as with 'content_type_abbr'. Up to you, though :-).

spec/components/attachment_spec.rb
benthorner: I was wondering about this - are we wanting to couple the file types we support in our app to the ones govspeak can render? If so, then perhaps it should throw an exception for something unexpected; if not, then maybe we should consider 'Unknown type' or something?

kevindew: I'd suggest they're only loosely coupled - since this should gradually reflect all GOV.UK attachments. It may well transpire that this reflects more content types than Content Publisher supports. Due to the loose coupling I don't think exception would be the right route for an unrecognised content_type.

I think saying "Unknown type" wouldn't provide much value to a user, so I think the right course would be to not show anything. That'd be consistent with just omitting content_type, which is supported.

app/views/govuk_publishing_components/components/_attachment.html.erb
kevindew: I've dropped the readable_* methods now

PR #853, opened 2019-05-13
Update to SassC Rails

### LINKED ISSUE #601: Switch from sass to sassc (and libsass)
https://github.com/sass/sassc-rails
https://github.com/sass/libsass

This should improve our Sass compile time by up to 4x. We've seen that since the Rails 5.1 upgrade development time is a lot slower. And with integration tests we had to bump poltergeist’s timeout limit.

We did this on Whitehall:
https://github.com/alphagov/whitehall/pull/2528
https://github.com/alphagov/whitehall/pull/2709

Need to be aware of:
* https://github.com/alphagov/whitehall/pull/2333
* Might need to avoid Ruby string concatenation in sass files

injms:
Ruby Sass is deprecated so needs to be replaced with something that is supported. [SassC Rails](https://github.com/sass/sassc-rails) is faster and a drop in replacement, according to its readme, and it seems to be both of these.

Fixes #601.

[Trello ticket](https://trello.com/c/V5y9atIh/80-draft-investigate-swapping-out-sass-rails-for-sassc-rails).

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -10,7 +10,7 @@ PATH
       rails (>= 5.0.0.1)
       rake
       rouge
-      sass-rails (>= 5.0.4)
+      sassc-rails (>= 2.0.1)
 
 GEM
   remote: https://rubygems.org/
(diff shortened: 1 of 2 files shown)


andysellick:
@injms also if this is faster, can we get this in `finder-frontend` too?

injms:
> @injms also if this is faster, can we get this in `finder-frontend` too?

Sure thing - I'm working on it.

injms:
> 🎉 🎉 🎉
> 
> Looks good. My local's still warning that Ruby Sass is deprecated, but I assume that's an automatic message? Any way of getting rid of that?

The Ruby Sass looks to be included because of the deprecated [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit) - I thought it best to replace that as a separate  task.

PR #856, opened 2019-05-15
Fix attachment rendering within Govspeak

kevindew:

This PR does a few things to demonstrate attachment components in a govspeak context (info in commits) and then looks at fixing a styling issue caused by specificity problems between govspeak and attachment component.

Before:

Screen Shot 2019-05-15 at 09 48 28

After:

Screen Shot 2019-05-15 at 09 49 29

### DIFF
--- a/app/views/govuk_publishing_components/component_guide/component_doc/_component_output.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/component_doc/_component_output.html.erb
@@ -1,7 +1,7 @@
 <% component = capture do %>
   <% if example.has_block? %>
     <%= render component_doc.partial_path, example.html_safe_data do %>
-      <%= example.block.html_safe %>
+      <%= ERB.new(example.block).result(binding).html_safe %>
     <% end %>
   <% else %>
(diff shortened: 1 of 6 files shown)


kevindew:
Thanks @alex-ju this is good for review now

app/views/govuk_publishing_components/component_guide/component_doc/_component_output.html.erb
```diff
@@ -1,15 +1,15 @@
 <% component = capture do %>
   <% if example.has_block? %>
     <%= render component_doc.partial_path, example.html_safe_data do %>
-      <%= ERB.new(example.block).result(binding).html_safe %>
+      <%= render inline: example.block %>
```
emmabeynon: Is the previous commit needed if we're changing this line again in this commit?

kevindew: Thanks for looking Emma.

Yeah I think this is useful as two commits. First one makes the change that we render the block. Second one changes the approach in multiple places.

PR #858, opened 2019-05-15
Add request different format to attachment component

alex-ju:
Add request different format section using the Details component instead of requiring JavaScript for revealing the additional content.

[Trello card](https://trello.com/c/OcuVomSb)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -10,6 +10,11 @@ $thumbnail-icon-border-colour: govuk-colour("grey-3");
 .gem-c-attachment {
   @include govuk-font(19);
   @include govuk-clearfix;
+  position: relative;
+
+  .govuk-details__summary {
+    @include govuk-font($size: 14);
+  }
 }
 
 .gem-c-attachment__thumbnail {
@@ -34,6 +39,10 @@ $thumbnail-icon-border-colour: govuk-colour("grey-3");
   stroke: $thumbnail-icon-border-colour;
 }
 
+.gem-c-attachment__details {
+  padding-left: $thumbnail-width + $thumbnail-border-width * 2 + govuk-spacing(5);
+}
+
 .gem-c-attachment__title {
   @include govuk-font($size: 27);
   margin: 0 0 govuk-spacing(3);
@@ -42,6 +51,10 @@ $thumbnail-icon-border-colour: govuk-colour("grey-3");
 .gem-c-attachment__metadata {
   @include govuk-font($size: 14);
   margin: 0 0 govuk-spacing(3);
+
+  &:last-of-type {
+    margin-bottom: 0;
+  }
 }
 
 .gem-c-attachment__abbr {
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -1,7 +1,7 @@
 <%
   attachment = GovukPublishingComponents::Presenters::Attachment.new(attachment)
   target ||= "_self"
-  hide_help_text ||= false
+  hide_opendocument_metadata ||= false
   attributes = []
 
   if attachment.content_type_name
@@ -57,7 +57,7 @@
       <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
     <% end %>
 
-    <% unless hide_help_text %>
+    <% unless hide_opendocument_metadata %>
       <% if attachment.opendocument? %>
         <%= tag.p class: "gem-c-attachment__metadata" do %>
           <%= t("components.attachment.opendocument_html", target: target) %>
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -46,20 +46,31 @@
     <% end %>
   <% end %>
 
-  <%= tag.h2 class: "gem-c-attachment__title" do %>
-    <%= link_to attachment.title, attachment.url,
-          class: "govuk-link",
-          target: target %>
-  <% end %>
+  <%= tag.div class: "gem-c-attachment__details" do %>
+    <%= tag.h2 class: "gem-c-attachment__title" do %>
+      <%= link_to attachment.title, attachment.url,
+            class: "govuk-link",
+            target: target %>
+    <% end %>
 
-  <% if attributes.any? %>
-    <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
-  <% end %>
+    <% if attributes.any? %>
+      <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
+    <% end %>
 
-  <% unless hide_help_text %>
-    <% if attachment.opendocument? %>
-      <%= tag.p class: "gem-c-attachment__metadata" do %>
-        <%= t("components.attachment.opendocument_html", target: target) %>
+    <% unless hide_help_text %>
+      <% if attachment.opendocument? %>
+        <%= tag.p class: "gem-c-attachment__metadata" do %>
+          <%= t("components.attachment.opendocument_html", target: target) %>
+        <% end %>
+      <% end %>
+    <% end %>
+
+    <% if attachment.alternative_format_contact_email %>
+      <%= tag.p t("components.attachment.request_format_text"), class: "gem-c-attachment__metadata" %>
+      <%= render "govuk_publishing_components/components/details", {
+        title: t("components.attachment.request_format_cta")
+      } do %>
+        <%= t("components.attachment.request_format_details_html", alternative_format_contact_email: attachment.alternative_format_contact_email) %>
       <% end %>
     <% end %>
   <% end %>
--- a/app/views/govuk_publishing_components/components/docs/attachment.yml
+++ b/app/views/govuk_publishing_components/components/docs/attachment.yml
@@ -69,3 +69,12 @@ examples:
         filename: BEIS_Information_Asset_Register_.ods
         content_type: application/vnd.oasis.opendocument.spreadsheet
         file_size: 20000
+  with_contact_email:
+    data:
+      attachment:
+        title: "Department for Transport information asset register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/747661/department-for-transport-information-asset-register.csv
+        filename: department-for-transport-information-asset-register.csv
+        content_type: application/pdf
+        file_size: 20000
+        alternative_format_contact_email: defra.helpline@defra.gsi.gov.uk
--- a/app/views/govuk_publishing_components/components/docs/attachment.yml
+++ b/app/views/govuk_publishing_components/components/docs/attachment.yml
@@ -53,7 +53,7 @@ examples:
         filename: 20130110_Iraq_wave01.txt
         content_type: text/plain
         file_size: 108515
-      hide_help_text: true
+      hide_opendocument_metadata: true
   embedded_in_govspeak:
     description: |
       This component can be embedded in Govspeak with the `[Attachment:]` code.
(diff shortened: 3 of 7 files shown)


spec/components/attachment_spec.rb
```diff
@@ -74,6 +74,18 @@ def component_name
     assert_select "a[href='https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software']"
   end
 
+  it "shows request a different format section if a contact email is provided" do
```
benthorner: Might be clearer to say 'section to request a different form' here.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -46,20 +47,31 @@
...
-  <% unless hide_help_text %>
-    <% if attachment.opendocument? %>
-      <%= tag.p class: "gem-c-attachment__metadata" do %>
-        <%= t("components.attachment.opendocument_html", target: target) %>
+    <% if contact_email %>
+      <%= tag.p t("components.attachment.request_format_text"), class: "gem-c-attachment__metadata" %>
+      <%= render "govuk_publishing_components/components/details", {
+        title: t("components.attachment.request_format_cta")
+      } do %>
+        <%= t("components.attachment.request_format_details_html", contact_email: contact_email) %>
```
kevindew: This should be inside the help text conditional - as that should include all supplementary data beyond the file name, content_type, size and number of pages.

Side note I don't think help_text is the best name for that block so any suggestions on how we can either name it better or document it better are welcomed.

app/views/govuk_publishing_components/components/docs/attachment.yml
```diff
@@ -54,3 +54,12 @@ examples:
...
       hide_help_text: true
+  with_contact_email:
+    data:
+      attachment:
+        title: "Department for Transport information asset register"
+        url: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/747661/department-for-transport-information-asset-register.csv
+        filename: department-for-transport-information-asset-register.csv
+        content_type: application/pdf
+        file_size: 20000
+      contact_email: defra.helpline@defra.gsi.gov.uk
```
kevindew: This should be part of the attachment block of data, as that is data that would be included in rendering this via govspeak

config/locales/en.yml
```diff
@@ -25,6 +25,9 @@ en:
   components:
     attachment:
       opendocument_html: "This file is in an <a href='https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software' target=%{target} class='govuk-link'>OpenDocument</a> format"
+      request_format_text: "This file may not be suitable for users of assistive technology."
+      request_format_cta: "Request a different format"
+      request_format_details_html: "If you use assistive technology and need a version of this document in a more accessible format, please email <a href='mailto:%{contact_email}' class='govuk-link'>%{contact_email}</a>. Please tell us what format you need. It will help us if you say what assistive technology you use."
```
kevindew: we should pass in the target attribute here for consistency

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -2,6 +2,7 @@
   attachment = GovukPublishingComponents::Presenters::Attachment.new(attachment)
   target ||= "_self"
   hide_help_text ||= false
+  contact_email ||= nil
```
kevindew: This should be managed by the attachment presenter

app/views/govuk_publishing_components/components/docs/attachment.yml
alex-ju: I was wondering about that. Is the attachment block mirroring some 'model' structure? or is it freestyle in terms of naming things?

kevindew: Ideally it should follow the precedent set by existing attachment data sent to govspeak to avoid unnecessary breaks in consistency and backwards compatibility. However the previous way is mostly Whitehall internal data and there is not a backing of a schema or similar to enforce it.

Ideally in the future they should be done by a schema though.

The Whitehall naming is alternative_format_contact_email: https://github.com/alphagov/whitehall/blob/5c760eea912320c5a80ef59791df47161d889f2f/app/helpers/govspeak_helper.rb#L22-L24

It doesn't look like this was ever fully implemented into govspeak though, as a single address was hardcoded: https://github.com/alphagov/govspeak/blob/f626b20e9e0a742cc4c9f4a946a80c87aa63a8da/lib/govspeak/presenters/attachment_presenter.rb#L102-L104

app/views/govuk_publishing_components/components/_attachment.html.erb
benthorner: It might help to know that we're only going to pass the `contact_email` attribute from Content Publisher when we actually want to render it. Not really keen on the 'pass all the data and then hide it' approach, as it seems pretty coupled to our use case. Keeping the conditional just around the open document rendering might also make it easier to rename, perhaps to a more positive 'show open document' kind of attribute?

soniaturcotte:
> Looks good to me - only thing is perhaps to assert that we don't show the alternative format by default? Also, I notice we've got a vertical line to the left of the secondary text, which is another departure from the current look on GOV.UK. Is @soniaturcotte OK with this?

This is the standard design system component (with the arrow and line) however it's not what currently exists on frontend. I’d like to double check with Mark H (in absense of Mia) before we change. Downside is that it won’t be consistent across pages.

app/views/govuk_publishing_components/components/docs/attachment.yml
alex-ju: Thanks for pointing to those. The Whitehall bit was missing from my puzzle.

app/views/govuk_publishing_components/components/_attachment.html.erb
kevindew: For govspeak it's necessary to use the all data approach as we don't know if it'll be used by an Attachment or an AttachmentLink. So there is already precedent for it. I can understand your concern, but I find it easier to think of the attachment as a model of data not a means of options.

With regards to not passing in contact_email in some cases, that seems a reversal of the precedents set so far where we have one method used both attachments and attachments link. So it'd be good if didn't intentionally veer into inconsistency without consideration.

The expectation of the email is that if one is not provided it falls back to a default one - not sure if we'll provide that or govspeak but that's what was previously part of govspeak.

On a side note, positive attributes - like you're suggestion of show - seem to not be at all common in govuk_publishing_components. Possibly due to the more verbose syntax needed than the or equal operator. When Emma and I looked into it there only seemed to be hide ones so did not want to introduce an inconsistency.

config/locales/en.yml
alex-ju: I'm wondering is we shouldn't default to '_blank' for this one instead of sharing the same target value with the opendocument link

kevindew: I'm not quite sure what you're suggesting in your comment, but since it's a mailto I'm having second thoughts about my comment. So I now think it's fine with no target.

app/views/govuk_publishing_components/components/_attachment.html.erb
benthorner: So I think we're both agreeing on the 'model of data' approach, but there's already quite a precedent for only rendering certain things if certain pieces of that data are present.

The default email is also something that I felt was more a concern of our app than a rendering issue, as it seems more like a content issue than a rendering one. Since we have to do some presence checks in our app anyway, it seemed better to have all the logic in one place.

I found a way to modify the helper stuff we have to both simplify them and only pass the contact email where necessary. Admittedly, I did do this on the assumption we would be doing all the decision logic in our app. I'll try and get a PR up so you can see the two together...

config/locales/en.yml
alex-ju: target `_blank` has effect if your email address is set to be a web-based email client (e.g. Gmail) so target still makes sense, since you will be directed to the email provider and loose page context if it's not set. Doesn't have an effect for desktop-based clients though.

alex-ju:
@kevindew addressed all the comments except for the one regarding the help text which is still being discussed.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -10,6 +10,11 @@ $thumbnail-icon-border-colour: govuk-colour("grey-3");
 .gem-c-attachment {
   @include govuk-font(19);
   @include govuk-clearfix;
+  position: relative;
+
+  .govuk-details__summary {
+    @include govuk-font($size: 14);
```
benthorner: Given @soniaturcotte's latest comments, do we want to use a different font here?

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -46,20 +46,31 @@
...
+    <% end %>
 
-  <% if attributes.any? %>
-    <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
-  <% end %>
+    <% if attributes.any? %>
+      <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
+    <% end %>
+
+    <% unless hide_help_text %>
```
benthorner: How about 'hide_opendocument_metadata'?

config/locales/en.yml
```diff
@@ -25,6 +25,9 @@ en:
   components:
     attachment:
       opendocument_html: "This file is in an <a href='https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software' target=%{target} class='govuk-link'>OpenDocument</a> format"
+      request_format_text: "This file may not be suitable for users of assistive technology."
+      request_format_cta: "Request a different format"
+      request_format_details_html: "If you use assistive technology and need a version of this document in a more accessible format, please email <a href='mailto:%{alternative_format_contact_email}' target='_blank' class='govuk-link'>%{alternative_format_contact_email}</a>. Please tell us what format you need. It will help us if you say what assistive technology you use."
```
benthorner: Should we make the target configurable, like we do for the opendocument guidance?

app/views/govuk_publishing_components/components/_attachment.html.erb
benthorner: Just writing up here: we discussed this in-person and agreed to continue as-is, after considering the pros and cons of each approach.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
soniaturcotte: No same font, just don't have to remove the arrow and grey border inset

config/locales/en.yml
alex-ju: As discussed, I think we'd want to enforce this to be blank since to make sure users are not loosing the context (the page they're on at the point they request another format).

app/views/govuk_publishing_components/components/_attachment.html.erb
alex-ju: Updated with the suggested name and rebased.

spec/components/attachment_spec.rb
```diff
@@ -74,14 +74,26 @@ def component_name
     assert_select "a[href='https://www.gov.uk/guidance/open-document-format-odf-guidance-for-uk-government/overview-of-productivity-software']"
   end
 
+  it "shows section to request a different form if a contact email is provided" do
```
benthorner: I think we mean 'format' here.

spec/components/attachment_spec.rb
```diff
@@ -74,14 +74,26 @@ def component_name
...
+        title: "Attachment",
+        url: "attachment",
+        content_type: "application/vnd.oasis.opendocument.spreadsheet",
+        alternative_format_contact_email: "defra.helpline@defra.gsi.gov.uk"
+      },
+    )
+    assert_select "a[href='mailto:defra.helpline@defra.gsi.gov.uk']"
+  end
+
   it "does not show help text if disabled" do
```
benthorner: We probably want to rephrase this to reflect the rename.

spec/components/attachment_spec.rb
alex-ju: Good spot.

spec/components/attachment_spec.rb
alex-ju: Yep, updated

PR #861, opened 2019-05-15
Add SearchResultsPage schema

sihugh:
This adds a naive implementation of https://schema.org/SearchResultsPage which
permits us to add basic schema information to finder pages.  Using this approach
also adds other page metadata (contained in the machine_readable_metadata partial).

We're paying more attention to how these are indexed for search because finders
have different requirements for search visibility.  Specialist finders need
to be findable via Google because these search pages are a critical for a subset
of users.

More general search pages have a more nuanced need.  We generally want people
to avoid navigating from Google search to an internal search page unless it
is for a specific reason.

We have already excluded search results page with keywords from being indexed,
and we similarly exclude everything past the first page.  We'll probably iterate
this approach.

https://trello.com/c/OVkRpKDk/735-add-metadata-for-finders
https://trello.com/c/4EAytHVV/747-plan-noindexing-of-finder-pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add SearchResultsPage schema.org schema (PR #861)
+
 ## 16.19.0
 
 * Adds support to allow parent finder to show as breadcrumbs (PR #831)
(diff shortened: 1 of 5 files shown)


sihugh:
That's a really good point, thanks @barrucadu . I'll write something up for that.

PR #872, opened 2019-05-20
Update inline button style

vanitabarrett:
Update the inline button style so they appear adjacent to each other for as long as possible before stacking one on top of the other, so it takes up less vertical space.

## Before
Screen Shot 2019-05-20 at 14 42 29

## After
Screen Shot 2019-05-20 at 14 42 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Update inline button styling on mobile (PR #872)
+
 ## 16.21.0
 
 * Add request different format section to attachment component (PR #858)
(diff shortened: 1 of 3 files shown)


vanitabarrett:
[Implement design tweaks to the cookie banner](https://trello.com/c/keYvts3O/145-implement-design-tweaks-to-the-cookie-banner)

injms:
Whilst this isn't a big change, it would affect apps that are already using inline buttons. Do you think it's worth having this layout as a new option?

vanitabarrett:
@injms I don't think anywhere else uses inline buttons (I only added the option last week for the new cookie banner).

injms:
@vanitabarrett ah, okay! Thanks for explaining 😄

vanitabarrett:
@injms No problem, thanks for the speedy review!

PR #875, opened 2019-05-21
Update dependencies

injms:
Taking advantage of the amazingly named twiddle-waka (`~>`) to update dependencies.

Most of the dependencies seem to be patch version bumps. I've run all the tests locally and checked a large number of the components to make sure they're all okay.


injms:
Hmm, binning this as the linter isn't liking things anymore.

PR #876, opened 2019-05-21
Update contents-list component to use govuk-frontend

alex-ju:
This PR updates the contents list component to use govuk-frontend instead of govuk_frontend_toolkit.

**Notes:**
`$gutter` = `govuk-spacing(6)` = `30px`

[Trello card](https://trello.com/c/ajPzDAOX)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -4,10 +4,10 @@
   // Always render the contents list above a
   // back to contents link
   position: relative;
-  margin: 0 0 $gutter-two-thirds 0;
+  margin: 0 0 govuk-spacing(4) 0;
   z-index: 1;
-  background: $white;
-  box-shadow: 0 20px 15px -10px $white;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -49,17 +49,17 @@
...
-  @include media(tablet) {
-    padding-top: $gutter-one-quarter;
+  @include govuk-media-query($from: tablet) {
+    padding-top: govuk-spacing(6) / 4;
   }
 }
 
 .gem-c-contents-list__list-item--dashed {
-  $contents-spacing: $gutter-half + 10;
+  $contents-spacing: govuk-spacing(3) + govuk-spacing(2);
```
nickcolley: Not sure if possible, but perhaps these can be squished together now?

alex-ju: Yep, they can..  was looking to see if there's any reason of them being split (such as making it more obvious where the value comes from, but it doesn't seem to be the case.

PR #878, opened 2019-05-21
Update error alert component to use govuk-frontend

alex-ju:
This PR:
- updates the error-alert component to use govuk-frontend instead of govuk_frontend_toolkit
- uses `govuk-spacing` mixin instead of locally defined `gem-spacing-scale-`
- updates component name for consistency

[Trello card](https://trello.com/c/ajPzDAOX)


alex-ju:
@sihugh good catch, updated! the downsides of copy-paste..

PR #879, opened 2019-05-22
Add SearchAction for manuals

sihugh:
Manuals like [The Highway Code](https://www.gov.uk/guidance/the-highway-code) have an embedded search box which allows searches scoped to that manual.

By advertising this in our machine readable metadata it's more obvious to external consumers of the site that these things are available.

This is part of the work to make our search more visible to external consumers of the site.

Possibly useful to review by commit because the majority of the change is extracting the `potentialSearchAction` schema from organisations to make it available for use in manuals.

https://trello.com/c/4EAytHVV/747-plan-indexing-of-finder-pages

### DIFF
--- a/spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
+++ b/spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
@@ -43,7 +43,7 @@ RSpec.describe GovukPublishingComponents::Presenters::SchemaOrg do
 
       search_action = {
         "@type": "SearchAction",
-        "target": "http://www.dev.gov.uk/search/all?keywords={query}&manual=%2Fguidance%2Fplane-manual",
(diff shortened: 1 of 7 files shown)


spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
```diff
@@ -53,7 +53,7 @@
 
       search_action = {
         "@type": "SearchAction",
-        "target": "http://www.dev.gov.uk/search/all?keywords={query}&organisations[]=ministry-of-magic",
+        "target": "http://www.dev.gov.uk/search/all?keywords={query}&organisations%5B%5D=ministry-of-magic",
```
bilbof: Would it be worth also adding a unit test for the `PotentialSearchActionSchema` class?

sihugh: I thought about it, but the schema bits are generally pretty trivial and they're exercised pretty well by the `schema_org_spec`.

PR #880, opened 2019-05-22
Refactor step by step navigation logic

oscarwyatt:
Makes it easier to work with


oscarwyatt:
@alex-ju I hadn't thought of it being a breaking change but I suppose it could (shouldn't, but could). I'm happy to update the PR to indicate that but I don't really have time to prepare PRs - is it better to just shelve the whole thing then?

PR #883, opened 2019-05-22
Fix gemspec comment

sihugh:
The comment that applied to the inclusion of the webmock gem had become
decoupled.  I've re-attached them and changed the wording of the comment
so that it's hopefully clearer that they belong together.


sihugh:
I didn't think it was @injms!

PR #892, opened 2019-05-28
Add machine readable breadcrumbs for step by steps

sihugh:
We've removed the `partOf` metadata for step by steps which linked back to the "parent" because we were using a partial HowTo schema which is not supported by Google's search indexing process.

This PR adds in machine readable breadcrumbs for pages that show the step by step header which should reinstate some of that experience for search engines.

I didn't want to assert too much on the exact format of the breadcrumbs schema output as they're defined in https://github.com/alphagov/govuk_publishing_components/blob/e3fd7fa321f4a38218a400fa795bb120c4681a4b/lib/govuk_publishing_components/presenters/breadcrumbs.rb (not the step by step header itself) and are used in the contextual breadcrumbs component too.

It's implementing https://schema.org/BreadcrumbList which looks like this in the examples:

Screenshot 2019-05-28 14 17 31

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ## Unreleased
 
+* Add machine readable breadcrumbs to the step by step header (PR #892)
+
 ## 16.25.0
 
 * Support data attributes in modal dialogue (PR #890)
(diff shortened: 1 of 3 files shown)


andysellick:
@sihugh can you add in a link to this PR description to the schema stuff? I just want to sanity check it.

sihugh:
@andysellick Is that what you meant?  You can check how it our breadcrumbs look at the moment by checking a page with breadcrumbs in the structured data tool, like: https://search.google.com/structured-data/testing-tool/u/0/#url=https%3A%2F%2Fwww.gov.uk%2Feducation%2Frunning-and-managing-a-school

PR #896, opened 2019-05-29
Add #sanitize to code examples containing HTML

koetsier:
Copy-pasting examples with attributes containing HTML caused
those attributes to be rendered incorrectly because Rails automatically
escapes all strings.

This commit finds all attributes containing HTML and adds
the 'sanitize' helper to the code examples allowing them to be copy-pasted as-is.

Screenshot:

image

### DIFF
--- a/app/models/govuk_publishing_components/component_example.rb
+++ b/app/models/govuk_publishing_components/component_example.rb
@@ -29,7 +29,18 @@ module GovukPublishingComponents
     end
 
     def pretty_data
-      JSON.pretty_generate(data).gsub('\\n', "\n    ").gsub(/"(\w*)":/, '\1:').strip
+      json_key_regex = /"(\w*)":/ # matches quoted keys ending with a colon, i.e. "key":
+      output = JSON.pretty_generate(data).gsub('\\n', "\n    ").gsub(json_key_regex, '\1:')
+
+      quoted_string_regex = /"((?:[^"\\]|\\.)*)"/ # matches "some text" - ignores escaped quotes, i.e. \"
+      output.gsub(quoted_string_regex) do |group|
+        match = Regexp.last_match[1]
+        contains_html?(match) ? "sanitize(#{group})" : group
+      end
+    end
+
+    def contains_html?(input)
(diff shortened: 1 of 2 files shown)


koetsier:
As part of a PR I wrote recently, I copy-pasted the example in https://finder-frontend.herokuapp.com/component-guide/notice/with_markup_in_the_title into Whitehall and it showed me a bunch of unescaped HTML. This was unexpected behaviour because the example suggests that that snippet of code can be used as-is. In fact behind the scenes - when it is rendered in the documentation - ```.html_safe``` is appended to all strings. It caused me to think that I had done something wrong or had the wrong version of the publishing_components gem and I wasted considerable time.

If HTML is not expected to be included, then it would make more sense not to have any HTML in the examples. In fact the notice example I mentioned explicitly mentions markup in the title to be a feature and the only way to make that happen is by marking that string to be safe.

alex-ju:
@koetsier if the data doesn't come from storage and it's not elaborated enough to be captured before the use, then rails helpers should be use to mitigate a potential XSS vulnerability.

This is how I think the example you mentioned should look like:

```
<%= render "govuk_publishing_components/components/notice", {
  title: tag.span("Advisory Committee on Novel Foods and Processes has a ") + link_to("separate website", "http://www.food.gov.uk/acnfp")
} %>
```

I agree with you that these examples should be 'ready to be used' and 'work out of the box', but I don't think this is the approach to move forward.

koetsier:
I see your point and using helpers would mitigate XSS, although your example still contain the possibility of an XSS attack, i.e.:
```link_to("separate website", "javascript:alert(1)")``` will execute when clicking the link.

Maybe ```sanitize``` could be a solution? https://apidock.com/rails/ActionView/Helpers/SanitizeHelper/sanitize? It won't accept <script> tags and works with a whitelist of tags it accepts. 

example:
image

It also has the method name at the front making it a bit clearer what we're doing

alex-ju:
`sanitize` seems like a sensible middle ground and allow the automation of how we display code examples

PR #900, opened 2019-05-30
Add the closed step by step for secondary content

DilwoarH:
## What
    
Make it possible for the step by step side bar to appear in the side navigation for all content that uses the contextual navigation.
    
(Note: we will keep all content the same for the breadcrumbs and side nav as for step by steps elsewhere, we are not changing 'part of' for example.)
    
Update developer documentation with these changes.
    
**Logic:**

- If only one secondary step by step and no other step by steps linked - show secondary step by step and expand it
- If multiple secondary step by steps linked and no other step by steps linked - show "Part of" step by step with secondary step by steps
- If multiple secondary step by steps exceed 5 - do not show anything
- If one primary step by step and one/multiple secondary - Show primary and hide secondary
- If multiple primary and one/multiple secondary - Show "Part of" step by step (with only primary)
- If you are on an active step by step - no secondary step by steps should show

If the content is tagged to the business finder and is secondary content on a step by step then the secondary content takes precedence.
    
## Why
We want to add a new feature to 'step by step publisher' that makes it possible to add step by step navigation to secondary content. (Secondary content = content that is about the service but is not useful enough to include in the step by step).
    
**MVP = add the closed step by step to the side nav for these pages**
    
**Nice to have =  make is possible to assign a step in the step by step to open for that piece of secondary content, where relevant.**

## Before
Screen Shot 2019-06-11 at 10 31 39

## After
Screen Shot 2019-06-11 at 12 53 00

## Multiple secondary step by steps
Screen Shot 2019-06-11 at 09 57 50

## Combined step by steps < 5 (Secondary, related)
Screen Shot 2019-06-11 at 11 02 03

## Combined step by steps > 5 (Secondary, related)
Screen Shot 2019-06-11 at 10 31 39

Ref: https://trello.com/c/eC8EXI21/120-step-by-step-add-the-closed-step-by-step-to-the-side-nav-for-secondary-content

Related PR: https://github.com/alphagov/collections-publisher/pull/618

### DIFF
--- a/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
+++ b/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
@@ -21,18 +21,30 @@ module GovukPublishingComponents
         end
       end
 
+      def secondary_step_by_steps
+        @secondary_step_by_steps ||= parsed_secondary_to_step_navs.map do |step_nav|
+          StepByStepModel.new(step_nav)
+        end
+      end
+
       def show_sidebar?
         show_header? && current_step_nav.steps.present?
       end
 
       def show_header?
-        step_navs.count == 1 || active_step_by_step?
+        step_navs.count == 1 || active_step_by_step? || show_secondary_step_by_step?
       end
 
       def show_related_links?
-        return true if active_step_by_step?
-
-        step_navs.any? && step_navs.count < 5
+        if active_step_by_step?
+          true
+        elsif step_navs.any? && step_navs.count < 5
+          true
+        elsif show_related_links_for_secondary_step_by_steps?
+          true
+        else
+          false
+        end
       end
 
       def show_also_part_of_step_nav?
@@ -40,7 +52,16 @@ module GovukPublishingComponents
       end
 
       def related_links
-        step_by_step_navs = active_step_by_step? ? [active_step_by_step] : step_navs
+        step_by_step_navs = if active_step_by_step?
+                              [active_step_by_step]
+                            elsif step_navs.any?
+                              step_navs
+                            elsif show_related_links_for_secondary_step_by_steps?
+                              secondary_step_by_steps
+                            else
+                              []
+                            end
+
         format_related_links(step_by_step_navs)
       end
 
@@ -71,6 +92,10 @@ module GovukPublishingComponents
         end
       end
 
+      def primary_step_by_steps?
+        step_navs_combined_list.any?
+      end
+
       def active_step_by_step?
         active_step_nav_content_id.present? && active_step_by_step.present?
       end
@@ -81,14 +106,34 @@ module GovukPublishingComponents
         @active_step_navs.first
       end
 
+      def secondary_step_by_step?
+        secondary_step_by_steps.any?
+      end
+
+      def secondary_step_by_step
+        secondary_step_by_steps.first
+      end
+
+      def show_secondary_step_by_step?
+        !primary_step_by_steps? && secondary_step_by_step? && secondary_step_by_steps.count === 1
+      end
+
+      def show_related_links_for_secondary_step_by_steps?
+        !primary_step_by_steps? && secondary_step_by_step? && secondary_step_by_steps.count < 5
+      end
+
     private
 
       attr_reader :content_item, :current_path
 
       def current_step_nav
-        return active_step_by_step if active_step_by_step?
-
-        step_navs.first
+        if active_step_by_step?
+          active_step_by_step
+        elsif primary_step_by_steps?
+          step_navs.first
+        elsif show_secondary_step_by_step?
+          secondary_step_by_step
+        end
       end
 
       def active_step_nav_content_id
@@ -112,6 +157,10 @@ module GovukPublishingComponents
 
       def parsed_related_to_step_navs
         content_item.dig("links", "related_to_step_navs").to_a
+      end
+
+      def parsed_secondary_to_step_navs
+        content_item.dig("links", "secondary_to_step_navs").to_a
       end
 
       def configure_for_sidebar(step_nav_content)
--- a/spec/features/contextual_navigation_spec.rb
+++ b/spec/features/contextual_navigation_spec.rb
@@ -120,6 +120,14 @@ describe "Contextual navigation" do
     and_no_step_by_step_header
   end
 
+  scenario "There's 2 primary step by step, 1 secondary step by step and I am interacting with one of the primary" do
+    given_there_are_two_primary_step_by_steps_and_one_secondary_to_step_navs
+    and_i_visit_that_page_by_clicking_on_a_step_by_step_link
+    then_i_see_the_step_by_step_that_i_am_interacting_with
+    and_i_see_the_other_step_by_step_as_an_also_part_of_list
+    and_i_dont_see_the_secondary_step_by_step_in_the_also_part_of_list
+  end
+
   include GdsApi::TestHelpers::ContentStore
 
   def given_theres_a_page_with_a_step_by_step
@@ -168,6 +176,21 @@ describe "Contextual navigation" do
     content_store_has_random_item(links: {
       secondary_to_step_navs: 1.times.map { random_step_nav_item("step_by_step_nav") },
       part_of_step_navs: part_of_step_navs
+    })
+  end
+
+  def given_there_are_two_primary_step_by_steps_and_one_secondary_to_step_navs
+    part_of_step_navs = 2.times.map { random_step_nav_item("step_by_step_nav") }
+    part_of_step_navs[0]["title"] = "PRIMARY STEP BY STEP - NOT INTERACTING WITH"
+    part_of_step_navs[1]["content_id"] = "8ad999bd-8603-40eb-97c0-999cb22047cd"
+    part_of_step_navs[1]["title"] = "PRIMARY STEP BY STEP - INTERACTING WITH"
+
+    secondary_to_step_navs = 1.times.map { random_step_nav_item("step_by_step_nav") }
+    secondary_to_step_navs[0]["title"] = "SECONDARY STEP BY STEP"
+
+    content_store_has_random_item(links: {
+      part_of_step_navs: part_of_step_navs,
+      secondary_to_step_navs: secondary_to_step_navs
     })
   end
 
@@ -301,6 +324,12 @@ describe "Contextual navigation" do
       expect(page).to have_content('Also part of')
       expect(page).to have_content("PRIMARY STEP BY STEP - NOT INTERACTING WITH")
       expect(page).not_to have_content("PRIMARY STEP BY STEP - INTERACTING WITH")
+    end
+  end
+
+  def and_i_dont_see_the_secondary_step_by_step_in_the_also_part_of_list
+    within '.gem-c-step-nav-related:last-child' do
+      expect(page).not_to have_content("SECONDARY STEP BY STEP")
     end
   end
 
--- a/spec/features/contextual_navigation_spec.rb
+++ b/spec/features/contextual_navigation_spec.rb
@@ -64,6 +64,41 @@ describe "Contextual navigation" do
     and_the_parent_based_breadcrumbs
   end
 
+  scenario "There's a secondary step by step list and no primary step by step list" do
+    given_theres_a_page_with_a_secondary_step_by_step
+    and_i_visit_that_page
+    then_i_see_the_step_by_step
+    and_the_step_by_step_header
+  end
+
+  scenario "There's between 2-5 secondary step by step lists and no primary step by step list" do
+    given_there_are_two_secondary_step_by_step_lists
+    and_i_visit_that_page
+    then_i_just_see_the_step_by_step_related_links
+  end
+
+  scenario "There's 6 or more secondary step by step lists and no primary step by step list" do
+    given_there_are_six_secondary_step_by_step_lists
+    and_i_visit_that_page
+    then_theres_no_step_by_step_at_all
+    and_no_step_by_step_header
+  end
+
+  scenario "There's 3 secondary step by steps and 2 primary step by step lists" do
+    given_there_are_three_secondary_step_by_steps_and_two_primary_to_step_navs
+    and_i_visit_that_page
+    then_i_just_see_the_step_by_step_related_links_with_just_two_links
+    and_i_dont_see_the_secondary_step_by_step_related_links
+  end
+
+  scenario "There's 1 primary step by step and 1 secondary step by step" do
+    given_there_are_one_primary_step_by_steps_and_one_secondary_to_step_navs
+    and_i_visit_that_page
+    then_i_see_the_step_by_step
+    and_the_step_by_step_header
+    then_i_see_the_primary_step_by_step
+  end
+
   include GdsApi::TestHelpers::ContentStore
 
   def given_theres_a_page_with_a_step_by_step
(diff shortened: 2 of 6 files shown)


leenagupte:
@DilwoarH Can you look at why the tests are failing?

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -40,7 +47,14 @@ def show_also_part_of_step_nav?
       end
 
       def related_links
-        step_by_step_navs = active_step_by_step? ? [active_step_by_step] : step_navs
+        step_by_step_navs = if active_step_by_step?
+                              [active_step_by_step]
+                            elsif secondary_step_by_step?
+                              step_navs_combined_list
```
leenagupte: `step_navs_combined_list` also contains `related_to_step_navs`.  Are `related_to_step_navs` supposed to show up in `related_links`? What happens if a content item is "related_to" one step nav and "secondary content" on another?

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,25 @@ def active_step_by_step
...
 
+      def secondary_step_by_step?
+        secondary_to_step_navs.any?
+      end
+
+      def secondary_step_by_step
+        secondary_to_step_navs.first
+      end
+
+      def show_secondary_step_by_step?
```
leenagupte: Is this function needed? Could it be replaced with `

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -103,6 +130,7 @@ def step_navs_combined_list
         step_nav_list = []
         step_nav_list += step_navs if step_navs.any?
         step_nav_list += related_to_step_navs if related_to_step_navs.any?
+        step_nav_list += secondary_to_step_navs if secondary_to_step_navs.any?
```
leenagupte: Adding `secondary_to_step_navs` here also means that `show_also_part_of_step_nav?` could start returning true as `also_part_of_step_nav` will also contain the secondary step nav.

I think we should seriously consider refactoring this module to make the step by step rules easier to understand.

spec/features/contextual_navigation_spec.rb
```diff
@@ -64,6 +64,27 @@
...
+    and_i_visit_that_page
+    then_i_see_the_step_by_step
+    and_the_step_by_step_header
+  end
+
+  scenario "There's between 2-5 secondary step by step lists" do
+    given_theres_are_two_secondary_step_by_step_lists
+    and_i_visit_that_page
+    then_i_just_see_the_step_by_step_related_links
+    and_no_step_by_step_header
```
leenagupte: This step isn't needed as the expectation is already included in: `then_i_just_see_the_step_by_step_related_links`

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: In the current solution, both will show up. In my opinion, this is the correct behaviour. The user should be able to see the secondary step navs linked as well as any related step navs.

leenagupte: Is this something we should double check with Sam?

DilwoarH: Yes, I will raise with Sam.

DilwoarH:
Agreed logic:
- If only one secondary step by step and no other step by steps linked - show secondary step by step and expand it
- If multiple secondary step by steps linked and no other step by steps linked - show "Part of" step by step with secondary step by steps
- If one primary step by step and one/multiple secondary - Show primary and hide secondary
- If multiple primary and one/multiple secondary - Show "Part of" step by step (with only primary)

DilwoarH:
@leenagupte - I have updated the component to match the new logic. I have also added some missing tests.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -40,7 +47,14 @@ def show_also_part_of_step_nav?
       end
 
       def related_links
-        step_by_step_navs = active_step_by_step? ? [active_step_by_step] : step_navs
+        step_by_step_navs = if active_step_by_step?
+                              [active_step_by_step]
+                            elsif show_related_links_for_secondary_step_by_steps?
+                              secondary_step_by_steps
+                            else
+                              step_navs
```
leenagupte: Should this logic be the other way around? So, you check if there are any primary step by steps first, and if there aren't then see about finding any secondary step navs?

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,33 @@ def active_step_by_step
         @active_step_navs.first
       end
 
+      def secondary_step_by_steps
```
leenagupte: Is this method needed? I think you can just reuse `secondary_to_step_navs`.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,33 @@ def active_step_by_step
...
+        step_navs_combined_list.count === 0 && secondary_step_by_step? && secondary_to_step_navs.count < 5
+      end
+
     private
 
       attr_reader :content_item, :current_path
 
       def current_step_nav
         return active_step_by_step if active_step_by_step?
+        return secondary_step_by_step if show_secondary_step_by_step?
```
leenagupte: Also here, shouldn't the secondary step navs only be evaluated only if there aren't any primary step navs?

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,33 @@ def active_step_by_step
...
+      def secondary_step_by_step?
+        secondary_step_by_steps.any?
+      end
+
+      def secondary_step_by_step
+        secondary_to_step_navs.first
+      end
+
+      def show_secondary_step_by_step?
+        step_navs_combined_list.count === 0 && secondary_step_by_step? && secondary_step_by_steps.count === 1
```
leenagupte: I think for ease of reading, `step_navs_combined_list.count === 0` could be moved into the `secondary_step_by_step?` method, e.g:

`step_navs_combined_list.count === 0 && secondary_step_by_steps.any?`

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: This is fine because there is a check within the `show_related_links_for_secondary_step_by_steps?` to check that there are no step navs.

If we do decide to change we'd have to do something like this:

```
step_by_step_navs = if active_step_by_step?
                              [active_step_by_step]
                            elsif step_navs.any?
                              step_navs
                            elsif show_related_links_for_secondary_step_by_steps?
                              secondary_step_by_steps
                            else
                              []
                            end
```

which is messy no?

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: this has been removed

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: as above - in the show_secondary_step_by_step we do a check to see if there are any step navs

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
leenagupte: I think it makes more sense logic-wise, and makes it more readable to put secondary step navs after, primary ones. 
It also means that you could remove `show_related_links_for_secondary_step_by_steps?` and just use `secondary_step_by_steps.any?`.

leenagupte: `show_related_links_for_secondary_step_by_steps?` doesn't check for primary step by steps, it looks like it only checks if there are less than 5 secondary ones.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -21,16 +21,23 @@ def related_to_step_navs
...
       end
 
       def show_header?
-        step_navs.count == 1 || active_step_by_step?
+        step_navs.count == 1 || active_step_by_step? || show_secondary_step_by_step?
       end
 
       def show_related_links?
         return true if active_step_by_step?
+        return true if show_related_links_for_secondary_step_by_steps?
```
leenagupte: I wonder if this whole check can be reduced to:
```
related_links.any? && related_links.count < 5
```

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,29 @@ def active_step_by_step
         @active_step_navs.first
       end
 
+      def secondary_step_by_step?
+        secondary_step_by_steps.any? && step_navs_combined_list.count === 0
```
leenagupte: Ah, I think `step_navs_combined_list.count === 0` is being used to check whether there are any primary step by steps? 

Perhaps this check could be moved to a method, so it's easier to see what it does, e.g. 
```
def primary_step_navs?
  step_navs_combined_list.count > 0
end
```

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,29 @@ def active_step_by_step
...
+      end
+
+      def secondary_step_by_step
+        secondary_step_by_steps.first
```
leenagupte: I'm not sure if this is needed as it's only being used in one place, to get the `current_step_nav`

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,29 @@ def active_step_by_step
...
+      end
+
+      def show_secondary_step_by_step?
+        secondary_step_by_step? && secondary_step_by_steps.count === 1
```
leenagupte: Does this check work? `secondary_step_by_step?` only returns true if `step_navs_combined_list.count === 0`, so `secondary_step_by_step?` and `secondary_step_by_steps.count === 1` will never be true at the same time.

leenagupte: I think for readability, it might be better to expand out the checks. e.g.
```
!primary_step_by_step? && secondary_step_by_steps.any? && secondary_step_by_steps.count === 1
```

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,12 +95,29 @@ def active_step_by_step
...
+      end
+
+      def show_related_links_for_secondary_step_by_steps?
+        secondary_step_by_step? && secondary_step_by_steps.count < 5
```
leenagupte: As with `show_secondary_step_by_step?` I think this might be more readable if it's expanded out, e.g.

```
!primary_step_by_step? && secondary_step_by_steps.any? && secondary_step_by_steps.count < 5
```

spec/features/contextual_navigation_spec.rb
```diff
@@ -64,6 +64,40 @@
...
+    and_i_visit_that_page
+    then_i_just_see_the_step_by_step_related_links_with_just_two_links
+  end
+
+  scenario "There's 1 primary step by step and 1 secondary step by step" do
+    given_there_are_one_primary_step_by_steps_and_one_secondary_to_step_navs
+    and_i_visit_that_page
+    then_i_see_the_step_by_step
+    and_the_step_by_step_header
+    then_i_see_the_primary_step_by_step
```
leenagupte: Could this have another step confirming that we don't see the secondary step by step?

spec/features/contextual_navigation_spec.rb
```diff
@@ -64,6 +64,40 @@
...
+    given_there_are_six_secondary_step_by_step_lists
+    and_i_visit_that_page
+    then_theres_no_step_by_step_at_all
+    and_no_step_by_step_header
+  end
+
+  scenario "There's 3 secondary step by steps and 2 primary step by step lists" do
+    given_there_are_three_secondary_step_by_steps_and_two_primary_to_step_navs
+    and_i_visit_that_page
+    then_i_just_see_the_step_by_step_related_links_with_just_two_links
```
leenagupte: Could this have another step confirming that we don't see the secondary step by step? Or be changed to confirm that it's definitely only the primary step by steps that in the related links?

spec/features/contextual_navigation_spec.rb
```diff
@@ -98,6 +98,27 @@
...
+    and_no_step_by_step_header
+  end
+
+  scenario "There are three related to step by step lists, no primary step by step list and one secondary to step nav" do
+    given_there_are_three_related_to_step_nav_and_one_secondary_to_step_nav
+    and_i_visit_that_page
+    then_theres_no_step_by_step_at_all
+    and_no_step_by_step_header
+  end
+
```
leenagupte: These tests are great! 👍

spec/features/contextual_navigation_spec.rb
DilwoarH: Done.

spec/features/contextual_navigation_spec.rb
DilwoarH: I think in this case, we can infer that we will not see secondary if we see primary.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
DilwoarH: I've left it as this keeps the logic of choosing the secondary step by step away from the other functions

leenagupte: I don't agree as the only place it's used, we're using `step_navs.first` in the condition above, but it's not a blocker.

lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
```diff
@@ -81,14 +106,34 @@ def active_step_by_step
         @active_step_navs.first
       end
 
+      def secondary_step_by_step?
```
leenagupte: I think this name should be plural, i.e. `secondary_step_by_steps?`

spec/features/contextual_navigation_spec.rb
```diff
@@ -242,14 +247,23 @@ def and_i_visit_that_page
   end
 
   def and_i_visit_that_page_by_clicking_on_a_step_by_step_link
-    visit "/contextual-navigation/page-with-contextual-navigation?step-by-step-nav=8ad782bd-8603-40eb-97c0-434cb22047cd"
+    visit "/contextual-navigation/page-with-contextual-navigation?step-by-step-nav=8ad999bd-8603-40eb-97c0-999cb22047cd"
```
leenagupte: Why did the content_id change?

DilwoarH: The content item is always the same when generating random step by steps. I realised the previous test that we were doing was giving false positives. To make sure that we 100% see the active step by step I've set a unique ID. ref https://github.com/alphagov/govuk_publishing_components/pull/900/files#diff-6a1129fe0126f5e5082740817b714c2bR140

leenagupte: Could you add that to the commit note?

DilwoarH:
@leenagupte - tests have been added

spec/features/contextual_navigation_spec.rb
DilwoarH: Done 👍

spec/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation_spec.rb
```diff
@@ -474,6 +474,520 @@
     end
   end
 
+  context("secondary step by steps") do
```
leenagupte: These are great! Perhaps one more that tests what happens when there is an active step by step too?

DilwoarH: done :)

spec/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation_spec.rb
```diff
@@ -474,6 +474,584 @@
...
+          expect(step_nav_links.show_related_links_for_secondary_step_by_steps?).to be false
+          expect(step_nav_links.show_header?).to be false
+          expect(step_nav_links.show_related_links?).to be false
+          expect(step_nav_links.show_sidebar?).to be false
+        end
+      end
+    end
+
+    context "secondary step by step with active step by step" do
+      it "shows header for related to step nav when a step by step is active and there is also a secondary step by step" do
```
leenagupte: This can't be right, can it? I thought "related_to" step navs don't show on the content item?

PR #901, opened 2019-05-30
Remove experimental status of Attachment components

kevindew:
Trello: https://trello.com/c/BQ5bBkMw/881-wrap-up-technical-loose-ends-on-attachments

As these are now used in production by Content Publisher and Govspeak
these are no longer considered experimental. Hopefully there won't be
need for any drastic API changes to them in the near future.

As these are purely internal changes I don't believe they need a changelog entry.


kevindew:
Thanks @alex-ju 👍

PR #902, opened 2019-05-30
Use "Component Guide" as title and breadcrumbs root instead of `component_guide_title`

alex-ju:
We're currently using the `component_guide_title` defined in config of the consumer application for the header, heading one and breadcrumb root. While the link in the header points to the root of the application, the breadcrumb root is pointing to the component guide route. This PR updates the template to keep the custom defined title only for the header and use a "Component Guide" for the rest.

Example using Content Publisher app.

### Before
Screen Shot 2019-05-30 at 17 32 43

Screen Shot 2019-05-30 at 17 33 50

### After
Screen Shot 2019-05-30 at 17 32 26

Screen Shot 2019-05-30 at 17 33 37

### DIFF
--- a/app/views/govuk_publishing_components/component_guide/index.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/index.html.erb
@@ -1,4 +1,4 @@
-<%= render 'govuk_publishing_components/components/title', title: GovukPublishingComponents::Config.component_guide_title %>
+<%= render 'govuk_publishing_components/components/title', title: "Component Guide" %>
 
 <div class="component-markdown">
   <p>Components are packages of template, style, behaviour and documentation that live in your application.</p>
--- a/app/controllers/govuk_publishing_components/component_guide_controller.rb
+++ b/app/controllers/govuk_publishing_components/component_guide_controller.rb
@@ -48,7 +48,7 @@ module GovukPublishingComponents
 
     def index_breadcrumb
       {
-        title: GovukPublishingComponents::Config.component_guide_title,
+        title: "Component Guide",
         url: component_guide_path
       }
     end


app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -1,4 +1,4 @@
-<%= render 'govuk_publishing_components/components/title', title: GovukPublishingComponents::Config.component_guide_title %>
+<%= render 'govuk_publishing_components/components/title', title: "Component Guides" %>
```
kevindew: Did you mean this to be Component Guide - without the s?

DilwoarH: How comes, the title doesn't have the correct value in it?

alex-ju: Right, fixed it now.

@DilwoarH: well, technically the title (as in the `<title>` tag) has the correct value - which is the one set in config of the app, but the heading (`<h1>`) of the page should always be "Component Guide".

DilwoarH: Thanks for explaining Alex, I also noticed that in the first example, the breadcrumbs is `content publisher` then on the second it's `Components guide` but I only see a change to the heading title component. How has this change affected the breadcrumbs?

alex-ju: @DilwoarH you're right, I didn't commit the breadcrumbs replacement. Added now, good spot.

alex-ju:
On a second thought, hardcoding it may not be the best idea.

PR #905, opened 2019-06-03
Switch tests from Poltergeist (PhantomJS) to govuk_test (Chrome)

brucebolt:
- Removes Poltergeist, which is no longer maintained.  Replaced with Selenium driver through the govuk_test gem.
- By using govuk_test, this will override the driver used by jasmine (Selenium instead of PhantomJS).
- Fixes tests which relied on features from Poltergeist.


spec/component_guide/component_example_accessibility_testing_spec.rb
```diff
@@ -11,19 +11,14 @@
...
-
+  it 'shows accessibility violations on the page and through browser console' do
+    visit '/component-guide/test-component-with-a11y-issue'
     expect(page).to have_selector('.js-test-a11y-failed.js-test-a11y-finished')
 
+    expect(page.driver.browser.manage.logs.get(:browser).map { |e| e.message if e.message.match(/Accessibility issues/) }).not_to be_empty
+
     selector_with_error = page.first('.selector').text
-    expect(page).to have_selector(selector_with_error)
+    expect(page).to have_selector(selector_with_error, visible: false)
```
ChrisBAshton: Why is this not visible?

spec/component_guide/component_example_accessibility_testing_spec.rb
```diff
@@ -11,19 +11,14 @@
     expect(page).to have_selector('.js-test-a11y-success.js-test-a11y-finished')
   end
 
-  it 'throws JavaScript errors if a component has an accessibility issue' do
```
nickcolley: Removing these tests suggests that the govuk_test module will not fail when serious accessibility issues are found. The current implementation relies on poltergeist blowing up when it encounters a JavaScript error being raise in the page.

brucebolt: With Selenium, we can't detect a JavaScript error, in the same way as Poltergeist was doing.  What I've done is replaced this with https://github.com/alphagov/govuk_publishing_components/pull/905/files#diff-54d9df40ed0e8a23b378c9114c300610R18, which checks that the error is logged to the browser's console.

spec/component_guide/component_example_accessibility_testing_spec.rb
brucebolt: The element we're selecting is ``.  It seems Poltergeist/PhantomJS saw this as visible, whereas Selenium/Chrome doesn't.

ChrisBAshton: May be worth adding a comment, e.g.

```ruby
# The deliberately inaccessible HTML `<img src="">` is invisible, so needs `visible: false`
expect(page).to have_selector(selector_with_error, visible: false)
```

spec/component_guide/component_example_accessibility_testing_spec.rb
nickcolley: Just to be clear, the current functionality means that when a serious accessibility issue is found with axe, it'll fail the build.

Unless I'm mistaken the new behaviour will not fail builds for components built with the guide?

nickcolley: The expectation you have added would need to run automatically on every example, for the behaviour to be consistent with what it is now.

brucebolt: The current code will only test for failures with `/component-guide/test-component-with-a11y-issue`, which is what I've replicated in the new test.  Unless I'm missing something?

nickcolley: The current behaviour:

1. The tests will visit a component
2. If the component has a serious accessibility error, it'll throw a JavaScript error
3. Poltergeist will fail the build whenever it encounters a JavaScript error

The new behaviour:

1. The tests will visit a component
2. If the component has a serious accessibility error, it'll throw a JavaScript error
3. govuk_test will not will fail the build whenever it encounters a JavaScript error

So we'd either need to be OK with this, or add additional logic to this gem that ensures everytime the tests visit a component they assert that there's no JavaScript errors.

brucebolt: The test that I've updated here only checks an example component to ensure the JavaScript error is raised.  I cannot see anywhere this test would get called when another component is created.  I haven't touched the code that raises the JavaScript error.

The integration test that is documented to be run when creating a new component is https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/minitest/component_guide_test.rb.  This hasn't been changed and the JavaScript error will still be raised.

nickcolley: The JavaScript error is raised sure, but if govuk_test does not automatically fail the build when it encounters JavaScript errors (is my assumption) then this pull request has broken the behaviour that we were testing here. I'm not sure merging it was a good idea, puts us at risk of deploying code with accessibility issues.

brucebolt: Please correct me if I'm wrong, but it looks like failing on a JavaScript error is somewhat redundant.  The integration test for component accessibility checks the class `.js-test-a11y-success` has been added to all components (https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/minitest/component_guide_test.rb#L17).  If that isn't true (i.e. class `.js-test-a11y-failed` has been added), the build will fail.  This seems to be the desired behaviour (i.e. all components must have passed the axe tests and been marked as such).

nickcolley: @brucebolt that does look like it might catch the same error, to test it out I would add some failing markup to a real component and see if it fails the build. fingers crossed...

PR #906, opened 2019-06-04
Update ODT guidance link

theseanything:
The ODT guidance has been published under a new path. This updates link with attachments to the guidance.


alex-ju:
@theseanything looks good. It seems like it needs a test to be updated to reflect the change
```
rspec ./spec/components/attachment_spec.rb:66 # Attachment shows OpenDocument help text if OpenDocument format
```

PR #907, opened 2019-06-05
Support error items on date component and refactor error items logic

kevindew:
Trello: https://trello.com/c/rGjDviSL/848-adapt-scheduling-implementation-for-iterated-workflow

This enables the date input field to accept error_items as per the form input components.

It also moves the logic to flatten error_items into a singular string out of individual components and into the error_message components. Refactoring this has allowed fixing a potential XSS issue in the previous common implementation

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,8 +8,8 @@
   of the commit log.
 
 ## 16.28.0
-* Make inline buttons 50% width from mobile - tablet (PR #898)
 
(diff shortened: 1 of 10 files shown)


kevindew:
> This is great - I wish I'd thought of this :-). I wonder if we could lose the annoying `has_error` boilerplate I added by doing `error_items ||= [error_message].compact`? Maybe I should be the one doing that ;-).

That's a good idea. It's a little more complicated though as you'd need `{ item: error_message }` which would lose the compact convenience.

PR #908, opened 2019-06-05
Replace Accessible Media Player with pure Youtube embed iframe

kr8n3r:
## What
Replaces the [Accessible Media Player](https://github.com/alphagov/Accessible-Media-Player) with a pure Youtube iframe player. This won't work on GOV.UK [until the content security policy has been updated](https://github.com/alphagov/govuk_app_config/pull/99) to allow `youtube-nocookie.com`

## Why
There are a number of accessibility issues with the current video player, so other shortlisted players were [prototyped and reviewed](https://docs.google.com/document/d/1ZhrZH-nKCXs-uMz8HIXPx7FgAs4p3fsrmNMItDj_dt0/edit). The recommendation was to use the Youtube player with some tweaks to make it more accessible:
- Add a `title` attribute to the iframe, taken from the fallback link text,
- Disable option to allow single key shortcuts ([`disablekb`](https://developers.google.com/youtube/player_parameters)) due to [WCAG SC 2.1.4](https://www.w3.org/WAI/WCAG21/quickref/#character-key-shortcuts)
- use the no-cookie version

## Before (Nomensa)
Screen Shot 2019-07-08 at 15 54 06
Screen Shot 2019-07-08 at 15 54 11

## After (Youtube)
Screen Shot 2019-07-08 at 15 53 40
Screen Shot 2019-07-08 at 15 53 56

## Examples

- [Embed video preview url](https://govuk-publishing-compon-pr-908.herokuapp.com/component-guide/govspeak/with_youtube_embed)
- [Embed-disabled preview url](https://govuk-publishing-compon-pr-908.herokuapp.com/component-guide/govspeak/with_youtube_embed_disabled)

## Notes
This PR also:
- rewrites the video player code so there is no dependency on jQuery or jQuery UI

## Accessibility Testing of the Youtube player

<details>
<summary>Summary</summary>

Work by @selfthinker 

I have implemented all 4 contending video players (Plyr, Able Player, MediaElement.js and Video.js) plus the default YouTube player and looked a bit deeper into things, especially accessibility.

There are a couple of things which all the other players inherit from YouTube. I've only tested YouTube videos with them as that is where the majority of GOV.UK videos are hosted (or maybe even 100% of them?).

**Keyboard access**

The YouTube video starts with an image with the generic name "photo image of" that is linked to the channel and then a hidden heading with the name of the channel that is also linked to the channel. That means if you navigate the page without a screen reader, the second tab within the iframe is always hidden. If you navigate the page with a screen reader, the first link never makes sense and the second link rarely makes sense, depending on how well the channel is named, although it always misses the information that it is the name of the channel.

It's nice that YouTube and Able Player show labels for most controls when they have focus or when you hover over them. That way it will be clearer for people with cognitive impairments. Plyr has a setting that would show them as well.

The focus ring on some of MediaElement.js's controls is quite faint.

All video players keep keyboard focus within the fullscreen view and loop to the top when you've reached the end of that screen.

Most of the players have keyboard shortcuts. YouTube uses a single key, which would be against WCAG SC 2.1.4, but that can be disabled. Plyr only uses such shortcuts when the focus is on the player, but that can also be disabled. And Able Player uses modifier keys which can also be changed, so it is never a letter just on its own.

**Screen readers**

I have so far tested with NVDA in Firefox and VoiceOver on iOS.
It is interesting how the various players announce that there is a video. The only way how you might know there is one with Plyr (and also the current Nomensa player) is because you will stumble upon a "play button". Not great. 

YouTube is a bit better as at least you will know there is a frame (whatever might be in it). Mediaplayer.js has the words "video player" just before it and has given the player a role of "application" with the same title, and Able Player has a heading saying "media player". Much better.

All players have some kind of feedback after clicking the play button to announce that the button has changed to a pause button. 

Except Plyr gives no feedback at all and even says "button play" after the button has visibly changed to a pause button. YouTube adds the letter "k" which will probably be confusing to screen reader users. On the tooltip the "k" is in brackets which makes it a bit clearer that it's the keyboard shortcut. Able Player goes a bit further and doesn't just announce what the button was changed to but also what clicking it has done to the video, i.e. "pause playing", which I find a bit confusing.

Some labels in the YouTube player are announced after the button, which is confusing. The amount of subscribers is a hidden heading, also confusing. The Plyr player often doesn't provide any feedback after activating a control, which is especially confusing when you are supposed to do something afterwards as in a dropdown menu.

MediaElement.js has half a keyboard trap. When a video has started the up and down arrows turn the volume up and down but cannot be used for the standard NVDA navigation. Tabbing still works, though.

All players show the "related videos" etc including the close button for those when pausing, but Plyr and MediaElement.js doesn't allow you to interact with them in any way, neither via clicking nor tabbing or via navigating with a screen reader. That means there is some functionality visible which cannot be accessed at all. Not good.

All players use the internal iOS video player (although that is often configurable) when used in Safari on iOS. Only Plyr doesn't do that. Plyr's fullscreen implementation still allows you to navigate the rest of the page with VoiceOver, which can be disorientating.
Voice control

I only tested with Windows Speech Recognition as I was testing when I was working from home and didn't have access to Dragon. I will re-test this with Dragon when I'm in the office.
I'm not too familiar with voice control tools, so might easily have missed something.

I could get every player to work with either "play" or "click play" or "play video" except Able Player. It wasn't working most of the time, it seems to only work with "click" when already hovering over the controls. ("Mousegrid" always works, though.)

I had trouble getting Plyr to pause. The YouTube player seemed to work best in this case.

**Changing colours**

None of the players are usable at all when changing colours in Firefox. Most elements are completely invisible.

Colours are not great but usable when changing colours via the Windows "high contrast" settings. Only MediaElement.js in IE11 is mostly invisible.

**Captions**

The only players which support switching captions on and off when they are hosted on YouTube are the YouTube player and Able Player.

**Browser tests**

Everything works in our supported browsers. In order for Plyr to work in IE11, we would need to implement a polyfill which I haven't bothered with for this test. So, Plyr on my playground won't work in IE11 but adding support is possible.

MediaElement.js on my playground isn't responsive. But there is a responsive setting which doesn't seem to be the default.

**Conclusion**

Just judging by these findings, Able Player or the YouTube player seem to be the best choice.

The question is, how much better is Able Player and in what areas that might justify choosing it over the native YouTube player? It seems to be better for screen reader users but worse for voice control users. It has support for transcripts (even interactive ones), audio descriptions (text-based or audio/video-based) and sign language (via synchronising a second video). But which of those would we actually use?
As far as I can see, both players would fail WCAG but the only proper barrier is the voice control issue with Able Player. (And invisible controls when changing colours, but all players have that issue.)
</details>

### Discussion

Additionally, we'd like to add a visually hidden heading before the iframe so it's clearer to screen reader users that there is a video, e.g. "Video player".

Issue is, the heading level needs to fit into the heading hierarchy (e.g. an h2 if the heading before is an h1, an h3 if the heading before is an h2, etc, but _I think_ we cannot programmatically establish the heading before.

[Youtube Iframe API ](https://developers.google.com/youtube/iframe_api_reference) implementation

## Trello Cards

- [Original Trello ticket](https://trello.com/c/EMOld1t2/152-scope-technical-implementation-limitations-for-new-video-player-for-govuk)
- [Change Nomensa to Youtube video player](https://trello.com/c/EMOld1t2/3-5-change-nomensa-to-youtube-video-player)


app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,48 +1,78 @@
...
   YoutubeLinkEnhancement.prototype.embedVideo = function ($link, videoId) {
-    var $mediaPlayer = $('<span class="media-player" />')
-
-    $link.parent().replaceWith($mediaPlayer)
-
-    var protocol = this.protocol()
+    var parent = $link.parentNode
+    var youtubeVideoContainer = '<div class="youtube-video-container">' +
+    // '<span id="' + videoId + '"></span></div>'
+    '<iframe id="' + videoId + '" type="text/html" src="https://www.youtube-nocookie.com/embed/' +
```
nickcolley: ❤️  for no cookie 👍 💯

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,48 +1,78 @@
...
-      captions: this.captions($link),
-      url: protocol + '//www.youtube.com/apiplayer?enablejsapi=1&version=3&playerapiid='
+  YoutubeLinkEnhancement.prototype.createVideoPlayer = function (playlistItem) {
+    return new window.YT.Player(playlistItem.id, {
+      videoId: playlistItem.videoId,
+      host: 'https://www.youtube-nocookie.com',
+      playerVars: {
+        enablejsapi: 1,
+        rel: 0, // don't show related videos
+        disablekb: 1, // insure the player does not respond to keyboard controls (WCAG)
```
nickcolley: Could you link to the WCAG guideline in this comment please?

kr8n3r: done

kevindew:
@kr8n3r woop this is an exciting change - be glad to say goodbye to the extra dependencies.

As part of your checks for whether this is good to merge could you co-ordinate with Publishing Workflow team (myself or @alex-ju are good candidates) so we can check it works in our app still. We use it to preview documents and need to re-initialise the component each preview. I ask this because it was quite delicate getting the original version to work with this. Thanks.

kr8n3r:
@kevindew yeah will check with the team once I have it in a state that's passing test and is ready for review.

kr8n3r:
Outstanding work:
 - test the functionality

Previous test only tested for the existence of the container (which is govspeak) and it didn't test if the video (iframe) actually gets created.

Currently the test fails is it completes before the API finishes (player is ready) as its async.
Any ideas how to test this would be appreciated.  Jasmine 2.4 is used here.

kr8n3r:
This will need to be updated with the new cookie work happening
- https://github.com/alphagov/govuk_publishing_components/pull/916
- https://github.com/alphagov/govuk_publishing_components/pull/912

While we are using the `no-cookie' youtube domain for embedding videos that don't store cookies when the videos are embedded, it DOES store them once you hit play.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,50 +1,101 @@
-// = require govuk_publishing_components/vendor/jquery-ui-1.10.2.custom
-// = require govuk_publishing_components/vendor/jquery.player.min
+window.GOVUK = window.GOVUK || {}
+// linting is not happy unless YT is defined
+var YT
```
kevindew: This looks a bit fishy here in the global scope. I think there are comments we can use to tell the linter to not complain about a variable.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,50 +1,101 @@
...
+    })
+  }
 
-    $mediaPlayer.player({
-      id: YoutubeLinkEnhancement.nextId(),
-      media: videoId,
-      captions: this.captions($link),
-      url: protocol + '//www.youtube.com/apiplayer?enablejsapi=1&version=3&playerapiid='
+  YoutubeLinkEnhancement.prototype.createVideoPlayer = function (playlistItem) {
+    return new YT.Player(playlistItem.id, {
```
kevindew: It might be clearer to have this reference window.YT and then remove the line from https://github.com/alphagov/govuk_publishing_components/pull/908/files#diff-b4722dbdd6fe718214588e4addc706d4R46 (which I found rather confusing)

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,50 +1,101 @@
...
-        _this.embedVideo($link, videoId)
+        _this.setupVideo($link, videoId)
       }
-    })
+    }
   }
 
-  YoutubeLinkEnhancement.prototype.hasDisabledEmbed = function ($link) {
-    return $link.attr('data-youtube-player') === 'off'
+  YoutubeLinkEnhancement.prototype.playerApiReady = function () {
```
kevindew: I don't think this function (and other global state ones) should be on the prototype of YoutubeLinkEnhancement as we'd only want these run once I'd expect rather than for every instance of an embed.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,50 +1,101 @@
...
+        this.players[i] = currentPlayer
+      }
+    }.bind(this)
   }
 
-  YoutubeLinkEnhancement.prototype.embedVideo = function ($link, videoId) {
-    var $mediaPlayer = $('<span class="media-player" />')
+  YoutubeLinkEnhancement.prototype.insertApiScript = function () {
+    var tag = document.createElement('script')
+    tag.src = 'https:///www.youtube.com/iframe_api'
```
kevindew: do you mean to have the third slash?

vanitabarrett:
@kevindew FYI this should be ready to test now 👍

vanitabarrett:
@kevindew I think I've fixed the problem with rendering multiple identical videos and the invalid HTML. Would you be able to check again the issues you were having in Content Publisher and see if they're resolved now?

kevindew:
> @kevindew I think I've fixed the problem with rendering multiple identical videos and the invalid HTML. Would you be able to check again the issues you were having in Content Publisher and see if they're resolved now?

Thank you! I'll try check it out by the end of the day.

kevindew:
> > @kevindew I think I've fixed the problem with rendering multiple identical videos and the invalid HTML. Would you be able to check again the issues you were having in Content Publisher and see if they're resolved now?
> 
> Thank you! I'll try check it out by the end of the day.

Hey @vanitabarrett, sorry for the delay. I've checked it out and the problem where the same video doesn't load is resolved however the problem where videos can't be re-initialised persists.

Here's a capture of what happens:

video-problem

vanitabarrett:
@kevindew The changes in your branch look good, I just had one comment which I've left there. Happy for you to push them to this branch 🙂

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -1,50 +1,106 @@
...
-    var protocol = this.protocol()
+    var youtubeVideoContainer = document.createElement('div')
+    youtubeVideoContainer.className += 'youtube-video-container'
+    youtubeVideoContainer.innerHTML = '<span id="' + videoWrapperId + '"></span>'
+
+    parentContainer.replaceChild(youtubeVideoContainer, parentPara)
+    this.playList.push({
+      id: videoWrapperId,
+      videoId: videoId,
+      title: $link.textContent
```
selfthinker: Could we add the word "video" in there somewhere? Maybe "Video of" in front of the link text? (Not sure if we need to ask a content designer about this.) One of the problems with the YouTube video player is that it doesn't announce anywhere that this is a video, you just find out by coming across a play button.
My initial proposal suggested to add a hidden heading before the iframe. Not sure why Jani didn't implement that. It would be okay to not have that if at least the title would say it's a video.

kevindew:
> @kevindew The changes in your branch look good, I just had one comment which I've left there. Happy for you to push them to this branch 🙂

Great thanks. I've just done so. 

With regards to your comment:

> I think I'm getting a bit confused here... it looks like we're renaming $element to $jqElement (which I'm assumig stands for 'jQuery Element'), but we've changed the logic to pass HTML rather than a jQuery object?

Yup you're correct that it refers to a jQuery element - would there be a nicer way to make that clearer? I know we use a $ prefix to refer to a jQuery result and for a HTML element so I wanted to make that a bit clearer.

I changed YouTubeLinkEnhancement to accept a HTML element rather than a jQuery one since it no longer uses any jQuery and it means we can lose 
https://github.com/alphagov/govuk_publishing_components/pull/908/commits/a6f9a635f03b01e9388dc3e7e79ab5be6a9803e4#diff-b4722dbdd6fe718214588e4addc706d4L14
where it looks like we have an array of elements and meant it had to be tested with jQuery: https://github.com/alphagov/govuk_publishing_components/pull/908/commits/a6f9a635f03b01e9388dc3e7e79ab5be6a9803e4#diff-ee8d4d8b36beb87a40beebb8b4e9c665L31

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
vanitabarrett: @selfthinker We could add 'Video of:' to the link text. Would this be more confusing if, for example, the video title already included 'video'? For example, we have a video on GOV.UK with the title "Video guide for exporters seeking trade finance and insurance solutions" (https://www.gov.uk/government/collections/watch-our-video-guide)

vanitabarrett: Have implemented the 'Video of: ' approach for now

kevindew: Is this the only text we insert as part of this process? I wonder if we want to consider that the content might not be English?

Although thinking about it that shipped might have already sailed if the whole YouTube interface is in English.

kr8n3r:
thanks everyone for taking this on and making it solid 👍

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
kr8n3r: @selfthinker not implemented as most videos have "Video of" in the title for the link-only fallback. If anything, those without should get updated.

@kevindew the API does provide a [hl parameter](https://developers.google.com/youtube/player_parameters#hl) with changes the player interface. 
I guess we would then need to either read the html lang or particular section lang attribute where the player is placed

selfthinker: It would be good to know how many links/titles already have the word video in it and how likely that is. Regarding other languages, I guess adding " (video)" at the end is slightly safer as the word "video" is pretty much the same across most languages (rarely spelt differently). But the word "of" obviously isn't.

PR #909, opened 2019-06-05
Upgrade to Capybara 3

injms:
Capybara 3 brings in some breaking changes - but the only one that seems to have affected Publishing Components is that `has_content` doesn't play nicely with `\n` and newlines. Some tests had to be tweaked to account for this change.

### DIFF
--- a/spec/component_guide/component_example_spec.rb
+++ b/spec/component_guide/component_example_spec.rb
@@ -38,9 +38,12 @@ describe 'Component example' do
   it 'yields a block in component examples' do
     visit '/component-guide/test-component-with-block'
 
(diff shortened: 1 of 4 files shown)


spec/component_guide/component_example_spec.rb
```diff
@@ -38,9 +38,14 @@
   it 'yields a block in component examples' do
     visit '/component-guide/test-component-with-block'
 
-    expect(page).to have_selector('.component-guide-preview .test-component-with-block', text: 'Here\'s a block')
+    puts page.html
```
sihugh: This might have been inadvertently left

injms: Indeed.

PR #911, opened 2019-06-06
Add aria-live flag to notice component

vanitabarrett:
I came across this when trying to use the notice component as a notice telling the user their settings were saved. We scroll to the top of the page so users who can visually see confirmation of their settings. However, in order to get the notice content to be read out correctly for screen reader users, the aria-live needs to be on the section tag of the notice component itself (not a wrapper), hence adding this flag to the component

This will be used on the new cookie settings page, when the user presses "save changes": https://new-cookie-banner.herokuapp.com/help/cookies

Screen Shot 2019-06-06 at 13 45 08

### DIFF
--- a/app/views/govuk_publishing_components/components/_notice.html.erb
+++ b/app/views/govuk_publishing_components/components/_notice.html.erb
@@ -3,6 +3,7 @@
   description_text ||= false
   description_govspeak ||= false
   description ||= yield || false
+  aria_live ||= false
   local_assigns[:margin_bottom] ||= 8
   local_assigns[:margin_bottom] = 8 if local_assigns[:margin_bottom] > 9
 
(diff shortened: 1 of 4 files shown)


vanitabarrett:
[Productionize the new cookie settings page](https://trello.com/c/fADkc6L5/160-productionize-the-new-cookie-settings-page)

app/views/govuk_publishing_components/components/_notice.html.erb
```diff
@@ -11,10 +12,12 @@
   css_classes = %w(gem-c-notice)
   css_classes << (shared_helper.get_margin_bottom)
 
+  aria_attributes = aria_live ? {label: 'Notice', live: 'polite'} : {label: 'Notice'}
```
DilwoarH: should we allow the option for `assertive` too?

vanitabarrett: I'm not sure if there's a use case for having `aria-live: assertive` on this component? I'm a bit wary of adding `assertive` if it's not really needed because it could have a negative effect on user experience if misused.

alex-ju: I usually favour open APIs, hence allowing any value to `aria_live` instead of using a boolean flag. However, on govuk_publishing_components we have this 'principle' that we don't want to allow something unless it's necessary/there's a need for it - and this aligns with this principle.
We can see this in action on how the label for this component doesn't allow a custom value.

Maybe it's something to be discussed at one of our meetings.

PR #913, opened 2019-06-10
Use a content security policy

kevindew:
This adds the GOV.UK content security policy to the demo version of this app and adds a nonce to all inline javascript. This will then allow other apps to add a [nonce generator](https://edgeguides.rubyonrails.org/security.html#content-security-policy) which disables unsafe inline scripts and the components won't have violations.

This also adds unsafe_eval to the Content Security Policy on component-guide routes as the [axe-core](https://github.com/dequelabs/axe-core/issues/1175) script uses a dependency that makes use of eval.  This will resolve reports like:

Screen Shot 2019-06-10 at 10 11 47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ## Unreleased
 
+* Add unsafe-eval to Content Security Policy for component-guide routes (PR #913)
+* Add nonce attribute to inline scripts if Content Security Policy use a nonce generator PR (PR #913)
 * Add aria-live flag to notice component (PR #911)
 
 ## 16.29.0
--- a/app/controllers/govuk_publishing_components/application_controller.rb
+++ b/app/controllers/govuk_publishing_components/application_controller.rb
@@ -5,6 +5,17 @@ module GovukPublishingComponents
     before_action :set_x_frame_options_header
     before_action :set_disable_slimmer_header
 
+    content_security_policy do |p|
+      # don't do anything if the app doesn't have a content security policy
+      next unless p.directives.any?
+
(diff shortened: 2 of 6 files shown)


kevindew:
I've found I've got a CSP issue to resolve in Firefox before I can merge this 😭

kevindew:
So it turns out that since we still use jQuery 1.12.* we will get CSP violations if `unsafe-inline` is switched off. This is due to [some jQuery code](https://github.com/jquery/jquery/blob/e09907ce152fb6ef7537a3733b1d65ead8ee6303/src/event/support.js#L7-L20) which sets inline onsubmit / onchange / onfocusin attributes to an element. Looks like this code is removed on later jQuery versions.

Annoyingly this only raises an issue in Firefox, with Chrome and Safari not having a problem.

cc @tijmenb this is relevant if you're wondering how we can work towards turning unsafe-inline off for GOV.UK

andysellick:
@kevindew what would the minimum version of jquery need to be to avoid these issues? Could we upgrade? (I realise the enormity of this suggestion)

kevindew:
> @kevindew what would the minimum version of jquery need to be to avoid these issues? Could we upgrade? (I realise the enormity of this suggestion)

jQuery 2 no longer has that so a version of 2.2 would suffice. I believe historically we've not upgraded from jQuery 1.x due to it supporting IE8 however I assume that time has moved on and that's not the same concern it once was.

I think @alex-ju is working towards removing jQuery altogether as a dependency for these components.

kevindew:
Replaced by https://github.com/alphagov/govuk_publishing_components/pull/1145

PR #916, opened 2019-06-10
Cookie consent mechanism

vanitabarrett:
We need to check what cookies a user has consented to before we set each cookie.

This also tidies up the cookie banner code so the old banner and new banner are more separate. This will make it easier for us to remove the old banner code once the new banner has been deployed.

**Note: the new cookie banner tests currently include a workaround by stubbing the return value of JSON.parse. Without this stubbing, it looks as though the value returned includes escaped quotes, which means an additional JSON.parse is required (but this works fine in the browser).**

Edited to add: it looks like the problem mentioned above could be related to phantomJS. I tried to update our version of jasmine and use chromeheadless instead, but this results in some test failures around step-by-step nav which I'm not familiar with. Leaving a [WIP branch here for now](https://github.com/alphagov/govuk_publishing_components/compare/update-jasmine?expand=1), for when I have more time to take a look

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -10,42 +10,65 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.showConfirmationMessage = this.showConfirmationMessage.bind(this)
     this.$module.setCookieConsent = this.setCookieConsent.bind(this)
 
+    this.$module.newCookieBanner = document.querySelector('.gem-c-cookie-banner--new')
+
+    // Temporary check while we have 2 banners co-existing.
+    // Once the new banner has been deployed, we will be able to remove code relating to the old banner
+    // Separating the code out like this does mean some repetition, but will make it easier to remove later
+    if (this.$module.newCookieBanner) {
+      this.setupNewCookieMessage()
+    } else {
+      this.setupCookieMessage()
+    }
+  }
+
+  CookieBanner.prototype.setupCookieMessage = function () {
+    this.$hideLink = this.$module.querySelector('a[data-hide-cookie-banner]')
+    if (this.$hideLink) {
+      this.$hideLink.addEventListener('click', this.$module.hideCookieMessage)
+    }
+
+    this.showCookieMessage()
+  }
+
+  CookieBanner.prototype.setupNewCookieMessage = function () {
+    this.$hideLink = this.$module.querySelector('button[data-hide-cookie-banner]')
+    if (this.$hideLink) {
+      this.$hideLink.addEventListener('click', this.$module.hideCookieMessage)
+    }
+
+    this.$acceptCookiesLink = this.$module.querySelector('button[data-accept-cookies]')
+    if (this.$acceptCookiesLink) {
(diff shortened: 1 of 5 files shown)


vanitabarrett:
[govuk_publishing_components: The consent cookie actually affects which cookies are set](https://trello.com/c/1Dg5Vxh8/162-govukpublishingcomponents-the-consent-cookie-actually-affects-which-cookies-are-set)

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -10,43 +10,66 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.showConfirmationMessage = this.showConfirmationMessage.bind(this)
     this.$module.setCookieConsent = this.setCookieConsent.bind(this)
 
-    var newCookieBanner = document.querySelector('.gem-c-cookie-banner--new')
+    this.$module.newCookieBanner = document.querySelector('.gem-c-cookie-banner--new')
 
-    // Hide the cookie banner on the cookie settings page, to avoid circular journeys
-    if (newCookieBanner && window.location.pathname === '/help/cookies') {
```
DilwoarH: Is this intentional? Do we now want to show it in the cookie screen?

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -82,20 +101,47 @@
...
   }
 
-  window.GOVUK.setCookie = function (name, value, options) {
-    if (typeof options === 'undefined') {
-      options = {}
+  window.GOVUK.checkConsentCookie = function (cookieName, cookieValue) {
+    var currentConsentCookie = window.GOVUK.getCookie('cookie_policy')
+
+    // If we're setting the consent cookie OR deleting a cookie, allow by default
+    if (cookieName === 'cookie_policy' || (cookieValue === null || false)) {
```
ChrisBAshton: The `|| false` is redundant here. I think what you may have meant is this:

```suggestion
    if (cookieName === 'cookie_policy' || (cookieValue === null || cookieValue === false)) {
```

spec/javascripts/components/cookie-banner-new-spec.js
```diff
@@ -0,0 +1,113 @@
...
+        '</div>' +
+      '</div>'
+
+    document.body.appendChild(container)
+    element = document.querySelector('.gem-c-cookie-banner--new')
+
+    // For some reason, JSON.parse on the cookie works in the browser, but fails in Jasmine tests.
+    // It seems to be due to extra escaping of quotes when the code is run in the tests, which means JSON.parse doesn't
+    // work as expected. So we'll stub this value instead.
+    spyOn(JSON, "parse").and.returnValue({"essential":true,"settings":true,"usage":true,"campaigns":true});
```
ChrisBAshton: Is this passing the linter? Appears to be missing some spacing in the JSON.

DilwoarH: I don't think the spec folder gets linted.

DilwoarH: https://github.com/alphagov/govuk_publishing_components/blob/master/package.json#L6

PR #917, opened 2019-06-11
Add options to document list component

andysellick:
Adds the following...

Add 'subtext' option, small grey text to appear beneath metadata. This is needed for search results in finders, (where we want to use the document list component). This is not styled other than to match the font size of the description, specific styles will be added in finder-frontend for any elements that will be passed into the subtext area.

Screen Shot 2019-06-11 at 10 29 39

Add 'highlight' option, including optional highlight text. This is needed to highlight the 'top result' in finders.

Screen Shot 2019-06-11 at 16 20 16

Trello card: https://trello.com/c/DAblviIN/793-investigate-using-more-components-in-finder-frontend

Some examples of what this looks like currently:

- [subtext](https://www.gov.uk/search/all?parent=&keywords=%22every+child+matters%22&level_one_taxon=&manual=&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=relevance) (was broken but fixed in [this PR](https://github.com/alphagov/finder-frontend/pull/1198))
- [highlight option](https://www.gov.uk/find-eu-exit-guidance-business?parent=&keywords=eori&order=relevance) can be seen here, but you need to set a request header of `GOVUK-ABTest-FinderAnswerABTest` to `B` (I'd suggest using [this chrome extension](https://chrome.google.com/webstore/detail/modheader/idgpnmonknjnojddfkpgkljpfnnfcklj)). In the real world it wouldn't have the metadata beneath the description text, I just left this in for the component guide to make sure the layout still worked okay.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -80,3 +80,19 @@
   width: 100%;
   margin-right: 25px;
 }
+
+.gem-c-document-list__item--highlight {
+  border: 1px solid govuk-colour("grey-2");
+  padding: govuk-spacing(6);
+  margin-bottom: govuk-spacing(6);
+
+  .gem-c-document-list__item-title {
+    @include govuk-font(24, bold);
+  }
+}
+
+.gem-c-document-list__highlight-text {
+  @include govuk-font(16, bold);
+  margin: 0 0 govuk-spacing(3) 0;
+
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -39,8 +39,16 @@
 
 .gem-c-document-list__item-description {
   @include govuk-text-colour;
-  @include govuk-font($size: 16, $line-height: 1.5);
   margin: 0 0 5px 0;
+}
+
+.gem-c-document-list__subtext {
+  margin: govuk-spacing(1) 0 0 0;
+}
+
+.gem-c-document-list__item-description,
+.gem-c-document-list__subtext {
+  @include govuk-font($size: 16, $line-height: 1.5);
 }
 
 .gem-c-document-list__item-metadata {
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -14,7 +14,12 @@
     <ol class="gem-c-document-list<%= margin_bottom_class %><%= margin_top_class %>">
   <% end %>
     <% items.each do |item| %>
-      <li class="gem-c-document-list__item<%= within_multitype_list_class %> <%= brand_helper.brand_class %>">
+      <% highlight_class = " gem-c-document-list__item--highlight" if item[:highlight] %>
+
+      <li class="gem-c-document-list__item<%= within_multitype_list_class %> <%= brand_helper.brand_class %> <%= highlight_class %>">
+        <% if item[:highlight] && item[:highlight_text] %>
+          <p class='gem-c-document-list__highlight-text'><%= item[:highlight_text] %></p>
+        <% end %>
         <%=
           link_to(
             item[:link][:text],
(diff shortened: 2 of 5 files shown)


vanitabarrett:
@andysellick Have you got links to finders with these on? Specifically, the subtext one

app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
```diff
@@ -43,6 +43,12 @@
   margin: 0 0 5px 0;
 }
 
+.gem-c-document-list__subtext {
+  @include govuk-font(14);
+  margin: govuk-spacing(1) 0 0 0;
```
vanitabarrett: Could this just be `margin-top: govuk-spacing(1)` instead?

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -14,7 +14,10 @@
     <ol class="gem-c-document-list<%= margin_bottom_class %><%= margin_top_class %>">
   <% end %>
     <% items.each do |item| %>
-      <li class="gem-c-document-list__item<%= within_multitype_list_class %> <%= brand_helper.brand_class %>">
+      <li class="gem-c-document-list__item<%= within_multitype_list_class %> <%= brand_helper.brand_class %> <%= 'gem-c-document-list__item--highlight' if item[:highlight] %>">
```
vanitabarrett: Should we follow the pattern we use for `within_multitype_list_class` and construct the class at the top of the file to avoid too much logic within the HTML itself?

andysellick:
> @andysellick Have you got links to finders with these on? Specifically, the subtext one

I haven't, unfortunately. The only way I know what these things look like is by putting debug code into my local so it outputs. But I'll try to find some examples when I look at addressing the reviews.

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: Yes, have changed now, thanks.

app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
andysellick: No - the paragraph seems to be inheriting bottom margin from somewhere, which I wanted to remove.

andysellick:
> I'm wondering if the subtext addition could be structured as part of `gem-c-document-list__item-metadata` (potentially with a modifier for the grey font colour) instead of a new element (`gem-c-document-list__subtext`)

@alex-ju that's an approach I hadn't considered probably because of the way the data in `finder-frontend` gets assembled before it (would be) passed to the component. I'm in no huge rush to merge this, shall we chat about it offline? Would appreciate your input.

andysellick:
> I'm wondering if the subtext addition could be structured as part of `gem-c-document-list__item-metadata` (potentially with a modifier for the grey font colour) instead of a new element (`gem-c-document-list__subtext`)

@alex-ju just to summarise the conversation we just had - I'm going to leave the subtext option in place, but remove the styling from the component, as the appearance of this is specific to the application I'm going to use it in (finder-frontend) and there's additional styling already in finder-frontend for the debug stuff that will also be inside the subtext element.

andysellick:
@vanitabarrett turns out the subtext 'First published by' is broken 😱 but it's being fixed in [this PR](https://github.com/alphagov/finder-frontend/pull/1198) 🎉

andysellick:
@vanitabarrett the 'First published by' is fixed and deployed, I've added an example link to the PR but it's also live at the link below. I've removed the styling for the subtext option as it feels more appropriate for it to go in finder-frontend (there's going to be more than just this inside this element).

Are you happy with this PR?

https://www.gov.uk/search/all?parent=&keywords=%22every+child+matters%22&level_one_taxon=&manual=&public_timestamp%5Bfrom%5D=&public_timestamp%5Bto%5D=&order=relevance

vanitabarrett:
@andysellick Looks good to me 👍 The only comment I have is maybe the first commit message needs to be changed to reflect the fact that you're not styling it to match the existing look, and that styles will be applied within finder-frontend?

Issue #915, opened 2019-06-10
Handle HTTP 422 response in the feedback component


kr8n3r:
The feedback component in alphagov/feedback app receives a 422 response ([Unprocessable Entity](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/422)) if the user leaves all fields empty. Code [here](https://github.com/alphagov/feedback/blob/20f46dd1f69f389f368484c77e33703a91ce9456/app/controllers/contact/govuk/problem_reports_controller.rb) 

While the error message is displayed, it's  not particularly useful and is at odds with [WCAG SC: 3.3.1](https://www.w3.org/TR/2008/REC-WCAG20-20081211/#minimize-error-identified)

The feedback app response on submissions is 
```
errors: ["What wrong can't be blank", "What doing can't be blank"]
0: "What wrong can't be blank"
1: "What doing can't be blank"
message: "<h2>Sorry, we're unable to send your message as you haven't given us any information.</h2> <p>Please tell us what you were doing or what went wrong.</p>"
status: "error"
```
so ideally the component would cater for that use case in [here](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/feedback.js#L125)

The service service manual frontend app caters for it 
https://github.com/alphagov/service-manual-frontend/blob/master/app/assets/javascripts/modules/improve-this-page.js#L286

PR #918, opened 2019-06-11
Update feedback component error handling

kr8n3r:
At the moment, the `alphagov/feedback` application responds with a 422 HTTP status if all fields are left empty in the form.

**Current behaviour**
Screen Shot 2019-06-11 at 10 34 54

The feedback components javascript was setup to always apply the constructed error message first and then check for the `responseJSON.message` property, which would always return undefined.

This resulted in 422 status error message to never being applied.

This PR updates the logic to account for that.
Resolves: #915 
Trello ticket: https://trello.com/c/r5jiQ7Wk/130-feedback-component-insufficient-error-message-11

**New behaviour** (mock api response)
Screen Shot 2019-06-11 at 11 02 30

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update feedback component error handling (PR #918)
 * Support data attributes for error summary items (PR #924)
 * Disable youtube embeds if campaign cookies turned off (PR #919)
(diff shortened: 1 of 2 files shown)


kr8n3r:
was really hoping for a heroku deployment to workaround CORS

PR #923, opened 2019-06-13
Delete cookies when a user remove consent

vanitabarrett:
When a user removes consent for a particular cookie category we need to delete all cookies of that type. For example: if a user removes consent for usage cookies, we want to delete usage cookies for that user.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
@@ -115,6 +115,19 @@
 
     for (var cookieType in options) {
       cookieConsent[cookieType] = options[cookieType]
+
+      // Delete cookies of that type if consent being set to false
+      if (!options[cookieType]) {
+        for (var cookie in COOKIE_CATEGORIES) {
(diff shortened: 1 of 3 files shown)


vanitabarrett:
[Remove GA cookies if consent for 'usage' cookies is withdrawn](https://trello.com/c/wo6wNeeq/181-remove-ga-cookies-if-consent-for-usage-cookies-is-withdrawn)

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -74,17 +74,6 @@
     window.GOVUK.setCookie('cookie_policy', JSON.stringify(approvedConsent))
   }
 
-  window.GOVUK.denyAllCookieTypes = function () {
```
DilwoarH: Is this function used by anything?

vanitabarrett: @DilwoarH No - users can't opt-out of 'essential' cookies so this method is not being used, and cannot be used as it stands unless our cookie policy changes.

PR #925, opened 2019-06-17
Reset seen_cookie_message for new banner

vanitabarrett:
Some users will already have acknowledged the old cookie banner, meaning they will have the seen_cookie_message set to true. If we deploy the new cookie banner, these users won't see it.

We therefore need to reset that cookie for users who have not yet seen the new cookie banner.

**Note: this can safely be merged and deployed as it won't have any effect until the new banner is switched on via a flag in Static**

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -42,6 +42,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$acceptCookiesLink.addEventListener('click', this.$module.setCookieConsent)
     }
 
+    // Force the new cookie banner to show if we don't think the user has seen it before
+    // This involves resetting the seen_cookie_message cookie, which may be set to true if they've seen the old cookie banner
+    if (!window.GOVUK.cookie('cookie_policy')) {
+      if (window.GOVUK.cookie('seen_cookie_message') === 'true') {
+        window.GOVUK.cookie('seen_cookie_message', false)
+      }
+    }
+
     this.showNewCookieMessage()
   }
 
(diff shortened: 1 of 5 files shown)


vanitabarrett:
[Ensure that the "seen_cookie_message" is reset when we deploy](https://trello.com/c/hGke8eKi/167-ensure-that-the-seencookiemessage-is-reset-when-we-deploy)

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -59,16 +67,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-    }
-
-    var shouldHaveCookieMessage = (this.$module && window.GOVUK.cookie('seen_cookie_message') !== 'true')
+    } else {
+      var shouldHaveCookieMessage = (this.$module && window.GOVUK.cookie('seen_cookie_message') !== 'true')
 
-    if (shouldHaveCookieMessage) {
-      this.$module.style.display = 'block'
+      if (shouldHaveCookieMessage) {
+        this.$module.style.display = 'block'
```
andysellick: This line is only changed because it's indented, but I wondered what the thinking was behind setting the style like this compared to setting a class that controls `display`? I can't actually think of a good argument for doing it with classes (this way saves lines of code, for example) so this may be a non-point, I just wondered.

vanitabarrett: I don't think there was a particular reason (I think this code was existing already and I followed the pattern). If there were more styles being applied I'd use a class, but not sure it needs it in this case... unless there's a standard pattern for using classes? Happy to change if needed.

vanitabarrett:
@andysellick Are you happy to approve this?

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
andysellick: That's fine then. As long as it's been thought through, and that result seems reasonable.

andysellick:
> @andysellick Are you happy to approve this?

I'm happy for it to be approved 😄. I'm not sure I understand it enough to review it thoroughly.

PR #935, opened 2019-06-17
Fix resetting cookie mechanism for new banner

vanitabarrett:
As a result of a rebase to some previous PRs, a bug got through where the new consent cookie was being set before the new banner was deployed. As we rely on this cookie not being set to reset `seen_cookie_message` and show the new banner to everyone, this needed to be fixed before going live.


vanitabarrett:
[Fix resetting cookie mechanism](https://trello.com/c/eVfj26HZ/193-fix-resetting-cookie-mechanism)

PR #937, opened 2019-06-18
Change step by step navigation active step link state

andysellick:
Adds underline to the currently active step link, following user research.

Before:

Screen Shot 2019-06-18 at 11 09 32

After:

Screen Shot 2019-06-18 at 11 09 41

Also changes the active link 'dot' to a line, to look like this:

Screen Shot 2019-06-18 at 14 50 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Change step by step navigation active step link state (PR #937)
 * Fix dialog element role in modal dialogue component (PR #920)
 
 ## 17.1.1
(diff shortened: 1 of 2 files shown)


DilwoarH:
Should it be styled more like this instead:

Screen Shot 2019-06-18 at 11 59 06

andysellick:
Not sure. Waiting on Kate.

DilwoarH:
We should maybe remove the $dot-size variable to avoid confusion for new developers.

andysellick:
> We should maybe remove the $dot-size variable to avoid confusion for new developers.

I already did.

PR #940, opened 2019-06-19
Fix new cookie mechanism not being session based

vanitabarrett:
We had assumed that not passing `days` to `window.GOVUK.cookie()` set a default expiry, but it doesn't set any expiry date (therefore creating a session cookie).

This PR:

- Sets the expiry date of the consent cookie and the `seen_cookie_message` cookie to 1 year
- Adds a default expiry of 30 days to `window.GOVUK.cookie()`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 
 ## Unreleased
 
+* Set the expiry of consent cookie to 1 year (PR #940)
+* Sets the default expiry of cookies to 30 days if not specified (PR #940)
+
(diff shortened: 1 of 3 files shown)


vanitabarrett:
[Some people are seeing the banner again after interacting with it](https://trello.com/c/xekgTFkH/201-some-people-are-seeing-the-banner-again-after-interacting-with-it)

vanitabarrett:
@tijmenb Will add the tests after the fix is complete, as there's a bit of a rush on getting this done 👍

PR #945, opened 2019-06-20
Add tests for new cookie expiry date logic

vanitabarrett:
We added logic in #940 which:

- Set the default expiry of cookies to 30 days, if not explicitly provided. This is because we were caught out by the expiry not being set and ending up with a session cookie instead.
- Set the expiry of the consent cookie to 1 year
- Set the expiry of seen_cookie_message to 1 year, to match the consent cookie

The above change was urgent and we didn't have time to write tests. This PR adds tests for that logic.


vanitabarrett:
[Add tests for cookie expiry date logic](https://trello.com/c/nCe8kU6r/211-add-tests-for-cookie-expiry-date-logic)

PR #946, opened 2019-06-20
Increase inline button desktop spacing

vanitabarrett:
## What
Inline buttons did not have any explicit spacing set between them. The only small spacing that appeared was down to (potentially) HTML whitespace.

This sets a margin-right of 10px on inline buttons, forcing the spacing to increase on desktop.
Mobile spacing should remain unchanged.

## Why
The spacing between the inline buttons used on the new cookie banner was too small.

## Before
Screen Shot 2019-06-20 at 14 29 40

## After
Screen Shot 2019-06-20 at 14 29 26

**Link:** https://govuk-publishing-compon-pr-946.herokuapp.com/component-guide/cookie-banner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Increase spacing between inline buttons on desktop (PR #946)
+
 ## 17.4.0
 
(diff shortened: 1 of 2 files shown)


vanitabarrett:
[Increase spacing between buttons in cookie banner](https://trello.com/c/YFltG8i2/212-increase-spacing-between-buttons-in-cookie-banner)

PR #948, opened 2019-06-21
Hide 'Accept Cookies' button when Javascript not available

vanitabarrett:
## What
Hide the "Accept Cookies" button in the cookie banner when Javascript is not available

This change ties in with https://github.com/alphagov/frontend/pull/1881

## Why
At the moment, when Javascript is not enabled the cookie banner looks no different. However, the 'Accept Cookies' button does not work without Javascript - our cookie consent mechanism is based on Javascript at the moment. We therefore want to hide the "Accept Cookies" button. Users can still click "Cookie Settings" to go to `/help/cookies` and get more information.

## Visual Changes
no-js

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix confirmation message being read to screenreaders (PR #952)
+* Hide 'Accept Cookies' button when Javascript not available (PR #948)
 
 ## 17.6.0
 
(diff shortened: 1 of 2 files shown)


vanitabarrett:
[Handle settings page requiring Javascript](https://trello.com/c/8U7NqGxP/204-handle-settings-page-requiring-javascript)

PR #949, opened 2019-06-24
Improve accessibility of the cookie banner

vanitabarrett:
## What
- Adding `role="region"` and `aria-label="cookie banner"` to the cookie banner.
- Switching focus to the cookie banner confirmation message automatically once a user has clicked "Accept Cookies", to enable the confirmation message to be read automatically to screen-readers.

## Why
**Adding `role=region` and `aria-label=cookie banner`:** When running our [accessibility review](https://trello.com/c/X7EnjPLR/192-accessibility-review), we had a warning about the cookie banner not being included in any landmark region. This meant that a user navigating the page by landmark would miss the cookie banner. 

Note: Other landmark roles didn't seem to fit this use case, so I've gone for a generic 'region' with aria-label, but happy to take other suggestions!

**Confirmation message**: without this change, a screen-reader user clicking on "Accept cookies" would have no automatic feedback of anything happening.

## Before (landmarks on GOV.UK homepage)
Screen Shot 2019-06-24 at 11 27 22

## After (landmarks on GOV.UK homepage)
Screen Shot 2019-06-24 at 11 28 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add landmark to cookie banner (PR #949)
+* Cookie banner confirmation message is read automatically to screenreaders (PR #949)
+
 ## 17.5.0
 
 * Increase spacing between inline buttons on desktop (PR #946)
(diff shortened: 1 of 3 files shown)


vanitabarrett:
[Check Accessibility warnings and fix if necessary](https://trello.com/c/MMiiT5nF/213-check-accessibility-warnings-and-fix-if-necessary)

vanitabarrett:
Thanks @alex-ju , good point. I've moved the multiple querySelector calls for the confirmation message to `start()`.

PR #952, opened 2019-06-25
Fix cookie confirmation for screenreaders

vanitabarrett:
## What
Removes `aria-live="polite"` as we are now explicitly setting focus on the confirmation message after accepting cookies (done in https://github.com/alphagov/govuk_publishing_components/pull/949).

## Why
The combination of `aria-live="polite"` and `aria-label="cookie banner"` doesn't seem to work well together - instead of announcing the confirmation message when accepting cookies, it read out the aria-label instead.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


vanitabarrett:
[Check Accessibility warnings and fix if necessary](https://trello.com/c/MMiiT5nF/213-check-accessibility-warnings-and-fix-if-necessary)

PR #957, opened 2019-06-25
Stop the hardcoding of a google client ID in the feedback footer

JonathanHallam:
## What
Stop hardcoding a GA Client ID

## Why
As it stands the hardcoding of a google client id in the feedback footer is making it impossible to link feedback data to GA.

## Visual Changes
No changes. The only thing that this could have changed would be the no email survey link but I've checked to make sure that it won't.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Stop the hardcoding of a google client ID in the feedback footer (PR #957)
+
 ## 17.10.0
 
 * Enable data attributes for table header links (PR #962)
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -20,6 +20,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$promptQuestions = $element.find('.js-prompt-questions')
       this.$promptSuccessMessage = $element.find('.js-prompt-success')
       this.$somethingIsWrongForm = $element.find('#something-is-wrong')
+      this.$surveyForm = $element.find('#page-is-not-useful')
+      this.$surveyWrapper = $element.find('#survey-wrapper')
 
       var that = this
       var jshiddenClass = 'js-hidden'
@@ -47,6 +49,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         trackEvent(getTrackEventParams(that.$pageIsUsefulButton))
         showFormSuccess()
         revealInitialPrompt()
+      })
+
+      this.$pageIsNotUsefulButton.on('click', function (e) {
+        var gaClientId
+        var dummyGaClientId = '111111111.1111111111'
+        if (window.GOVUK.cookie('_ga') === null || window.GOVUK.cookie('_ga') === '') {
+          gaClientId = dummyGaClientId
+        } else {
+          gaClientId = window.GOVUK.cookie('_ga').split('.').slice(-2).join('.')
+        }
+        setHiddenValuesNotUsefulForm(gaClientId)
       })
 
       $element.find('form').on('submit', function (e) {
@@ -88,6 +101,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       function setHiddenValues () {
         that.$somethingIsWrongForm.append('<input type="hidden" name="javascript_enabled" value="true"/>')
         that.$somethingIsWrongForm.append($('<input type="hidden" name="referrer">').val(document.referrer || 'unknown'))
+      }
+
+      function setHiddenValuesNotUsefulForm (gaClientId) {
+        var currentPathName = window.location.pathname.replace(/[^\s=?&]+(?:@|%40)[^\s=?&]+/, '[email]')
+        var finalPathName = encodeURI(currentPathName)
+
+        that.$surveyForm.append($('<input name="email_survey_signup[ga_client_id]" type="hidden">').val(gaClientId || '0'))
+        that.$surveyWrapper.append('<a href="https://www.smartsurvey.co.uk/s/gov-uk-banner/?c=' + finalPathName + '&amp;gcl=' + gaClientId + '" class="gem-c-feedback__email-link govuk-link" id="take-survey" target="_blank" rel="noopener noreferrer">Don’t have an email address?</a>')
       }
 
       function updateAriaAttributes (linkClicked) {
--- a/spec/javascripts/components/feedback-spec.js
+++ b/spec/javascripts/components/feedback-spec.js
@@ -18,7 +18,7 @@ describe("Feedback component", function () {
         '<a href="#" class="gem-c-feedback__close js-close-form" aria-controls="something-is-wrong" aria-expanded="true" data-track-category="Onsite Feedback" data-track-action="GOV.UK Close Form">Close</a>' +
 
         '<div class="grid-row">' +
-          '<div class="column-two-thirds">' +
+          '<div class="column-two-thirds" id="survey-wrapper">' +
             '<div class="gem-c-feedback__error-summary js-hidden js-errors" tabindex="-1"></div>' +
 
             '<input type="hidden" name="url" value="http://example.com/path/to/page">' +
@@ -63,7 +63,6 @@ describe("Feedback component", function () {
             '<input class="gem-c-input " id="input-111111" name="email_survey_signup[email_address]" type="text">' +
 
             '<input class="gem-c-feedback__submit" type="submit" value="Send me the survey">' +
-            '<a href="FIXME" class="">Don\'t have an email address?</a>' +
           '</div>' +
         '</div>' +
       '</form>' +
@@ -104,6 +103,27 @@ describe("Feedback component", function () {
     loadFeedbackComponent();
 
     expect($('#something-is-wrong').find("[name=referrer]").val()).toBe("unknown");
+  });
+
+  it("should append a hidden 'ga_client_id' field to the form with the appropriate value", function() {
+    loadFeedbackComponent();
+    window.GOVUK.setCookie('_ga', 'GA1.3.512324446.1561716924', {})
+    $('.js-page-is-not-useful').click();
+    expect($('#page-is-not-useful').find("[name='email_survey_signup[ga_client_id]']").val()).toBe("512324446.1561716924");
+  });
+
+  it("should append a hidden 'ga_client_id' field to the from with a default value if no client id is present", function() {
+    loadFeedbackComponent();
+    window.GOVUK.setCookie('_ga', '', {})
+    $('.js-page-is-not-useful').click();
+    expect($('#page-is-not-useful').find("[name='email_survey_signup[ga_client_id]']").val()).toBe("111111111.1111111111");
+  });
+
+  it("should append the \'Don’t have an email address?\' link at the bottom of the form", function(){
+    loadFeedbackComponent();
+    $('.js-page-is-not-useful').click();
+
+    expect($('#survey-wrapper').find('#take-survey').length).toBe(1);
   });
 
   describe("clicking the 'page was useful' link", function () {
@@ -460,6 +480,7 @@ describe("Feedback component", function () {
     });
 
     it("submits the feedback to the feedback frontend", function () {
+      window.GOVUK.setCookie('_ga', '', {})
       loadFeedbackComponent();
       fillAndSubmitPageIsNotUsefulForm();
 
@@ -467,7 +488,8 @@ describe("Feedback component", function () {
       expect(request.url).toBe('/contact/govuk/email-survey-signup');
       expect(request.method).toBe('POST');
       expect(request.data()).toEqual({
-        'email_survey_signup[email_address]': ["test@test.com"]
+        'email_survey_signup[email_address]': ["test@test.com"],
+        'email_survey_signup[ga_client_id]': ['111111111.1111111111']
       });
     });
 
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         revealInitialPrompt()
       })
 
+      this.$pageIsNotUsefulButton.on('click', function (e) {
+        var cookie = document.cookie.split("; ") ;
+        var i ;
+        var clientIDFound = false;
+        for (i = 0; i < cookie.length; i++) {
+          if (cookie[i].startsWith("_ga")) {
```
ChrisBAshton: `.startsWith` is an [ES6 feature](http://kangax.github.io/compat-table/es6/#test-String.prototype.startsWith), which unfortunately is not supported in the [browsers we need to support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices).

You'll need to use `indexOf`, which is widely supported.

e.g. `if (cookie[i].indexOf('_ga') !== -1)`

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            document.gaClientId = cookie[i].split("=")[1].split(".").slice(-2).join(".")
```
ChrisBAshton: You're trying to set a global variable? The traditional way of doing this is to use `window`, i.e. `window.gaClientId = ...`.

But it's generally better to use scoped attributes (`this`). On line  24 you can have `this.gaClientId = undefined;`, which means on this line (59) you can do `this.gaClientId = ...`.

But then you're going to face the [`this` problem](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/this), because your `on('click'` declaration is taking an "anonymous function" as a callback (the `function (e) {}` bit). So the value of `this` is pointing to the anonymous function and not the `Modules.Feedback` object.

So you'll want to use `that.gaClientId` (we set `that = this` on line 25 to get around this issue, it's quite a common workaround). The alternative is [binding your function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_objects/Function/bind).

app/assets/javascripts/govuk_publishing_components/components/feedback.js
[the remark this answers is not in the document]
JonathanHallam: will look into this, thanks!

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            clientIDFound = true
+          }
+        }
+        if (clientIDFound == false) {
+          document.gaClientId = "111111111.1111111111"
```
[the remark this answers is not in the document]
JonathanHallam: It was done when it was first set up but doesn't mean anything. I'm falling back to this because I figured it was clearly a dummy value

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -20,6 +20,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$promptQuestions = $element.find('.js-prompt-questions')
       this.$promptSuccessMessage = $element.find('.js-prompt-success')
       this.$somethingIsWrongForm = $element.find('#something-is-wrong')
+      this.$pageIsNotUsefulForm = $element.find('#page-is-not-useful')
```
ChrisBAshton: This doesn't sound very promising about our forms! Perhaps `this.$surveyForm` is better?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -90,6 +107,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         that.$somethingIsWrongForm.append($('<input type="hidden" name="referrer">').val(document.referrer || 'unknown'))
       }
 
+      function setHiddenValuesNotUsefulForm () {
+        var currentPathname = window.location.pathname;
+        var encodedPathname = encodeURI(currentPathname);
+        var finalPathName = encodedPathname.replace(/[^\s=\/?&]+(?:@|%40)[^\s=\/?&]+/, '[email]')
```
ChrisBAshton: I suggest doing any regex replacement _before_ you encode the path name - probably easier to understand!

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         revealInitialPrompt()
       })
 
+      this.$pageIsNotUsefulButton.on('click', function (e) {
+        var cookie = document.cookie.split("; ") ;
```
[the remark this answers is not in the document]
JonathanHallam: 👍

app/assets/javascripts/govuk_publishing_components/components/feedback.js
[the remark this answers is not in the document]
JonathanHallam: 👍

JonathanHallam: I've addressed a lot of these comments by using `window.GOVUK.cookie('_ga')` rather than this funky iteration so will resolve them.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
JonathanHallam: 👍

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         revealInitialPrompt()
       })
 
+      this.$pageIsNotUsefulButton.on('click', function (e) {
+        if (window.GOVUK.cookie('_ga') === null) {
```
vanitabarrett: This won't catch if the cookie value has been set to blank (which is what happens when someone opts out of Google Analytics cookies via https://www.gov.uk/help/cookies )

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +50,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+          window.GOVUK.setCookie('_ga', '111111111.1111111111', {})
```
vanitabarrett: I'm not sure we want to be setting the cookie to the dummy value - this may get confusing with our opt-in/opt-out method. I think explicitly setting a variable (as you were doing before) and passing this through to `setHiddenValuesNotUsefulForm()` is probably preferable.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -90,6 +98,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         that.$somethingIsWrongForm.append($('<input type="hidden" name="referrer">').val(document.referrer || 'unknown'))
       }
 
+      function setHiddenValuesNotUsefulForm () {
+        var currentPathName = window.location.pathname.replace(/[^\s=?&]+(?:@|%40)[^\s=?&]+/, '[email]')
+        var finalPathName = encodeURI(currentPathName)
+
+        that.$surveyForm.append($('<input name="email_survey_signup[ga_client_id]" type="hidden">').val(window.GOVUK.cookie('_ga') || '0'))
```
vanitabarrett: Same comment as above - I'd prefer to fetch the value of the cookie and assign that to a variable we use here, rather than setting the value of the cookie and reading from that again.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -90,6 +98,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        that.$surveyForm.append('<a href="https://www.smartsurvey.co.uk/s/gov-uk-banner/?c=' + finalPathName + '&amp;gcl=' + window.GOVUK.cookie('_ga') + 'class="gem-c-feedback__email-link govuk-link" id="take-survey" target="_blank" rel="noopener noreferrer">Don’t have an email address?</a>')
```
vanitabarrett: Same comment as above - I'd prefer to fetch the value of the cookie and assign that to a variable we use here, rather than setting the value of the cookie and reading from that again.

spec/javascripts/components/feedback-spec.js
```diff
@@ -106,6 +105,20 @@ describe("Feedback component", function () {
     expect($('#something-is-wrong').find("[name=referrer]").val()).toBe("unknown");
   });
 
+  it("should append a hidden 'ga_client_id' field to the form", function() {
+    loadFeedbackComponent();
+    $('a.js-page-is-not-useful').click();
+    console.log($('#page-is-not-useful').find('#testing-feedback'))
```
vanitabarrett: Stray `console.log`

spec/javascripts/components/feedback-spec.js
```diff
@@ -106,6 +105,20 @@ describe("Feedback component", function () {
     expect($('#something-is-wrong').find("[name=referrer]").val()).toBe("unknown");
   });
 
+  it("should append a hidden 'ga_client_id' field to the form", function() {
+    loadFeedbackComponent();
+    $('a.js-page-is-not-useful').click();
```
vanitabarrett: Should be able to target this without the `a` in the selector.

spec/javascripts/components/feedback-spec.js
```diff
@@ -106,6 +105,20 @@ describe("Feedback component", function () {
...
+    console.log($('#page-is-not-useful').find('#testing-feedback'))
+    expect($('#page-is-not-useful').find("[name='email_survey_signup[ga_client_id]']").val()).toBe("111111111.1111111111");
```
vanitabarrett: You already have a test further down that tests the default value is being set correctly. This test should check what happens if the ga cookie is available. You should be able to run `window.GOVUK.cookie('_ga', 'a value here') at the start of your test so you can check this.

spec/javascripts/components/feedback-spec.js
```diff
@@ -106,6 +105,20 @@ describe("Feedback component", function () {
...
+  });
+
+  it("should append the \'Don’t have an email address?\' link at the bottom of the form", function(){
+    loadFeedbackComponent();
+    $('a.js-page-is-not-useful').click();
```
vanitabarrett: Should be able to target this without the `a` in the selector.

spec/javascripts/components/feedback-spec.js
```diff
@@ -189,6 +202,13 @@ describe("Feedback component", function () {
       expect(GOVUK.analytics.trackEvent).
         toHaveBeenCalledWith('yesNoFeedbackForm', 'ffNoClick');
     });
+
+    it("sets a default Google Analytics Client ID if none is available", function () {
+      loadFeedbackComponent();
+      $('a.js-page-is-not-useful').click();
```
vanitabarrett: Should be able to target this without the `a` in the selector.

vanitabarrett:
Do you have an image of what this looks like compared to the old feedback form, so we can check the visual changes?

JonathanHallam:
@vanitabarrett No image yet, have failed to get static running locally which I need to get any of the frontend apps up where it might be visible. I'm off volunteering tomorrow but I'll get that done on Friday.

vanitabarrett:
@JonathanHallam Great! I'll be WFH on Friday but give me a shout if you need a hand with getting it running

app/assets/javascripts/govuk_publishing_components/components/feedback.js
JonathanHallam: 👍

app/assets/javascripts/govuk_publishing_components/components/feedback.js
JonathanHallam: 👍

spec/javascripts/components/feedback-spec.js
JonathanHallam: 👍

spec/javascripts/components/feedback-spec.js
JonathanHallam: 👍

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -49,6 +51,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         revealInitialPrompt()
       })
 
+      this.$pageIsNotUsefulButton.on('click', function (e) {
+        var gaClientId
+        if (window.GOVUK.cookie('_ga') === null || window.GOVUK.cookie('_ga') === '') {
+          gaClientId = '111111111.1111111111'
```
vanitabarrett: To make it clear what this is doing, it might be worth saving "111111111.1111111111" to a variable, e.g: dummyGAClientID.

JonathanHallam: okay have done this and fixed a test that was failing. Hopefully builds will pass now.

vanitabarrett:
@JonathanHallam Probably worth squashing all the commits into one 🙂

CHANGELOG.md
```diff
@@ -7,29 +7,9 @@
...
-## 17.7.0
-
-* Add an inverse flag to the phase banner (PR #954)
-
-## 17.6.1
-
-* Fix confirmation message being read to screenreaders (PR #952)
-* Hide 'Accept Cookies' button when Javascript not available (PR #948)
-
-## 17.6.0
```
andysellick: Why has all of this been removed from the changelog? You shouldn't be removing anything, it's a record of what's changed.

vanitabarrett: @JonathanHallam @andysellick I'm guessing this might have happened as a result of a bad rebase?

andysellick: Yeah, we're looking at it now.

PR #959, opened 2019-06-26
Remove old cookie banner code

vanitabarrett:
## What
Removes old cookie banner code. Includes:

- Removing specific HTML, CSS and JS for the old cookie banner
- Removing the need for the `new_cookie_banner: true` flag
- Removing old cookie banner tests and combining into one test file
- Keeping certain CSS and JS selectors for `gem-c-cookie-banner--new` in case of static caching on deploy. These will be removed once this change has been deployed.

## Why
The old cookie banner was deployed on 18/06/2019 (one week ago). We want to remove the old cookie banner code as it is no longer being used and having old code and examples in the component guide could be confusing for other developers.

## Before
Screen Shot 2019-06-26 at 11 39 17

## After
Screen Shot 2019-06-26 at 11 39 00


vanitabarrett:
[Remove outdated cookie banner code once we've switched to the new banner](https://trello.com/c/Upyyk2yu/215-remove-outdated-cookie-banner-code-once-weve-switched-to-the-new-banner)

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -1,5 +1,4 @@
-$govuk-cookie-banner-background: #d5e8f3;
-$govuk-cookie-banner-background-new: govuk-colour("white");
+$govuk-cookie-banner-background: govuk-colour("white");
 $govuk-cookie-banner-text-green: #00823b;
```
andysellick: Probably outside the scope of this PR, but does this need to be a variable - it looks like it's only used once?

spec/javascripts/components/cookie-banner-spec.js
```diff
@@ -1,46 +1,114 @@
 /* eslint-env jasmine, jquery */
 /* global GOVUK */
 
-var container
+var container, element;
 
 var GOVUK = window.GOVUK || {};
 
-describe('Cookie banner', function () {
+describe('New cookie banner', function () {
```
andysellick: If this is making the new cookie banner into the regular cookie banner, shouldn't this be called `Cookie banner`?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
vanitabarrett: @andysellick I think this is more to appease the linter

PR #969, opened 2019-07-01
[DO NOT MERGE] Change title margin option

andysellick:
**Do not merge this PR until other PRs have been raised to account for the change in this component. Worst case scenario is that some titles won't have the correct bottom margin, so not exactly a breaking change but we want to catch it nonetheless.**

## What

Updates the title component to use the govuk-spacing-scale for bottom margin, instead of using a custom scale previously defined. Uses the `shared_helper` for this.

## Why

This brings this component in line with other components that are now using the same mechanism for a variable bottom margin.

## Visual Changes

The component previously had a default margin top and bottom of 45px, scaling to 15px on mobile. The govuk spacing scale doesn't have an exact equivalent, so I've gone for the nearest, which is 40px, scaling to 25px on mobile.

For reference when updating apps:

- 4 (gutter, 30px) change to 6
- 5 (gutter-half, 15px) change to 3

Related PRs:

- https://github.com/alphagov/collections/pull/1166
- https://github.com/alphagov/finder-frontend/pull/1238

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/title.yml
+++ b/app/views/govuk_publishing_components/components/docs/title.yml
@@ -43,12 +43,12 @@ examples:
       average_title_length: long
   with_margin:
     description: |
-      Configurable bottom margin for the component. This has been built in anticipation of being compatible with a theoretical future model for component margins.
+      The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). It defaults to having a margin bottom of 50px.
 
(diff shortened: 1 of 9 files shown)


app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -43,12 +43,12 @@ examples:
       average_title_length: long
   with_margin:
     description: |
-      Configurable bottom margin for the component. This has been built in anticipation of being compatible with a theoretical future model for component margins.
+      The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](http://govuk-frontend-review.herokuapp.com/docs/#settings/spacing-variable-govuk-spacing-points). It defaults to having a margin bottom of 40px.
```
nickcolley: Consider linking to the official design system documentation here, as the frontend review application is subject to change over time.

https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale

andysellick: Excellent suggestion. I've now updated this in this yml file and all others that had this link. Thanks.

andysellick:
> This looks great @andysellick! Only one suggestion: Could we default to 8?

Updated.

PR #971, opened 2019-07-03
Add description text

DilwoarH:
## What
Add support for description text

## Why
A lot of "one question per page" views on GOV.UK have a description under the title. This means that "hacks" are being put in place to deliver user need.

Example:
https://www.gov.uk/prepare-business-uk-leaving-eu?page=2

## Visual Changes
Screen Shot 2019-07-03 at 12 55 44

Screen Shot 2019-07-03 at 15 10 19

### DIFF
--- a/app/views/govuk_publishing_components/components/_checkboxes.html.erb
+++ b/app/views/govuk_publishing_components/components/_checkboxes.html.erb
@@ -10,6 +10,10 @@
       <%= tag.fieldset class: "govuk-fieldset", "aria-describedby": cb_helper.fieldset_describedby do %>
         <%= cb_helper.heading_markup %>
 
+        <% if cb_helper.description %>
+          <%= tag.p cb_helper.description, class: "govuk-body" %>
+        <% end %>
+
         <% if cb_helper.hint_text %>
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -10,6 +10,10 @@
       <%= tag.fieldset class: "govuk-fieldset", "aria-describedby": cb_helper.fieldset_describedby do %>
         <%= cb_helper.heading_markup %>
 
+        <% if cb_helper.description %>
+          <%= tag.p cb_helper.description, id: "#{id}-description", class: "govuk-body" %>
```
andysellick: What's the ID for on this paragraph? Do we need it?

DilwoarH: I don't think so, I'll remove it

leenagupte:
@DilwoarH I think you needed to add an "Unreleased Changes" section in the Change note.

DilwoarH:
> @DilwoarH I think you needed to add an "Unreleased Changes" section in the Change note.

Yes you're right, I noticed after merging. I've added it to the bump up PR #982

PR #975, opened 2019-07-09
Reset focus when user closes feedback form before submission

maxgds:
https://trello.com/c/RVYM3DPP/1-3-survey-focus-skipping-content-12
## What
This resets the focus back to the link that was activated to
open the form when it is closed.

## Why
Currently if the user closes the "not useful" or "something is wrong" feedback
form the focus is lost. This is an accessibility issue as keyboard users end up
back at the start of the page, and screenreader users are unaware of where they
are on the page. 

### Before
Screen Shot 2019-07-09 at 08 57 23
 

### After
Screen Shot 2019-07-09 at 08 57 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNRELEASED
+
+* Reset focus when feedback component is closed before submit (PR #975)
+
 ## 17.12.0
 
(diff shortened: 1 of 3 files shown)


injms:
Not a major thing and not caused by this change, but would it be worth fixing the underline that's being lost in Edge and Firefox on Windows?
image

I think it's a minor thing because this component seems to be mostly used above a footer with a blue border  which makes the white underline visible.

image

PR #977, opened 2019-07-09
Update search component to use govuk-frontend

alex-ju:
## What
Update `search.scss` to use govuk-frontend mixins and variables

## Why
To help [removing the dependency on the deprecated govuk_frontend_toolkit package](https://trello.com/c/ajPzDAOX).

## Visual Changes
No visual changes.

Screen Shot 2019-07-09 at 17 10 16

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -10,15 +10,13 @@ $large-input-size: 50px;
   display: block;
 
   h1 {
-    @include core-19($line-height: $input-size, $line-height-640: $input-size);
+    @include govuk-font($size: 19, $line-height: $input-size);
     margin: 0;
   }
 }
 
 .gem-c-search__input[type="search"] { // overly specific to prevent some overrides from outside
-  @include box-sizing(border-box);
-  @include core-19($line-height: (28 / 19), $line-height-640: (28 / 16));
-  @include appearance(none);
+  @include govuk-font($size: 19, $line-height: (28 / 19));
   @include govuk-focusable;
 
   padding: 6px;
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -42,20 +44,20 @@ $large-input-size: 50px;
 @if ($is-ie == false) or ($ie-version >= 8) {
```
kr8n3r: this i believe is still from govuk_frontend_toolkit

alex-ju: Good spot! Will look into it

alex-ju: Updated and removed these conditions which from my understanding were aimed to prevent some styles from being passed to IE6 and 7. The reasons for removing them is that:
- this library relies on govuk-frontend, which supports Internet Explorer down to version 8
- removing the conditions (adding these styles to older browsers) doesn't seem to affect the way the component is being displayed in IE6 and 7.

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -38,131 +40,124 @@ $large-input-size: 50px;
...
+    @include govuk-font($size: 19, $line-height: $input-size);
 
-      position: absolute;
-      left: 15px;
-      top: 1px;
-      z-index: 1;
-      color: $secondary-text-colour;
-    }
+    position: absolute;
+    left: 15px;
```
kr8n3r:
```suggestion
    left: govuk-spacing(3);
```

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
kr8n3r: 👍

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
alex-ju: Added the suggestion. I skipped it as Andy has some concerns about using the spacing scale in other places then margins and paddings.

PR #978, opened 2019-07-10
Move tab style overrides to the gem

vanitabarrett:
## What 
Trello: https://trello.com/c/dnszYbmt/7-accessibility-of-tabs-on-govuk

Moves CSS 'overrides' for the tab component from Frontend into the gem.
Follow on PR in Frontend: https://github.com/alphagov/frontend/pull/1898

## Why
The tab styles are affected by `.content-block` styles which are [defined in Static](https://github.com/alphagov/static/blob/a815620cada7ea1c65428c1c3b3ac4dbe28977bf/app/assets/stylesheets/helpers/_text.scss). In Frontend, there are [overrides for this problem](https://github.com/alphagov/frontend/blob/master/app/assets/stylesheets/application.scss#L70). However, they don't work properly with our current tabs on mobile (the left spacing is incorrect). This corrects the overrides and moves them into the gem, so they're in one common place and more visible to those making changes to the tab component.

## Before
Screen Shot 2019-07-10 at 11 20 06

## After
Screen Shot 2019-07-10 at 11 19 56

__Note: there should be no visible change within the component guide or other apps that don't use `content-block`:__

Screen Shot 2019-07-10 at 11 21 06

https://govuk-publishing-compon-pr-978.herokuapp.com/component-guide/tabs

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
@@ -7,3 +7,20 @@
     border: 0;
   }
 }
+
+// We have some styles within GOVUK (.content-block) which can leak into the list styles for this component.
+// These styles are defined in Static: https://github.com/alphagov/static/blob/a815620cada7ea1c65428c1c3b3ac4dbe28977bf/app/assets/stylesheets/helpers/_text.scss
+// This sets more specific selectors so those unwanted styles are overidden
+// scss-lint:disable QualifyingElement
+// sass-lint:disable no-qualifying-elements
+ul.govuk-tabs__list {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
```diff
@@ -7,3 +7,19 @@
     border: 0;
   }
 }
+
+// We have some styles within GOVUK (.content-block) which can leak into the list styles for this component.
```
maxgds: I'd suggest linking or at least referencing exactly where that code is as you have in the PR description, so people can potentially clean up later with a quick check.

PR #981, opened 2019-07-11
Fix duplicate link bug with feedback form and fix spacing on small viewport

vanitabarrett:
# Fix duplicate link bug with feedback form
## What
Add a check to the feedback form logic so that we don't insert the links if they're already present, to avoid duplication.

## Why
The hidden input for GA ID and the "Don't have an email address?" link are inserted using Javascript. This is done when the user clicks 'No' within the feedback form and the form opens. However, there is no check for whether the links have already been inserted. This means that the links are appended each time the user opens the 'No' form.

*Replication Steps:*

- Click "No" on the feedback form.
- Without interacting with the form itself, close the form via the Close button
- Click "No" on the feedback form again
- Duplicate 'Don't have an email address?' links will be visible

## Before
Screen Shot 2019-07-11 at 09 31 15

## After
Screen Shot 2019-07-11 at 09 37 37

# Fix spacing on small viewport
## What
Add spacing around the feedback form on small screen sizes

## Before
Screen Shot 2019-07-11 at 09 52 17

## After
Screen Shot 2019-07-11 at 10 08 05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Stop duplicate links within feedback form (PR #981)
+* Add spacing to the feedback form (PR #981)
+
 ## 17.12.2
 
 * Make some tab component styles more specific to override `.content-block` (PR #978)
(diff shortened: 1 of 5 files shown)


vanitabarrett:
@maxgds I've added a fix for the spacing issue, should be ready to review now

PR #984, opened 2019-07-12
Moves axe-core to devDependencies

MuriloDalRi:
## What
Moves axe-core to devDependencies.

## Why
This is a dev dependency and does not need to be installed in production.


nickcolley:
Now that I think about it, axe-core is used in the publishing guide pages itself at runtime, which is likely why it's not a devDependency...

MuriloDalRi:
> Now that I think about it, axe-core is used in the publishing guide pages itself at runtime, which is likely why it's not a devDependency...

I'll close this then

PR #986, opened 2019-07-12
Add view changes link to PR template

DilwoarH:
## What
Add view changes link to PR template

## Why
This will help users view changes easily, this was included in the template previously but has been removed since.

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -7,3 +7,8 @@
 
 ## Visual Changes
 <!-- If the change results in visual changes, show a before and after -->
+
+<!--
+## View Changes
+https://govuk-publishing-compon-pr-[PR-NUMBER].herokuapp.com/
+-->


vanitabarrett:
This was previously removed because @tijmenb [found that a lot of PRs were kept with [PR-NUMBER] in the description](https://github.com/alphagov/govuk_publishing_components/commit/710c82102a0fc137b564d486aad7f3ae99a53620#diff-ac354b58e960ffc4b699664e7f8a832d) and we now have the deployed environments button at the bottom of the PR when it's deployed to Heroku.

However I think there's still value in including a link within the PR description as it makes it more visible, so I'm happy to try this again and monitor if it's used or not.

DilwoarH:
> This was previously removed because @tijmenb [found that a lot of PRs were kept with [PR-NUMBER] in the description](https://github.com/alphagov/govuk_publishing_components/commit/710c82102a0fc137b564d486aad7f3ae99a53620#diff-ac354b58e960ffc4b699664e7f8a832d) and we now have the deployed environments button at the bottom of the PR when it's deployed to Heroku anyway.
> 
> However I think there's still value in including a link within the PR description as it makes it more visible, so I'm happy to try this again and monitor if it's used or not. Interested to know what others think?

Thnks for this @vanitabarrett - wasn't aware. I'm happy to try again. Is there anyone else that has a different view?

alex-ju:
When I review I use the "View deployment" link (when the app is still alive on Heroku). When I raise a PR I usually add a link to the preview app to specific page(s) to ease the review.

I personally don't feel a need for it, but if you feel that it brings value I'm happy to conform. I may be a bigger challenge of getting people to use the template in the first place.

andysellick:
I think a common problem we have is when PRs are raised that address issues on specific pages but then don't link to those specific pages. That's more of a problem for changes in apps, though. I don't know what the solution is to that, other than to model best practice and gently nag people to do it. 

As Vanita said, we used to have this change and people didn't use it much, so not sure it's worth trying to reintroduce. Usually for this repo it's not hard to find the deployment link, as it's always in /component-guide.

docs/pull_request_template.md
```diff
@@ -7,3 +7,6 @@
 
 ## Visual Changes
 <!-- If the change results in visual changes, show a before and after -->
+
+## View Changes
+https://govuk-publishing-compon-pr-[PR-NUMBER].herokuapp.com/
```
injms: To avoid having a pull request with [PR-NUMBER] visible, we could wrap it in a comment? This would mean that the link can still be easily added if required, but won't show unless it's edited.

```suggestion
<!-- https://govuk-publishing-compon-pr-[PR-NUMBER].herokuapp.com/ -->
```

DilwoarH:
@andysellick @vanitabarrett @injms @alex-ju 

I've commited @injms 's suggestion with putting it as a comment. I think that was a good suggestion. Let me know what you think.

PR #987, opened 2019-07-12
bump to version 17.14.0

DilwoarH:
Includes:
* Change paragraph to div for descriptions to support govspeak (PR #985)
* Update cookie banner component to hide if cross-origin message has `hideCookieBanner` set to `true` (PR #988)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 17.14.0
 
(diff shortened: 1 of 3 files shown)


vanitabarrett:
I would say this is probably a patch release, so 17.13.1

DilwoarH:
@vanitabarrett good point, updated

alex-ju:
Now that #988 is in, could we rebase and make it 17.14.0 again? 😁

DilwoarH:
@alex-ju updated

PR #989, opened 2019-07-14
Set up data tracking for navigation on secondary content

DilwoarH:
## What

**Note: this card is for setting up data tracking, it is not about doing the analysis**

We want to be able to track usage of the step by step side bar and the breadcrumb on secondary content.

Is the data tracking in place? What are the metrics we are trying to measure?

Would we be able to answer these questions:

- Does adding step by step navigation make a difference to user journeys (eg next page paths, exits) for *users who land directly* (eg from Google) on a secondary content page with step by step navigation? And compared to users who come from GOV.UK pages?
- Does the step by step navigation being closed make a difference to user journeys compared to the step by step being opened at a certain step, for secondary content?

## Why
We want to add a new feature to 'step by step publisher' that makes it possible to add step by step navigation to secondary content. (Secondary content = content that is about the service but is not useful enough to include in the step by step).

https://trello.com/c/LU6AAcOk/14-step-by-step-set-up-data-tracking-for-navigation-on-secondary-content

### DIFF
--- a/lib/govuk_publishing_components/presenters/meta_tags.rb
+++ b/lib/govuk_publishing_components/presenters/meta_tags.rb
@@ -1,9 +1,9 @@
 module GovukPublishingComponents
   module Presenters
     class MetaTags
-      attr_reader :content_item, :details, :links, :local_assigns
+      attr_reader :content_item, :details, :links, :local_assigns, :request
 
-      def initialize(content_item, local_assigns)
+      def initialize(content_item, local_assigns, request)
         # We have to call deep_symbolize_keys because we're often dealing with a
         # parsed JSON document which will have string keys by default, but our
         # components use symbol keys and we want consistency.
@@ -11,6 +11,7 @@ module GovukPublishingComponents
         @details = @content_item[:details] || {}
         @links = @content_item[:links] || {}
         @local_assigns = local_assigns
+        @request = request
       end
 
       def meta_tags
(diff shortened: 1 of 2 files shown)


lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -98,6 +99,12 @@ def add_step_by_step_tags(meta_tags)
         stepnavs = links[:part_of_step_navs] || []
         stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
         meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content.present?
+
+        step_nav_helper = PageWithStepByStepNavigation.new(content_item, request.path, request.query_parameters)
+        if step_nav_helper.show_secondary_step_by_step?
+          meta_tags["govuk:secondary-stepnav-shown"] = step_nav_helper.secondary_step_by_step[:content_id]
+        end
```
ChrisBAshton: I'd consider doing all the setup at the beginning of the method and doing the meta_tags assignment at the end of the function, using one-line-ifs for both for consistency:

```ruby
def add_step_by_step_tags(meta_tags)
  stepnavs = links[:part_of_step_navs] || []
  stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
  stepnav = PageWithStepByStepNavigation.new(content_item, request.path, request.query_parameters)

  meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content.present?
  meta_tags["govuk:secondary-stepnav-shown"] = stepnav.secondary_step_by_step[:content_id] if stepnav.show_secondary_step_by_step?
end
```

DilwoarH: The reason why I chose not to use inline ifs was that I wanted to improve the readability of the code. Where lines are more than 100 characters long, is it very difficult to read and follow (https://batsov.com/articles/2013/06/26/the-elements-of-style-in-ruby-number-1-maximum-line-length/ - this article goes into some detail about this if you're interested).

I've improved the logic a bit more

PR #990, opened 2019-07-15
Add no underline option to document list component

andysellick:
## What

Add an option to the document list component to remove underlines from links.

## Why

We intend to use the component for search results in finder-frontend, but the current design doesn't have underlines on the links. Adding them adds to the visual clutter of the search results pages and leads to a much bigger conversation, so it seems more sensible to have this option available.

## Visual Changes

Option to not have underlines looks like this:

Screen Shot 2019-07-15 at 14 49 48

### DIFF
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -1,9 +1,13 @@
 <%
   items ||= []
+
+  classes = "gem-c-document-list"
+  classes << " gem-c-document-list--top-margin" if local_assigns[:margin_top]
+  classes << " gem-c-document-list--bottom-margin" if local_assigns[:margin_bottom]
+  classes << " gem-c-document-list--no-underline" if local_assigns[:remove_underline]
+
   within_multitype_list ||= false
   within_multitype_list_class = " gem-c-document-list__multi-list" if within_multitype_list
-  margin_top_class = " gem-c-document-list--top-margin" if local_assigns[:margin_top]
-  margin_bottom_class = " gem-c-document-list--bottom-margin" if local_assigns[:margin_bottom]
   title_with_context_class = " gem-c-document-list__item-title--context"
 
   brand ||= false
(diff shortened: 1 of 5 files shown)


nickcolley:
Is visual clutter a user need? How does this impact accessibility?

andysellick:
> Is visual clutter a user need? How does this impact accessibility?

What I'm trying to do is use a component without having to start this wider conversation. This change will mean that search results don't change - and search pages have already been checked for user needs/accessibility/etc. 

If I don't make this change, I have to start that conversation for all of our search pages, which is a much bigger thing.

app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -156,6 +156,27 @@ examples:
           public_updated_at: 2017-07-19 15:01:48
           document_type: 'Statistical data set'
         subtext: 'First published during the 1996 Conservative Government'
+  without_underline:
+    description: Where the component is used for search results, the current design does not include underlines on links.
```
injms: I think that removing the underlines for this component is a reasonably finder-frontend-specific use case that's worth mentioning that next to the component variation. The wording is definitely a suggestion, but I hope get's that across.

```suggestion
    description: The current search design does not include underlines on links and has been tested without underlines. Other uses will require further user testing.
```

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -5,13 +5,14 @@
...
   title_with_context_class = " gem-c-document-list__item-title--context"
+  underline_class = " gem-c-document-list--no-underline" if local_assigns[:remove_underline]
 
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
 %>
 <% if items.any? %>
   <% unless within_multitype_list %>
-    <ol class="gem-c-document-list<%= margin_bottom_class %><%= margin_top_class %>">
+    <ol class="gem-c-document-list<%= margin_bottom_class %><%= margin_top_class %><%= underline_class %>">
```
injms: Just a readability suggestion - do you think it's worth pushing these to a single variable to then be used in the class of the `ol` element? 

For example:
```ruby
list_class = ['gem-c-document-list', underline_class, margin_top_class, margin_bottom_class]
```

then

```ruby
<ol class="<% list_class %>">
```

andysellick: I did, but I was hoping to get away with just one more... okay, I'll change it 😁

andysellick: Changed.

PR #993, opened 2019-07-17
(DNM until design consensus) - Replace govspeak steps background images with css counters

kr8n3r:
## What
Replace the use of background images with CSS to indicate step numbers.
Component preview: https://govuk-publishing-compon-pr-993.herokuapp.com/component-guide/govspeak/steps/preview

## Why
CSS counters are supported on IE8 and above so we can safely use them
to visually indicate the steps instead of images.

Added bonus is that numbers retain crispness with zoom :)
## Bonus
- numbers retain crispness with zoom :)
- as images are part of _govuk_frontend_toolkit_ this further reduces the dependency on it

### There are minor visual difference, so needs a designer consultation.

**Old:**
Screen Shot 2019-07-17 at 10 34 30

**New:**
Screen Shot 2019-07-17 at 10 34 26

**New with long text:**
Screen Shot 2019-07-18 at 12 05 18

**New (left) vs old (right):**
Screen Shot 2019-07-17 at 10 44 39

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Replace govspeak steps background images with css counters (PR #993)
+
 ## 17.16.0
 
 * Track when primary or secondary step by step are shown (PR #989)
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
@@ -1,23 +1,36 @@
 .gem-c-govspeak .steps {
   padding-left: 0;
   margin-left: 0;
+  counter-reset: step-counter;
   overflow: hidden;
 
   > li {
-    background-position: 0 .87em;
-    background-repeat: no-repeat;
+    counter-increment: step-counter;
     list-style-type: decimal;
     margin-left: 0;
     padding: .75em 0 .75em 2.2em;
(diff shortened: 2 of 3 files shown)


kr8n3r:
_Where are step icon images located, as they can now be deleted?_

jeremyhhy2:
Can the numbers in the circle be a bit smaller?
Would you ever get double figures? What does that look like?

kr8n3r:
@jeremyhhy2 
font-size is currently 16px, can drop to 14 but will need top alignment offset adjustment

with double digits
Screen Shot 2019-07-18 at 12 23 30

jeremyhhy2:
@markhurrell  @miaallers What do you think?

@kr8n3r Have you tried adjusted the numbers to the right slightly?

Feels like the numbers are not centred inside the circle, maybe because we are using 16px? The number is like 1 or 2 pixels off to the left?

kr8n3r:
@jeremyhhy2 I've added more items to the preview page https://govuk-publishing-compon-pr-993.herokuapp.com/component-guide/govspeak/steps/preview

you can play around in the browser and suggest changes

alex-ju:
@kr8n3r, @miaallers suggests that we could use the tabular version of the font for numbers. what do you think of that?

kr8n3r:
tabulars are wacky :) see below
Screen Shot 2019-07-18 at 14 03 28

would need drop font-size and increase padding. that would result in larger padding on 0-9 than double digits

kr8n3r:
closing in favour of leaving it as it is

PR #988, opened 2019-07-12
Hide cookie banner if cross-origin message has hideCookieBanner set to true

alex-ju:
## What
Update cookie banner component to hide if cross-origin message has `hideCookieBanner` set to `true`.

The cookie banner script listens for cross-origin messages and if they're coming from a `publishing.service.gov.uk` domain and have `hideCookieBanner` set to `true` it hides the banner. 

## Why
This behaviour is required for [previewing GOV.UK pages from draft stack in Content Publisher without obstructing the view](https://github.com/alphagov/content-publisher/pull/1240).

## Visual Changes
No visual changes.

[Trello card](https://trello.com/c/QXbQKBG3)

PR #995, opened 2019-07-18
Remove cross origin approach and detect parent document location

alex-ju:
## What
This is an alternative to the approach taken on #988 (which causes the cookie banner to flicker before removing it). This approach checks if the component is rendered in an iframe and the parent document location is a publishing domain (*.publishing.service.gov.uk) to prevent the cookie banner from being displayed.

## Why
This behaviour is required for [previewing GOV.UK pages from draft stack in Content Publisher without obstructing the view](https://github.com/alphagov/content-publisher/pull/1240).

## Visual Changes
No visual changes

[Trello card](https://trello.com/c/QXbQKBG3)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -14,10 +14,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.cookieBannerConfirmationMessage = this.$module.querySelector('.gem-c-cookie-banner__confirmation')
 
     this.setupCookieMessage()
-
-    // Listen for cross-origin communication messages (e.g. hideCookieBanner for when previewing GOV.UK pages
-    // in publishing applications
-    this.listenForCrossOriginMessages()
   }
 
   CookieBanner.prototype.setupCookieMessage = function () {
@@ -44,7 +40,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   CookieBanner.prototype.showCookieMessage = function () {
     // Hide the cookie banner on the cookie settings page, to avoid circular journeys
-    if (this.$module.cookieBanner && window.location.pathname === '/help/cookies') {
+    // or when presented in an iframe by a publishing application
+    if (this.isInCookiesPage() || (this.isInIframe() && this.parentIsPublishingDomain())) {
       this.$module.style.display = 'none'
     } else {
       var shouldHaveCookieMessage = (this.$module && window.GOVUK.cookie('seen_cookie_message') !== 'true')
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -44,7 +40,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   CookieBanner.prototype.showCookieMessage = function () {
     // Hide the cookie banner on the cookie settings page, to avoid circular journeys
-    if (this.$module.cookieBanner && window.location.pathname === '/help/cookies') {
+    // or when presented in an iframe by a publishing application
+    if (this.isInCookiesPage() || (this.isInIframe() && this.parentIsPublishingDomain())) {
       this.$module.style.display = 'none'
```
benthorner: Is this line necessary, if it's hidden by default?

spec/javascripts/components/cookie-banner-spec.js
```diff
@@ -113,20 +113,4 @@ describe('Cookie banner', function () {
 
     expect(newCookieBanner).not.toBeVisible()
   })
-
```
benthorner: Sadness. Might we have a test? ;-)

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -89,36 +86,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-    }
+  CookieBanner.prototype.isInIframe = function () {
+    return window.parent && window.location !== window.parent.location
+  }
 
-    // Polyfill origin.endsWith(trustedDomain) for IE
-    var offset = origin.length - trustedDomain.length
-    var trustedOrigin = offset >= 0 && origin.lastIndexOf(trustedDomain, offset) === offset
+  CookieBanner.prototype.parentIsPublishingDomain = function () {
+    var publishingDomain = 'publishing.service.gov.uk'
```
benthorner: It would be good if this worked the same way in the dev environment too :-).

spec/javascripts/components/cookie-banner-spec.js
alex-ju: Sure, will think of what's a good way of testing it - raised it quickly to get feedback on.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
alex-ju: I will dig a bit more before removing it, but it seems to not have any effects.

kevindew:
Nice one Alex 👍 

I'm wondering if we need to check the hostname at all and it might be sufficient to just make the call that this message isn't shown in an iframe. Shouldn't really be an issue as only the draft stack can be embedded in an iframe.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
alex-ju: Updated refactored the conditional statement and removed it.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
alex-ju: Removed the `parentIsPublishingDomain` requirement based on Kevin's suggestion.

spec/javascripts/components/cookie-banner-spec.js
alex-ju: Updated with a test to check the cookie banner is hidden when used in an iframe (`window.parent` is different than `window`)

alex-ju:
@kevindew Updated to drop the domain checking.

PR #1002, opened 2019-07-23
Changes content item to have symbolic keys

DilwoarH:
## What
Changes content item to have symbolic keys.

## Why
We have to call deep_symbolize_keys because we're often dealing with a
parsed JSON document which will have string keys by default, but our
components use symbol keys and we want consistency.

## Visual Changes
None

## CHANGE NOTE NOT ADDED
This was not added to change notes as this is a bug fix and does not provide any usefulness in change notes

https://trello.com/c/LU6AAcOk/14-step-by-step-set-up-data-tracking-for-navigation-on-secondary-content

### DIFF
--- a/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
+++ b/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb
@@ -4,7 +4,7 @@ module GovukPublishingComponents
     # Only used by the step by step component
     class PageWithStepByStepNavigation
       def initialize(content_store_response, current_path, query_parameters = {})
-        @content_item = content_store_response.to_h
+        @content_item = content_store_response.to_h.deep_symbolize_keys
         @current_path = current_path
         @query_parameters = query_parameters
       end
@@ -152,15 +152,15 @@ module GovukPublishingComponents
       end
 
       def parsed_step_navs
-        content_item.dig("links", "part_of_step_navs").to_a
+        content_item.dig(:links, :part_of_step_navs).to_a
       end
 
       def parsed_related_to_step_navs
-        content_item.dig("links", "related_to_step_navs").to_a
+        content_item.dig(:links, :related_to_step_navs).to_a
       end
 
       def parsed_secondary_to_step_navs
-        content_item.dig("links", "secondary_to_step_navs").to_a
+        content_item.dig(:links, :secondary_to_step_navs).to_a
       end
 
       def configure_for_sidebar(step_nav_content)


lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -100,7 +100,7 @@ def add_step_by_step_tags(meta_tags)
         stepnavs_content = stepnavs.map { |stepnav| stepnav[:content_id] }.join(",")
         meta_tags["govuk:stepnavs"] = stepnavs_content if stepnavs_content.present?
 
-        step_nav_helper = PageWithStepByStepNavigation.new(content_item, request.path, request.query_parameters)
+        step_nav_helper = PageWithStepByStepNavigation.new(content_item.deep_stringify_keys, request.path, request.query_parameters)
```
leenagupte: Having the code use labels in one place and strings in another is making the code get messy. I'd consider either: 
1. Making this class use string keys as well, or
2. If you want to use both strings and labels, change [line 10](https://github.com/alphagov/govuk_publishing_components/blob/1611c4d4fb8323e0bc978ddcc72587c2ace956af/lib/govuk_publishing_components/presenters/meta_tags.rb#L10) to use `with_indifferent_access`

DilwoarH: I don't feel like refactoring the whole class for one single use case is worthwhile especially when things like `with_indifferent_access` is being deprecated in future releases of ruby (ref: https://github.com/rails/rails/pull/28157). There is also a lot of performance issues with using `with_indifferent_access` (ref: https://gist.github.com/tiagoamaro/c82a27aceedfc901b081#file-results_with_4-2-5-txt).

```
Comparison:
Common Hash#[] access:  8306261.5 i/s
Common Hash#fetch access:  6053147.2 i/s - 1.37x slower
HashWithIndifferentAccess #[] String:  3803546.0 i/s - 2.18x slower
HashWithIndifferentAccess#fetch String:  1993671.6 i/s - 4.17x slower
HashWithIndifferentAccess#fetch Symbol:  1932004.0 i/s - 4.30x slower
HashWithIndifferentAccess #[] Symbol:  1422367.3 i/s - 5.84x slower
Hash#with_indifferent_access #[] String:   470876.8 i/s - 17.64x slower
Hash#with_indifferent_access #fetch String:   414701.6 i/s - 20.03x slower
Hash#with_indifferent_access #fetch Symbol:   410033.7 i/s - 20.26x slower
Hash#with_indifferent_access #[] Symbol:   381347.2 i/s - 21.78x slower
```

If this was used in multiple places, I would agree with your approach of changing it everywhere but I don't think this one use case justifies the change across the whole component.

leenagupte: So, I've had a look, and everything else in `PageWithStepByStepNavigation` is using symbol keys. This is because the keys are being symbolised here: https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb#L197

The only lines using strings are the ones that populate [parsed_step_nav](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/page_with_step_by_step_navigation.rb#L154), `parsed_related_to_step_navs` and `parsed_secondary_to_step_navs`, so I wonder how this is all working?

leenagupte: I'm happy to go ahead with the approach we talked about in real life to change `PageWithStepByStepNavigation` to use symbols everywhere.

DilwoarH: @leenagupte updated

PR #1004, opened 2019-07-23
Add origin to YouTube Player API

thomasleese:
The documentation for the YouTube Player API states that:

> If you are using the IFrame API, which means you are setting the enablejsapi parameter value to 1, you should always specify your domain as the origin parameter value.

https://developers.google.com/youtube/player_parameters#origin

We're currently seeing a Smokey error that suggests the origin is not being set correctly for the player:

```
https://s.ytimg.com/yts/jsbin/www-widgetapi-vfl2284kc/www-widgetapi.js 105 Failed to execute 'postMessage' on 'DOMWindow': The target origin provided ('https://www.youtube-nocookie.com') does not match the recipient window's origin ('https://draft-origin.publishing.service.gov.uk'). (Capybara::Chromedriver::Logger::JsError)
```

I'm not sure if this change will fix the Smokey error, but the documentation suggests that we should be setting `origin` and I've found other references online to that being a solution to this problem: https://stackoverflow.com/a/50518247

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Remove expectation that sprockets is installed when used in a Rails app (PR #999)
 * Fix tabs list overwrite on mobile (PR #1003)
+* Set the origin when rendering the YouTube player (PR #1004)
 
 ## 17.18.0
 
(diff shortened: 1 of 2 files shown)


thomasleese:
I'm not sure if there are any security implications about using `window.location.origin`, but I wasn't sure what the correct way of passing this value in using govuk_publishing_components is.

chao-xian:
Can confirm that this runs locally ok. Merging

chao-xian:
[Fix Smokey errors with youtube-nocookie](https://trello.com/c/YvdU9D39/33-fix-smokey-errors-with-youtube-nocookie)

PR #1008, opened 2019-07-26
Replace subscription links images with SVG

andysellick:
## What
The subscription links component uses PNG files for the images. Although there was a higher res option for mobile devices (oddly for only one of the icons) the result was still slightly blurry on some devices. This PR replaces these images with inline SVGs.

Before:

Screen Shot 2019-07-26 at 08 50 44

After:

Screen Shot 2019-07-26 at 08 50 33

IE8 doesn't support SVGs and I've not put in a fallback because even though we support it, it's a very old browser. I also don't believe these images are essential to understanding what this component is. IE8 doesn't show the images and the text moves left to fill the space, so it looks like this:

Screen Shot 2019-07-26 at 09 01 54

Tested in IE8 to current, current Edge, Chrome, Firefox, and a series of mobile devices. Interestingly IE9 (and maybe 10) still has problems with SVGs but the solution is to scale the SVG to the actual size it should appear (also set width and height, but ditch the viewBox attribute). This is why I've introduced a css scale transform to the Sass in order to reduce the size of the icons for the 'small' option (although I'm not sure where this is actually used, anyone know?).

As a bonus, because these images are now inline, this change removes the accessibility warning thrown up by axe in the component guide, where it couldn't determine the contrast ratio of the text due to a background image.

## Why
The images looked blurry on some devices and it's been bugging me for a while now.

## View Changes
https://govuk-publishing-compon-pr-1008.herokuapp.com/subscription-links/


app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -24,17 +24,19 @@
     >
       <% if sl_helper.email_signup_link.present? %>
         <li class="gem-c-subscription-links__list-item<%= ' gem-c-subscription-links__list-item--small' if local_assigns[:small_form] == true %>" >
+          <svg xmlns="http://www.w3.org/2000/svg" width="21" height="15.75" class="gem-c-subscription-links__icon"><path d="M19.687 0H1.312C.589 0 0 .587 0 1.313v13.124c0 .726.588 1.313 1.313 1.313h18.374c.725 0 1.313-.587 1.313-1.313V1.313C21 .587 20.412 0 19.687 0zm-2.625 2.625L10.5 7.875l-6.563-5.25h13.126zm1.313 10.5H2.625V3.937L10.5 10.5l7.875-6.563v9.188z"/></svg>
```
injms: Is it worth adding `role="presentation"` to the SVGs? I know it can be helpful for `img` elements, but I'm unsure about SVG elements.

andysellick: I was intending to put `aria-hidden="true"` to ensure that screen readers ignored them but voiceover on safari ignored them already. I also don't know if `role="presentation"` would be useful or not.

36degrees: I'd also add `focusable="false"` as SVGs are focusable (part of the tab order) by default in Internet Explorer 😕 

You can see how we treat the SVG in the header here:
https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/header/template.njk#L7-L27

36degrees: Bonus points if you put `fill="currentColor"` on the `path` – that way if users override colours in their browser the icons will modify themselves too.

Rather than this:

Screenshot_2019-07-26 Subscription links component - Component Guide

They'll look like this:

Screenshot_2019-07-26 Subscription links component - Component Guide(1)

andysellick: So many good ideas this morning, thanks guys. I've updated, please take a look.

app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -24,17 +24,19 @@
     >
       <% if sl_helper.email_signup_link.present? %>
         <li class="gem-c-subscription-links__list-item<%= ' gem-c-subscription-links__list-item--small' if local_assigns[:small_form] == true %>" >
+          <svg xmlns="http://www.w3.org/2000/svg" width="21" height="15.75" class="gem-c-subscription-links__icon" focusable="false" fill="currentColor" aria-hidden="true"><path d="M19.687 0H1.312C.589 0 0 .587 0 1.313v13.124c0 .726.588 1.313 1.313 1.313h18.374c.725 0 1.313-.587 1.313-1.313V1.313C21 .587 20.412 0 19.687 0zm-2.625 2.625L10.5 7.875l-6.563-5.25h13.126zm1.313 10.5H2.625V3.937L10.5 10.5l7.875-6.563v9.188z"/></svg>
```
36degrees: I think the `fill` needs to be on the `path` rather than the SVG.

andysellick: I tested it and it worked. Or is there some browser quirk I'm not aware of?

36degrees: So it does 👍Nice.

PR #1010, opened 2019-07-30
Update to govuk-frontend 3.0

alex-ju:
## What
Update `govuk_publishing_components` to use `govuk-frontend` 3.0.
Fix focus states on a few components to align them with the colour contrast requirements

## Why
- Keep frontend code in sync with the Design System updates
- Meet [WCAG 2.1 requirements regarding colour contrast](https://www.w3.org/TR/WCAG21/#contrast-minimum)

## Notes
- this PR is ~~raised for visibility and be used as a base for discussions~~ ready for review
- for more details regarding changes see [`govuk-frontend` 3.0 changelog](https://github.com/alphagov/govuk-frontend/releases/tag/v3.0.0)

## Compatibility checks
These changes were tested with applications in the following states using the appropriate compatibility flags.

- [x] applications with a direct dependency on `govuk_elements_rails`, `govuk_frontend_toolkit` and an indirect dependency on `govuk_template` (e.g. Service manual frontend)
- [x] applications with a direct dependency on `govuk_frontend_toolkit` and an indirect dependency on `govuk_template` (e.g. Frontend)
- [x] applications with an indirect dependency on `govuk_template` - dependent on static (e.g. [Calendars](https://github.com/alphagov/calendars/compare/update-to-govuk-frontend-3-0))
- [x] applications without dependencies on legacy projects (e.g. [Content Publisher](https://github.com/alphagov/content-publisher/compare/update-to-govuk-frontend-3-0))

## Visual Changes
There are visual changes for almost all components, especially related to [focus states](https://design-system.service.gov.uk/get-started/focus-states/) (which are harder to get with a basic visual comparison). I would recommend using the preview app when reviewing as images don't capture all the states.

## View Changes
https://govuk-publishing-compo-pr-1010.herokuapp.com/component-guide

[Trello card](https://trello.com/c/bYtf5EEe)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
@@ -4,7 +4,7 @@
   'use strict'
 
   if (document.querySelectorAll && document.addEventListener) {
-    var els = document.querySelectorAll('a.js-header-toggle')
+    var els = document.querySelectorAll('.js-header-toggle')
     var i
     var _i
     for (i = 0, _i = els.length; i < _i; i++) {
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -286,7 +286,7 @@ html {
 // Rouge syntax highlighting
 // Based on https://github.com/alphagov/tech-docs-template/blob/master/template/source/stylesheets/palette/_syntax-highlighting.scss
 
-$code-00: scale-color(govuk-colour("grey-4"), $lightness: 50%); // Default Background
+$code-00: scale-color(govuk-colour("light-grey"), $lightness: 50%); // Default Background
 $code-01: #f5f5f5; // Lighter Background (Unused)
 $code-02: #bfc1c3; // Selection Background
 $code-03: darken($govuk-secondary-text-colour, 2%); // Comments, Invisibles, Line Highlighting
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -7,8 +7,6 @@
 @import "govuk/helpers/all";
 @import "govuk/core/all";
 
-@include _govuk-font-face-gds-transport;
-
 .component-guide-wrapper {
   padding-bottom: $govuk-gutter * 1.5;
 }
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -7,7 +7,7 @@
 @import "govuk/helpers/all";
 @import "govuk/core/all";
 
-@include _govuk-font-face-nta;
+@include _govuk-font-face-gds-transport;
 
 .component-guide-wrapper {
   padding-bottom: $govuk-gutter * 1.5;
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -286,7 +286,7 @@ html {
 // Rouge syntax highlighting
 // Based on https://github.com/alphagov/tech-docs-template/blob/master/template/source/stylesheets/palette/_syntax-highlighting.scss
 
-$code-00: scale-color(govuk-colour("light-grey"), $lightness: 50%); // Default Background
+$code-00: scale-color(govuk-colour("light-grey", $legacy: "grey-4"), $lightness: 50%); // Default Background
 $code-01: #f5f5f5; // Lighter Background (Unused)
 $code-02: #bfc1c3; // Selection Background
 $code-03: darken($govuk-secondary-text-colour, 2%); // Comments, Invisibles, Line Highlighting
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -2,10 +2,10 @@
 
 @import "govuk_publishing_components/components/helpers/govuk-frontend-settings";
 @import "govuk_publishing_components/components/helpers/markdown-typography";
-@import "govuk-frontend/settings/all";
-@import "govuk-frontend/tools/all";
-@import "govuk-frontend/helpers/all";
-@import "govuk-frontend/core/all";
+@import "govuk/settings/all";
+@import "govuk/tools/all";
+@import "govuk/helpers/all";
+@import "govuk/core/all";
 
 @include _govuk-font-face-nta;
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -5,7 +5,7 @@ $thumbnail-background: govuk-colour("white");
 $thumbnail-border-colour: rgba(11, 12, 12, .1);
 $thumbnail-shadow-colour: rgba(11, 12, 12, .4);
 $thumbnail-shadow-width: 0 2px 2px;
-$thumbnail-icon-border-colour: govuk-colour("mid-grey");
+$thumbnail-icon-border-colour: govuk-colour("mid-grey", $legacy: "grey-3");
 
 .gem-c-attachment {
   @include govuk-font(19);
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -5,7 +5,7 @@ $thumbnail-background: govuk-colour("white");
 $thumbnail-border-colour: rgba(11, 12, 12, .1);
 $thumbnail-shadow-colour: rgba(11, 12, 12, .4);
 $thumbnail-shadow-width: 0 2px 2px;
-$thumbnail-icon-border-colour: govuk-colour("grey-3");
+$thumbnail-icon-border-colour: govuk-colour("mid-grey");
 
 .gem-c-attachment {
   @include govuk-font(19);
--- a/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
@@ -107,15 +107,3 @@ $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
     content: none;
   }
 }
-
-// Begin adjustments for font baseline offset
-// These should be removed when the font is updated with the correct baseline
-// For the 1px addition please see https://github.com/alphagov/govuk/pull/365#discussion_r154349428
-
-$offset: 2;
-
-.gem-c-button--secondary,
-.gem-c-button--secondary-quiet {
-  padding-top: (govuk-spacing(2) - $govuk-border-width-form-element + $offset); // s1
-  padding-bottom: (govuk-spacing(2) - $govuk-border-width-form-element - $offset + 1); // s1
-}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
@@ -3,9 +3,9 @@
 $gem-secondary-button-colour: #00823b;
 $gem-secondary-button-hover-colour: darken($gem-secondary-button-colour, 5%);
 $gem-secondary-button-background-colour: govuk-colour("white");
-$gem-secondary-button-hover-background-colour: govuk-colour("grey-4");
+$gem-secondary-button-hover-background-colour: govuk-colour("light-grey");
 
-$gem-quiet-button-colour: govuk-colour("grey-1");
+$gem-quiet-button-colour: govuk-colour("dark-grey");
 $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
 
 // Because govuk-frontend adds a responsive bottom margin by default for each component
--- a/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
@@ -3,9 +3,9 @@
 $gem-secondary-button-colour: #00823b;
 $gem-secondary-button-hover-colour: darken($gem-secondary-button-colour, 5%);
 $gem-secondary-button-background-colour: govuk-colour("white");
-$gem-secondary-button-hover-background-colour: govuk-colour("light-grey");
+$gem-secondary-button-hover-background-colour: govuk-colour("light-grey", $legacy: "grey-4");
 
-$gem-quiet-button-colour: govuk-colour("dark-grey");
+$gem-quiet-button-colour: govuk-colour("dark-grey", $legacy: "grey-1");
 $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
 
 // Because govuk-frontend adds a responsive bottom margin by default for each component
--- a/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
@@ -1,4 +1,4 @@
-@import "govuk-frontend/components/button/button";
+@import "govuk/components/button/button";
 
 $gem-secondary-button-colour: #00823b;
 $gem-secondary-button-hover-colour: darken($gem-secondary-button-colour, 5%);
@@ -110,7 +110,7 @@ $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
 
 // Begin adjustments for font baseline offset
 // These should be removed when the font is updated with the correct baseline
-// For the 1px addition please see https://github.com/alphagov/govuk-frontend/pull/365#discussion_r154349428
+// For the 1px addition please see https://github.com/alphagov/govuk/pull/365#discussion_r154349428
 
 $offset: 2;
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
@@ -5,7 +5,7 @@
   box-sizing: border-box;
   border-left-style: solid;
   border-left-width: 4px;
-  border-color: govuk-colour("grey-2");
+  border-color: govuk-colour("mid-grey");
   margin-top: govuk-spacing(2);
   margin-bottom: govuk-spacing(2);
   padding: govuk-spacing(2) govuk-spacing(4);
--- a/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
@@ -5,7 +5,7 @@
   box-sizing: border-box;
   border-left-style: solid;
   border-left-width: 4px;
-  border-color: govuk-colour("mid-grey");
+  border-color: govuk-colour("mid-grey", $legacy: "grey-2");
   margin-top: govuk-spacing(2);
   margin-bottom: govuk-spacing(2);
   padding: govuk-spacing(2) govuk-spacing(4);
--- a/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
@@ -1,4 +1,4 @@
-@import "govuk-frontend/components/checkboxes/checkboxes";
+@import "govuk/components/checkboxes/checkboxes";
 
 .govuk-checkboxes--nested {
   margin-left: -(govuk-spacing(4) + 2px); //22px
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -33,10 +33,6 @@
   }
 }
 
-.gem-c-feedback__grid-row {
-  @extend %govuk-grid-row;
-}
-
 .gem-c-feedback__column-two-thirds {
   @include govuk-grid-column(two-thirds)
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -33,10 +33,12 @@
   }
 }
 
-@include govuk-grid-row($class: "gem-c-feedback__grid-row");
+.gem-c-feedback__grid-row {
+  @extend %govuk-grid-row;
+}
 
 .gem-c-feedback__column-two-thirds {
-  @include govuk-grid-column(two-thirds, $class: false)
+  @include govuk-grid-column(two-thirds)
 }
 
 .gem-c-feedback__prompt {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -33,10 +33,6 @@
   }
 }
 
-.gem-c-feedback__column-two-thirds {
-  @include govuk-grid-column($width: two-thirds)
-}
-
 .gem-c-feedback__prompt {
   @include govuk-clearfix;
   background-color: govuk-colour("blue");
@@ -136,12 +132,6 @@
   display: block;
   padding: 4px 0 0;
   color: $govuk-error-colour;
-}
-
-.gem-c-feedback__survey-container {
-  @include govuk-media-query($until: tablet) {
-    padding: govuk-spacing(4);
-  }
 }
 
 .gem-c-feedback__form {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -105,7 +105,7 @@
 .gem-c-feedback__error-summary {
   margin-bottom: govuk-spacing(3);
   padding: govuk-spacing(3);
-  border: solid $govuk-border-width-mobile $govuk-error-colour;
+  border: solid $govuk-border-width-narrow $govuk-error-colour;
   clear: both;
 
   &:focus {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -34,7 +34,7 @@
 }
 
 .gem-c-feedback__column-two-thirds {
-  @include govuk-grid-column(two-thirds)
+  @include govuk-grid-column($width: two-thirds)
 }
 
 .gem-c-feedback__prompt {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
@@ -37,10 +37,10 @@
 
 .gem-c-highlight-boxes__item {
   box-sizing: border-box;
-  border: 1px solid govuk-colour("mid-grey");
+  border: 1px solid govuk-colour("mid-grey", $legacy: "grey-2");
   padding: govuk-spacing(3) * 1.5;
   height: 100%;
-  box-shadow: 7px 7px 0 govuk-colour("white"), 8px 8px 0 govuk-colour("mid-grey");
+  box-shadow: 7px 7px 0 govuk-colour("white"), 8px 8px 0 govuk-colour("mid-grey", $legacy: "grey-2");
 }
 
 .gem-c-highlight-boxes--inverse {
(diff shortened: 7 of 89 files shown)


app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
```diff
@@ -4,7 +4,7 @@
   'use strict'
 
   if (document.querySelectorAll && document.addEventListener) {
-    var els = document.querySelectorAll('a.js-header-toggle')
+    var els = document.querySelectorAll('.js-header-toggle')
     var i
     var _i
     for (i = 0, _i = els.length; i < _i; i++) {
```
injms: I'm curious and intrigued about this loop, since it seems to assigned `els.length` to `_i` within parenthesis, and I wonder what this does and why?

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -5,7 +5,7 @@ $thumbnail-background: govuk-colour("white");
 $thumbnail-border-colour: rgba(11, 12, 12, .1);
 $thumbnail-shadow-colour: rgba(11, 12, 12, .4);
 $thumbnail-shadow-width: 0 2px 2px;
-$thumbnail-icon-border-colour: govuk-colour("grey-3");
+$thumbnail-icon-border-colour: govuk-colour("mid-grey");
```
36degrees: If you want these components to work in downstream applications with compatibility mode enabled, you'll want to pass `$legacy: "grey-3"` here so that it knows which colour to use from the legacy palette, whenever you use a colour that's been renamed or added.

Without this, if you tried to use the attachment component in an app that has compatibility mode enabled, you'd get an 'Unknown colour mid-grey' error.

http://govuk-frontend-review.herokuapp.com/docs/#function-govuk-colour

You might also want to find a way to surface or test with legacy mode within publishing components, so that you don't find out things like this are broken after you've published!

app/assets/stylesheets/component_guide/application.scss
```diff
@@ -2,12 +2,12 @@
...
-@import "govuk-frontend/tools/all";
-@import "govuk-frontend/helpers/all";
-@import "govuk-frontend/core/all";
+@import "govuk/settings/all";
+@import "govuk/tools/all";
+@import "govuk/helpers/all";
+@import "govuk/core/all";
 
-@include _govuk-font-face-nta;
+@include _govuk-font-face-gds-transport;
```
36degrees: As an aside, the font-face should be included automatically the first time it's used – you shouldn't have to do this. If that's not working for you, it'd be good to know, as we consider this a private mixin at the minute.

app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
```diff
@@ -66,7 +66,7 @@
 
   &:active,
   &:focus {
-    color: $govuk-link-colour;
+    color: $govuk-text-colour;
```
36degrees: Unsure, because of the `:active` state, but should this be `$govuk-focus-text-colour`?

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -1,4 +1,4 @@
-@import "govuk-frontend/components/input/input";
+@import "govuk/components/input/input";
 
 .gem-c-input--search-icon {
   background: govuk-colour("white") url("data:image/svg+xml,%3Csvg xmlns='http://www.w3.org/2000/svg' viewBox='0 0 36 36' width='40' height='40'%3E%3Cpath d='M25.7 24.8L21.9 21c.7-1 1.1-2.2 1.1-3.5 0-3.6-2.9-6.5-6.5-6.5S10 13.9 10 17.5s2.9 6.5 6.5 6.5c1.6 0 3-.6 4.1-1.5l3.7 3.7 1.4-1.4zM12 17.5c0-2.5 2-4.5 4.5-4.5s4.5 2 4.5 4.5-2 4.5-4.5 4.5-4.5-2-4.5-4.5z' fill='currentColor'%3E%3C/path%3E%3C/svg%3E") no-repeat -5px -3px;
```
36degrees: Unrelated to this change, and super minor, but as far as I know `currentColor` doesn't work in background images – so the magnifying glass icon is `#000000`

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -33,10 +33,8 @@
   }
 }
 
-@include govuk-grid-row($class: "gem-c-feedback__grid-row");
-
 .gem-c-feedback__column-two-thirds {
-  @include govuk-grid-column(two-thirds, $class: false)
+  @include govuk-grid-column($width: two-thirds)
```
36degrees: Could the feedback component just be updated to use `govuk-grid-column-two-thirds`? It looks like the row class was changed to `govuk-grid-row` in the markup.

app/assets/stylesheets/component_guide/application.scss
alex-ju: This should've been removed after dropping the compatibility mode.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
alex-ju: Sounds good. Wasn't aware of the `$legacy` attribute - haven't seen it anywhere yet.

app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
alex-ju: Yep, they all (for all the fixed focus colours) should be `$govuk-focus-text-colour`

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
alex-ju: Good call for improvement!

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
alex-ju: Yes, wasn't constrained to do it for the row, but still got wiggle room for columns. I personally think it should use both classes from `govuk-frontend`.

app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
alex-ju: 🤷‍♂️haven't looked at it to be honest. Hope this will soon be removed as we have it covered by the header script from govuk-frontend.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
alex-ju: Adressed in 4c479d7

app/assets/stylesheets/component_guide/application.scss
alex-ju: Removed

app/assets/stylesheets/govuk_publishing_components/components/_highlight-boxes.scss
alex-ju: Checked focus states for all components and made use of `$govuk-focus-text-colour` where needed in ff411ec

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
alex-ju: Updated feedback component to use govuk-frontend classes for layout in 6ab1174

app/assets/stylesheets/govuk_publishing_components/components/_button.scss
```diff
@@ -110,7 +110,7 @@ $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
 
 // Begin adjustments for font baseline offset
 // These should be removed when the font is updated with the correct baseline
-// For the 1px addition please see https://github.com/alphagov/govuk-frontend/pull/365#discussion_r154349428
+// For the 1px addition please see https://github.com/alphagov/govuk/pull/365#discussion_r154349428
```
andysellick: Not sure this one should have been changed from govuk-frontend to govuk, possibly a victim of find and replace?

app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
```diff
@@ -27,7 +27,7 @@
 
 .gem-c-checkboxes {
   // this is complex but needed to override the govuk-frontend styles in
-  // https://github.com/alphagov/govuk-frontend/blob/b8058640b9602ecb6e1f66f887553190cbae7b46/src/components/hint/_hint.scss#L16
+  // https://github.com/alphagov/govuk/blob/b8058640b9602ecb6e1f66f887553190cbae7b46/src/components/hint/_hint.scss#L16
```
andysellick: Also not sure this one should have been changed from govuk-frontend to govuk.

config/initializers/assets.rb
```diff
@@ -34,7 +34,8 @@
 )
 
 Rails.application.config.assets.paths += %W(
-  #{__dir__}/../../node_modules/govuk-frontend/assets/images
-  #{__dir__}/../../node_modules/govuk-frontend/assets/fonts
+  #{__dir__}/../../node_modules/govuk-frontend/govuk/assets/images
+  #{__dir__}/../../node_modules/govuk-frontend/govuk/assets/fonts
+  #{__dir__}/../../node_modules/govuk-frontend/
```
andysellick: Question: if the whole of the `node_modules/govuk-frontend` dir is a path, why do we need the images and fonts subdirs explicitly specified?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -33,10 +33,12 @@
...
 }
 
-@include govuk-grid-row($class: "gem-c-feedback__grid-row");
+.gem-c-feedback__grid-row {
+  @extend %govuk-grid-row;
+}
 
 .gem-c-feedback__column-two-thirds {
-  @include govuk-grid-column(two-thirds, $class: false)
+  @include govuk-grid-column(two-thirds)
```
andysellick: Nitpick: no semicolon at the end of this line (although presumably it passed the linter?).

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -33,10 +33,12 @@
   }
 }
 
-@include govuk-grid-row($class: "gem-c-feedback__grid-row");
+.gem-c-feedback__grid-row {
+  @extend %govuk-grid-row;
```
andysellick: V3 changelog says `You can no longer use the govuk-grid-row mixin` - that's not this, is it? I haven't checked the review app (it's not working atm).

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -136,7 +138,10 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
 
   &:focus,
   &:hover {
+    @include govuk-focused-text;
+
     outline: none;
+    box-shadow: none;
```
andysellick: Probably outside of scope, but I think the close button looks pretty good with the box shadow. Certainly helps with the contrast between its focussed state and the white space below.

With:

Screen Shot 2019-08-05 at 17 23 17

Without:

Screen Shot 2019-08-05 at 17 23 09

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -30,14 +29,22 @@ $large-input-size: 50px;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
+
+  &:focus {
+    outline: $govuk-focus-width solid $govuk-focus-colour;
+    outline-offset: inherit;
```
andysellick: Probably out of scope, but the search box is long overdue an overhaul to make it look like GOVUK Frontend. We should talk about that.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -237,7 +237,9 @@ $top-border: solid 2px $grey-3;
   }
 
   .gem-c-step-nav__button {
-    @include govuk-focusable;
+    &:focus {
+      outline: $govuk-focus-width solid $govuk-focus-colour;
+    }
```
andysellick: Not sure what's happened here but the show/hide all button on step navs (which has this class) seems to have lost its focus state.

Screen Shot 2019-08-05 at 17 29 14

nickcolley: Perhaps one for a follow up pull request but that style will not pass non-text guidelines.

https://design-system.service.gov.uk/get-started/focus-states/

app/views/govuk_publishing_components/components/feedback/_survey_signup_form.html.erb
```diff
@@ -12,7 +12,7 @@
     aria-expanded="true"
     role="button">Close</a>
 
-  <div class="gem-c-feedback__grid-row gem-c-feedback__survey-container">
+  <div class="govuk-grid-row">
```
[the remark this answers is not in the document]
alex-ju: I think this was re-introduced at rebasing. The padding is unnecessary from my point of view too, except for the home page where the layout should provide this padding on mobile (this was the reason for introducing it recently if I recall well).

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -36,7 +36,7 @@
     </div><div class="govuk-header__content">
 
     <% if navigation_items.any? %>
-    <button role="button" class="govuk-header__menu-button js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
+    <button role="button" class="govuk-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
```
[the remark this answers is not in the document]
36degrees: This is a change from GOV.UK Frontend – we [introduced namespacing of JS classes](https://github.com/alphagov/govuk-frontend/pull/1444) to fix an [issue using the header component with GOV.UK Template](https://github.com/alphagov/govuk-frontend/issues/1270).

app/assets/stylesheets/govuk_publishing_components/components/helpers/_variables.scss
```diff
@@ -11,7 +11,7 @@ $gem-spacing-scale-7: 50px;
 $gem-spacing-scale-8: 60px;
 
 $gem-text-colour: $text-colour;
-$gem-secondary-text-colour: $grey-1;
+$gem-secondary-text-colour: govuk-colours("dark-grey");
```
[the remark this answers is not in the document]
36degrees: [Aside: This'll compile just fine because Sass will assume that `govuk-colours` is a native CSS construct (like e.g. `background-image`) – this sort of thing can easily slip in, so we've written [a test to try and catch it](https://github.com/alphagov/govuk-frontend/blob/08b408840deabc40cb9d2438b0d19862cda41f05/src/govuk/all.test.js#L116-L155)]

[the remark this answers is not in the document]
alex-ju: Ha, good found, and even better test, but it's a bit hard to replicate it in the world of Jasmine. The mixin failed silently and it had no visual effects since this variable is no longer used (was replaced by `$govuk-secondary-text-colour` all over the gem. This should be removed now.

andysellick:
Some more thoughts...

- the accessible autocomplete doesn't have the new focus styles (because it's mostly pulled in from another repo) but we're going to address that separately
- links that wrap have a potential overlap issue with the line below slightly cropping the line above, as seen in the attachment component (note how the bottom of the 'P' in 'department' is missing). Might be something to talk to the Design System team about (also we might need to adjust the line height in this component)

Screen Shot 2019-08-06 at 13 21 54

Here's an only semi-related but interesting example from the contents list component, where the second line is indented, leaving a trailing bit of the first line's underline visible (I thought it was a quote mark when I first spotted it)

Screen Shot 2019-08-06 at 13 48 53

andysellick:
Also fairly obvious but worth pointing out that apps will break with this update - I just tried it in finder-frontend and it failed because apparently we're using the `govuk-focusable` mixin in that app.

app/assets/stylesheets/govuk_publishing_components/components/_button.scss
alex-ju: The initial replace was 'erroneous', but the adjustments (and the comments) were removed in a following commit.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
alex-ju: This code was removed in a following commit.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
alex-ju: Yes, it's that one. This was removed in a following commit.

app/views/govuk_publishing_components/components/_textarea.html.erb
```diff
@@ -20,7 +20,7 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   css_classes = %w(govuk-textarea)
-  css_classes << "js-character-count" if character_count
+  css_classes << "govuk-js-character-count" if character_count
```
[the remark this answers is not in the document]
alex-ju: It's not unnecessary. We used to have a clash on `js-header-toggle` being picked up by 2 scripts. I needed to find a workaround for that in the past, using a more explicit selector. Now these can co-exist until we move to the new header component.

app/assets/stylesheets/govuk_publishing_components/components/_checkboxes.scss
alex-ju: Fixed, missed when replacing. Good spot.

config/initializers/assets.rb
alex-ju: I suspect this is because they're referenced without the full path (e.g. `my-font.otf` instead of `/govuk/assets/fonts/my-font.otf`)

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Good call for potential improvement

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
alex-ju: I agree about this needs an overhaul. Again, good call for an improvement.

andysellick:
Tables look good with the new, slightly lighter grey...

Screen Shot 2019-08-07 at 10 21 48

Old:

Screen Shot 2019-08-07 at 10 21 42

andysellick:
The new focus style takes up a bit more vertical space than it used to. The worst example of this so far that I've found is the image card (also noticeable with the textarea, with text beneath it).

Screen Shot 2019-08-07 at 11 06 00

I don't know if this is actually a problem though. You're most likely to notice this if you're using a keyboard, in which case moving the focus won't be a problem. Also the one you're currently focussed on is fine.

Might be worth looking at some of these issues after this PR is merged, just noting it here for reference.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
alex-ju: I did an update on the step-by-step-nav to improve focus states.

Screen Shot 2019-08-07 at 11 08 42

This component is also a good candidate for an overhaul looking at the use of `!important`

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Added a commit to improve the focus on the modal's close button
Screen Shot 2019-08-07 at 11 11 17

config/initializers/assets.rb
andysellick: I've tested this and those extra paths are definitely needed. Weird. Oh well.

app/assets/stylesheets/component_guide/application.scss
```diff
@@ -2,12 +2,10 @@
 
 @import "govuk_publishing_components/components/helpers/govuk-frontend-settings";
 @import "govuk_publishing_components/components/helpers/markdown-typography";
-@import "govuk-frontend/settings/all";
-@import "govuk-frontend/tools/all";
-@import "govuk-frontend/helpers/all";
-@import "govuk-frontend/core/all";
-
-@include _govuk-font-face-nta;
```
DilwoarH: does this no longer need importing?

andysellick:
This is unrelated but worth adding to the 'to check later' list... the layout header with links doesn't seem to collapse early enough, plus when the links do collapse the dropdown menu doesn't seem to work?

Screen Shot 2019-08-07 at 13 14 28

app/assets/stylesheets/component_guide/application.scss
alex-ju: Nope, that was a hack to use the new font in the component guide app as it wasn't loaded because of the compatibility flags. Now that we have them turned off the font comes by default and this is no longer needed.

andysellick:
Page title (with context link) lacks a focus state for the link. This was a problem prior to this PR, so arguably out of scope, but we should fix it.

Screen Shot 2019-08-07 at 13 27 10

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
nickcolley: Looking good, however there looks like some issues around hover + focus:

Screen Shot 2019-08-07 at 13 46 59

alex-ju: Yep, I'm thinking how to tackle that by untangling styles rather than chaining selectors.

vanitabarrett:
Sorry if I missed this in our review meeting of this earlier - it looks like some components have been updated to use the new focus style, but some haven't. What is the reason behind that? 

E.g:

Screen Shot 2019-08-07 at 14 22 17

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
```diff
@@ -26,7 +26,7 @@
 
   $chart-border: govuk-colour("white"); // Chart border colour
   $key-border: govuk-colour("white"); // Key border colour
-  $bar-colours: govuk-colour("blue"), govuk-colour("turquoise"), govuk-colour("green"), govuk-colour("light-green"), govuk-colour("yellow"), govuk-colour("orange"), govuk-colour("red"), govuk-colour("bright-red");
+  $bar-colours: govuk-colour("blue"), govuk-colour("turquoise"), govuk-colour("green"), govuk-colour("light-green"), govuk-colour("yellow"), govuk-colour("orange"), govuk-colour("red"), govuk-colour("red");
```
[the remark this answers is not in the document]
alex-ju: Yes, that's right. We could use purple, or any other [colour from the palette](https://design-system.service.gov.uk/styles/colour/#colour-palette). Open to suggestions 🎨

vanitabarrett:
Do we need to update the govspeak 'highlight answer' to use the green instead of the turquoise?
https://govuk-publishing-compo-pr-1010.herokuapp.com/component-guide/govspeak/highlight_answer/preview

Screen Shot 2019-08-07 at 14 50 02

andysellick:
The hover + focus state for next/previous navigation could do with a bit of attention...

Focus:

Screen Shot 2019-08-07 at 14 53 23

Focus + hover:

Screen Shot 2019-08-07 at 14 53 29

andysellick:
Not sure we can do much about this one and it predates this PR, but focussed share links with branding applied might not always pass contrast requirements. At some point we should check.

Screen Shot 2019-08-07 at 16 05 28

vanitabarrett:
@andysellick I did look through those as I was concerned about focus states, but I think they all pass colour contrast (AA)

andysellick:
Skip link hasn't picked up the new focus style, might be defining it itself?

Screen Shot 2019-08-07 at 16 08 45

andysellick:
Step by step navigation context text fails contrast requirements when focussed (probably did before this PR).

Screen Shot 2019-08-07 at 16 11 55

andysellick:
We might want to tweak the image card component a little... if you focus and hover some of the links, you get a double underline.

Screen Shot 2019-08-07 at 16 18 11

The component overrides the default `text-decoration: none` on hover + focus from govuk-frontend to add underline. We should remove it in this specific case i.e. leave underline on when hovering.

andysellick:
Again seems to be unrelated but the sort option for the table component seems to be non-functional - clicking it goes back to the guide homepage (pointing at /).

Screen Shot 2019-08-07 at 16 52 41

@alex-ju once I've finished going through the components I'll create issues for anything I've raised here that shouldn't be fixed as part of this PR.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
```diff
@@ -97,7 +97,7 @@ $govuk-modal-wide-breakpoint: $govuk-page-width + $govuk-modal-margin * 2 + $gov
 }
 
 .gem-c-modal-dialogue__header {
-  padding: 8px govuk-spacing(3) 10px;
+  padding: 9px govuk-spacing(3) 10px;
```
andysellick: I don't really mind, but why have an explicit `px` value alongside a `govuk-spacing` value?

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
andysellick: I think we should address this after V3 has been merged. Issue logged here: https://github.com/alphagov/govuk_publishing_components/issues/1027

andysellick:
@alex-ju I've logged issues for anything I've found that's unrelated to this PR and marked those comments accordingly. Some of them should be addressed after this PR is merged and some can be worked on right now - I'll take a look at some of those when I have a chance.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -1,15 +1,15 @@
...
 $table-header-border-width: 2px;
-$table-header-background-colour: govuk-colour("grey-3");
+$table-header-background-colour: govuk-colour("light-grey");
 $sort-link-active-colour: govuk-colour("white");
 $sort-link-arrow-size: 14px;
 $sort-link-arrow-size-small: 8px;
 $sort-link-arrow-spacing: $sort-link-arrow-size / 2;
 $table-row-hover-background-colour: rgba(43, 140, 196, .2);
-$table-row-even-background-colour: govuk-colour("grey-4");
+$table-row-even-background-colour: govuk-colour("light-grey");
```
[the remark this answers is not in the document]
andysellick: I've gone through all the components and haven't spotted any problems relating to this.

app/assets/stylesheets/govuk_publishing_components/components/_modal-dialogue.scss
alex-ju: Yeah, it's an odd mix isn't it. I can use pixels for all values for consistency. Tried to make sure the vertical rhythm is preserved, that's why I always use `govuk-spacing` for top and bottom.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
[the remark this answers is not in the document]
alex-ju: Updated to use `bright-purple` for now, the closest in shade with the previous one.

alex-ju:
@vanitabarrett Thanks for the review. Tried to address the comments. Do let me know if the review is still in progress.
> Do we need to update the govspeak 'highlight answer' to use the green instead of the turquoise?
> https://govuk-publishing-compo-pr-1010.herokuapp.com/component-guide/govspeak/highlight_answer/preview

Updated to use the new colour

@andysellick Thanks for the review. Tried to address all the comments (aside from the ones with an associated issue - 🙌 for that). For those comments without a thread I tried to reply below.

> The hover + focus state for next/previous navigation could do with a bit of attention...

Updated to preserve the focus state when hovering

> Skip link hasn't picked up the new focus style, might be defining it itself?

This is how the component is coming from `govuk-frontend`. We can propose enhancements upstream or do overwrites if we have good reasons for that.

> We might want to tweak the image card component a little... if you focus and hover some of the links, you get a double underline.

Fixed the focus/hover states on image card component

app/views/govuk_publishing_components/components/feedback/_survey_signup_form.html.erb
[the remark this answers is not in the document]
alex-ju: Removed the unused class.

alex-ju:
Rebased again before merging. Thanks everyone who commented on this, that helped a lot!

PR #1012, opened 2019-07-31
Add full width option to select component

andysellick:
## What
Adds a full width option to the select component, which sets the select element itself to be `width: 100%;`.

Screen Shot 2019-07-31 at 08 23 47

## Why
This isn't included in the Design System for reasons but we're using this component in finder-frontend's search and adding CSS there to make it full width, so it seemed more in line with our component principles to add it into the component itself.

Screen Shot 2019-07-31 at 08 25 11

## Visual Changes
No visual changes unless you select this option, obviously.

## View Changes
https://govuk-publishing-compon-pr-1012.herokuapp.com/select

Trello card: https://trello.com/c/w34QlutZ/910-investigate-adding-an-option-to-the-select-component-for-width-100

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,12 +7,16 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## 17.9.1
+## Unreleased
+
+* Add full width option to select component (PR #1012)
+
+## 17.19.1
 
 * Replace subscription links images with SVG (PR #1008)
 * Change subscription links CSS (PR #1007)
 
-## 17.9.0
+## 17.19.0
 
 * Remove expectation that sprockets is installed when used in a Rails app (PR #999)
 * Fix tabs list overwrite on mobile (PR #1003)
--- a/app/views/govuk_publishing_components/components/_select.html.erb
+++ b/app/views/govuk_publishing_components/components/_select.html.erb
@@ -2,6 +2,7 @@
   options ||= []
   id ||= false
   label ||= false
+  full_width ||= false
 
   select_helper = GovukPublishingComponents::Presenters::SelectHelper.new(options)
   data_module = "data-module=track-select-change" unless select_helper.data_tracking?.eql?(false)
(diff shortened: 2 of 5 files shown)


andysellick:
@alex-ju I asked the Design System team a while ago why there wasn't an option for this in govuk-frontend, and they gave me a good reason for it, which I've since forgotten. I think one possible reason might be we don't want selects to be enormous compared to their content, as it can look a bit weird, particularly compared to every other form element, which isn't full width.

andysellick:
@alex-ju I had `--full-width` originally! And then I noticed that I'd done it as `gem-c-select--full-width` but technically since it was a sub element of the component it needed to be `gem-c-select__select--full-width` and I thought it was just getting a bit ridiculously verbose at that point. But I can change it.

andysellick:
@alex-ju oh I see, yes, no, we don't want to do width: 100% by default, definitely not 😁

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add full width option to select component (PR #1012)
+
 ## 17.9.1
```
vanitabarrett: Not your change, but I just noticed we've gone from version 17.18.x to 17.9.x ... I think we need to change the Changelog to be 17.19.x (as the version.rb file itself is correct)

app/views/govuk_publishing_components/components/_select.html.erb
```diff
@@ -11,7 +12,7 @@
     <label class="govuk-label" for="<%= id %>">
       <%= label %>
     </label>
-    <select class="govuk-select" id="<%= id %>" name="<%= id %>" <%= data_module %> >
+    <select class="govuk-select <%= 'gem-c-select__select--full-width' if full_width %>" id="<%= id %>" name="<%= id %>" <%= data_module %> >
```
injms: Bikeshedding, so truly actually just a suggestion. 

The components seem inconsistent with each other with how classes are built up - but the joy is that it all works. 

Can we set the class as a variable, the push the `gem-c-select__select--full-width` class with an if? Any future extra classes can then be added on by `<<` on a new line.

```ruby
class = "govuk-select"
class << "gem-c-select__select--full-width" if full_width
```

andysellick: I tend to use that technique if there's more than two classes, because it gets unreadable at that point. But I think for two it's probably okay.

CHANGELOG.md
andysellick: Good spot, have corrected.

Looks like it was actually my change, though. Sorry...

PR #1013, opened 2019-07-31
Add visual regression testing tool Percy

injms:
## What
Visual regression testing compares screenshots, highlighting any changes made. Taking the screenshots is reasonably easily, but the storage and approval of any changes is more difficult. Enter Percy - a useful tool to do both store the screenshots and approve them. 

Percy is run in a GitHub Action which is triggered when a pull request is marked as **ready for review**. This is to cut down on the number of screenshots taken - a single round of visual regression tests takes about 300 screenshots. Triggering this on every push would likely use up all of our allowance.

Percy has been set up to only use one browser, but due to a known bug with Percy this won't happen until Percy can compare against the main branch.

## Why
With the number of components contained in this repo, it can be useful to run a visual regression test on every component to easily flag any changes that could have bled over.

## Known problems

 - ~Transition countdown changes each day as the deadline approached, causing a visual difference to be flagged~
 - ~Contextual sidebar contains the transition countdown component, causing a visual difference to be flagged~
 - ~Image card has random images, so can trigger Percy to flag a change (fixed in https://github.com/alphagov/govuk_publishing_components/pull/1795)~
 - Sometimes organisation logos don't appear as they time out, causing a a visual difference to be flagged 🤷‍♀️ I can't reliably replicate this problem, but I'll raise it as an issue for future reference.
 - ~Only runs when a draft pull request is marked 'ready for review'.~

## Visual Changes
No visual changes.

[Trello card](https://trello.com/c/swt30fXi)

### DIFF
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -1,0 +1,35 @@
+on:
+  push:
+    branches:
+      - main
+      - master
+  pull_request:
+    types:
+      - ready_for_review
+
+jobs:
+  visual-regression-test-components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+        with:
+          bundler-cache: true # Also runs `bundle install`
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12"
+      - name: Get Yarn cache directory path
+        id: yarn-cache-directory-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Check for cached Yarn dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
+          key: yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: yarn-
+      - run: yarn install --frozen-lockfile
+      - uses: percy/exec-action@v0.3.1
+        with:
+          command: bundle exec rspec ./spec/visual_regression_tests/all_components.rb
+        env:
+          PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -4,8 +4,6 @@ on:
       - main
       - master
   pull_request:
-    types:
-      - ready_for_review
 
 jobs:
   visual-regression-test-components:
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -30,6 +30,6 @@ jobs:
       - run: yarn install --frozen-lockfile
       - uses: percy/exec-action@v0.3.1
         with:
-          command: bundle exec rspec ./spec/visual_regression_tests/all_components.rb
+          command: bundle exec rspec --tag visual_regression
         env:
           PERCY_TOKEN: ${{ secrets.PERCY_TOKEN }}
--- a/spec/visual_regression_tests/all_components.rb
+++ b/spec/visual_regression_tests/all_components.rb
@@ -1,0 +1,41 @@
+require "rails_helper"
+require "percy"
+require "uri"
+
+describe "visual regression test runner Percy", type: :feature, js: true do
+  it "takes a screenshot of each component" do
+    # Freeze time for transition countdown
+    travel_to(Time.zone.local(2020, 12, 1, 6, 0, 0))
+
+    visit("/component-guide")
+
+    component_links = find("#list-all-components-in-the-gem")
+      .all("a")
+      .map do |link|
+        URI(link[:href])
+      end
+
+    component_links.each do |link|
+      skip_as_govspeak_times_out = [
+        "/component-guide/govspeak",
+      ]
+
+      skip_non_visible_components = [
+        "/component-guide/meta_tags",
+        "/component-guide/machine_readable_metadata",
+        "/component-guide/admin_analytics",
+        "/component-guide/google_tag_manager_script",
+        "/component-guide/layout_for_admin",
+      ]
+
+      all_components_to_skip = skip_as_govspeak_times_out + skip_non_visible_components
+
+      next if all_components_to_skip.include?(link.path)
+
+      visit("#{link}/preview")
+      name = title.gsub(/(: Default|) preview - Component Guide/, "")
+      page.find(:css, "#wrapper", wait: 10)
+      Percy.snapshot(page, name: name)
+    end
+  end
+end
--- a/spec/visual_regression_tests/all_components_spec.rb
+++ b/spec/visual_regression_tests/all_components_spec.rb
@@ -2,7 +2,7 @@ require "rails_helper"
 require "percy"
 require "uri"
 
-describe "visual regression test runner Percy", type: :feature, js: true do
+describe "visual regression test runner Percy", visual_regression: true do
   it "takes a screenshot of each component" do
     # Freeze time for transition countdown
     travel_to(Time.zone.local(2020, 12, 1, 6, 0, 0))
--- a/spec/visual_regression_tests/all_components_spec.rb
+++ b/spec/visual_regression_tests/all_components_spec.rb
@@ -4,38 +4,47 @@ require "uri"
 
 describe "visual regression test runner Percy", visual_regression: true do
   it "takes a screenshot of each component" do
-    # Freeze time for transition countdown
-    travel_to(Time.zone.local(2020, 12, 1, 6, 0, 0))
+    # Freeze time for consistency
+    travel_to Time.zone.local(2020, 12, 1, 6, 0, 0)
+
+    skip_as_govspeak_times_out = %w[
+      /component-guide/govspeak/preview
+    ]
+
+    skip_non_visible_components = %w[
+      /component-guide/meta_tags/preview
+      /component-guide/machine_readable_metadata/preview
+      /component-guide/admin_analytics/preview
+      /component-guide/google_tag_manager_script/preview
+      /component-guide/layout_for_admin/preview
+    ]
+
+    all_components_to_skip = skip_as_govspeak_times_out + skip_non_visible_components
 
     visit("/component-guide")
 
     component_links = find("#list-all-components-in-the-gem")
       .all("a")
       .map do |link|
-        URI(link[:href])
+        URI("#{link[:href]}/preview")
       end
 
+    # The public layout component has two previews - one with multiple versions
+    # as part of the component guide, and another one at `/public`.
+    component_links << URI("/public")
+
     component_links.each do |link|
-      skip_as_govspeak_times_out = [
-        "/component-guide/govspeak",
-      ]
-
-      skip_non_visible_components = [
-        "/component-guide/meta_tags",
-        "/component-guide/machine_readable_metadata",
-        "/component-guide/admin_analytics",
-        "/component-guide/google_tag_manager_script",
-        "/component-guide/layout_for_admin",
-      ]
-
-      all_components_to_skip = skip_as_govspeak_times_out + skip_non_visible_components
-
       next if all_components_to_skip.include?(link.path)
 
-      visit("#{link}/preview")
+      visit(link)
+
       name = title.gsub(/(: Default|) preview - Component Guide/, "")
-      page.find(:css, "#wrapper", wait: 10)
+
+      page.find(:css, "body > #wrapper", wait: 10)
+
       Percy.snapshot(page, name: name)
     end
+
+    travel_back
   end
 end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,11 @@
 
 ## Unreleased
 * Tidy up untranslated content and formatting on accordion docs ([PR #1958](https://github.com/alphagov/govuk_publishing_components/pull/1958)) PATCH
+* Add visual regression testing tool Percy ([PR #1013](https://github.com/alphagov/govuk_publishing_components/pull/1013)) PATCH
 
 ## 24.9.1
 
 * Change to the exceptions list for Brexit branding ([PR #2011](https://github.com/alphagov/govuk_publishing_components/pull/2011))
-
 
 ## 24.9.0
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -202,6 +202,7 @@ GEM
     parallel (1.20.1)
     parser (3.0.0.0)
       ast (~> 2.4.1)
+    percy-capybara (4.3.3)
     phantomjs (2.1.1.0)
     plek (4.0.0)
     pry (0.13.1)
@@ -375,6 +376,7 @@ DEPENDENCIES
   govuk_test (~> 2)
   jasmine (~> 3.7.0)
   jasmine_selenium_runner (~> 3.0.0)
+  percy-capybara (~> 4.0, >= 4.0.2)
   pry-byebug
   rake
   rspec-rails (~> 5.0)
--- a/app/views/govuk_publishing_components/component_guide/index.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/index.html.erb
@@ -79,7 +79,7 @@
 
 <% unless ENV["MAIN_COMPONENT_GUIDE"] %>
   <h2 class="component-doc-h2">Components in this application (<%= @component_docs.length %>)</h2>
-  <ul class="component-list">
+  <ul class="component-list" id="list-components-in-this-application">
     <% @component_docs.each do |component_doc| %>
       <li>
         <%= link_to component_doc.name, component_doc_path(component_doc.id), class: "govuk-link" %>
@@ -93,7 +93,7 @@
   <h2 class="component-doc-h2">All gem components (<%= @gem_component_docs.length %>)</h2>
 <% end %>
 
-<ul class="component-list">
+<ul class="component-list" id="list-all-components-in-the-gem">
   <% @gem_component_docs.each do |component_doc| %>
     <li>
       <%= link_to component_doc.name, component_doc_path(component_doc.id), class: "govuk-link" %>
--- a/govuk_publishing_components.gemspec
+++ b/govuk_publishing_components.gemspec
@@ -30,6 +30,7 @@ Gem::Specification.new do |s|
   s.add_development_dependency "govuk_test", "~> 2"
   s.add_development_dependency "jasmine", "~> 3.7.0"
   s.add_development_dependency "jasmine_selenium_runner", "~> 3.0.0"
+  s.add_development_dependency "percy-capybara", "~> 4.0", ">= 4.0.2"
   s.add_development_dependency "pry-byebug"
   s.add_development_dependency "rake"
   s.add_development_dependency "rspec-rails", "~> 5.0"
--- a/package.json
+++ b/package.json
@@ -32,6 +32,7 @@
   "devDependencies": {
     "standardx": "^7.0.0",
     "stylelint": "^13.8.0",
-    "stylelint-config-gds": "^0.1.0"
+    "stylelint-config-gds": "^0.1.0",
+    "@percy/agent": "^0.9.0"
   }
 }
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -10,7 +10,7 @@ WebMock.disable_net_connect!(allow_localhost: true)
 RSpec.configure do |config|
   config.example_status_persistence_file_path = "tmp/failures.txt"
 
-  config.filter_run_excluding not_applicable: true
+  config.filter_run_excluding not_applicable: true, visual_regression: true
 
   config.expect_with :rspec do |expectations|
     # This option will default to `true` in RSpec 4. It makes the `description`
@@ -36,4 +36,9 @@ RSpec.configure do |config|
   # inherited by the metadata hash of host groups and examples, rather than
   # triggering implicit auto-inclusion in groups with matching metadata.
   config.shared_context_metadata_behavior = :apply_to_host_groups
+
+  config.before do |example|
+    # Visual regression tests need the JS driver to be used for screenshots
+    Capybara.current_driver = Capybara.javascript_driver if example.metadata[:visual_regression]
+  end
 end
--- a/yarn.lock
+++ b/yarn.lock
@@ -2,185 +2,204 @@
 # yarn lockfile v1
 
 
-"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.10.4":
-  version "7.10.4"
-  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.10.4.tgz#168da1a36e90da68ae8d49c0f1b48c7c6249213a"
-  integrity sha512-vG6SvB6oYEhvgisZNFRmRCUkLz11c7rp+tbNTynGqc6mS1d5ATd/sGyV6W0KZZnXRKMTzZDRgQT3Ou9jhpAfUg==
+"@babel/code-frame@^7.0.0", "@babel/code-frame@^7.12.13":
+  version "7.12.13"
+  resolved "https://registry.yarnpkg.com/@babel/code-frame/-/code-frame-7.12.13.tgz#dcfc826beef65e75c50e21d3837d7d95798dd658"
+  integrity sha512-HV1Cm0Q3ZrpCR93tkWOYiuYIgLxZXZFVG2VgK+MBWjUqZTundupbfx2aXarXuw5Ko5aMcjtJgbSs4vUGBS5v6g==
   dependencies:
-    "@babel/highlight" "^7.10.4"
+    "@babel/highlight" "^7.12.13"
+
+"@babel/compat-data@^7.13.8":
+  version "7.13.12"
+  resolved "https://registry.yarnpkg.com/@babel/compat-data/-/compat-data-7.13.12.tgz#a8a5ccac19c200f9dd49624cac6e19d7be1236a1"
+  integrity sha512-3eJJ841uKxeV8dcN/2yGEUy+RfgQspPEgQat85umsE1rotuquQ2AbIub4S6j7c50a2d+4myc+zSlnXeIHrOnhQ==
 
 "@babel/core@>=7.9.0":
-  version "7.12.3"
-  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.12.3.tgz#1b436884e1e3bff6fb1328dc02b208759de92ad8"
-  integrity sha512-0qXcZYKZp3/6N2jKYVxZv0aNCsxTSVCiK72DTiTYZAu7sjg73W0/aynWjMbiGd87EQL4WyA8reiJVh92AVla9g==
+  version "7.13.10"
+  resolved "https://registry.yarnpkg.com/@babel/core/-/core-7.13.10.tgz#07de050bbd8193fcd8a3c27918c0890613a94559"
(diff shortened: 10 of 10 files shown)


kevindew:
Awesome stuff Ian, this looks super cool. I've only looked briefly at the code and had a few questions rather than a review at this point.

What is Percy comparing against? I can't see at what point Percy knows that something is good. Looking at https://percy.io/55ba5e1a/govuk_publishing_components/builds/8039557 seems to show a comparison between current and earlier versions of your branch. How do set the baseline?

I'm guessing we can't run this on Jenkins because of the requirement of it being a PR and that PR being ready for review. However I was wondering about when we hopefully migrate this repo to Github Actions. Would this be a task that we could configure as part of a wider CI action and have the Percy steps conditional? It feels like most of the action file would be the same pre-requisities for doing the rest of CI so would be a shame to duplicate.

Btw, there are developer docs on conventions in GitHub actions that could help make the actions file here more consistent: https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#govuk-rails-application-with-postgres-redis-yarn-and-govuk-content-schemas-dependencies - there's also been a long running issue with using [secrets in GitHub actions](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#header) which I think is just about resolved. May just need to give GOV.UK Senior Tech a nudge to say it's happening.

There may also need to be a tweak in [govuk-saas-config](https://github.com/alphagov/govuk-saas-config/tree/master/github) to make the percy check a required check. I'm not sure whether the sass config script would override settings in the repo.

injms:
Thanks for taking a look @kevindew - really appreciate it 👍 

> What is Percy comparing against?

Percy compares against a chosen branch - for this repo it's the master branch, but it could be any branch set in the Percy's admin settings. If Percy can't do that it'll compare against last set of screenshots from the current branch. 

So Percy will take screenshots of the master branch once this has been merged in; any subsequent pull requests will be compared against the master branch's screenshots.

> Would this be a task that we could configure as part of a wider CI action and have the Percy steps conditional? 

I couldn't find a way of saying "run all test on pushes and pull requests, but only run this step if marked as ready", which is why I didn't put it in a workflow named `ci.yml`. If it can be done then super, but I don't think it should be a blocker to getting this merged in.

> Btw, there are developer docs on conventions in GitHub actions that could help make the actions file here more consistent...

Brill - thanks for sharing. I'll update the action to follow these conventions.

> ...there's also been a long running issue with using [secrets in GitHub actions](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#header) which I think is just about resolved. May just need to give GOV.UK Senior Tech a nudge to say it's happening.

My understanding is that this was about using secrets to deploy sites, not to run tests; and that [this issue has been resolved](https://trello.com/c/BrOSVOpO). I'll ask a Senior Tech to take a look to be on the safe side.

> There may also need to be a tweak in [govuk-saas-config](https://github.com/alphagov/govuk-saas-config/tree/master/github) to make the percy check a required check. I'm not sure whether the sass config script would override settings in the repo.

I think that this shouldn't be a required check quite yet - people that need to approve should make sure they have a Browserstack account (Browserstack are Percy's parent company) so they have the correct permissions. Once we're sure of that people can access this tool, then this can be changed to a required check.

chao-xian:
Just one comment - some docs around this would be great, for example to include the stuff about needing a Browserstack account and permissions.

kevindew:
> Percy compares against a chosen branch - for this repo it's the master branch, but it could be any branch set in the Percy's admin settings. If Percy can't do that it'll compare against last set of screenshots from the current branch.
> 
> So Percy will take screenshots of the master branch once this has been merged in; any subsequent pull requests will be compared against the master branch's screenshots.

Ah neat, is this something that is configured with a Percy UI?

> > Would this be a task that we could configure as part of a wider CI action and have the Percy steps conditional?
> 
> I couldn't find a way of saying "run all test on pushes and pull requests, but only run this step if marked as ready", which is why I didn't put it in a workflow named `ci.yml`. If it can be done then super, but I don't think it should be a blocker to getting this merged in.

Cool - hopefully we won't be too far away from switching this project to a GitHub Action and can try integrate it.

> My understanding is that this was about using secrets to deploy sites, not to run tests; and that [this issue has been resolved](https://trello.com/c/BrOSVOpO). I'll ask a Senior Tech to take a look to be on the safe side.

The restriction was unfortunately any usage of secrets at all. But yes, you're right that this should be resolved. I raised it as this will be our first sanctioned usage of a GitHub Action secret. 

> I think that this shouldn't be a required check quite yet - people that need to approve should make sure they have a Browserstack account (Browserstack are Percy's parent company) so they have the correct permissions. Once we're sure of that people can access this tool, then this can be changed to a required check.

Ah sure - so I assume everyone can see the results of a Percy build but only authenticated users have the ability to approve? I've got a Browserstack account myself but no idea how widespread they are.

richardTowers (approved):
👋  weighing in just to make it clear that this is An Officially Sanctioned Use of GitHub Actions With Secrets (TM).

In this case, the secret in question isn't sensitive from an IA perspective (unlike deploy keys, or tokens that give write access to artefact stores), so we don't need to think about it in too much detail.

We still need to think about the tradeoffs we're making with GitHub Actions in the more general case, so this shouldn't be seen as a precedent implying "we can put whatever secrets we like in GHA without thinking about the consequences of doing so". For now, we should think about it on a case by case basis.

injms:
> > Percy compares against a chosen branch - for this repo it's the master branch, but it could be any branch set in the Percy's admin settings. If Percy can't do that it'll compare against last set of screenshots from the current branch.
> > So Percy will take screenshots of the master branch once this has been merged in; any subsequent pull requests will be compared against the master branch's screenshots.
> 
> Ah neat, is this something that is configured with a Percy UI?

Yes, it's set with the Percy admin UI - so when the master to main change happens it'll need to be updated.

> > My understanding is that this was about using secrets to deploy sites, not to run tests; and that [this issue has been resolved](https://trello.com/c/BrOSVOpO). I'll ask a Senior Tech to take a look to be on the safe side.
> 
> The restriction was unfortunately any usage of secrets at all. But yes, you're right that this should be resolved. I raised it as this will be our first sanctioned usage of a GitHub Action secret.

Ooh, didn't realise that this was the first. I asked @richardTowers to take a look to see if this is an acceptable use of a secret.

> > I think that this shouldn't be a required check quite yet - people that need to approve should make sure they have a Browserstack account (Browserstack are Percy's parent company) so they have the correct permissions. Once we're sure of that people can access this tool, then this can be changed to a required check.
> 
> Ah sure - so I assume everyone can see the results of a Percy build but only authenticated users have the ability to approve? I've got a Browserstack account myself but no idea how widespread they are.

I'm going to add some documentation to make this clearer  (thanks for the suggestion @chao-xian :) )

> Oh btw thanks for updating the actions file it looks great and I have a few minor comments that I'll add later (got to dash somewhere now). However I did notice the action didn't seem to run when you pushed the change so wondered if there might be a problem to look into?

D'oh - should be fixed now; the setup-node was missing a `with` for the version number.

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  visual-regression-test-components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "14"
```
kevindew, line 18: Do we need to specify a node version? I feel like we'd be best placed leaving it blank unless this project had either a .nvmrc file or specified a node version in package.json.

14 doesn't seem the right choice to specify as Jenkins is running node 12.

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Check for cached Ruby gems
+        uses: actions/cache@v2
+        with:
+          path: vendor/bundle
+          key: ${{ runner.os }}-gems-${{ hashFiles('**/Gemfile.lock') }}
+          restore-keys: ${{ runner.os }}-gems-
```
kevindew, line 24:
```suggestion
          path: vendor/bundle
          key: bundle-${{ hashFiles('**/Gemfile.lock') }}
          restore-keys: bundle
```

This suggestion brings it in line with https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#simple-ruby-application - we don't gain anything by specifing the os given we're only building on one.

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+      - name: Get Yarn cache directory path
+        id: yarn-cache-directory-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Check for cached Yarn dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-yarn-
+      - run: bundle install --jobs 4 --retry 3 --path vendor/bundle
```
kevindew, line 34: Any reason why you've gone for `--path vendor/bundle` over `--deployment`? the latter installs to vendor/bundle and also checks the Gemfile is up to date so is slightly preferable.

As I understand both warn as being deprecated as bundle now wants you to specify them in separate commands (which is a bit annoying - we'll need to update the docs to specify the newer approach)

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+          restore-keys: ${{ runner.os }}-gems-
+      - name: Get Yarn cache directory path
+        id: yarn-cache-directory-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Check for cached Yarn dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: ${{ runner.os }}-yarn-
```
kevindew, line 33: Ah you've taken to caching the yarn cache over the node_modules directory? Does this offer advantages?  Some googling tells me [it's not recommended](https://stackoverflow.com/a/62244232) to cache node_modules - but I don't understand why? whereas another doc seems to suggest it was better : https://dev.to/mpocock1/how-to-cache-nodemodules-in-github-actions-with-yarn-24eh.

I don't mind which approach we take but I would want want to make sure [our docs](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#govuk-rails-application-with-postgres-redis-yarn-and-govuk-content-schemas-dependencies) match whichever is preferred.

Again we shouldn't need the os specified here. 

It also reminds me I need to set-up a shared yarn cache for govuk-docker.

spec/visual_regression_tests/all_components.rb
```diff
@@ -0,0 +1,41 @@
+require "rails_helper"
```
kevindew, line 1: As this file is a spec it should be suffixed with `_spec.rb`

spec/visual_regression_tests/all_components.rb
```diff
@@ -0,0 +1,41 @@
...
+require "percy"
+require "uri"
+
+describe "visual regression test runner Percy", type: :feature, js: true do
```
kevindew, line 5: Does this need to be part of Rspec? It doesn't really feel like this code is doing a test as there are no assertions. I was wondering if the use of RSpec was just due to ease of Capybara availability? If we could easily spin up the server and capybara in a rake task would that be preferable? (I don't mind helping on that)

kevindew:
> 👋 weighing in just to make it clear that this is An Officially Sanctioned Use of GitHub Actions With Secrets (TM).
> 
> In this case, the secret in question isn't sensitive from an IA perspective (unlike deploy keys, or tokens that give write access to artefact stores), so we don't need to think about it in too much detail.
> 
> We still need to think about the tradeoffs we're making with GitHub Actions in the more general case, so this shouldn't be seen as a precedent implying "we can put whatever secrets we like in GHA without thinking about the consequences of doing so". For now, we should think about it on a case by case basis.

Great thanks @richardTowers, do you want to update the [docs](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html) with the updated stance.

What about the use of a secret to release a gem (which is the missing piece that'll keep this project on both Jenkins and GitHub actions) - is the approach in [rubocop-govuk](https://github.com/alphagov/rubocop-govuk/blob/706bc29f3721f0b2295f6c344141e83ca741b9d1/.github/workflows/default.yml) sanctioned or not? And if still ambiguous, how can we move past the stalemate?

> D'oh - should be fixed now; the setup-node was missing a `with` for the version number.
> 
> Hmm, it didn't run. I suspect this action only runs when the `ready_for_review` event is fired, not on subsequent pushes after that.

Oh, maybe just have it run on PR open? I guess it's a minority of PR's that are opened as draft.

.github/workflows/visual-regression-tests.yml
injms: `actions/setup-node` needs a version given to it as there's no default mentioned in the container's [action.yml config](https://github.com/actions/setup-node/blob/27082cecf3ff7a1742dbd5e12605f0cb59dce2d9/action.yml#L8-L9). 

(The `actions/setup-ruby` container that GitHub provides also needs an explicit version, which is why this uses `ruby/setup-ruby` which reads the  version from `.ruby-version`.)

I'll change it to 12 to match Jenkins.

.github/workflows/visual-regression-tests.yml
injms: No reason - I took it from the [Ruby gem example](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#a-ruby-gem) in the docs. I'll update it to deployment.

kevindew: Oh yes, sorry. Most gems can't use `--deployment` as they don't have their Gemfile.lock in the repo. This app is somewhat special in it's half gem half app state.

spec/visual_regression_tests/all_components.rb
injms: Is there a way of excluding this from the generic 'run everything that ends in `_spec.rb`' command? Otherwise this will take up developers time running the test for no end result.

spec/visual_regression_tests/all_components.rb
injms: I haven't got any strong opinion on what triggers the visual regression test to run - a rake task is a good idea, and the only thing is that it shouldn't be run as part of the _default_ rake task.

I don't think that this should be a blocker. I'd like to get this in front of developers to find out whether this is actually useful or not before spending more time on it - and running it as a GitHub Action makes this an easy change to make at a later date.

If this this workflow turns out to be useful then I'm happy to spend time on making it run in a better, more correct way.

.github/workflows/visual-regression-tests.yml
injms: Ah, I meant to remove this - thank you. I initially added this OS differentiator as I thought that might give us the ability to take screenshots on Windows, Linux, and macOS - but then discovered that's not how Percy works.

.github/workflows/visual-regression-tests.yml
injms: My understanding is that yarn first tries to install dependencies from the cache folder to the node_modules folder. If the dependency is not cached it's fetched from somewhere on the internet, then cached and installed.

So if the node_modules folder is cached it results in the node_modules folder being overwritten when `yarn install` is run. But if yarn's cache folder is cached, then the that's used to install the dependencies into the node_modules folder - hence this set up.

I can update the documentation with this method if that'd be useful?

spec/visual_regression_tests/all_components.rb
kevindew: Ah I see, hmm there's a flag approach that can be used to tell it to not run or not ([relevant stack overflow](https://stackoverflow.com/questions/41288405/not-run-a-spec-by-default-unless-flag-is-set))

I'd suggest this may be greater evidence that this should be outside specs since this looks like a spec, runs in the Rspec framework but then we don't want to treat it as a spec.

spec/visual_regression_tests/all_components.rb
kevindew: Yup agree on the default rake task. We define those in the [Rakefile](https://github.com/alphagov/govuk_publishing_components/blob/dcda1a5ad918b436c454ff4c2837e77fea7cc58c/Rakefile#L360) so we shouldn't be at risk.

I can see your concern about this being a blocker. There are lots of ways to consider it. It feels like the ideal approach would be for this to run implicitly as part of the unit tests (maybe part of some sort of render assertion) which could be switched on and off with an env var. But a few things to think about.

I did have a tinker around to see if I could work out how this could be disambuated from RSpec and came up with putting it in a `bin` script. See what you think as to whether this seems a viable way to avoid this being tangled with tests: https://github.com/alphagov/govuk_publishing_components/commit/87bc88cd3308ecf1a1fc6bccdf2ecd75803e7469

.github/workflows/visual-regression-tests.yml
kevindew: From a quick play it looks like it checks the node_modules directory first. For example:

```
➜  govuk_publishing_components git:(master) yarn
yarn install v1.22.10
[1/4] 🔍  Resolving packages...
[2/4] 🚚  Fetching packages...
[3/4] 🔗  Linking dependencies...
[4/4] 🔨  Building fresh packages...
✨  Done in 45.96s.
➜  govuk_publishing_components git:(master) yarn cache clean
yarn cache v1.22.10
success Cleared cache.
✨  Done in 2.24s.
➜  govuk_publishing_components git:(master) yarn install
yarn install v1.22.10
[1/4] 🔍  Resolving packages...
success Already up-to-date.
✨  Done in 0.30s.
```

I guess using the yarn cache means the cache is a bit broader, ie if PR A removes dependency 1, PR B can still access the cached one. But that seems minor so I'm not really clear on what advantages it gives.

Yup it'd be great to update the documentation to use it, would be great if the commit message justified why it's a better approach so that's recorded.

injms: I thought that [`yarn install --frozen-lockfile`](https://classic.yarnpkg.com/en/docs/cli/install/#toc-yarn-install-pure-lockfile) was the yarn equivalent to [`npm ci`](https://docs.npmjs.com/cli/v6/commands/npm-ci#description). Turns out that `npm ci` removes the node_modules folder before installing dependencies without updating the lockfile, and yarn only installs without updating the lockfile. (Shame that there's no yarn equivalent of `ci` as it's super useful.)

I think it's cleaner to install from yarn's cache as only exact dependencies from the lockfile would be used - caching the node_modules folder could lead to other dependencies being present and other hard to diagnose side effects.

I'll update the examples in the documentation and add the reasoning in the commit message there as well.

spec/visual_regression_tests/all_components.rb
chao-xian: Sorry to wade in with an opinion. I can see both sides of the argument here - it does feel a bit polluty to have this part of rspec but as that's what came out of the box plus we don't yet know what problems are really perhaps we should look to get this merged and in use, and collect some data? We have Kev's mini-spike to fall back on if we need to which is great. Feels like the sooner FE devs can make use of this the better.

injms: Thanks for taking the time to have a deep dive into this Kevin - I really appreciate the time you've taken investigate the rake task and the rest of the pull request.

I've had a ponder on whether this should or shouldn't be a rake task - and despite my earlier thoughts, now I think that this is spec / test and should be run using RSpec. 

 - Though the assertion isn't in the file, there a visual "assertion" that runs using Percy - so it is a test and should be treated one
 - The answer to "Where's the visual regression test?" should be "In the folder with the other tests"
 - I think that the set up for running this in RSpec is a lot clearer to understand than in the script - there's a lot of things that comes for free with RSpec that need to be set up in the script.
 - Frontend developers need to be able to easily update this - and RSpec is a tool that we're familiar with. This means that this is less likely to just gather dust when something needs updating in the future.
 - Excluding it from the default set of tests is because it's expensive to run in terms of both time and money, not because it isn't a test

If this causes problems in the future, we can and should revisit this - but for now using RSpec is fine and shouldn't be a blocker for merging this in.

kevindew: Sure, thanks for your perspectives here and no problem for chiming in @chao-xian - it's great to have other insights.

I get what you mean with your points. I do actually agree that Percy and RSpec should be together. My concerns here are that we're somewhat using RSpec as a mechanism to run a script, rather than actually integrating Percy and RSpec or organising the file like a spec - so it might be better a separation. However you have both raised good points too.

I do appreciate that the reorganisation aspect is beyond the scope of this first pass, so the remaining part I'm keen to resolve is doing non standard RSpec usage so this runs in some places not others. I've tried to come up with a counter proposal for that by using tags: https://github.com/alphagov/govuk_publishing_components/commit/09f8596820df372a4a41dd250e4410bcdddfad87

How does that sound?

.github/workflows/visual-regression-tests.yml
kevindew: Ah - I didn't know about `npm ci`. Cool thanks re: updating the docs 👍

chao-xian:
Where are we with this now? What are the minimum steps we could take to get this over the line, if there is anything else at all? I'm not clear what else is needed @injms @kevindew @richardTowers

injms:
I've cherry picked @kevindew's 09f8596820df372a4a41dd250e4410bcdddfad87 onto this branch - using `--tag` is much more elegant and future proof. 

Still musing on how to run the tests as they're needed without burning through all of the monthly allowance in one pull request.

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,40 @@
...
+
+jobs:
+  visual-regression-test-components:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - uses: ruby/setup-ruby@v1
+      - uses: actions/setup-node@v1
+        with:
+          node-version: "12"
```
injms, line 18: Note to self - see if this can be updated to 14 LTS

kevindew: It might be best to keep it on 12 anyway since that is the Node version on GOV.UK CI.

kevindew:
> > Still musing on how to run the tests as they're needed without burning through all of the monthly allowance in one pull request.
> 
> So until we have limitless money to spend on this, the best solution I can think of is to only run Percy when a pull request is marked as ready for review.
> 
> Advantages:
> 
> * Only runs when a pull request is ready
> * Doesn't run on every push, so doesn't burn though all the credits super quick
> * Runs on every pull request
> 
> Disadvantages:
> 
> * Doesn't run on every push, which makes it fiddly to re-run the VRTs as the pull request needs to put in and out of draft
> * Noisy, as there's a slack notification that pings every time a pull request is ready for review

I was trying to read up on the `ready_for_review` event to understand exactly when it occurs (which is barely documented), and the best I could find was:  https://frontside.com/blog/2020-05-26-github-actions-pull_request/#how-do-workflows-trigger-on-pull_request which leads me to think that this job will only get triggered in the scenario that someone opens a draft PR and then it'll run only once when they set that as ready for review. So if someone opened a PR, forgoing the draft stage this would never run. Is that how you understand it or anticipate it?

.github/workflows/visual-regression-tests.yml
injms: Good to know - I'll keep it on 12. Ta!

injms:
> ... which leads me to think that this job will only get triggered in the scenario that someone opens a draft PR and then it'll run only once when they set that as ready for review. So if someone opened a PR, forgoing the draft stage this would never run. Is that how you understand it or anticipate it?

My understanding / hope was that it would run on normal pull requests as they were raised, and when a pull request was changed from draft to ready. Reading that article makes me think otherwise - that the trigger only happens when going from draft to ready, rather than when ready.

I think I'm going to have to change this to the default - running when a pull request is raised, pushed to, or re-opened. That'll ensure that this is a useful test.

We could prevent non-essential runs by [ignoring some paths in the repo][ignore] or [only watching certain paths in the repo][include] - I'll do that in another pull request as I'll want to investigate what is okay to include or ignore.

If we burn through the credits then we burn through the credits - that won't prevent a pull request from being merged in since it is not a mandatory check.

[include]: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths
[ignore]: https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths

kevindew:
> > ... which leads me to think that this job will only get triggered in the scenario that someone opens a draft PR and then it'll run only once when they set that as ready for review. So if someone opened a PR, forgoing the draft stage this would never run. Is that how you understand it or anticipate it?
> 
> My understanding / hope was that it would run on normal pull requests as they were raised, and when a pull request was changed from draft to ready. Reading that article makes me think otherwise - that the trigger only happens when going from draft to ready, rather than when ready.
> 
> I think I'm going to have to change this to the default - running when a pull request is raised, pushed to, or re-opened. That'll ensure that this is a useful test.
> 
> We could prevent non-essential runs by [ignoring some paths in the repo](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-ignoring-paths) or [only watching certain paths in the repo](https://docs.github.com/en/actions/reference/workflow-syntax-for-github-actions#example-including-paths) - I'll do that in another pull request as I'll want to investigate what is okay to include or ignore.
> 
> If we burn through the credits then we burn through the credits - that won't prevent a pull request from being merged in since it is not a mandatory check.

That sounds reasonable and a path that isn't surprising. It looked like even trying to set a check to ignore draft PR's had some surprising complications: https://github.community/t/dont-run-actions-on-draft-pull-requests/16817/8

Be great if we can eventually get a plan so this can be less of a concern.

spec/visual_regression_tests/all_components_spec.rb
```diff
@@ -0,0 +1,48 @@
+require "rails_helper"
+require "percy"
+require "uri"
+
+describe "visual regression test runner Percy", visual_regression: true do
+  it "takes a screenshot of each component" do
+    # Freeze time for consistency
+    travel_to(Time.zone.local(2020, 12, 1, 6, 0, 0))
```
kevindew, line 8: This stub won't be reset after the test which could lead to some surprising situations for any Ruby running after the test. You can fix this by running the code in a block or using a `travel_back` method at the end of the test

kevindew (commented):
Just a couple of minor things, then I think this is good to merge.

.github/workflows/visual-regression-tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      - name: Get Yarn cache directory path
+        id: yarn-cache-directory-path
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - name: Check for cached Yarn dependencies
+        uses: actions/cache@v2
+        with:
+          path: ${{ steps.yarn-cache-directory-path.outputs.dir }}
+          key: yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: yarn-
+      - run: bundle install --jobs 4 --retry 3 --deployment
```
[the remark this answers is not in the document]
injms: Well that's a lot nicer - done 👍

spec/visual_regression_tests/all_components_spec.rb
injms: Thanks for spotting this - fixed ⌛

kevindew (approved):
Super, I'm looking forward to seeing it in action

PR #1017, opened 2019-08-01
Add tests for feedback form fields

bilbof:
This ensures that tests will fail if required fields are removed from the component HTML.

Co-Author: @benthorner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add tests for email feedback form
```
andysellick: Can you add the PR number as per the other entries?

bilbof: Sure thing, done :)

PR #1018, opened 2019-08-02
Fix for select height zoom problem

andysellick:
## What
This is a fix for a problem noticed during user research where a participant had their phone text size set very high. This resulted in the text in the select component being cropped slightly. This can be reproduced using Firefox's 'zoom text only' option, and it looks like this:

Screen Shot 2019-08-02 at 14 18 00

I've raised an [issue in govuk-frontend](https://github.com/alphagov/govuk-frontend/issues/1519) but it's likely that their fix will not help on GOV.UK as we're using compatibility mode for elements and toolkit, and this issue apparently relates to the base font size.

The problem is caused by the height of selects being set at 40px. The best solution seems to be to set the height using `em`, so that it scales with the text size of the select. An alternative is to set it using `rem` but since the component guide doesn't set a base font size this results in the height appearing differently there and in finder-frontend.

I think (correct me if I'm wrong) that as long as there's an explicit font size on the select (which there is) even if it's specified using `em` (which it is) as long as the height is greater than `1em` (which it is now) then the height of the select will always exceed the height of the text, so this problem shouldn't occur.

Screen Shot 2019-08-02 at 14 26 32

## Why
Well, it's broken.

## Visual Changes
I've overridden the height of `40px` with a height of `2.14em`, which appears to be the same, so there should be no visual changes. I'm pretty sure we're only using this component in finder-frontend, so we only need to test the search and finders pages.

## View Changes
https://govuk-publishing-compo-pr-1018.herokuapp.com/component-guide/select

Trello card: https://trello.com/c/VWJHWlBC/922-fix-facet-text-zoom-problem

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_select.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_select.scss
@@ -3,3 +3,11 @@
 .gem-c-select__select--full-width {
   width: 100%;
 }
+
+// Solution to text inside selects becoming unreadable if font size in
+// the browser is increased. This is currently a problem in govuk-frontend
+.gem-c-select {
+  .govuk-select {
+    height: 2.14em;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Add tests for email feedback form (PR #1017)
 * Add `lang` attribute option to subscription links (PR #1019)
+* Fix for select height zoom problem (PR #1018)
 
 ## 17.20.0
 
@@ -18,6 +19,7 @@
 * Change statistics header link to link to new finder (PR #1015)
 
 ## 17.19.1
+
 * Replace subscription links images with SVG (PR #1008)
 * Change subscription links CSS (PR #1007)
 


app/assets/stylesheets/govuk_publishing_components/components/_select.scss
```diff
@@ -3,3 +3,11 @@
 .gem-c-select__select--full-width {
   width: 100%;
 }
+
+// Solution to text inside selects becoming unreadable if font size in
+// the browser is increased. This is currently a problem in govuk-frontend
+.gem-c-select {
+  .govuk-select {
```
nickcolley: It'd be good to catchup with @alex-ju I think you may need to cater for both cases:

1. compatibility mode turned on
2. compatibility mode turned off

I still think we could fix this in govuk-frontend...

andysellick: > I still think we could consider a fix upstream in govuk-frontend first?

Yes, if that's an option. Sorry, the impression I had was that it wasn't. Happy to have that happen, although we need a fix quickly and I didn't want to put pressure on anyone.

nickcolley: > I've raised an issue in govuk-frontend but it's likely that their fix will not help on GOV.UK as we're using compatibility mode for elements and toolkit, and this issue apparently relates to the base font size.

I think I may have miscommunicated, we already output different units for the font scale depending on if the compatibility mode is set, we may be able to use similar logic to solve this issue too.

nickcolley: Spoke with Andy, and we'll try for a long term fix on govuk-frontend but that shouldnt block quicker improvements here.

andysellick:
@alex-ju this change shouldn't affect select multiples, should it? It's specific to the select component, which doesn't have a multiple option.

I just want to check this won't break anything.

alex-ju:
@andysellick yep, sorry if I caused confusion, the select in `govuk_publishing_components` doesn't allow a multiple attribute at the moment, but the one in `govuk-frontend` does.

andysellick:
@alex-ju have tested this with compatibility mode on and off, doesn't seem to make a difference, works fine in both.

alex-ju:
Needs a rebase to get it into [17.21.0](https://github.com/alphagov/govuk_publishing_components/pull/1020)

PR #1021, opened 2019-08-05
Normalise falsey values to nil to prevent `lang="false"`

injms:
## What
Checks that the locale attributes being passed into the subscription links components are falsey - if falsey then `nil` is used.

## Why
When a `false` was being used as a parameter in the subscription components, it was being cast and used as a string - leading to `lang="false"`. Not a big deal, since (according to the spec) this should be ignored, but worth fixing.

## Visual Changes
None.

## View Changes
https://govuk-publishing-compo-pr-1021.herokuapp.com/

### DIFF
--- a/app/views/govuk_publishing_components/components/_subscription-links.html.erb
+++ b/app/views/govuk_publishing_components/components/_subscription-links.html.erb
@@ -16,8 +16,8 @@
 
   hide_heading ||= false
 
-  email_signup_link_text_locale = local_assigns[:email_signup_link_text_locale]
-  feed_link_text_locale = local_assigns[:feed_link_text_locale]
+  email_signup_link_text_locale = local_assigns[:email_signup_link_text_locale].present? ? local_assigns[:email_signup_link_text_locale] : nil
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_subscription-links.html.erb
```diff
@@ -16,8 +16,8 @@
 
   hide_heading ||= false
 
-  email_signup_link_text_locale = local_assigns[:email_signup_link_text_locale]
-  feed_link_text_locale = local_assigns[:feed_link_text_locale]
+  email_signup_link_text_locale = local_assigns[:email_signup_link_text_locale].present? ? local_assigns[:email_signup_link_text_locale] : nil
```
tijmenb: You could shorten this with [`.presence`](https://apidock.com/rails/Object/presence) to `local_assigns[:email_signup_link_text_locale].presence` (although it might not be clearer)

injms: Done - thanks for the suggestion.

PR #1022, opened 2019-08-05
Use components.publishing.service.gov.uk

tijmenb:
## What

The main component guide is now hosted on https://components.publishing.service.gov.uk.

## Why

We can't use the crown and font on non *.gov.uk sites. It looks better.


nickcolley:
:art: :100:

PR #1035, opened 2019-08-08
[WIP][DO NOT MERGE] Add public/frontend layout component

vanitabarrett:
This is very WIP.

http://govuk-publishing-compo-pr-1035.herokuapp.com/public

To Do:

- [x] Add header
- [x] Add footer using layout footer component
- [x] Allow page content to be passed to component
- [x] Add example to docs
- [x] Add a11y criteria
- [ ] Fix logo spacing
- [x] Skip to main page link
- [ ] Compare with layout files in static/rendered html to see what else we might be missing
- [x] Compare with [govuk frontend layout](https://github.com/alphagov/govuk-frontend/blob/899264cd89559d31f0a790682254d94ecf66418b/package/govuk/template.njk)
- [ ] Make Brexit footer links in line with other footer links
- [x] Search bar shouldn't be shown all the time
- [x] Search icon not displaying
- [x] Search bar label not working
- [x] Fix search bar layout
- [x] Ensure we're using gem-c classes
- [ ] Header component displays small border on smaller screen sizes
- [ ] Whitehall layout with different search bar positioning and links to finders ([example](https://www.gov.uk/government/people/boris-johnson))
- [x] Favicon and other assets
- [x] Full width option, e.g: for topic page
- [x] Allow setting of body classes
- [x] Width container (logo lining up with cookie banner?)
- [x] Cookie banner
- [ ] omit_header ?
- [ ] Emergency banners
- [ ] Global banner
- [ ] Survey banner
- [x] Add tests
- [ ] Test responsiveness
- [ ] Browser testing
- [x] Check stylesheet and JS links in `head`
- [ ] Check JS functionality and move into gem if needed
- [ ] Check tracking
- [ ] Test in frontend apps
- [ ] Address comments

## Notes

This implementation makes the following assumptions:

- Expects an `application.js`, `application.css` and `print.css` file. Each Frontend app will have to follow this pattern.
- Meta tags, e.g: for analytics, are not included as a param. Applications can add these by using the [meta tags component](https://components.publishing.service.gov.uk/component-guide/meta_tags) within the `:head` section which this component renders.

To test locally in a frontend app: update the frontend app Gemfile to point to this branch; remove slimmer (and all references to it) from the Gemfile; update the relevant layout file to use the component rather than hardcode the layout.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
@@ -1,5 +1,4 @@
 .gem-c-layout-for-public {
-  @include govuk-font(16);
   margin: 0;
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
@@ -1,4 +1,5 @@
 .gem-c-layout-for-public {
   margin: 0;
+  font-family: $govuk-font-family;
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
@@ -1,0 +1,4 @@
+.gem-c-layout-for-public {
+  @include govuk-font(16);
+  margin: 0;
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
@@ -2,3 +2,4 @@
   @include govuk-font(16);
   margin: 0;
 }
+
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -55,6 +55,8 @@
       <%= yield %>
     </main>
 
+    <%= render "govuk_publishing_components/components/feedback", {} %>
+
     <% unless local_assigns[:hide_footer_links] %>
       <%= render "govuk_publishing_components/components/layout_footer", {
         with_border: true,
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -2,6 +2,7 @@
   title ||= "GOV.UK - The best place to find government services and information"
   html_lang ||= "en"
   full_width ||= false
+  without_search ||= false
   body_classes ||= ""
   main_classes ||= ""
 -%>
@@ -48,7 +49,7 @@
 
     <%= render "govuk_publishing_components/components/layout_header", {
       environment: "public",
-      search: true,
+      search: !without_search,
       remove_bottom_border: full_width
     } %>
 
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -416,6 +416,9 @@
         }
       } %>
     <% end %>
+
+    <%# if no GOVUK-namespaced module has loaded we can assume JS has failed and remove the class %>
+    <script>if (typeof window.GOVUK === 'undefined') document.body.className = document.body.className.replace('js-enabled', '');</script>
   </body>
 </html>
 <!-- Thanks Martha -->
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -1,6 +1,7 @@
 <%
   title ||= "GOV.UK - The best place to find government services and information"
   html_lang ||= "en"
+  full_width ||= false
 -%>
 <!DOCTYPE html>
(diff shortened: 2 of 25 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,414 @@
...
+    <meta name="theme-color" content="#0b0c0c" />
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <%# The default og:image is added below :head so that scrapers see any custom metatags first, and this is just a fallback %>
+    <meta property="og:image" content="<%= asset_path "opengraph-image.png" %>">
+
+    <%= yield :head %>
+  </head>
+  <body class="gem-c-layout-for-public">
```
36degrees: I'd recommend using `govuk-template__body` here instead – as per https://design-system.service.gov.uk/styles/page-template/

app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
```diff
@@ -0,0 +1,4 @@
+.gem-c-layout-for-public {
+  @include govuk-font(16);
```
36degrees: Don't set a font size on the `html` or `body` – in a non-compatibility mode world all of the font sizes are defined in rem, but we rely on the user being able to change the root font-size in order to resize the text without zooming.

Also, because `govuk-font` uses the responsive typography scale, this'd result in the root font size being 14px on mobile, which means on mobile all of the text would be 14/16ths of what it should be.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,414 @@
+<%
+  title ||= "GOV.UK - The best place to find government services and information"
+  html_lang ||= "en"
+%>
+
+<!DOCTYPE html>
```
36degrees: Will this whitespace get trimmed automatically? I believe [some older browsers will be forced into 'quirks mode' if there is whitespace before the doctype](https://github.com/alphagov/govuk-frontend/pull/949).

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
36degrees: I think you'll also need the [script that adds `js-enabled`](https://github.com/alphagov/govuk-frontend/blob/439564855f1f7c86dfd5dd618bed369b709e1ee5/package/govuk/template.njk#L31) in order for JS-enabled components to work correctly.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,414 @@
...
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <%# The default og:image is added below :head so that scrapers see any custom metatags first, and this is just a fallback %>
+    <meta property="og:image" content="<%= asset_path "opengraph-image.png" %>">
+
+    <%= yield :head %>
+  </head>
+  <body class="gem-c-layout-for-public">
+    <%= render "govuk_publishing_components/components/layout_header", {
```
36degrees: Does layout_header include a skip link?

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -24,13 +25,15 @@
           </span>
         </span>
 
-        <span class="govuk-header__product-name">
-          <%= product_name %>
-        </span>
+        <% unless local_assigns[:hide_product_details] %>
```
alex-ju: I'm aware this is not ready for review, but since it's open I would suggest we hide the wrapping span if no `product_name` is provided instead of introducing a new flag.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,414 @@
...
+    <meta charset="utf-8" />
+    <title><%= title %></title>
+
+    <%= csrf_meta_tags %>
+
+    <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
+    <%= stylesheet_link_tag 'print', media: 'print' %>
+    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+
+    <%= stylesheet_link_tag "fonts.css", media: "all", integrity: true, crossorigin: "anonymous" %>
```
36degrees: I don't believe this `fonts.css` is needed any more.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
alex-ju: We have these suggestions exemplified in https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
vanitabarrett: Added to checklist in PR description

maxgds:
Closing this in favour of https://github.com/alphagov/govuk_publishing_components/pull/1265 where work will continue. It was necessary to rebase to pick up a lot of changes that had gone into master, which I did on a separate branch to keep it non destructive if I messed up. Merging the two back together isn't a simple merge so it makes sense to continue on that as the live branch: https://github.com/alphagov/govuk_publishing_components/pull/1265

PR #1037, opened 2019-08-09
Add inverse flag to contents list component

vanitabarrett:
## What
Screen Shot 2019-08-09 at 12 09 58

## Why
We have a need for using the contents list component within a topic page header, which is blue.

https://govuk-publishing-compo-pr-1037.herokuapp.com/component-guide/contents_list/inverse

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -10,10 +10,25 @@
   box-shadow: 0 20px 15px -10px govuk-colour("white");
 }
 
+.gem-c-contents-list--inverse {
+  background: transparent;
+  box-shadow: none;
+}
+
 .gem-c-contents-list__title {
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -10,12 +10,27 @@
...
 }
 
+.gem-c-contents-list--inverse .gem-c-contents-list__title,
+.gem-c-contents-list--inverse .gem-c-contents-list__link:link,
+.gem-c-contents-list--inverse .gem-c-contents-list__link:visited {
+  color: govuk-colour("white");
+}
+
+.gem-c-contents-list--inverse .gem-c-contents-list__link:focus {
+  color: govuk-colour("black");
```
alex-ju: Minor: Could we use `$govuk-focus-text-colour` for focus states? Returns the same value, but it makes it consistent in any future changes will happen to the focus states.

PR #1041, opened 2019-08-09
Make the gem smaller

tijmenb:
We currently ship everything in `node_modules`, which includes the dev dependencies. This changes it to only include in the gem files from govuk-frontend. This will significantly decrease the size of the resulting gem.


andysellick:
@tijmenb are we able to test this other than by deploying it?

nickcolley:
I think this is safe now because GOV.UK Frontend has no dependencies but might change in the future :+1: :100:

PR #1046, opened 2019-08-14
Expand share links component

andysellick:
## What
Add two new things to the share links component:

- a 'columns' option to line up the links in responsive columns
- an icon for whatsapp sharing

Controversial: the 'columns' option uses flexbox, which isn't supported fully (or at all) in IE10 and downwards. In those browsers the columns are not even. I'm not sure there's a way of doing it in those browsers - even if we put in specifics for every breakpoint (e.g. mobile width 100%, tablet width 50%, etc.) there's no guarantee that'll work because we don't know how much text might be in the links.

## Why
A page needs to be shared on four social networks including whatsapp and the design calls for a column arrangement.

## Visual Changes
Here's the columns option at various widths.

Screen Shot 2019-08-14 at 08 45 37

Screen Shot 2019-08-14 at 08 45 45

Screen Shot 2019-08-14 at 08 45 55

...etc. 

Here's the not-quite-working IE10 version. The behaviour is exactly as intended (the links fill the available space then wrap onto the next line) but the columns aren't quite the right width.

Screen Shot 2019-08-14 at 09 50 01

## View Changes
https://govuk-publishing-compo-pr-1046.herokuapp.com/component-guide/share_links


nickcolley:
This approach is good, the fallback is considered, I would push back on the design and see if it there's really a need to change the layout like this but perhaps there's a good reason for it.

andysellick:
> This approach is good, the fallback is considered, I would push back on the design and see if it there's really a need to change the layout like this but perhaps there's a good reason for it.

Thanks Nick. This is for a new page, the design is to have four share links as the only things on one row of the full width page (see below but imagine two more links). I think until now we've only ever had two share links on any page so it's not been a problem. It feels like we're going to face this problem eventually so might as well tackle it here.

Screen Shot 2019-08-14 at 09 55 10

I've updated the description above to include a screenshot of what happens on older IE. I'm a little conflicted - it's not horribly bad but it's not perfect either, and as I said, I don't know what alternative to try, considering that the text length could vary for each component use.

app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
```diff
@@ -58,6 +58,17 @@ $share-button-height: 32px;
...
 }
 
+.gem-c-share-links--columns {
+  .gem-c-share-links__list {
+    display: flex;
+    flex-wrap: wrap;
+  }
+
+  .gem-c-share-links__list-item {
+    flex: 0 0 200px;
```
injms: I think it might be worth adding a touch of padding-right to `.gem-c-share-links__list-item`, as otherwise you could end up with text running right up to the icon on the right:

Screen Shot 2019-08-14 at 13 24 14

injms: Definitely a contrived example and a rare use - I leave up to you whether it's worth fixing :)

andysellick: No, that's a good point. I've added a teeny bit of padding right (and padding left for rtl) and it seems to have fixed it. Thanks!

PR #1048, opened 2019-08-14
Restore node_modules gem dependencies

andysellick:
## What
Restoring some dependencies from `node_modules` - we used to include everything but this was unnecessary so it was reduced to just `govuk-frontend`, but it looks like some other things are also needed.

- axe-core/axe.js

## Why
The last versions of the gem (18.1.0 and 18.0.1) appear to be broken because of missing files not included from `node_modules`.

### DIFF
--- a/govuk_publishing_components.gemspec
+++ b/govuk_publishing_components.gemspec
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/alphagov/govuk_publishing_components"
   s.license     = "MIT"
 
-  s.files = Dir["{node_modules/govuk-frontend,app,config,db,lib}/**/*", "DEVELOPMENT.md", "LICENCE.md", "Rakefile", "README.md"]
+  s.files = Dir["{node_modules/govuk-frontend,node_modules/axe-core,app,config,db,lib}/**/*", "DEVELOPMENT.md", "LICENCE.md", "Rakefile", "README.md"]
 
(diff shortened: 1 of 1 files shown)


govuk_publishing_components.gemspec
```diff
@@ -14,7 +14,7 @@ Gem::Specification.new do |s|
   s.homepage    = "https://github.com/alphagov/govuk_publishing_components"
   s.license     = "MIT"
 
-  s.files = Dir["{node_modules/govuk-frontend,app,config,db,lib}/**/*", "DEVELOPMENT.md", "LICENCE.md", "Rakefile", "README.md"]
+  s.files = Dir["{node_modules/govuk-frontend,node_modules/axe-core,app,config,db,lib}/**/*", "DEVELOPMENT.md", "LICENCE.md", "Rakefile", "README.md"]
```
tijmenb: Come to think of it, should this be the same as https://github.com/alphagov/govuk_publishing_components/blob/master/package.json#L16-L18? So jquery as well?

In that case we can automate it with something like 

```rb
JSON.parse(File.read("package.json"))["dependencies"].keys
```

andysellick: I was thinking it might also need jquery, yes. Thanks - probably going to end up doing a few versions of the gem this afternoon!

Where would the example line you gave go?

andysellick: Seems to have worked without jquery...

PR #1053, opened 2019-08-15
Add a "use_month_year_dates" option to document lists

barrucadu:
## What
Adds an option to change the date format for document lists to only show the month and year.

## Why
We have things where dates are provisional, like statistics announcements.  These all have a timestamp for the 1st of the chosen month, but may not actually be released on that day - they can be released any time in the month.  So displaying the day is incorrect.

Statistics announcements pages display dates like so: https://www.gov.uk/government/statistics/announcements/local-authority-revenue-expenditure-and-financing-england-2019-to-2020-provisional-outturn

This change is so that search pages can be consistent: https://www.gov.uk/search/research-and-statistics?content_store_document_type=upcoming_statistics&organisations%5B%5D=ministry-of-housing-communities-and-local-government

## Visual Changes
No change to existing behaviour.  The new thing looks like this:
Screen Shot 2019-08-15 at 11 09 39

## View Changes
https://govuk-publishing-compo-pr-1053.herokuapp.com/component-guide/document_list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## unreleased
+
+* Add option to hide day numbers in timestamps in document lists (PR #1053)
+
 ## 18.1.2
 
 * Restore jquery as gem dependency (PR #1051)
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -10,6 +10,9 @@
   within_multitype_list_class = " gem-c-document-list__multi-list" if within_multitype_list
   title_with_context_class = " gem-c-document-list__item-title--context"
 
+  use_month_year_dates ||= false
+  date_format = use_month_year_dates ? "%B %Y" : "%e %B %Y"
+
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
 %>
@@ -46,8 +49,8 @@
             <% item[:metadata].compact.each do |item_metadata_key, item_metadata_value| %>
               <li class="gem-c-document-list__attribute">
                 <% if item_metadata_key.to_s.eql?("public_updated_at") %>
-                  <time datetime="<%= item_metadata_value.iso8601 %>">
-                    <%= l(item_metadata_value, format: '%e %B %Y') %>
+                  <time datetime="<%= use_month_year_dates ? item_metadata_value.strftime('%Y-%m') : item_metadata_value.iso8601 %>">
+                    <%= l(item_metadata_value, format: date_format) %>
                   </time>
                 <% else %>
                   <%= item_metadata_value %>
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -46,8 +49,8 @@
             <% item[:metadata].compact.each do |item_metadata_key, item_metadata_value| %>
               <li class="gem-c-document-list__attribute">
                 <% if item_metadata_key.to_s.eql?("public_updated_at") %>
-                  <time datetime="<%= item_metadata_value.iso8601 %>">
-                    <%= l(item_metadata_value, format: '%e %B %Y') %>
+                  <time datetime="<%= use_month_year_dates ? item_metadata_value.strftime('%Y-%m') : item_metadata_value.iso8601 %>">
```
ChrisBAshton: I'd move this logic up to line 15, making a `datetime` variable

barrucadu: I thought of doing that initially, but it's a bit awkward because we want to call different methods, so it'd be something like

```ruby
datetime = use_month_year_dates? ->(date) { date.strftime("%Y-%m") } : ->(date) { date.iso8601 }
```

And be called like `datetime(item_metadata_value)`

Which I'm not sure is better

ChrisBAshton: Ah I'm with you, didn't spot that it's working on a temporary variable in the loop.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## unreleased
+
+* Add option to hide day numbers in timestamps in document lists (PR #1053)
```
andysellick: Nitpick: can this match the PR name please?

spec/components/document_list_spec.rb
```diff
@@ -74,6 +74,47 @@ def component_name
...
+    assert_select "#{li}[href='/government/publications/parental-responsibility-measures-for-behaviour-and-attendance']", text: "School behaviour and attendance: parental responsibility measures"
+    assert_select "#{attribute} time", text: "January 2017"
+    assert_select "#{attribute} time[datetime='2017-01']"
+    assert_select ".gem-c-document-list__attribute", text: "Statutory guidance"
+
+    assert_select "#{li}[href='/become-an-apprentice']", text: "Become an apprentice"
+    assert_select ".gem-c-document-list__item-description", text: 'Becoming an apprentice - what to expect'
+    assert_select "#{attribute} time", text: "July 2017"
+    assert_select "#{attribute} time[datetime='2017-07']"
+  end
```
andysellick: On the basis that this test is for the date stuff only, I think lines 107, 110, 112, and 113 are unnecessary? Isn't this being tested elsewhere in this file already?

barrucadu:
Closing this as we've found a way to handle statistics announcements that works in all cases, not just in cases where the date is "any time within this one month"

PR #1055, opened 2019-08-16
Fix image card responsiveness

andysellick:
## What

- when image card drops below 320px (unlikely, admittedly) it collapses into a single column but the margins are broken
- also the sass is a bit confusing
- this commit restructures the sass to make it more readable and fixes this problem
- behaviour should be (for standard image card): from 320px two columns, from 640px one column
- behaviour is perhaps not ideal (mobile with two columns is a little squished in some cases) but changing it is outside scope
- this commit does not change the behaviour, it only fixes the brokenness below 320px

## Why
The likelihood of anyone using a device narrower than 320px is very low, but this PR also makes the code more readable. Having said that, I'm not 100% sure I understand the new govuk-frontend V3 mixins for breakpoints and columns.

## Visual Changes
Before:

Screen Shot 2019-08-16 at 15 59 26

After:

Screen Shot 2019-08-16 at 16 40 34

I've checked this in collections (component is used on org pages) and the behaviour appears to be as before but includes the fix.

## View Changes
https://govuk-publishing-compo-pr-1055.herokuapp.com/component-guide/image_card

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix image card responsiveness (PR #1055)
+
 ## 18.1.2
 
 * Restore jquery as gem dependency (PR #1051)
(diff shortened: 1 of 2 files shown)


injms:
Not unlikely in the future - also since Apple Watch can now open links. 

The watch browser can either say it has a screen of 320px wide, or - if we add a meta tag - it will claim to have a screen < 320px wide. More about [designing web content for the Apple Watch](https://developer.apple.com/videos/play/wwdc2018/239/).

PR #1056, opened 2019-08-19
Make contextual title accept and present lang parameter

maxgds:
## What
Add ability to pass in a language param for 

## Why
Sometimes the page is translated but we don't have a translation for the page type string used in the contextual header, we need to mark this up for semantics and accessibility

## View Changes
https://govuk-publishing-compo-pr-1056.herokuapp.com/

### DIFF
--- a/app/views/govuk_publishing_components/components/_title.html.erb
+++ b/app/views/govuk_publishing_components/components/_title.html.erb
@@ -1,6 +1,7 @@
 <%
   average_title_length ||= false
   context ||= false
+  context_locale ||= false
   context_text = context.is_a?(Hash) ? context[:text] : context
   context_href = context.is_a?(Hash) ? context[:href] : false
   context_data = context.is_a?(Hash) ? context[:data] : false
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_title.html.erb
```diff
@@ -1,6 +1,7 @@
 <%
   average_title_length ||= false
   context ||= false
+  context_locale ||= false
```
vanitabarrett: This is currently setting `lang="false"` if no context_locale is passed. Perhaps only add the lang attribute if this isn't false?

spec/components/title_spec.rb
```diff
@@ -55,4 +60,9 @@ def component_name
     render_component(title: 'Margin wat', margin_bottom: 17)
     assert_select "[class='^=govuk-\!-margin-bottom-']", false
   end
+
+  it "adds a lang attribute when set" do
```
vanitabarrett: I'm not sure on the difference between the two tests you've added? Is one checking specifically because the context is now a link? If so, can we reflect that in the test name, to make it clear?

app/views/govuk_publishing_components/components/_title.html.erb
```diff
@@ -13,10 +14,11 @@
   classes = %w(gem-c-title govuk-!-margin-top-8)
   classes << "gem-c-title--inverse" if inverse
   classes << (shared_helper.get_margin_bottom)
+
 %>
 <%= content_tag(:div, class: classes) do %>
   <% if context %>
-    <p class="gem-c-title__context">
+    <p class="gem-c-title__context" lang='<%= context_locale %>'>
```
injms:
```suggestion
    <p class="gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
```

should only display `lang` if the context_locale is present.

spec/components/title_spec.rb
maxgds: Yeah, it was making sure it worked with and without a link, updated now.

app/views/govuk_publishing_components/components/_title.html.erb
maxgds: Fixed

PR #1059, opened 2019-08-20
Name and value can be set on button component

injms:
## What

Adds two optional parameters to the button component to let the `value` and `name` attributes be set.

## Why

A button's value is only sent as part of a form when the button is used to submit the form - having multiple buttons set with different values means that we can tell which button submitted a form. 

## Visual Changes

No visual changes - just code changes.

Before:
```html
<button class="gem-c-button govuk-button" type="submit">
  This is the button text
</button>
```

After:
```html
<button 
  class="gem-c-button govuk-button" 
  type="submit" 
  name="this_is_the_name" 
  value="this_is_the_value"
>
  This is the button text
</button>
```

## View Changes
https://govuk-publishing-compo-pr-1059.herokuapp.com/component-guide/button/with_name_and_value_set

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -99,7 +99,10 @@ examples:
       text: "Second button"
       inline_layout: true
   with_name_and_value_set:
-    description: By default, the button has no value or name set so it will not pass information when the form is submitted. This allows a name and value to be added so a button can add information to the form submission.
+    description: |
+      By default, the button has no value or name set so it will not pass information when the form is submitted. This allows a name and value to be added so a button can add information to the form submission.
+
+      Please note that Internet Explorer 6 and 7 have **breaking bugs** when submitting a form with multiple buttons - this can [change what value is submitted by the button](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Notes). Make sure to check your user needs and browser usage.
     data:
       text: "This is the button text"
       value: "this_is_the_value"
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -98,4 +98,10 @@ examples:
     data:
       text: "Second button"
       inline_layout: true
+  with_name_and_value_set:
+    description: By default, the button has no value or name set so it will not pass information when the form is submitted. This allows a name and value to be added so a button can add information to the form submission.
+    data:
+      text: "This is the button text"
+      value: "this_is_the_value"
+      name: "this_is_the_name"
 
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -5,7 +5,7 @@ module GovukPublishingComponents
     class ButtonHelper
       attr_reader :href, :text, :title, :info_text, :rel, :data_attributes,
                   :margin_bottom, :inline_layout, :target, :type, :start,
-                  :secondary, :secondary_quiet, :destructive
+                  :secondary, :secondary_quiet, :destructive, :name, :value
 
       def initialize(local_assigns)
         @href = local_assigns[:href]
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/docs/button.yml
```diff
@@ -98,4 +98,10 @@ examples:
     data:
       text: "Second button"
       inline_layout: true
+  with_name_and_value_set:
+    description: By default, the button has no value or name set so it will not pass information when the form is submitted. This allows a name and value to be added so a button can add information to the form submission.
```
vanitabarrett: Have you got an example use case for needing this? Just curious!

spec/components/button_spec.rb
```diff
@@ -141,4 +141,32 @@ def component_name
...
   end
+
+  it "renders with a name and a value attribute  when name and value are set" do
+    render_component(text: "Button one", value: "this_value", name: "this_name")
+
+    assert_select ".gem-c-button[name='this_name']", 1
+    assert_select ".gem-c-button[value='this_value']", 1
+  end
+
+  it "renders an without a name or a value attribute when only value is set" do
```
vanitabarrett:
```suggestion
  it "renders without a name or a value attribute when only value is set" do
```

spec/components/button_spec.rb
```diff
@@ -141,4 +141,32 @@ def component_name
...
+    render_component(text: "Button one", value: "this_value")
+
+    assert_select ".gem-c-button[name]", 0
+    assert_select ".gem-c-button[value]", 0
+  end
+
+  it "renders an without a name or a value attribute when only name is set" do
```
vanitabarrett:
```suggestion
  it "renders without a name or a value attribute when only name is set" do
```

spec/components/button_spec.rb
```diff
@@ -141,4 +141,32 @@ def component_name
...
+    render_component(text: "Button one", name: "this_name")
+
+    assert_select ".gem-c-button[name]", 0
+    assert_select ".gem-c-button[value]", 0
+  end
+
+  it "renders an without a name or a value attribute neither name or value is set" do
```
vanitabarrett:
```suggestion
  it "renders without a name or a value attribute when neither name or value is set" do
```

lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -36,6 +38,8 @@ def html_options
         options[:data] = data_attributes if data_attributes
         options[:title] = title if title
         options[:target] = target if target
+        options[:name] = name if name.present? && value.present?
+        options[:value] = value if value.present? && name.present?
```
andysellick: Nitpick: I'd maybe switch the logic on this line around to match the previous one, just to increase the readability a little, i.e.

```
options[:name] = name if name.present? && value.present?
options[:value] = value if name.present? && value.present?
```

spec/components/button_spec.rb
```diff
@@ -141,4 +141,32 @@ def component_name
 
     assert_select ".gem-c-button--inline", text: "Button one"
   end
+
+  it "renders with a name and a value attribute  when name and value are set" do
```
andysellick: Nitpick: is that a double space between attribute and when?

spec/components/button_spec.rb
injms: 🤦‍♂️

lib/govuk_publishing_components/presenters/button_helper.rb
injms: Good idea - done.

spec/components/button_spec.rb
injms: 🤦‍♂

app/views/govuk_publishing_components/components/docs/button.yml
injms: Um, not one that's live - just me trying to gracefully get a form to either display a filter page or a results page without losing the information either currently in the form or that's just been entered by the user.

36degrees:
It's unlikely to be a concern, but note that if you're using multiple buttons and using the `value` to distinguish between different buttons being clicked, there are [some bugs in IE7 (and IE6!) to be aware of](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#Notes).

You may opt to use `input type="submit"` instead, as it doesn't have these issues, but that has its own downsides.

There's more context in https://github.com/alphagov/govuk_elements/pull/593.

injms:
Thanks @36degrees - I've added in a note to the documentation since it's a bug that could break things quite significantly, so even if the browser's aren't used it's worth being aware of.

PR #1061, opened 2019-08-20
Add summary list component

alex-ju:
## What
This component is based on the [summary list component in govuk-frontend](https://design-system.service.gov.uk/components/summary-list/).

The API was modelled after the [summary component in Content Publisher](https://content-publisher.integration.publishing.service.gov.uk/component-guide/summary) as it should be replaced with this component after it gets released.

## Why
This component is used already in Content Publisher and it's needed in Collections Publisher. It's also the only missing component from the Design System

## View Component
https://govuk-publishing-compo-pr-1061.herokuapp.com/component-guide/summary_list

[Trello card](https://trello.com/c/arYq0tST)

### DIFF
--- a/app/views/govuk_publishing_components/components/_summary_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_summary_list.html.erb
@@ -1,0 +1,67 @@
+<%
+  id ||= nil
+  title ||= nil
+  edit ||= {}
+  items ||= []
+  block ||= yield
+%>
+<% if title || items.any? %>
+  <%= tag.div class: "gem-c-summary-list", id: id do %>
+    <% if title %>
+      <%= tag.h3 title, class: "govuk-heading-m" %>
+      <% if edit.any? %>
+        <%= tag.ul class: "govuk-summary-list__actions-list" do %>
+          <%= tag.li class: "govuk-summary-list__actions-list-item" do %>
+            <%= link_to edit.fetch(:href),
+                      class: "govuk-link gem-c-summary-list__edit-section-link",
+                      title: "Edit #{title}",
+                      data: edit.fetch(:data_attributes, {}) do %>
+               Edit <%= tag.span title, class: "govuk-visually-hidden" %>
+            <% end %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
+
+    <% if items.any? %>
+      <%= tag.dl class: "govuk-summary-list" do %>
+        <% items.each do |item| %>
+          <%= tag.div class: "govuk-summary-list__row" do %>
+
+            <%= tag.dt item[:field], class: "govuk-summary-list__key" %>
+            <%= tag.dd item[:value], class: "govuk-summary-list__value" %>
+
+            <% if item.fetch(:edit, {}).any? || item.fetch(:delete, {}).any? %>
+              <%= tag.dd class: "govuk-summary-list__actions" do %>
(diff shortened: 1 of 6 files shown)


injms:
Is the [with block example](https://govuk-publishing-compo-pr-1061.herokuapp.com/component-guide/summary_list/with_block/preview) needed or working as expected?

alex-ju:
@injms Good observation. It looks quite generic, but it's used in Content Publisher to render blank/empty states (see Topics in screenshot below) or custom blocks (see Lead image in screenshot below). I will add an edit section to the example and a description to explain when to use it.

Screen Shot 2019-08-21 at 09 19 50

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -0,0 +1,67 @@
...
+                      <% end %>
+                    <% end %>
+                  <% end %>
+                <% end %>
+              <% end %>
+            <% end %>
+          <% end %>
+        <% end %>
+      <% end %>
+    <% end %>
```
ChrisBAshton: I wonder if we can cut down any of the nesting by doing more conditional checks up front and less 'wrapping' deeper inside the nest?

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -0,0 +1,67 @@
...
+                      <% end %>
+                    <% end %>
+                  <% end %>
+                  <% if item.fetch(:delete, {}).any? %>
+                    <%= tag.li class: "govuk-summary-list__actions-list-item" do %>
+                      <%= link_to item[:delete].fetch(:href),
+                                  class: "govuk-link",
+                                  title: "Delete #{item[:field]}",
+                                  data: item[:edit].fetch(:data_attributes, {}) do %>
+                        Delete<%= tag.span item[:field], class: "govuk-visually-hidden" %>
```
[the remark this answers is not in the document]
alex-ju: Good spot! Updated

app/views/govuk_publishing_components/components/_summary_list.html.erb
alex-ju: Hmm.. it's not great in terms of nesting, but this is the level of granularity we want the mark-up to be, as there are lots of optional bits where we don't want to render markup unless we have a value for it. Open to suggestions if you see a way to make the code more readable while rendering things optionally.

app/views/govuk_publishing_components/components/docs/summary_list.yml
```diff
@@ -0,0 +1,65 @@
+name: Summary list
+description: Use the summary list to summarise information, for example, a user’s responses at the end of a form.
+accessibility_criteria: |
+  Action links in the component must:
+
+  * indicate what the action reffers to (e.g. Change _name_)
```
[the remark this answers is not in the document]
alex-ju: Thanks!

app/views/govuk_publishing_components/components/_summary_list.html.erb
ChrisBAshton: I suppose you could make a helper `list-item` component, as the list items all look the same apart from the title prefix ("Delete", "Edit" etc)

But also wouldn't want to go for 'the wrong abstraction'! Happy to leave with you.

alex-ju:
Thanks for your reviews @injms and @ChrisBAshton. Updated with your suggestions, do let me know if you have any other comments on this.

PR #1062, opened 2019-08-21
Consolidate input aria documentation and tests

andysellick:
**This PR has been updated and no longer does what it originally did.**

## What
Consolidate the aria attribute documentation for the input component.

A possible future change (for more than just this component) could be to pass any required aria attributes to components rather than having to add new options for each new use case, as below.

```
  label: {
    text: "This is an example only and may not work with a screen reader"
  },
  name: "labelledby",
  aria: {
    describedby: "wrapper",
    controls: "wrapper"
  }
} %>
```

```
aria ||= []
<%= tag.input name: name,
    aria: aria
%>
```

## Visual Changes
No visual changes.

## View Changes
https://govuk-publishing-compo-pr-1062.herokuapp.com/component-guide/input

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/input.yml
+++ b/app/views/govuk_publishing_components/components/docs/input.yml
@@ -37,22 +37,18 @@ examples:
         text: "Has an id"
       name: "hasid"
       id: "hasid"
-  aria_described_by:
+  with_aria_attributes:
     description: |
-      Allows the addition of an aria-describedby attribute. Note that this will be overridden in the event of an error, where the error will be used for the describedby attribute value.
+      The component accepts two possible aria attributes: aria-describedby and aria-controls.
 
-      The example below uses the ID of the main container for demonstration purposes as there aren't any useful elements with IDs in the component guide. In real use this would be passed the ID of an element that correctly described the input.
+      [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) is used to indicate the ID of the element that describes the input. This will be overridden in the event of an error, where the error will be used for the describedby attribute value. This example uses the ID of the main container for demonstration purposes as there aren't any useful elements with IDs in the component guide. In real use this would be passed the ID of an element that correctly described the input.
+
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -37,23 +37,22 @@ examples:
...
+      [aria-describedby](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-describedby_attribute) is used to indicate the ID of the element that describes the input. This will be overridden in the event of an error, where the error will be used for the describedby attribute value. This example uses the ID of the main container for demonstration purposes as there aren't any useful elements with IDs in the component guide. In real use this would be passed the ID of an element that correctly described the input.
+
+      aria-controls allows the addition of an aria-controls attribute, for use in places like the finders where the page is updated dynamically after value changes to the input.
+
+      [aria-invalid](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-invalid_attribute) indicates invalid entry, such as a phone number where a postcode is expected. It should be programmatically set as the result of a validation process.
     data:
       label:
         text: "This is an example only and may not work with a screen reader"
```
injms: I'd say this shouldn't work, since there is no corresponding element that this controls - the example would need the wrapper to work. Unless there's a way of easily adding an extra element to the example?

```suggestion
        text: "This is an example to look at HTML output and does not work with a screen reader"
```

nickcolley:
Could we avoid doing this by updating the pattern on that page to match the current recommendations for errors? I don't believe we promote using `aria-invalid` at the moment, so might be able to simplify things this way.

nickcolley:
I've made an example of this: https://govuk-frontend-local-council.glitch.me/

nickcolley:
@36degrees did an investigation here: https://github.com/alphagov/govuk-frontend/issues/1168

andysellick:
@nickcolley you're absolutely right. I was so busy trying to implement this feature I didn't stop to consider if we actually needed it or not. I've reworked this PR to not have this addition but still have the consolidation of the aria examples and tests, which I think is worth keeping. Thanks!

PR #1063, opened 2019-08-22
RFC 3: change how we include gem assets in applications

andysellick:
👉 [Rendered version](https://github.com/alphagov/govuk_publishing_components/blob/a29d7f94aef5046790d46909d963dc924084f5cf/docs/rfc-003.md)

### DIFF
--- a/docs/rfc-003.md
+++ b/docs/rfc-003.md
@@ -1,0 +1,66 @@
+# Include specific component assets in applications
+
+## Problem and background
+
+The [current method for including the components](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#4-include-the-assets) in a frontend application (such as `government-frontend`) is to include `_all_components.scss` and `_all_components_print.scss` in the Sass of that application. This has some benefits:
+
+- all components are automatically included in the application
+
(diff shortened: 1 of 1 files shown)


vanitabarrett:
Should we be putting this in the govuk-rfc repo?

andysellick:
The past two RFCs (admittedly also written by me, so I could be generally wrong on this) were in this repo: https://github.com/alphagov/govuk_publishing_components/pulls?utf8=%E2%9C%93&q=is%3Apr+is%3Aclosed+rfc

sihugh:
I agree with @vanitabarrett!

docs/rfc-003.md
```diff
@@ -0,0 +1,66 @@
...
+// currently
+//= require govuk_publishing_components/all_components
+```
+
+```
+// proposed
+//= require govuk_publishing_components/components/accordion
+//= require govuk_publishing_components/components/button
+// (etc.)
+```
```
injms: I think this would also require explicit initialisation of each component as well.

andysellick:
@vanitabarrett @sihugh fine... https://github.com/alphagov/govuk-rfcs/pull/110

PR #1065, opened 2019-08-22
Fix nested checkboxes js

andysellick:
##  What
Checkboxes javascript is broken - checking a parent of a nested group should check all the children, and checking a child should uncheck the parent:

Screen Shot 2019-08-22 at 14 27 40

## Why
`govuk-frontend` V3 introduced namespacing for component JS prefixed with `govuk-`. We have custom js that acts on that module call, but it wasn't updated.

## Visual Changes
No visual changes.

## View Changes
https://govuk-publishing-compo-pr-1065.herokuapp.com/checkboxes/checkbox_items_with_nested_checkboxes/

### DIFF
--- a/spec/javascripts/components/checkboxes-spec.js
+++ b/spec/javascripts/components/checkboxes-spec.js
@@ -3,12 +3,12 @@
 
 describe('Checkboxes component', function () {
   function loadCheckboxesComponent () {
-    var checkboxes = new GOVUK.Modules.Checkboxes()
+    var checkboxes = new GOVUK.Modules.GovukCheckboxes()
     checkboxes.start($('.gem-c-checkboxes'))
   }
 
   var FIXTURE =
-  '<div id="checkboxes-1ac8e5cf" class="gem-c-checkboxes govuk-form-group " data-module="checkboxes">' +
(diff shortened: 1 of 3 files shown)


spec/javascripts/components/checkboxes-spec.js
```diff
@@ -3,12 +3,12 @@
...
 describe('Checkboxes component', function () {
   function loadCheckboxesComponent () {
-    var checkboxes = new GOVUK.Modules.Checkboxes()
+    var checkboxes = new GOVUK.Modules.GovukCheckboxes()
     checkboxes.start($('.gem-c-checkboxes'))
   }
 
   var FIXTURE =
-  '<div id="checkboxes-1ac8e5cf" class="gem-c-checkboxes govuk-form-group " data-module="checkboxes">' +
+  '<div id="checkboxes-1ac8e5cf" class="gem-c-checkboxes govuk-form-group " data-module="govuk-checkboxes">' +
```
alex-ju: This means we won't support [conditional revealing](https://components.publishing.service.gov.uk/component-guide/checkboxes/checkbox_items_with_conditional_reveal), so we're fixing one example, but we're breaking another.

PR #1069, opened 2019-08-27
Remove govuk_frontend_toolkit sass dependencies

andysellick:
## What
@alex-ju i needed to do this for another reason, apologies if I'm duplicating work you've already done.

Remove the last sass dependencies from the gem, which were in the step by step component. Due to the slightly complex nature of this component I've had to write a custom font mixin, as the govuk-frontend mixins have undesired responsive font-sizing behaviour.

The gem still seems to require `govuk_frontend_toolkit` to initialise the JS properly, so I've not removed it fully.

Also improves the print styles for the step nav component.

## Why
We want to remove this dependency.

## Visual Changes
Changes should only affect the step by step navigation component. Visually should be identical, but there do seem to be some very minor differences, which I think is probably due to the font face change. Also had to choose a different grey for the horizontal line dividing steps, so it's now darker than before (I believe this is preferable to it being much lighter).

Before:

Screen Shot 2019-08-27 at 15 17 32

After:

Screen Shot 2019-08-27 at 15 17 39

## View Changes
https://govuk-publishing-compo-pr-1069.herokuapp.com/component-guide/step_by_step_nav

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
@@ -2,7 +2,7 @@ $stroke-width: 2px;
 $stroke-width-large: 3px;
 $number-circle-size: 26px;
 $number-circle-size-large: 35px;
-$top-border: solid 2px $grey-3;
+$top-border: solid 2px govuk-colour("mid-grey", $legacy: "grey-3");
 
 @mixin step-nav-vertical-line ($line-style: solid) {
   content: "";
@@ -10,8 +10,8 @@ $top-border: solid 2px $grey-3;
   z-index: 2;
   width: 0;
   height: 100%;
-  border-left: $line-style $stroke-width $grey-2;
-  background: $white;
+  border-left: $line-style $stroke-width govuk-colour("mid-grey", $legacy: "grey-2");
+  background: govuk-colour("white");
 }
 
 @mixin step-nav-line-position {
@@ -25,12 +25,27 @@ $top-border: solid 2px $grey-3;
   border-width: $stroke-width-large;
 }
 
+// custom mixin as govuk-font does undesirable things at different breakpoints
+// we want to ensure that both large and small step navs have the same size font on mobile
+// this will stop text resizing if compatibility mode is turned off
+@mixin step-nav-font($size, $tablet-size: $size, $weight: normal, $line-height: 1.3, $tablet-line-height: $line-height) {
+  @include govuk-typography-common();
+  font-size: $size + px;
+  font-weight: $weight;
+  line-height: $line-height;
+
+  @include govuk-media-query($from: tablet) {
+    font-size: $tablet-size + px;
+    line-height: $tablet-line-height;
+  }
+}
+
 .gem-c-step-nav {
-  margin-bottom: $gutter;
+  margin-bottom: govuk-spacing(6);
 
   &.gem-c-step-nav--large {
-    @include media(tablet) {
-      margin-bottom: $gutter * 2;
+    @include govuk-media-query($from: tablet) {
+      margin-bottom: govuk-spacing(9);
     }
   }
 
@@ -45,7 +60,7 @@ $top-border: solid 2px $grey-3;
 }
 
 .gem-c-step-nav__button {
-  color: $link-colour;
+  color: $govuk-link-colour;
   cursor: pointer;
   background: none;
   border: 0;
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -339,7 +348,7 @@ $top-border: solid 2px $grey-3;
   list-style: none;
 
   .gem-c-step-nav--large & {
-    @include media(tablet) {
+    @include govuk-media-query($from: tablet) {
       padding-bottom: 20px;
```
kr8n3r: govuk-spacing (4)

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
```diff
@@ -59,6 +58,9 @@ $stroke-width: 3px;
   background: $white;
```
vanitabarrett: `govuk-colour("white")`

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
```diff
@@ -114,4 +116,9 @@ $stroke-width: 3px;
   margin-bottom: .3em;
 }
 
+.gem-c-step-nav__circle-step-label,
```
vanitabarrett: Line 111: `padding-left: 30px` could be `padding-left: govuk-spacing(6)`

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
andysellick: There are loads of explicit px definitions in this file that could be replaced with the govuk-spacing mixin. I don't see the benefit in changing them all - the only benefit would be if the spacing scale changed, but then that would break the component.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -312,23 +321,23 @@ $top-border: solid 2px $grey-3;
 // contents of the steps, such as paragraphs and links
 
 .gem-c-step-nav__paragraph {
-  padding-bottom: $gutter-half;
+  padding-bottom: govuk-spacing(3);
   margin: 0;
   font-size: inherit;
 
   + .gem-c-step-nav__list {
     margin-top: -5px;
```
kr8n3r: - govuk-spacing(5)

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
kr8n3r:
```suggestion
      padding-bottom: govuk-spacing(4);
```

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
```diff
@@ -59,6 +58,9 @@ $stroke-width: 3px;
   background: $white;
   border-radius: 100px;
   text-align: center;
+  font-size: 19px;
```
kr8n3r: i guess we're not using the font mixin as we want consistent font-size across breakpoint?

36degrees: Note that by doing so, when compatibility mode is disabled this text won't be resizable (as unlike everything that uses the typography mixin, it won't be defined in rem). Might be something to worry about in the future though.

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
andysellick: Fixed, thanks.

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
andysellick: Correct, this is to ensure a consistent font size across breakpoints. I'll add a comment about the compatibility mode, as that's a good point, but something to worry about later.

andysellick:
@vanitabarrett @kr8n3r thanks for your reviews. There's a lot of places throughout these sass files (and beyond) where we could replace px values for govuk-spacing() values, but I'm not convinced of the value of that. 

The spacing scale in this context isn't going to change, and if it did it would have disastrous consequences for component layouts. So all we're doing is obfuscating the Sass slightly - I've been working with the scale for months and still have to look it up sometimes to remind myself of the numbers involved. And we'll always have px values in Sass files anyway, the scale can't account for everything, and we wouldn't want it to.

Thoughts?

nickcolley:
> There's a lot of places throughout these sass files (and beyond) where we could replace px values for govuk-spacing() values, but I'm not convinced of the value of that.

:+1:  to keep things moving any feedback on that can be raised as issues and done in future pull requests since the main thing you're trying to do here is remove the dependency not refactor all spacing.

PR #1071, opened 2019-08-29
Explicitly set focus states

maxgds:
## What
This explicitly sets the focus state for some links that otherwise don't inherit all of the necessary styles.

## Why
This is to support the work in government-frontend detailed here: https://trello.com/c/0OfIgpul/51-5-update-government-frontend-to-govukpublishingcomponents-v18-with-compatibility-mode-on

## Visual Changes
Screenshots are taken from examples of the component in use on gov.uk

### Before:
Screen Shot 2019-08-29 at 09 50 44

### After:
Screen Shot 2019-08-29 at 09 49 07

### Before:
Screen Shot 2019-08-29 at 09 48 21
### After:
Screen Shot 2019-08-29 at 09 49 29
### Before:
Screen Shot 2019-08-29 at 09 48 54
### After:
Screen Shot 2019-08-29 at 10 03 48

## View Changes
https://govuk-publishing-compo-pr-1071.herokuapp.com/

Note: The black colour is not coming through on some of these components and I'm still trying to track down why. This PR fixes the contents list including colour, and also fixes the underlines for the other components so seemed worth raising on the way to that full fix.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Restore margin bottom to the image card (PR #1079)
 * Allow summary_list to render without borders (PR #1073)
 * Remove govuk_frontend_toolkit sass dependencies (PR #1069)
+* Explicitly set focus states (PR #1071)
 
 ## 18.3.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -40,11 +40,6 @@
   list-style-type: none;
 }
 
-.gem-c-contents-list__link {
-  @include govuk-link-common;
-  @include govuk-link-style-default;
-}
-
 .gem-c-contents-list--no-underline {
   .gem-c-contents-list__link {
     text-decoration: none;
@@ -53,6 +48,8 @@
     &:active {
       text-decoration: underline;
     }
+
+    @include govuk-template-link-focus-override;
   }
 }
 
@@ -60,6 +57,8 @@
   .gem-c-contents-list__list-item--parent > & {
     font-weight: bold;
   }
+
+  @include govuk-template-link-focus-override;
 }
 
 .gem-c-contents-list__list-item {
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -10,6 +10,7 @@
 
 @import "components/helpers/variables";
 @import "components/helpers/brand-colours";
+@import "components/mixins/govuk-template-link-focus-override";
 @import "components/mixins/media-down";
 @import "components/mixins/margins";
 @import "components/mixins/clearfix";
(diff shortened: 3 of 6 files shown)


andysellick:
@maxgds bit confused by this PR... I'm looking at the live https://components.publishing.service.gov.uk/component-guide/share_links and the focus states already look like this:

Screen Shot 2019-08-29 at 17 21 38

Is it that these styles aren't coming through to government-frontend for some reason? Is that because it hasn't been upgraded to the latest version of the components gem yet, or some other CSS in the app is conflicting with it?

injms:
> @maxgds bit confused by this PR... I'm looking at the live https://components.publishing.service.gov.uk/component-guide/share_links and the focus states already look like this:
> 
> Screen Shot 2019-08-29 at 17 21 38
> 
> Is it that these styles aren't coming through to government-frontend for some reason? Is that because it hasn't been upgraded to the latest version of the components gem yet, or some other CSS in the app is conflicting with it?

These styles are fine in the components preview - however, in `government-frontend` (and probably other frontend apps as well) has a [style  from `govuk_template`](https://github.com/alphagov/govuk_template/blob/master/source/assets/stylesheets/_accessibility.scss#L63) that overpowers a single class selector.

The outcome is:

Screen Shot 2019-08-30 at 13 06 08


Adding the focus state explicitly should help overcome this.

andysellick:
@maxgds @injms ah that makes sense, thanks. In that case would it make more sense to put these overrides in `government-frontend` rather than the gem?

Also will upgrading `government-frontend` to V3 fix this, and if not, what will?

injms:
> In that case would it make more sense to put these overrides in `government-frontend` rather than the gem?

It also makes sense. I think this problem is with anything that has `govuk_template` as a dependency, so adding it here prevents duplication and makes it easier to remove when the time is right. (We could add a `TODO` note to this effect.)

> Also will upgrading `government-frontend` to V3 fix this, and if not, what will?

Once `govuk_template` is gone then this will be truly fixed - the upgrade to v3 in `government-frontend` won't fix this on it's own.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -53,13 +48,21 @@
     &:active {
       text-decoration: underline;
     }
+
+    &:focus {
+      @include govuk-focused-text;
```
andysellick: This all looks fine. If the focus state isn't quite right can I suggest creating a new mixin to use throughout the code in place of this one, something like:

```
@mixin link-focus-override {
  &:focus,
  &:active:focus,
  &:link:focus,
  &:visited:focus {
    @include govuk-focused-text;
    color: govuk-colour("black") !important;
    // plus anything else
  }
}
```

You may not need that exact combination of selectors, you'll need to try it out and see what works.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Explicitly set the focus state on links in some components (PR #1071)
```
andysellick: Nitpick: can this match the PR name?

maxgds:
@andysellick @injms Changes applied as suggested - I've made the mixin name very explicit so hopefully its existence doesn't get overlooked when we can remove it. Could you take another look?

PR #1074, opened 2019-08-29
Adjust share links column width

andysellick:
## What
Make the columns narrower in the 'columns' option for share links.

Before:

Screen Shot 2019-08-29 at 21 20 32

After:

Screen Shot 2019-08-29 at 21 20 42

## Why
This option is only used in one place, where the link text is much smaller and more realistic (e.g. 'Facebook' preceded by 'Share on' in visually hidden text) and this width works much better as the screen gets smaller.

## View Changes
https://govuk-publishing-compo-pr-1074.herokuapp.com/component-guide/share_links/

Trello card: https://trello.com/c/ky5WMr5L/181-look-into-share-links-layout-on-mobile

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Adjust share links column width (PR #1074)
+
 ## 18.3.0
 
 * `value` and `name` can be set on button component (PR #1059)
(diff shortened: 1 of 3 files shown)


DilwoarH:
Why don't we drop the email one too?

Just have `Email`

andysellick:
> Why don't we drop the email one too?
> 
> Just have `Email`

I want to show an example of what happens if the text is too long, so developers don't assume it'll auto expand to the right size.

PR #1076, opened 2019-08-30
Override edit link text on summary-link component

ChrisBAshton:
## What

- Adds option to override 'Edit' link text
- Fixes SASS configuration (looks like it was accidentally removed in #1069)
- Refactors fixture data to use YAML inheritance, as the number of examples are growing

## Why
We have a requirement to render a 'Reorder' link in place of an 'Edit' link.
This still makes sense in the context of the screen reader text.
'Edit' is now the default edit link text if none is specified, but can be
overridden with the new 'edit.link_text' property.

Trello: https://trello.com/c/UlgsMO0Q/45-overhaul-the-steps-section-and-move-the-add-step-button-to-it-and-show-it-in-the-summary-component

## Visual Changes

Fixed CSS import:

|Before|After|
|------|------|
|Screen Shot 2019-09-03 at 08 37 32|Screen Shot 2019-09-03 at 08 38 01|

Overwritten 'Edit' link text:

Screen Shot 2019-09-03 at 08 46 41

## View Changes
https://govuk-publishing-compo-pr-1076.herokuapp.com/

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/summary_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/summary_list.yml
@@ -41,6 +41,24 @@ examples:
         data_attributes:
           gtm: "edit-title-summary-body"
 
+  with_custom_link_on_section:
+    data:
+      edit:
+        href: "custom-link"
+        link_text: "Reorder"
+      title: "Items"
+      items:
+      - field: "Item 1"
+        value: "Value 2"
+      - field: "Item 2"
+        value: "Value 2"
+      - field: "Item 3"
+        value: "Value 3"
+    accessibility_criteria: |
+      Take care that the provided `link_text` still makes sense to screen readers when combined with the title.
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/summary_list.yml
```diff
@@ -53,6 +53,13 @@ examples:
       - field: "Body"
         value: "After the government decided in 2013 to expand the remit of the Committee to include public service providers, the Committee on Standards in Public Life produced our first report on the issue: Ethical Standards for Providers of Public Services in 2014."
 
+  with_custom_link_on_section:
+    data:
+      <<: *default-example-data
+      edit:
+        href: "custom-link"
+        link_text: "Reorder"
```
leenagupte: I'd consider using either different example link text or a different example in the summary list items, as it looks odd having a "Reorder" link for a "Title, summary and body".

ChrisBAshton: Done 👍

PR #1077, opened 2019-08-30
Adds inset-prompt component

ChrisBAshton:
## What

This was copied from the inset-prompt component in content-publisher:
https://content-publisher.integration.publishing.service.gov.uk/component-guide/inset_prompt

...but with the following modifications:

- CSS migrated to work with v18 of govuk_publishing_components
- Added tests
- Make component render nothing if no data passed

## Why

The component is now needed by collections-publisher, so it makes sense
to extract it out into the shared component library here.

Trello: https://trello.com/c/TcBpLRuC/

## View Changes

https://govuk-publishing-compo-pr-1077.herokuapp.com/component-guide/inset_prompt

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## unreleased
 
 * Allow summary_list to render without borders (PR #1073)
+* Adds inset-prompt component (PR #1077)
 
 ## 18.3.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -50,6 +50,7 @@
 @import "components/hint";
 @import "components/image-card";
 @import "components/input";
+@import "components/inset-prompt";
 @import "components/inset-text";
 @import "components/inverse-header";
 @import "components/label";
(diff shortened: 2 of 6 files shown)


kr8n3r:
could it be an extended [inset-text](https://components.publishing.service.gov.uk/component-guide/inset_text)?

andysellick:
@ChrisBAshton am I right in thinking that this component will only be used in non-public facing applications? If so, we don't want to add it to the gem as it'll add unnecessary page weight to GOV.UK. If it can be included in the specific application rather than the gem, that would be great.

There's an RFC open discussing this https://github.com/alphagov/govuk-rfcs/pull/110

ChrisBAshton:
> could it be an extended [inset-text](https://components.publishing.service.gov.uk/component-guide/inset_text)?

It probably could be @kr8n3r - not sure why it wasn't approached like that in content-publisher in the first place. I also think it could be less prescriptive if we did it like that - just make inset-text take a `do %> ... <% end %>` block to take any arbitrary contents.

> am I right in thinking that this component will only be used in non-public facing applications? 

Yes - good point. I think that's another good reason to turn this into a simple extension of inset_text!

ChrisBAshton:
Closing, as we've established this shouldn't be merged as-is. But would appreciate any comments as to next steps; either:

1. Moving this PR as-is to the collections-publisher repo, or
2. Editing the inset-text component to take an optional block, into which we can pass a collection of other components to manually recreate the contents of the inset-prompt component.

leenagupte:
Personally, I think the better option is to try and extend `inset-text` if possible, to avoid having almost identical components in multiple publishing apps.

ChrisBAshton:
Raised https://github.com/alphagov/govuk_publishing_components/pull/1078

PR #1078, opened 2019-08-30
Allow inset-text to take a block

ChrisBAshton:
## What

Can now pass an arbitrary block to the inset-text component to
have content rendered inset. If both 'text' and a block are
provided, 'text' takes precedence (mutually exclusive).

## Why

This is a compromise solution, rather than copying the inset-prompt
component from content-publisher (for use in collections-publisher),
for the reason that under current GOV.UK architecture the component
CSS would be added to the stylesheets for our end users (not just
the users using the publishing app). See:
https://github.com/alphagov/govuk_publishing_components/pull/1077

Once merged, we will use this latest inset-text component and
manually recreate what the inset-prompt component looks like by
passing components to the inset-text component as a block.

Trello: https://trello.com/c/TcBpLRuC/48-add-inset-prompt-component-to-govukpublishingcomponents

## Visual Changes

Screen Shot 2019-08-30 at 15 43 56

## View Changes
https://govuk-publishing-compo-pr-1078.herokuapp.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Remove govuk_frontend_toolkit sass dependencies (PR #1069)
 * Explicitly set focus states (PR #1071)
 * Override edit link text on summary-link component (#1076)
+* Allow inset-text to take a block (PR #1078)
 
 ## 18.3.1
 
--- a/app/views/govuk_publishing_components/components/_inset_text.html.erb
+++ b/app/views/govuk_publishing_components/components/_inset_text.html.erb
@@ -1,7 +1,10 @@
 <%
   id ||= "inset-text-#{SecureRandom.hex(4)}"
 %>
-
 <%= tag.div id: id, class: "gem-c-inset-text govuk-inset-text" do %>
-  <%= text %>
+  <% if defined? text %>
+    <%= text %>
+  <% elsif block_given? %>
+    <%= yield %>
+  <% end %>
 <% end %>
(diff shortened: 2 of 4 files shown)


andysellick:
We do have other components that accept a block like this, and the guidance around that is that the contents should be styled separately from the component, ie. the component takes no responsibility for what it is passed. We should at least document that in this component.

leenagupte:
> We do have other components that accept a block like this, and the guidance around that is that the contents should be styled separately from the component, ie. the component takes no responsibility for what it is passed. We should at least document that in this component.

Do you have an example?

andysellick:
There's a few, although they don't seem to document what I was talking about 😞 

- https://components.publishing.service.gov.uk/component-guide/details
- https://components.publishing.service.gov.uk/component-guide/inverse_header
- plus a few others, including the layout (although that's a special case in a sense)

ChrisBAshton:
> We do have other components that accept a block like this, and the guidance around that is that the contents should be styled separately from the component, ie. the component takes no responsibility for what it is passed. We should at least document that in this component.

Happy to add a clarifying note. Where should that be added, @andysellick - the `description` at the top of the YAML file? Or is there a way of adding a description inside the block example?

andysellick:
@ChrisBAshton https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/docs/step_by_step_nav.yml#L76

PR #1094, opened 2019-09-06
Ensure input styles are not overridden by static

andysellick:
## What
- static has some base styles for input[type="text"], input[type="search"] {} that alter the margin and padding of this component in ways we don't want
- this change overrides those specifically for this component only
- hopefully at some point in the future this can be removed

## Why
The input component in `frontend` looks like this, and this change fixes that.

Screen Shot 2019-09-06 at 10 32 28

## View Changes
https://govuk-publishing-compo-pr-1094.herokuapp.com/component-guide

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   of the commit log.
 
 ## Unreleased
+
+* Ensure input styles are not overridden by static ([PR #1094](https://github.com/alphagov/govuk_publishing_components/pull/1094))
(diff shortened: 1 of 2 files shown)


andysellick:
@nickcolley we get these problems every now and again and it's usually the classic problem of nice new BEM being overridden by old nested selector classes. I didn't want to bother you to add weight to govuk-frontend to fix things specifically in gov.uk.

nickcolley:
@andysellick I meant in this repository, how will people in the future remember to remove this code?

PR #1085, opened 2019-09-03
Fixes focus state spacing on 'extensive' related navigation links

injms:
## What
Increases the line height of the inline links and tweaks the toggle link in the ['extensive' related navigation](https://components.publishing.service.gov.uk/component-guide/related_navigation/with_extensive_world_locations) component.

Partially fixes #1028; and probably shouldn't be merged until #1054 has been merged.

## Why

The overlap of the focus state made it look broken and having the focus state of the toggle link stretch across the entire parent element looks slightly odd, as no other link had that behaviour.

## Visual Changes
Toggle link before:
Screen Shot 2019-09-03 at 14 54 34

Toggle link after:
Screen Shot 2019-09-03 at 14 53 51

Inline links before:
Screen Shot 2019-09-03 at 14 54 20

Inline links after:
Screen Shot 2019-09-03 at 14 53 58

PR #1084, opened 2019-09-03
Add chevron banner component

vanitabarrett:
## What
Add chevron banner component for use on branded EU Exit pages only.

## Why
We are currently using a chevron banner on both the homepage (Frontend) and the landing page (Collections). As it's not a component, we've implemented the chevron twice (once in each app) and there are small differences between each one as a result. It also means if we need to make changes to the banner, we need to make them twice. This tidies up the code into one common place.

## Visual Changes
Screen Shot 2019-09-04 at 10 28 43

https://govuk-publishing-compo-pr-1084.herokuapp.com/component-guide/chevron_banner

PR #1095, opened 2019-09-06
Version 20.1.1

andysellick:
* Ensure input styles are not overridden by static ([PR #1094](https://github.com/alphagov/govuk_publishing_components/pull/1094))
* Fixes focus state spacing on 'extensive' related navigation links ([PR #1085](https://github.com/alphagov/govuk_publishing_components/pull/1085))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@andysellick Could you take another look at https://github.com/alphagov/govuk_publishing_components/pull/1084 ? It would be good if we could get that into this version too.

PR #1098, opened 2019-09-06
Chevron banner tweaks

vanitabarrett:
## What
Adds the following fixes:

- Reduces the padding in the chevron banner so the chevron is less distorted on smaller screens
- Fixes the hover state so the background image is only replaced mobile and above

## Visual Changes
**Improve layout on smaller screens (before)**
Screen Shot 2019-09-06 at 13 06 19 

**Improve layout on smaller screens (after)**
Screen Shot 2019-09-06 at 13 07 41

**Hover state below 320px (before)**
Screen Shot 2019-09-06 at 13 05 59

**Hover state below 320px (after)**
Screen Shot 2019-09-06 at 13 11 21

https://govuk-publishing-compo-pr-1098.herokuapp.com/component-guide/chevron_banner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Improve chevron banner component ([PR #1098](https://github.com/alphagov/govuk_publishing_components/pull/1098))
+
 ## 20.2.0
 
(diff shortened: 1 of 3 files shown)


vanitabarrett:
@andysellick Ah, I think that might be a result of one of these changes. Will fix now 👍

vanitabarrett:
@andysellick That should be fixed now - it looks like the compressor I was using was removing some properties from the SVG which caused the stroke-width to be smaller than it should be.

andysellick:
@vanitabarrett looks great!

PR #1102, opened 2019-09-06
Add lang attribute to heading component

vanitabarrett:
Screen Shot 2019-09-06 at 15 59 24

## What
Add the ability to pass a language into the heading component, which is added as a `lang` attribute.

## Why
In some apps we are passing in a translated string which is falling back to the default (normally English) because the translation does not exist in the chosen language. In these cases, we want to mark these strings as not being in the chosen language.

https://govuk-publishing-compo-pr-1102.herokuapp.com/

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/heading.yml
+++ b/app/views/govuk_publishing_components/components/docs/heading.yml
@@ -68,3 +68,11 @@ examples:
       brand: 'department-for-environment-food-rural-affairs'
       padding: true
       border_top: 5
+  with_lang_attribute:
+    description: |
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/heading.yml
```diff
@@ -68,3 +68,7 @@ examples:
       brand: 'department-for-environment-food-rural-affairs'
       padding: true
       border_top: 5
+  with_lang_attribute:
+    data:
```
injms: Can we note in the description for this that the `lang` attribute needs to be a valid BCP47 string? I've linked to https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#Language_tag_syntax before.

injms:
Aaaand I completely missed that this was a draft. Sorry!

PR #1104, opened 2019-09-06
Version 20.2.2

alex-ju:
## 20.2.2

* Fix data attributes in delete summary list item ([PR #1103](https://github.com/alphagov/govuk_publishing_components/pull/1103))

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (20.2.1)
+    govuk_publishing_components (20.2.2)
       gds-api-adapters
       govuk_app_config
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -321,5 +321,8 @@ DEPENDENCIES
   webmock (~> 3.6.0)
   yard
 
+RUBY VERSION
+   ruby 2.6.3p62
```
injms: Should this different to the `2.6.3` in `.ruby-version` file?

alex-ju: It should be `2.6.3`, but patches are allowed it seems.

injms: Interesting, thanks!

PR #1107, opened 2019-09-09
Allow custom heading size on checkboxes and radios

alex-ju:
## What
Add support for custom heading size on checkboxes and radios when not being used as page heading.

## Why
Because we currently hardcode the `--m` size for legends and we need to support `--s` size too for when the components are being used on a page with multiple input groups.

When using checkboxes/radios on a page with more than one form group the headings should match the styling of the labels with the rest of the inputs on the page (which can be achieved by setting an `--s` size to the heading).

## Visual Changes
Added examples with custom heading size for checkboxes and radio component

## View Changes
https://govuk-publishing-compo-pr-1107.herokuapp.com/component-guide/checkboxes/with_custom_heading_size

https://govuk-publishing-compo-pr-1107.herokuapp.com/component-guide/radio/with_custom_heading_size

[Trello card](https://trello.com/c/ZB4lH4G8)

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -73,7 +73,7 @@ examples:
       heading: "Choose your favourite skittles"
       description: |
         Skittles consist of hard sugar shells imprinted with the letter "S".
-        The interior consists mainly of sugar, corn syrup, and hydrogenated 
+        The interior consists mainly of sugar, corn syrup, and hydrogenated
         palm kernel oil along with fruit juice, citric acid, natural and artificial flavors.
       hint_text: "Taste the rainbow"
       items:
@@ -94,7 +94,7 @@ examples:
       is_page_heading: true
       description: |
         Skittles consist of hard sugar shells imprinted with the letter "S".
-        The interior consists mainly of sugar, corn syrup, and hydrogenated 
+        The interior consists mainly of sugar, corn syrup, and hydrogenated
         palm kernel oil along with fruit juice, citric acid, natural and artificial flavors.
       hint_text: "Taste the rainbow"
       items:
@@ -157,6 +157,18 @@ examples:
       name: "favourite_colour"
       heading: "What is your favourite colour?"
       is_page_heading: true
+      items:
+        - label: "Red"
+          value: "red"
+        - label: "Green"
+          value: "green"
+        - label: "Blue"
+          value: "blue"
+  with_custom_heading_size:
+    data:
+      name: "favourite_colour"
+      heading: "What is your favourite colour?"
+      heading_size: "s"
       items:
         - label: "Red"
           value: "red"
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -164,6 +164,18 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  with_custom_heading_size:
+    data:
+      name: "favourite_colour"
+      heading: "What is your favourite colour?"
+      heading_size: "s"
```
leenagupte: Perhaps this should be "small" rather than "s"?

alex-ju: I do get it can be confusing but the classes are defined in govuk-frontend/Design System, so we have no control of that and I tried to avoid having an associated map based on which to generate these modifiers.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -168,6 +168,19 @@ examples:
         text: "Green"
       - value: "blue"
         text: "Blue"
+  with_custom_heading_size:
```
leenagupte: What's changed here?

spec/components/radio_spec.rb
```diff
@@ -119,6 +119,20 @@ def component_name
     assert_select "legend h1", "What is your favourite skittle?"
   end
 
+  it "renders radio-group with custom heading size" do
+    render_component(
+      name: "favourite-skittle",
+      heading: "What is your favourite skittle?",
+      heading_size: "s",
```
leenagupte: What happens if an invalid value is passed in? e.g. "q"?

alex-ju: [Will default to 'm'](https://github.com/alphagov/govuk_publishing_components/pull/1107/files#diff-fe62b3a78b7005dd4f069f961652b72cR10) which is the hard-coded value at the moment.

app/views/govuk_publishing_components/components/docs/radio.yml
alex-ju: Defined a new example for the custom `heading_size` set to "s"

spec/components/radio_spec.rb
andysellick: @alex-ju are you sure?

alex-ju: Not with the pushed code, no. Will update it to use an associated hash, full names for heading sizes and handle the default properly.

andysellick: Just tested it and it adds to the class whatever you pass. You could try something like:

```
css_classes << "govuk-fieldset__legend--#{heading_size}" if ['s', 'm', 'l', 'xl'].include?(heading_size)
```

alex-ju: Yep, that would be consistent with the approach we use for margins.

alex-ju: Updated to default to 'm' when both no size is being passed in or the value that's being passed in is not a pre-defined value. Also added tests for these.

PR #1109, opened 2019-09-09
Fix focus states on cookie-banner confirmation message

### LINKED ISSUE #1014: Cookie banner focus is unstyled
I'm not sure if this is intentional but raising it just incase it is not.

When submitting the cookie banner it manages focus to the banner but the focus outline is using a default style.

I think... because you're setting focus to have this message read out, and it's not focussable you could hide the outline here...?

```css
.gem-c-cookie-banner__confirmation:focus {
    outline: none;
}
```

Or maybe it should have a consistent outline, not sure :P

alex-ju:
## What
Remove the outline when the element is focused to be read out by screen readers and update the focus state on the hide confirmation message button.

## Why
To fix #1014

### Before (confirmation message)
Screen Shot 2019-09-09 at 15 39 16

### After (confirmation message)
Screen Shot 2019-09-09 at 15 38 19

### Before (hide message button) - no, it's not a mistake, it really looks like the screen capture above
Screen Shot 2019-09-09 at 15 38 19

### After (hide message button)
Screen Shot 2019-09-09 at 15 38 32

### View Changes
https://govuk-publishing-compo-pr-1109.herokuapp.com/component-guide/cookie_banner

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
@@ -69,6 +69,13 @@ $govuk-cookie-banner-text-green: #00823b;
   display: none;
   position: relative;
   padding: govuk-spacing(1);
+
+
+  // This element is focused using JavaScript so that it's being read out by screen readers
+  // for this reason we don't want to show the default outline or emphasise it visually using `govuk-focused-text`
+  &:focus {
+    outline: none;
+  }
 }
 
 .gem-c-cookie-banner__confirmation-message,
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -69,6 +69,10 @@ $govuk-cookie-banner-text-green: #00823b;
   display: none;
   position: relative;
   padding: govuk-spacing(1);
+
+  &:focus {
+    outline: none;
```
nickcolley: Worth a comment explaining why this is OK to do for future travellers.

alex-ju: Explained in the commit message, but maybe not everyone is a git history fan, so I've added a comment in the file too as suggested.

nickcolley:
Thanks for doing this Alex 👍

PR #1110, opened 2019-09-09
Remove dependency on GOV.UK Frontend Toolkit 🎉

### LINKED ISSUE #500: This depends on deprecated govuk_frontend_toolkit
We hit some issues using this gem working with content-publisher: https://github.com/alphagov/content-publisher/pull/217 in a javascript context which highlighted some of the problems with the remaining dependency on govuk_frontend_toolkit.

A problem we have for any apps is that nearly all the modules in https://github.com/alphagov/govuk_publishing_components/tree/master/app/assets/javascripts/govuk_publishing_components need to use [govuk_frontend_toolkit](https://github.com/alphagov/govuk_frontend_toolkit) to initialise their modules. This presents a problem for developers trying to migrate away from jQuery as govuk_frontend_toolkit requires jQuery.

A solution to this would be to change the initialisation method for our components to use a different approach - as the current one we have is only a very thin layer. We'd have to decide whether we want to continue with a self initialisation approach or to take the approach from [govuk-frontend](https://github.com/alphagov/govuk-frontend) to have them be initialised separately. 

We also seem to use govuk_frontend_toolkit for various CSS files. I'm not sure if that's now covered by govuk-frontend, @alex-ju ?

So actions to resolve this would be:

- Determine a new way to initialise javascript components that doesn't use govuk_frontend_toolkit
- See if we can replace the stylesheets that are used

I'm not sure if I'm missing any others

/cc @alex-ju

alex-ju:
## What
Move `modules.js` from `govuk_frontend_toolkit` into `govuk_publishing_components` (without changes - for now) and remove dependency on `govuk_frontend_toolkit`.

It's been a bit over an year since we started to slowly migrate from `govuk_frontend_toolkit` to `govuk-frontend` and this is the last final bit. The migration was a collective effort, so thanks a lot to everyone who contributed to this and made it possible! 🙌 

## Why
[Remove dependency on deprecated projects](https://trello.com/c/ajPzDAOX)
Fixes #500

## Visual Changes
No visual changes.

## Notes
This is not a breaking change, nor a feature - more of an internal change.

~~The tests for this seem to be flaky, so I'd like to have a look at them and try to fix them up before merging. Will do this in a separate commit to ease the review.~~ Fixed the `FilterComponents` spec - no flaky tests anymore.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -1,0 +1,61 @@
+;(function (global) {
+  'use strict'
+
+  var $ = global.jQuery
+  var GOVUK = global.GOVUK || {}
+  GOVUK.Modules = GOVUK.Modules || {}
+
+  GOVUK.modules = {
+    find: function (container) {
+      container = container || $('body')
+
+      var modules
+      var moduleSelector = '[data-module]'
+
+      modules = container.find(moduleSelector)
+
+      // Container could be a module too
+      if (container.is(moduleSelector)) {
+        modules = modules.add(container)
+      }
+
+      return modules
+    },
+
+    start: function (container) {
+      var modules = this.find(container)
+
+      for (var i = 0, l = modules.length; i < l; i++) {
+        var module
+        var element = $(modules[i])
+        var type = camelCaseAndCapitalise(element.data('module'))
+        var started = element.data('module-started')
+
+        if (typeof GOVUK.Modules[type] === 'function' && !started) {
+          module = new GOVUK.Modules[type]()
+          module.start(element)
+          element.data('module-started', true)
+        }
+      }
+
+      // eg selectable-table to SelectableTable
+      function camelCaseAndCapitalise (string) {
+        return capitaliseFirstLetter(camelCase(string))
+      }
+
+      // http://stackoverflow.com/questions/6660977/convert-hyphens-to-camel-case-camelcase
+      function camelCase (string) {
+        return string.replace(/-([a-z])/g, function (g) {
+          return g.charAt(1).toUpperCase()
+        })
+      }
+
+      // http://stackoverflow.com/questions/1026069/capitalize-the-first-letter-of-string-in-javascript
+      function capitaliseFirstLetter (string) {
+        return string.charAt(0).toUpperCase() + string.slice(1)
+      }
+    }
+  }
+
+  global.GOVUK = GOVUK
+})(window)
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/govuk_publishing_components/modules.js
```diff
@@ -0,0 +1,61 @@
+;(function (global) {
```
andysellick: Were the files moved from the toolkit copied unmodified? Just want to check how thoroughly this needs to be reviewed.

spec/javascripts/govuk_publishing_components/modules.spec.js
```diff
@@ -0,0 +1,95 @@
+/* eslint-env jasmine, jquery */
+
+var $ = window.jQuery
+
+describe('GOVUK Modules', function () {
```
andysellick: Was there any documentation accompanying the modules code? If not, should we write some? Feels like a good opportunity (although probably in a separate PR).

app/assets/javascripts/govuk_publishing_components/modules.js
alex-ju: I copied them (both the file and the spec) 'as-is', for now, with the intention to update them in a future iteration (I'd really love to remove the jQuery dependency out of it and potentially improve it).

spec/javascripts/govuk_publishing_components/modules.spec.js
alex-ju: Good point. There's the [Modules section](https://github.com/alphagov/govuk_frontend_toolkit/blob/master/docs/javascript.md#modules) in the JavaScript doc. I'll extract it and add it to this repo.

alex-ju:
> Most of the apps will still pull this code in from the toolkit,
Yes, there are a couple with direct dependency on govuk_frontend_toolkit. Hope we'll get rid of those too rather soon.

> so until we remove/modify those dependencies it'll run twice
Nope, the migrated `modules.js` will only run once since they're part of `dependencies.js` which was introduced for and imported by projects that don't use `static` (e.g. Content Publisher, Content Data, Collection Publisher). All the public facing applications initialise modules via static. We should work to converge them at some point, but until then this helps moving the script to `govuk_publishing_components` and use with the new layouts (admin_layout and the new public-facing layout @vanitabarrett started the work on).

> we should probably test if that happens safely.
Test in progress with various apps. Will update the PR comment with a checklist.

nickcolley:
Just a note: one other alternative to removing this code would be to manually initialise components similar to what we do in GOV.UK Frontend.

PR #1112, opened 2019-09-11
Refactor contents list component

vanitabarrett:
## What
First commit: refactors the contents list component to move more logic out of the view and into the helper. This makes it easier for us to add another option to the component in the second commit.

Second commit: Adds an option to the component to set a font-size other than the default 16px.

## Why
On the new [Brexit landing page](https://www.gov.uk/brexit), we want the contents list to match the font-size of the surrounding text, which is 19px. At the moment, we are overriding the `gem-c-contents-list__link` class within Collections, which isn't ideal. This adds the option to the component itself, so we can remove the override.

## Visual Changes
Screen Shot 2019-09-11 at 12 06 09

https://govuk-publishing-compo-pr-1112.herokuapp.com/component-guide/contents_list

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -10,6 +10,16 @@
   box-shadow: 0 20px 15px -10px govuk-colour("white");
 }
 
+.gem-c-contents-list--font-size-19 .gem-c-contents-list__title,
+.gem-c-contents-list--font-size-19 .gem-c-contents-list__list {
+  @include govuk-font($size: 19, $line-height: 1.5);
+}
+
+.gem-c-contents-list--font-size-24 .gem-c-contents-list__title,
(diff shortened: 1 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -10,6 +10,16 @@
...
 }
 
+.gem-c-contents-list--font-size-19 .gem-c-contents-list__title,
+.gem-c-contents-list--font-size-19 .gem-c-contents-list__list {
+  @include govuk-font($size: 19, $weight: regular, $line-height: 1.5);
+}
+
+.gem-c-contents-list--font-size-24 .gem-c-contents-list__title,
+.gem-c-contents-list--font-size-24 .gem-c-contents-list__list {
+  @include govuk-font($size: 24, $weight: regular, $line-height: 1.5);
```
andysellick: Is `$weight` needed here, and on line 15?

PR #1117, opened 2019-09-13
Version 20.5.1

vanitabarrett:
Includes:

* Fix IE-specific CSS for chevron banner ([PR #1116](https://github.com/alphagov/govuk_publishing_components/pull/1116))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 20.5.1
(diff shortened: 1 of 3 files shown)


cbaines:
Doesn't this version also include https://github.com/alphagov/govuk_publishing_components/pull/1110 ?

vanitabarrett:
@cbaines Hm, yes, potentially. Looks like that change wasn't added to the Changelog, but that may be because Alex considers it an "internal" gem change which doesn't really need to be added to the changelog.

PR #1118, opened 2019-09-13
Fix changelog heading for v 20.5.0

alex-ju:
## What
Fix changelog heading for version 20.5.0. Replacing a non-standard spacing character with the standard one.

## Why
Because it doesn't allow the heading to be rendered correctly


vanitabarrett:
🤔 What's the difference?

alex-ju:
@vanitabarrett it's replacing a non-standard spacing character with the standard one

PR #1127, opened 2019-09-19
Generate FAQ schema from page body

sihugh:
## What

At present, we're presenting a whole chapter of a guide in a single question/answer. This limits the impact of the content design we apply, because it doesn't break up the content in rich results in the same way as we do on GOV.UK.  

We're also presenting all the chapters of the guide as structured data on the first chapter of the guide. This might run us foul of external search indexers because not all of the text is visible on the first page.

This PR moves to only generating structured markup for the chapter on which it's displayed.  This allows us to split up the page into more bite-sized sections.  We parse the govspeak generated HTML and uses the text in h2 headings as the "question", the id of the h2 to generate URLs with anchors, and the other markup as the "answer".

This enables us to display structured markup on on every chapter in a guide and also on other pages (such as mainstream answers).  It also means that the Google results clash less with step by steps because it's more obvious that this is guidance, and not a "how to".

We will still avoid using this markup on travel advice pages because of the latency in results being updated.

I briefly disabled the  `Style/IfInsideElse` cop because the alternatives seemed less appealing.

Example usage can be seen on the pages linked in https://github.com/alphagov/government-frontend/pull/1492

## Visual Changes 

The rich Google result for default page in https://www.gov.uk/apply-to-come-to-the-uk:

### Before (one rich result for a whole guide)
Screenshot 2019-09-19 09 39 46

### After (one rich result per page in the guide)
Screenshot 2019-09-19 09 46 07

## View Changes
https://govuk-publishing-compo-pr-1127.herokuapp.com/component-guide/machine_readable_metadata/a_guide


lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb
```diff
@@ -19,36 +19,70 @@ def structured_data
...
+        doc.xpath("html/body").children.each_with_object({}) do |element, q_and_as|
+          if question_element?(element)
+            question = element.text
+            q_and_as[question] = { anchor: element["id"] }
+          elsif q_and_as[question].nil?
+            q_and_as[question] = { answer: element.to_s }
+          elsif q_and_as[question][:answer].nil?
+            q_and_as[question][:answer] = element.to_s
+          else
+            q_and_as[question][:answer] << element.to_s
```
bilbof: You may get a NoMethodError here here if `q_and_as` is still an empty hash.

sihugh:
Going to hold off merging for a couple of days while we measure the last change

PR #1128, opened 2019-09-19
Update search component styles

### LINKED ISSUE #1027: Search box component needs to look like GOVUK Frontend
The search box component is currently entirely custom built and doesn't use any GOVUK Frontend styles. The result of this is that with the upcoming V3 update, it won't change to be in line with the rest of the site. This is particularly noticeable for the input part of the component.

Current search component (focussed)

Screen Shot 2019-08-08 at 07 37 32

V3 input element (focussed)

Screen Shot 2019-08-08 at 07 37 57

This issue can wait until the [V3 PR](https://github.com/alphagov/govuk_publishing_components/pull/1010) has been merged.

kr8n3r:
## What
- Updates border colour and style to match inputs
- Updates focus states to match inputs
- Updates label colour when label is set to be outside in input or when javascript is disabled so it matches label component colour
- general reordering and consolidation of styles

## Why
To update to latest styles, matching other components.
Fixes #1027

## Visual Changes
<details>
<summary>Default states before and after</summary>

**Before:**
components publishing service gov uk_component-guide_search

**After:**
0 0 0 0_3212_component-guide_search
</details>
<details>
<summary>Focus states before and after</summary>

**Before:**
Screen Shot 2019-09-19 at 15 48 32

**After:**
Screen Shot 2019-09-19 at 15 44 55

</details>

## View Changes
https://govuk-publishing-compo-pr-1128.herokuapp.com/component-guide/search

## Further improvements
an additional option is to completely modernise it and use an inline svg for the icon and add aria-label="Search" on the button to describe it.
Fallback for SVG would be an `image` tag within, like https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/header/template.njk#L34

then we can use the button background yellow focus colour and change the icon colour accordingly

thumbs or down to add this to the PR

Trello ticket: https://trello.com/c/xvVfdgAc/1024-update-search-component-styles

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -7,11 +7,26 @@ $large-input-size: 50px;
 }
 
 .gem-c-search__label {
+  @include govuk-font($size: 19, $line-height: $input-size);
   display: block;
 
   h1 {
     @include govuk-font($size: 19, $line-height: $input-size);
     margin: 0;
+  }
+
+  .js-enabled & {
+    position: absolute;
+    left: govuk-spacing(3);
+    top: 1px;
+    z-index: 1;
+    color: $govuk-secondary-text-colour;
+  }
+
+  // match label colour with the label component colour
+  // when javascript is enabled and inline_label option  is set to false
+  .js-enabled .gem-c-search--separate-label & {
+    color: $govuk-text-colour;
   }
 }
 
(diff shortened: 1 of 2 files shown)


kr8n3r:
an additional option is to completely modernise it and use an inline svg for the icon and add aria-label="Search" on the button to describe it.
Fallback for SVG would be an `image` tag within, like https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/header/template.njk#L34

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -13,49 +13,99 @@ $large-input-size: 50px;
     @include govuk-font($size: 19, $line-height: $input-size);
     margin: 0;
   }
+
+  .js-enabled & {
+    @include govuk-font($size: 19, $line-height: $input-size);
```
andysellick: I know this wasn't part of your change but I think this line should be moved outside of the `.js-enabled` section here, as we want the font to be correct even if JS isn't enabled.

PR #1129, opened 2019-09-19
Bump to 21.0.0

sihugh:
This is a breaking change for guides in government-frontend.  The FAQ schema is not used anywhere else yet.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 21.0.0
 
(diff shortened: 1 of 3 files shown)


andysellick:
@sihugh so this is a breaking change, but only for `government-frontend`? All other apps are safe to upgrade?

sihugh:
@andysellick yes, that's right.  I'm not sure if you're being sarcastic or if I just wasn't clear 🙈

andysellick:
@sihugh no, sorry, just wanted to be clear before I hit the approve button - if it was a breaking change for everything then we'd need to immediately start fighting dependabot battles :)

PR #1130, opened 2019-09-23
Add name option to select component

andysellick:
## What
Adds an option to the select component to add a 'name' attribute that is different from the 'id' attribute. If none is passed, 'name' defaults to 'id' (as before this change was made).

## Why
Smart answers is being converted to use components and the selects for dates have differing 'id' and 'name' attributes.

## Visual Changes
No visual changes.

## View Changes
https://govuk-publishing-compo-pr-1130.herokuapp.com/component-guide/select

### DIFF
--- a/app/views/govuk_publishing_components/components/_select.html.erb
+++ b/app/views/govuk_publishing_components/components/_select.html.erb
@@ -3,6 +3,7 @@
   id ||= false
   label ||= false
   full_width ||= false
+  name ||= id
 
   select_helper = GovukPublishingComponents::Presenters::SelectHelper.new(options)
   data_module = "data-module=track-select-change" unless select_helper.data_tracking?.eql?(false)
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_select.html.erb
```diff
@@ -3,6 +3,7 @@
   id ||= false
   label ||= false
   full_width ||= false
+  name_attr ||= id
```
alex-ju: I think we use `name` for this for most of the components

injms: If the name defaults to the `id`, what happens to the name when the `id` defaults to false?

andysellick: Oh, do we? I typed in `name` and my editor highlighted it, so I assumed it was reserved.

injms:
Aside from Alex's comment, looks good to me 👍

app/views/govuk_publishing_components/components/_select.html.erb
andysellick: Ah yes that works. Will change.

PR #1132, opened 2019-09-23
Bump to version 21.1.0

kr8n3r:


lib/govuk_publishing_components/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukPublishingComponents
-  VERSION = '21.0.0'.freeze
+  VERSION = '21.0.1'.freeze
```
andysellick: You mean `21.1.0`?

kr8n3r: 👍

PR #1133, opened 2019-09-23
Prevent sticky / fixed elements from overlapping feedback component

injms:
## What

Sets the feedback component to be `position: relative` so sticky or fixed elements that precede the feedback component don't overlap it; and sets the feedback element container's background to be white to stop any underlapping elements from showing through.

## Why

A sticky or fixed element was overlapping the feedback component on Service Manual Frontend and this would fix it.

Problem can be seen on the [Agile and government services: an introduction](https://www.gov.uk/service-manual/agile-delivery/agile-government-services-introduction) page in the Service Manual. Scroll to the feedback component and open either one of the feedback forms by clicking 'No' or 'Is there anything wrong with this page?'. Once the feedback form has opened, scroll to the bottom of the page to see the sticky menu will overlap with the feedback form.

## Visual Changes

Before:
Screen Shot 2019-09-23 at 16 58 44 1

After:
Screen Shot 2019-09-23 at 16 58 32

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -1,6 +1,8 @@
 .gem-c-feedback {
-  max-width: $govuk-page-width;
+  background: govuk-colour("white");
   margin: 0 auto;
+  max-width: $govuk-page-width;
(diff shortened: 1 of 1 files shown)


andysellick:
@injms can you link to the page where this is a problem?

PR #1135, opened 2019-09-24
Add "Button looks like a link" variant

kr8n3r:
## What
Add button that looks like a link variant of a button 

## Why
The toggle to expand content, such as [show updates](https://www.gov.uk/government/publications/sample-accessible-document-policy#history) are links but should be buttons, like[ "Show more search options" ](https://www.gov.uk/search/all) which we updated thanks to some [prodding](https://github.com/alphagov/finder-frontend/issues/1323) 

## View Changes
https://govuk-publishing-compo-pr-1135.herokuapp.com/component-guide/button/button_looks_like_link/preview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add "Button looks like a link" variant [#PR 1135](https://github.com/alphagov/govuk_publishing_components/pull/1135)
+
 ## 21.2.0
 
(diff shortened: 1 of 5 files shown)


36degrees:
I'd consider making this a component in its own right, rather than a modifier on a button. Given you don't really want any of the styling that comes from the button component, and this is having to effectively 'unset' the majority of the button styling, it seems likely that future 'non-breaking' changes to the button component could break this.

kr8n3r:
ok will close this for now until I have time to work on the new component

PR #1143, opened 2019-09-26
Add select component error state without message

andysellick:
## What
Allows the select component to appear in an error state and accept a given id for an error element (used for aria-describedby) but without an error message appearing. 

This PR also updates the doc and tests for the error message component - I needed to be able to pass an ID to it, it turns out you can already do that but it wasn't documented.

This PR also adds an error state to the fieldset component.

## Why
This is intended for use where the select component error message should be rendered separately from the component, which is needed in the [smart answers upgrade](https://github.com/alphagov/smart-answers/pull/4132).

## Visual Changes
This is the select component in the new state:

Screen Shot 2019-09-26 at 16 12 59

This the fieldset error state:

Screen Shot 2019-09-26 at 16 06 27

This is how all of this combined will look in smart answers:

Screen Shot 2019-09-26 at 15 32 31

This change brings the component markup slightly more in line with the [markup from govuk-frontend](https://design-system.service.gov.uk/components/date-input/#error-messages), which wraps the fieldset element in a div. This has had the consequence of giving the component a bottom margin, but since this is a change from govuk-frontend I'm assuming it's okay.

## View Changes
https://govuk-publishing-compo-pr-1143.herokuapp.com/component-guide

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/fieldset.yml
+++ b/app/views/govuk_publishing_components/components/docs/fieldset.yml
@@ -12,10 +12,10 @@ examples:
       legend_text: 'Do you have a passport?'
       text: |
         <!-- Use the radio component, this is hardcoded only for this example -->
-        <input type="radio" id="default-yes" name="default"t>
+        <input type="radio" id="default-yes" name="default">
         <label for="default-yes">Yes</label>
 
-        <input type="radio" id="default-no" name="default"t>
+        <input type="radio" id="default-no" name="default">
         <label for="default-no">No</label>
   with_custom_legend_size:
     description: Make the legend different sizes. Valid options are 's', 'm', 'l' and 'xl'.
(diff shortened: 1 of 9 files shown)


alex-ju:
Sorry, haven't done a proper review on this, but the date component should look like [the one in the Design System](https://design-system.service.gov.uk/components/date-input/#error-messages) which potentially involves some updates in the way we wrap different elements.

andysellick:
Thanks @alex-ju , will have a go at adjusting that.

andysellick:
@alex-ju ready for re-review, I've added another commit.

app/views/govuk_publishing_components/components/docs/fieldset.yml
```diff
@@ -12,10 +12,10 @@ examples:
       legend_text: 'Do you have a passport?'
       text: |
         <!-- Use the radio component, this is hardcoded only for this example -->
-        <input type="radio" id="default-yes" name="default"t>
```
alex-ju: 🤦‍♂  We should update these examples over time.

andysellick: Yeah, I know! I did spend a moment wondering if that was deliberate and some rails thing I didn't know about...

PR #1146, opened 2019-10-02
Upgrade Ruby to 2.6.5

thomasleese:
Upgrade Ruby to 2.6.5, see commits for more details.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Upgrade Ruby version to 2.6.5. ([PR #1146](https://github.com/alphagov/govuk_publishing_components/pull/1146))
```
cbaines: Is this a change for the CHANGELOG? What effect does it have on consumers of the gem?

thomasleese: Perhaps not, I don't think it does have any effect on the consumers. I'll remove this commit.

PR #1152, opened 2019-10-04
Replace regex with similar method from Rails

sihugh:
https://apidock.com/rails/ActiveSupport/Inflector/parameterize

We want to be able to generate links to steps from other places on the site. Using the rails builtin methods for doing this seems like an easy way to converge - even if one of the existing tests does end up looking a little weird. (I'm not sure that 'Step 3?"`_!' is a piece of text that we're likely to see for real.)

### DIFF
--- a/spec/components/step_by_step_nav_spec.rb
+++ b/spec/components/step_by_step_nav_spec.rb
@@ -165,7 +165,7 @@ describe "step nav", type: :view do
 
     assert_select step1 + " #step-panel-step-1-1.gem-c-step-nav__panel"
     assert_select step1and + " #step-panel-step-1-and-2.gem-c-step-nav__panel"
-    assert_select step3 + "#step-3"
+    assert_select step3 + "#step-3-_"
   end
 
   it "renders correct list elements and includes length of lists" do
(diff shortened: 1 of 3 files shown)


spec/components/step_by_step_nav_spec.rb
```diff
@@ -165,7 +165,7 @@ def stepnav
 
     assert_select step1 + " #step-panel-step-1-1.gem-c-step-nav__panel"
     assert_select step1and + " #step-panel-step-1-and-2.gem-c-step-nav__panel"
-    assert_select step3 + "#step-3"
+    assert_select step3 + "#step-3-_"
```
andysellick: Not sure I understand this change, can you clarify?

sihugh: This is where the parameterize method differs.  It won't replace underscores, so we end up with the original underscore and a hyphen joining it to the preceding "word".  

It looks a bit weird, but I didn't want to fudge the test, and it's a good demonstration of what it does.  In reality, I don't think we'd ever have a title with underscores in it, so the weird looking URL wouldn't occur.

There's nothing functionally wrong with this btw, it just looks odd

sihugh: It's generated from`Step 3?"`_!`

PR #1156, opened 2019-10-07
Add contextual guidance component

alex-ju:
## What
This PR:
- moves the [contextual-guidance component from Content Publisher](https://content-publisher.integration.publishing.service.gov.uk/component-guide/contextual_guidance) to this gem.
- updates the component as follows:
  - ensures the component generates the required layout (2/3 for input 1/3 for guidance) - instead of relying on manually placing the components in the right structure
  - refactors the code so that the guidance refers to the input using a `for` attribute - instead of defining for each input what's the associated guidance and then querying it from there. This helps improving isolation for this component as no special attributes are need for the input fields.

### Notes to reviewer
The component was ported as-is in the first commit then updated on the following ones to make the refactor clearer. If you're not familiar with how the component worked in Content Publisher it may be a better idea to review the final changes. The updated version of this component was [tested in Content Publisher](https://github.com/alphagov/content-publisher/compare/contextual-guidance?expand=1) to ensure a clear upgrade path.

## Why
This component is needed in Collection Publisher to show guidance for various input fields.

## Visual Changes
No visual changes

## View Changes
https://govuk-publishing-compo-pr-1156.herokuapp.com/component-guide/contextual_guidance

[Trello card](https://trello.com/c/Ju8oaE5y)

### DIFF
--- a/app/views/govuk_publishing_components/components/_contextual_guidance.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_guidance.html.erb
@@ -6,17 +6,19 @@
   data_attributes[:module] = "contextual-guidance"
 %>
 
-<%= tag.div class: "govuk-grid-row" do %>
-  <%= tag.div class: "govuk-grid-column-two-thirds" do %>
-    <%= tag.div class: "gem-c-contextual-guidance__input-field" do %>
-      <%= yield %>
+<%= tag.div class: "gem-c-contextual-guidance", id: id, data: data_attributes do %>
+  <%= tag.div class: "govuk-grid-row" do %>
+    <%= tag.div class: "govuk-grid-column-two-thirds" do %>
+      <%= tag.div class: "gem-c-contextual-guidance__input-field" do %>
+        <%= yield %>
+      <% end %>
     <% end %>
-  <% end %>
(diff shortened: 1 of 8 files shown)


app/views/govuk_publishing_components/components/_contextual_guidance.html.erb
```diff
@@ -0,0 +1,25 @@
+<%
+  html_for ||= nil
```
kevindew: Should this attribute actually be required? I'm not sure this does anything without it.

alex-ju: I've updated this to not have a default value and added a test to make sure it fails if `html_for` is not provided.

alex-ju:
> This looks good to me, did you resolve the issue that it wasn't fully encapsulated? I wasn't sure if the `&:last-of-type` was maybe still that?

I consider it now being fully encapsulated, only that when we have a list of components the last one is styled differently. We have a similar approach when styling the `app-c-image-meta` and `app-c-attachment-meta` in Content Publisher.

I think that instead of creating a component that generates a list of items (and handles the last one) I'd rather keep the component as an item that can be placed in a loop and handle the last item via CSS selectors.

PR #1158, opened 2019-10-10
Bring specific org focus states in line with others

vanitabarrett:
## What
We have some specific organisations which are styled separately. The hover and focus states for these organisations were set to remain as the brand colours, whereas all other organisations have text set to black on hover and focus. This brings those other organisations in line with the rest.

## Why
The branding hover and focus states should be consistent across organisations

### Hover state
Screen Shot 2019-10-10 at 14 56 16
Screen Shot 2019-10-10 at 15 06 48

### Focus state
Screen Shot 2019-10-10 at 14 56 08
Screen Shot 2019-10-10 at 15 06 55

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
@@ -15,7 +15,7 @@
 
         &:hover,
         &:focus {
-          color: govuk-colour("black");
+          color: $govuk-focus-text-colour;
         }
       }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
```diff
@@ -46,7 +46,7 @@
 
     &:hover,
     &:focus {
-      color: darken(govuk-organisation-colour("office-of-the-leader-of-the-house-of-commons"), 10%);
+      color: govuk-colour("black");
```
alex-ju: I suggest we use `$govuk-focus-text-colour`, just in case we'll change the colours again 😁

vanitabarrett:
Note: not sure why this build is failing at the moment.

```
Sprockets::ArgumentError:
      #   link_tree argument must be a directory
      #   ./spec/dummy/app/assets/config/manifest.js:2
```

PR #1172, opened 2019-10-18
Revert "Update to sprockets 4"

alex-ju:
Reverts "Update to sprockets 4" done in #1160, #1163, #1165
Reverts releases (these were unpublished from rubygems)
Pin sprockets to 3.x for the CI
Pin foreman to 0.85 for the CI

[Diff between previous (stable) release 21.5.0 and this revert](https://github.com/alphagov/govuk_publishing_components/compare/release-21-5-0...revert-update-dependencies?expand=1).


govuk_publishing_components.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rake"
   s.add_dependency "rouge"
   s.add_dependency "sassc-rails", ">= 2.0.1"
+  s.add_dependency "sprockets", "~> 3.7.2"
```
benthorner: Minor, but we could more accurately do "< 4" here.

govuk_publishing_components.gemspec
```diff
@@ -26,8 +26,8 @@ Gem::Specification.new do |s|
   s.add_dependency "sassc-rails", ">= 2.0.1"
   s.add_dependency "sprockets", "~> 3.7.2"
 
-  s.add_development_dependency "capybara", "~> 3.25"
-  s.add_development_dependency "foreman", "~> 0.85"
+  s.add_development_dependency "capybara", "= 3.25"
```
benthorner: Do we need to do this?

alex-ju: Unfortunately, yes 😞otherwise CI will fail

alex-ju:
> I had a quick go using this (as a local gem) in Content Publisher, but encountered an error...

That error tells me your build was using sprockets 4 - not sure how though, but the current gem version is not pulling this in. I tested content-publisher, frontend, collections and finder-frontend with this branch and they all work fine.

benthorner:
Interesting. I just did `npm install`. Also tried `rm -rf node_modules; npm install`. All in govuk-docker, by the way. I wonder if it could still be something to do with local node_modules?

alex-ju:
@benthorner it worked fine for me yesterday. now looking at the dependabot's PR in Content Publisher it seems to [fail in CI](https://ci.integration.publishing.service.gov.uk/job/content-publisher/job/dependabot%252Fbundler%252Fgovuk_publishing_components-21.6.0/6/execution/node/125/log/) because of a capybara-related ambiguity. Will try and look into it when I have a few moments.

PR #1173, opened 2019-10-21
Add inverse option to translation component

vanitabarrett:
## What
Add inverse option to translation component so it displays as white text when the flag is passed. This can be used when using the component on a darker background

## Why
We have a use case for using the translation component on a darker background.

Screen Shot 2019-10-21 at 12 14 00

https://govuk-publishing-compo-pr-1173.herokuapp.com/component-guide/translation_nav


alex-ju:
🏴󠁧󠁢󠁷󠁬󠁳󠁿

PR #1177, opened 2019-10-24
Add GovernmentService schema.org schema

karlbaker02:
## What

This PR adds a new `GovernmentService` schema to be used by the machine readable metadata component. At the moment this is a basic skeleton of a schema, which will be iterated upon as we build out the data used to power it.

Trello: https://trello.com/c/g6uvswCG/69-add-governmentservice-schema-to-transaction-pages

## Why

So that we can use the `GovernmentService` schema to help improve results shown on search engines.

## Visual Changes

Example outputs of this schema run against Google's Structured Data Testing Tool

Book your driving test

Screenshot 2019-10-24 at 15 21 20

Apply for or renew a Blue Badge

Screenshot 2019-10-24 at 15 38 27

## View Changes

https://govuk-publishing-compo-pr-1177.herokuapp.com/

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,10 @@
 - Don't include changes that are purely internal. The CHANGELOG should be a
   useful summary for people upgrading their application, not a replication
   of the commit log.
+  
+## Unreleased
+
+* Add GovernmentService schema.org schema to machine readable metadata component ([PR #1177](https://github.com/alphagov/govuk_publishing_components/pull/1177))
 
 ## 21.6.1
 
--- a/lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
+++ b/lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
@@ -1,0 +1,62 @@
+module GovukPublishingComponents
+  module Presenters
+    class GovernmentServiceSchema
+      attr_reader :page
+
+      def initialize(page)
+        @page = page
+      end
+
+      def structured_data
+        # http://schema.org/GovernmentService
+        {
+          "@context" => "http://schema.org",
+          "@type" => "GovernmentService",
+          "name" => page.title,
+          "description" => page.description,
+          "url" => page.canonical_url
+        }
+          .merge(provider)
+          .merge(related_services)
+      end
+
+    private
+
+      def related_services
+        related_links = page.content_item.dig('links', 'ordered_related_items')
+
+        return {} unless related_links.present?
+
+        {
+          "isRelatedTo" => related_links.each_with_object([]) do |link, items|
+            if link['schema_name'] == 'transaction'
+              items << {
+                "@type" => "GovernmentService",
+                "name" => link['title'],
+                "url" => link['web_url']
+              }
+            end
+          end
+        }
+      end
+
+      def provider
+        organisations = page.content_item.dig('links', 'organisations')
+
+        return {} unless organisations.present?
+
+        providers = organisations.map do |organisation|
+          {
+            "@type" => "GovernmentOrganization",
+            "name" => organisation['title'],
+            "url" => organisation['web_url']
+          }
+        end
+
+        {
+          "provider" => providers
+        }
+      end
+    end
+  end
+end
--- a/app/views/govuk_publishing_components/components/docs/machine_readable_metadata.yml
+++ b/app/views/govuk_publishing_components/components/docs/machine_readable_metadata.yml
@@ -62,6 +62,14 @@ examples:
         <p>Your next of kin may need a strong stomach</p>
       schema: :faq
       canonical_url: https://www.gov.uk/how-to-train-your-dragon
+  government_service_schema:
+    data:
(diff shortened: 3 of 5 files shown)


lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
```diff
@@ -0,0 +1,73 @@
...
+        @page = page
+      end
+
+      def structured_data
+        # http://schema.org/GovernmentService
+        {
+          "@context" => "http://schema.org",
+          "@type" => "GovernmentService",
+          "name" => page.title,
+          "serviceType" => page.description,
```
sihugh: Did we think this might be some sort of enumeration?  Something like https://standards.esd.org.uk/?uri=list%2Finteractions# (even though they're discontinued)

CHANGELOG.md
```diff
@@ -6,6 +6,10 @@
 - Don't include changes that are purely internal. The CHANGELOG should be a
   useful summary for people upgrading their application, not a replication
   of the commit log.
+  
+## Unreleased
+
+* Add GovernmentService schema.org schema ([PR #1177](https://github.com/alphagov/govuk_publishing_components/pull/1177))
```
theseanything: Can we be more specific?
```suggestion
* Add GovernmentService schema.org schema to machine readable metadata component([PR #1177](https://github.com/alphagov/govuk_publishing_components/pull/1177))
```

karlbaker02: I was following the prior art of previous commits here - I'm happy with either.

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
```diff
@@ -0,0 +1,73 @@
...
+          "description" => description,
```
sihugh: In our other schemas this is usually the `page.description` - possibly best to stick with that unless we have a strong reason not to?

CHANGELOG.md
andysellick: The changelog should match the PR title. If more detail is needed they should both be updated, but hopefully the description gives enough information.

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
sihugh: Also, this has raised a good point, that the Article schema doesn't seem to include the page body for transactions in it (because of the weird properties like `more_information` that you've seen).  We could probably do that there and keep this more simple for now?

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
```diff
@@ -0,0 +1,73 @@
...
+    private
+
+      def related_services
+        related_links = page.content_item.dig('links', 'ordered_related_items')
+
+        return {} unless related_links.present?
+
+        {
+          "isRelatedTo" => related_links.each_with_object([]) do |link, items|
+            if link['schema_name'] == 'transaction'
```
sihugh: Nice 👍

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
```diff
@@ -0,0 +1,73 @@
...
+          end
+        }
+      end
+
+      def provider
+        organisations = page.content_item.dig('links', 'organisations')
+
+        return {} unless organisations.present?
+
+        organisation = organisations.first
```
sihugh: could we return an array of orgs here?  Something like:

```
orgs = organisations.map do |org|{
  {
    "@type" => "GovernmentOrganization",
    "name" => org['title'],
    "url" => org['web_url']
   }
end

{ "provider" => orgs }
```

CHANGELOG.md
theseanything: Yup - for me I found the title wasn't enough to understand what's happened even and reads as an incomplete sentence. (Sorry its a minor thing)

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
karlbaker02: `provider` as far as I can tell isn't an array: https://schema.org/provider

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
karlbaker02: To confirm, this is getting rid of `serviceType` for now and using our `description` in the `description` property? I think this might work better actually, as `serviceType` is just a categorisation of the service.

lib/govuk_publishing_components/presenters/machine_readable/government_service_schema.rb
karlbaker02: As discussed, I'm going to remove `serviceType` for now because on reflection `description` is the more appropriate property to use for the description contained within our content schema for `transaction`s.

PR #1181, opened 2019-10-28
Make chevron component more resilient

vanitabarrett:
Trello: https://trello.com/c/R2Ah3m9O/163-make-chevron-component-more-resilient

## What 
This rewrites the component to handle varying text lengths better, and remove some of the hacks we had to put in place to get it to work.

**Summary:** 

- switches from a background-image to using an inline SVG with `fill` and `stroke` being applied dynamically via CSS rather than requiring multiple different SVGs
- switches to "mobile" mode (chevron inside a rectangle) a bit sooner (450px rather than 320px)

## Why
The chevron component was developed quickly and under time pressure, and therefore was quite brittle to any changing content within it as we hadn't expected the text to change. We've had to put overrides in place within Collections - this isn't good, and we want to be able to remove these.

## Visuals
While we try to create the same experience in IE browsers, this hasn't been fully possible due to different [SVG support on IE browsers](https://caniuse.com/#search=svg). On IE8, we fallback to a full width rectangle (without chevron) as SVG support means the chevron simply fails to render. On IE9+, we have had to hardcode a maximum width to get the chevron to display in the correct location. This is a best guess with what works with the content we have on desktop - smaller sizes may not render perfectly.

### IE8
Screen Shot 2019-10-30 at 10 50 37
Screen Shot 2019-10-30 at 10 50 46

### IE9+
Screen Shot 2019-10-30 at 10 51 30
Screen Shot 2019-10-30 at 10 51 44

https://govuk-publishing-compo-pr-1181.herokuapp.com/component-guide/chevron_banner


andysellick:
The component escapes from the container between around 450px and 485px. I don't know if this would be a problem in the wild, maybe worth testing it on the homepage to be sure.

Screenshot 2019-10-29 at 16 09 22

app/views/govuk_publishing_components/components/_chevron_banner.html.erb
```diff
@@ -7,17 +7,20 @@
...
-
   <div class="gem-c-chevron-banner <%= "gem-c-chevron-banner--hover-border" if hover_border %>">
-    <div class="gem-c-chevron-banner__link-container">
-      <%= link_to(text, href, class: "gem-c-chevron-banner__link govuk-link", data: data_attributes) %>
-    </div>
+    <%= link_to href, class: "gem-c-chevron-banner__link", data: data_attributes, role: "text" do %>
+      <span class="gem-c-chevron-banner__text"><%= text %></span>
+      <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 -.2 175 109.5" class="gem-c-chevron-banner__chevron" preserveAspectRatio="xMidYMid meet">
+        <defs>
+          <path id="a" d="M0 0h93v109H0z"></path>
```
andysellick: I _think_ this is causing the aXe error in the component guide. Although even if it isn't, I don't think it's needed?

alex-ju: I agree, the SVG could be simplified to something like the code below or even more
```
      <svg xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" viewBox="0 0 175 109" class="gem-c-chevron-banner__chevron" preserveAspectRatio="xMidYMid meet">
        <path class="gem-c-chevron-banner__chevron-point" fill="#E61E32" fill-rule="nonzero" d="M91.611 51.664c1.0371 1.1432 1.0375 2.8869.000802 4.0305l-44.991 49.634c-.37092.40919-.33989 1.0416.069307 1.4125.18394.16674.42334.25909.6716.25909h36.892c.55978 0 1.0939-.2346 1.4727-.6468l46.534-50.643c1.0549-1.148 1.0546-2.9127-.000709-4.0604L85.3617.64689c-.37873-.41188-.91268-.64628-1.4722-.64628h-36.891c-.55228 0-1 .44772-1 1 0 .24841.092454.48792.25936.6719l45.352 49.992z" transform="translate(0 1)"></path>
        <path class="gem-c-chevron-banner__chevron-base" fill="#E61E32" fill-rule="nonzero" d="M-37.997 1l.35768 107h40.408c.55978 0 1.0939-.2346 1.4727-.6468l46.534-50.643c1.0549-1.148 1.0546-2.9127-.0007095-4.0604l-46.898-51.003c-.37873-.41188-.91268-.64628-1.4722-.64628h-40.402z" vector-effect="non-scaling-stroke"></path>
      </svg>

```

Will have a think about how to avoid that overflow though..

vanitabarrett:
Note: the overflow issue isn't completely fixed, but I've slightly changed the way in which the chevron is being positioned which makes it less likely to happen. 

As far as I can see, the overflow issue only happens when the text inside the chevron is over 2 lines (on desktop), causing the chevron to have to scale a lot more on smaller screens which leads to the overflow. I couldn't see an easy way of fixing this, and I'm hoping the way in which we're using the chevron means we won't ever have that much text inside it.

PR #1187, opened 2019-11-04
Allow tracking of the details component

vanitabarrett:
Trello:
https://trello.com/c/Dtle7GcF/197-modify-details-component-to-enable-tracking-analytics
to unblock https://github.com/alphagov/finder-frontend/pull/1702

## What
Allow tracking of the details component, e.g: on the Brexit checker.

## Why
The details component, by default, has `data-module="govuk-details` which makes it difficult for us to add our usual `data-module="track-click"`. We also want some more custom tracking of the open and close status of the component. This adds tracking JS to the component. 

Passing all the necessary data attributes, including `track-label` simply passes the details element to our usual [TrackClick method in static](https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/track-click.js). This then works in exactly the same way as the rest of our tracking.

Passing in all the necessary data attributes **except `track-label** means that the track-label by default becomes the open/close status of the component. In this case, we manually add an event listener and fire a track event on click.

Screen Shot 2019-11-04 at 15 18 16

https://govuk-publishing-compo-pr-1187.herokuapp.com/

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/details.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/details.js
@@ -1,0 +1,48 @@
+/* eslint-env jquery */
+// = require govuk/components/details/details.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  Modules.GovukDetails = function () {
+    this.start = function (element) {
+      var customTrackLabel = element[0].getAttribute('data-track-label')
+
+      // If a custom label has been provided, we can simply call the tracking module
+      if (customTrackLabel) {
+        var trackDetails = new window.GOVUK.Modules.TrackClick()
+        trackDetails.start(element)
+      } else {
+        // If no custom label is set, we use the open/close status as the label
+        var detailsComponent = element[0]
+        var detailsClick = detailsComponent.querySelector('[data-details-track-click]')
+        var that = this
+
+        $(detailsClick).click(function (e) {
+          that.trackDefault(detailsComponent)
+        })
+      }
+    }
+
+    this.trackDefault = function (element) {
+      if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+        var componentStatus = (element.getAttribute('open') == null) ? 'open' : 'closed'
+        var trackCategory = element.getAttribute('data-track-category')
+        var trackAction = element.getAttribute('data-track-action')
+        var trackOptions = element.getAttribute('data-track-options')
+
+        if (typeof trackOptions !== 'object' || trackOptions === null) {
+          trackOptions = {}
+        }
+
+        trackOptions['label'] = componentStatus
+
+        if (trackAction && trackCategory) {
+          window.GOVUK.analytics.trackEvent(trackCategory, trackAction, trackOptions)
+        }
+      }
+    }
+  }
+})(window.GOVUK.Modules)
--- a/app/views/govuk_publishing_components/components/docs/details.yml
+++ b/app/views/govuk_publishing_components/components/docs/details.yml
@@ -22,6 +22,18 @@ examples:
       block: |
         We need to know your nationality so we can work out which elections you’re entitled to vote in. If you can’t provide your nationality, you’ll have to send copies of identity documents through the post.
   with_data_attributes:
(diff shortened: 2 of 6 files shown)


vanitabarrett:
Thank you helping fix the tests @alex-ju ! 🙇‍♀

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -0,0 +1,48 @@
...
+        var trackAction = element.getAttribute('data-track-action')
+        var trackOptions = element.getAttribute('data-track-options')
+
+        if (typeof trackOptions !== 'object' || trackOptions === null) {
+          trackOptions = {}
+        }
+
+        trackOptions['label'] = componentStatus
+
+        if (trackAction && trackCategory && window.GOVUK.analytics) {
```
alex-ju: we've already checked for `window.GOVUK.analytics` to exist in the outer conditional statement so I think we can remove it from here

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -0,0 +1,48 @@
...
+      if (customTrackLabel) {
+        var trackDetails = new window.GOVUK.Modules.TrackClick()
+        trackDetails.start(element)
+      } else {
+        // If no custom label is set, we use the open/close status as the label
+        var detailsComponent = element[0]
+        var detailsClick = detailsComponent.querySelector('[data-details-track-click]')
+        var that = this
+
+        $(detailsClick).click(function (e) {
```
andysellick: Is this line jquery? Only it looks like the rest of the file doesn't use it, so would be good to avoid it entirely if possible.

vanitabarrett:
@alex-ju Agreed - I wasn't sure whether to make this component-specific or not. I opted for making it specific to the details component just because it seemed like quite a custom bit of tracking, but happy to move it when we come to a consensus on where these things should live.

app/views/govuk_publishing_components/components/docs/details.yml
```diff
@@ -22,6 +22,18 @@ examples:
       block: |
         We need to know your nationality so we can work out which elections you’re entitled to vote in. If you can’t provide your nationality, you’ll have to send copies of identity documents through the post.
   with_data_attributes:
+    description: Can be used for tracking. By default, `track-label` is set to the status ("open" or "closed") unless a track_label is passed into the component.
+    data:
+      title: Help with nationality
+      data_attributes:
+        track_category: "track-category"
```
huwd: The umbrella term eg `brexitCheckerHelper` Something that identifies all helper messages with a specific feature

vanitabarrett: @huwd I don't want to make the example too specific to the Brexit checker, as it's really an analytics pattern decided by PAs rather than the component itself. But I've used actual examples now rather than placeholder strings.

spec/javascripts/components/details-spec.js
```diff
@@ -0,0 +1,80 @@
...
+  var FIXTURE
+
+  GOVUK.analytics = {
+    trackEvent: function () {}
+  }
+
+  var callback = jasmine.createSpy()
+  GOVUK.Modules.TrackClick = function () {
+    var that = this
+    that.start = function () {
```
andysellick: Does calling `this.start` here not work? Or is `that` used somewhere else?

app/assets/javascripts/govuk_publishing_components/components/details.js
vanitabarrett: This was originally written without jQuery, but we seem to have a mismatch in the way we write tests (relying heavily on jQuery). I had an issue with tests passing locally, but failing in CI. It seemed to be because of this mismatch, so the click event wasn't being triggered. The simplest fix was to move this line to jQuery - not ideal, but moving everything the other way (tests written without jQuery) looked like a bigger piece work. 

I've made a note, so will revisit it once this has been deployed.

andysellick: Fair enough.

vanitabarrett:
Waiting for confirmation on the tracking attributes before this can be merged

vanitabarrett:
Update on the tracking from Paul. In summary, the logic should be ok as is.

> We're already inconsistent in how we use eventLabel vs eventAction so it doesn't really matter. Choose which ever is easiest for you! What's important is that we record which question the event takes place on. We normally depend on the URL for this but this doesn't work within the checker. I think there are labels for questions like 'travelling' and 'employment' which are used on the checkbox event.

PR #1197, opened 2019-11-14
Update background colour for search buttons

alex-ju:
## What
This is the only place where we use the `light-blue` in the library and as far as I know, for no strong reasons.

## Why
Using `blue` increases the contrast with the magnifying glass icon and makes the component more consistent with other elements on the page.

## Visual Changes
Before
Screen Shot 2019-11-14 at 14 14 34

After
Screen Shot 2019-11-14 at 14 14 15

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update background colour for search buttons ([PR #1197](https://github.com/alphagov/govuk_publishing_components/pull/1197))
 * Support document list items without links ([PR #1194](https://github.com/alphagov/govuk_publishing_components/pull/1194))
 
 ## 21.10.0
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update background colour for search buttons ([PR #1197](https://github.com/alphagov/govuk_publishing_components/pull/1197))
```
DilwoarH: Do we need this for a guidance page update?

DilwoarH:
This component isn't being used anywhere:

https://github.com/search?q=org%3Aalphagov+%22govuk_publishing_components%2Fcomponents%2Fsearch%22&type=Code

Should we investigate?

andysellick:
Screenshot 2019-11-14 at 15 51 02

Must be something weird with github's search, it's definitely used ☝️

DilwoarH:
> Screenshot 2019-11-14 at 15 51 02
> 
> Must be something weird with github's search, it's definitely used ☝️

Yes ignore me, my search query was bad. It's used here:
https://github.com/alphagov/frontend/blob/master/app/views/homepage/index.html.erb#L18

alex-ju:
> This component isn't being used anywhere:

The default version is used on [the search page ](https://www.gov.uk/search)(don't ask how to get there though) 😬

PR #1198, opened 2019-11-14
Add heading and inline options to organisation logo

injms:
## What
This adds an extra configuration option to set the wrapping element of the organisation logo to be a heading.

This also adds the option to make that element not fill up as wide as the parent element. Since this could break things, this option is off by default.

## Why

Wrapping the organisation logo in a heading allows it to be part of the content of the page, as required by the [Ministers page](https://www.gov.uk/government/ministers). 

Making the component inline removes a large invisible area that users could mistakenly tap and trigger the link. It also makes the focus state more consistent.

## Visual Changes

No visual changes when changing the wrapping element.

### Inline before:
image

### Inline after:
image

## View Changes
https://govuk-publishing-compo-pr-1198.herokuapp.com/component-guide/organisation_logo

### DIFF
--- a/app/views/govuk_publishing_components/components/_organisation_logo.html.erb
+++ b/app/views/govuk_publishing_components/components/_organisation_logo.html.erb
@@ -1,21 +1,41 @@
 <%
   logo_helper = GovukPublishingComponents::Presenters::OrganisationLogoHelper.new(local_assigns)
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(organisation[:brand])
+
   organisation ||= {}
+  heading_level ||= false
+  inline ||= false
+
+  # Check if `heading_level` is a number; if so, check if it's an appropriate
+  # number.
+  use_heading = heading_level.is_a?(Integer) ?
+                  (heading_level >= 1 && heading_level <= 6) :
+                  false
+
+  # Set the wrapping element to be a heading or a `div`
+  wrapping_element = use_heading ? "h#{heading_level}" : "div"
+
+  wrapper_classes = %w(gem-c-organisation-logo)
+  wrapper_classes << brand_helper.brand_class
+
+  container_classes = [
+    logo_helper.logo_container_class,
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_organisation_logo.html.erb
```diff
@@ -1,21 +1,41 @@
...
+  # Check if `heading_level` is a number; if so, check if it's an appropriate
+  # number.
+  use_heading = heading_level.is_a?(Integer) ?
+                  (heading_level >= 1 && heading_level <= 6) :
+                  false
+
+  # Set the wrapping element to be a heading or a `div`
+  wrapping_element = use_heading ? "h#{heading_level}" : "div"
+
+  wrapper_classes = ["gem-c-organisation-logo"]
```
alex-ju: we're generally using the `%w(gem-c-organisation-logo)` notation when working with arrays of classes — I personally have no preference, but this would help with consistency.

spec/components/organisation_logo_spec.rb
```diff
@@ -27,7 +27,7 @@ def component_name
   end
 
   it "doesn't include a link when a URL is not provided" do
-    render_component(organisation: { name: "Linked" })
```
alex-ju: 😬

injms: ⛏

app/views/govuk_publishing_components/components/_organisation_logo.html.erb
injms: Okaydokey - I've changed it to use the word array shorthand.

injms:
> Good improvement! I left a non-blocking comment and **don't forget the changelog entry**. Have you got the chance to check it renders fine at the top of an organisation page?

The changes have to be explicitly turned on, so it'll render correctly wherever it's  currently being used - the default behaviour hasn't changed.

PR #1200, opened 2019-11-18
Remove false positive accessibility errors

alex-ju:
## What
Update `accessibility_excluded_rules` on a few components to remove false positives from the component guide.

## Why
Our components have fair few accessibility issues, but having not-so-relevant errors makes it harder to focus on those that actually need to be addressed.

## Visual Changes
No visual changes

https://govuk-publishing-compo-pr-1200.herokuapp.com/component-guide

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/accordion.yml
+++ b/app/views/govuk_publishing_components/components/docs/accordion.yml
@@ -215,7 +215,7 @@ examples:
   different_heading_level:
     description: This will alter the level of the heading, not the appearance of the heading.
     data:
-      heading_level: 5
+      heading_level: 3
       items:
         - heading:
             text: "Writing well for the web"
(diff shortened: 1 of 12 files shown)


vanitabarrett:
@alex-ju Are we able to address these errors too? Looks like a problem with the skip-link target not existing, but shouldn't be an actual accessibility issue with the component.

- [Layout header](https://govuk-publishing-compo-pr-1200.herokuapp.com/component-guide/layout_header).
- [Metadata](https://govuk-publishing-compo-pr-1200.herokuapp.com/component-guide/metadata/updated_with_links_to_see_details).

alex-ju:
> @alex-ju Are we able to address these errors too? Looks like a problem with the skip-link target not existing, but shouldn't be an actual accessibility issue with the component.

Good finds @vanitabarrett. I pushed commits to cover these two cases too.

PR #1206, opened 2019-11-25
Check if CSP is available

chao-xian:
This is causing tests to fail for apps that use the gem but not
on Rails 5.2, which is where CSP is introduced as a feature.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


chao-xian:
[Follow up Trello card to address the overall issue](https://trello.com/c/HBmTGp23).

chao-xian:
@alex-ju thanks! Do you think this needs a Changelog entry?

alex-ju:
> @alex-ju thanks! Do you think this needs a Changelog entry?

It doesn't affect the components 'API' or how they are being displayed, so I'd say it's not necessary, but feel free to add if you want to.

PR #1207, opened 2019-11-25
Change tags and styling for feedback component

SamJamCul:
This change addresses the issue encountered when a user that has blocked
css styling sees the feedback component. Originally, the options
presented to the user are all strung together in one line, making it
potentially confusing. This change separates them out into a list,
meaning that when there is no styling for the page, each option is on
its own separate row.

The styling is effectively neutralised if css is turned on, and all of
the default behaviours for the lists are removed.

See this [Trello card](https://trello.com/c/otALwPuk/1170-y-n-links-not-obvious-when-styles-are-removed) for more details

### DIFF
--- a/app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
@@ -5,8 +5,7 @@
 <div class="gem-c-feedback__prompt gem-c-feedback__js-show js-prompt" tabindex="-1">
   <div class="gem-c-feedback__js-prompt-questions js-prompt-questions">
     <h2 class="gem-c-feedback__prompt-question">Is this page useful?</h2>
-    <ul class="gem-c-feedback__option-list">
-      <%= link_to contact_govuk_path, {
+    <%= link_to contact_govuk_path, {
         class: 'gem-c-feedback__prompt-link',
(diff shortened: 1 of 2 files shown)


SamJamCul:
Some more info - it's addressing point 8 in [this comment] (https://github.com/alphagov/govuk_publishing_components/pull/163#issuecomment-366267858)

And here's the before with no CSS 
image

And then here's the new change 
image

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
```diff
@@ -5,59 +5,63 @@
...
-        'track-action' => 'GOV.UK Open Form'
-      },
-      'aria-controls': 'something-is-wrong',
-      'aria-expanded': false,
-      role: 'button',
-    } do %>
-      Is there anything wrong with this page?
-    <% end %>
+    <ul class="gem-c-feedback__option-list">
+      <%= link_to contact_govuk_path, {
```
injms: Only list items can be direct children of unordered lists - so this link should be outside of the `ul` element.

SamJamCul: Thanks Ian, I've shifted it so that it's outside the ul, should be tidied up now

PR #1210, opened 2019-11-27
[WIP - Do not merge] Add nodeList to Array polyfill helper

kr8n3r:
## What
Add a helper to create and array from a list of Nodes in vanilla javascript.
Check for Array.from support: if browser supports it use that, else employ the services of a good'ol for-loop

## Why
Help to move to a jQuery-free world

## To do

- [ ] Add tests

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/polyfills/nodelist-to-array.js
+++ b/app/assets/javascripts/govuk_publishing_components/polyfills/nodelist-to-array.js
@@ -1,0 +1,19 @@
+// helper to loop over nodes and create an array for cross-browser compatibillity
+// if browser support Array.from we'll call that
+// else iterate over the array with a good'ol for-loop
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Polyfills = window.GOVUK.Polyfills || {};
+
+(function (Polyfills) {
+  'use strict'
+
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/polyfills/nodelist-to-array.js
```diff
@@ -0,0 +1,19 @@
+// helper to loop over nodes and create an array for cross-browser compatibillity
+// if browser support Array.from we'll call that
+// else iterate over the array with a good'ol for-loop
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Polyfills = window.GOVUK.Polyfills || {};
+
+(function (Polyfills) {
```
nickcolley: What's your thinking of doing this instead of using the polyfill in GOV.UK Frontend? Not saying it's the wrong approach since I'm not 100% confident with our polyfills but interested to hear your thoughts.

kr8n3r: it was my first point of call, but the polyfill in the frontend is attached to the global scope so we can't initialise in in publishing components. additionally it doesn't work on IE8

kr8n3r:
closing as there is no demand for it

PR #1216, opened 2019-12-06
Simplify CSS Grid on feedback component

kr8n3r:
## What
- remove redundant row definitions and use the implicit rows.

- remove grid column and row definitions from the parent container,
as they're inherited from the child when needed

- replace column definition with a simple two width definitions,
as we only care about the minimum width of the first column

## Why
Remove redundant declarations. Keep it simple.

## View Changes

https://govuk-publishing-compo-pr-1216.herokuapp.com/component-guide/feedback/preview

## Visual Changes
No visual changes

### iOS Safari 12 (iphone 8)
Screenshot 2019-12-06 at 08 37 43

### Desktop Firefox, Safari, Chrome
Screenshot 2019-12-06 at 08 22 30

### Mobile Chrome and Samsung internet
Screenshot 2019-12-06 at 08 23 43
Screenshot 2019-12-06 at 08 36 38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ## Unreleased
 
+* Simplify CSS Grid on feedback component ([#1216](https://github.com/alphagov/govuk_publishing_components/pull/1216))
+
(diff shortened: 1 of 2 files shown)


kr8n3r:
@andysellick yeah that alignment isn't the best, but as it's unrelated to the grid stuff it's probably best to address it separately in #1217

PR #1223, opened 2019-12-12
Allow custom heading size on fieldset component

alex-ju:
## What
Allow custom heading size on fieldset component.

## Why
So we can generate custom headings within legends consistently across GOV.UK

## Visual Changes
No visual changes

## View Changes
Before: https://components.publishing.service.gov.uk/component-guide/fieldset
After: https://govuk-publishing-compo-pr-1223.herokuapp.com/component-guide/fieldset

[Trello card](https://trello.com/c/gYQgI6Xg)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Allow custom heading size on fieldset component ([#1223](https://github.com/alphagov/govuk_publishing_components/pull/1223))
+
 ## 21.13.5
 
 * Fix checkbox small option ([#1221](https://github.com/alphagov/govuk_publishing_components/pull/1221))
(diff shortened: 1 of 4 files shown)


kr8n3r:
it looks good to me, but is this a breaking change @alex-ju ?

alex-ju:
> it looks good to me, but is this a breaking change @alex-ju ?

No, we still allow sanitised html to be passed in as values (including to the `legend_text` in this component) so it's backwards compatible, we're just not promoting it. I introduced the `yield` block as an alternative to `text` attribute a while back but haven't updated the examples.

kr8n3r:
> > it looks good to me, but is this a breaking change @alex-ju ?
> 
> No, we still allow sanitised html to be passed in as values (including to the `legend_text` in this component) so it's backwards compatible, we're just not promoting it. I introduced the `yield` block as an alternative to `text` attribute a while back but haven't updated the examples.

yeah, i notice that after i've posted and then deleted the comment

🚢

PR #1244, opened 2019-12-23
Revert "Simplify document list component markup when having only one item"

alex-ju:
Reverts alphagov/govuk_publishing_components#1226

One thing that was missed when carrying out this work is that `finder-frontend` (both backend/Rails and frontend/JavaScript) relies heavily on this component having a specific markup structure (div>ul>li) – hence we can not simplify the markup within the component. An alternative option would be to extract the document bit to its own component and update the document-list to make use of it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Revert 'Simplify document list component markup when having only one item' ([#1244](https://github.com/alphagov/govuk_publishing_components/pull/1244))
+
 ## 21.16.2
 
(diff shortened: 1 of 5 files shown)


kr8n3r:
can work I around that?

alex-ju:
> can the search team work around that?

Maybe we can chat about it in Jan, but this made me realise that in the short run my approach is not suited for all purposes.

kr8n3r:
i guess you could pass a flag `removeThisListMarkup: true` `:) where this is false by default.

do you want to merge/release this revert soon?

alex-ju:
> i guess you could pass a flag `removeThisListMarkup: true` `:) where this is false by default.

Yep, that could be an option too, but rather specific.

Yes, please. It can do with a tick. I'd rather revert now and unblock all apps from being updated to the latest release and address the problem with a more suited approach.

PR #1247, opened 2019-12-31
Add Dataset schema.org schema to machine readable metadata component

laurentqro:
## What

Add Dataset schema.org schema to machine readable metadata component.

## Why

Things that appear in the published statistics list in the [research and stats finder](https://www.gov.uk/search/research-and-statistics) can be marked up with structured data so that they are recognised as statistical datasets (https://developers.google.com/search/docs/data-types/dataset).

One benefit of this is that they might appear in Google's dataset search tool: https://toolbox.google.com/datasetsearch.

## Trello card

https://trello.com/c/0W3LfcGl/1232-implement-schemaorg-dataset-schema-on-published-statistics-document-types

### DIFF
--- a/lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
+++ b/lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
@@ -1,0 +1,34 @@
+module GovukPublishingComponents
+  module Presenters
+    class DatasetSchema
+      attr_reader :page
+
+      def initialize(page)
+        @page = page
+      end
+
+      def structured_data
+        # http://schema.org/Dataset
+        data = CreativeWorkSchema.new(@page).structured_data
+          .merge(description)
+          .merge(name)
+        data["@type"] = "Dataset"
+        data
+      end
+
+    private
+
+      def description
+        {
+          "description" => page.description || page.body
+        }
+      end
+
+      def name
+        {
+          "name" => page.title
+        }
+      end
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
```diff
@@ -0,0 +1,34 @@
...
+        @page = page
+      end
+
+      def structured_data
+        # http://schema.org/Dataset
+        data = CreativeWorkSchema.new(@page).structured_data
+          .merge(description)
+          .merge(name)
+        data["@type"] = "Dataset"
+        data
```
barrucadu: It's a real shame we don't have attachments appear as links in the content items.  If we did, this could also set the `distribution` property and link straight to the file.

laurentqro: @barrucadu maybe one to add to the backlog?

barrucadu: Could be, but worth talking to Simon first - there may be other things we'd like to do that would be unblocked by adding attachments to content items, which could inform how we approach that.  For example, having publishing-api tell asset-manager to make an asset live when the corresponding edition is published.

And to do it properly and consistently, we'd need to implement this in *every* publishing app that does attachments, not just whitehall.

PR #1251, opened 2020-01-08
Add print icons

vanitabarrett:
## What
Adds print icons to the component gem

## Why
This icon is currently only used in government-frontend, so hasn't been promoted to the gem as there is no common need for it. 

However, we now want a print link in finder-frontend too. This should probably be a common component, but the logic for opening the printing dialog is more complex than anticipated. A Javascript solution could be implemented, but people with JS turned off would lose the ability to print. For now, let's just move the icon into the gem.


injms:
This icon could an SVG as well - but that shouldn't block this pull request.

PR #1257, opened 2020-01-16
Check for ordered related items and parent in breadcrumb logic

vanitabarrett:
Zendesk: https://govuk.zendesk.com/agent/tickets/3890644
Trello: https://trello.com/c/wP0pwnM3/824-related-links-impacting-visibility-of-breadcrumbs-on-guidance

## What
The current ordered_related_items breadcrumb logic relies on having a parent in the links hash, but we don't actually check for this until we try to add breadcrumbs. This adds a check early on, so we don't end up with empty breadcrumbs.

## Why
The ordered related items breadcrumbs logic relies on the content item having a parent in the links hash. This isn't always present (in fact, it seems to be a feature of mainstream browse and HTML attachments rather than ordered_related_items).
This means documents with ordered_related_items have empty breadcrumbs, which isn't ideal, especially when they're tagged to taxons which could be shown.
This is more of a temporary fix/workaround until we make a decision on what we expect the breadcrumb behaviour to be when ordered_related_items are present, but parent items aren't.

## Before
Screenshot 2020-01-16 at 09 55 38

## After
Screenshot 2020-01-16 at 09 55 29

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Check for ordered related items and parent in breadcrumb logic ([#1257](https://github.com/alphagov/govuk_publishing_components/pull/1257))
+
 ## 21.19.1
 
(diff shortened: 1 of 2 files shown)


sihugh:
:shipit:

PR #1260, opened 2020-01-20
change dataset schema description to use body primarily

SamJamCul:
This is related to the ticket [here](https://trello.com/c/5sN5oxbt/1288-fix-errors-in-dataset-schema), and is aiming to mitigate the errors we've seen with descriptions being bad lengths.

### DIFF
--- a/lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
+++ b/lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
@@ -20,7 +20,7 @@ module GovukPublishingComponents
 
       def description
         {
-          "description" => page.description || page.body
+          "description" => (page.body || page.description).slice(0..4999)
         }
       end
 
--- a/spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
+++ b/spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
@@ -55,7 +55,7 @@ RSpec.describe GovukPublishingComponents::Presenters::SchemaOrg do
       content_item = GovukSchemas::RandomExample.for_schema(frontend_schema: "statistical_data_set") do |random_item|
         random_item.merge(
           "details" => {
-            "body" => "Dataset body",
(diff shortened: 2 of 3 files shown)


SamJamCul:
I've added the test onto the end of a pre-existing test which I had to modify anyway, as we're using the body rather than the description primarily. I'd be happy to split this out into another test if it makes things better though!

lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
```diff
@@ -29,6 +29,10 @@ def name
           "name" => page.title
         }
       end
+
+      def truncate_description(string)
+        string.length > 5000 ? string.slice(0..4999) : string
```
barrucadu: `slice` doesn't extent the list if it's shorter than your ranger, so this can be `string.slice(0..4999)`.  or the whole thing could even be `(page.body || page.description).slice(0..4999)`

spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
```diff
@@ -75,7 +75,7 @@
 
       expect(structured_data['@type']).to eql("Dataset")
       expect(structured_data['name']).to eql("Dataset Title")
-      expect(structured_data['description']).to eql("Dataset description")
+      expect(structured_data['description']).to eql("Dataset body" + ("a" * 4988))
```
barrucadu: I think it would be better instead to check that the length is 5000.

lib/govuk_publishing_components/presenters/machine_readable/dataset_schema.rb
SamJamCul: done, got rid of the extra method and sliced the original description method

spec/lib/govuk_publishing_components/presenters/schema_org_spec.rb
SamJamCul: done, now checks the length of the description to check that it's only 5000 chars long

PR #1265, opened 2020-01-22
Add public/frontend layout component 

maxgds:
This branch is now ready to be reviewed for merging. Fuller details on what was done are outlined below in the notes used while it was in progress but the basic update is to add a public layout component along with header and footer ones in order to replace the code in Static that currently builds the base template. This allows better use of the Design System and removes reliance on legacy code such as govuk_template. 

The full commit history has been preserved for now as the PR involves a lot of changes, but that should be squashed before merging. 

---

It is a rebased version of the work started by Vanita as the original (https://github.com/alphagov/govuk_publishing_components/tree/add-layout) had drifted from master. The original (now closed) PR with comments is here: https://github.com/alphagov/govuk_publishing_components/pull/1035

https://govuk-publis-rebased-ad-xpqn54.herokuapp.com/component-guide

Vanita's "to do" list and notes are below: 

## To Do:

- [x] Add header
- [x] Add footer using layout footer component
- [x] Allow page content to be passed to component
- [x] Add example to docs
- [x] Add a11y criteria
- [x] Fix logo spacing
- [x] Skip to main page link
- [x] Compare with [govuk frontend layout](https://github.com/alphagov/govuk-frontend/blob/899264cd89559d31f0a790682254d94ecf66418b/package/govuk/template.njk)
- [x] Search bar shouldn't be shown all the time
- [x] Search icon not displaying
- [x] Search bar label not working
- [x] Fix search bar layout
- [x] Ensure we're using gem-c classes
- [x] Whitehall layout with different search bar positioning and links to finders ([example](https://www.gov.uk/government/how-government-works))
- [x] Favicon and other assets
- [x] Full width option, e.g: for topic page
- [x] Allow setting of body classes
- [x] Width container (logo lining up with cookie banner?)
- [x] Cookie banner
- [x] Add tests
- [x] Check stylesheet and JS links in `head`
- [x] Test responsiveness
- [x] omit_header ?
- [x] Check tracking
- [x] Check JS functionality and move into gem if needed
- [x] Make Brexit footer links in line with other footer links
- [x] Compare with layout files in static/rendered html to see what else we might be missing

### For next iteration

We'll stick with Static and Slimmer but switch to this template within Static for now, so the following doesn't need to happen yet.

- [ ] Emergency banners
- [ ] Global banner
- [ ] Survey banner

Extra attention needs to be given to the cross browser testing as we implement in static and frontend apps

## Notes

This implementation makes the following assumptions:

- Expects an `application.js`, `application.css` and `print.css` file. Each Frontend app will have to follow this pattern.
- Meta tags, e.g: for analytics, are not included as a param. Applications can add these by using the [meta tags component](https://components.publishing.service.gov.uk/component-guide/meta_tags) within the `:head` section which this component renders.

To test locally in a frontend app: update the frontend app Gemfile to point to this branch; remove slimmer (and all references to it) from the Gemfile; update the relevant layout file to use the component rather than hardcode the layout.


app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -57,5 +58,16 @@
     <% end %>
 
     </div>
+    </div>
```
andysellick: **Reviewing commit by commit**

Nitpick: indentation.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -57,5 +58,16 @@
...
+
+    <% if local_assigns[:search] %>
+        <div class="govuk-grid-column-one-third">
```
andysellick: **Reviewing commit by commit**

Nitpick: indentation.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,416 @@
...
+    <title><%= title %></title>
+
+    <%= csrf_meta_tags %>
+
+    <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
+    <%= stylesheet_link_tag 'print', media: 'print' %>
+    <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
+
+    <%= stylesheet_link_tag "fonts.css", media: "all", integrity: true, crossorigin: "anonymous" %>
+    <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
```
andysellick: **Reviewing commit by commit**

Presumably all this is copied from static/slimmer somewhere? Could the commit message provide some more context on this? There's a lot of code here.

app/views/govuk_publishing_components/components/docs/layout_for_public.yml
```diff
@@ -5,6 +5,8 @@ body: |
 
 display_preview: false
 display_html: true
+accessibility_criteria: |
+  TBD
```
andysellick, line 9: **Reviewing commit by commit**

TBD

spec/components/layout_for_public_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+    assert_select "title", visible: false, text: "GOV.UK - The best place to find government services and information"
+  end
+
+  it "adds a custom <title> tag" do
+    render_component(title: "Custom GOV.UK Title")
+
+    assert_select "title", visible: false, text: "Custom GOV.UK Title"
+  end
+
+  it "can support full width layouts" do
```
andysellick, line 20: **Reviewing commit by commit**

Does this need a corresponding test to check non-full width layouts?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -43,7 +43,7 @@
       search: true,
     } %>
 
-    <main class="<%= "govuk-width-container" unless full_width %>" id="content">
+    <main class="<%= "govuk-width-container govuk-main-wrapper" unless full_width %>" id="content">
```
andysellick: In the Design System docs these two classes are applied on separate elements, is it okay to combine them like this?

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -11,7 +11,7 @@
   end
 %>
 
-<header class="gem-c-layout-header govuk-header gem-c-layout-header--<%= environment %> govuk-grid-row" role="banner" data-module="govuk-header">
+<header class="gem-c-layout-header govuk-header gem-c-layout-header--<%= environment %>" role="banner" data-module="govuk-header">
```
andysellick: **Reviewing commit by commit**

Obviously the `govuk-grid-row` class here was causing a problem but beneath it are some column classes, are they wrapped by a grid row somewhere higher up?

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -2,7 +2,9 @@
   class_name = "gem-c-search--on-white"
   class_name = "gem-c-search--on-govuk-blue" if local_assigns[:on_govuk_blue].eql?(true)
   size ||= ""
+  no_margin ||= false
```
andysellick, line 5: **Reviewing commit by commit**

We could implement this using the `shared_helper` code to do margin, see https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/shared_helper.rb

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -18,11 +18,11 @@
 
     <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
 
-    <%# the colour used for theme-color is the standard palette $black from
-        https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss %>
+    <!-- the colour used for theme-color is the standard palette $black from GOVUK Frontend -->
```
andysellick: **Reviewing commit by commit**

This line is now a HTML comment, which means it will appear in the markup of the page. I think this and line 25 should go back to being rails comments, we don't want to add any page weight we don't need.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,416 @@
...
+<!DOCTYPE html>
+  <!--[if lt IE 9]><html class="lte-ie8" lang="<%= html_lang %>"><![endif]-->
+  <!--[if gt IE 8]><!--><html lang="<%= html_lang %>"><!--<![endif]-->
+  <head>
+    <meta charset="utf-8" />
+    <title><%= title %></title>
+
+    <%= csrf_meta_tags %>
+
+    <%= stylesheet_link_tag 'application', media: 'all', 'data-turbolinks-track': 'reload' %>
```
vanitabarrett: Does anyone know what the `data-turbolinks-track` attribute does and if it's still needed?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,416 @@
...
+    <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
+
+    <%# the colour used for theme-color is the standard palette $black from
+        https://github.com/alphagov/govuk_frontend_toolkit/blob/master/stylesheets/colours/_palette.scss %>
+    <meta name="theme-color" content="#0b0c0c" />
+
+    <meta name="viewport" content="width=device-width, initial-scale=1">
+
+    <%# The default og:image is added below :head so that scrapers see any custom metatags first, and this is just a fallback %>
+    <meta property="og:image" content="<%= asset_path "opengraph-image.png" %>">
```
vanitabarrett: Have we moved this asset into the gem?

app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
```diff
@@ -2,3 +2,4 @@
   @include govuk-font(16);
   margin: 0;
 }
+
```
vanitabarrett, line 5: Stray new line?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -18,12 +18,18 @@
 
     <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
 
-    <!-- the colour used for theme-color is the standard palette $black from GOVUK Frontend -->
+    <link rel="shortcut icon" href="<%= asset_path 'favicon.ico' %>" type="image/x-icon" />
```
vanitabarrett, line 22: Have all of these assets been moved into the gem?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -16,8 +16,6 @@
     <%= stylesheet_link_tag 'print', media: 'print' %>
     <%= javascript_include_tag 'application', 'data-turbolinks-track': 'reload' %>
 
-    <!--[if lt IE 9]><%= javascript_include_tag "ie.js", integrity: true, crossorigin: "anonymous" %><![endif]-->
```
vanitabarrett: Just wanted to check - is this no longer needed?

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
   }
 }
 
+.gem-c-layout-header--search-left {
+
+  .gem-c-header__menu-button {
```
vanitabarrett, line 32: This button doesn't seem to be working in the component guide - is that just a quirk of the component guide or is something broken?

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -0,0 +1,17 @@
...
+          <%= link_to(
+            item[:text],
+            item[:href],
+            class: 'govuk-header__link',
+            ) %>
+        </li>
+      <% end %>
+    </ul>
+  </nav>
+  <% end %>
```
vanitabarrett: Nitpick: indentation

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -12,35 +12,29 @@
...
+
+  header_classes = "gem-c-layout-header govuk-header gem-c-layout-header--#{environment}"
+  header_classes << " gem-c-layout-header--no-bottom-border" if remove_bottom_border
+  header_classes << " gem-c-layout-header--search-left" if search_left
 %>
 
-<% if search_left %>
-  <header class="gem-c-layout-header gem-c-layout-header--search-left govuk-header gem-c-layout-header--<%= environment %> <%= "gem-c-layout-header--no-bottom-border" if remove_bottom_border %>" role="banner" data-module="govuk-header">
+<header class="<%= header_classes %>"  role="banner" data-module="govuk-header">
     <div class="govuk-header__container <%= width_class %>">
```
vanitabarrett, line 22: Difficult to tell with the diff, but this indentation doesn't look quite right

app/views/govuk_publishing_components/components/layout_header/_search.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= render "govuk_publishing_components/components/search", {
+  id: "site-search-text",
+  label_text: "Search",
+  no_border: true,
+  no_margin: true
+} %>
+
```
vanitabarrett, line 7: Stray new line

app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -7,6 +7,8 @@ govuk_frontend_components:
   - header
 accessibility_excluded_rules:
   - landmark-banner-is-top-level # The header element can not be top level in the examples
+  - duplicate-id-aria # IDs will be duplicated in component examples list
```
vanitabarrett, line 10: This works, but it looks like we're getting an error about having multiple instances of `role=banner` on one page.

```
Problem: Document must not have more than one banner landmark (landmark-no-duplicate-banner)
```

I wonder if we can do something similar to ignore these errors?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
alex-ju: As far as I can see none of the public-facing application is using turbolinks. I suggest we remove this unless we find evidence of it being used.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -48,7 +49,7 @@
 
     <%= render "govuk_publishing_components/components/layout_header", {
       environment: "public",
-      search: true,
+      search: !without_search,
```
andysellick: I'm wondering if this could be rewritten to avoid the `!`. Would it be worth turning this on its head so the parameter is `hide_search` and then in the template itself the logic is `unless hide_search`?

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -3,6 +3,8 @@
   class_name = "gem-c-search--on-govuk-blue" if local_assigns[:on_govuk_blue].eql?(true)
   size ||= ""
   no_margin ||= false
+  no_border ||= false
+  class_name = "#{class_name} gem-c-search--no-border" if no_border
```
andysellick, line 7: I can't tell if I'm reading this right or not, but it seems like each of these options is mutually exclusive i.e. it's not possible to have the component be both e.g. large and without a border. Do we not need this behaviour?

app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -36,8 +36,8 @@ examples:
       - text: Hidden on desktop
         href: "item-3"
         show_only_in_collapsed_menu: true
-
   with_left_search_and_navigation:
+    description: This supports pages where the search appears onthe left with multiple navigation links on the right, such as the [How government works](https://www.gov.uk/government/how-government-works) page
```
andysellick, line 42: `onthe`

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -93,8 +93,6 @@
   }
 }
 
-
-
```
andysellick: Could this (and other similar linting commits) be merged with an appropriate commit? I wouldn't normally ask but there are a lot of commits in this PR.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
   }
 }
 
+.gem-c-layout-header--search-left {
```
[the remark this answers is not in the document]
maxgds: The rationale is that we're not trying to make design changes here, we're trying to swap out the architecture and this case needs to be handled. At the time this was coded we had no designer available, and there are definitely pages using this pattern.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -33,7 +35,7 @@
 
     <%= yield :head %>
   </head>
-  <body class="gem-c-layout-for-public govuk-template__body">
+  <body class="gem-c-layout-for-public govuk-template__body <%= body_classes %>">
```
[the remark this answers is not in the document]
maxgds: It's an equivalent of the existing gem-c-layout-for-admin which has been live for ages.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -2,6 +2,7 @@
   title ||= "GOV.UK - The best place to find government services and information"
   html_lang ||= "en"
   full_width ||= false
+  without_search ||= false
```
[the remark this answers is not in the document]
maxgds: Is that not what's happening here? By default you get search, if you don't want it you set "without_search" to be true?

app/views/govuk_publishing_components/components/_search.html.erb
danacotoran: I'm just looking at this currently and the options are not mutually exclusive. 
`class_name` builds on itself by concatenating the existing `class_name` and the new class if the conditions are met. So it is possible to have the component large, and with a border, and with no margin etc.

However, this seems like a slightly confusing way to achieve that outcome. We can refactor this to use a word array instead, which should improve readability.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -1,8 +1,9 @@
 <%
   meta ||= []
   navigation ||= []
+  with_border ||= false
```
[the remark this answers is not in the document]
danacotoran: I've done a GH search for `layout_footer` in the `alphagov` org and I could not find an instance where this component it is being used with the blue border at the top.
The footer on the LP has a blue border... but it looks like that footer is not a component, but is largely hardcoded in the template in `static`

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -18,12 +18,18 @@
...
+    <link rel="mask-icon" href="<%= asset_path 'govuk-mask-icon.svg' %>" color="#0b0c0c">
+    <link rel="apple-touch-icon" sizes="180x180" href="<%= asset_path "govuk-apple-touch-icon-180x180.png" %>">
+    <link rel="apple-touch-icon" sizes="167x167" href="<%= asset_path "govuk-apple-touch-icon-167x167.png" %>">
+    <link rel="apple-touch-icon" sizes="152x152" href="<%= asset_path "govuk-apple-touch-icon-152x152.png" %>">
+    <link rel="apple-touch-icon" href="<%= asset_path "govuk-apple-touch-icon.png" %>">
+
     <meta name="theme-color" content="#0b0c0c" />
     <meta name="viewport" content="width=device-width, initial-scale=1">
 
     <!-- The default og:image is added below :head so that scrapers see any custom metatags first, and this is just a fallback -->
```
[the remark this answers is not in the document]
danacotoran: Updated!

app/views/govuk_publishing_components/components/docs/layout_for_public.yml
danacotoran: I think this was copied over from `/govuk_publishing_components/components/docs/layout_for_admin.yml` which also has "TBD" as its accessibility criteria 😅

This raises a good point though –  should we define accessibility criteria for something that is essentially an HTML wrapper for other components/content?... and if yes, what would that criteria be (assuming that its contents have their own accessibility criteria)

andysellick: We shouldn't define criteria for anything that might be passed to the component, because in this case that would need the whole of WCAG, pretty much. But there might be some specific criteria around setting a `lang` attribute, having a sensible page title, that kind of thing?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -2,6 +2,8 @@
   title ||= "GOV.UK - The best place to find government services and information"
   html_lang ||= "en"
   full_width ||= false
+  body_classes ||= ""
+  main_classes ||= ""
```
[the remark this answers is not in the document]
danacotoran: @andysellick thank you for this comment, that's a really good point.

As of right now, I could not identify a need to have custom classes for body/main on this. I think if the need for custom styling arises, we can always extend the component to accept additional parameters. For now I've removed the ability to pass class names into the component.

spec/components/layout_header_spec.rb
```diff
@@ -67,4 +67,10 @@ def component_name
 
     assert_select ".gem-c-layout-header .gem-c-search"
   end
+
+  it "renders the search bar on the left when requested" do
+    render_component(environment: 'public', search: true, search_left: true)
+
+    assert_select ".gem-c-layout-header--search-left"
```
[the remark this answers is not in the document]
danacotoran: Hi @andysellick, I'm not 100% sure what you mean with this one. It seems to me that the tests cover all the cases but perhaps I am missing something very obvious.

Would you mind stating specifically what's missing or what can be done to make it more thorough? thanks! 🙏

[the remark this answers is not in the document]
danacotoran: Ah yes I see what you mean.
That commit does look deceivingly large, but as far as the `layout_header` component is concerned, seems to me that the new thing being introduced is the ability to have the search bar on the left, which is controlled through the `.gem-c-layout-header--search-left` class, which is what the test checks for. 
Another visual detail introduced by those changes is that when the search bar moves to the left, the navigation items also shift to the right. However this is done with CSS, using the same class. 

```
.gem-c-layout-header--search-left {
  .govuk-header__navigation--end {
    padding-top: 0;
    text-align: left;
  }
}
```

So I am not sure how we would check that the layout difference has been applied to the nav items, specifically.
I could update the wording on   `it "renders the search bar on the left when requested"` to also specify that nav items are moving to the right at the same time?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -49,6 +49,7 @@
     <%= render "govuk_publishing_components/components/layout_header", {
       environment: "public",
       search: true,
+      remove_bottom_border: full_width
```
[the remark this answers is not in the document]
danacotoran: Not really. I think the logic is that it should be true when `full_width` is true, and false when `full_width` is false. 
So basically it should always have a bottom border, unless the layout is full width.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Not sure I've implemented this exactly the way you suggested, but I've refactored a bit. 
So now we have `search: show_search`, where `show_search` defaults to `true` (and can be set to false via local assigns).

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
danacotoran: Still in the process of tidying up the history on this branch but will mark this as resolved as I've already squashed the particular commit this comment was about.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: I'd leave `govuk-width-container` on `main` as that seems to be the class we need to achieve the width constraints. 
The other `govuk-main-wrapper` class seems to add extra padding which we might not always be desirable or needed (it might become a head ache if we end up having to override it often. I reckon we can always add padding/margins to the inner elements instead, if necessary?).  

But then again I haven't got full context of why this was done the way it was.  Please let me know your thoughts on this?

spec/components/layout_header_spec.rb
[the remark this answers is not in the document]
andysellick: Fair enough then, as long as it's comprehensive.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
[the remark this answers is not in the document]
andysellick: Sorry, yes, realised later that `full_width` is of course a variable, not a string.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
andysellick: Sounds good, thanks.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
andysellick: I think we need both classes but I don't know if it's okay to combine them like this. Perhaps @36degrees can illuminate?

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
danacotoran: As of right now, it seems to work as expected for me! Perhaps it used to be broken and has been fixed in the meanwhile. 
Will be marking this as resolved but feel free to "unresolve" if there is still an issue for you.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: @andysellick I've checked with DS team and they have confirmed that this is not an appropriate use of these classes, so good shout there.

> They should be separate, because any phase banner, breadcrumbs or back link should sit inside the width container but outside of the main wrapper. There’s some guidance around this on https://design-system.service.gov.uk/styles/layout/#page-wrappers which might be useful.

On the basis of the quoted comment above, I've also added a new block that sits outside `main.govuk-main-wrapper`, but inside `.govuk-width-container`.  This introduces the ability to pass through breadcrumbs/backlink/any other element that should exist as a sibling of `.govuk-main-wrapper`.

Please feel free to flag this conversation as resolved if all looks good to you.

app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
```diff
@@ -1 +1,9 @@
 @import "govuk/components/footer/footer";
+
+.gem-c-layout-footer--border {
```
owenatgov, line 3: Could you nest this to tidy it up? So something like:

```
.gem-c-layout-footer {
    &--border {...}
    .govuk-footer__list {...}
}
```

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -18,8 +20,80 @@
...
+  }
+}
+
+.gem-c-layout-header--search-left {
+  .gem-c-header__menu-button {
+    margin-top: - govuk-spacing(7);
+    left: 0;
+  }
+
+  .gem-c-header__nav-wrapper {
```
owenatgov, line 39: Similar nesting comment with `.gem-c-header` on this line, line 54, 90, 99 and 107.

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -180,6 +180,12 @@ $large-input-size: 50px;
   }
 }
 
+.gem-c-search--no-border {
```
owenatgov, line 183: Again with nesting here and line 189.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,438 @@
...
+      <%= render "govuk_publishing_components/components/layout_footer", {
+        with_border: true,
+        navigation: [
+          {
+            title: "Prepare for Brexit",
+            columns: 2,
+            items: [
+              {
+                href: "/business-uk-leaving-eu",
+                text: "Prepare your business or organisation for Brexit",
```
owenatgov, line 75: 2 questions. These may well be annoying lack-of-context questions so soz in advance:

- Hard-coding this navigation seems a little risky. Is there no way for content designers to set this via publishing/whatever it's called?
- Failing point 1, is there a locale file for this content to go in?

danacotoran: Hi @owenatgov not annoying at all, you've actually brought up a good point here.

I think the reason why it's hard coded in here is because, well.... it was probably hard coded in the template it was copied over from. 😅

I am actually not sure if the components gem draws any content from any of the _publishing/whatever it's called_ 's. My feeling is that it does not.

See, generally we pass content through to the gem components from the rendering app (content which may or may not come from a publishing app, or the rendering app's localisation files, or be hardcoded in the rendering app etc).

I'm not totally sure how we should handle this case, but I can't see a reason why we couldn't move this data to the gem's locale file? I can check with team and get back to you on that.

andysellick: Currently there's no way of content designers changing the footer links, so we probably don't need to worry about that. We should definitely have text like this in a locale file though.

Something else I've noticed is that this is a lot of data to put in the view. It would be better to have that data object come from a helper method, to unclutter the template.

danacotoran: @andysellick  I've discussed this with @maxgds, @injms and @owenatgov and we've concluded that moving this content into a locale file sounds like a good idea. However, team thinks it needs scoping and discussion with the community, so we decided to leave it as is for now and create a card to address it in the future: https://trello.com/c/Jrb7JV0B

andysellick: Fair enough. Can we move all the data into a helper method though?

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
```diff
@@ -10,7 +10,11 @@ body: |
 display_preview: false
 display_html: true
 accessibility_criteria: |
-  TBD
+  The layout template must:
+
+    * have a `lang` attribute
+    * contain a [skip link](/component-guide/skip_link)
```
andysellick, line 16: Viewing this branch locally, the skip link when focussed looks different to how it does on the live site. Is this deliberate or an unexpected consequence of the preview? Live site:

Screenshot 2020-07-24 at 15 01 37

PR:

Screenshot 2020-07-24 at 15 00 55

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,438 @@
...
+      } %>
+    <% end %>
+    <div class="<%= "govuk-width-container" unless full_width %>"  id="content">
+      <%= yield :before_content %>
+      <main class="govuk-main-wrapper">
+        <%= yield %>
+      </main>
+    </div>
+
+    <%= render "govuk_publishing_components/components/feedback" %>
```
andysellick, line 63: The feedback component was recently changed to not restrict its own width anymore, instead allowing applications the ability to do that. The result is that it's gone full width in this component.

Screenshot 2020-07-24 at 15 03 33


The fix is to wrap this call in a `govuk-width-container` class div.

Screenshot 2020-07-24 at 15 07 01

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,438 @@
...
+                  data: {
+                    track_category: "footerClicked",
+                    track_action: "brexitLinks",
+                    track_label: "Prepare your business or organisation for the UK leaving the EU"
+                  }
+                }
+              },
+              {
+                href: "/prepare-eu-exit",
+                text: "Prepare for Brexit if you live in the UK",
```
andysellick, line 86: Really weird small/huge thing here - the focus state is clipped on this link. See this comparison with another link.

Screenshot 2020-07-24 at 15 09 41

I think this is something to do with the css columns stuff that's being used here. You can see it's doing some slightly odd wrapping of the element. 

Screenshot 2020-07-24 at 15 11 08


I thought it would be possible to fix by adding some margin or padding to the bottom of that list item, but the nature of the column wrapping means it just pushes down in the next column. Not sure what to suggest.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -0,0 +1,438 @@
...
+    </script>
+
+    <%= render "govuk_publishing_components/components/skip_link", {
+      href: "#content"
+    } %>
+
+    <%= render "govuk_publishing_components/components/cookie_banner" %>
+
+    <% unless omit_header %>
+      <%= render "govuk_publishing_components/components/layout_header", {
```
andysellick, line 49: Noticed that the header on mobile is different to the live site - the search box is shown by default, instead of being exposed by clicking the search icon. Is this intentional? Personally I like the new version, makes it much less complicated, but wanted to check.

Current:

Screenshot 2020-07-24 at 15 20 50

New:

Screenshot 2020-07-24 at 15 20 59

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
danacotoran: Having looked into it no, I don't think this is an unintended consequence of the preview, but rather a consequence of replacing the old skip link with [the Design System one](https://design-system.service.gov.uk/components/skip-link/default/index.html) which has different styling.

The skip link on the has live site some extra wrappers with styling from [govuk_template CSS](https://github.com/alphagov/govuk_template/blob/134e8b910c5175d5f94b1f5c48fdc342bb5065c2/source/assets/stylesheets/_accessibility.scss#L85).

I'd be keen to stick with the DS styling (I actually prefer it to what we've got on the live site currently 🤷🏻‍♀️) but let me check with a designer and get back to you on that.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Thanks for catching this layout issue!
I've done as suggested and wrapped the component with a `govuk-width-container` which seems to resolve the problem.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: I've raised this on the Design System slack channel as I believe this issue might be best fixed at the source rather than trying a localised fix on our side. 

It's a really weird one, well done spotting it. I've tried a number of potential solutions but nothing seems to work quite as intended? 😕  

Chrome displays this strange clipping behaviour, but additionally, it looks like other browsers are struggling with this in different ways. 
In Safari for instance, the focus state looks correct, but the links move up and down in a rather jarring way as you tab through the nav items https://gds.slack.com/files/U012K9PKS5N/F017XNZQ3EE/screen_recording_2020-07-29_at_14.52.09.mov

danacotoran: This has now been raised as an [issue](https://github.com/alphagov/govuk-frontend/issues/1888) on the Design System team's repo.
@36degrees mentioned that the team are already aware of the issues with CSS `columns` and planning to revisit the header and footer soon. This bug should be resolved as part of that work 🤞

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
danacotoran: @andysellick I have discussed this with @miaallers and she's happy to go with the Design System styling for this (can't see a reason why not).

andysellick: Good response. I have only one remaining question - is it supposed to have that thin black line near the bottom of the highlighted yellow?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: I'm sorry @andysellick, for some reason I missed this comment before. 
Yes, I am happy to move the data into a helper. However I'm not sure which helper this should be and how the data should be stored?... is there any example that you know of that I could use as a model for storing data into a helper method?

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
danacotoran: @andysellick I think this is a known Chrome bug –  I believe it was fixed in the Design System here https://github.com/alphagov/govuk-frontend/pull/1879/files

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
andysellick: I think probably a new one? Maybe in this directory: https://github.com/alphagov/govuk_publishing_components/tree/master/lib/govuk_publishing_components/presenters

Here's an example of a helper that returns a chunk of text to be reused in the accessibility criteria - I was thinking a similar method that simply returned the data for the footer.

https://github.com/alphagov/govuk_publishing_components/blob/master/app/models/govuk_publishing_components/shared_accessibility_criteria.rb

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
andysellick: If it was fixed in the Design system can it be fixed here too?

danacotoran: @andysellick Is it not fixed for you? 
If you can still see the outdated styling on your local version, it would probably help if you run `yarn install` in the project. This should update `govuk-frontend` to the latest version  (it should be 3.8.0).

andysellick: Ah yes that seems to have done it, thanks.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: @andysellick I've moved the footer data into a helper, hopefully it's correct. I did it before I saw your latest comment so it's not exactly the same format as the accessibility criteria example you shared, but it works.

When you get a moment would you mind giving this branch another once-over? Please let me know if I've missed any comments (sometimes they'll disappear and I'm not sure how to find them again 😅) or if you have any more feedback on this.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -0,0 +1,379 @@
...
+                  track_category: "footerClicked",
+                  track_action: "footerLinks",
+                  track_label: "Transparency and freedom of information releases",
+                },
+              },
+            },
+          ],
+        },
+      ].freeze
+      FOOTER_META = {
```
andysellick, line 342: Nitpick: I'd put a newline between these two things.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
andysellick: Helper looks good, I think. Good job. Will do.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: This should have been resolved now (asked for design input and it should stay as on the live site)

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: We're pulling the assets from GOVUK Frontend 👍

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: As far as I can tell the assets are now being imported from GOVUK Frontend

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Based on @alex-ju suggestion, I have removed the reference to `turbolinks` as I could not find any evidence that it was doing anything.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Honestly not 100% sure why this was removed to begin with. 
But seeing as this is used on the live site and that I can't find any evidence that it is _not_ useful, I think the lowest risk way to proceed with this is keep it for now. It can always be removed in the future if we find no need for it.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
vanitabarrett: Thanks for clarifying @danacotoran - do we need to rename? I can see `govuk-opengraph-image.png` in the GOVUK Frontend repo

danacotoran: @vanitabarrett I think you're looking at the outdated version of `_layout_for_public.html.erb`.
Looks like we renamed this to call for `govuk-opengraph-image.png` at some point  🤔https://github.com/alphagov/govuk_publishing_components/blob/3d67d5a36428168100207e0d441834c280e2d6de/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L34

vanitabarrett: D'oh, makes sense! Thanks 🙂

PR #1270, opened 2020-01-28
Breadcrumb truncation on mobile devices 

RobBond-GDS:
Mobile breadcrumb to be show collapsed to home and parent page on mobile devices. Alignment for long taxons to wrap under home item and not offset (done by moving chevron to "after" and not "before" css element.
 
## What
Apply mobile enhancement designs to the GOV.UK Breadcrumb component, specifically:

- Collapse breadcrumb to show only the first and last item (the parent item) on devices less than 640px wide
- Override and Increase the touch target padding on mobile devices 
- Ensure long taxons wrap appropriately 
- Override chevrons to display ":after" a breadcrumb item and not ":before"

## Why
The current breadcrumb shows the full taxonomy path. This causes the breadcrumb to wrap onto multiple lines, taking up valuable screen space and making it less usable on mobile due to the touch targets being too close together. The current placing of the chevron in the code breaks up the breadcrumb in inappropriate places

The reasoning is taken from this snippet in the trello card:

> This [article by NNG](https://www.nngroup.com/articles/breadcrumbs/) lays out the pros and cons of showing a full breadcrumb path on mobile. In short, the drawbacks of having a multi-line breadcrumb, and the need to save space on mobile, make it less useful to users. By truncating the breadcrumb we at least allow users to go backwards by one level, or home.

Designs have been discussed with Tim Paul and advised to proceed with the intention to feedback into the design system at a later date.
 

Screenshot 2020-02-05 at 14 57 28
Before update of breadcrumb

Screenshot 2020-02-05 at 14 56 20
After update of breadcrumb

Screenshot 2020-02-10 at 13 46 09
Long breadcrumb shown aligned to the left with arrows "after"


app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -20,3 +20,55 @@
 .gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:before {
   border-color: govuk-colour("white");
 }
+
+.govuk-breadcrumbs {
+  @include media-down(mobile) {
```
kr8n3r: I know this is WIP so not a review, merely suggestion to use `govuk-media-query` mixin

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -20,3 +20,55 @@
...
+    margin-bottom:0px;
+    &.govuk-breadcrumbs--collapse-on-mobile .govuk-breadcrumbs__list-item {
+      position: absolute;
+      left: -9999em;
```
kr8n3r: maybe use `govuk-visually-hidden` mixin

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: good idea...will do!

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: good idea...will do!

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -22,16 +22,12 @@
 }
 
 .govuk-breadcrumbs {
-  @include media-down(mobile) {
+  @include govuk-media-query($from: mobile, $until: tablet) {
```
kr8n3r: you can just use `$until: tablet` here

vanitabarrett:
@RobBond-GDS Can you remove the node modules from this commit so it's easier to review changes?

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,10 +13,109 @@
...
+  padding-left: 0;
+  margin-right: 20px;
+
+  &:before {
+    display: none;
+  }
+
+  &:after {
+    content: "";
+    display: inline-block;
```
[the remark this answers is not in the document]
vanitabarrett: I'm not sure the wrapping is working as intended. At 320px:

Screenshot 2020-02-05 at 11 54 32

The code change I suggested above seems slightly better, although it doesn't align to Home:
Screenshot 2020-02-05 at 11 59 29

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,32 @@
 @import "govuk/components/back-link/back-link";
+
+
+@include govuk-media-query($from: mobile, $until: tablet) {
```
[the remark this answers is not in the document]
RobBond-GDS: added the back link in error will remove and make separate PR

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
```diff
@@ -1,17 +1,17 @@
 <%
   breadcrumbs ||= []
-  collapse_on_mobile ||= false
+  collapse_on_mobile ||= true
   inverse ||= false
-  collapse_class =  collapse_on_mobile && breadcrumbs.any? { |crumb| crumb[:is_page_parent] } ? "gem-c-breadcrumbs--collapse-on-mobile" : ""
+  # collapse_class =  collapse_on_mobile && breadcrumbs.any? { |crumb| crumb[:is_page_parent] } ? "gem-c-breadcrumbs--collapse-on-mobile" : ""
```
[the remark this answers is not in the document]
RobBond-GDS: will remove comments

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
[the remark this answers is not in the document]
RobBond-GDS: have reduced the code for overrides

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,10 +13,109 @@
...
   color: govuk-colour("white");
 }
 
-.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:before {
+.gem-c-breadcrumbs--inverse.govuk-breadcrumbs--collapse-on-mobile .govuk-breadcrumbs__list-item:after {
   border-color: govuk-colour("white");
 }
+
+.govuk-breadcrumbs--collapse-on-mobile .govuk-breadcrumbs__list-item {
+
```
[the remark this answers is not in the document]
RobBond-GDS: removed stray line

node_modules/axe-core/CHANGELOG.md
```diff
@@ -0,0 +1,905 @@
+# Changelog
```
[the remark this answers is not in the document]
RobBond-GDS: removed node_modules folder

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,10 +14,49 @@
...
+  margin-left: 0;
+  padding-left: 0;
+  margin-right: 20px;
+  @include chevron($direction: "right", $color: govuk-colour("dark-grey"));
```
vanitabarrett: I'm still unsure we need to re-write the arrow/chevron part of the component.

We could do something like this to make the links wrap between mobile and tablet:

```
  @include govuk-media-query($from: mobile, $until: tablet) {
    &:last-of-type {
      position: absolute;
      margin-right: 20px;
    }
  }
```

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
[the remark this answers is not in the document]
RobBond-GDS: Removed from this PR, will create new PR with relevant ticket

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: the mixin is so that same code could be used for the "back link" ticket and have same styling as breadcrumb 

https://trello.com/c/ZYJ7nnYt/21-make-back-icon-more-consistent-on-mobiles

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,10 +14,49 @@
...
+      margin-right: 0;
+      max-width: calc(80% - 14px);
+    }
+
+    &:first-of-type {
+      display: inline-block;
+      padding-left: 0;
+    }
+
+    &.gem-c-breadcrumbs--parent {
```
andysellick: Where is this class used? I can't see it in the review app.

app/assets/stylesheets/govuk_publishing_components/components/mixins/_chevron.scss
```diff
@@ -0,0 +1,37 @@
+// An angled chevron
+// $direction - direction the chevron will point to: up, down, left, margin-right
```
andysellick: Typo? `margin-right`?

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
```diff
@@ -1,17 +1,16 @@
 <%
   breadcrumbs ||= []
-  collapse_on_mobile ||= false
+  collapse_on_mobile ||= true
```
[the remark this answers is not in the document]
andysellick: Related question - if collapse on mobile is going to be the default behaviour, in what situation would we want to change that?

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
```diff
@@ -1,17 +1,16 @@
...
   invert_class = inverse ? "gem-c-breadcrumbs--inverse" : ""
   breadcrumb_presenter = GovukPublishingComponents::Presenters::Breadcrumbs.new(breadcrumbs, request.path)
 %>
-
 <script type="application/ld+json">
   <%= raw breadcrumb_presenter.structured_data.to_json %>
 </script>
 
-<div class="gem-c-breadcrumbs govuk-breadcrumbs <%= invert_class %>" data-module="track-click">
+<div class="gem-c-breadcrumbs govuk-breadcrumbs <%= invert_class %> <%= collapse_class %>" data-module="track-click">
```
andysellick: This is getting a little untidy, here's an example of an alternative way of doing this: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_textarea.html.erb#L23-L25

app/assets/stylesheets/govuk_publishing_components/components/mixins/_chevron.scss
RobBond-GDS: Yep typo! will amend accordingly

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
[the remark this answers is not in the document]
RobBond-GDS: On further discussion the collapsing of the breadcrumb to first and last is the only behaviour. No option to change so true setting to become the only setting making things much simpler, will clean up the logic!

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -13,10 +14,49 @@
...
+  @include govuk-media-query($until: tablet) {
+    display: none;
+    margin-bottom: 0;
+    padding: 14px 0;
+
+    &:last-of-type {
+      display: inline-block;
+      padding-left: 0;
+      margin-right: 0;
+      max-width: calc(80% - 14px);
```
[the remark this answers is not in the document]
RobBond-GDS: as the breadcrumb collapses to first and last item, the last item if long is expected to wrap neatly under itself, having it with the max-width allows this

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: This was left from the original code prior to this work

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
RobBond-GDS: On further discussion the collapsing of the breadcrumb to first and last is the only behaviour. No option to change so true setting to become the only setting making things much simpler, will clean up the logic!

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -1,6 +1,7 @@
 @import "govuk/components/breadcrumbs/breadcrumbs";
+@import "mixins/chevron";
 
-.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item .govuk-breadcrumbs__link {
+.gem-c-breadcrumbs--inverse.govuk-breadcrumbs--collapse-on-mobile .govuk-breadcrumbs__list-item .govuk-breadcrumbs__link {
```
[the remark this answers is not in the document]
RobBond-GDS: On further discussion, the collapsing of the breadcrumb to first and last is the only behaviour. Therefore additional selectors for conditional display are no longer need, so have removed an simplified

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,44 @@
...
+  @include govuk-media-query($until: tablet) {
+    display: none;
+    margin-bottom: 0;
+    padding: 14px 0;
+
+    &:last-of-type {
+      display: inline-block;
+      padding-left: 0;
+      margin-right: 0;
+      max-width: 76%;
```
andysellick: This still seems to be assuming that the first item in the breadcrumb is no more than the remaining width because it only contains the word 'Home'.

This isn't the case, as shown in the component guide itself (and on [pages like this](https://www.gov.uk/service-manual/helping-people-to-use-your-service)). Example below (I've artificially extended the length of the last item's text, but something like this could happen).

Screenshot 2020-02-07 at 08 25 37

If you're going to set the width of the last item, you'll have to set the width of the first item as well.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,44 @@
   color: govuk-colour("white");
 }
 
-.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:before {
+.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:after {
   border-color: govuk-colour("white");
 }
+
+.govuk-breadcrumbs .govuk-breadcrumbs__list-item {
```
andysellick: I don't think this needs to be nested like this - will it work as `.govuk-breadcrumbs__list-item {`?

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,44 @@
...
+  margin-left: 0;
+  padding-left: 0;
+  margin-right: 20px;
```
andysellick: Why are we overriding so much of `govuk-frontend` here? Their breadcrumbs seem to work perfectly well.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
andysellick: Can it be removed?

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
```diff
@@ -3,6 +3,8 @@ description: "Navigational breadcrumbs, showing page hierarchy"
 body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
   Links have tracking data but links to the homepage (any link with a url of "/") will be tracked separately as `homeLinkClicked`
+
+  Note: By default on mobile devices, in portrait orientation, the breadcrumb will collapse to the Home and Last Sub Section
```
andysellick: Suggested change: `Only the first and last items in the breadcrumb will be displayed on a page width less than 640px` (or whatever it is).

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,44 @@
...
+  @include chevron($direction: "right", $color: govuk-colour("dark-grey"));
+
+  &:before,
+  &:last-of-type:after {
+    display: none;
+  }
+
+  @include govuk-media-query($until: tablet) {
+    display: none;
```
andysellick: Not that I want to throw another spanner in the works, but it's just occurred to me that the change being proposed here reduces the functionality of the breadcrumb to only a link to the homepage, whereas previously it provides both a visual representation of where the user is and navigation options for moving 'up'. 

Has there been any user research into this? I'm concerned that this reduction, while making better use of the vertical space, reduces the usefulness of the breadcrumb significantly. One alternative would be to do this change as proposed, but when the user clicks on any part of the breadcrumb, to show the entire thing?

vanitabarrett: @andysellick I think the commit/PR name is a bit misleading - it's actually showing the Home and parent page on mobile, rather than the home and current page. Is that correct @RobBond-GDS ?

andysellick: That would be better, yes, if that's the case then the PR description needs to be updated as well, along with the documentation for this component.

RobBond-GDS: Yes it is home and parent page, I have updated the PR description and included the snippet for the research and reasoning as mentioned in the design doc on the trello card

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: Removed this and is fine

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: Overrides are to facilitate touch target and chevron origin switching from "before" to "after"

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
RobBond-GDS: Update docs as per suggestion

Gemfile.lock
```diff
@@ -317,5 +317,8 @@ DEPENDENCIES
   webmock (~> 3.6.0)
   yard
 
+RUBY VERSION
+   ruby 2.6.5p114
```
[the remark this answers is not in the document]
RobBond-GDS: I believe they got added automatically on local bundle install...not sure

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: Removed

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: added max-width to the first of type as well as the last

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
```diff
@@ -3,6 +3,8 @@ description: "Navigational breadcrumbs, showing page hierarchy"
 body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
   Links have tracking data but links to the homepage (any link with a url of "/") will be tracked separately as `homeLinkClicked`
+
+  Note: Only the first and last items in the breadcrumb will be displayed on a page width less than 640px
```
andysellick: Could this be expanded slightly to explain that this will show the parent page?

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,37 @@
...
+      display: inline-block;
+      padding-left: 0;
+      margin-right: 0;
+      max-width: 72%;
+    }
+
+    &:first-of-type {
+      display: inline-block;
+      padding-left: 0;
+      max-width: 22%;
```
andysellick: Sorry, but this still doesn't work properly when the text in the first breadcrumb is something other than 'Home'.

Screenshot 2020-02-12 at 11 18 47

You might have to use absolute positioning on the chevron to crack this.

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
andysellick: What's the plan with making the change so the last item isn't the current page? I assume it'll need deploying at the same time but the PR doesn't mention it.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -17,6 +18,37 @@
   color: govuk-colour("white");
 }
 
-.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:before {
+.gem-c-breadcrumbs--inverse .govuk-breadcrumbs__list-item:after {
```
andysellick: I'm still uncertain as to the value of shifting the chevron from `before` to `after`. In the screenshot in the PR description (assuming this is the 'after' state?) I'd be concerned about users not understanding that the link wrapped underneath (with no chevron) is part of the breadcrumb.

RobBond-GDS: That is the design that was decided on when discussions were had with TIm Paul

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
RobBond-GDS: have tweaked again...

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
andysellick: Can you provide more detail of this? What was the justification?

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
andysellick: Seems to be better now.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
andysellick: My confusion is that in the situation we're talking about, when there's only two breadcrumbs, there's only one chevron between them. To my mind it doesn't therefore matter whether it's a before or an after element - but my concern is that changing it from one to the other in order to override the styles from govuk-frontend adds extra lines of code that I don't understand the justification for. Does that make sense?

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
andysellick: Am I right in thinking this has already been deployed?

alex-ju:
> I'm still unsure more generally on overriding the design system component so much, it doesn't feel like something we should be doing. I'd welcome the views of other frontend devs on this though, perhaps @alex-ju @andysellick ?

Sorry for missing this mention for so long. I do agree that before merging this piece of code it should be raised with the Design System team. This is a component we inherit from them and I would be reluctant to change it unless agreed with them and/or making the change upstream.

andysellick:
@RobBond-GDS @alex-ju I think this PR may have been superseded by other PRs now? If so can this be closed? Or am I confused?

RobBond-GDS:
There was discussion and agreement with Design System/Tim Paul and Stephen
McArthy.

This PR, however, is now redundant and i shall be pulling it, the work that
was done was split into smaller PR's

https://github.com/alphagov/govuk_publishing_components/pull/1290
https://github.com/alphagov/govuk_publishing_components/pull/1296
https://github.com/alphagov/govuk_publishing_components/pull/1298

PR #1276, opened 2020-02-03
Make maxlength or maxwords required

andysellick:
## What
Make number passed to the character count component (either `maxlength` or `maxwords`) required, render nothing if not passed.

## Why
I just tried to use this component and got the maxlength parameter in the wrong place, but instead of failing gracefully the component loaded and then broke, which was briefly very confusing. If one of these parameters is required, let's make it required.

## Visual Changes
No changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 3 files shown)


andysellick:
@ChrisBAshton changelog entry added, thanks.

PR #1278, opened 2020-02-04
Alternative header layout

maxgds:
## What
Adds a new layout alternative to the layout header.
https://trello.com/c/vMiHP2yB/194-build-header-layout-for-search-on-left

## Why
This is needed for pages in whitehall such as https://www.gov.uk/government/how-government-works

Notes: The layout change meant that the skeleton of the html needed to change. To avoid repeating blocks they are now in partials. I tried to do this with content blocks but it broke the component's examples page as each render of the content block appended itself to the previous ones before displaying. 

This is being pushed back into a non-master branch so that individual chunks of work can be reviewed and monitored on Trello rather than pushing to one huge PR.

## Screenshot
Screenshot 2020-02-04 at 15 08 10


kr8n3r:
is it known that the nav menu is not visible on mobile?

maxgds:
> is it known that the nav menu is not visible on mobile?

No, it wasn't... it was working previously I must have broken something. Looking into it.

maxgds:
> is it known that the nav menu is not visible on mobile?

Fixed now, I'd accidentally optimised away the div that set the relative position.

maxgds:
> A few initial visual discrepancies (I haven't dived into the code yet). 

The layout at mobile is different anyway - the search bar is full width deliberately rather than half width to some arbitrary point where it gets hidden on live. Likewise the menu items don't try to cram into 2 columns. I thought it made more sense for the menu button to appear on the left than on the right if we're not putting the search over there. If you think it should be on the right, why?

I'll look into the separated icon, though.

andysellick:
I'm not sure which side it should be on, I was just pointing out the difference.

Another slightly odd behaviour - if you expand the menu on mobile then switch to desktop the whole of the menu is pushed down slightly. Obviously this is a niche use case and doesn't represent a barrier to anything, it's just a little odd.

header

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -1,10 +1,12 @@
+
```
andysellick: Nitpick: can we remove this blank line? I think it gets rendered in the source.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -13,63 +15,49 @@
...
-    </nav>
-    <% end %>
-
-    </div>
-    </div>
-
-    <% if search %>
-        <div class="govuk-grid-column-one-third gem-c-layout-header__search">
+<% if search_left %>
+  <header class="gem-c-layout-header gem-c-layout-header--search-left govuk-header gem-c-layout-header--<%= environment %> <%= "gem-c-layout-header--no-bottom-border" if remove_bottom_border %>" role="banner" data-module="govuk-header">
```
andysellick: This is getting quite complex - could it be replaced with something easier to read, for example: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_textarea.html.erb#L23-L25

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
...
+    .govuk-header__navigation-item {
+      padding: 0 0 govuk-spacing(1) 0;
+    }
+
+    .gem-c-header__logo {
+      margin-top: 0;
+    }
+
+    .gem-c-header__content {
+      @include govuk-grid-column(two-thirds, $at: desktop);
```
vanitabarrett: You don't need `$at: desktop` here as this code is already within a media query

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -0,0 +1,25 @@
...
+        ></path>
```
andysellick: I appreciate this isn't your change, but the code is broken up a bit oddly here - could this trailing `>` be reconnected to the previous line?

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
...
+      padding-left: govuk-spacing(6);
+      padding-right: govuk-spacing(1);
+    }
+
```
vanitabarrett: There seems to be some extra space here

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
   }
 }
 
+.gem-c-layout-header--search-left {
+
```
vanitabarrett: Extra space

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -0,0 +1,17 @@
...
+  <nav class="gem-c-header__nav">
+    <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
+      <% navigation_items.each_with_index do |item, index| %>
+        <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
+          <%= "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu] %>">
+          <%= link_to(
+            item[:text],
+            item[:href],
+            class: 'govuk-header__link',
+            ) %>
```
andysellick: Nitpick: I think this should be unindented by one.

app/views/govuk_publishing_components/components/_layout_header.html.erb
vanitabarrett: Extra new line

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -13,63 +15,49 @@
...
-      </ul>
-    </nav>
-    <% end %>
-
-    </div>
-    </div>
-
-    <% if search %>
-        <div class="govuk-grid-column-one-third gem-c-layout-header__search">
+<% if search_left %>
```
vanitabarrett: What's the reasoning behind this `if` statement? It looks like there's quite a bit of shared code between the `if` and the `else` - would it not be possible to conditionally add the `gem-c-layout-header--search-left` to the existing HTML and work from there?

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -13,63 +15,49 @@
...
+        <div class="govuk-grid-column-full govuk-grid-column-one-third-from-desktop gem-c-layout-header__search">
           <%= render "govuk_publishing_components/components/search", {
             id: "site-search-text",
             label_text: "Search",
             no_border: true,
             no_margin: true
           } %>
         </div>
-    <% end %>
+        <div class="govuk-header__content gem-c-header__content govuk-grid-column-full govuk-grid-column-two-thirds-from-desktop">
```
vanitabarrett: Is the `govuk-grid-column-two-thirds-from-desktop` class needed? It looks like we're already adding it in the CSS?

app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -36,6 +36,29 @@ examples:
       - text: Hidden on desktop
         href: "item-3"
         show_only_in_collapsed_menu: true
+
```
vanitabarrett: Stray new line

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -0,0 +1,25 @@
+
```
vanitabarrett: Extra new line

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
...
+  .gem-c-header__menu-button {
+    position: absolute;
```
andysellick: Looking at the CSS hierarchy it looks like absolute positioning is already defined for this element.

Screenshot 2020-02-06 at 13 29 23

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
...
+  .govuk-header__navigation--end {
+    padding-top: 0;
+    text-align: left;
+  }
+
+  .gem-c-layout-header__search {
+    margin-bottom: govuk-spacing(2);
+  }
+
+  .gem-c-layout-header__search {
```
andysellick: Is this declaration the same as line 53? Could these two be merged?

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -28,6 +28,63 @@
...
+    }
+
+    .gem-c-header__content {
+      @include govuk-grid-column(two-thirds, $at: desktop);
+      padding-left: govuk-spacing(6);
+      padding-right: govuk-spacing(1);
+    }
+
+  }
+
```
andysellick: Nitpick: I'd remove this blank line and line 83.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -51,6 +108,10 @@
   }
 }
 
+.govuk-header__navigation--end {
+  text-align: right;
+}
```
andysellick: Where is this text alignment used? It seems to get overridden.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -13,63 +15,49 @@
...
+          <%= render "govuk_publishing_components/components/layout_header/navigation_items", navigation_items: navigation_items  %>
+        </div>
+      </div>
+    </div>
+  </header>
 
-  </div>
-</header>
+<% else %>
+  <header class="gem-c-layout-header govuk-header gem-c-layout-header--<%= environment %> <%= "gem-c-layout-header--no-bottom-border" if remove_bottom_border %>" role="banner" data-module="govuk-header">
```
andysellick: This bit could also be clearer with this approach: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_textarea.html.erb#L23-L25

app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -36,6 +36,29 @@ examples:
...
+  with_left_search_and_navigation:
+    data:
```
andysellick: Can I suggest a description be added here, to explain where this variation on the header will be used?

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -0,0 +1,25 @@
...
+        ></path>
+        <image src="<%= asset_path('govuk-logotype-crown.png') %>" class="govuk-header__logotype-crown-fallback-image"></image>
```
andysellick: This isn't your change, but shouldn't this be an `img` tag? I don't think there's such a thing as an `image` tag. I'm still recovering from a loss of sleep earlier in the week, so I might be crazy.

app/views/govuk_publishing_components/components/_layout_header.html.erb
maxgds: It's because the div structure needs to be different with the nav rendered last if the search is on the left, otherwise the menu button for the mobile nav is hard to place properly. Looking at it, though, I can move more of it outside the if/else to reduce the repetition.

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
maxgds: https://lynn.ru/examples/svg/en.html

andysellick: That should be documented somewhere.

kr8n3r: can take inspiration from https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/header/template.njk#L34

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -0,0 +1,25 @@
+
+<div class="govuk-header__logo gem-c-header__logo">
+  <a href="/" class="govuk-header__link govuk-header__link--homepage">
+    <span class="govuk-header__logotype">
+      <svg role="presentation" focusable="false" class="govuk-header__logotype-crown" xmlns="http://www.w3.org/2000/svg" viewbox="0 0 132 97" height="32" width="36">
```
kr8n3r: role=presentation is be replaced with aria-hidden=true
see https://github.com/alphagov/govuk-frontend/pull/1724

maxgds:
@kr8n3r @andysellick @vanitabarrett @injms 
I believe I've addressed the multiple comments and concerns on the initial version of this branch. Would be good to get feedback on it as it stands now. Updates are broken down into separate commits.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -38,11 +85,16 @@
...
-  white-space: nowrap;
   margin-top: govuk-spacing(1);
   margin-bottom: govuk-spacing(3);
+
+  @include govuk-media-query($from: desktop) {
+    white-space: nowrap;
+  }
 }
 
+
```
vanitabarrett: Additional new lines here

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
vanitabarrett: Looks like this new line is still here... sorry to be picky!

vanitabarrett:
Is this placement of the menu link here correct? Could the spacing below the Menu link match the spacing above, as it looks quite close to the blue border?

Screenshot 2020-02-12 at 10 07 44

vanitabarrett:
Not sure if this is a problem which existed before, but it looks like a search bar on it's own needs some bottom spacing on smaller devices so it doesn't touch the blue bar

Screenshot 2020-02-12 at 10 08 15

injms:
There are a couple of false positives with the automated accessibility tests caused by the same components being repeated on the example page - it's not a problem with the component, as this is (hopefully) something that wouldn't happen in the real world use.

We can turn off the checks by adding `duplicate-id-aria` and `form-field-multiple-labels` to the `accessibility_excluded_rules` in the [layout header's YAML file](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/docs/layout_header.yml#L8).

An alternative fix for the `form-field-multilple-labels` problem is to make the id of the search component unique by appending `#{SecureRandom.hex(4)}` to the id - which is [what's done on the accordion component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_accordion.html.erb#L4).

maxgds:
> Is this placement of the menu link here correct? Could the spacing below the Menu link match the spacing above, as it looks quite close to the blue border?

Fixed the spacing issues and the stray new lines introduce/overlooked previously.

maxgds:
> There are a couple of false positives with the automated accessibility tests caused by the same components being repeated on the example page

I've ignored those checks. The accordion example is different - you could potentially have multiple accordions on a single page, but you shouldn't have multiple headers at any time, so it needs the randomising approach whereas we don't here.

Issue #1279, opened 2020-02-05
Search component: input with value + and inline label overlap


kr8n3r:
When an input in the search component has a value and an inline label, they overlap until the input receives focus.

See screenshot
<img width="916" alt="Screenshot 2020-02-05 at 14 32 26" src="https://user-images.githubusercontent.com/3758555/73851164-eff42980-4824-11ea-9c2f-7f80b26a7820.png">

See it in action https://www.gov.uk/search/all?keywords=components

Should we move to a hidden label and placeholder text for this arrangement?

PR #1280, opened 2020-02-06
WIP - Proposal: Search inline label rework

kr8n3r:
## What
Replace current absolutely-positioned inline label with placeholder + visually hidden label

## Why
See issue #1279 

## Visual Changes
None

| [Before](https://components.publishing.service.gov.uk/component-guide/search/preview)  | [After](https://govuk-publis-search-inl-12zvwo.herokuapp.com/component-guide/search/preview) |
| ------------- | ------------- |
| components publishing service gov uk_component-guide_search_preview   | 0 0 0 0_3212_component-guide_search_preview |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -8,25 +8,11 @@ $large-input-size: 50px;
 
 .gem-c-search__label {
   @include govuk-font($size: 19, $line-height: $input-size);
-  display: block;
+  color: $govuk-secondary-text-colour;
 
   h1 {
     @include govuk-font($size: 19, $line-height: $input-size);
     margin: 0;
-  }
-
-  .js-enabled & {
-    position: absolute;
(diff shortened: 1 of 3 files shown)


andysellick:
What's our stance on placeholder text? I thought we were against it for accessibility reasons.

kr8n3r:
on it's own it's not accessible. that's why we'd include a visually hidden label.
prior art: design system website search

nickcolley:
I believe the original approach was down to browser support as placeholders are only supported in Internet Explorer 10 and above: https://caniuse.com/#feat=input-placeholder

We should also consider the contrast of the text in the placeholder, it is possible to style this, should we set it to a colour in the palette or leave it to the browser? This sort of decision might be worth pushing upstream if there's confidence around it.

kr8n3r:
some A11Y work on floating label
https://a11yproject.com/patterns/
https://github.com/a11yproject/a11yproject.com/issues/505

kr8n3r:
closing as there is no demand for this

PR #1286, opened 2020-02-11
Add visually hidden text to share links component

### LINKED ISSUE #1064: Add visually hidden text to share links component
We've [implemented a workaround within Collections](https://github.com/alphagov/collections/blob/2fd99c9f1456301cec48cbe78f35fc0a4e34aa8e/app/views/brexit_landing_page/_common.html.erb#L35) for this, but it should be an option in the component itself.

kr8n3r:
## What
- Adds an `hidden_text` link to specify custom visually hidden link text
- Adds a default 'Share on` visually hidden text to each share link if `hidden_text` value is not provided

## Why
Improve accessibility and remove the need for [workarounds](#1064 )
Fixes #1064 

## Visual Changes
No visual changes

https://govuk-publis-share-link-crmidx.herokuapp.com/component-guide/share_links/with_custom_visually_hidden_text

### DIFF
--- a/app/views/govuk_publishing_components/components/_share_links.html.erb
+++ b/app/views/govuk_publishing_components/components/_share_links.html.erb
@@ -24,6 +24,16 @@
 
     <ul class="gem-c-share-links__list">
       <% links.each do |link| %>
+        <% link_text = capture do %>
+          <span class="govuk-visually-hidden">
+            <% if link[:hidden_text] %>
+              <%= link[:hidden_text] %>
+            <% else %>
+              Share on
+            <% end %>
+          </span>
+          <%= link[:text] %>
+        <% end %>
         <li class="gem-c-share-links__list-item">
           <%
             if track_as_sharing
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/share_links.yml
```diff
@@ -65,6 +65,16 @@ examples:
           icon: 'facebook'
         }
       ]
+  with_custom_visually_hidden_text:
+    data:
```
andysellick: If I'm reading the code right, it looks like if no visually hidden text is provided it defaults to 'Share on'. I think that's worth documenting here with a description, otherwise we might end up with share links that duplicate it unknowingly e.g. '(Share on) Share on Facebook'.

app/views/govuk_publishing_components/components/_share_links.html.erb
```diff
@@ -24,6 +24,16 @@
 
     <ul class="gem-c-share-links__list">
       <% links.each do |link| %>
+        <% link_text = capture do %>
+          <span class="govuk-visually-hidden">
+            <% if link[:hidden_text] %>
+              <%= link[:hidden_text] %>
+            <% else %>
+              Share on
```
andysellick: Having made my previous comment, I'm wondering should we default to having hidden text? What's the benefit? (we're making more work for ourselves, once this gets rolled out we'll need to modify any apps that have a workaround or include 'Share on' in their link text already, which we otherwise could do at our leisure).

kr8n3r: the benefit is that we're actually improve what's being read out. quick look at the usage of the component that in most cases the link text is just the service (Twitter, Facebook).

app/views/govuk_publishing_components/components/docs/share_links.yml
kr8n3r: description added

PR #1288, opened 2020-02-12
Downgrade 'rouge' to a development_dependency

ChrisBAshton:
## What
Downgrade 'rouge' to a dev_dependency

## Why
Upgrades to 'rouge' open PRs in repositories that use govuk_publishing_components.
Example: https://github.com/alphagov/whitehall/pull/5342

But this isn't actually used by the components themselves, so should be stripped
out of the gem.

## Visual Changes
N/A. Heroku app still shows syntax highlighting as expected.

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -8,7 +8,6 @@ PATH
       plek
       rails (>= 5.0.0.1)
       rake
-      rouge
       sassc-rails (>= 2.0.1)
       sprockets (< 4)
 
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -100,7 +99,7 @@ GEM
       rubocop (~> 0.64)
       rubocop-rspec (~> 1.28)
       scss_lint
-    govuk_app_config (2.0.2)
+    govuk_app_config (2.0.3)
```
ChrisBAshton: This slipped in probably because someone forgot to run `bundle install` before their change got merged to master.

maxgds:
Hi, @ChrisBAshton I know this is closed and merged already, but it is causing issues for me working locally. When I run calendars, pointed at a local instance of the components gem then try to view the component guide examples (eg http://localhost:3011/component-guide/layout_for_public) they error with a message about being unable to find rouge. Changing it from a dev dependency and bundle installing got it working again locally. Does this perhaps need reverting?

ChrisBAshton:
I didn't realise apps use govuk_publishing_components as a component viewer, as well as for its components. Looks like this change would affect lots of apps: https://github.com/search?q=mount+GovukPublishingComponents%3A%3AEngine&type=Code

I'll raise a PR to revert.

PR #1296, opened 2020-02-14
Mobile breadcrumb to wrap long taxon and correctly align chevron top

RobBond-GDS:
On mobile devices update breadcrumb behaviour for long taxons to wrap and appear aligned with horizontally. Ensure chevron appears aligned correctly. As per trello card:

## What
Apply mobile enhancement designs to the GOV.UK Breadcrumb component, specifically:

- Breadcrumb to display first and last items horizontally on devices less than 640px wide
- Ensure the chevron is aligned to the top of the horizontal breadcrumb

## Why
The reasoning is taken from this snippet in the trello card:

> This [article by NNG](https://www.nngroup.com/articles/breadcrumbs/) lays out the pros and cons of showing a full breadcrumb path on mobile. In short, the drawbacks of having a multi-line breadcrumb, and the need to save space on mobile, make it less useful to users. By truncating the breadcrumb we at least allow users to go backwards by one level, or home.

Long taxons currently wrap beneath the first item.

## Visual Changes

Screenshot 2020-02-14 at 16 11 34

Screenshot 2020-02-14 at 16 12 11

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
@@ -23,6 +23,10 @@
 
 @include govuk-media-query($until: tablet) {
 
+  .govuk-breadcrumbs__list {
+    display: flex;
+  }
+
   .govuk-breadcrumbs__list-item:not(:last-child):not(:first-child) {
     display: none;
   }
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -32,6 +36,11 @@
     padding-bottom: 14px;
   }
 
+  .govuk-breadcrumbs__list-item:before {
+    margin: initial;
```
andysellick: `initial` isn't supported by IE, so the breadcrumb looks like the below. I think the same effect could be achieved using `margin: 0`?

Screenshot 2020-02-17 at 09 02 04

RobBond-GDS: good spot, have implemented.

PR #1298, opened 2020-02-17
Mobile breadcrumb update guidance

RobBond-GDS:
On mobile devices update breadcrumb behaviour as per :

## What
Update guidance documentation to remove patterns with "is_current_page"

## Why
When a breadcrumb on a mobile device is collapsed, if last item is set as "is_current_page" then it is not possible to navigate to parent items. To eliminate the inconsistency, this option has been removed

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
+++ b/app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
@@ -3,6 +3,8 @@ description: "Navigational breadcrumbs, showing page hierarchy"
 body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
   Links have tracking data but links to the homepage (any link with a url of "/") will be tracked separately as `homeLinkClicked`
+
+  Note: Only the first and last (or parent) item in the breadcrumb will be displayed when the page width is less than the "tablet" govuk-breakpoint.
 shared_accessibility_criteria:
   - link
 accessibility_criteria:
@@ -28,7 +30,6 @@ examples:
       - title: 'Sub-section'
         url: '/section/sub-section'
         title: 'Education of disadvantaged children'
-        is_current_page: true
       inverse: true
     context:
       dark_background: true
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
```diff
@@ -3,6 +3,8 @@ description: "Navigational breadcrumbs, showing page hierarchy"
 body: |
   Accepts an array of breadcrumb objects. Each crumb must have a title and a URL.
   Links have tracking data but links to the homepage (any link with a url of "/") will be tracked separately as `homeLinkClicked`
+
+  Note: Only the first and last (or parent) item in the breadcrumb will be displayed on a page width less than 640px.
```
alex-ju: Can we use a breakpoint (e.g. tablet) instead of a fixed value in this sentence? It may be that 640px is not right, as we don't have a breakpoint with this value.

RobBond-GDS: updated not to refer to tablet breakpoint not a specific value

andysellick:
https://github.com/alphagov/finder-frontend/blob/master/app/presenters/finder_breadcrumbs_presenter.rb#L20
https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/breadcrumbs.rb

lib/govuk_publishing_components/presenters/breadcrumbs.rb
```diff
@@ -21,7 +21,7 @@ def structured_data
       def item_list_element
```
vanitabarrett: `request_path` is referenced elsewhere in this file - I think we can remove all these as it's not used any more.

RobBond-GDS: Removed `request_path`

vanitabarrett:
@RobBond-GDS Looks good, but not sure why there's a merge commit here - can we rebase to remove it?

RobBond-GDS:
Have rebased as requested

PR #1299, opened 2020-02-17
Change back link arrow to chevron

RobBond-GDS:
Update styling of back link component

https://trello.com/c/ZYJ7nnYt/21-make-back-icon-more-consistent-on-mobiles

## What
Update styling on back link component and increase touch target on mobile devices

## Why
Back links use an inconsistent icon from the breadcrumbs, The link also has a small touch target on mobiles. We should have consistent design patterns across GOV.UK. 

## Visual Changes
Screenshot 2020-02-17 at 13 43 32
Before change - mobile

Screenshot 2020-02-17 at 13 44 11
Before Change - desktop

Screenshot 2020-02-17 at 13 42 50
After change - mobile

Screenshot 2020-02-17 at 13 44 54
After Change - desktop

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Change back link arrow to chevron [PR #1299](https://github.com/alphagov/govuk_publishing_components/pull/1299)
 * Mobile breadcrumb update guidance [PR #1298](https://github.com/alphagov/govuk_publishing_components/pull/1298)
 * Add page heading captions to checkboxes and radio boxes components ([PR #1304](https://github.com/alphagov/govuk_publishing_components/pull/1304))
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
@@ -1,1 +1,38 @@
 @import "govuk/components/back-link/back-link";
+
+.govuk-back-link {
+  &[href] {
+    text-decoration: underline;
+    border-bottom: 1px solid transparent;
+
+    &:focus {
+      text-decoration: none;
+    }
+  }
+
+  &:before {
+    content: "";
+    display: block;
+    position: absolute;
+    top: 0;
+    bottom: 1px;
+    left: 3px;
+    width: 7px;
+    height: 7px;
+    -webkit-transform: rotate(225deg);
+    -ms-transform: rotate(225deg);
+    transform: rotate(225deg);
+    border: solid;
+    border-width: 1px 1px 0 0;
+    clip-path: inherit;
+  }
+}
+
+.govuk-back-link::after {
+  content: "";
+  position: absolute;
+  top: -14px;
+  right: 0;
+  left: 0;
+  bottom: -14px;
+}


maxgds:
I'm not convinced we should be diverging from the design system at this level. We're in a world where many things are inconsistent and we're trying to adopt the design system as fully as possible to overcome that so this feels like a backwards step. If the touch area isn't good enough I think it should be fixed at source rather than over-ridden. If you haven't already done so I'd suggest talking to the team about it: https://design-system.service.gov.uk/components/back-link/

injms:
From the looks of the [talk around the back link component itself](https://github.com/alphagov/govuk-design-system-backlog/issues/32), it would seem that the Home Office have done some research around improvements to the back button. There's also this [discussion about harmonising the design of the back link and breadcrumbs](https://github.com/alphagov/govuk-design-system-backlog/issues/33#issuecomment-465957906). I agree with @maxgds - it's better to try and fix this at source.

RobBond-GDS:
Thanks for the feedback, as it stands the mobile team has already had
conversations with the design team and the tasks that have been proposed
have been given the go-ahead by them as far as implementing on GOV.UK is
concerned. This is with a view to feeding back into the design system at a
later date.

maxgds:
If this is to be merged in, I think you need to give the chevron some further attention. On your screenshot the < seems to be misaligned vertically (it looks too low) on the mobile view and when I view it in chrome locally with dev tools responsive it's also misaligned but in the opposite direction by a couple of pixels.

RobBond-GDS:
> If this is to be merged in, I think you need to give the chevron some further attention. On your screenshot the < seems to be misaligned vertically (it looks too low) on the mobile view and when I view it in chrome locally with dev tools responsive it's also misaligned but in the opposite direction by a couple of pixels.

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,45 @@
...
+    text-decoration: underline;
+    border-bottom: 1px solid transparent;
+
+    &:focus {
+      text-decoration: none;
+    }
+  }
+
+  &:before {
+    content: "";
```
andysellick: Is it not possible to use the govuk-shape mixin here? e.g.

`@include govuk-shape-arrow($direction: left, $base: 10px, $height: 6px);`

RobBond-GDS:
I shall check its alignment with the breadcrumb component as that would be the pattern it is to compliment

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,45 @@
...
+    -ms-transform: rotate(2255deg);
+    transform: rotate(225deg);
+    border: solid;
+    border-width: 1px 1px 0 0;
+    clip-path: inherit;
+  }
+}
+
+@include govuk-media-query($until: tablet) {
+
```
andysellick: Nitpick: empty line.

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,45 @@
...
+  .govuk-back-link {
+    position: relative;
```
andysellick: What's this positioning for? Looks like it's already defined in the inspector.

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
RobBond-GDS: This would give a solid triangle, design calls for a left-facing chevron

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
RobBond-GDS: This applies the same pseudo-element to increase the touch target after the link as used in the breadcrumb PR

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
RobBond-GDS: have removed the empty line

RobBond-GDS:
Slice

Took screenshot of both breadcrumb and back link, overlaid on on the other. As you can see they chevron is aligned the same on both and same size

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
andysellick: What I mean is that position: relative is already set on this element - it doesn't need setting again.

RobBond-GDS: See what you mean, have removed duplicate

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,40 @@
...
+    -webkit-transform: rotate(225deg);
+    -ms-transform: rotate(2255deg);
+    transform: rotate(225deg);
+    border: solid;
+    border-width: 1px 1px 0 0;
+    clip-path: inherit;
+  }
+}
+
+@include govuk-media-query($until: tablet) {
```
vanitabarrett: This may be a question for the team, but I wonder if we want the larger touch target to apply all the time including on desktop? I'm not sure what the recommendations are, but it seems feasible that someone could be viewing the site in desktop view but on a touchscreen device.

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,40 @@
...
+    }
+  }
+
+  &:before {
+    content: "";
+    display: block;
+    position: absolute;
+    top: 0;
+    bottom: 1px;
+    left: 3.31px;
```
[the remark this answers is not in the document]
RobBond-GDS: so the chevron remains in the bounds of the focused state

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
```diff
@@ -1 +1,40 @@
...
+    width: 7px;
+    height: 7px;
+    -webkit-transform: rotate(225deg);
+    -ms-transform: rotate(2255deg);
```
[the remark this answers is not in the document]
RobBond-GDS: amended typo

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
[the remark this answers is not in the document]
vanitabarrett: `3px` seems to work ok?

CHANGELOG.md
```diff
@@ -7,8 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+
```
vanitabarrett: Can you remove this additional space and squash please? I think this PR needs another rebase too as someone has just versioned the gem.

RobBond-GDS:
Squashed and rebased

app/assets/stylesheets/govuk_publishing_components/components/_back-link.scss
RobBond-GDS: Asked PM and response was "That makes sense to me - would that have any adverse effect on the look of anything around the Back link on desktop? If not, happy to go with that change for desktop too. " 

As it wouldn't affect desktop have removed media query so it was just on mobile

PR #1293, opened 2020-02-13
Fix component guide layout in IE11

alex-ju:
## What
Fix component guide layout in IE11 by using [the Design System page template](https://design-system.service.gov.uk/styles/page-template/) and render breadcrumbs only if there are items.

## Why
@tombye said it's broken while testing at the Empathy lab – and it was bugging me for a while

## Visual Changes
<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td>

Screenshot 2020-02-13 at 15 27 48

</td><td>

Screenshot 2020-02-13 at 15 28 08

</td></tr>
</table>

PR #1302, opened 2020-02-18
Enable custom margin top on title component

alex-ju:
## What
Allow setting a custom margin-top on title component.

## Why
This change will allow the title component to be used in a [Design System page template](https://design-system.service.gov.uk/styles/page-template/) (which sets the padding on the main container – see examples below) by specifying a `margin_top: 0`. This property is introduced to avoid a breaking change and ease the transition to the new page template. When all applications will be using the Design System page template the component will be updated, specifying `margin_top: 0` will not be required and this example will be removed.

### Title in a Design System-based page template
design-system-main

design-system-title

### Title in the current GOV.UK page template
govuk-title

## Visual Changes
No visual changes

https://github.com/alphagov/govuk_publishing_components/pull/1293 is currently dependent on this change and presumably the new layout/template change too.

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/title.yml
+++ b/app/views/govuk_publishing_components/components/docs/title.yml
@@ -50,7 +50,7 @@ examples:
       context: Publication
       title: My page title which is often really long and verbose and has lots of extra words it doesn't need
       average_title_length: long
-  with_margin:
+  with_margin_bottom:
     description: |
       The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). It defaults to having a margin bottom of 50px.
 
@@ -78,4 +78,13 @@ examples:
       margin_bottom: 0
     context:
       dark_background: true
-
+  using_design_system_template:
+    description: |
+      This option allows the removal of top margin from the component so that it works within a [Design System page template](https://design-system.service.gov.uk/styles/page-template/default/index.html), where spacing above the title is already provided by padding on the wrapping div.
+    embed: |
+      <main class="govuk-main-wrapper">
+        <%= component %>
+      </main>
+    data:
+      title: My page title
+      margin_top: 0
--- a/lib/govuk_publishing_components/presenters/shared_helper.rb
+++ b/lib/govuk_publishing_components/presenters/shared_helper.rb
@@ -1,12 +1,17 @@
 module GovukPublishingComponents
   module Presenters
     class SharedHelper
-      attr_reader :options, :margin_bottom, :heading_level
+      attr_reader :options, :margin_top, :margin_bottom, :heading_level
 
       def initialize(local_assigns)
         @options = local_assigns
+        @margin_top = @options[:margin_top] || nil
         @margin_bottom = @options[:margin_bottom] || 3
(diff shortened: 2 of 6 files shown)


spec/components/title_spec.rb
```diff
@@ -41,26 +41,46 @@ def component_name
...
+    assert_select '.gem-c-title.govuk-\!-margin-top-0'
+  end
+
+  it "applies a valid margin top" do
+    render_component(title: 'Margin 4', margin_top: 4)
+    assert_select '.gem-c-title.govuk-\!-margin-top-4'
+  end
+
+  it "ignores an invalid margin top" do
+    render_component(title: 'Margin wat', margin_top: 17)
```
andysellick: 😁

app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -78,4 +78,15 @@ examples:
       margin_bottom: 0
     context:
       dark_background: true
-
+  using_design_system_template:
```
andysellick: Can you update the other example of margin in this file from 'With margin' to 'With margin bottom' ?

app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -78,4 +78,15 @@ examples:
...
+    description: |
+      Page title component used in a [Design System page template](https://design-system.service.gov.uk/styles/page-template/default/index.html).
+      This example is a reference for transitioning the current template used in GOV.UK to the Design System one without introducing a breaking change.
+      When all applications will be using the Design System template the component will be updated, specifying `margin_top: 0` will not be required and this example will be removed.
```
andysellick: Nitpick: these line breaks aren't honoured in the output unless there's an empty line between them, is this what you intended?

app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -78,4 +78,15 @@ examples:
       margin_bottom: 0
     context:
       dark_background: true
-
+  using_design_system_template:
+    description: |
+      Page title component used in a [Design System page template](https://design-system.service.gov.uk/styles/page-template/default/index.html).
```
andysellick: Suggestion: `This option allows the removal of top margin from the component so that it works within a Design System page template, where spacing above the title is already provided by padding on the wrapping div.`

lib/govuk_publishing_components/presenters/shared_helper.rb
```diff
@@ -1,14 +1,19 @@
...
       def initialize(local_assigns)
         @options = local_assigns
+        @margin_top = @options[:margin_top] || nil
         @margin_bottom = @options[:margin_bottom] || 3
         @heading_level = @options[:heading_level] || 2
       end
 
+      def get_margin_top
+        [*0..9].include?(@margin_top) ? "govuk-!-margin-top-#{margin_top}" : ""
+      end
```
andysellick: Shared helper for the win 👍

app/views/govuk_publishing_components/components/docs/title.yml
alex-ju: Yeah, arbitrary (phrase-based) soft-wrapping 😁. I can move them all on one line and let editors do the soft-wrap.

alex-ju: Even better, I replaced the content with your suggestion 👍

PR #1310, opened 2020-02-20
Make cookie banner text and preferences URL customisable

vanitabarrett:
## What
Adds the ability to pass the following into the cookie banner and override the default:

- Cookie banner message
- The URL on the 'set cookie preferences' button

## Why
This makes the banner more usable by other services which may have their own cookie settings pages and want to remove references to "GOV.UK" in the cookie banner messaging.

Our specific use case at the moment is data.gov.uk, which is GDS-owned but lives on a separate URL.

## Visual Changes
Screenshot 2020-02-20 at 11 17 00
Screenshot 2020-02-20 at 11 17 06

Heroku: https://govuk-publis-make-cooki-p0dhlc.herokuapp.com/component-guide/cookie_banner

### DIFF
--- a/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
@@ -1,6 +1,7 @@
 <%
   id ||= 'global-cookie-message'
   text ||= raw("We use <a class='govuk-link' href='/help/cookies'>cookies to collect information</a> about how you use GOV.UK. We use this information to make the website work as well as possible and improve government services.")
+  cookie_preferences_href ||= "/help/cookies"
 %>
 
 <div id="<%= id %>" class="gem-c-cookie-banner govuk-clearfix" data-module="cookie-banner" role="region" aria-label="cookie banner" data-nosnippet>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,5 +1,7 @@
 <%
   id ||= 'global-cookie-message'
+  text ||= raw("We use <a class='govuk-link' href='/help/cookies'>cookies to collect information</a> about how you use GOV.UK. We use this information to make the website work as well as possible and improve government services.")
```
andysellick: Do we need to think about translation options for this default text?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -8,7 +10,7 @@
       <div class=" govuk-grid-column-two-thirds">
         <div class="gem-c-cookie-banner__message">
           <span class="govuk-heading-m">Tell us whether you accept cookies</span>
```
andysellick: While we're at it, is it worth having a custom option this heading text as well?

vanitabarrett: Hm, I did consider it but thought I'd keep this change as small as possible - at the moment we don't have a use case for changing the title.

maxgds: I was going to ask the same about the button text but concluded no use case, no need.

andysellick: Fair enough.

PR #1312, opened 2020-02-20
Add cookie category list override functionality

vanitabarrett:
**Note: this change is up for debate - it's just my best guess at how we could approach this, but I'm open to other ideas.**

## What
Adds cookie function which allows a service to override the default list of cookies and categories allowed on the site. Here is an example of how this might be used:

```
# Defining the override method within the service
window.GOVUK.overrideCookieCategories = function() {
  var cookieCategories = {
    'cookies_policy': 'essential',
    'cookies_preferences_set': 'essential',
    'cookie_specific_to_service': 'settings',
  }

  return cookieCategories
}
```

## Why
If other services want to use the cookie banner functionality, they will want to define their own cookies to allow. In this specific case, we want to use the cookie banner on data.gov.uk.

Documentation added here:

- https://govuk-publis-spike-cook-hptqol.herokuapp.com/component-guide/cookie_banner
- https://github.com/alphagov/govuk_publishing_components/blob/8eeecba8ee89fbf27ebbe55170ed7a3d6adbc30a/docs/using-cookie-banner.md

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   of the commit log.
 
 ## Unreleased
+
+* Add cookie category override function [PR #1312](https://github.com/alphagov/govuk_publishing_components/pull/1312)
 * Make cookie banner text and preferences URL customisable [PR #1310](https://github.com/alphagov/govuk_publishing_components/pull/1310)
 
 ## 21.24.0
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,13 +9,13 @@
 
 ## Unreleased
 
-* Add cookie category override function [PR #1312](https://github.com/alphagov/govuk_publishing_components/pull/1312)
-* Make cookie banner text and preferences URL customisable [PR #1310](https://github.com/alphagov/govuk_publishing_components/pull/1310)
+* Add cookie category override function ([PR #1312](https://github.com/alphagov/govuk_publishing_components/pull/1312))
+* Make cookie banner text and preferences URL customisable ([PR #1310](https://github.com/alphagov/govuk_publishing_components/pull/1310))
 
 ## 21.24.0
 
-* Change back link arrow to chevron [PR #1299](https://github.com/alphagov/govuk_publishing_components/pull/1299)
-* Mobile breadcrumb update guidance [PR #1298](https://github.com/alphagov/govuk_publishing_components/pull/1298)
+* Change back link arrow to chevron ([PR #1299](https://github.com/alphagov/govuk_publishing_components/pull/1299))
+* Mobile breadcrumb update guidance ([PR #1298](https://github.com/alphagov/govuk_publishing_components/pull/1298))
 * Add page heading captions to checkboxes and radio boxes components ([PR #1304](https://github.com/alphagov/govuk_publishing_components/pull/1304))
 
 ## 21.23.1
--- a/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
@@ -11,7 +11,7 @@
     'campaigns': false
   }
 
-  var COOKIE_CATEGORIES = {
+  var DEFAULT_COOKIE_CATEGORIES = {
     'cookies_policy': 'essential',
     'seen_cookie_message': 'essential',
     'cookie_preferences_set': 'essential',
@@ -99,6 +99,7 @@
 
   window.GOVUK.setConsentCookie = function (options) {
     var cookieConsent = window.GOVUK.getConsentCookie()
+    var cookieCategories = window.GOVUK.cookieCategories()
 
     if (!cookieConsent) {
       cookieConsent = JSON.parse(JSON.stringify(DEFAULT_COOKIE_CONSENT))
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,8 @@
   of the commit log.
 
 ## Unreleased
+
+* Add cookie category override function [PR #1312](https://github.com/alphagov/govuk_publishing_components/pull/1312)
```
andysellick: Just noticed that the formatting of these entries changed midway through 21.24.0, not sure if it was intentional. Used to be `PR name ([PR num](link))`

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -207,14 +211,23 @@
     }
   }
 
+  window.GOVUK.cookieCategories = function () {
+    if (window.GOVUK.overrideCookieCategories) {
+      return window.GOVUK.overrideCookieCategories()
+    } else {
```
andysellick: We probably need some kind of documentation on how to do this?

vanitabarrett: Yeah, good point... Not sure where the best place for that would be. I could add a new file to `/docs`?

vanitabarrett: Or perhaps to the component guide, as it's related to the cookie banner component.

andysellick: I'd add it wherever there is similar documentation and link to that from the component guide, with a summary explanation.

vanitabarrett: @andysellick Added some documentation on [using the cookie banner](https://github.com/alphagov/govuk_publishing_components/blob/8eeecba8ee89fbf27ebbe55170ed7a3d6adbc30a/docs/using-cookie-banner.md) and linked to this from the component guide.

vanitabarrett:
@andysellick I'm not sure - as far as I know there are no immediate plans to start advertising this to other departments, but using this on data.gov.uk was also an experiment into how easily our solution could be applied outside of GOVUK.

I tried to keep the documentation quite general, as it seemed strange to write documentation specific to data.gov.uk (especially as it's already been mostly done now).

I'd imagine if we wanted to actually start making recommendations for developers outside GDS to start using the cookie feature, we might want to think about packaging it up separately from the gem.

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -207,14 +211,23 @@
     }
   }
 
+  window.GOVUK.cookieCategories = function () {
+    if (window.GOVUK.overrideCookieCategories) {
+      return window.GOVUK.overrideCookieCategories()
```
alex-ju: This is now guarded by CSP otherwise it may have been too permissive – in a sense that some could attach a function to the `GOVUK` global and we were executing it.

andysellick:
@alex-ju I hadn't thought about that. I'm not sure it impacts?

alex-ju:
> I hadn't thought about that. I'm not sure it impacts?

@andysellick Not in terms of impact, I was thinking more in terms of architecture. How we generally allow other departments/domains to attach their functionality to our scripts.

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
vanitabarrett: @alex-ju This will be executing within that site/domain though, not on GOV.UK? Or have I misunderstood?

andysellick:
@alex-ju at the moment all of the analytics code is in static, so if someone else were using the gem they would still need their own analytics code. So it should be fine? I think? Let's have this conversation in person.

vanitabarrett:
Yes, at the moment DGU has its own analytics code and it doesn't cross over with GOVUK analytics. I can write this up as something to make note of if anyone starts considering this as a wider solution to services outside GDS though.

vanitabarrett:
@alex-ju @andysellick Are people happy for this to be merged now, or would you rather wait until we've discussed the things above?

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
alex-ju: Yes, on their domain, which I guess is fine – unless someone exploits their site, which doesn't have to do with us.

vanitabarrett:
Noting down before I forget: a short-term solution might just be to add any DGU-specific cookies to the list of cookies that already exists in the gem, without providing any kind of override functionality.

vanitabarrett:
This PR has been superseded by https://github.com/alphagov/govuk_publishing_components/pull/1315

We've decided that this approach is probably a bit too open at the moment, as we have concerns around opening up the use of this gem to outside GDS, in terms of:

- Maintaining external contributions
- Opening up GOVUK-specific code to external departments, e.g: analytics (when these are moved out of Static)

As this is only needed for data.gov.uk at the moment, we've decided that a better short-term solution is to add any DGU-specific cookies to the cookies list in the gem itself, as we have control over both the gem and DGU within GDS.

However, one option if we needed to open this up again might be to add a whitelist to any sensitive javascript (e.g: analytics and cookie consent) so it can only run on domains specified within a whitelist in the gem. That way, we can still retain control over who is using our scripts.

PR #1313, opened 2020-02-20
Update main README and usage guidance

alex-ju:
## What
- Try to clarify the scope of the project, relationship with GOV.UK Design System and browser support in the main [README.md](https://github.com/alphagov/govuk_publishing_components/blob/update-guidance/README.md)
- Replace bowl/vm/vagrant approach with [govuk-docker](https://github.com/alphagov/govuk-docker) in [run-component-guide.md](https://github.com/alphagov/govuk_publishing_components/blob/update-guidance/docs/run-component-guide.md)

## Why
To keep our docs tidy.

## Visual Changes
No visual change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@ GOV.UK Publishing Components ·
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
 =====================
 
-GOV.UK Publishing Components is a Ruby gem to document and distribute components in GOV.UK frontend applications.
+GOV.UK Publishing Components is a Ruby gem to document and distribute components for GOV.UK applications.
 
 It provides:
 - a shared library of components
--- a/README.md
+++ b/README.md
@@ -3,11 +3,12 @@ GOV.UK Publishing Components ·
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
 =====================
 
-This gem:
+GOV.UK Publishing Components is a Ruby gem to document and distribute components in GOV.UK frontend applications.
 
-- Provides shared components for applications
-- Provides helpers to generate component payloads
-- Provides an application to preview components
+It provides:
+- a shared library of components
+- helpers to generate component payloads
+- an application to preview components and provide guidance
 
 Components should be added to this gem if they are required in more than one application, otherwise they should be added to that application.
 
@@ -19,7 +20,7 @@ Components should be added to this gem if they are required in more than one app
 - [Develop a component](/docs/develop-component.md)
 - [Run the component guide](/docs/run-component-guide.md)
 - [Move a component from an application to the gem](/docs/moving-components-upstream-into-this-gem.md)
-- [Publish this gem](/docs/publishing-to-rubygems.md)
+- [Publish/release a new version of the gem](/docs/publishing-to-rubygems.md)
 
 ## Architecture / structure
 
@@ -32,18 +33,12 @@ There are 2 types of helper classes in this app:
 - [AppHelpers](lib/govuk_publishing_components/app_helpers). Are exposed to the applications using this gem. They should be documented using RDoc.
 - [Component Presenters](lib/govuk_publishing_components/presenters). Anything in these classes is only for use within the components. They should be marked `@private`.
 
-## GOV.UK Frontend
+## Relationship with GOV.UK Design System
 
-This gem is the way GOV.UK consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), part of the GOV.UK Design System.
+GOV.UK Publishing Components also makes [GOV.UK Design System](https://design-system.service.gov.uk/) styles and components available to GOV.UK frontend applications. This gem consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) via [NPM](https://www.npmjs.com/).
 
-We use NPM to pull in the GOV.UK Frontend styles and Javascript, but we don't commit
-the `node_modules` directory. To install GOV.UK Frontend on your machine, run:
-
-```
-npm install
-```
-
-You'll need to have installed Node and NPM for this.
+## Browser and assistive technology support
+GOV.UK Publishing Components shares the same standards in terms of browser and assistive technology support with [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support).
 
 ## Documentation
 
@@ -51,7 +46,7 @@ You'll need to have installed Node and NPM for this.
 
 Run the documentation locally with:
 
-```
+```sh
 bundle exec yard server --reload
 ```
 
@@ -59,7 +54,7 @@ bundle exec yard server --reload
 
 The default rake task runs all tests:
 
-```
+```sh
 bin/rake
 ```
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -32,34 +33,28 @@ There are 2 types of helper classes in this app:
...
-We use NPM to pull in the GOV.UK Frontend styles and Javascript, but we don't commit
-the `node_modules` directory. To install GOV.UK Frontend on your machine, run:
-
-```
-npm install
-```
-
-You'll need to have installed Node and NPM for this.
+## Browser and assistive technology support
+GOV.UK Publishing Components shares the same standards in terms of browser and assistive technology support with [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support).
```
nickcolley: Is this true? We support IE8 but your project could decide to follow the service manual?

alex-ju: This update is followed by a discussion between frontend developers in GOV.UK that happened yesterday. The reasoning behind this is that, as far as I know, the decision for the current browser matrix that GOV.UK Frontend uses was based on data from GOV.UK.

alex-ju: Happy for this matter to be discussed more widely if we think otherwise – only that I currently can't make any of the frontend meetings 😞.

nickcolley: > the decision for the current browser matrix that GOV.UK Frontend uses was based on data from GOV.UK.

GOV.UK Frontend's matrix is based on supporting the lowest denominator across *all* services. Services themselves should follow the service manual as a basis and adjust according to the specific audience.

README.md
```diff
@@ -3,11 +3,12 @@ GOV.UK Publishing Components ·
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
 =====================
 
-This gem:
+GOV.UK Publishing Components is a Ruby gem to document and distribute components in GOV.UK frontend applications.
```
andysellick: Nitpick: is `GOV.UK frontend` confusing here? I initially read it as `govuk-frontend`.

Similar problem on line 38, we're talking about `GOV.UK frontend` and `GOV.UK Frontend`.

alex-ju: "in GOV.UK applications" perhaps? I get the confusion, harder to stay away from it.

andysellick: Yeah, I couldn't think of a good alternative. GOV.UK publishing?

nickcolley: GOV.UK **Publishing** as you are the publishing service part of GOV.UK (government online) 😈

injms: Maybe?

```suggestion
GOV.UK Publishing Components is a Ruby gem to document and distribute components for GOV.UK frontend applications.
```

README.md
```diff
@@ -32,34 +33,28 @@ There are 2 types of helper classes in this app:
 - [AppHelpers](lib/govuk_publishing_components/app_helpers). Are exposed to the applications using this gem. They should be documented using RDoc.
 - [Component Presenters](lib/govuk_publishing_components/presenters). Anything in these classes is only for use within the components. They should be marked `@private`.
 
-## GOV.UK Frontend
+## Relationship with GOV.UK Design System
 
-This gem is the way GOV.UK consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend), part of the GOV.UK Design System.
+GOV.UK Publishing Components also makes [GOV.UK Design System](https://design-system.service.gov.uk/) styles and components available to GOV.UK frontend applications. This gem consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) via [NPM](https://www.npmjs.com/).
```
injms:
```suggestion
GOV.UK Publishing Components also makes [GOV.UK Design System](https://design-system.service.gov.uk/) styles and components available to GOV.UK's frontend applications. This gem consumes [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend) via [NPM](https://www.npmjs.com/).
```

docs/run-component-guide.md
```diff
@@ -1,24 +1,24 @@
-## Running the gem as a standalone app
+## Running the component guide as a standalone app
 
 If you want to work on this gem directly, without having to add it to another
-application, you can run it in the development vm or on your local machine with:
+application, you can run it on your local machine or using [govuk-docker](https://github.com/alphagov/govuk-docker). You'll need [Ruby](https://www.ruby-lang.org/en/downloads/) and [Node.js](https://nodejs.org/en/) installed to get this project running.
```
injms: I don't think you need Node.js on the local machine to get this running - I thought that Node is within a docker container?

injms:
```suggestion
application, you can run it on your local machine or using [govuk-docker](https://github.com/alphagov/govuk-docker). 

To run it locally you'll need [Ruby](https://www.ruby-lang.org/en/downloads/) and [Node.js](https://nodejs.org/en/) installed to get this project running.
```

docs/run-component-guide.md
```diff
@@ -1,24 +1,24 @@
...
-$ cd /var/govuk/govuk_publishing_components
-$ bundle
-$ npm install
+cd ~/govuk/govuk-docker
+make govuk_publishing_components
 
-$ cd /var/govuk/govuk-puppet/development-vm
-$ bowl publishing-components
+cd ~/govuk/govuk_publishing_components
+govuk-docker up govuk_publishing_components-app
```
injms: If you `cd` into the folder you can just run `govuk-docker-up`; the command `govuk-docker up govuk_publishing_components-app` can be run from any folder.

```suggestion
govuk-docker-up
```

README.md
alex-ju: My only concern with using 'GOV.UK Publishing' is that people may perceive them as only the ones sitting on a `.publishing.service.gov.uk` domain (e.g. Signon, Content Publisher) instead of the entire suite of applications GOV.UK owns (including the ones that are visible on www.gov.uk).

alex-ju:
This PR seems to have stalled. Any blockers that I should address or things I can improve to get a yay (or a nay)?

README.md
richardTowers: I agree with Ian and Alex:

```suggestion
GOV.UK Publishing Components is a Ruby gem to document and distribute components for GOV.UK applications.
```

nickcolley: GOV.UK is Government online, it's a problem that GOV.UK Publishing sees it self as 'all of GOV.UK', it should see itself as a service like Verify, Notify, Pay.

GOV.UK Publishing is the programme on the 6th floor that owns all of those suite of applications.

richardTowers: I agree with you in principle, but I don't think this readme is necessarily the right place to blaze a trail for that distinction.

There's a trade-off we need to make between potentially confusing external users (who don't understand what we mean when we say "GOV.UK" as opposed to "GOV.UK Publishing"), and confusing internal users (who already use the phrase "GOV.UK publishing applications" to refer to a subset of the applications that use this gem).

Maybe: _GOV.UK Publishing Components is a Ruby gem to document and distribute components for applications maintained by the GOV.UK team at GDS_ would be less controversial?

At least removing "frontend" clears up the confusion with "govuk-frontend" though.

README.md
alex-ju: We already state this in the [GOV.UK developer docs](https://docs.publishing.service.gov.uk/manual/browser-support.html) it seems so this change shouldn't be controversial.

PR #1319, opened 2020-02-21
Fix layout header width issue

andysellick:
## What / why
- the new gem allows individual sass for components to be imported into applications
- this isn't quite perfect yet, because the component guide imports its own sass and the application's sass
- both now import govuk-frontend (this isn't ideal, I realise) but differ in the components they import, the guide imports all, the application imports some
- now that the two are different, including the application's sass after the component guide means that govuk-frontend's styles for the layout header are overridding those in the layout header component
- fix is to swap the ordering of these in the guide markup, but ideally a better solution can be found as clearly some duplication is happening here

## Visual Changes

Component guide should look like this:

Screenshot 2020-02-21 at 16 50 48

but looks like this:

Screenshot 2020-02-21 at 15 55 35

This PR fixes that.

Just for interest, here's how the CSS should be, but the problem was that the `gem-c-header__logo` class was being overridden. It might also be worth looking into how this CSS works as an alternative option to fix this long term.

Screenshot 2020-02-21 at 16 58 17

### DIFF
--- a/app/views/layouts/govuk_publishing_components/application.html.erb
+++ b/app/views/layouts/govuk_publishing_components/application.html.erb
@@ -16,8 +16,8 @@
     <%= csrf_meta_tags %>
     <%= favicon_link_tag "govuk_publishing_components/favicon-production.png" %>
 
-    <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
     <%= stylesheet_link_tag "#{GovukPublishingComponents::Config.application_stylesheet}" %>
+    <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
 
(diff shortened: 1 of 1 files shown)


kr8n3r:
yes the correct cascade order should be 
width:33% and then width:auto from our component header, as we override the govuk-frontend styles because of https://github.com/alphagov/govuk-frontend/issues/1502
so swapping the order

andysellick:
I think a broader question might be 'why do we have the application's CSS included in the component guide?' but I think the answer might be so we spot things like this. We could however consider removing it in future.

andysellick:
Thanks for the context @kr8n3r

PR #1320, opened 2020-02-24
Improve suggested sass functionality

andysellick:
## What / why

This change relates to the work done in https://github.com/alphagov/govuk_publishing_components/pull/1159 to allow individual components to be imported into an application, specifically relating to an oversight in the 'suggested sass for this application' functionality.

- the component guide was checking for what components are in use, but some components import other components, so it was missing those
- added a check for components within components and added to suggested sass output

Screenshot 2020-02-24 at 09 47 07

## Visual Changes
No visual changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ## Unreleased
 
-* Allow individual components to be imported into apps ([PR #1159](https://github.com/alphagov/govuk_publishing_components/pull/1159))
 * Update Sass documentation ([PR #1321](https://github.com/alphagov/govuk_publishing_components/pull/1321))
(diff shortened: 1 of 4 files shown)


injms:
When looking at the component guide on government-frontend locally, I'm seeing `nta` being used in the black bar heading and the title - doesn't seem to be loading, so it falls back to Arial:

image

Rest of the test is `GDS Transport`, which is loaded correct. 

Not sure if this is my local install been strange, but I thought it best to flag as a potential thing.

kr8n3r:
@injms it's most likely due to `$govuk-compatibility-govuktemplate: true` setting in goverment-frontend stylesheet

PR #1290, opened 2020-02-13
Mobile breadcrumb to show first and last items only

RobBond-GDS:
On mobile devices update breadcrumb behaviour as per :

https://trello.com/c/YuObLgSq/10-fix-breadcrumbs-on-mobile

## What
Apply mobile enhancement designs to the GOV.UK Breadcrumb component, specifically:

- Collapse breadcrumb to show only the first and last item (the parent item) on devices less than 640px wide
- Increase the touch target padding on mobile devices
- Collapse on Mobile to be de facto behaviour

## Why

The reasoning is taken from this snippet in the trello card:

> This [article by NNG](https://www.nngroup.com/articles/breadcrumbs/) lays out the pros and cons of showing a full breadcrumb path on mobile. In short, the drawbacks of having a multi-line breadcrumb, and the need to save space on mobile, make it less useful to users. By truncating the breadcrumb we at least allow users to go backwards by one level, or home.

## Visual Changes
Screenshot 2020-02-05 at 14 57 28
Before Change

Screenshot 2020-02-05 at 14 56 20
After Change

PR #1324, opened 2020-02-26
Dont show breadcrumb item with no url

RobBond-GDS:
Part of updating breadcrumb behaviour as per :

https://docs.google.com/document/d/1OFG1Ln6VeCR_bc0qmUtRwMJU62Ljw4F3HO6hq2yloXk/

## What
Add logic to check for breadcrumb item to not show if it does not have a corresponding "url" 

## Why
Breadcrumb items with no "url", do not aid navigation. Specifically, when a breadcrumb is collapsed on a mobile device to show only first and last items, there is no method to navigate to the parent page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Dont show breadcrumb item with no url ([PR #1324](https://github.com/alphagov/govuk_publishing_components/pull/1324))
+
 ## 21.26.1
 
 * Add visually hidden text to share links component ([PR #1286](https://github.com/alphagov/govuk_publishing_components/pull/1286/))
--- a/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
@@ -13,21 +13,19 @@
   <ol class="govuk-breadcrumbs__list">
     <% breadcrumbs.each_with_index do |crumb, index| %>
       <% breadcrumb = GovukPublishingComponents::Presenters::Breadcrumb.new(crumb, index) %>
-      <li class="govuk-breadcrumbs__list-item" aria-current="<%= breadcrumb.aria_current %>">
         <% if breadcrumb.is_link? %>
-          <%= link_to(
-            breadcrumb[:title],
-            breadcrumb.path,
-            data: breadcrumb.tracking_data(breadcrumbs.length),
-            class: "govuk-breadcrumbs__link",
-            aria: {
-              current: breadcrumb.aria_current,
-            }
-          ) %>
-        <% else %>
-          <%= breadcrumb[:title] %>
+          <li class="govuk-breadcrumbs__list-item" aria-current="<%= breadcrumb.aria_current %>">
+              <%= link_to(
+                breadcrumb[:title],
+                breadcrumb.path,
+                data: breadcrumb.tracking_data(breadcrumbs.length),
+                class: "govuk-breadcrumbs__link",
+                aria: {
+                  current: breadcrumb.aria_current,
+                }
+              ) %>
+          </li>
         <% end %>
-      </li>
     <% end %>
   </ol>
 </div>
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## 21.26.1
 
+* Dont show breadcrumb item with no url ([PR #1324](https://github.com/alphagov/govuk_publishing_components/pull/1324))
```
vanitabarrett: This shouldn't be included under `## 21.26.1`, it should be under an `## Unreleased` heading.

kevindew:
@RobBond-GDS this looks to be a breaking change - I'm a bit confused why it was released as a patch release?

This breaks [Content Publisher](https://github.com/alphagov/content-publisher) where we make use of the breadcrumb component to display hierarchical information without links. So we can't currently update govuk_publishing_components. Example:

Screenshot 2020-02-27 at 16 27 52

kevindew:
I'm also a bit confused why we had to change the component for this rather just not pass in items without links at point of usage. It seems like potentially likely to cause bugs if some items are passed are just ignored.

vanitabarrett:
Thanks for letting us know @kevindew! Sorry, we weren't aware that anyone on GOVUK was using breadcrumbs in that way. 

[This change](https://github.com/alphagov/govuk_publishing_components/pull/1290) means that passing in text but no url will mean mobile users will be stuck with just one navigation link (normally the homepage in the case of GOVUK). Does this collapsing change also break your usage of breadcrumbs in Content Publisher?

andysellick:
@kevindew apologies, I didn't realise this was a breaking change when I did the latest version.

kevindew:
> Thanks for letting us know @kevindew! Sorry, we weren't aware that anyone on GOVUK was using breadcrumbs in that way.
> 
> [This change](https://github.com/alphagov/govuk_publishing_components/pull/1290) means that passing in text but no url will mean mobile users will be stuck with just one navigation link (normally the homepage in the case of GOVUK). Does this collapsing change also break your usage of breadcrumbs in Content Publisher?

Ah yes, thanks for letting us know about that. That does indeed affect our presentation on mobile devices and we lose the hierarchical information. 

Hmm we seem to be caught in a bit of an awkward where this component is diverging from the general purpose use of the design system one to a quite specific GOV.UK frontend navigation. I wonder if there's something we can do to uncouple these cases.

Any thoughts @alex-ju ?

kr8n3r:
finders-frontend is using it in a similar way to content-publisher I just noticed via a failing test

vanitabarrett:
@kr8n3r have you got an example?

kr8n3r:
yep, when you go from an organisation latest documents to the search.
example https://www.gov.uk/search/all?organisations%5B%5D=competition-and-markets-authority&order=updated-newest&parent=competition-and-markets-authority

vanitabarrett:
I think the finder-frontend case is something which we're trying to avoid and make more consistent on GOVUK, so I think the solution there would be to stop providing the "Search" part of the breadcrumb.

As for the Content Publisher use, perhaps the breadcrumb component isn't the most appropriate? The only other way I can think of having the two co-existing is putting this new behaviour back behind a flag so it doesn't automatically apply (or can be overridden).

kr8n3r:
yeah, Jeremy is looking at breadcrumbs on finders so that can be sorted in the near future.

kevindew:
It looks like we have quite a few of the finder-frontend style cases:

https://github.com/alphagov/search-admin/blob/01ebc9123086d086bc851c490986d8b57f7113f1/app/views/recommended_links/show.html.erb#L1-L11

https://github.com/alphagov/release/blob/b9ffa6e0013aa93e81b4bea996c19a1f449f882f/app/views/deployments/new.html.erb#L3-L9

I wonder if the best route forward is to revert the changes done as minor/patch releases and re-instate them behind opt-in flags for the time being? It feels like there is a wider discussion about the scope of this component to decide if it's only for a particular use case which may end up being a major release (I'd guess we'd also want a non url to result in an exception than being silently ignored). How's that sound or is there another route we can take?

vanitabarrett:
Sounds good @kevindew , I'll take a look at reverting those changes tomorrow

kevindew:
> Sounds good @kevindew , I'll take a look at reverting those changes tomorrow

Thank you 👍 happy to help too, feel free to ping me for reviews.

alex-ju:
> Hmm we seem to be caught in a bit of an awkward where this component is diverging from the general purpose use of the design system one to a quite specific GOV.UK frontend navigation. I wonder if there's something we can do to uncouple these cases.
> 
> Any thoughts @alex-ju ?

I agree we should try and keep the component generic and aligned with the Design System. We shall try to accommodate both (if not multiple) scenarios by component configuration.

vanitabarrett:
PR here to address these comments: https://github.com/alphagov/govuk_publishing_components/pull/1330

PR #1327, opened 2020-02-27
Streamline feedback

andysellick:
## What
Changes the feedback component to make it more standalone. Specifically:

- component receives a chunk of markup from the backend for error messages, containing links, paras and a h2. This was previously styled with `%extend` but now the JS adds the relevant `govuk-frontend` links.
- component had style overrides to stop static from re-styling the text inputs in the 'is there anything wrong with this page?' section. Looks like this is no longer necessary, and also removes another `%extend`.

## Why
In the new world of individual component sass import, it turns out we need to import the feedback component styles into static, where it is used. Unfortunately with the use of `%extend` this was not possible without also pulling in most of `govuk-frontend`, adding unnecessary CSS weight and duplication with apps. This change should remove the need to do that.

## Visual Changes

Only one minor change, using the `govuk-heading-m` class on the headings adds a little bit more margin underneath them, but I think this is fine.

Before:

Screenshot 2020-02-27 at 09 35 05

After:

Screenshot 2020-02-27 at 09 35 17


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -93,24 +93,6 @@
...
-  }
-
-  p {
-    @include govuk-text-colour;
-    @include govuk-font(19);
-    margin: govuk-spacing(2) 0;
-  }
-
-  a {
-    @extend %govuk-link;
```
kr8n3r: we could replace this extend with 
    @include govuk-link-common;
    @include govuk-link-style-default;
    @include govuk-link-print-friendly;

andysellick: Cool. What are the equivalents for h2 and p?

andysellick: One question though - wouldn't this route add additional CSS weight? Whereas the approach I'm using (apply the relevant classes using JS) doesn't. I think I'd prefer to not add more CSS if possible.

kr8n3r: for heading-m
```
@include govuk-text-colour;
@include govuk-font($size: 24, $weight: bold);
```

kr8n3r: so we can then also remove the JS for adding classes (which would require at least a partial import of frontend anyway)

andysellick: `govuk-frontend` is going to be imported by whatever app is rendering the page with the feedback component on anyway.

kr8n3r: apps yes. I though the concern was for static to have to import frontend to use those classes

andysellick: With this change static shouldn't have to import anything from `govuk-frontend` other than the mixins, ie. no added CSS weight.

kr8n3r: unless i'm reading it incorrently, if you add `.govuk-heading-m`, etc class they won't be styled in static as it doesn't import those partials. by using mixin's you get the styles in static with a minor hit

andysellick: Ah - are you saying that's a page that's only styled by static? I was assuming all pages are rendered by an app as well as static. If that's the case, I'll go with your suggestion.

Related: for the second commit, is there an equivalent for inputs using the `@include` stuff you suggested above?

kr8n3r: judging by https://github.com/alphagov/static/blob/master/app/views/root/_error_page.html.erb 
hopefully someone with more knowledge can advise

as for the input, side from border and font-size, not much more
https://github.com/alphagov/govuk-frontend/blob/master/package/govuk/components/input/_input.scss#L10

andysellick: @kr8n3r I've updated as per your suggestion, and removed the JS changes. I've also checked the component on the 404 page, and made some specific overrides to make inputs appear correctly when it's just static rendering the page. We'll need to remove them at a later date, but fine for now.

PR #1329, opened 2020-02-27
Fix feedback component tests

andysellick:
## What
- there are several tests in the feedback component to check that when parts of the component are interacted with specific inputs or error messages receive focus
- these tests were failing intermittently, despite not having changed recently
- unsure of the cause of this, but the result was that the body was focussed and not the intended input
- changing the tests to check that focus is called on the correct elements instead of checking to see what the activeElement is seems to work

## Why
These errors have been causing builds to fail for days. Hopefully this has fixed them.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove font_size option on contents list ([PR #1325](https://github.com/alphagov/govuk_publishing_components/pull/1325))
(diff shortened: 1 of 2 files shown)


andysellick:
Obviously it's hard to successfully prove that these changes have fixed an intermittently failing test if the build passes, so I've re-run the build in jenkins three times and re-run the tests locally loads more (it was also failing intermittently locally) and all seems okay.

vanitabarrett:
Seems to be passing consistently now! Thanks for fixing this ❤️

PR #1330, opened 2020-02-27
Make breadcrumb component more flexible to support different uses

vanitabarrett:
## What

- Revert default breadcrumb behaviour back to non-collapsing on mobile
- Adds a `collapse_on_mobile` flag. This defaults to false, but it is recommended that anyone using the breadcrumbs for navigation sets this to true.
- Set collapse_on_mobile to true in the contextual breadcrumbs component
- Reverts [#1324](https://github.com/alphagov/govuk_publishing_components/pull/1324) to allow breadcrumb items without URLs.
- Mobile touch targets only apply if `collapse_on_mobile` is set to true. We can look at fixing this in the future, but I think it will need some design work.

## Why
As discussed in https://github.com/alphagov/govuk_publishing_components/pull/1324 , we had missed some use cases for this component and our changes essentially broke this component for them.

My original approach was to introduce a `collapse_on_mobile` flag defaulting to true (opt-out approach). However, this may have caused problems if contributing the change to the design system which would probably have to be an opt-in approach.

Final approach is reverting default behaviour back to non-collapsing behaviour on mobile. This can be switched on with the `collapse_on_mobile` flag. This is a little bit more work as we will need to identify all breadcrumb uses and update the apps to use this flag, but means this is not a breaking change and should make contributing the change to the design system easier.

## collapse_on_mobile set to true
Screenshot 2020-02-27 at 19 50 32

## collapse_on_mobile set to false (or removed)
Screenshot 2020-02-27 at 19 50 43

https://govuk-publis-provide-br-aju5m4.herokuapp.com/component-guide/breadcrumbs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
 
 * Remove font_size option on contents list ([PR #1325](https://github.com/alphagov/govuk_publishing_components/pull/1325))
 * Fix feedback component tests ([PR #1329](https://github.com/alphagov/govuk_publishing_components/pull/1329))
+* Revert 'Dont show breadcrumb item with no url' ([PR #1330](https://github.com/alphagov/govuk_publishing_components/pull/1330))
+* Make breadcrumb collapsing behaviour opt-in and provide `collapse_on_mobile` flag ([PR #1330](https://github.com/alphagov/govuk_publishing_components/pull/1330))
 
 ## 21.26.2
 
--- a/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
@@ -1,15 +1,19 @@
 <%
   breadcrumbs ||= []
   inverse ||= false
-  invert_class = inverse ? "gem-c-breadcrumbs--inverse" : ""
+  collapse_on_mobile ||= false
   breadcrumb_presenter = GovukPublishingComponents::Presenters::Breadcrumbs.new(breadcrumbs)
+
+  classes = "gem-c-breadcrumbs govuk-breadcrumbs"
+  classes << " gem-c-breadcrumbs--collapse-on-mobile" if collapse_on_mobile
+  classes << " gem-c-breadcrumbs--inverse" if inverse
 %>
 
 <script type="application/ld+json">
   <%= raw breadcrumb_presenter.structured_data.to_json %>
 </script>
 
-<div class="gem-c-breadcrumbs govuk-breadcrumbs <%= invert_class %>" data-module="track-click">
+<div class="<%= classes %>" data-module="track-click">
   <ol class="govuk-breadcrumbs__list">
     <% breadcrumbs.each_with_index do |crumb, index| %>
       <% breadcrumb = GovukPublishingComponents::Presenters::Breadcrumb.new(crumb, index) %>
--- a/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
@@ -17,19 +17,21 @@
   <ol class="govuk-breadcrumbs__list">
     <% breadcrumbs.each_with_index do |crumb, index| %>
       <% breadcrumb = GovukPublishingComponents::Presenters::Breadcrumb.new(crumb, index) %>
+        <li class="govuk-breadcrumbs__list-item" aria-current="<%= breadcrumb.aria_current %>">
         <% if breadcrumb.is_link? %>
-          <li class="govuk-breadcrumbs__list-item" aria-current="<%= breadcrumb.aria_current %>">
-              <%= link_to(
-                breadcrumb[:title],
-                breadcrumb.path,
-                data: breadcrumb.tracking_data(breadcrumbs.length),
-                class: "govuk-breadcrumbs__link",
-                aria: {
-                  current: breadcrumb.aria_current,
-                }
-              ) %>
-          </li>
+          <%= link_to(
+            breadcrumb[:title],
+            breadcrumb.path,
+            data: breadcrumb.tracking_data(breadcrumbs.length),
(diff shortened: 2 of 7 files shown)


app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
```diff
@@ -2,14 +2,16 @@
...
+  collapse_on_mobile_class = collapse_on_mobile ? "gem-c-breadcrumbs--collapse-on-mobile" : ""
   breadcrumb_presenter = GovukPublishingComponents::Presenters::Breadcrumbs.new(breadcrumbs)
 %>
 
 <script type="application/ld+json">
   <%= raw breadcrumb_presenter.structured_data.to_json %>
 </script>
 
-<div class="gem-c-breadcrumbs govuk-breadcrumbs <%= invert_class %>" data-module="track-click">
+<div class="gem-c-breadcrumbs govuk-breadcrumbs <%= invert_class %> <%= collapse_on_mobile_class %>" data-module="track-click">
```
andysellick: Might be more readable to stack the classes like in other components: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_document_list.html.erb#L4

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
```diff
@@ -78,3 +78,13 @@ examples:
         url: '/education'
       - title: 'Education of disadvantaged children appended with some extra long content to make this a very very very very long taxon'
         url: '/education'
+  stop_collapsing_on_mobile:
+    data:
```
andysellick: I know it's already been explained at the top of the page, but worth adding a quick description here as well? `On a mobile screen, only the first and last breadcrumbs are shown. Show all breadcrumbs on mobile using this option.`?

spec/components/breadcrumbs_spec.rb
```diff
@@ -128,14 +128,16 @@ def assert_link_with_text_in(selector, link, text)
     assert_select ".gem-c-breadcrumbs--inverse"
   end
 
-  it "ignore breadcrumb items without url" do
+  it "renders breadcrumb items without link as  text" do
```
andysellick: The nitpickiest nitpick: is there an extra space between 'as' and 'text'?

spec/components/breadcrumbs_spec.rb
```diff
@@ -128,14 +128,16 @@ def assert_link_with_text_in(selector, link, text)
...
     render_component(
       breadcrumbs: [
         { title: 'Topic', url: '/topic' },
         { title: 'Current Page' },
       ]
     )
-    assert_select('.govuk-breadcrumbs__list-item:last-child', 'Topic')
+
+    assert_link_with_text_in('.govuk-breadcrumbs__list-item:first-child', '/topic', 'Topic')
+    assert_select('.govuk-breadcrumbs__list-item:last-child', 'Current Page')
```
andysellick: Does this test actually prove there's no link in the final list item? Or do we need to be more explicitly checking there's no link element?

CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ## Unreleased
 
 * Remove font_size option on contents list ([PR #1325](https://github.com/alphagov/govuk_publishing_components/pull/1325))
+* Revert 'Dont show breadcrumb item with no url' ([PR #1330](https://github.com/alphagov/govuk_publishing_components/pull/1330))
+* Add `collapse_on_mobile` flag which can be used to override default collapsing behaviour ([PR #1330](https://github.com/alphagov/govuk_publishing_components/pull/1330))
```
andysellick: This is outside of our conventions for changelog entries but I see why you've done it. Could we make the PR title and this line match?

vanitabarrett: @andysellick This PR covers both changes, so I didn't want to make the PR title too specific otherwise it may imply that the revert isn't included here?

kevindew:
Thank you @vanitabarrett for getting this put together so fast, really appreciate it 👍 

The only thing I'm concerned about is that by default this component will behave differently to the design system approach. It feels like we'd normally expect components to behave as per the design system unless we provide modifiers. We could end up in a strange situation if the design system adopts this collapse approach behind a flag where we have to pass a flag to disable it but the design system needs a flag to enable it.

Do we have precedent for this type of divergence before?

I don't mind too much, particularly if the general consensus from GOV.UK frontend devs is that this is the right thing to do, but just thought I'd raise.

vanitabarrett:
@kevindew Good point, I don't know of any other examples of divergence like that. I would guess that because we still define our own HTML for the components and just use design system styling and JS that this shouldn't cause a problem... but I agree it may be confusing. 

I think the mobile team will be revisiting the best way to contribute these changes back to the design system. I'd be happy to merge this as a temporary solution at least, but open to other ideas!

alex-ju:
> The only thing I'm concerned about is that by default this component will behave differently to the design system approach. **It feels like we'd normally expect components to behave as per the design system unless we provide modifiers.** We could end up in a strange situation if the design system adopts this collapse approach behind a flag where we have to pass a flag to disable it but the design system needs a flag to enable it.

Thank you for raising this so quickly (although the late hour concerns me a bit)!
I admit I haven't looked at the code, but I envision that the `collapse_on_mobile` flag would be false by default too. Mainly for the same reasons Kevin already mentioned above.

vanitabarrett:
@kevindew @alex-ju Thanks for your comments, I agree that behaviour would be less confusing. I've implemented an opt-in approach in these 2 commits, if you'd like to review? If you're happy with this approach, I'll squash the commits to remove the original opt-out approach.

https://github.com/alphagov/govuk_publishing_components/pull/1330/commits/6aab21a3c8edea6a0783f86ba7041bd35cc57b39
https://github.com/alphagov/govuk_publishing_components/pull/1330/commits/c0584dc9b18476f7a643837bd0bb043f8b082fc8

PR #1334, opened 2020-02-28
Fix handling of `on_govuk_blue` parameter value in the search component

kr8n3r:
## What

Check for `true` (boolean) value for `on_govuk_blue` parameter.

We used to check only if the parameter exist, not its value.
This made it difficult to conditionally toggle classes.

Now we check it the local_assign value is `true` before we add `gem-c-search--on-govuk-blue` CSS class to the component.

## Why
In finder-frontend we're looking to unify the page templates and we have scenarios where the search box will appear on either a white or blue background. At the moment the search component renders will appropriate classes as soon as `on_govuk_blue` local parameter is present without checking for `true`, which means we cannot alternate the appearance

## Visual Changes
No visual changes

### DIFF
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -1,6 +1,6 @@
 <%
   class_name = "gem-c-search--on-white"
-  class_name = "gem-c-search--on-govuk-blue" if local_assigns.include?(:on_govuk_blue)
+  class_name = "gem-c-search--on-govuk-blue" if local_assigns[:on_govuk_blue].eql?(true)
   size ||= ""
   class_name = "#{class_name} gem-c-search--large" if size == 'large'
   class_name = "#{class_name} gem-c-search--separate-label" if local_assigns.include?(:inline_label)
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -1,6 +1,9 @@
 <%
   class_name = "gem-c-search--on-white"
-  class_name = "gem-c-search--on-govuk-blue" if local_assigns.include?(:on_govuk_blue)
+  class_name = "gem-c-search--on-govuk-blue" if local_assigns.include?(:on_govuk_blue) &&
```
alex-ju: I can I suggest we do the following, for consistency
```suggestion
class_name = "gem-c-search--on-govuk-blue" if local_assigns[:on_govuk_blue].eql?(true)
```

kr8n3r: that was quick :)

kr8n3r: much more elegant!

PR #1337, opened 2020-03-03
Tweak pagination component spacing and add border

vanitabarrett:
Trello: https://trello.com/c/4ZsqNg3r/11-layout-improve-previous-and-next-navigation-component-on-mobile

## What
This PR makes the following changes to the pagination (previous and next) component:

- Increase spacing above and below the component
- Increase spacing between previous and next links
- Add border-top to links

Changes made as per the [design doc](https://docs.google.com/document/d/1ryu7LdOiJdKGqG_lUZOwDPOpLpSNw0MVAGaWx6ocmuw/edit#)

## Why
The pagination component needs to be a separate and distinct section of the page. These changes help separate the component from the rest of the page content. The spacing and border changes also add clearer separation between the previous and next links themselves.

## Before
Screenshot 2020-03-03 at 11 00 56
Screenshot 2020-03-03 at 11 01 30

## After
Screenshot 2020-03-03 at 11 00 52
Screenshot 2020-03-03 at 11 01 36

### Without label
Screenshot 2020-03-03 at 11 11 55
Screenshot 2020-03-03 at 11 10 46

### With label
Screenshot 2020-03-03 at 11 21 31
Screenshot 2020-03-03 at 11 20 54

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
@@ -1,6 +1,6 @@
 .gem-c-pagination {
   display: block;
-  margin: govuk-spacing(6) (- govuk-spacing(3) );
+  margin: govuk-spacing(8) 0;
 }
 
 .gem-c-pagination__list {
@@ -11,13 +11,17 @@
 .gem-c-pagination__item {
   @include govuk-font($size: 16, $line-height: (20 / 16));
   list-style: none;
+
+  &:first-child {
+    margin-bottom: govuk-spacing(4);
+  }
 }
 
 .gem-c-pagination__link {
   @extend %govuk-link;
   display: block;
-  padding: govuk-spacing(3);
   text-decoration: none;
+  padding-bottom: govuk-spacing(4);
 
   &:hover,
   &:active,
@@ -37,6 +41,7 @@
 
 .gem-c-pagination__link-title {
   display: block;
+  padding-top: govuk-spacing(3);
 }
 
 .gem-c-pagination__link-divider {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
@@ -36,11 +36,16 @@
 
   &:focus {
     @include govuk-focused-text;
+
+    .gem-c-pagination__link-title {
+      border-top-color: transparent;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
```diff
@@ -11,13 +11,18 @@
...
+    margin-bottom: 0;
+  }
 }
 
 .gem-c-pagination__link {
   @extend %govuk-link;
   display: block;
-  padding: govuk-spacing(3);
   text-decoration: none;
+  padding-bottom: 20px;
```
maxgds: Any reason not to use the govuk-spacing here?

app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
```diff
@@ -54,6 +69,7 @@
   margin-bottom: 1px;
   height: .482em;
   width: .63em;
+  fill: $govuk-secondary-text-colour;
```
maxgds: Why a text colour rather than the standard border colour? $govuk-border-colour

vanitabarrett: The border colour is a slightly paler grey, which means the arrow can't be seen very easily:

Screenshot 2020-03-03 at 12 04 13

However, the govuk-colour("mid-grey") is the same colour, so I can use the mixin instead as it's not text content.

vanitabarrett:
@andysellick When you have time, it would be good to get your review on this too as you had some concerns about the previous PR 🙂

injms:
I think that the focus state design is... interesting. Other patterns and components, such as the accordion, tabs, and the facets on search, have a different touch area and highlight area to stop the yellow from being too overpowering. I think that could be something worth considering here, especially on larger screens.

app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
```diff
@@ -11,13 +11,18 @@
 .gem-c-pagination__item {
   @include govuk-font($size: 16, $line-height: (20 / 16));
   list-style: none;
+  margin-bottom: govuk-spacing(4);
+
+  &:last-child {
```
andysellick: Since there's presumably only ever going to be two items in this component, would it be simpler to just put the margin-bottom on the first child?

vanitabarrett:
@injms I would say that's probably a separate piece of work, if that's ok? This PR hasn't actually changed the existing focus state. I'll make a note to ask Mia about it though

PR #1339, opened 2020-03-04
Linting fix

vanitabarrett:
## What
Switch from `govuk-lint` (deprecated) to `rubocop-govuk`.

## Why
`govuk-lint` is now deprecated.


vanitabarrett:
Kind of related to this, it looks like we're using `sass-lint` rather than the new GOVUK recommended approach of using `scss-lint-govuk`. I think we chose sass-lint to be in line with GOVUK Frontend, but should we be consistent with the rest of GOVUK?

govuk_publishing_components.gemspec
```diff
@@ -29,12 +29,12 @@ Gem::Specification.new do |s|
   s.add_development_dependency "capybara", "~> 3.25"
   s.add_development_dependency "foreman", "= 0.85"
   s.add_development_dependency "gds-api-adapters", "~> 59.1"
-  s.add_development_dependency "govuk-lint", "~> 3.11"
   s.add_development_dependency "govuk_schemas", "~> 3.2"
   s.add_development_dependency "govuk_test", "~> 1"
   s.add_development_dependency "jasmine", "~> 3.4.0"
   s.add_development_dependency "pry-byebug"
   s.add_development_dependency "rspec-rails", "~> 3.8"
+  s.add_development_dependency "rubocop-govuk", "~> 2"
```
issyl0:
```suggestion
  s.add_development_dependency "rubocop-govuk", "~> 3"
```

3.0.0 is the latest, released yesterday, and pulls in yet more RuboCop rules (and some improvements!)

issyl0:
You can also now back out the changes to the linting of files in the `bin/` directory as [RuboCop will never pick up on them with gem version 3](https://github.com/alphagov/rubocop-govuk/pull/14) as they're entirely generated files that we shouldn't change.

PR #1342, opened 2020-03-05
Update govspeak button styles again

andysellick:
## What
- remove 'button' class from extend for buttons in govspeak
- this change relates to an update of govspeak that switches from the toolkit buttons to the new component guide buttons
- after consideration I don't believe we need to extend both the gem-c-button class and the govuk-button class, just the gem-c-button class seems to work fine, adding both adding unnecessary CSS weight
- this change follows on from PR #1282 

## Why
This all ties in to the removal of the `.button` class from static: https://github.com/alphagov/static/pull/2035

Trello card: https://trello.com/c/d8D7JTJE/203-%F0%9F%8F%B4%E2%98%A0%EF%B8%8F-remove-design-patterns-buttons-from-static

## Visual Changes
None.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
@@ -3,13 +3,10 @@
 // scss-lint:disable PlaceholderInExtend
 // sass-lint:disable placeholder-in-extend
 .gem-c-govspeak {
-  .button,
-  .govuk-button,
   .gem-c-button {
     @extend .govuk-button;
   }
 
-  .button-start,
   .govuk-button--start {
     @extend .govuk-button--start;
   }
--- a/app/views/govuk_publishing_components/components/docs/govspeak.yml
+++ b/app/views/govuk_publishing_components/components/docs/govspeak.yml
@@ -673,21 +673,9 @@ examples:
           <p>This is a summary</p>
         </div>
   buttons:
-    description: |
-      We're in process of migrating govspeak from using the old (toolkit) button styles to using the (govuk-frontend) button component from this gem. The example below demonstrates both kinds of buttons.
-
-      This documentation will be updated once that work is complete.
     data:
       block: |
         <p>
-          <a href="#" class="button" role="button">
-            Old button
-          </a>
-          <a href="#" class="button button-start" role="button">
-            Old start button
-          </a>
-        </p>
-        <p>
-          <a role="button" class="gem-c-button govuk-button" href="https://gov.uk/random">New button</a>
-          <a class="gem-c-button govuk-button govuk-button--start" role="button" href="https://gov.uk/random"> New start button <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false"><path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"></path></svg></a>
+          <a role="button" class="gem-c-button govuk-button" href="https://gov.uk/random">Button</a>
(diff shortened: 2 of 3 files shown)


app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -673,21 +673,9 @@ examples:
...
-            Old button
-          </a>
-          <a href="#" class="button button-start" role="button">
-            Old start button
-          </a>
-        </p>
-        <p>
-          <a role="button" class="gem-c-button govuk-button" href="https://gov.uk/random">New button</a>
-          <a class="gem-c-button govuk-button govuk-button--start" role="button" href="https://gov.uk/random"> New start button <svg class="govuk-button__start-icon" xmlns="http://www.w3.org/2000/svg" width="17.5" height="19" viewBox="0 0 33 40" role="presentation" focusable="false"><path fill="currentColor" d="M0 0h13l20 20-20 20H0l20-20z"></path></svg></a>
+          <a role="button" class="gem-c-button govuk-button" href="https://gov.uk/random">Button</a>
```
injms: Since `gem-c-button` extends `govuk-button` with no additional styles, are both classes needed?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
```diff
@@ -3,13 +3,10 @@
...
 // sass-lint:disable placeholder-in-extend
 .gem-c-govspeak {
-  .button,
-  .govuk-button,
   .gem-c-button {
     @extend .govuk-button;
   }
 
-  .button-start,
   .govuk-button--start {
```
injms: Does this `.govuk-button--start` extend itself?

app/views/govuk_publishing_components/components/docs/govspeak.yml
andysellick: `gem-c-button` doesn't extend `govuk-button`, because that would duplicate `govuk-frontend` CSS inside the component gem CSS.

`gem-c-button` specifically provides a reduction in the bottom margin that `govuk-button` has: https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_button.scss#L14

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
andysellick: Yes, but necessarily. The reason for this whole bit of Sass is to override the styles within govspeak elements for links. It's a bit inefficient, in that `.govuk-button--start` is already defined, and we're duplicating the CSS by doing this, but without it the styles for that class get overwritten by govspeak - the button has link text colour instead of white, etc. It seems illogical but it's a necessary evil.

Now that you mention this I wonder if there's a world where we don't need this - by being more thoughtful about styling links within govspeak. We should be able to detect if a link is a button and if not, apply the `govuk-link` class to it.

injms: `:not()` for the win!

app/views/govuk_publishing_components/components/docs/govspeak.yml
injms: Okay makes sense. Similar to the other comment, I take it the extra specificity gained by extending inside a `gem-c-govspeak` is needed because of overriding link styles?

https://github.com/alphagov/govuk_publishing_components/blob/fffb3678ad30dbaf50cf1bb85fef1bf2b01873a3/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss#L5-L8

andysellick: Correct. It's not ideal, and I'm thinking about a better way of doing this, but that's outside the scope of this PR.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
andysellick: Yeah, something like that. But outside of the scope of this PR.

PR #1344, opened 2020-03-05
Enable custom classes and aria-label on button component

alex-ju:
## What
Add support for custom `class` and `aria-label` to the button component.

## Why
Currently need for the [reorderable list component](https://github.com/alphagov/content-publisher/pull/1819) in Content Publisher as follows:
- custom `class` is required to be able to query buttons from JavaScript using `js-` prefixed class (e.g. `.js-app-button-up` to identify a button that moves items upwards) without adding excessive wrapper elements.
- `aria-label` is required to give screen reader users context on what button acts on (e.g. "Up" button would have a "Move Budget 2019 up" as `aria-label`).

## Visual Changes
No visual changes.

[Trello card](https://trello.com/c/H3T7JWtS)

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -107,4 +107,12 @@ examples:
       text: "This is the button text"
       value: "this_is_the_value"
       name: "this_is_the_name"
-
+  with_js_classes:
+    description: Use `js-` prefixed classes only as interaction hooks – to query and operate on elements via JavaScript
+    data:
+      text: "Button"
+      classes: "js-selector-1 js-selector-2"
+  with_aria_label:
+    data:
+      text: "Button"
+      aria_label: "Button with custom label"
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -6,7 +6,7 @@ module GovukPublishingComponents
       attr_reader :href, :text, :title, :info_text, :rel, :data_attributes,
                   :margin_bottom, :inline_layout, :target, :type, :start,
                   :secondary, :secondary_quiet, :destructive, :name, :value,
-                  :classes
+                  :classes, :aria_label
 
       def initialize(local_assigns)
         @href = local_assigns[:href]
@@ -26,6 +26,7 @@ module GovukPublishingComponents
         @name = local_assigns[:name]
         @value = local_assigns[:value]
         @classes = local_assigns[:classes]
+        @aria_label = local_assigns[:aria_label]
       end
 
       def link?
@@ -42,6 +43,7 @@ module GovukPublishingComponents
         options[:target] = target if target
         options[:name] = name if name.present? && value.present?
         options[:value] = value if name.present? && value.present?
+        options[:aria] = { label: aria_label } if aria_label
         options
       end
 
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -25,7 +25,12 @@ module GovukPublishingComponents
         @destructive = local_assigns[:destructive]
         @name = local_assigns[:name]
         @value = local_assigns[:value]
-        @classes = local_assigns[:classes]
+        if local_assigns.include?(:classes)
+          @classes = local_assigns[:classes].split(" ")
+          unless @classes.all? { |c| c.start_with?("js-") }
+            raise(ArgumentError, "The button component expects classes to be prefixed with `js-`")
+          end
+        end
         @aria_label = local_assigns[:aria_label]
       end
 
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -5,7 +5,8 @@ module GovukPublishingComponents
     class ButtonHelper
       attr_reader :href, :text, :title, :info_text, :rel, :data_attributes,
                   :margin_bottom, :inline_layout, :target, :type, :start,
-                  :secondary, :secondary_quiet, :destructive, :name, :value
+                  :secondary, :secondary_quiet, :destructive, :name, :value,
+                  :classes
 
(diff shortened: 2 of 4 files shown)


lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -24,6 +25,8 @@ def initialize(local_assigns)
         @destructive = local_assigns[:destructive]
         @name = local_assigns[:name]
         @value = local_assigns[:value]
+        @classes = local_assigns[:classes]
```
andysellick: This allows an application to add an arbitrary class to the component, am I correct? If so I'm concerned about the precedent this sets - we don't want a world where applications apply custom classes to components for styling, breaking component isolation.

I appreciate the need here but is there an alternative way of doing it? If you want to add a unique hook to a button to trigger JS, could it be done with an ID or a data attribute instead?

alex-ju: > This allows an application to add an arbitrary class to the component, am I correct? If so I'm concerned about the precedent this sets - we don't want a world where applications apply custom classes to components for styling, breaking component isolation.

That is correct. There is no precedent, aside from me trying to introduce this shortly after joining GOV.UK. We already live in a world where applications add wrappers to components to adjust them accordingly (which in my opinion is no different than this, aside from generating excessive markup) or add local overrides to achieve a similar effect (where an app modifier can be used instead). Component isolation is a principle we should stick to when developing frontend code and while I agree we should guard our principles, we shouldn't solely rely on components code for this.

[In our principles](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md#a-component-is-isolated-when) we state that a component is isolated when:

_1. its styles and scripts have no effect on the page or other components_
We're using BEM as a strategy for this. This doesn't mean, for example, that someone can't grab a `govuk-breadcrumbs__list` class and use it in an `ul` outside the component – hence violating this principle.

_2. it does not rely on external selectors to style its children_
We're using the component guide to check how components look in isolation and make sure they don't rely on external styles. [An example of a violation](https://government-frontend.herokuapp.com/component-guide/government_navigation) is when a component depends on font styles being defined globally. Applying a modifier to a component from outside doesn't mean the component relies on it.

I conclusion, I consider this change does not violate our principles if used correctly – happy to discuss this matter as a community and reach a consensus.

> I appreciate the need here but is there an alternative way of doing it? If you want to add a unique hook to a button to trigger JS, could it be done with an ID or a data attribute instead?

Following our guidance, we should use [`js-` classes for interaction hooks](https://github.com/alphagov/govuk_publishing_components/blob/f6a9f48c4ac7efc549ffa799199e40e220384426/docs/javascript-modules.md#use-js--prefixed-classes-for-interaction-hooks) and not `id`s or `data` attributes.

andysellick: I agree that isolation should be practiced across the whole of FE development and not just relied upon by the way components are structured but this feels like opening a door that we'll have difficulty shutting - especially when the Design System guidance talks about passing classes to components if you want to restyle them, which is one of the key differences between their component model and ours.

If our guidance is to use `js-` classes for interaction hooks, how about an option to do that instead - any class passed would have `js-` appended to it.  That way the documentation would be unambiguous - passing this class is only for JS hooks, not for styling.

alex-ju: Shouldn't we be concerned then that people may use `js-` classes to style components?

andysellick: Maybe, except the documentation would be able to explicitly state the purpose of this, so anyone using it to do that would be knowingly circumventing the purpose of the option. My concern is that a generic `class` option invites a wider potential misuse, both through intent and lack of understanding.

alex-ju: I understand your concern. Mine is that people who are looking to make app-level adjustments to a component have only two alternatives: wrapping the element and leaking styles into the component using more specific selectors or use app-level overwrites for the component that may have wider effects than desired. I think both of these are not great either and we can already find them in frontend applications.

kevindew: I don't think this change makes much difference if people want to break the component isolation - this can be done pretty trivially by targeting .gem-c-button in a nesting - for example: https://github.com/alphagov/content-data-admin/blob/eed9ca67fdceb69e08280a812f9712ed7250559e/app/assets/stylesheets/content/_index.scss#L77-L80

It looks like people have got around this in the past by forgoing the component system and just writing the HTML, which seems an unfortunate outcome: https://github.com/alphagov/finder-frontend/blob/aa8d39862e8e8570b8e7844bb67960d394733f48/app/views/finders/_facet_collection.html.erb#L55-L57

If the only classes we want to allow are ones prefixed with js could we do something like:

```
@classes = local_assigns.fetch(:classes, [])
raise "provided classes must be prefixed with js-" unless @classes.all? { |c| c.match(/Ajs-/) }
```
I'd prefer doing that than automatically prefixing inputted classes as that feels surprising.

andysellick: Oooh, yes, both of those last suggestions ☝️ 

The other thing I was wondering about is why is our guidance that data attributes shouldn't be used for JS hooks? Is there a good reason for it?

alex-ju: While the suggested approach doesn't resolve [my concern above regarding extending components](https://github.com/alphagov/govuk_publishing_components/pull/1344#discussion_r389030957), I'm happy to raise that separately and move on with this PR by restricts the classes to `js-` prefixed only and raises an exception otherwise. Would you be able to review this again, @andysellick when you have time?

app/views/govuk_publishing_components/components/docs/button.yml
```diff
@@ -107,4 +107,12 @@ examples:
       text: "This is the button text"
       value: "this_is_the_value"
       name: "this_is_the_name"
-
+  with_js_classes:
+    description: Use `js-` prefixed classes only as interaction hooks – to query and operate on elements via JavaScript
```
andysellick: 👍

PR #1346, opened 2020-03-09
Make feedback component more responsive and usable on mobile

vanitabarrett:
## What
- Implement new design for feedback component, with larger touch targets and more responsive layout.
- Tidy up unused feedback component code
- Remove outdated sections of feedback component documentation

## Why
Small touch targets and small spacing between items makes it visually frustrating, but for mobile users this might make it harder to click on the intended link.
The previous design also failed when a user zoomed, as the links started to overflow the container.

## Before and After (desktop)
Screenshot 2020-03-09 at 10 42 08

Screenshot 2020-03-09 at 10 41 59

## Before and After (mobile)
Screenshot 2020-03-09 at 10 42 35
Screenshot 2020-03-09 at 10 42 50

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -1,23 +1,16 @@
 .gem-c-feedback {
   background: govuk-colour("white");
-  margin: 0 auto;
+  margin: govuk-spacing(6) auto 0 auto;
   max-width: $govuk-page-width;
   position: relative;
 
   @include govuk-media-query($from: tablet) {
+    margin-top: govuk-spacing(9);
     border-bottom: 1px solid govuk-colour("white");
   }
 
   .visually-hidden {
     @include govuk-visually-hidden;
-  }
-}
-
-.gem-c-feedback--top-margin {
-  margin-top: govuk-spacing(6);
-
-  @include govuk-media-query($from: tablet) {
-    margin-top: govuk-spacing(9);
   }
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -117,6 +117,29 @@
   }
 }
 
+.gem-c-feedback__option-list {
+  display: inline-block;
+  list-style-type: none;
+  margin: 0;
+  padding: 0;
+  margin-right: govuk-spacing(2);
+  margin-top: govuk-spacing(2);
+
+  @include govuk-media-query($from: tablet) {
+    margin-top: 0;
+  }
+}
+
+.gem-c-feedback__option-list-item {
+  display: inline-block;
+}
+
+.gem-c-feedback__option-list-item:first-child {
+  margin-right: govuk-spacing(7);
+}
+
+
+// Feedback form styles
 .gem-c-feedback__error-summary {
   margin-bottom: govuk-spacing(3);
   padding: govuk-spacing(3);
@@ -219,28 +242,5 @@
 
   &:focus {
     outline: $govuk-focus-width solid $govuk-focus-colour;
-  }
-}
-
-.js-enabled {
-  .gem-c-feedback__option-list {
-    display: inline-block;
-    list-style-type: none;
-    margin: 0;
-    padding: 0;
-    margin-right: govuk-spacing(2);
-    margin-top: govuk-spacing(2);
-
-    @include govuk-media-query($from: tablet) {
-      margin-top: 0;
-    }
-  }
-
-  .gem-c-feedback__option-list-item {
-    display: inline-block;
-  }
-
-  .gem-c-feedback__option-list-item:first-child {
-    margin-right: govuk-spacing(7);
   }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -248,16 +248,7 @@
     display: inline-block;
   }
 
-  .gem-c-feedback__option-list-item--useful {
+  .gem-c-feedback__option-list-item:first-child {
     margin-right: govuk-spacing(7);
-    @include govuk-media-query($until: tablet) {
-      display: inline-block;
-    }
-  }
-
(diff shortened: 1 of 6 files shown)


kr8n3r:
nice one for splitting the questions and being able to remove the grid 👍

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -35,39 +39,87 @@
   }
 }
 
+.js-enabled .gem-c-feedback__prompt-questions {
+  display: inline-block;
```
andysellick: Why is this set? It seems to be overriding a `display: block`, removing it doesn't seem to make any difference?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -35,39 +39,87 @@
...
+}
+
+.gem-c-feedback__prompt-questions {
+  position: relative;
+  text-align: center;
+  border-bottom: 1px solid govuk-colour("white");
+  padding: govuk-spacing(5) govuk-spacing(5) govuk-spacing(3) govuk-spacing(5);
+  box-sizing: border-box;
+  width: 100%;
```
andysellick: This width and the position relative on line 47 don't appear to do anything, but I could be wrong.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -186,79 +238,30 @@
...
-      // older grid spec
-      grid-row-gap: govuk-spacing(3);
-      // newer grid spec
-      row-gap: govuk-spacing(3); // sass-lint:disable-line no-misspelled-properties
-    }
+    margin-right: govuk-spacing(2);
   }
 
   .gem-c-feedback__option-list-item {
+    position: relative;
```
andysellick: Am I right in thinking this is to capture the pseudo element on the link? Only the link itself already has position relative, so I'm not sure this is needed?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -186,79 +238,30 @@
...
+    display: inline-block;
```
andysellick: It's quite hard to follow the inheritance tree through all of this (due to the current duplication of the CSS, not this PR) but I think there's some simplification to be had here. The heading `gem-c-feedback__prompt-question` is also floated and has margin to position it, but I think if all of the floating was removed and everything set to inline-block the required layout can be achieved with fewer lines of code.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -1,19 +1,16 @@
...
-.gem-c-feedback--top-margin {
-  margin-top: govuk-spacing(6);
-
   @include govuk-media-query($from: tablet) {
     margin-top: govuk-spacing(9);
+    border-bottom: 1px solid govuk-colour("white");
+  }
+
+  .visually-hidden {
+    @include govuk-visually-hidden;
```
andysellick: This isn't your change (in fact it's probably mine! but I can't remember) but do you know why the include is here, rather than using the class directly in the component? Seems like something worth fixing.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -1,19 +1,16 @@
...
-    @include govuk-visually-hidden;
-  }
-}
-
-.gem-c-feedback--top-margin {
-  margin-top: govuk-spacing(6);
-
   @include govuk-media-query($from: tablet) {
     margin-top: govuk-spacing(9);
+    border-bottom: 1px solid govuk-colour("white");
```
andysellick: Suggestion: currently this swaps between the `.gem-c-feedback` element having the border bottom above tablet and the `.gem-c-feedback__prompt-questions` having it on mobile (where both need it to have two, when stacked) - would it be possible to just have the border bottom on the `.gem-c-feedback__prompt-questions` elements and not need any media queries involved?

andysellick: Actually, I've just realised why this is difficult - at a certain breakpoint the heights of the two elements are inconsistent. Hmm. Maybe don't worry about this. Although maybe there's a solution..

Screenshot 2020-03-11 at 13 47 42

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -80,6 +117,30 @@
   }
 }
 
+.js-enabled {
+  .gem-c-feedback__option-list {
```
andysellick: Just wondering why there's so many rules nested beneath `.js-enabled` - doesn't the component simply not appear if JS is disabled?

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
vanitabarrett: Applying the border to `.gem-c-feedback__prompt-questions` was my first approach, but I also noticed the issue you had above so switched approaches. I tried again, but I think either way will require media queries.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
vanitabarrett: Yes, good point, I don't think this is needed any more since I tidied up some of the CSS. Will remove

vanitabarrett:
@andysellick Thanks for your review! I've addressed all your comments now, apart from the one about the border - I tried to find an alternative solution but couldn't find anything that would be simpler.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: Yeah, fair enough.

injms:
The link "Don't have an email address?" is the same text size as "Yes" and "No", but doesn't have the pseudo element to increase the touch target. Should the pseudo element be added to the link to also give it a larger touch target?

vanitabarrett:
Thanks @injms , good spot! I've increased the touch target on that link now

alex-ju:
🙌 for using pseudo-elements on this

PR #1352, opened 2020-03-11
Fix overflowing text in attachments

alex-ju:
## What
Fix overflowing URLs and email addresses in govspeak-rendered attachments and the attachment component.

## Why
To make sure the page layout doesn't break

### govspeak attachments
<table>
<tr><td>

Screenshot 2020-03-11 at 10 41 45

</td><td>

Screenshot 2020-03-11 at 10 39 32

</td></tr>
</table>

Reference: [the `url` class generated by govspeak](https://github.com/alphagov/govspeak/blob/f8131efa42c2838a6c94288bd0404e13b0f17ad2/lib/govspeak/presenters/attachment_presenter.rb#L39)

### attachment component

<table>
<tr><td>

Screenshot 2020-03-11 at 10 58 49

</td><td>

Screenshot 2020-03-11 at 10 58 07

</td></tr>
</table>

govspeak attachments problem on the live website: https://www.gov.uk/government/publications/covid-19-track-coronavirus-cases reported by @timpaul.

[attachments component problem tracked in Content Publisher](https://trello.com/c/IGI9Hf16).

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -70,6 +70,11 @@
         @include govuk-font($size: 14);
       }
 
+      .url {
+        word-break: break-word;
+        word-wrap: break-word;
+      }
+
       .changed,
       .references,
       .unnumbered-paper {
@@ -98,6 +103,8 @@
           @include govuk-font($size: 14);
           margin: 0;
         }
+        word-break: break-word;
+        word-wrap: break-word;
       }
 
       .js-hidden {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -70,6 +70,11 @@
         @include govuk-font($size: 14);
       }
 
+      .url {
+        word-break: break-word;
+        word-wrap: break-word;
```
ChrisBAshton: Is it worth documenting why we're setting both `word-break` and `word-wrap`? Or does this exist as prior art somewhere? I don't think I've seen the two properties used together before.

alex-ju: Sure, I can add more details in the commit message. `word-wrap` is meant to be the non-standard Microsoft implementation of `overflow-wrap` that I was trying to leverage since `word-break` is not supported.

PR #1354, opened 2020-03-11
Extract document-item component from document-list

alex-ju:
Raising this for visibility. Requires linting to pass in CI.

## What
Simplify document list markup when used in a table.

## Why
Each document is unnecessarily wrapped into an ordered list within the document list component (when presented in a table in Content Publisher). When rendering a single document with the component the ordered list should not be rendered.

## Visual Changes
TBD

[Trello card](https://trello.com/c/pJHJYOcK)


alex-ju:
Not a priority for Content Publisher anymore.

PR #1356, opened 2020-03-12
Fix duplicate CSS for components loading in component guide

andysellick:
## What
Since the change to allow applications to only include the sass for components they need, we've had duplicate CSS for all components loading in the component guide.

This change includes the sass for components used in the guide in the suggested sass and removes the sass for components from the guide stylesheet. There are two possible configurations for an application:

- the application includes the sass for all components, as before. Both the application and the guide have no styling issues or duplication
- the application includes the sass for only the components it is using, plus the components in use within the guide (8 components, although there may be some overlap, e.g. both probably use `breadcrumbs`). Both the application and the guide have no styling issues or duplication

This change fixes the duplicate CSS problem but it introduces two other possible issues worth noting:

- applications using the new suggested sass may include CSS from a few components that are only in use in the guide (I think currently this is likely to be `layout_header` and `layout_footer`) so a small amount of CSS will be loaded unnecessarily
- for applications that have already switched to the new individual sass feature (`collections` and `finder-frontend`) this may represent a slightly breaking change, although the worst that will happen is parts of the component guide might not be styled

## Why
The component guide loads its own stylesheet plus the stylesheet for the application it is running in. If it's not running in an application it loads the stylesheet from the dummy app.

In the old world, both the dummy app and the application included the sass for all components, and the guide stylesheet didn't include any, so there was no duplication. When the individual sass feature was introduced a problem arose - there are components in use by the guide that might not be in use by the application, so parts of the guide weren't being styled. To fix this, the sass for all components was added to the guide stylesheet, but this introduced this duplication.

This change now addresses this problem.

## Visual Changes
None.

Trello card: https://trello.com/c/FpI4ntWf/234-fix-duplicate-css-in-component-guide

### DIFF
--- a/app/controllers/govuk_publishing_components/component_guide_controller.rb
+++ b/app/controllers/govuk_publishing_components/component_guide_controller.rb
@@ -46,7 +46,7 @@ module GovukPublishingComponents
       additional_files = "@import 'govuk_publishing_components/govuk_frontend_support';\n"
       additional_files << "@import 'govuk_publishing_components/component_support';\n" unless print_styles
 
-      components = components_in_use("#{@application_path}/app/views/**/*.html.erb")
+      components = components_in_use("#{@application_path}/app/views/")
       extra_components = []
 
       components.each do |component|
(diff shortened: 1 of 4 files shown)


app/controllers/govuk_publishing_components/component_guide_controller.rb
```diff
@@ -110,5 +111,9 @@ def component_breadcrumb(component_doc, component_example = nil)
         h[:url] = component_doc_path(component_doc.id) if component_example
       end
     end
+
+    def components_used_by_component_guide
+      %w(breadcrumbs details layout-footer layout-header lead-paragraph search textarea title)
```
alex-ju: Can we automate these, in a similar way we do to determine which components are used in a certain app?

andysellick: Done, thanks for the suggestion.

PR #1375, opened 2020-03-19
Update attachment component to accommodate publications

alex-ju:
## What
Add reference details to attachments

## Why
To accommodate publications attachments and bring them close to how they are now rendered by Whitehall (in terms of presented data)

## Visual Changes
[Preview attachment component](https://govuk-publis-publicatio-e3wxt3.herokuapp.com/component-guide/attachment)

[Trello card](https://trello.com/c/UET4vx6k/1439-update-attachment-component-to-accommodate-new-attachment-design)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -66,6 +66,10 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey", $legacy: "grey-3");
   }
 }
 
+.gem-c-attachment__metadata--compact {
+  margin-bottom: 0;
+}
+
 .gem-c-attachment__abbr {
   text-decoration: none;
   cursor: help;
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -68,6 +68,11 @@
       <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
     <% end %>
 
+    <% if attachment.is_official_document %>
+      <%= tag.p link_to("Order a copy", "https://www.gov.uk/guidance/how-to-buy-printed-copies-of-official-documents", class: "govuk-link", target: "_blank"),
+        class: "gem-c-attachment__metadata" %>
+    <% end %>
+
     <% unless hide_opendocument_metadata %>
       <% if attachment.opendocument? %>
         <%= tag.p class: "gem-c-attachment__metadata" do %>
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -60,6 +60,10 @@
       <%= tag.p "Ref: #{attachment.reference}", class: "gem-c-attachment__metadata gem-c-attachment__metadata--compact" %>
     <% end %>
 
+    <% if attachment.unnumbered_reference.present? %>
+      <%= tag.p attachment.unnumbered_reference, class: "gem-c-attachment__metadata gem-c-attachment__metadata--compact" %>
+    <% end %>
+
     <% if attributes.any? %>
       <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
     <% end %>
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -56,6 +56,10 @@
             data: data_attributes %>
     <% end %>
 
+    <% if attachment.reference.present? %>
+      <%= tag.p "Ref: #{attachment.reference}", class: "gem-c-attachment__metadata gem-c-attachment__metadata--compact" %>
+    <% end %>
+
     <% if attributes.any? %>
       <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
     <% end %>
--- a/lib/govuk_publishing_components/presenters/attachment.rb
+++ b/lib/govuk_publishing_components/presenters/attachment.rb
@@ -61,6 +61,12 @@ module GovukPublishingComponents
         reference.join(", ")
       end
 
+      def unnumbered_reference
+        unnumbered_reference = "Unnumbered command paper" if attachment_data[:unnumbered_command_paper].eql?(true) && !attachment_data[:command_paper_number]
+        unnumbered_reference = "Unnumbered act paper" if attachment_data[:unnumbered_hoc_paper].eql?(true) && !attachment_data[:hoc_paper_number]
+        unnumbered_reference
+      end
+
       class SupportedContentType
         attr_reader :content_type_data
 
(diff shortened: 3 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -66,6 +66,10 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey", $legacy: "grey-3");
   }
 }
 
+.gem-c-attachment__metadata--compact {
```
benthorner: Could you help me understand the need for this style? As I understand it, we can have up to three lines of metadata for an attachment (ref, unnumbered, file type). I would have thought this needs to apply to one of the other lines, as well, if it applies to any.

lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -67,6 +67,10 @@ def unnumbered_reference
         unnumbered_reference
       end
 
+      def is_official_document
+        attachment_data[:command_paper_number].present? || attachment_data[:hoc_paper_number].present?
```
benthorner: We should also show the link if the document is unnumbered, as [per the screenshot on the card](https://trello.com/c/UET4vx6k/1439-update-attachment-component-to-accommodate-new-attachment-design).  A document is still official and can still be ordered (apparently), even if it doesn't have a number (e.g. using the unique ref).

spec/components/attachment_spec.rb
```diff
@@ -105,10 +105,60 @@ def component_name
...
+      attachment: {
+          title: "The government financial reporting review",
+          url: "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/791567/the_government_financial_reporting_review_web.pdf",
+          filename: "department-for-transport-information-asset-register.csv",
+          content_type: "application/pdf",
+          file_size: 20000,
+          number_of_pages: 7,
+          isbn: "978-1-5286-1173-2",
+          unique_reference: "2259",
+          command_paper_number: "67",
```
benthorner: Minor: a more realistic value here is like `Cd. 1234`.

benthorner:
Nice one @alex-ju 👍. This looks good to me. I've spotted a few small things that we should fix before this goes in. Another minor thing: it might help to add a link to the component guide (or a screenshot), in case someone is reviewing this who's not familiar with the repo.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
alex-ju: As per design, the reference-related lines don't have margin-bottom, compared to the other metadata lines which require a margin-bottom – this is the reason I introduced this modifier, to be able to style them differently.

benthorner: Ahhh, sorry. I just realised this applies to **all** metadata lines. All good 👍

lib/govuk_publishing_components/presenters/attachment.rb
alex-ju: Updated, not sure how I missed that.

lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -61,6 +61,12 @@ def reference
         reference.join(", ")
       end
 
+      def unnumbered_reference
+        unnumbered_reference = "Unnumbered command paper" if attachment_data[:unnumbered_command_paper].eql?(true) && !attachment_data[:command_paper_number]
```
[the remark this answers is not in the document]
alex-ju: One of the AC was: "If both a numbered paper parameter and the equivalent unnumbered paper parameter is passed, the numbered one takes priority (unnumbered is not rendered)" – that's why I introduced this logic in here.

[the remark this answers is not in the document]
benthorner: Yes, I forgot to mention: it's a bit arbitrary. As a counter example, we're showing both an act paper and a command paper number, even though a document can't be both. It's a minor point, so we could just leave it as-is.

app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -56,10 +56,23 @@
...
+    <% if attachment.unnumbered_reference.present? %>
+      <%= tag.p attachment.unnumbered_reference, class: "gem-c-attachment__metadata gem-c-attachment__metadata--compact" %>
+    <% end %>
+
     <% if attributes.any? %>
       <%= tag.p raw(attributes.join(', ')), class: "gem-c-attachment__metadata" %>
     <% end %>
 
+    <% if attachment.is_official_document %>
+      <%= tag.p link_to("Order a copy", "https://www.gov.uk/guidance/how-to-buy-printed-copies-of-official-documents", class: "govuk-link", target: "_blank"),
```
kevindew: Sorry to be late - should we set this as a link with no visited state since it's a bit confusing that if you visit one of these ever it looks like you've ordered a copy of every attachment.

lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -47,6 +47,30 @@ def alternative_format_contact_email
...
+          hoc_reference = "HC #{attachment_data[:hoc_paper_number]}"
+          hoc_reference += " #{attachment_data[:parliamentary_session]}" if attachment_data[:parliamentary_session].present?
+          reference << hoc_reference
+        end
+
+        reference.join(", ")
+      end
+
+      def unnumbered_reference
+        unnumbered_reference = "Unnumbered command paper" if attachment_data[:unnumbered_command_paper].eql?(true) && !attachment_data[:command_paper_number]
```
kevindew: Does Whitehall have any translations for these as I recall most of the other copy in this component is available in multiple languages

lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -47,6 +47,30 @@ def alternative_format_contact_email
...
+        unnumbered_reference = "Unnumbered act paper" if attachment_data[:unnumbered_hoc_paper].eql?(true) && !attachment_data[:hoc_paper_number]
+        unnumbered_reference
+      end
+
+      def is_official_document
+        attachment_data[:command_paper_number].present? || attachment_data[:hoc_paper_number].present? || attachment_data[:unnumbered_command_paper].eql?(true) || attachment_data[:unnumbered_hoc_paper].eql?(true)
```
kevindew: Oh man it's grizzly how we've got these potentially contradictory booleans and fields 😬

lib/govuk_publishing_components/presenters/attachment.rb
ChrisBAshton: Whitehall does have translations for things like “Unnumbered command paper”, but, for some reason, [only in Estonian](https://github.com/alphagov/whitehall/blob/master/config/locales/et.yml#L211).

govuk_publishing_components only has locale files for English and Welsh. I’m anticipating that adding an Estonian version will open a can of worms. For example, I don't know how the component preview generation might react in the presence of locale files which have missing keys. (I suspect nothing, but would rather not take the risk).

app/views/govuk_publishing_components/components/_attachment.html.erb
ChrisBAshton: Addressed by https://github.com/alphagov/govuk_publishing_components/pull/1427

PR #1386, opened 2020-03-20
Explicitly set accordion heading line height

andysellick:
## What / why

Our accordion component is unintentionally deviating from govuk-frontend, resulting in a slightly smaller touch area on mobile devices.

- govuk-frontend component does not appear to explicitly set the line height for the h2 in an accordion
- inside some applications h2 elements have some default styling which was subtly reducing the line height
- end result was on mobile, unopened accordions were 54px tall, not 60px as intended
- this fixes this by explicitly defining the h2 font on mobile

## Visual Changes

Before | After
----- | -----
Screenshot 2020-03-20 at 14 49 56 | Screenshot 2020-03-20 at 14 50 07

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
@@ -1,5 +1,15 @@
 @import "govuk/components/accordion/accordion";
 
+.gem-c-accordion {
+  .govuk-accordion__section-heading {
+    // this is a temporary addition to fix the line height when it
+    // is being overridden by styles from govuk-template in collections
+    @include govuk-media-query($until: desktop) {
+      @include govuk-font(24, $weight: bold, $line-height: 1.6);
+    }
+  }
+}
+
 .govuk-accordion--condensed {
   .govuk-accordion__section-button {
     @include govuk-media-query($from: tablet) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add custom margin option to accordion component ([PR #1470](https://github.com/alphagov/govuk_publishing_components/pull/1470))
+* Explicitly set accordion heading line height ([PR #1386](https://github.com/alphagov/govuk_publishing_components/pull/1386))
 
 ## 21.41.4
 


DilwoarH:
Change looks good - but should this be a flag as this will change everywhere?

andysellick:
I should have explained in the PR... this fixes an unintentional deviation from `govuk-frontend`. It shouldn't be an option, because it should be fixed everywhere.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,5 +1,13 @@
 @import "govuk/components/accordion/accordion";
 
+.gem-c-accordion {
+  .govuk-accordion__section-heading {
+    @include govuk-media-query($until: desktop) {
```
alex-ju: The default at the moment seems to be `1.25` for both mobile and desktop. How did we come up with `1.6`?

nickcolley:
Can this be raised as a bug on GOV.UK Frontend and fixed upstream?

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
andysellick: When it's at 1.6 the height of the touch area is in line with the accordion in the Design System.

andysellick:
These are fair comments. It's possible I should have done a better job of explaining the problem.

We're using the component in collections. Styles from static are setting line height on h2 elements to inherit. This inherits all the way up to the body, where styles from govuk-template set the line height to 1.5. This also seems to be complicated by other styles from govuk-template, and the user agent stylesheet (although why that's not a problem for govuk-frontend I'm not sure).

So obviously a slightly complex setup fix directly. This was discovered when testing the new covid-19 landing page, obviously something that needs to happen fast and work well. Although we will be dealing with static/govuk-template/etc. properly in time, the most efficient solution right now seems to be this, but I'm happy to discuss alternatives if you have them.

andysellick:
Also I'm not sure this is a bug... this is something that can happen whenever two sets of CSS are mixed. I'm not sure I want govuk-frontend to start filling up with explicit CSS to override stuff that may or not be there - I worry this could be the tip of an iceberg only a few people might collide with.

hannalaakso:
@andysellick I've investigated this, we should be able to fix this upstream: https://github.com/alphagov/govuk-frontend/issues/1789

andysellick:
@alex-ju I've raised this with the Design System team and they intend to fix it upstream (https://github.com/alphagov/govuk-frontend/issues/1789). Are you happy to approve and we can remove this when a fix is ready?

PR #1398, opened 2020-03-23
Rename colour variable

andysellick:
## What
Rename recently added colour variable to have a `gem-c` prefix.

## Why
Because it's already used elsewhere and might cause a conflict. Also this is good practice.

## Visual Changes
None.


andysellick:
@alex-ju fastest approve ever! 😁

PR #1402, opened 2020-03-24
Fix conditional reveal on checkboxes

alex-ju:
## What
Fix conditional reveal on checkboxes
Add examples for pre-checked items with conditonal reveal

## Why
GOV.UK Frontend script for conditional reveal (`checkboxes.js`) requires a `data-aria-controls` not a `data-controls` attribute as currently implemented. This being misconfigured means the script doesn't show the conditional content when pre-checked. This is a major problem when presenting errors to a field within the conditionally revealed block.

## Visual Changes
No visual changes, only functional changes: [Preview checkboxes with conditional revealed content in a pre-checked state](https://govuk-publis-fix-condit-svxz7n.herokuapp.com/component-guide/checkboxes/checkbox_items_with_conditional_reveal_checked) – you'll also see a bunch of axe errors which are not introduced by this change and shall be addressed, but I'll raise a separate issue for this.

<table>
<tr><td valign="top">

Screenshot 2020-03-24 at 19 04 34

</td><td valign="top">

Screenshot 2020-03-24 at 19 04 12

</td></tr>
</table>

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -285,9 +285,26 @@ examples:
         - label: "Phone"
           value: "phone"
           conditional: <div class="govuk-form-group"><label class="govuk-label" for="contact-by-phone">Phone number</label><input class="govuk-input govuk-!-width-one-third" id="contact-by-phone" name="contactingme" type="tel"></div>
-        - label: "Other"
-          value: "other"
-          conditional: "See no styling."
+        - label: "Text message"
+          value: "text"
+          conditional: <div class="govuk-form-group"><label class="govuk-label" for="contact-by-text">Mobile phone number</label><input class="govuk-input govuk-!-width-one-third" id="contact-by-text" name="contactingme" type="tel"></div>
+  checkbox_items_with_conditional_reveal_checked:
+    data:
+      name: "contacting-checked"
+      id: "contacting-checked"
+      heading: "How would you like to be contacted?"
+      hint_text: "Please select all options that are relevant to you."
+      items:
+        - label: "Email"
+          value: "email"
+          conditional: <div class="govuk-form-group"><label class="govuk-label" for="contact-by-email">Email address</label><input class="govuk-input govuk-!-width-one-third" id="contact-by-email" name="contactingme" type="email"></div>
+          checked: true
+        - label: "Phone"
+          value: "phone"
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -288,6 +288,23 @@ examples:
...
+        - label: "Email"
+          value: "email"
+          conditional: <div class="govuk-form-group"><label class="govuk-label" for="contact-by-email">Email address</label><input class="govuk-input govuk-!-width-one-third" id="contact-by-email" name="contactingme" type="email"></div>
+          checked: true
+        - label: "Phone"
+          value: "phone"
+          conditional: <div class="govuk-form-group"><label class="govuk-label" for="contact-by-phone">Phone number</label><input class="govuk-input govuk-!-width-one-third" id="contact-by-phone" name="contactingme" type="tel"></div>
+        - label: "Other"
+          value: "other"
+          conditional: "See no styling."
```
injms: I don't understand this conditional - why "See no styling."?

alex-ju: Ah, I copied only this over from the example above and added a `checked: true`. I can look for a decent example.

injms: I mean, it's definitely not a blocker - I just wondered why a person want to show 'See no styling' 😕. Don't worry about another example 😄

alex-ju: 🤷‍♂ Ideally, they should all be real-life examples, since we introduce each feature with a certain need – but we got skittles examples instead 🙂

injms: You mean `which-skittles-do-you-like.service.gov.uk` doesn't exist? 🤯

alex-ju: Replaced that with contact by text via mobile phone 😁

andysellick: I think `See no styling` is meant to highlight the fact that the component isn't meant to provide styles for the content of the conditional element - that's left to the developer.

injms: Ah, that makes sense - thanks!

PR #1411, opened 2020-03-27
Fix back link in Safari

### LINKED ISSUE #1387: Back link arrow doesn't display correctly in Safari
The back link renders strangely in Safari (13.0.5), with the arrow only showing one side of the chevron:

Screenshot 2020-03-21 at 09 26 02

36degrees:
## What

Safari only supports clip-path with the -webkit prefix, which is being being used to apply the clip-path in GOV.UK Frontend. In order to unset it, we also need to use the -webkit prefix here.

## Why

See #1387

### Before

Screenshot 2020-03-27 at 15 58 04

### After

Screenshot 2020-03-27 at 15 58 33

Fixes #1387

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix back link arrow rendering in Safari ([PR #1411](https://github.com/alphagov/govuk_publishing_components/pull/1411))
+
 ## 21.36.0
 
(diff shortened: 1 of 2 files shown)


36degrees:
I've only tested this change by making it in Dev Tools at the minute, so please hold whilst I get the dev environment set up and check it actually does what I've said it does…

_Installing sassc 2.2.1 with native extensions_ ⏱

36degrees:
Yep, all good.

Screenshot 2020-03-27 at 16 05 18

alex-ju:
I tested it too before approving it 💚

PR #1415, opened 2020-04-01
Add heading option to warning component

andysellick:
## What
- allows the component to be rendered as a heading
- uses the shared helper code for headings

## Why
We're using the styles from the component on a heading element on [travel pages like this one](https://www.gov.uk/foreign-travel-advice/usa/entry-requirements) but not the component itself. This change will allow us to switch to using the component.

## Visual Changes
Screenshot 2020-04-01 at 10 50 06

### DIFF
--- a/app/views/govuk_publishing_components/components/_warning_text.html.erb
+++ b/app/views/govuk_publishing_components/components/_warning_text.html.erb
@@ -4,6 +4,9 @@
   text_icon ||= '!'
   large_font ||= false
   highlight_text ||= false
+  heading_level ||= 0
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   text_classes = %w(govuk-warning-text__text)
   text_classes << "gem-c-warning-text__text--no-indent" if text_icon.empty?
@@ -15,8 +18,17 @@
   <% unless text_icon.empty? %>
     <%= tag.span text_icon, class: "govuk-warning-text__icon", "aria-hidden": "true" %>
   <% end %>
-  <%= tag.strong class: text_classes do %>
+  <% inner_text = capture do %>
     <%= tag.span text_assistive, class: "govuk-warning-text__assistive" %>
     <%= text %>
+  <% end %>
+  <% if heading_level > 0 %>
+    <%= content_tag(shared_helper.get_heading_level, class: text_classes) do %>
+      <%= inner_text %>
+    <% end %>
+  <% else %>
+    <%= tag.strong class: text_classes do %>
+      <%= inner_text %>
+    <% end %>
   <% end %>
 <% end %>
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_warning_text.html.erb
```diff
@@ -15,8 +18,15 @@
   <% unless text_icon.empty? %>
     <%= tag.span text_icon, class: "govuk-warning-text__icon", "aria-hidden": "true" %>
   <% end %>
-  <%= tag.strong class: text_classes do %>
-    <%= tag.span text_assistive, class: "govuk-warning-text__assistive" %>
-    <%= text %>
+  <% if heading_level > 0 %>
```
injms, line 21: Would this need refining? Would it render `<h123>` with `heading_level: 123`?

app/views/govuk_publishing_components/components/_warning_text.html.erb
```diff
@@ -15,8 +18,15 @@
...
+    <%= content_tag(shared_helper.get_heading_level, class: text_classes) do %>
+      <%= tag.span text_assistive, class: "govuk-warning-text__assistive" %>
```
vanitabarrett, line 23: I wonder if we can combine these two lines into something reusable, as they're always used regardless of whether it's a heading or not?

```
<%= tag.span text_assistive, class: "govuk-warning-text__assistive" %>
<%= text %>
```

app/views/govuk_publishing_components/components/_warning_text.html.erb
andysellick: The shared helper code already handles this - if you pass anything invalid it defaults to a H2 - https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/shared_helper.rb#L22

app/views/govuk_publishing_components/components/_warning_text.html.erb
andysellick: Yeah, I wondered about that, but everything I tried reduced the two pairs of lines to one but added at least two more lines of code i.e. no actual lines saved. In the end I gave up on the basis that this is at least clear to read. Happy to accept a suggestion though?

vanitabarrett: For me it was less about the number of lines of code and more about having multiple places to update if we want to make changes in the future.

We could maybe pull it out into a helper or just do something like this:
```
<% text_capture = capture do %>
  <%= tag.span text_assistive, class: "govuk-warning-text__assistive" %>
  <%= text %>
<% end %>

...


if heading
    <%= text_capture %>
else
    <%= text_capture %>
```

andysellick: 👍

PR #1418, opened 2020-04-01
Allow livestream youtube URLs

vanitabarrett:
## What
Expand our video player code to handle livestream links.
This PR also fixes a Javascript error that was happening because the video title property couldn't be reached.

## Why
Our video player code expected a video URL that looks like this: 
`https://www.youtube.com/watch?v=y6hbrS3DheU` - essentially a youtube URL with a videoId (v) query param.

This means that it didn't support livestream URLs which look like this: 
`https://www.youtube.com/embed/live_stream?channel=UCoMdktPbSTixAyNGwb-UYkQ`
These URLs are structured slightly differently with 'live_stream' and a channel ID instead of videoId.

## Visual Changes
Screenshot 2020-04-02 at 10 19 10

Preview URLs:

- [Standard Youtube Embed](https://govuk-publis-spike-live-ftkn8m.herokuapp.com/component-guide/govspeak/with_youtube_embed)
- [Livestream Youtube Embed](https://govuk-publis-spike-live-ftkn8m.herokuapp.com/component-guide/govspeak/with_youtube_livestream)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -54,6 +54,8 @@
     youtubeVideoContainer.className += 'gem-c-govspeak__youtube-video'
     youtubeVideoContainer.innerHTML = '<span id="' + elementId + '" data-video-id="' + id + '"></span>'
 
+    options['title'] = $link.textContent
+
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -20,11 +20,19 @@
```
DilwoarH, line 20: Nothing major but should this just return if there isn't a link?

PR #1419, opened 2020-04-01
🚧 Copy over remaining JavaScripts from Frontend Toolkit

injms:
## What
Copying over three JavaScript files from Frontend Toolkit, so we can move towards removing Frontend Toolkit as a dependency for the frontend apps that render `www.gov.uk`.

## Why

A recent audit showed that these JavaScript files are still being used on `www.gov.uk` - deleting them would probably break things.

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/lib/shim-links-with-button-role.js
```diff
@@ -0,0 +1,36 @@
+// migrated from govuk_frontend_toolkit
+
+// javascript 'shim' to trigger the click event of element(s)
```
alex-ju, line 3: We don't need to migrate this one as it's covered by [button.js in govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/button/button.js#L21-L30).

app/assets/javascripts/govuk_publishing_components/lib/selection-buttons.js
```diff
@@ -0,0 +1,126 @@
+// migrated from govuk_frontend_toolkit
+
+// NOTICE
+// IF you are using GOV.UK Elements 3.0,0 or above then you do not need to use
```
alex-ju, line 4: We don't have any projects still using GOV.UK Elements – hence this script is not needed or used across GOV.UK applications.

app/assets/javascripts/govuk_publishing_components/lib/auto-track-event.js
```diff
@@ -0,0 +1,31 @@
+// migrated from govuk_frontend_toolkit
```
alex-ju, line 1: This is only used in old publishing tools via [govuk_admin_template which has its own copy of this code](https://github.com/alphagov/govuk_admin_template/blob/0dfd5cffe8ec6ffa2d29767531f1e64826773b7e/app/assets/javascripts/govuk-admin-template/modules/auto_track_event.js).

alex-ju: I was wrong with this, [`wrapper layout in static` still uses it as a tertiary dependency](https://github.com/alphagov/static/blob/976e5659c8c199ce9a7a79425eae0474dc880a5c/app/assets/javascripts/application.js#L4) that does automagic tracking in `frontend` 🤦

injms:
Closed because this pull request is not passing 😡 and an (almost) [identical pull request](https://github.com/alphagov/govuk_publishing_components/pull/1606) is passing 🤷‍♂️

PR #1437, opened 2020-04-07
Make FAQs look deeper in the page for questions

oscarwyatt:
Make FAQs look deeper in the page for questions

## What
Make FAQs look deeper in the page for questions. Previously, it looked only at immediate children of the body for h2 elements, this looks recursively through the DOM

## Why
Some pages (eg https://www.gov.uk/government/publications/coronavirus-outbreak-faqs-what-you-can-and-cant-do/coronavirus-outbreak-faqs-what-you-can-and-cant-do) have the h2s deeper in the DOM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Make FAQs look deeper in the page for questions ([PR #1437](https://github.com/alphagov/govuk_publishing_components/pull/1437))
+
 ## 21.41.1
 
 * Fix button alignment on cookie-banner ([PR #1450](https://github.com/alphagov/govuk_publishing_components/pull/1450))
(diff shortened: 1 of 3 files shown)


lib/govuk_publishing_components/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukPublishingComponents
-  VERSION = "21.38.4".freeze
+  VERSION = "21.39.4".freeze
```
andysellick, line 2: @oscarwyatt we don't modify the version number of the gem as part of a code change PR. Changes are made as normal PRs and then when a new version of the gem is needed we follow [this procedure](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md).

andysellick: e.g. https://github.com/alphagov/govuk_publishing_components/pull/1429

PR #1438, opened 2020-04-07
Add cookie banner variation for services

alex-ju:
## What
Add cookie banner variation for services.
With this addition, we also enable custom title and confirmation messages for the cookie banner.

## Why
For services asking for cookies used for analytics only, we have a simpler approach (similarly with other GDS services, such as Notify, Wifi). We introduce this variation to be used across C19-related services.

## Visual Changes
[Preview example](https://govuk-publis-update-coo-zbw2lb.herokuapp.com/component-guide/cookie_banner/in_services_asking_for_analytics_only)

Co-Authored-By: @bilbof

[Trello card](https://trello.com/c/TObzdyej)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add application stylesheet where needed ([PR #1436](https://github.com/alphagov/govuk_publishing_components/pull/1436))
+* Add cookie banner variation for services ([PR #1438](https://github.com/alphagov/govuk_publishing_components/pull/1438))
 
 ## 21.38.4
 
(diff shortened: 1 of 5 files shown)


leenagupte:
@alex-ju Isn't this the same as: https://govuk-publis-update-coo-zbw2lb.herokuapp.com/component-guide/cookie_banner/with_custom_content ?

alex-ju:
> @alex-ju Isn't this the same as: https://govuk-publis-update-coo-zbw2lb.herokuapp.com/component-guide/cookie_banner/with_custom_content ?

I'm a bit confused about the question as the `with_custom_content` example is introduced in this PR. The difference between the `with_custom_content` and `in_services_asking_for_analytics_only` examples is that the latter has a different set of actions compared to the default banner (the one we use in the GOV.UK main website).

leenagupte:
> > @alex-ju Isn't this the same as: https://govuk-publis-update-coo-zbw2lb.herokuapp.com/component-guide/cookie_banner/with_custom_content ?
> 
> @leenagupte I'm a bit confused about the question as the `with_custom_content` example is introduced in this PR. The difference between the `with_custom_content` and `in_services_asking_for_analytics_only` examples is that the latter has a different set of actions compared to the default banner (the one we use in the GOV.UK main website).

Sorry! Didn't read the code, just saw that there were two virtually identical variations.

PR #1440, opened 2020-04-08
Add Youtube Embed Analytics

DilwoarH:
## What
We need to be able to track when users are viewing the Livestream on GOV.UK.

We should use the YouTube API to send data to GA.

https://developers.google.com/youtube/iframe_api_reference#Retrieving_playlist_information

## Why
It will help us determine if it is a useful feature to include on GOV.UK.

Most commonly people use GTM to track videos on sites but we have security concerns about this.

Using the Youtube API we should be able to get around this but it is more complicated. It will give us more accurate information about viewing numbers and how users interact with the video, E.g do users pause it, start watching late and watch at increased speed? Currently, we are unable to do so and at best we can estimate the number of users but it is not accurate. For us to work this out we'd need this data to be passed to GA or equivalent as events.

## Done when
It's done when we can accurately record how many users viewed the live stream at any point.

It would be nice to be able to record how users are interacting with the video with events to trigger during the course of the stream.

https://trello.com/c/zAQSVUqi/124-add-livestream-tracking

## Demo

https://govuk-publis-youtube-an-h7le99.herokuapp.com/component-guide/govspeak/with_youtube_analytics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased changes
+
+* Adds tracking for youtube ([PR #1440](https://github.com/alphagov/govuk_publishing_components/pull/1438))
+
 ## 21.39.0
 * Add cookie banner variation for services ([PR #1438](https://github.com/alphagov/govuk_publishing_components/pull/1438))
 
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -21,17 +21,30 @@
     for (var i = 0; i < $youtubeLinks.length; ++i) {
       var $link = $youtubeLinks[i]
       var href = $link.getAttribute('href')
+      var hasTracking = $link.getAttribute('data-youtube-player-analytics') == "true"
+      var options = {
+        link: $link
+      }
+
+      if (hasTracking) {
+        options.tracking = {
+          hasTracking: hasTracking,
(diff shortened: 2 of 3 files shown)


DilwoarH:
Not sure how I would test this, tried a few methods didn't work.

DilwoarH:
> Sorry, ignore my previous comment, I was testing without the `data-youtube-player-analytics-category` defined!
> 
> Looks good, the only thing I noticed was that if a user skips backwards or forwards in the video, a 'VideoPaused' and 'VideoPlaying' event fires in quick succession. I'm guessing because technically behind the scenes the video is being paused and then played again from a different point, but I don't know if this might be misleading for PAs?

Good spot, I'll read through the documentation to see if there is anything we can do to work around that.

DilwoarH:
> @DilwoarH Have you tested this locally against a frontend app? I just tried testing against Collections and I got a Javascript error:
> 
> ```
> Uncaught TypeError: Cannot read property 'hasTracking' of undefined
> ```

Thanks for finding this, I pushed a fix

DilwoarH:
> > Sorry, ignore my previous comment, I was testing without the `data-youtube-player-analytics-category` defined!
> > Looks good, the only thing I noticed was that if a user skips backwards or forwards in the video, a 'VideoPaused' and 'VideoPlaying' event fires in quick succession. I'm guessing because technically behind the scenes the video is being paused and then played again from a different point, but I don't know if this might be misleading for PAs?
> 
> Good spot, I'll read through the documentation to see if there is anything we can do to work around that.

Hey, I had a look, don't think there is a nice way of doing this without putting loads of custom code in. 

I will let the analytics team know this is something that happens. 

Also - FYI added in all the states now as requested by the team.

https://github.com/alphagov/govuk_publishing_components/pull/1440/files#diff-b4722dbdd6fe718214588e4addc706d4R121

PR #1441, opened 2020-04-08
Version 21.38.6

bilbof:
## What

Releases https://github.com/alphagov/govuk_publishing_components/pull/1438

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,12 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 21.39.0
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -7,10 +7,12 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 21.38.6
```
injms, line 10: Nit pick - I think that this is a minor upgrade rather than a patch. Happy either way though.

alex-ju: Agree with Ian, as we're introducing a new feature, otherwise it's good to go.

bilbof:
Thanks! I've made that change.

PR #1443, opened 2020-04-09
Fix hide link on cookie banner

bilbof:
## What

This adds an event listener to all cookie banner hide buttons. Previously only the first hide button would hide the banner.

querySelectorAll returns a list of elements, whereas querySelector returns only the first element found. Since there are two hide links now we want to add an event listener to all of them.

## Visual Changes

https://govuk-publis-bilbof-fix-dk83zo.herokuapp.com/component-guide/cookie_banner/in_services_asking_for_analytics_only

https://trello.com/c/oT8l91fv/198-bugfix-hide-button-on-cookie-banner-component-should-hide-banner

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Adds tracking for youtube ([PR #1440](https://github.com/alphagov/govuk_publishing_components/pull/1438))
 * Fix excess newlines for the Attachment Link component ([PR #1442](https://github.com/alphagov/govuk_publishing_components/pull/1442))
+* Bugfix: Ensure cookie banner hide button always hides the cookie banner ([PR #1443](https://github.com/alphagov/govuk_publishing_components/pull/1443)).
 
 ## 21.39.0
 
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -17,9 +17,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   CookieBanner.prototype.setupCookieMessage = function () {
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNRELEASED
+
+* Bugfix: Ensure cookie banner hide button always hides the cookie banner.
```
alex-ju, line 12:
```suggestion
* Ensure cookie banner hide button always hides the cookie banner. ([PR #1443](https://github.com/alphagov/govuk_publishing_components/pull/1443))
```

alex-ju:
> Is it possible that we can have two of this element in the same component? it looks like it's defined twice in the component within an else condition - maybe I'm missing something?

Yes, we have two variations of cookie banner (the default one, used across gov.uk and the newly introduced version, used on services). In the second variation you can dismiss the banner by not accepting the cookies (the 'No' button) or by accepting the cookies and then dismissing the confirmation message (the 'Hide' button).

> I wonder if we have a nicer example of iterating through a querySelectorAll that we can use rather than a for loop. I forever forget what is common and supported though.

`forEach` is not supported by IE11 and above, in lack of a polyfill the old school `for` is still the norm.

CHANGELOG.md
alex-ju: Needs a rebase before merging for the changelog (as another PR got merged meanwhile)

kevindew:
> Yes, we have two variations of cookie banner (the default one, used across gov.uk and the newly introduced version, used on services). In the second variation you can dismiss the banner by not accepting the cookies (the 'No' button) or by accepting the cookies and then dismissing the confirmation message (the 'Hide' button).

Thanks - I missed that the second one wasn't in the conditional.

bilbof:
Thanks both!

PR #1459, opened 2020-04-20
Load application styles after gem styles in the component guide

alex-ju:
## What
Load application styles after gem styles (same order as scripts and all assets in apps).

## Why
In frontend applications, styles/scripts from the gem are loaded first, then the application-level styles/scripts which sometimes depend or inherit from the ones in the gem. With this change, we apply the same order to the component guide so that components look and behave the same as they do when used in the application.

## Visual Changes
Some visual fixes for components presented in the guide across apps.


andysellick:
@alex-ju good change. What precipitated it?

alex-ju:
Unexpected styling on some existing app-level components (Content Publisher and new COVID-19 services) – mainly link styles being overridden.

PR #1460, opened 2020-04-21
Allow input component to accept an error id

andysellick:
## What
- allows an id for the error message to be passed if the component is in an error state
- normally auto generates one but there are situations where we want to define the error id e.g. if using the error summary that links to the errors highlighted

Screenshot 2020-04-21 at 09 04 49

## Why
Needed for the feedback application.

## Visual Changes
None.


andysellick:
@alex-ju so the behaviour on https://www.gov.uk/contact/govuk is wrong then? If you choose 'a specific page' and clear the shown input, then submit the form, the error message at the top links to the error message for the input, not the input itself.

andysellick:
Closing this PR as the behaviour it was designed to allow for is incorrect, and is therefore unneeded.

PR #1461, opened 2020-04-22
Expand scope of suggested sass functionality

andysellick:
## What / why
- was previously looking for instances of component use only in the view, but government-frontend had one component call in a ruby helper, which was overlooked, resulting in an unstyled component
- now includes all ruby files in /app/ when looking for components

## Visual Changes
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


vanitabarrett:
The change itself looks good 👍 Does this have any impact on how quick the suggestions are to generate?

andysellick:
@vanitabarrett good question. Probably. I've not measured it. I haven't noticed the component guide being slow since I implemented this change. If it is slow, or becomes slow, I'm happy to move this functionality off the guide homepage.

PR #1470, opened 2020-04-24
Add custom margin option to accordion component

andysellick:
## What
Adds an option to set the margin bottom for the accordion component. Defaults to 30px as previous behaviour.

## Why
We need to change the bottom margin of the component in collections.

## Visual Changes
Screenshot 2020-04-24 at 11 37 23

### DIFF
--- a/app/views/govuk_publishing_components/components/_accordion.html.erb
+++ b/app/views/govuk_publishing_components/components/_accordion.html.erb
@@ -1,11 +1,14 @@
 <%
+  local_assigns[:margin_bottom] ||= 6
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   id ||= "default-id-#{SecureRandom.hex(4)}"
   items ||= []
   condensed ||= false
+
   accordion_classes = %w(gem-c-accordion govuk-accordion)
   accordion_classes << 'govuk-accordion--condensed' if condensed
+  accordion_classes << (shared_helper.get_margin_bottom)
 
   data_attributes ||= {}
   data_attributes[:module] = 'govuk-accordion'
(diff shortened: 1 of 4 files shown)


spec/components/accordion_spec.rb
```diff
@@ -96,6 +96,33 @@ def component_name
...
+    test_data = {
+      items: [{
+                heading: { text: "Heading 1" },
+                summary: { text: "Summary 1." },
+                content: { html: "<p>Content 1.</p>" },
+              }],
+    }
+
+    render_component(test_data)
+    assert_select '.gem-c-accordion.govuk-\!-margin-bottom-6'
```
DilwoarH, line 109: I would personally avoid using `govuk-!-margin-bottom-6` in a component as it adds an`!important` css.

`margin-bottom: 30px !important;`

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -1,11 +1,14 @@
 <%
+  local_assigns[:margin_bottom] ||= 6
```
DilwoarH, line 2: Are we sure we want to set this as the default?

spec/components/accordion_spec.rb
andysellick: This is using the `shared_helper.rb`, which is used by a number of other components. The good thing about it is that if we decide to change these classes we can do it for all components that use this, by modifying that.

This seems like a wider conversation and one that shouldn't impact this PR though.

app/views/govuk_publishing_components/components/_accordion.html.erb
andysellick: Yes, because 30px was what it was set to originally, and we don't want to change the default behaviour.

spec/components/accordion_spec.rb
DilwoarH: Ok - I'll approve the PR but this is something important we should bring up with everyone because these classes could potentially cause styling issues when implemented.

PR #1472, opened 2020-04-24
Allow individual script imports

alex-ju:
## What
Allow individual, component-based, JavaScript imports.

## Why
Follow up on #1159 to complete [RFC 108](https://github.com/alphagov/govuk-rfcs/pull/110)
To avoid excessive code from being imported in the application.

## Notes
~~This PR will remain draft until tested with applications consuming the gem but comments/reviews are welcome.~~

Tested with:
 - [x] [`content-publisher`](https://github.com/alphagov/content-publisher/tree/test-individual-script-imports)
 - [x] [`finder-frontend`](https://github.com/alphagov/finder-frontend/tree/test-individual-script-imports)
 - [x] [`frontend`](https://github.com/alphagov/frontend/tree/test-individual-script-imports)

For applications dependent on `static`/`slimmer`, `static` requires to be updated first to use the modules system from this gem instead of `govuk_frontend_toolkit`.

### Modules in GOV.UK
JavaScript modules in GOV.UK are actually design patterns for keeping pieces of code independent of other components. We currently have 3 design patterns for writing JavaScript in GOV.UK applications.

#### GOV.UK Frontend Toolkit Modules
Mostly found in legacy scripts, either migrated from Toolkit, Elements, or written in the same era.
```
  var that = this
  that.start = function (element) {
    //
  }
}
```

#### GOV.UK Publishing Modules
Found across GOV.UK applications, on relatively new scripts (2 years or younger). Similar to the `govuk-frontend` approach described below, it relies on the classic prototype pattern but with a `start` function that takes the element that initialises on. This `start` function was preserved so that the module initialisation script (`modules.js`) would work for both legacy and new scripts without any alteration required.
```
function TestModule () { }
TestModule.prototype.start = function (element) {
  //
}
```

#### GOV.UK Frontend Modules
Used in `govuk-frontend` scripts. Relies on the classic prototype pattern with an  `init` function. The element is passed to the constructor function.
```
function TestModule (element) {
  this.element = element
}
TestModule.prototype.init = function () {
  //
}
```

## Visual Changes
No visual changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,12 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Allow individual JavaScript imports ([PR #1472](https://github.com/alphagov/govuk_publishing_components/pull/1472))
+
+  To import individual components refer to ['Include the assets section in docs'](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md) and the 'Suggested imports for this application' section available on `/component-guide` in your application.
+
 ## 21.50.1
 
 * Tweak action link ([PR #1518](https://github.com/alphagov/govuk_publishing_components/pull/1518))
@@ -202,7 +208,7 @@
 
 ## 21.26.0
 
-* Allow individual components to be imported into apps ([PR #1159](https://github.com/alphagov/govuk_publishing_components/pull/1159))
+* Allow individual Sass imports ([PR #1159](https://github.com/alphagov/govuk_publishing_components/pull/1159))
 
 ## 21.25.0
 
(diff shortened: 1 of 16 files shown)


andysellick:
@alex-ju this looks good - will there be a 'suggested JS' feature in the guide like for the Sass?

alex-ju:
> @alex-ju this looks good - will there be a 'suggested JS' feature in the guide like for the Sass?

It would be great to have a similar feature indeed, but I would rather do it as a separate PR – not to overload this with changes and because it feels like an independent piece of work.

andysellick:
@alex-ju that's sensible, but can you provide that code in a comment so we can test this in an application?

alex-ju:
> @alex-ju that's sensible, but can you provide that code in a comment so we can test this in an application?

@andysellick I realised it might be simpler to add the suggested imports to this PR instead of providing a few examples that would limit testing. Added links to branches in repos I tested this with.

I found some time today and update the PR to contain suggested imports as shown below.
Screenshot 2020-05-04 at 17 22 20

alex-ju:
Thank you, @bilbof! I rebased and updated the JavaScript usage section and referred it from the changelog.

PR #1473, opened 2020-04-27
Adapt phase banner to allow embedding govspeak

kevindew:
This changes the phase banner to allow a govspeak component to be
embedded and still render ok. The reason for this is so that content
designers can be provided with the means to specify the phase banner
text which may require links embedded into the text.

To do this I've made changes to the HTML structure of the phase
component so that it no longer is only inline elements in a p element.
This allows the govspeak component to be inserted without the HTML
becoming invalid.

To allow the govspeak to be rendered in a style appropriate for this
component I've created two new modifiers for the component: small
and unspaced. These allow using the equivalent of govuk-frontends
body-small size and removing typography margins.

Example:

Screenshot 2020-04-27 at 18 50 16

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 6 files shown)


kevindew:
Doesn't look like we need this 🎉

PR #1476, opened 2020-04-29
Update step by step header component style

DilwoarH:
Removes top border
Changes border bottom colour to blue
Adds margin top of 15px

---

## Context
We know that 70% of traffic to coronavirus content is direct to content rather than via landing page. We want to provide users with a clear route to the landing page, and to useful related content. This is to help them discover support that they might not already be aware of or think to search for.

## What
We need to adapt the step by step superbreadcrumb component to be able to be used in this case.

The feature will provide users with a link to the landing page or hub that the content is tagged to using the taxonomy, using the 'superbreadcrumb'.

It will also provide users with a set of related links which will be drawn from the curated content associated with the lowest level topic that the content item is tagged to.

## Why
Providing users with clear routes to discover useful content relevant to them.

https://trello.com/c/IOjnO55X/196-update-superbreadcrumb-component-ready-for-coronavirus-tagged-content

## Before

Screenshot 2020-04-29 at 12 22 52
Screenshot 2020-04-29 at 12 23 03

## After

Apr-29-2020 11-40-26

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-header.scss
@@ -3,26 +3,24 @@
   position: relative;
   padding: 10px;
   background: govuk-colour("light-grey", $legacy: "grey-4");
-  border-top: solid 1px govuk-colour("mid-grey", $legacy: "grey-2");
-  border-bottom: solid 1px govuk-colour("mid-grey", $legacy: "grey-2");
+  border-bottom: solid 1px govuk-colour("blue");
+  margin-top: govuk-spacing(3);
 
   @include govuk-media-query($from: tablet) {
-    padding: govuk-spacing(3);
+    padding: govuk-spacing(5);
   }
 }
 
 // scss-lint:disable SelectorFormat
 
 .gem-c-step-nav-header__part-of {
-  @include govuk-font(16, $weight: bold);
-
+  @include govuk-font(19, $weight: bold);
   display: block;
-  padding-bottom: .2em;
 }
 
 .gem-c-step-nav-header__title {
   @extend %govuk-link;
-  @include govuk-font(24, $weight: bold);
+  @include govuk-font(24, $weight: bold, $line-height: 1);
 }
 
 // scss-lint:enable SelectorFormat
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix margin on buttons with info text ([PR #1474](https://github.com/alphagov/govuk_publishing_components/pull/1474))
+* Update step by step header component style ([PR #1476](https://github.com/alphagov/govuk_publishing_components/pull/1476))
 
 ## 21.42.0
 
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-header.scss
```diff
@@ -3,8 +3,8 @@
   position: relative;
   padding: 10px;
   background: govuk-colour("light-grey", $legacy: "grey-4");
-  border-top: solid 1px govuk-colour("mid-grey", $legacy: "grey-2");
-  border-bottom: solid 1px govuk-colour("mid-grey", $legacy: "grey-2");
+  border-bottom: solid 1px govuk-colour("blue");
```
alex-ju, line 6: Do we know what was the reasoning for changing the border colour? We'll still have to preserve the `$legacy` attribute since the compatibility mode is still on in some applications.

DilwoarH:
> On this page https://www.gov.uk/guidance/the-highway-code?step-by-step-nav=e01e924b-9c7c-4c71-8241-66a575c2f61f the component collides with the inverse header beneath it. We might want to think about that, whether there's a way of adding margin bottom to the component without increasing the spacing everywhere else, or just adding a small amount?

@samdub - this is how it would look with the page above.

Screenshot 2020-04-29 at 12 32 18

Screenshot 2020-04-29 at 12 34 22

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav-header.scss
DilwoarH: Not sure what the reason is for the blue line but here is the trello - https://trello.com/c/IOjnO55X/196-update-superbreadcrumb-component-ready-for-coronavirus-tagged-content

I've added in the legacy colour

alex-ju: Thank you! Don't have access to the board, unfortunately.

36degrees: Just to clarify, you only _need_ to provide a `$legacy` attribute if the colour with that name does not exist in the legacy palette – which is the case with `mid-grey` (the legacy palette has `grey-1` through `grey-4` instead).

As there is a colour called `blue` in both the new and legacy palettes, it'll just use the blue from whichever respective palette, without needing to pass `$legacy`.

alex-ju: ah, thank you for pointing this out @36degrees. my bad, @DilwoarH 🤦 , `blue` is still `blue`, I was thinking we'll go with `mid-gray`.

DilwoarH: Thanks for confirming this @36degrees

@alex-ju - most of the information about the context is in the PR description. Here is a copy of the design that I was given:
Corona_content_nav_mob-1

alex-ju: Well, assuming this was well thought by a designer, I'm happy to get it in.

kevindew:
I'm glad to see the border top going - it was a pain point on https://github.com/alphagov/frontend/pull/2343

DilwoarH:
Before: https://components.publishing.service.gov.uk/component-guide/step_by_step_nav_header/preview
After: https://govuk-publis-update-ste-ppy66n.herokuapp.com/component-guide/step_by_step_nav_header/preview

@alex-ju - probably need a re-review as I've made a few more changes

DilwoarH:
Design reviewed by Sam and Mia offline

PR #1481, opened 2020-05-04
Add suffix to inputs

GDSNewt:
## What

Add a new suffix option to the input component. 

## Why

Users were entering data which was unusable, i.e entering text when we desired numbers. This change will help us indicate to users how we would like them to input data. 

### Cross-browser testing
Latest Chrome, Firefox, Safari
Chrome

Latest Edge
Screenshot 2020-05-05 at 15 10 22

IE 11
IE11

IE 10
IE10

IE9
IE9

IE8
IE8

[Trello card](https://trello.com/c/7VVBYT7z/257-add-suffix-to-input-fields-on-product-details-question)

Co-Authored-By: @alex-ju

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
@@ -16,3 +16,37 @@
     padding-left: govuk-spacing(6);
   }
 }
+
+.gem-c-input__wrapper {
+  display: flex;
+  display: -ms-flexbox; // sass-lint:disable-line no-duplicate-properties
+
+  .gem-c-input {
+    display: inline-block;
+    flex: 0 1 auto;
+  }
+
+  .gem-c-input:focus {
+    z-index: 1; // increase z-index on focus so it appears over the suffix
+  }
+}
+
+.gem-c-input__suffix {
+  @include govuk-font($size: 19);
+
+  background-color: govuk-colour("light-grey");
+  border: $govuk-border-width-form-element solid $govuk-input-border-colour;
+  border-left: 0;
+  box-sizing: border-box;
+  cursor: default; // emphasise non-editable status of prefixes and suffixes
+  display: inline-block;
+  flex: 0 0 0;
+  padding: govuk-spacing(1);
+  white-space: nowrap;
+  width: auto;
+  height: 40px;
+  @if $govuk-typography-use-rem {
+    height: govuk-px-to-rem(40px);
+  }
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -16,3 +16,24 @@
     padding-left: govuk-spacing(6);
   }
 }
+
+.gem-c-input__wrapper {
+  display: flex;
```
andysellick, line 21: Do we need a fallback for this? I'm guessing probably not but thought it was worth asking.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -16,3 +16,24 @@
...
+}
+
+.gem-c-input:focus {
+  z-index: 1; // increase z-index on focus so it appears over the suffix
+}
+
+.gem-c-input__suffix {
+  // scss-lint:disable PlaceholderInExtend
+  // sass-lint:disable placeholder-in-extend
+  @extend .govuk-input;
```
andysellick, line 31: I've only briefly played with this but it seems like instead of doing this `@extend` the `govuk-input` class could be added to this element. Would save a few lines of code.

(I point this out partly because I have a vague memory of `@extend` causing us headaches at some point to do with compilation, but I can't remember where unfortunately)

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
alex-ju: Excellent point, Andy. I've updated the code to fallback to display inline for IE9 and below and used the prefixed property for IE10. Also added screen captures to the PR description to ease the review.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
alex-ju: No real headaches (aside from generating quite a bit of css that may be too much for this scenario), but not ideal and sanctioned by linter when used on classes (not placeholders). 

The main reason we went with `@extend` is that we want to make sure the suffix styles are aligned with the input styles in case the later change. There is a bit of logic around the height of the input which is generated differently based on `$govuk-typography-use-rem` flag and I wanted to avoid it. However, I removed the `@extend` and got the logic in – we can propose the Design System team to expose `.govuk-input` styles to be used as a placeholder if there will be future needs around it.

PR #1488, opened 2020-05-07
Add 'or' divider to checkboxes component

alex-ju:
## What
Allow 'or' divider on checkboxes (similarly with the radio component).

## Why
To be used in a pair with the exclusive option.
Example of a page using this pattern: https://coronavirus-business-volunteers.service.gov.uk/expert-advice-type where checkboxes with an 'or' option are not part of the same fieldset. This makes it harder for screen readers to use those forms.

## Visual Changes
<table>
<tr><th>Without</th><th>With</th></tr>
<tr><td valign="top">

Screenshot 2020-05-07 at 19 37 18

</td><td valign="top">

Screenshot 2020-05-07 at 19 36 43

</td></tr>
</table>

[Trello card](https://trello.com/c/kJe25doD)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Prevent govspeak tables from breaking the layout ([PR #1493](https://github.com/alphagov/govuk_publishing_components/pull/1493))
 * Prevent govspeak links from breaking the layout ([PR #1492](https://github.com/alphagov/govuk_publishing_components/pull/1492))
 * Remove PHE brand colour ([PR #1489](https://github.com/alphagov/govuk_publishing_components/pull/1489))
+* Add 'or' divider to checkboxes component ([PR #1488](https://github.com/alphagov/govuk_publishing_components/pull/1488))
 * Fix legacy colour on input component ([PR #1487](https://github.com/alphagov/govuk_publishing_components/pull/1487))
 * Add `exclusive` option to checkboxes component ([PR #1478](https://github.com/alphagov/govuk_publishing_components/pull/1478))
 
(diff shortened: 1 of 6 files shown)


maxgds:
Technically this seems sound, necessary and works. I find myself questioning the semantics, though. Is "or" really a list item? Is the alternative option (or options) part of the original list? Feels like it should be more like
<ul>
<li> a </li>
<li> b </li>
<li> c </li>
</ul>
<p>or</p>
<ul>
<li> none of the above </li>
</ul>

alex-ju:
> Technically this seems sound, necessary and works. I find myself questioning the semantics, though. Is "or" really a list item? Is the alternative option (or options) part of the original list? Feels like it should be more like
> 
> ```
> <li> a </li>
> <li> b </li>
> <li> c </li>
> </ul>
> <p>or</p>
> <ul>
> <li> none of the above </li>
> </ul>```
> ```

Good point about the semantics, Max. The implementation of [the component in `govuk-frontend` uses `div`s instead of `li`s](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/checkboxes/template.njk#L52) which may not be as semantic on a simple example, but allows for separators like this to be added easier.

I'll think about the suggestion above to see if there are any limitations.

andysellick:
I'd worry slightly about separating the two lists if they're related. I think there should be some kind of connection between the first set of options, the 'or' and the final option for screen readers, and including them all in the same list seems like the simplest way. 

If they were split into separate lists, a screen reader would announce (in the example in the screenshot) that there are three options, so users might be confused that none of the options apparently available apply to them.

alex-ju:
I'm tempted to 'rebase' this to match the Design System model (which is using more generic `divs`) which will also align it with the radios component. Objections?

PR #1497, opened 2020-05-14
Add action link component

andysellick:
## What
Adds the action link component from `collections` and adds a new option.

Screenshot 2020-05-14 at 21 56 14

Apart from the new option, the component is unchanged from `collections` apart from the `gem-c` namespace and syntax tweaks to get the tests to run, so the first commit in this PR shouldn't need much scrutiny.

## Change from original - subtext

This new option adds the ability to have subtext - text that appears alongside the link on desktop and beneath it on mobile. This is made complicated by two things; the need for the text to vertically align, and the need for a dividing line between the two but only on desktop (and not knowing how long that text will be in practice).

**Desktop:**

Screenshot 2020-05-14 at 21 57 34

**Mobile:**

Screenshot 2020-05-14 at 22 01 25

This is solved using `display: table-cell` and `vertical-align: middle` on desktop for the individual elements and by adding some new wrapping elements. The wrapping elements are necessary for the dividing line between the link and the subtext. Adding spacing to the right of the link would oddly increase the focus state size (although this was considered). 

We also need a parent class on the whole component when subtext is present to style the link differently.

This solution was chosen over flexbox because it works better cross browser without browser prefixes (we used flexbox in an early iteration of this component and getting it to work in all the IEs, even modern ones, was a headache never quite solved).

## Change from original - compact and dark, and margins

We previously had two options, dark and compact, that basically did the same thing, except the compact version put the link text closer to the arrow. This was for use on a background of the same colour, where the dark part of the arrow icon wouldn't be visible.

That was okay-ish when this component was only in an application, but it feels unnecessarily confusing now, so I've consolidated these changes together by removing the compact version and reducing the space between the arrow and the link slightly in the dark version.

I've also added in the shared helper so we can pass a margin bottom option to the component, which was also part of the compact version.

Compact version on a dark background before (top) and after (bottom):

Screenshot 2020-05-18 at 12 43 39

## Why

I'm adding this component to the gem because we are using it `collections` but there's a version of it in `static` as well and maintaining two of them has now passed the point of practicality. The plan is to add this to the gem then use it in `static` and `collections`.

Trello card: https://trello.com/c/l9xuuuTB/279-update-emergency-banner-design


injms:
I'm seeing some odd clipping in Safari on macOS when the link over two lines and focused:

Screenshot 2020-05-15 at 09 30 07

injms:
The dark version of the icon is different to the light version of the icon - it's not an inverse version. There's a space between the arrow and the left side of the circle on the dark version, whilst there's no space on the other version. Do you know if this is a deliberate choice?

andysellick:
@alex-ju I think we've diverged a bit since the initial borrowing, including all of the variants and different uses throughout the coronavirus landing page. We initially included the arrow in the focus state but unlike the NHS version our text can get quite verbose so it led to a very large yellow block when focussed, which we decided was unnecessary. In the case of the NHS, they specifically use it to signpost to a service, whereas we just use for important content.

andysellick:
@injms thanks will look at both those issues.

andysellick:
@injms I've fixed the line-height issue and reworked some of the component options in light of your comments (see the `Change from original - compact and dark, and margins` in the updated PR description).

Having said all that, it doesn't actually address your comment about the light icon not being the actual inverse of the dark one, but yes, the slight difference is deliberate.

PR #1502, opened 2020-05-18
Reduce re-importing of SCSS files

kevindew:
## What

Primarily this changes the _all_components helper file to use the same support scripts as we recommend users use to import individual components.

Aside from that I then removed a number of @import calls that were superfluous

## Why

Primarily I went in looking to see if there was much low hanging fruit available where we were making unnecessary @import calls, there wasn't actually too much of that now that we support both individual components and all component importing. However I did find a few of those and, given they can add up to steep performance penalties, felt they were best removed.

I also felt it was more prudent to do the same support set-up in the _all_components file as we recommend users take in individual components. This seemed to reduce the risk that these differ in configuration.

## Visual Changes

I very much hope there are none

@andysellick and @alex-ju: I felt you two might be best suited to know if this seems a good addition and whether I was accidentally breaking any patterns/consistencies we were aiming for.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Change `_all_components.scss` to use dependencies from support stylesheets  ([PR #1502](https://github.com/alphagov/govuk_publishing_components/pull/1502))
 * Allow individual JavaScript imports ([PR #1472](https://github.com/alphagov/govuk_publishing_components/pull/1472))
 
   To import individual components refer to ['Include the assets section in docs'](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md) and the 'Suggested imports for this application' section available on `/component-guide` in your application.
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -1,12 +1,5 @@
 // This file contains the styles for the Component Guide.
 
-@import "govuk_publishing_components/components/helpers/govuk-frontend-settings";
-@import "govuk_publishing_components/components/helpers/markdown-typography";
-@import "govuk/settings/all";
-@import "govuk/tools/all";
-@import "govuk/helpers/all";
-@import "govuk/core/all";
-
 @import "govuk_publishing_components/all_components";
 
 $gem-guide-border-width: 1px;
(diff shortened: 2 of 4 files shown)


andysellick:
This looks really good. I'd like to spend more time looking at it but don't want to hold you up. Have you tested it in an application, using both of the component inclusion models and checking the component guide? (sorry, I'm just suggesting more work here)

andysellick:
Have tested this in `collections` using the all components model and the individual components model, both work fine.

andysellick:
Have tested this in `collections` using the all components model and the individual components model, seems to work fine.

andysellick:
(@kevindew don't forget to update the changelog)

kevindew:
Great, thanks for trying it out Andy and glad to hear you got good results. 

I'll try find some today to do more checking myself (I'll plug it into Content Publisher) and try diffing output for any clues of differences. The area that most concerns me is the switch from "govuk/all" to "govuk/components/all" where I wonder if there could be a subtlety (such as an ordering change) that affects things. Helper wise, I'm less concerned as it'll blow up if they're missing.

change note wise, I didn't think this needed one since this is an internal change. You should only notice this if either a consumer is doing something odd or I've broken it 😱

kevindew:
I've gave it a run through with content publisher and not found any problems. I checked each component and the only one I saw a dubious result on was modal - however that was consistent with https://components.publishing.service.gov.uk/component-guide/modal_dialogue (if you click open modal there is a blur effect applied to the modal - note this wasn't actually the case using the modal in content publisher app)

Comparing with master branch I see the [dummy application css (which wraps all components)](https://github.com/alphagov/govuk_publishing_components/blob/5bdb678b35292b916c6467da3453ae8847eb45e1/spec/dummy/app/assets/stylesheets/application.scss) comes out with the same number of rules (2530) and exact same filesize (332525 bytes) however the ordering is slightly different - we seem to have govuk frontend helpers output before components for instance. The [print css file](https://github.com/alphagov/govuk_publishing_components/blob/5bdb678b35292b916c6467da3453ae8847eb45e1/spec/dummy/app/assets/stylesheets/print.scss) remains exactly the same (as we'd expect).

I've used the first result on Google for online diff to produce something that shows the change with a chunk of govuk-frontend css moving positions: https://www.diffchecker.com/1PO9k8Ig

kevindew:
Thanks @andysellick - I'll drop one in before I merge.

PR #1509, opened 2020-05-19
Add prefix option to input component

GDSNewt:
## What

Add a prefix option to the input component

## Why

Users were entering data which was unusable, i.e entering text when we desired numbers. This change will help us indicate to users how we would like them to input data.

## Visual Changes

**BEFORE**

Screenshot 2020-05-19 at 08 26 24

**AFTER**
Screenshot 2020-05-19 at 16 47 55

[Trello](https://trello.com/c/GCGNaCSH/367-add-prefix-to-input-fields-on-product-details-question)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,8 @@
 
   To import individual components refer to ['Include the assets section in docs'](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md) and the 'Suggested imports for this application' section available on `/component-guide` in your application.
 
+* Add prefix option to input component ([PR #1509](https://github.com/alphagov/govuk_publishing_components/pull/1509))
+
 ## 21.50.1
 
 * Tweak action link ([PR #1518](https://github.com/alphagov/govuk_publishing_components/pull/1518))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
@@ -31,7 +31,7 @@
   }
 }
 
-.gem-c-input__suffix {
+.gem-c-input__prefix, .gem-c-input__suffix {
   @include govuk-font($size: 19);
 
   background-color: govuk-colour("light-grey", $legacy: "grey-3");
@@ -40,13 +40,22 @@
   box-sizing: border-box;
   cursor: default; // emphasise non-editable status of prefixes and suffixes
   display: inline-block;
-  flex: 0 0 0;
-  padding: govuk-spacing(1);
   white-space: nowrap;
   width: auto;
+  text-align: center;
   height: 40px;
+  padding: govuk-spacing(1);
+  min-width: 40px;
   @if $govuk-typography-use-rem {
-    height: govuk-px-to-rem(40px);
+      min-width: govuk-px-to-rem(40px);
   }
   margin-top: 0;
+}
+
+.gem-c-input__prefix {
+  border-right: 0;
+}
+
+.gem-c-input__suffix {
+  border-left: 0;
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_input.scss
@@ -31,12 +31,12 @@
   }
 }
 
-.gem-c-input__prefix, .gem-c-input__suffix {
+.gem-c-input__prefix,
+.gem-c-input__suffix {
   @include govuk-font($size: 19);
 
   background-color: govuk-colour("light-grey", $legacy: "grey-3");
   border: $govuk-border-width-form-element solid $govuk-input-border-colour;
-  border-left: 0;
   box-sizing: border-box;
   cursor: default; // emphasise non-editable status of prefixes and suffixes
   display: inline-block;
@@ -47,7 +47,7 @@
   padding: govuk-spacing(1);
   min-width: 40px;
   @if $govuk-typography-use-rem {
-      min-width: govuk-px-to-rem(40px);
+    min-width: govuk-px-to-rem(40px);
   }
   margin-top: 0;
 }
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -22,6 +22,7 @@
   error_id = "error-#{SecureRandom.hex(4)}"
   search_icon ||= nil
   heading_size = false unless ['s', 'm', 'l', 'xl'].include?(heading_size)
+  prefix ||= nil
   suffix ||= nil
 
   css_classes = %w(gem-c-input govuk-input)
@@ -92,7 +93,13 @@
     }
   %>
 
-  <% if prefix %>
+  <% if prefix && suffix %>
+    <%= tag.div class: "gem-c-input__wrapper" do %>
+      <%= tag.span prefix, class: "gem-c-input__prefix", aria: { hidden: true } %>
+      <%= input_tag %>
+      <%= tag.span suffix, class: "gem-c-input__suffix", aria: { hidden: true } %>
+    <% end %>
+  <% elsif prefix %>
     <%= tag.div class: "gem-c-input__wrapper" do %>
       <%= tag.span prefix, class: "gem-c-input__prefix", aria: { hidden: true } %><%= input_tag %>
     <% end %>
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -92,7 +92,11 @@
     }
   %>
 
-  <% if suffix %>
+  <% if prefix %>
+    <%= tag.div class: "gem-c-input__wrapper" do %>
+      <%= tag.span prefix, class: "gem-c-input__prefix", aria: { hidden: true } %><%= input_tag %>
+    <% end %>
+  <% elsif suffix %> 
     <%= tag.div class: "gem-c-input__wrapper" do %>
       <%= input_tag %><%= tag.span suffix, class: "gem-c-input__suffix", aria: { hidden: true } %>
     <% end %>
--- a/spec/components/input_spec.rb
+++ b/spec/components/input_spec.rb
@@ -215,6 +215,16 @@ describe "Input", type: :view do
     assert_select ".govuk-label.govuk-label--xl"
   end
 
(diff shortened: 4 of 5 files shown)


CHANGELOG.md
```diff
@@ -8,7 +8,7 @@
   of the commit log.
 
 ## Unreleased
-
+* Add prefix option to input component ([PR #1508](https://github.com/alphagov/govuk_publishing_components/pull/1508))
```
andysellick, line 11: 1509

GDSNewt: 👍 Cheers Andy

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -22,8 +22,9 @@
   error_id = "error-#{SecureRandom.hex(4)}"
   search_icon ||= nil
   heading_size = false unless ['s', 'm', 'l', 'xl'].include?(heading_size)
+  prefix ||= nil
   suffix ||= nil
-
+
```
andysellick, line 27: Stray whitespace

spec/components/input_spec.rb
```diff
@@ -215,13 +215,23 @@ def component_name
...
   end
 
+  it "renders text input with prefix" do
+    render_component(
+      label: { text: "Cost per item, in pounds" },
+      name: "cost-per-item",
+      prefix: "£",
+    )
+    assert_select ".gem-c-input__wrapper .gem-c-input[name=cost-per-item]"
+    assert_select ".gem-c-input__wrapper .gem-c-input__affix[aria-hidden=true]", text: "£"
```
andysellick, line 225: Should this test check for the `--prefix` class? (same for line 235 but `--suffix`)

spec/components/input_spec.rb
```diff
@@ -215,13 +215,23 @@ def component_name
...
+
   it "renders text input with suffix" do
     render_component(
       label: { text: "Approximate lead times, in days" },
       name: "lead-times",
       suffix: "days",
     )
     assert_select ".gem-c-input__wrapper .gem-c-input[name=lead-times]"
-    assert_select ".gem-c-input__wrapper .gem-c-input__suffix[aria-hidden=true]", text: "days"
+    assert_select ".gem-c-input__wrapper .gem-c-input__affix--suffix[aria-hidden=true]", text: "days"
```
simonwhatley, line 235: This should use `gem-c-input__suffix`

spec/components/input_spec.rb
```diff
@@ -215,13 +215,23 @@ def component_name
...
   end
 
+  it "renders text input with prefix" do
+    render_component(
+      label: { text: "Cost per item, in pounds" },
+      name: "cost-per-item",
+      prefix: "£",
+    )
+    assert_select ".gem-c-input__wrapper .gem-c-input[name=cost-per-item]"
+    assert_select ".gem-c-input__wrapper .gem-c-input__affix--prefix[aria-hidden=true]", text: "£"
```
simonwhatley, line 225: This should use `gem-c-input__prefix`

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -49,4 +48,12 @@
     height: govuk-px-to-rem(40px);
   }
   margin-top: 0;
+
+  &.gem-c-input__affix--prefix {
```
simonwhatley, line 52: The prefix and suffix should be separate element classes rather than modifiers.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -31,12 +31,11 @@
   }
 }
 
-.gem-c-input__suffix {
+.gem-c-input__affix {
```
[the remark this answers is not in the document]
andysellick: This is already following BEM convention. I think this is fine.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
andysellick: Why? They're basically the same element but for the different border.

GDSNewt:
> The prefix (and suffix) should have a minimum width which is the height of the input and the text aligned centre. Currently, the text is quite constrained and difficult to read if you have visual impairments.
> 
> For example:
> Screenshot 2020-05-19 at 08 58 56
> 
> NB. In the example screenshot, I have used `min-height: 40px` and `text-align: center`.


**Hi Simon, I was able to achieve the desired effect by being more explicit with the padding in the styling. Is this better?**

Screenshot 2020-05-19 at 10 46 55
Screenshot 2020-05-19 at 10 46 51

app/views/govuk_publishing_components/components/_input.html.erb
GDSNewt: I think this is okay, there was already a space between 'suffix' and 'css_classes', I assume for readability.

spec/components/input_spec.rb
GDSNewt: Good shout, changed as recommended

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: I meant the tab character, the newline is fine.

GDSNewt: 🤦 oh i'm an idiot. Thanks will change now

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
GDSNewt: Hi Simon, I made them modifiers to cut down on the amount of repeated code and to make maintenance easier. I will concede that separate element classes are perhaps simpler though. 

I'm not sure i'm following your point on adding separate modifiers to the prefix and suffix classes. In the case of errors they work on the input component itself rather than the prefix/suffix. 

Screenshot 2020-05-19 at 11 08 08

Is there a scenario you can envision in which we may need to add a modifier that is not applicable to both the prefix and suffix?

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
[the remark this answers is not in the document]
alex-ju: I agree with @simonwhatley on this one. The prefix and suffix spans are separate elements and they can potentially have modifiers of their own. We can always define common styles for both (e.g. `.gem-c-input__prefix, gem-c-input__suffix`) and then adjust the border in a separate definition.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -92,7 +93,11 @@
     }
   %>
 
-  <% if suffix %>
+  <% if prefix %>
```
simonwhatley, line 96: Prefix and suffix shouldn't be mutually exclusive. For example, income could be '£500/day' requiring the '£' prefix and 'day' suffix. (Of course, we could avoid this scenario with an appropriate question, for example "How much do you earn a day?")

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -31,19 +31,45 @@
   }
 }
 
+.gem-c-input__prefix {
```
alex-ju, line 34: I would suggest reducing the repetition of these declarations by grouping them
```
.gem-c-input__prefix, .gem-c-input__suffix {
  /* common styles */
}
```
followed by
```
.gem-c-input__prefix {
  /* individual styles */
}
```

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -31,19 +31,45 @@
...
+  @include govuk-font($size: 19);
+
+  background-color: govuk-colour("light-grey", $legacy: "grey-3");
+  border: $govuk-border-width-form-element solid $govuk-input-border-colour;
+  box-sizing: border-box;
+  cursor: default; // emphasise non-editable status of prefixes and suffixes
+  display: inline-block;
+  flex: 0 0 0;
+  border-right: 0;
+  padding-left: govuk-spacing(3);
```
alex-ju, line 44: @simonwhatley correct me if I'm wrong, but as I understand the expected outcome from a design perspective would require consistent padding (`govuk-spacing(1);`), as before, but limited to a minimum width equal to the height (which is 40px, or the equivalent value in rems).

This can be achieved with
```
  padding: govuk-spacing(1);
  min-width: 40px;
      min-width: govuk-px-to-rem(40px);
  }
```

simonwhatley: Yes, that's my expectation @alex-ju. 👍

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
GDSNewt: Ahhhhhh this is what i couldn't get my head around the first time and why I used modifiers in the initial PR. Will make the change.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
GDSNewt: I went with the weird padding arrangement over min-width because while it worked fine for the prefix example I couldn't seem to get it to work for the suffix. I tried having both a width setting of auto and a min-width setting of 40px, but from reading the css docs it seems that min-width always overrides width anyway. Please let me know if i'm missing something obvious. 

Screenshot 2020-05-21 at 09 33 28
Screenshot 2020-05-21 at 09 33 32

GDSNewt: It seems the above problem was being caused by the flex: 0 0 0; on line 42. It was not allowing the box to grow with the text, instead it was imposing a width limit of 40px.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -31,7 +31,7 @@
   }
 }
 
-.gem-c-input__suffix {
+.gem-c-input__prefix, .gem-c-input__suffix {
```
simonwhatley, line 34: @GDSNewt You don't need `border-left: 0;`for both prefix and suffix. It's only for the `.gem-c-input__suffix`.

It's also a good idea to but the classes in separate lines for readibility:

```
.gem-c-input__prefix,
.gem-c-input__suffix {

}
```

PR #1526, opened 2020-05-22
280 Add breadcrumbs from parent for html documents

reggieb:
## What
- Refactor breadcrumbs code
- Get breadcrumbs from parent if item is an html publication
- Ensure breadcrumbs from parent include superbreadcrumbs

## Why
- A refactor was needed to provide a place to select the source of the breadcrumbs (item or item's parent). That was made easier by moving logic out of the contextual_breadcrumbs partial, and into a new object `GovukPublishingComponents::Presenters::BreadcrumbSelector`
- The refactor also simplified some of the code by removing the needs to call `[:breadcrumbs]` on a group of breadcrumbs to get the breadcrumbs.
- Html publications do not contain data from which complete breadcrumbs can be built, but the parent document does. So for these documents the parent data needs to be retrieved and that used to build the breadcrumbs. 

### Html document that should have a supercrumb
html_document_superbreadcrumbs_before_change

### Html document after change showing supercrumb
html_document_superbreadcrumbs

### Html document without supercrumb before change
html_document_breadcrumbs_before_change

### Html document without supercrumb showing breadcrumbs from parent
html_document_breadcrumbs


lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -0,0 +1,105 @@
...
+        return content_item_options unless content_item_navigation.html_document_with_parent?
+        return parent_item_options if parent_item_navigation.priority_breadcrumbs
+
+        {
+          step_by_step: parent_item_options[:step_by_step],
+          breadcrumbs: parent_breadcrumbs,
+        }
+      end
+
+      def options(navigation)
```
hannako, line 58: Minor thing, but of the 8 branches in this method, 4 of them return the same thing: 
```
 {
    step_by_step: false,
    breadcrumbs: navigation.breadcrumbs,
  }
```

Maybe that is the default `option`?

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -0,0 +1,105 @@
...
+          {
+            step_by_step: false,
+            breadcrumbs: navigation.breadcrumbs,
+          }
+        elsif navigation.content_has_curated_related_items?
+          {
+            step_by_step: false,
+            breadcrumbs: navigation.breadcrumbs,
+          }
+        elsif navigation.content_is_tagged_to_a_live_taxon? && !navigation.content_is_a_specialist_document?
```
hannako, line 89: Did you want to call `use_taxon_breadcrumbs?` here?

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -0,0 +1,105 @@
+module GovukPublishingComponents
+  module Presenters
+    class BreadcrumbSelector
+      def self.call(*args)
+        new(*args).output
+      end
+
```
hannako, line 7: A question rather than a comment: I would have made everything in this presenter private, except the `call` method. Is there no point doing that in this case?

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
reggieb: That mirrors the previous implementation. I didn't want to mess with the logic - just move it from a view partial into an object. Also, as the core of the logic is within the `options` method, I thought it clearer, and easier to modify if all the logical endpoints were of the same format.

lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -16,32 +16,41 @@ def simple_smart_answer?
...
+            url: "/",
+          },
+          {
+            title: parent_finder["title"],
+            url: parent_finder["base_path"],
+          },
+        ]
+      end
+
+      def use_taxon_breadcrumbs?
```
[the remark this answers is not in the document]
reggieb: A bit of tidying up that it looks like I've started and haven't completed. It's now used in `BreadcrumbSelector`.

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
reggieb: Exactly

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
reggieb: I tend to use `private` fairly sparingly. It's not particular meaningful in Ruby as it's very easy to bypass (`send(:private_method_name)`). I also think the greatest complement is for someone to reuse code I've written in a way I hadn't imagined. So, I tend to assume that as I don't know how someone else might want to use the object I've built, then why should I guess which methods might or might not be useful to them. Also leaving them public makes debugging easier.

Therefore, I tend only to use `private` where using the method would break things or has security implication. For example, in controllers for methods that aren't actions and so could be exposed as actions due to a routing error if they weren't private.

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
hannako: That makes sense.

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
hannako: 👍

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
hannako: Thanks

yarn.lock
```diff
@@ -0,0 +1,2260 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
alex-ju, line 1: This project is using `npm` – this file should not be committed.

reggieb: Very good point. Thank you

app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
```diff
@@ -1,47 +1,16 @@
...
-    <%= render 'govuk_publishing_components/components/breadcrumbs', breadcrumbs: navigation.breadcrumbs, inverse: inverse, collapse_on_mobile: collapse_on_mobile %>
-  <% elsif navigation.content_is_tagged_to_a_live_taxon? && prioritise_taxon_breadcrumbs %>
-    <%# Rendering taxonomy breadcrumbs because the page is tagged to live taxons
-          and we want to prioritise them over all other breadcrumbs %>
+               breadcrumb_selector[:breadcrumbs] %>
+  <% elsif breadcrumb_selector[:breadcrumbs] %>
     <%= render 'govuk_publishing_components/components/breadcrumbs',
-               breadcrumbs: navigation.taxon_breadcrumbs[:breadcrumbs],
+               breadcrumbs: breadcrumb_selector[:breadcrumbs],
                inverse: inverse,
```
DilwoarH, line 13: How does this get set for when we want it to be true?

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@
   of the commit log.
 
 ## Unreleased
+* Add breadcrumbs from parent for html documents ([PR #1526] (https://github.com/alphagov/govuk_publishing_components/pull/1526)
```
andysellick, line 11: Can this match the PR title please? (you can amend the PR title to match this if you like)

app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
reggieb: This is existing code. I believe it is set when the partial is called so to set it to inverse true it would be called via:
```ruby
<%= render 'path/to/contextual_breadcrumbs', .... , inverse: true %>
```

CHANGELOG.md
reggieb: I've updated the PR title as I think that's a better description of the change. Originally, the ticket was just to add superbreadcrumbs to html documents, but on investigation it was realised that all html documents needed to pull the breadcrumbs from their parent (including superbreadcrumbs).

PR #1527, opened 2020-05-22
Undo recursive FAQPage presenter

sihugh:
## What

Removes the recursive bit of the FAQSchema presenter which is used to extract questions and answers from the HTML publication format.

## Why

I'm intending to do this more simply and specifically for HTML publications within government-frontend. This code is quite hard to debug without context, and it's not working quite right at the moment.

Note that this is not a breaking change - it still resolves a valid schema on HTML publications, but they are a bit less useful.

This reverts commit f0e470a27660e1d0bcfbd318e4957ba983191095.

### DIFF
--- a/lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb
+++ b/lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb
@@ -57,42 +57,22 @@ module GovukPublishingComponents
 
         question = page.title
 
+        # rubocop:disable Style/IfInsideElse
         doc.xpath("html/body").children.each_with_object({}) do |element, q_and_as|
-          _q_and_as, question = recursive_question_and_answers(element, question, q_and_as)
-        end
-      end
-
-      def recursive_question_and_answers(element, question, q_and_as)
-        if is_a_question?(element)
-          question = element.text
-          q_and_as[question] = { anchor: element["id"] }
-        else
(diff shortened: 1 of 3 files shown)


lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb
```diff
@@ -57,42 +57,22 @@ def question_and_answers(html)
...
-          if question_hash_is_unset?(q_and_as, question)
-            q_and_as[question] = { answer: element.to_s }
-          elsif answer_is_unset?(q_and_as, question)
-            q_and_as[question][:answer] = element.to_s
+          if question_element?(element)
+            question = element.text
+            q_and_as[question] = { anchor: element["id"] }
           else
-            q_and_as[question][:answer] << element.to_s
+            if question_hash_is_unset?(q_and_as, question)
```
kevindew, line 66: Sorry - I see this is only part of the revert but I wonder if we can switch this to elsif to avoid the need to add the linter enabling/disabling?

sihugh: Yeah, I looked at that too - will fix 👍

sihugh: Updated 👍

PR #1531, opened 2020-05-23
Bump to 21.52.0

sihugh:


sihugh:
> Do you need to delete `## Unreleased`?

I figure it saves someone else typing it next time

PR #1533, opened 2020-05-26
Modify Step nav header component to support custom tracking

DilwoarH:
We don't currently track users who click on the superbread crumbs (on the business & education hubs). As this feature is going to be rolled out with the new taxonomy it would be good to add tracking to understand if it is helping users navigate back through the pages.

Event category: breadcrumbClicked
Event action: superBreadcrumb
Event label: page path link (to hub page)

Its done when: When a users clicks on a super bread crumb an event is triggered in GA and we can track the page they were on and the page they went to.

https://trello.com/c/G2iuimyb/283-add-analytics-tracking-to-super-breadcrumb

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
@@ -26,3 +26,20 @@ examples:
       title: 'With a tracking id'
       path: '#'
       tracking_id: 'this-is-the-tracking-id'
+  with_custom_tracking:
+    data:
+      title: 'With a custom tracking'
+      path: '#'
+      tracking_category: "customTrackingCategoryClicked"
+      tracking_action: "customTrackingAction"
+      tracking_label: "customTrackingLabel"
+      tracking_dimension: "customTrackingDimension"
+      tracking_dimension_index: "23"
+  without_custom_dimension:
+    data:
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
```diff
@@ -26,3 +26,22 @@ examples:
...
+      tracking_dimension_index: "23"
+  without_custom_dimension:
+    data:
+      title: 'With a tracking id'
+      path: '#'
+      tracking_id: 'this-is-the-tracking-id'
+      tracking_category: "customTrackingCategoryClicked"
+      tracking_action: "customTrackingAction"
+      tracking_label: "customTrackingLabel"
+      tracking_dimension: ""
```
DilwoarH, line 47: @andysellick @vanitabarrett - not really happy that I have to send an empty string. Is there a way around this?

Had to do this because of this https://github.com/alphagov/govuk_publishing_components/pull/1533/files#diff-b75e540e4545c561ea33a018910a7bf1R14

vanitabarrett: That code looks like it should use the `title` by default if a `tracking_dimension` isn't passed?

DilwoarH:
Added some minor changes that I noticed, no change to logic.

PR #1538, opened 2020-05-27
Remove jQuery from modules system

alex-ju:
## What
Remove jQuery from the modules system

## Why
- there are projects that require jQuery only because the module system is based on it (such as Content Publisher)
- this will allow us to start removing jQuery gradually from GOV.UK

## Notes
This is a breaking change because all scripts using the GOV.UK Publishing and the legacy GOV.UK Frontend Toolkit pattern have an initialisation pattern that expects the argument to be a jQuery object. This includes scripts in `collections`, `finder-frontend`, `frontend`, `service-manual-frontend`, `whitehall`, `static` for the public-facing applications and most of the admin applications.

The upgrade process is fairly simple:
```
  this.start = function ($element) {
    console.log($element)
  }
}
```
will become
```diff
GOVUK.Modules.TestModule = function () {
  this.start = function (element) {
+   $element = $(element)
    console.log($element)
  }
}
```

## Visual Changes
No visual changes

[Trello card](https://trello.com/c/sghHecNi/91-remove-jquery-dependency)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -1,22 +1,23 @@
 ;(function (global) {
   'use strict'
 
-  var $ = global.jQuery
   var GOVUK = global.GOVUK || {}
   GOVUK.Modules = GOVUK.Modules || {}
 
   GOVUK.modules = {
     find: function (container) {
-      container = container || $('body')
+      container = container || document.querySelector('body')
+
+      if (container.jquery) container = container[0]
 
       var modules
       var moduleSelector = '[data-module]'
 
-      modules = container.find(moduleSelector)
+      modules = Array.prototype.slice.call(container.querySelectorAll(moduleSelector))
 
       // Container could be a module too
-      if (container.is(moduleSelector)) {
-        modules = modules.add(container)
+      if (container.getAttribute('data-module')) {
+        modules.push(container)
       }
 
       return modules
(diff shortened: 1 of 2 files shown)


kevindew:
Thanks for raising Alex 👍 From my comment on https://trello.com/c/sghHecNi/91-remove-jquery-dependency#comment-5ecea254d07a507c26a77d35 this seems to be the short, backwards incompatible shock approach? is that right?

Where I see the bc pains is code such as https://github.com/alphagov/content-publisher/blob/86235fa1639e944cd02d63e543983d5dfd919b83/app/assets/javascripts/components/autocomplete.js#L8-L9 where we unwrap the jQuery object no longer working (and presumably there's a whole bunch in other places that expect it wrapped)

alex-ju:
> Thanks for raising Alex 👍 From my comment on https://trello.com/c/sghHecNi/91-remove-jquery-dependency#comment-5ecea254d07a507c26a77d35 this seems to be the short, backwards incompatible shock approach? is that right?
> 
> Where I see the bc pains is code such as https://github.com/alphagov/content-publisher/blob/86235fa1639e944cd02d63e543983d5dfd919b83/app/assets/javascripts/components/autocomplete.js#L8-L9 where we unwrap the jQuery object no longer working (and presumably there's a whole bunch in other places that expect it wrapped)

This should be backward-compatible, as it works for both jQuery wrapped elements and basic nodeElements. Will test with a few apps before marking it as ready for review.

PR #1541, opened 2020-05-28
Find components within components

andysellick:
## What
- the suggested sass code was looking to see if any components contained other components, but stopped at that first level
- the feedback component contains several partials (identified and correctly ignored by the original code) but they include other components
- this change keeps burrowing down through partials/subcomponents until it can't find anything else
- updated the dummy app to include the feedback component so that this can be tested, which works because the dummy app doesn't include button but feedback does
- however I'm not sure how to do a concrete test for this i.e. if button was already in the dummy app and this code didn't work properly, the test would still pass

## Why
Noticed in the upgrade to individual components in `service-manual-frontend` that it hadn't included the input component, because it was 'hidden' in the feedback components partials

## Visual Changes
None.

### DIFF
--- a/app/controllers/govuk_publishing_components/component_guide_controller.rb
+++ b/app/controllers/govuk_publishing_components/component_guide_controller.rb
@@ -47,22 +47,7 @@ module GovukPublishingComponents
       additional_files = "@import 'govuk_publishing_components/govuk_frontend_support';\n"
       additional_files << "@import 'govuk_publishing_components/component_support';\n" unless print_styles
 
-      components = components_in_use
-      components_to_search = components
-      loop = true
-
-      while loop
-        extra_components = find_partials_in(components_to_search)
-
-        if extra_components.any?
-          components << extra_components
-          components_to_search = extra_components
-        else
-          loop = false
-        end
-      end
-
-      components = components.flatten.uniq.sort
+      components = find_all_partials_in(components_in_use)
 
       components.map { |component|
         "@import 'govuk_publishing_components/components/#{print_path}#{component.gsub('_', '-')}';" if component_has_sass_file(component.gsub("_", "-"), print_styles)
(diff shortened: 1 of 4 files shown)


spec/dummy/app/views/layouts/dummy_admin_layout.html.erb
```diff
@@ -18,5 +18,6 @@
       <%= yield %>
     </main>
   </div>
+  <%= render "govuk_publishing_components/components/feedback" %>
```
alex-ju, line 21: I'm not sure we want this in the `dummy_admin_layout.html.erb` (as we won't have it in an admin app), maybe the dummy (public-facing) `application.html.erb` layout would be more suitable.

app/controllers/govuk_publishing_components/component_guide_controller.rb
```diff
@@ -48,14 +48,20 @@ def components_in_use_sass(print_styles)
...
       components = components_in_use
-      extra_components = []
-
-      components.each do |component|
-        components_in_component = components_within_component(component)
-        extra_components << components_in_component
+      components_to_search = components
+      loop = true
+
+      while loop
```
alex-ju: This looks more like a depth-first iterative search than a recursive one, so we might want to update the PR title and function name in the next commit.

spec/dummy/app/views/layouts/dummy_admin_layout.html.erb
andysellick: Okay, thanks, have changed.

app/controllers/govuk_publishing_components/component_guide_controller.rb
andysellick: Have updated, okay to re-review?

andysellick:
Thanks @alex-ju. Do you think this should be added to the changelog, or just let it be folded into the next release?

alex-ju:
@andysellick, it may be valuable to add it as it can translate in updated suggestions for some apps. thanks for working on this!

alex-ju:
This turned out to be a breaking change (crashing on `/component-guide`) for `content-publisher` (and potentially other apps) when [rendering an asset from govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/aa86284e7081788b6d22c50cfb70ef3992aa7dca/app/views/govuk_publishing_components/components/_attachment.html.erb#L43). Kevin will raise an issue for us to look into it.

andysellick:
Oh no! Sorry @alex-ju - what can I do to help?

alex-ju:
Just to be aware, it's not affecting users in any way. I'll look into it and will ping you for a review.

PR #1543, opened 2020-05-29
304 Fix FAQ Schemas for html publications

reggieb:
## What
Adds a version of the FAQ schema (for generating structured data) that is specifically for use with html publications.

## Why
Html publications are not parsed correctly by the current FAQ schema. They have a simple structure generated from markup that add heading and paragraphs at a single level.

## How
The new schema identifies the first (from h1 to h5) level of heading where there are more than one, and uses that level headings as the questions and the content from that heading to the next heading of the same level (or the text end) as the answer.

See `spec/lib/govuk_publishing_components/presenters/machine_readable/html_publication_schema_spec.rb` for examples of how the questions are split.

## Structured Data Changes
Currently the [Coronavirus outbreak FAQs](https://www.gov.uk/government/publications/coronavirus-outbreak-faqs-what-you-can-and-cant-do/coronavirus-outbreak-faqs-what-you-can-and-cant-do) is of FAQPage type but consists of a single question and answer. The answer containing the whole of the `.govspeak` div.

304-structured-data-before-change

This PR alters this behaviour. Each `h2` within the `.govspeak` div is set as a question, and the answer is given as the text between that heading and the next.

304-structured-data-after-change


lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -16,7 +16,7 @@ def simple_smart_answer?
         content_item["document_type"] == "simple_smart_answer"
       end
 
-      def html_document_with_parent?
+      def html_publication_with_parent?
```
reggieb, line 19: This is a correction of naming spotted while working on this code.

lib/govuk_publishing_components/presenters/machine_readable/faq_page_schema.rb
```diff
@@ -24,9 +24,8 @@ def main_entity
       end
 
       def questions_and_answers_markup
-        question_and_answers(page.body)
-          .select { |_, value| value[:answer].present? }
-          .map do |question, value|
+        question_and_answers.select { |_, value| value[:answer].present? }
+                            .map do |question, value|
```
reggieb, line 28: Changed the spacing here so that it was easier to see where the chained calls ended and block code began.

PR #1544, opened 2020-06-01
Add margin bottom option to step nav header component

andysellick:
## What
Uses the shared helper to add a `margin_bottom` option to the step by step navigation header component.

## Why
We need it for [this page](https://www.gov.uk/government/publications/coronavirus-outbreak-faqs-what-you-can-and-cant-do/coronavirus-outbreak-faqs-what-you-can-and-cant-do) where it's colliding with the organisation logo.

Trello card: https://trello.com/c/dY8wHgpu/280-adding-breadcrumbs-and-superbreadcrumbs-to-all-html-publications

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
@@ -18,13 +18,17 @@
   if tracking_id
     tracking_options ||= ({ dimension96: tracking_id }).to_json
   end
+
+  local_assigns[:margin_bottom] ||= 0
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  classes = "gem-c-step-nav-header #{shared_helper.get_margin_bottom}"
 %>
 <% if title %>
   <script type="application/ld+json">
     <%= raw breadcrumb_presenter.structured_data.to_json %>
   </script>
 
-  <div class="gem-c-step-nav-header" data-module="track-click">
+  <div class="<%= classes %>" data-module="track-click">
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
```diff
@@ -18,13 +20,16 @@
   if tracking_id
     tracking_options ||= ({ dimension96: tracking_id }).to_json
   end
+
+  classes = %w(gem-c-step-nav-header)
+  classes << (shared_helper.get_margin_bottom)
```
reggieb, line 25: As both classes are single strings, it would be simpler just to build the string directly. So just:
```
classes = "gem-c-step-nav-header #{shared_helper.get_margin_bottom}"
```

andysellick: Thanks, good point, always happy to save a line of code.

app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
```diff
@@ -18,13 +18,17 @@
   if tracking_id
     tracking_options ||= ({ dimension96: tracking_id }).to_json
   end
+
+  local_assigns[:margin_bottom] ||= 0
```
DilwoarH, line 22: is there a reason why you use local_assigns rather than `margin_bottom ||= 0`?

andysellick: No. Good point, that would probably work just fine.

PR #1547, opened 2020-06-01
Remove unnecessary margin around expanded feedback component

andysellick:
**Don't merge this until https://github.com/alphagov/frontend/pull/2387 is merged**

## What
Remove the spacing around the form in the feedback component when the component is expanded.

## Why
The spacing is largely unnecessary and appears to be in place for one edge case only, the homepage, which has been [fixed in a separate PR](https://github.com/alphagov/frontend/pull/2387).

## Visual Changes

Here's the homepage at the moment when feedback is expanded:

Screenshot 2020-06-01 at 12 13 58

Here's the fix:

Screenshot 2020-06-01 at 12 44 02

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Remove unnecessary margin around expanded feedback component ([PR #1547](https://github.com/alphagov/govuk_publishing_components/pull/1547))
+
 ## 21.55.3
 
 * Use video title as eventLabel for YT video event tracking ([PR #1562](https://github.com/alphagov/govuk_publishing_components/pull/1562))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -182,7 +182,6 @@
 }
 
 .gem-c-feedback__form {
-  margin: govuk-spacing(3) govuk-spacing(2) 0 govuk-spacing(2);
   padding: govuk-spacing(3) 0;
   border-top: govuk-spacing(2) solid govuk-colour("blue");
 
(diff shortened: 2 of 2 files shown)


andysellick:
Thanks @alex-ju - I've already raised the change I think you're talking about, see https://github.com/alphagov/frontend/pull/2387

alex-ju:
> Thanks @alex-ju - I've already raised the change I think you're talking about, see [alphagov/frontend#2387](https://github.com/alphagov/frontend/pull/2387)

Yes, saw that. It certainly works well in `frontend` as shown in your PR, I'm wondering if it will suit other frontend applications too.

andysellick:
@alex-ju good shout. I've just tested it in `collections`, `government-frontend` and `finder-frontend` - and unfortunately it breaks in `finder-frontend`. I'll put this on hold for the moment until I can do some more testing and be sure this change won't break things.

Collections

Before | After 
------ | ------
Screenshot 2020-06-09 at 15 54 45 | Screenshot 2020-06-09 at 15 54 54

Government-frontend

Before | After 
------ | ------
Screenshot 2020-06-09 at 16 00 13 | Screenshot 2020-06-09 at 16 00 24

Finder-frontend

Before | After 
------ | ------
Screenshot 2020-06-09 at 16 06 00 | Screenshot 2020-06-09 at 16 06 29

Interestingly `finder-frontend`'s layout is clearly a little bit broken already - the feedback component on mobile is against the edge of the screen, unlike elsewhere:

Screenshot 2020-06-09 at 16 05 37

alex-ju:
> @alex-ju good shout. I've just tested it in `collections`, `government-frontend` and `finder-frontend` - and unfortunately it breaks in `finder-frontend`. I'll put this on hold for the moment until I can do some more testing and be sure this change won't break things.

I imagine that's because `finder-frontend` is the only one from the above list using the `header_footer_only` layout from `static` (`smart-answers` and `whitehall` are using `header_footer_only` too).

andysellick:
Just tested it on `smart-answers` and it seems fine, the layout is a bit cleaner:

Screenshot 2020-06-11 at 09 14 13

andysellick:
Pretty sure whitehall will be okay as well... everything is wrapped inside the `#wrapper` element, which has a max-width and margin auto, so shaving off the margin and removing the max width of the component shouldn't have any effect. This is me simulating this on live:

Screenshot 2020-06-11 at 09 40 40

PR #1552, opened 2020-06-02
Use collapsible breadcrumbs from govuk-frontend

alex-ju:
## What
The code for collapsible breadcrumbs is now [in `govuk-frontend`](https://github.com/alphagov/govuk-frontend/pull/1754) and [available to this gem](https://github.com/alphagov/govuk_publishing_components/pull/1549) so we use the modifier from `govuk-frontend`.

## Why
Keep codebase clean.

## Visual Changes
No visual changes.


app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -20,38 +20,3 @@
...
-  @include govuk-media-query($until: tablet) {
-    .govuk-breadcrumbs__list {
-      display: flex;
-    }
-
-    .govuk-breadcrumbs__list-item:not(:last-child):not(:first-child) {
-      display: none;
-    }
-
-    .govuk-breadcrumbs__list-item {
```
vanitabarrett, line 34: The govuk-frontend change doesn't include the touch target change as there were some questions about that. It might be worth leaving that in if we can, until we get an approach sorted out with the design system?

alex-ju: Good catch. I kept the definitions that cover touch target change and tested the examples again.

vanitabarrett:
The increased touch target doesn't seem to work when there are long breadcrumbs which wrap. I think this is a bug with the current implementation though, so happy for this to be merged and we can create a separate issue for it.

Screenshot 2020-06-02 at 15 13 34

alex-ju:
@vanitabarrett. Ha, interesting one. No urgency to get this in, as users are not gaining much with this change – more of a way to keep code to a minimum. I'd leave it open, maybe we'll have time to look at this in the near future.

alex-ju:
> The increased touch target doesn't seem to work when there are long breadcrumbs which wrap. I think this is a bug with the current implementation though, so happy for this to be merged and we can create a separate issue for it.
> 
> Screenshot 2020-06-02 at 15 13 34

I removed the relative positioning from the `<a>` element (now it goes one level to the `<li>`) when collapsing is on to address the bug flagged above and did a cross-browser check.

Screenshot 2020-06-23 at 17 43 54

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
```diff
@@ -48,10 +40,10 @@
     .govuk-breadcrumbs__link:after {
```
andysellick, line 40: Related - I think it's possible to remove line 32: `margin: 0` on `.govuk-breadcrumbs__list-item:before`. Testing it locally and it's already set.

vanitabarrett:
@andysellick The original intention of the change was to make it the default, but this meant it became a breaking change for people working on other apps (like Content Publisher) where the breadcrumbs are being used in slightly different ways. There's some context in https://github.com/alphagov/govuk_publishing_components/pull/1324 and https://github.com/alphagov/govuk_publishing_components/pull/1330 . Just raising as a heads up as I think we'd have to tread carefully if making it the default again.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
alex-ju: Indeed, L32 is not needed. Regarding the other comment, when rebasing I re-added the `position: relative` mentioned in my comment above 🤦 . Removed it again – it should look as in my screen capture again.

PR #1556, opened 2020-06-04
339 - Reorder the breadcrumb so superbreadcrumb not top option

reggieb:
## What
Reverse order of top three options in breadcrumb selector, so that content_tagged_to_a_finder comes first, step by step second, priority 3rd

## Why
Superbreadcrumb (priority option) is not always the correct option. 

https://trello.com/c/hxOu0Mc9/339-reorder-the-breadcrumb-so-superbreadcrumb-not-top-option

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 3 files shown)


reggieb:
Changelog added.

PR #1561, opened 2020-06-08
Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead

### LINKED ISSUE #1770: Print link component visible in the printed document
Screenshot_20201107-212111

[New National Restrictions from 5 November - GOV.UK.pdf](https://github.com/alphagov/govuk_publishing_components/files/5505454/New.National.Restrictions.from.5.November.-.GOV.UK.pdf)

danacotoran:
## What

Where possible and practicable, remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead.

## Why

Many components are hidden in the printable version due to being interactive or simply irrelevant to a user printing a page.
This is currently done in separate print stylesheets.

Thanks to the adoption of a dedicated `govuk-!-display-none-print` class in govuk-frontend, we can remove `display:none` from many components' print stylesheets and use the special class instead.

The next step would be to remove all references for now empty component stylesheets from the GOVUK apps.

## Visual Changes

No visual changes should occur as a result. 

Fixes #1770
Also, closes #2063 

https://trello.com/c/EHMAHSjh

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead. ([PR #1561](https://github.com/alphagov/govuk_publishing_components/pull/1561))
+
 ## 24.10.3
 
 * Remove phase banner restrictions ([PR #2057](https://github.com/alphagov/govuk_publishing_components/pull/2057))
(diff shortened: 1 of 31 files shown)


alex-ju (approved):
Excellent work on this! Would be great if we can create an issue to remember our future selves to remove the files in our next major/breaking release.

alex-ju:
Some would look for this PR title to match the changelog entry – feels like a low-effort thing to do

PR #1565, opened 2020-06-09
341 Replace bodged parent breadcrumbs with specialist topic breadcrumbs

reggieb:
## What
Replace bodged parent breadcrumbs with specialist topic breadcrumbs

Also some refactoring:
- Add a topic specific breadcrumb presenter `ContentBreadcrumbsBasedOnTopic`
- Reorder breadcrumb selector to put main methods at the top
- Make all breadcrumb selector respond to call (so calling object doesn't need to know inner structure)
- Remove legacy as a label as it's not useful
- Rename `ContentBreadcrumbsBasedOnParent` to `ContentBreadcrumbsBasedOnAncestors` to clarify when working directly with parent and when with a group of parents in context navigation.
- Removed breadcrumb outer hash as redundant.
- Separated finder breadcrumbs from default breadcrumbs to simplify logic

## Why
The contextual breadcrumbs has what seems like a bodge to allow overriding taxon breadcrumbs with breadcrumbs based on the parent link, but only if there's some related links set.

We want to look into replacing these with breadcrumbs based on specialist topics (these use the topics link type).


spec/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_parent_spec.rb
```diff
@@ -1,6 +1,6 @@
 require "spec_helper"
```
sihugh, line 2: We might want to rename this file to `...based_on_ancestors_spec.rb` to match the presenter

lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_topic.rb
```diff
@@ -0,0 +1,39 @@
...
+      def initialize(content_item)
+        @content_item = content_item
+      end
+
+      def breadcrumbs
+        breadcrumbs = [{ title: "Home", url: "/" }]
+        return breadcrumbs unless topics.present?
+
+        breadcrumbs << topic_breadcrumb
+        breadcrumbs.reverse!
```
sihugh, line 19: I don't think we want to reverse these ones - I've just checked out `/guidance/bond-insurance-policy` locally using this branch with government-frontend and it looks like `Export Finance > Home` when it should be the other way round.

I think we reverse them in other scenarios where we construct them backwards

sihugh: Actually, [that doc I chose](https://www.gov.uk/guidance/bond-insurance-policy) is interesting because it's got topics as it's breadcrumbs, but constructs the hierarchy for the nav by using linked parents. 🤔

PR #1572, opened 2020-06-12
369 Show breadcrumbs alongside superbreadcrumbs everywhere they appear

reggieb:
## What
We want to show breadcrumbs alongside superbreadcrumbs where we're using them. Superbreadcrumbs should appear below breadcrumbs. This is how they worked when originally implemented on step by steps.

Also refactor of `BreadcrumbSelector` to return an instance with public methods rather than a hash.

## Why
We want breadcrumbs to be a consistent pattern on every page of the site. But keep the benefits of the superbreadcrumbs on coronavirus content and step by steps.

### Before change

breadcrumbs_and_superbreadcrumb_before_change

### After change

breadcrumbs_and_superbreadcrumb_version_2

### Same in mobile view

breadcrumbs_and_superbreadcrumb_version_2_mobile

### DIFF
--- a/lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
+++ b/lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
@@ -1,10 +1,6 @@
 module GovukPublishingComponents
   module Presenters
     class BreadcrumbSelector
-      def self.call(*args)
-        new(*args).output
-      end
-
       attr_reader :content_item, :request, :prioritise_taxon_breadcrumbs
 
       def initialize(content_item, request, prioritise_taxon_breadcrumbs)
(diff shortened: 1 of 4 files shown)


lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -45,9 +57,8 @@ def parent_breadcrumbs
         }
       end
 
-      def output
+      def best_match_option
```
DilwoarH, line 60: Could this just be called `options` rather than `best match options`?

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -68,11 +79,6 @@ def options(navigation)
             step_by_step: true,
```
DilwoarH, line 79: Also, do we need this anymore if we're using `breadcrumb_selector.step_by_step`

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
reggieb: There is already an instance method called `options` which contains the collection of options. The purpose of this method is to determine which of those options should be used (and with which set of navigation data) in the current context. Hence the name.

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
reggieb: If you look at the `step_by_step` method, it returns the value matching `:step_by_step` within the options hash. So yes - this key is needed.

PR #1574, opened 2020-06-15
Update summary list component to allow delete action at the group level and custom heading levels

alex-ju:
## What
Update the summary-list component with the following to support the following features:
 - allow delete action at group level
 - render delete actions as destructive links (in red)
 - allow custom heading level and size on section titles

## Why
Required by design for the [business-volunteer-form](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form)

## Visual Changes
<table>
<tr><td valign="top">

Screenshot 2020-06-15 at 16 55 47

</td><td valign="top">

Screenshot 2020-06-15 at 16 55 35

</td></tr>
<tr><td valign="top">

Screenshot 2020-06-15 at 16 55 16

</td><td valign="top">

Screenshot 2020-06-15 at 16 55 06

</td></tr>
</table>

[Trello card](https://trello.com/c/Bszitcav/601-update-summary-list-component-to-allow-delete-action-at-the-group-level)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_summary-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_summary-list.scss
@@ -1,7 +1,6 @@
 @import "govuk/components/summary-list/summary-list";
 
 .gem-c-summary-list {
-  position: relative;
   border-bottom: 1px solid $govuk-border-colour;
 
   @include govuk-font(19);
@@ -14,12 +13,26 @@
   &:nth-last-of-type(1) {
     border-bottom: 0;
   }
+
+  .govuk-summary-list {
+    clear: both;
+  }
 }
 
-.gem-c-summary-list__edit-section-link {
-  position: absolute;
-  top: 0;
-  right: 0;
+.gem-c-summary-list__group-title {
+  @include govuk-media-query($from: tablet) {
+    display: inline-block;
+    margin-right: govuk-spacing(3);
+  }
+}
+
+.gem-c-summary-list__group-actions-list {
+  margin-bottom: govuk-spacing(3);
+
+  @include govuk-media-query($from: tablet) {
+    float: right;
+    width: auto;
+  }
 }
 
 .gem-c-summary__block {
--- a/app/views/govuk_publishing_components/components/_summary_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_summary_list.html.erb
@@ -3,21 +3,36 @@
   title ||= nil
   borderless ||= false
   edit ||= {}
+  delete ||= {}
   items ||= []
   block ||= yield
 %>
 <% if title || items.any? %>
   <%= tag.div class: "gem-c-summary-list #{"govuk-summary-list--no-border" if borderless}", id: id do %>
     <% if title %>
-      <%= tag.h3 title, class: "govuk-heading-m" %>
-      <% if edit.any? %>
-        <%= tag.ul class: "govuk-summary-list__actions-list" do %>
-          <%= tag.li class: "govuk-summary-list__actions-list-item" do %>
-            <%= link_to edit.fetch(:href),
-                      class: "govuk-link gem-c-summary-list__edit-section-link",
-                      title: "#{edit.fetch(:link_text, "Change")} #{title}",
-                      data: edit.fetch(:data_attributes, {}) do %>
-               <%= edit.fetch(:link_text, "Change") %> <%= tag.span title, class: "govuk-visually-hidden" %>
+      <%= tag.h3 title, class: "govuk-heading-m gem-c-summary-list__group-title" %>
+      <% if edit.any? || delete.any? %>
+        <%= tag.ul class: "govuk-summary-list__actions-list gem-c-summary-list__group-actions-list" do %>
+          <%- if edit.any? %>
+            <% edit_section_link_text = edit[:link_text] || "Change" %>
+            <%= tag.li class: "govuk-summary-list__actions-list-item" do -%>
+              <%= link_to edit.fetch(:href),
+                        class: "govuk-link",
+                        title: "#{edit_section_link_text} #{title}",
+                        data: edit.fetch(:data_attributes, {}) do %>
+                <%= edit_section_link_text %><%= tag.span " #{title}", class: "govuk-visually-hidden" -%>
+              <% end %>
+            <% end %>
+          <% end %>
+          <% if delete.any? %>
+            <% delete_section_link_text = delete[:link_text] || "Delete" %>
+            <%= tag.li class: "govuk-summary-list__actions-list-item" do -%>
+              <%= link_to delete.fetch(:href),
+                        class: "govuk-link",
+                        title: "#{delete_section_link_text} #{title}",
+                        data: delete.fetch(:data_attributes, {}) do %>
+                <%= delete_section_link_text %><%= tag.span " #{title}", class: "govuk-visually-hidden" -%>
+              <% end %>
             <% end %>
           <% end %>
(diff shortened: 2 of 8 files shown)


maxgds:
Quick point - I think marking destructive links as red and not differentiating them otherwise breaks WCAG guidelines: "Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element." https://www.w3.org/WAI/WCAG21/quickref/

andysellick:
> Quick point - I think marking destructive links as red and not differentiating them otherwise breaks WCAG guidelines: "Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element." https://www.w3.org/WAI/WCAG21/quickref/

Isn't the text "Delete" also conveying information?

maxgds:
> > Quick point - I think marking destructive links as red and not differentiating them otherwise breaks WCAG guidelines: "Color is not used as the only visual means of conveying information, indicating an action, prompting a response, or distinguishing a visual element." https://www.w3.org/WAI/WCAG21/quickref/
> 
> Isn't the text "Delete" also conveying information?

Well, I guess so, but if that conveys all the necessary info then why colour it red? If it needs to be enhanced to convey "watch out, this is dangerous!" I think "watch out this is dangerous!" needs to be perceivable by people who can't see that colour.

andysellick:
> Well, I guess so, but if that conveys all the necessary info then why colour it red? If it needs to be enhanced to convey "watch out, this is dangerous!" I think "watch out this is dangerous!" needs to be perceivable by people who can't see that colour.

I've often wondered about this as a general concept. If you assume the text is enough to reach the required level of accessibility, is then okay to make it slightly more obvious for those who can perceive colour? (assuming that colour doesn't make the experience worse for another user). 

We use colour to help sighted users because they can see many page elements at once, and we want to help them distinguish between them. A screen reader user can only perceive one element at a time, so they don't need elements to stand out, relying instead on things like semantic markup. And actually this works the other way round - a screen reader user will be told that a list is a list, but a sighted user has visual cues like bullets to convey the same meaning. You have to make sure that everyone understands the page they're using, but you do it in different ways for different users.

maxgds:
> I've often wondered about this as a general concept. If you assume the text is enough to reach the required level of accessibility, is then okay to make it slightly more obvious for those who can perceive colour? (assuming that colour doesn't make the experience worse for another user).

I'm going to drop it onto the FE meeting agenda for discussion.

app/assets/stylesheets/govuk_publishing_components/components/_summary-list.scss
```diff
@@ -14,12 +13,27 @@
...
+    display: inline-block;
+    margin-right: govuk-spacing(3);
+  }
+}
+
+.gem-c-summary-list__group-actions-list {
+  margin-bottom: govuk-spacing(3);
+
+  @include govuk-media-query($from: tablet) {
+    display: inline-block;
```
andysellick, line 33: `inline-block` isn't needed if this element is floated.

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -1,23 +1,47 @@
 <%
+  local_assigns[:heading_level] ||= 3
+  heading_size = 'm' unless ['s', 'm', 'l', 'xl'].include?(heading_size)
```
andysellick, line 3: We have a thing in the [heading helper](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/heading_helper.rb#L10) for doing font size on a heading already, although it looks like it wouldn't do what you want exactly in its current state. Might be worth expanding it and using it here (although not blocking if you need to get this out quickly).

Or should this component use the heading component for this?

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -1,23 +1,47 @@
...
-      <% if edit.any? %>
-        <%= tag.ul class: "govuk-summary-list__actions-list" do %>
-          <%= tag.li class: "govuk-summary-list__actions-list-item" do %>
-            <%= link_to edit.fetch(:href),
-                      class: "govuk-link gem-c-summary-list__edit-section-link",
-                      title: "#{edit.fetch(:link_text, "Change")} #{title}",
-                      data: edit.fetch(:data_attributes, {}) do %>
-               <%= edit.fetch(:link_text, "Change") %> <%= tag.span title, class: "govuk-visually-hidden" %>
+      <%= content_tag(
+            shared_helper.get_heading_level,
```
andysellick, line 18: Nitpick: is these lines indented too much?

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -1,23 +1,47 @@
...
+            title,
+            class: "govuk-heading-#{heading_size} gem-c-summary-list__group-title"
+          )
+      %>
+      <% if edit.any? || delete.any? %>
+        <%= tag.ul class: "govuk-summary-list__actions-list gem-c-summary-list__group-actions-list" do %>
+          <%- if edit.any? %>
+            <% edit_section_link_text = edit[:link_text] || "Change" %>
```
andysellick, line 26: Does this text (and "Delete" on line 37) need to go into the `en.yml` file?

app/assets/stylesheets/govuk_publishing_components/components/_summary-list.scss
alex-ju: Good spot, a leftover from the first iteration (where on mobile they were kept inline, but that would've made actions at the group level inconsistent with actions at the item level).

app/views/govuk_publishing_components/components/_summary_list.html.erb
alex-ju: I'm trying to use the `SharedHelper` to render heading levels and provide custom `heading_size` consistently with other components while preserving the default values for backward compatibility. The heading component has some issues with spacing currently, so I wouldn't nest it in this component – this is the main reason we're not using it for `page_heading` on other input elements.

app/views/govuk_publishing_components/components/_summary_list.html.erb
alex-ju: We were wrapping it in all other places, happy to squish it on one line.

app/views/govuk_publishing_components/components/_summary_list.html.erb
alex-ju: Great improvement suggestion, have moved the hardcoded text in locales. We should raise issues for components that still have hardcoded text in order to align them.

alex-ju:
> Realised I should actually review this, having already discussed it so much! Just a few comments/thoughts.

Thank you for the review @andysellick – tried to address the comments.

Happy to discuss the use of colour in this scenario @maxgds – I worked with the assumption that we're already doing colour-based improvements for sighted people with things like error messages.

app/views/govuk_publishing_components/components/_summary_list.html.erb
andysellick: Yeah, fair enough, I figured as much. I think we'll file this under 'things to tidy'.

app/views/govuk_publishing_components/components/_summary_list.html.erb
andysellick: This is actually the thing I usually forget about when working on components, but I suddenly remembered it's one of our core principles! Good idea to raise issues, I've added a note to the component audit card in the backlog: https://trello.com/c/OvXd2E84/103-component-audit

PR #1576, opened 2020-06-16
Provide more guidance for developers

andysellick:
## What
Update the links in the component guide main page to provide clearer guidance for developers. Specifically:

1. Remove most of the existing text/links to make the remaining stuff stand out more.
1. Add link to how to release a new version of the gem.
1. Keep the link to all component documentation.
1. Add a section at the bottom to act as a 'what now?' if developers can't find a component they need.

## Why

1. The existing guidance is too long and isn't answering the questions developers have.
1. We have guidance for versioning the gem but developers aren't finding it. I've seen multiple PRs in the last few weeks where versioning has been done incorrectly. I've also been recently asked by several developers how to version the gem.
1. We should still link to _some_ documentation.
1. I also keep getting questions like 'what do I do if there's no component for this?' The section at the bottom should clarify this.

## Visual changes

I've updated the main guidance at the top of the page to link to the things I think are most relevant. It now looks like this:

Screenshot 2020-07-03 at 09 45 32

I've also added a new section at the very bottom to provide specific guidance if no suitable component is found. I was originally thinking of doing something clever with the search/filter JS to only show this if no results are found, but just putting it at the bottom is a lot simpler and means it is always visible (so if your search only finds one result, or two, this information is still shown). It looks like this at the bottom:

Screenshot 2020-07-03 at 09 48 48

and like this after doing a search:

Screenshot 2020-07-03 at 09 49 09

### DIFF
--- a/app/views/govuk_publishing_components/component_guide/index.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/index.html.erb
@@ -2,12 +2,7 @@
 
 <div class="component-markdown">
   <p>Components are packages of template, style, behaviour and documentation that live in your application.</p>
-  <p>A component must:</p>
-  <ul>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md">meet the component principles</a></li>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md">follow component conventions</a></li>
-  </ul>
-  <p>See the <a href="https://github.com/alphagov/govuk_publishing_components">govuk_publishing_components gem</a> for further details, or <a href="https://docs.publishing.service.gov.uk/manual/components.html#component-guides">a list of all component guides</a>.</p>
+  <p>See the <a href="https://github.com/alphagov/govuk_publishing_components">govuk_publishing_components gem</a> for further details, or <a href="https://docs.publishing.service.gov.uk/manual/components.html#component-guides">a list of all component guides</a>. Read about how to <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md">release a new version of the gem</a>.</p>
 </div>
 
 <form role="search" data-module="filter-components" class="component-search">
@@ -86,3 +81,7 @@
     </li>
   <% end %>
 </ul>
+
+<div class="component-markdown">
+  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/develop-component.md">creating a new one</a>.</p>
+</div>


chao-xian:
This is a great idea for people who aren't Frontend Devs. I think there's a little too much assumed knowledge and contextual understanding of the Design System and Components as it is.

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -8,6 +8,7 @@
     <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md">follow component conventions</a></li>
   </ul>
   <p>See the <a href="https://github.com/alphagov/govuk_publishing_components">govuk_publishing_components gem</a> for further details, or <a href="https://docs.publishing.service.gov.uk/manual/components.html#component-guides">a list of all component guides</a>.</p>
+  <p>If you cannot find a component to fit your need, ask in #govuk-frontenders. If there is no suitable component, please use the <a href="https://design-system.service.gov.uk/styles/">Design System styles</a>.</p>
```
chao-xian, line 11: I think we should give devs a link to the general Design System docs as well.

vanitabarrett, line 11: I like the link to the design system docs, that's really good 👍 

While I think pointing people to `govuk-frontenders` is good, I'm wondering if it puts the onus on frontend devs to help instead of encouraging backend devs to explore the code and try and contribute? 
I think at some point a frontend dev may need to get involved, but perhaps we should try and guide people to explore other options first, e.g: trying to identify a component they could expand; raising a PR for review?

alex-ju: I'm concerned that with '_if there is no suitable component, please use the <a href="https://design-system.service.gov.uk/styles/">Design System styles</a>_' we will promote class composition instead of isolation and re-usability.

andysellick:
Thanks for the comments everyone. I've updated the contents of this PR based on the feedback received (see description). This still may not be perfect but I think it's heading in the right direction.

andysellick:
Yes, my change to the links at the top is based on recent experience. Specifically:

1. **release process**: quite a few developers in the last few weeks have either asked how to do a release or tried to do a release incorrectly
1. **moving a component from an application to the gem**: Sean asked about it (okay, that's one occurrence, so maybe this link isn't essential)
1. **all documentation**: I figure we need at least one link to the actual documentation
1. **all component guides**: I use this quite often

If there's a need to slim it down, we could lose links 2 and 4 but I think 1 is important (because it comes up so often and it's clearly linked from the main github page, so people are clearly not even looking for it) and link 3 (or something like it) is needed just to point at the rest of the documentation.

Regarding the new section at the bottom, I think we won't need to maintain that much as the component principles/conventions will always need to exist (and presumably won't change their links), and linking to the Design System to help with styling information is what kicked off this change in the first place.

Does that sound sensible @alex-ju?

alex-ju:
As @chao-xian mentioned, I think we should do our best to update the [existing documentation on components](https://docs.publishing.service.gov.uk/manual/components.html) where all devs are used to look for such info – but that's my personal view, happy for other devs.

andysellick:
We should, and that'll be the next thing. But I think we need to start here and give some big roadsigns - because, as previously mentioned, developers aren't looking very hard to find some of the things that could answer their questions.

chao-xian:
Honestly, I think that if developers aren't looking at the Design System docs then how have they managed to find the Component Guide? Putting this info here as opposed to improving our existing docs doesn't feel like the right way round to solving this. GOV.UK sorely lacks a Tech Writer and sometimes it really shows in these situations.

andysellick:
Developers have told me they have looked in the component guide and sometimes not found a suitable component. That's why I raised this PR, to provide a bit more clear signposting. I agree that the other docs also need improving, but that's a separate issue that we can also look at.

Can I get a bit more of a clear steer on this? Do we think this PR isn't needed, or that it doesn't do enough?

chao-xian:
I think that the component guide is perhaps being too prescriptive and should at most point to the docs for contributing in both the `govuk_publishing_components` and Design System.

andysellick:
Thanks for the feedback. I've made a few changes to simplify the text/links and updated the PR description to try to clarify my approach.

I think we agree that the component guide should mainly link to the documentation, not provide documentation itself. Currently it links to the gem repo, and not the [documentation page](https://docs.publishing.service.gov.uk/manual/components.html). I think that this is the right approach, because we've been historically much better at updating the documentation in the repo, but interested to hear other opinions.

chao-xian:
I don't think being historically better at updating repos as opposed to the Developer Docs, which is considered the single source of truth, is necessarily a good reason to link to the repo instead. As long as the docs have a link on to the repo it should be fine - the likelihood of _that_ changing is unlikely...? There are definite problems with the Developer Docs but avoiding it doesn't help to fix it.

andysellick:
Okay, fair point. I've changed the link to point to the developer docs.

andysellick:
Do we need to discuss/change this PR anymore or is someone happy to approve?

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -86,3 +84,11 @@
     </li>
   <% end %>
 </ul>
+
+<div class="component-markdown">
+  <h2 class="component-doc-h2">If you cannot find a suitable component</h2>
```
alex-ju, line 89: I think this whole section could be resumed to one line, since the page we link to has the principles and conventions at the top of the page and explains when to build a new component.

```suggestion
  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/develop-component.md" class="govuk-link">creating a new one</a></p>.
```

As mentioned in a previous comment, I have concerns about promoting the usage of Design System styles as it can easily become a gateway to a non-componentised frontend.

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -2,12 +2,10 @@
 
 <div class="component-markdown">
   <p>Components are packages of template, style, behaviour and documentation that live in your application.</p>
-  <p>A component must:</p>
   <ul>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md">meet the component principles</a></li>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md">follow component conventions</a></li>
+    <li><a href="https://docs.publishing.service.gov.uk/manual/components.html">all documentation for components</a></li>
+    <li>how to <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md">release a new version of the gem</a></li>
```
alex-ju, line 7: This is tightly coupled with the repo (you create a pull request for your change via GitHub, you create a release via GitHub, I don't see how the components guide fits in this workflow) and we link to it from the main readme file.

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -2,12 +2,10 @@
 
 <div class="component-markdown">
   <p>Components are packages of template, style, behaviour and documentation that live in your application.</p>
-  <p>A component must:</p>
   <ul>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_principles.md">meet the component principles</a></li>
-    <li><a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/component_conventions.md">follow component conventions</a></li>
+    <li><a href="https://docs.publishing.service.gov.uk/manual/components.html">all documentation for components</a></li>
```
alex-ju, line 6: This feels valuable but seems a bit abrupt. I think a statement that also point to the repo, as we used to have is more natural ("See the govuk_publishing_components gem for further details, or a list of all component guides.")

app/views/govuk_publishing_components/component_guide/index.html.erb
andysellick: My hope was that by putting the link to the Design System styles as the very last thing on the page that developers would use it as their last resort, but I take your point, so I've made the change above.

app/views/govuk_publishing_components/component_guide/index.html.erb
andysellick: That's all true, but as I said there's a lot of people who don't know this procedure and keep getting it wrong. Developers know where the component guide is, so I'm just trying to provide this as even clearer signposting (yes, it's pretty easy to find from the repo, but I think our procedure is perhaps different to what developers are used to, which is why they don't look for it).

alex-ju: As discussed, let's get this in and keep an eye on the next releases (let's say by the end of the year) if they improve, we'll keep it, otherwise we'll remove it and look at other ways to communicate this at the community level.

andysellick: Thanks!

app/views/govuk_publishing_components/component_guide/index.html.erb
andysellick: Yeah, I agree. I'll restore the original wording, thanks.

PR #1582, opened 2020-06-22
Add print link component

theseanything:
## What
This component adds a link that includes a print icon and can be used to help users print a page.

This component renders two different outputs depending on whether a `href` is specified:

By default, when no `href` is given, the component renders as a button and triggers a print action via JavaScript. In this case the component is hidden in environments where JavaScript is disabled and can be used as a progressive enhancement.

When a `href` is specified the component renders as an anchor tag and navigates to that `href` without JavaScript, suitable for applications which have paths that trigger a print event on load. (for application such as Government Frontend)

## Why
This pattern was originally implemented in the Government Frontend application, however we'd like to re-use this component in Smart Answers.

We render a button instead of an anchor tag when using JavaScript because the action is not a navigation event. Some browsers have different behaviour for anchor tags with no `href` and could cause accessibility issues. We've styled the button as a link to keep visual consistency.

## Visual Changes
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix layout on input with prefix/suffix ([PR #1581](https://github.com/alphagov/govuk_publishing_components/pull/1581))
+* Add print link component ([PR #1582](https://github.com/alphagov/govuk_publishing_components/pull/1582))
 
 ## 21.56.1
 
--- a/app/assets/javascripts/govuk_publishing_components/components/print-link.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/print-link.js
@@ -1,0 +1,14 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  Modules.PrintLink = function () {
+    this.start = function (element) {
+      element[0].addEventListener('click', function () {
+        window.print()
+      })
+    }
+  }
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
@@ -1,0 +1,52 @@
+.gem-c-print-link {
+  display: none;
+  margin-bottom: 2em;
+  margin-top: 2em;
+}
+
+.gem-c-print-link.gem-c-print-link--show-without-js {
+  display: block;
+}
+
+.js-enabled {
+  .gem-c-print-link {
+    display: block;
+  }
+}
+
+.gem-c-print-link__link {
+  background: image-url("govuk_publishing_components/icon-print.png") no-repeat 10px 50%;
+
+  margin-left: -10px;
+  padding: .5em .5em .5em 38px;
+
+  @include govuk-device-pixel-ratio($ratio: 2) {
+    background-image: image-url("govuk_publishing_components/icon-print-2x.png");
+    background-size: 16px 18px;
+  }
+
+  &:focus {
+    @include govuk-focused-text;
+  }
+}
+
+.gem-c-print-link__button {
+  @extend %govuk-body-s;
+  background: image-url("govuk_publishing_components/icon-print.png") no-repeat 10px 50%;
+  border: 0;
+  color: $govuk-link-colour;
+  cursor: pointer;
+  margin: 0;
+  margin-left: -10px;
+  padding: .5em .5em .5em 38px;
+  text-decoration: underline;
+
+  @include govuk-device-pixel-ratio($ratio: 2) {
+    background-image: image-url("govuk_publishing_components/icon-print-2x.png");
+    background-size: 16px 18px;
+  }
+
+  &:focus {
+    @include govuk-focused-text;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_print_link.html.erb
+++ b/app/views/govuk_publishing_components/components/_print_link.html.erb
@@ -1,0 +1,27 @@
+<%
+  text ||= t('components.print_link.text')
+  href ||= nil
+  data_attributes ||= {}
+
+  require_js ||= href.nil?
+  data_attributes[:module] = 'print-link' if require_js
+%>
+
+<% if require_js %>
+  <div class="gem-c-print-link" >
+    <%= content_tag(:button, text, {
+        class: %w(gem-c-print-link__button govuk-link),
+        data: data_attributes
+    }) %>
+  </div>
+<% else %>
+  <div class="gem-c-print-link gem-c-print-link--show-without-js">
+    <%= link_to(
+      text,
+      href,
+      class: %w(gem-c-print-link__link govuk-link),
+      rel: "alternate",
+      data: data_attributes
+    ) %>
+  </div>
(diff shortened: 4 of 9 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 21.57.0
```
andysellick, line 10: This needs to be changed to `## Unreleased` - see https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md

app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
```diff
@@ -0,0 +1,30 @@
...
+.gem-c-print-link.gem-c-print-link__show-without-js {
+  display: block;
+}
+
+.gem-c-print-link {
+  display: none;
+  margin-top: 2em;
+  margin-bottom: 2em;
+
+  .gem-c-print-link__link {
```
andysellick, line 16: Could this class be un-nested from the `.gem-c-print-link` class? We try to avoid nesting if possible (and sorry, this probably isn't your code if you've copied it directly from government-frontend).

CHANGELOG.md
theseanything: Ah good catch!

app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
theseanything: Ah no worries - yeah I though `link__link` was a bit peculiar. I've un-nested and removes the extra div wrapping the link.

theseanything:
@andysellick Thanks for the review! I've addressed the suggested changes.

app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
andysellick: Ah, sorry, I didn't quite mean that. I mean instead of 

```
.gem-c-print-link {
   .gem-c-print-link__link { }
}
```

just un-nest it like so:

```
.gem-c-print-link { }
.gem-c-print-link__link { }
```

I don't think you should remove the wrapping div - it looks like it's broken the tests and also the styling on focus:

Screenshot 2020-06-23 at 10 08 52

Should look like this:

Screenshot 2020-06-23 at 10 10 42

app/views/govuk_publishing_components/components/_print_link.html.erb
```diff
@@ -0,0 +1,11 @@
+<%
+  text ||= t('components.print_link.text')
+  href ||= "javascript:window.print()"
```
alex-ju, line 3: This will currently raise a CSP warning (and soon it will be an error) as it generates in-line JavaScript. We don't use in-line JavaScript in the component library so I suggest we follow the current pattern and move this line into its own JavaScript file using the current module system (contextual-guidance, cookie-banner, and modal-dialogue are decent references for writing a JS module, but if things are not clear I'm more than happy to help).

theseanything: Ah righto - will do! Do you know what value I should set as the href? I read that without a href value some browsers don't treat anchor links as normal (but maybe that's not an issue). Also wasn't sure if href="#" was appropriate either.

andysellick: Might be worth not setting a default and raising an error/not rendering if no `href` is passed to the component?

DilwoarH: Could we just create this as a button rather than a link? The functionality is more like a button than a link anyway.

andysellick: Maybe worth considering for a future iteration.

theseanything: Ah, so I'm thinking of the case where there is no href and we want to rely entirely on JS.

Using just JS does make it seem more like a button - but then the implementation in government-frontend does use a href and makes sense to be a link.

alex-ju: Ideally, we'd render either a `<button class="gem-c-print-link" data-module="print-link">Print entire page</button>` or a `<a class="gem-c-print-link" href="/print">Print entire page</a>` based on wether or not the component has an `href` passed in. If we're under pressure to deliver we can push it as is and raise and issue to clear debt later.

theseanything: Thanks! Yup - I think that makes sense, we aren't under enormous pressure so I'd to get it right and not add to the tech debt. I'm just seeing what design would be appropriate for the button - will update the PR and try to do the button or link implementation.

alex-ju: We needed something similar ([a button styled as a link](https://github.com/alphagov/content-publisher/blob/master/app/assets/stylesheets/core/_link.scss#L1-L10)) in content-publisher, if that's of any help.

theseanything:
@andysellick, @alex-ju - I've PR with a button/link implementation.

app/assets/javascripts/govuk_publishing_components/components/print-link.js
```diff
@@ -0,0 +1,14 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  Modules.PrintLink = function () {
+    this.start = function (element) {
+      element[0].addEventListener('click', function () {
```
DilwoarH, line 9: Doesn't the following work?

```suggestion
      element.addEventListener('click', function () {
```

theseanything: Ah no :( 
image

I was a bit confused by this - but appears the the code initialising the modules passes in an array of elements. I've updated the variable name to be plural to reflect this.

alex-ju: I'll try and cast a shade of light on this. The module system relies on jQuery (that's how it was implemented and [we're in the process of changing this, but it will require a breaking change](https://github.com/alphagov/govuk_publishing_components/pull/1538)). So the function expects `element` to be a jQuery object, but within the module you wrote you're using plain/vanilla JavaScript and this is the reason you need to use `element[0]`, to get the raw `nodeElement` from a jQuery-wrapped object. So the variable name was correct as it represents a single `element`.

theseanything: Ah thanks @alex-ju, that make a tonne of sense! I've changed the naming back.

app/assets/stylesheets/govuk_publishing_components/components/_print-link.scss
```diff
@@ -0,0 +1,52 @@
+.gem-c-print-link {
+  display: none;
+  margin-bottom: 2em;
+  margin-top: 2em;
+}
+
+.gem-c-print-link.gem-c-print-link__show-without-js {
```
alex-ju, line 7:
```suggestion
.gem-c-print-link--show-without-js {
```
This class, being used at the block level, feels more like a modifier class than a different element.

theseanything: fixed!

PR #1586, opened 2020-06-23
Hide services cookie banner when JavaScript is disabled

alex-ju:
## What
When JavaScript is disabled we should hide the cookie banner user in services as its sole purpose is to let the user accept/decline cookies used for analytics purposes, which are only being done if JavaScript is enabled.

## Why
To avoid confusing users when JavaScript is disabled

## Visual Changes
Changes to the cookie banner variation used in services (with `services_cookies` set, such as Covid-19 services).
No changes to the cookie banner used on GOV.UK

<table>
<tr><th>JavaScript enabled</th><th>JavaScript disabled</th></tr>
<tr><td valign="top">

Screenshot 2020-06-23 at 17 02 51

</td><td valign="top">

Screenshot 2020-06-23 at 17 03 10

</td></tr>
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
+
 ## 21.56.2
 
 * Fix layout on input with prefix/suffix ([PR #1581](https://github.com/alphagov/govuk_publishing_components/pull/1581))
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Hide services cookie banner when JavaScrips is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
```
injms, line 12:
```suggestion
* Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
```

injms, line 12: Why hide the banner for only services? The analytics on GOV.UK is also Javascript only, so it could be useful there as well.

alex-ju: not really, on GOV.UK we use it for user settings too, some of them handled server-side. I won't touch the main GOV.UK one outside of a mission work 😬

PR #1587, opened 2020-06-24
Add more font sizes to heading component

andysellick:
## What
Add more font sizes to the heading component. We had 19 and 24 (defaulting to 27), which correspond to govuk headings small and medium and this PR adds 36 (large) and 48 (extra large).

I'm not sure why we have 27 as the default as that doesn't match a govuk heading size, but I'm not going to remove it as it's in use all over the place.

## Why
We probably didn't have these font sizes as options before because we also have the Page title component, which provides the 48px size. However I've started to wonder why we need two components and whether one could be enough to provide all needed functionality. This change is an interim step towards merging these two components, plus I need headings of these sizes in one of our applications.

## Visual Changes

Screenshot 2020-06-24 at 09 48 51

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
@@ -1,15 +1,10 @@
 .gem-c-heading {
-  @include govuk-text-colour;
-  @include govuk-font(27, $weight: bold);
   margin: 0;
 }
 
-.gem-c-heading--font-size-24 {
-  @include govuk-font(24, $weight: bold);
-}
-
-.gem-c-heading--font-size-19 {
-  @include govuk-font(19, $weight: bold);
+.gem-c-heading--font-size-27 {
+  @include govuk-text-colour;
+  @include govuk-font(27, $weight: bold);
 }
 
 // special case for publications and consultations pages
--- a/app/views/govuk_publishing_components/components/docs/heading.yml
+++ b/app/views/govuk_publishing_components/components/docs/heading.yml
@@ -24,10 +24,13 @@ examples:
       text: 'Original consultation'
       heading_level: 3
   different_font_sizes:
-    description: Choose a different font size. Valid options are 24 (24px) and 19 (19px), with the component defaulting to 27px, appropriate for a H1. This option is not tied to the heading_level option in order to give flexibility.
+    description: |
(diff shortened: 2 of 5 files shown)


app/views/govuk_publishing_components/components/docs/heading.yml
```diff
@@ -24,10 +24,18 @@ examples:
...
       heading_level: 3
   different_font_sizes:
-    description: Choose a different font size. Valid options are 24 (24px) and 19 (19px), with the component defaulting to 27px, appropriate for a H1. This option is not tied to the heading_level option in order to give flexibility.
+    description: |
+      Choose a different font size, defaults to 27px. Valid options are:
+
+      - 48 (48px)
+      - 36 (36px)
+      - 24 (24px)
+      - 19 (19px)
```
injms, line 33: I've always found it slightly unintuitive that the headings in the Component gem doesn't emulate the Design System's `xl`, `l` etc naming convention whilst emulating the heading styles. It also means that the style names lead to a dissonance between the name and what it does - for example, `.gem-c-heading--font-size-48` isn't always 48px in size.

Whilst the old naming convention shouldn't be removed for backwards compatibility reasons, is it worth updating this component to match the Design System naming convention? 

The number-based names could still be kept as an undocumented option for compatibility - but in this pull request the names could be updated to match the Design System's names. It's not a breaking change, since the old sizes are still useable.

Using the DS styles could mean less CSS shippped since the components could use `govuk-heading-*` classes, and we could remove some of the `gem-heading--*` classes.

alex-ju: I share @injms's view that we should try and deprecate the value-based class names in favour of T-shirt sizes for both consistency with the Design System and, more importantly, maintainability.

andysellick: Thanks for the thinking. I've reworked the component to allow this, can you take another look?

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -1,15 +1,10 @@
 .gem-c-heading {
   @include govuk-text-colour;
```
injms, line 2: Since the text colour is set to with the mixin `govuk-text-colour` mixin on the `govuk-heading-*` classes anyway we could move this to the `.gem-c-heading--font-size-27` to prevent any unnecessary repetition.

andysellick: Great spot, have changed, thanks!

PR #1588, opened 2020-06-24
Improve print styles for govspeak callouts

theseanything:
## What and why
- This fixes the margin for info notices so text is centred.
- Removes the border for help and info notices as it disrupts the existing styling.
- Adds padding for call to action boxes to improve readability of text.
- Removes the grey background in call to action boxes to prevent wasted ink when printing.

## Visual Changes
Call to action:
image

Help notice:
image

Info notice:
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
+* Improve print styles for govspeak info, help and call to action callouts ([PR #1588](https://github.com/alphagov/govuk_publishing_components/pull/1588) )
 
 ## 21.56.2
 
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
+* Improve print styles for govspeak info, help and call to action callouts ([PR #1588](https://github.com/alphagov/govuk_publishing_components/pull/1588) )
```
andysellick, line 13: Can this match the PR title? (you could just update the PR title)

theseanything:
> Also the before/after screenshots for 'help notice' seem to be of different things, can you clarify?

Thanks @andysellick! Sorry, yeah the screenshot were a bit confusing, I've updated them to reflect more clearly what has changed.

> there are three changes listed in the description, but only one thing changed in the code, does this one change affect all of these things?

Yup, so I think the git diff makes it look like a smaller change than it is. But the end result changes to the CSS are:

- Removed print specific border and padding from help/info notices
- Added print specific padding and background to call-to-action

(I'll add this to the PR description too)

PR #1590, opened 2020-06-25
Extend action link component with a small icon variant

danacotoran:
## What

Extend action link component with a small icon variant.

## Why
There is a requirement for a small icon version of the default (dark arrow on yellow background) action link component.
The design has been tweaked in order to have two such links on the Coronavirus landing page header.

image

https://trello.com/c/oB87oGqD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
 * Improve print styles for govspeak info, help and call to action callouts ([PR #1588](https://github.com/alphagov/govuk_publishing_components/pull/1588) )
+* Extend action link component with a small icon variant ([PR #1590](https://github.com/alphagov/govuk_publishing_components/pull/1590) )
 
 ## 21.56.2
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
@@ -118,6 +118,16 @@
 }
 
 .gem-c-action-link--dark-icon {
+  &:before {
+    // sass-lint:disable no-duplicate-properties
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.png");
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.svg"), linear-gradient(transparent, transparent);
+    // sass-lint:enable no-duplicate-properties
+  }
+}
+
+.gem-c-action-link--dark-icon,
+.gem-c-action-link--small-icon {
   max-width: none;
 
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Hide services cookie banner when JavaScript is disabled ([PR #1586](https://github.com/alphagov/govuk_publishing_components/pull/1586))
 * Improve print styles for govspeak info, help and call to action callouts ([PR #1588](https://github.com/alphagov/govuk_publishing_components/pull/1588) )
+* Extend action link component with a small icon variant ([PR #1590](https://github.com/alphagov/govuk_publishing_components/pull/1590) )
```
andysellick, line 14: Can this match the PR title please?

app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -118,19 +118,31 @@
...
+    // sass-lint:disable no-duplicate-properties
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.png");
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.svg"), linear-gradient(transparent, transparent);
+    // sass-lint:enable no-duplicate-properties
+  }
+}
+
+.gem-c-action-link--small-icon {
+  &:before {
+    width: 35px;
```
andysellick, line 132: I'm a bit confused why small is 35px wide and dark is 30px, when the icons for each are the same size. Would it be possible to make them consistent?

danacotoran: Yeah the icons are the same size, but the header version requires more visual space between the icon and the text vs the light background/dark text version. 

See mockup (left) vs existing `dark_icon` version (right)

Screenshot 2020-06-25 at 11 59 05


We could make the spacing consistent with the `dark_icon` one – I just think it might look a bit squished? I can check with Mia if she thinks it looks ok with the smaller spacing though.

andysellick: I was thinking the other way round... the `dark_icon` variant looks a little squished. But yeah, see what Mia thinks. I'm just against having tiny differences that add more code if there's not a good reason to have those differences.

danacotoran: Yeah agree with you, I also dislike having these little exceptions for not very good reasons. 😕 
Might be better to increase spacing for the other variant indeed. Will discuss with Mia and get back to you on that.

danacotoran: @andysellick I've spoken to Mia and she agrees it would be best to keep things consistent. She suggested that we match the existing dark icon style for now (I've updated the PR to reflect that)

PR #1593, opened 2020-06-25
Minor adjustment to action link icon spacing

danacotoran:
Add 5px to increase space between arrow and text in the small/dark icon versions of the arrow link component, as requested by Mia

Before (left) vs after (right)

Screenshot 2020-06-25 at 17 48 31

https://trello.com/c/oB87oGqD

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 21.57.1
+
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 21.57.1
```
andysellick, line 10: I'm late to this party, but any change shouldn't also version the gem, we do those separately, see https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md

PR #1594, opened 2020-06-30
Fix search component label accessibility

andysellick:
## What
- when the page is linearized ([web developer toolbar](https://chrispederick.com/work/web-developer/), miscellaneous, linearize page) the absolutely positioned label for the search input (which is pretending to be placeholder text) appears before the search input and is unreadable on a blue background
- this fixes this by adding a white background to the label, so the appearance is unchanged in the normal view but readable when linearized

## Why
As raised by @selfthinker 

## Visual Changes

Before (normal view)

Screenshot 2020-06-30 at 16 15 02

Before (linearized)

Screenshot 2020-06-30 at 16 15 17

After (linearized)

Screenshot 2020-06-30 at 17 01 44

After when focussed (linearized)

Screenshot 2020-06-30 at 17 01 53

After with some custom colours applied

Screenshot 2020-06-30 at 17 10 44

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -9,6 +9,7 @@ $large-input-size: 50px;
 .gem-c-search__label {
   @include govuk-font($size: 19, $line-height: $input-size);
   display: block;
+  background: govuk-colour("white");
 
   h1 {
     @include govuk-font($size: 19, $line-height: $input-size);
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -18,7 +19,7 @@ $large-input-size: 50px;
   .js-enabled & {
     position: absolute;
     left: govuk-spacing(3);
```
injms, line 21: We've lost the spacing to the left of the label text when it's linearised - would this help fix it?

```suggestion
    left: 0;
    padding-left: govuk-spacing(3);
```

andysellick:
Thanks both. That's some odd behaviour - the overlap problem seems to be only occurring on the preview app, not on my local version. Will investigate.

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
andysellick: Good spot, thanks.

andysellick:
Should be fixed now. Have also tested it using some custom colours, appears to be correct.

PR #1595, opened 2020-07-01
Add list component

injms:
## What

This is a direct port of the list component from GOV.UK Frontend / Design System with no extra styles added.

## Why

Unordered lists were being hardcoded with a `each` loop in the templates. It seems a more elegant solution to have a component that we can pass an array of strings to display.

## Visual Changes

No before, as this is a new addition:

image

### DIFF
--- a/app/views/govuk_publishing_components/components/_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_list.html.erb
@@ -1,0 +1,26 @@
+<%
+  id ||= nil
+  extra_spacing ||= nil
+  list_type ||= "unordered"
+  visible_counters ||= nil
+  items ||= []
+  aria_label ||= nil
+
+  classes = %w(gem-c-list govuk-list)
+  classes << "govuk-list--bullet" if visible_counters && list_type === "unordered"
+  classes << "govuk-list--number" if visible_counters && list_type === "number"
+  classes << "govuk-list--spaced" if extra_spacing
+
+  # Default list type is unordered list.
+  list_tag = "ul"
+
+  # Set to ordered list to override default.
+  list_tag = "ol" if list_type === "number"
+%>
+<% if items.any? %>
+  <%= content_tag list_tag, class: classes, id: id, "aria-label": aria_label do %>
+    <% items.each do |item| %>
+      <li><%= sanitize(item) %></li>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/list.yml
+++ b/app/views/govuk_publishing_components/components/docs/list.yml
@@ -1,0 +1,64 @@
+name: List
+description: A list - unordered or ordered, with or without counters / bullet points.
+body: |
+  This is an ordered or unordered list, with or without visible bullets / numbers.
+
+  The `items` parameter can include HTML to display - such as links or another
+  list. This HTML can either be directly coded or come from another component.
+accessibility_criteria: |
+  The list must:
+
+  - inform the user how many items are in the list
+  - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
+
+  The list may:
+
+  - include an `aria-label` to contextualise the list if helpful
+govuk_frontend_components:
+  - list
+examples:
+  default:
+    description: "The default is an unordered list with no bullet points or numbers."
+    data: &default-example-data
+      items:
+      - "Tony&rsquo;s Chocolonely"
+      - '<a href="https://en.wikipedia.org/wiki/Cherry_Ripe_(chocolate_bar)" rel="noopener" class="govuk-link">Cherry Ripe</a>'
+      - "Snickers"
+      - "Chomp"
+      - "Penguin"
+      - "Boost"
+  unordered_list_with_aria-label:
+    description: "A list with an aria-label"
+    data:
+      aria_label: "A list of delicious chocolate bars."
+      <<: *default-example-data
+  unordered_list_with_bullet_points:
+    description: "An unordered list with visible bullet points."
+    data:
+      visible_counters: true
+      <<: *default-example-data
+  ordered_list_without_numbers:
+    description: "This is an ordered list where the numbers aren't visible."
+    data:
+      list_type: "number"
+      <<: *default-example-data
+  ordered_list_with_numbers:
+    description: |
+      This is an ordered list with the numbers visible.
+    data:
+      list_type: "number"
+      visible_counters: true
+      <<: *default-example-data
+  with_extra_spacing:
+    description: |
+      Increases the amount of spacing between the list items.
+    data:
+      extra_spacing: true
+      <<: *default-example-data
+  with_id_attribute:
+    description: |
+      Sets the `id` on the `ul` or `ol` element.
+    data:
+      id: 'super-fantastic-chocolate-list'
+      <<: *default-example-data
--- a/spec/components/list_spec.rb
+++ b/spec/components/list_spec.rb
@@ -1,0 +1,101 @@
+require "rails_helper"
+
+describe "List", type: :view do
+  def component_name
+    "list"
+  end
+
+  it "does not render anything if no data is passed" do
+    test_data = {}
+
+    assert_empty render_component(test_data)
+  end
+
+  it "does not render anything if nothing is in the items array" do
+    test_data = { items: [] }
+
+    assert_empty render_component(test_data)
+  end
+
+  it "renders an unordered list by default" do
+    render_component(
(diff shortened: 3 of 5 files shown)


andysellick:
Where will this be used? You mention hard coded lists in the description.

injms:
> Where will this be used? You mention hard coded lists in the description.

The use case that set this in motion was adding unordered lists inside a summary list for the coronavirus business volunteer service. This was using a [partial that looped through an array to display the list items](https://github.com/alphagov/govuk-coronavirus-business-volunteer-form/blob/47725cdfcb6deaa7a8ece2d63f3f32602691de6c/app/views/coronavirus_form/check_your_answers/helpers/_list.html.erb)

Whilst it's not a great bother to have a list as a partial - it works after all - I thought it was something that could be a nice component, similar to the heading component.

app/views/govuk_publishing_components/components/_list.html.erb
```diff
@@ -0,0 +1,26 @@
+<%
+  id ||= nil
+  extra_spacing ||= nil
+  list_style ||= nil
```
andysellick, line 4: I think this is unused?

app/views/govuk_publishing_components/components/_list.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  list_type ||= "unordered"
```
andysellick, line 5: I'm wondering if this would be more usable as a true/false flag, since there are only two options? Something like

```
bullets ||= false
...

classes << "govuk-list--bullet" if visible_counters && bullets
classes << "govuk-list--number" if visible_counters && bullets == false
...

list_tag = "ol" unless bullets
```

app/views/govuk_publishing_components/components/_list.html.erb
```diff
@@ -0,0 +1,26 @@
...
+  visible_counters ||= nil
+  items ||= []
```
andysellick, line 7: I haven't fully formulated this thought yet but it's around the idea that you have to pass a chunk of markup for each list item. I can't think of another way of doing it, because anything else would restrict what you could display and you don't want that (say, if the list item is text with two links in, or even some more complex formatting).

I think my concern is that for some uses of this component it would produce a situation where either the view gets quite cluttered or we start assembling markup outside of the view and injecting it. I don't know. Let's have a conversation about this. Thoughts? Is this a concern?

andysellick: On that subject, could you put an example into the YML of a list that contains links?

app/views/govuk_publishing_components/components/docs/list.yml
```diff
@@ -0,0 +1,59 @@
...
+accessibility_criteria: |
+  The list must:
+
+  - inform the user how many items are in the list
+  - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
+
+  The contents list may:
+
+  - include an aria-label to contextualise the list if helpful
```
andysellick, line 14: How would this be added? The component doesn't seem to have such an option.

app/views/govuk_publishing_components/components/docs/list.yml
```diff
@@ -0,0 +1,59 @@
...
+body: |
+  This is an ordered or unordered list, with or without visible bullets / numbers.
+accessibility_criteria: |
+  The list must:
+
+  - inform the user how many items are in the list
+  - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
+
+  The contents list may:
```
andysellick, line 12: We should be consistent with our terminology, especially since there's already a component named 'Contents list' - this is a little confusing. Is this meant to be here?

app/views/govuk_publishing_components/components/docs/list.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  - include an aria-label to contextualise the list if helpful
+shared_accessibility_criteria:
+  - link
+  - contents_list
```
andysellick, line 17: I don't think there's a shared accessibility criteria for `contents_list`? This doesn't seem to output anything.

app/views/govuk_publishing_components/components/docs/list.yml
```diff
@@ -0,0 +1,59 @@
...
+
+  - inform the user how many items are in the list
+  - convey the content structure
+  - indicate the current page when contents span different pages, and not link to itself
+
+  The contents list may:
+
+  - include an aria-label to contextualise the list if helpful
+shared_accessibility_criteria:
+  - link
```
andysellick, line 16: Is this required? The component itself doesn't contain links, unless they're passed to it. I guess since that's allowed this is okay, but then don't we need other accessibility criteria to cover anything else that might be passed? I'd rather keep it simple. Maybe just a criteria to say anything passed should meet accessibility criteria for itself. Or something.

spec/components/list_spec.rb
```diff
@@ -0,0 +1,73 @@
+require "rails_helper"
+
+describe "List", type: :view do
+  def component_name
+    "list"
+  end
+
+  it "does not render anything if no data is passed" do
+    test_data = {}
```
andysellick, line 9: Does it matter that this test passes an object instead of an array, which the component expects? It still passes, but is it a fair test? Or do we need both tests?

app/views/govuk_publishing_components/components/docs/list.yml
injms: Yes, sorry - missed that! Fixed now.

app/views/govuk_publishing_components/components/_list.html.erb
injms: Whilst there are only two options and a boolean would work, I think setting `list_type` to "ordered" or "unordered" gives a clearer intent and doesn't confuse with whether the counters are shown or not. 

It also gives us space to expand to include other ordinals - for example, square bullets or Roman numerals - without breaking backwards compatibility.

app/views/govuk_publishing_components/components/_list.html.erb
injms: It is! Removed.

app/views/govuk_publishing_components/components/_list.html.erb
injms: Sure - I'll add an example. It'll be the same as the example with an `&rsquo;` in it and it's always good to be clearer.

I think adding markup into each list item is okay, as other options are much too restrictive. We allow passing in markup into other components - such as the accordion - and having elements inside a `li` is allowed by the HTML spec. Every string is put through `sanitize()`, which should help stop it breaking if some erroneous markup is used.

app/views/govuk_publishing_components/components/docs/list.yml
injms: I've added that option - good spot!

app/views/govuk_publishing_components/components/docs/list.yml
injms: Good point - I've removed `link`

spec/components/list_spec.rb
injms: The component expects `{ items: ["one", "two", "three"] }` as a minimum - so a test to test for graceful failure with an empty `items` array is a good idea.

app/views/govuk_publishing_components/components/docs/list.yml
injms: Boo. Removed.

app/views/govuk_publishing_components/components/_list.html.erb
andysellick: Ah okay hadn't thought about other options like that. Just whenever I see strings passed as options it looks wrong to me.

spec/components/list_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+  it "adds an `id` correctly" do
+    render_component(
+      id: "this-is-a-test-id",
+      items: ["Test item", "Another test item"],
+    )
+
+    assert_select "ul#this-is-a-test-id li", text: "Test item"
+    assert_select "ul#this-is-a-test-id li:nth-child(2)", text: "Another test item"
+  end
+end
```
DilwoarH, line 89: Hey @injms - a common use case is the use of links within lists. It might be worth adding a test for this (and possibly some documentation too)

injms: Good call - I've added an explanation at the top the documentation and a test to ensure that links are being rendered correctly. Each example has a link in it since they all use the same data - shoutout to YAMLs ability to reference itself nicely.

PR #1597, opened 2020-07-06
Remove jQuery from details component

injms:
## What

On the details component the jQuery click listener has been replaced with a JavaScript event listener - swapped `click()` with `addEventListener()`. No test changes required.

Minor change in the way an object was referenced to make the Standard linter happy.

## Why

Part of the ongoing drive to remove jQuery as a dependency and improve code neatness.

## Visual Changes

None.


injms:
Waiting on #1598.

PR #1598, opened 2020-07-06
Set up Chrome for JavaScript tests

injms:
## What

Replace PhantomJS with Chrome for the JavaScript Jasmine tests .

## Why

Because [PhantomJS is no longer in development](https://github.com/ariya/phantomjs/issues/15344) and is out of date. This is impacting our ability to remove jQuery as a dependency because PhantomJS is throwing errors for things that [our browser support matrix](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices) supports, such as `addEventListener` or `NodeList.forEach()`

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/support/jasmine_helper.rb
+++ b/spec/javascripts/support/jasmine_helper.rb
@@ -1,15 +1,10 @@
-# Use this file to set/override Jasmine configuration options
-# You can remove it if you don't need it.
-# This file is loaded *after* jasmine.yml is interpreted.
-#
-# Example: using a different boot file.
-# Jasmine.configure do |config|
-#   config.boot_dir = '/absolute/path/to/boot_dir'
-#   config.boot_files = lambda { ['/absolute/path/to/boot_dir/file.js'] }
-# end
+require "jasmine_selenium_runner/configure_jasmine"
 
-# Prevent PhantomJS auto install, uses PhantomJS already on your path.
-Jasmine.configure do |config|
-  config.prevent_phantom_js_auto_install = true
-  config.show_console_log = false
+class ChromeHeadlessJasmineConfigurer < JasmineSeleniumRunner::ConfigureJasmine
+  def selenium_options
+    chrome_options = Selenium::WebDriver::Chrome::Options.new
+    chrome_options.headless!
+    chrome_options.add_argument("--no-sandbox")
+    { options: chrome_options }
+  end
 end
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -122,6 +122,9 @@ GEM
       rack (>= 1.2.1)
       rake
     jasmine-core (3.5.0)
+    jasmine_selenium_runner (3.0.0)
+      jasmine (~> 3.0)
+      selenium-webdriver (~> 3.8)
     json-schema (2.8.1)
       addressable (>= 2.4)
     kgio (2.11.3)
@@ -316,6 +319,7 @@ DEPENDENCIES
   govuk_schemas (~> 4.0)
   govuk_test (~> 1)
   jasmine (~> 3.5.1)
+  jasmine_selenium_runner (~> 3.0.0)
   pry-byebug
   rspec-rails (~> 4.0)
   rubocop-govuk (~> 3)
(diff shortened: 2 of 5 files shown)


Gemfile
```diff
@@ -7,6 +7,11 @@ source "https://rubygems.org"
 # development dependencies will be added by default to the :development group.
 gemspec
 
+group :development, :test do
+  # The Jasmine Selenium runner needs `require: false` to work properly - and
+  # this doesn't seem to be an option in the `gemspec` file.
+  gem "jasmine_selenium_runner", require: false
```
alex-ju, line 13: While I'm not one of those Ruby devs that consider Gemfile a big no for gems, for consistency we can move this to gemspec as `require: false` is default in gemspec so we don't need to worry abut that. I suggest we add this to this to the [gemspec](https://github.com/alphagov/govuk_publishing_components/blob/master/govuk_publishing_components.gemspec#L32) as `s.add_development_dependency "jasmine_selenium_runner", "~> 3.0.0"`

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -1,15 +1,15 @@
...
-# You can remove it if you don't need it.
-# This file is loaded *after* jasmine.yml is interpreted.
-#
-# Example: using a different boot file.
-# Jasmine.configure do |config|
-#   config.boot_dir = '/absolute/path/to/boot_dir'
-#   config.boot_files = lambda { ['/absolute/path/to/boot_dir/file.js'] }
-# end
+require "jasmine/runners/selenium"
+require "webdrivers/chromedriver"
```
alex-ju, line 2: we shouldn't need to require this explicitly, we should have it available via govuk_test

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -1,15 +1,15 @@
...
 
-# Prevent PhantomJS auto install, uses PhantomJS already on your path.
 Jasmine.configure do |config|
   config.prevent_phantom_js_auto_install = true
```
alex-ju, line 5: We can get rid of this line as we won't be relying on `PhantomJS` anymore – unless I'm missing something

spec/javascripts/support/jasmine_helper.rb
injms: Super - I've removed it.

spec/javascripts/support/jasmine_helper.rb
injms: 👍

Gemfile
injms: Hmm, this is a tricky and strange one then.

I originally added the `jasmine_selenium_runner` in the gemspec file - but then the tests were not running correctly. It defaulted to trying to find Firefox to run the tests in - which is the default setting. 

With the dependency in the gemspec file, and running `bundle exec rake app:jasmine:ci` the error is:

```
Selenium::WebDriver::Error::WebDriverError: Could not find Firefox binary (os=macosx). Make sure Firefox is installed or set the path manually with Selenium::WebDriver::Firefox::Binary.path=
```

Adding the dependency to the Gemfile meant that the tests ran correctly in headless chrome. 

I thought it was something to do with `require: false` - removing that from the Gemfile resulted in the same 'Can't find Firefox' error.

I'll dig a bit more and see why this could be. There isn't a consistency for Gemfile vs gemspec across the organisation, which makes things a bit trickier. Most of the implementations I've found that get headless Chrome to run Jasmine tests use a Gemfile, so I'll try and see what a gemspec set up does.

injms: A fix was found for content-publisher (https://github.com/alphagov/content-publisher/pull/2110) which I have copied into this pull request. All seems to be working 🤞

kevindew: Yay, glad you were able to make use of that 👍

kevindew: By the way - even though gems don't automatically pull in gems, this happens in our test suite because of this line: https://github.com/alphagov/govuk_publishing_components/blob/206cde79934a62577f2c368c8c869ac2933f4d14/spec/dummy/config/application.rb#L16 (which is a bit of a smell as a part of the test suite makes this app behave differently)

PR #1604, opened 2020-07-15
Make component auditing code more resilient

andysellick:
## What
The auditing code makes a lot of assumptions about being able to find specific directories, which means it's liable to errors or unexpected output when it can't find them. For example, when it's running on review apps sometimes it can't find itself, which meant that it was throwing up a ton of 'component does not exist' warnings when it shouldn't have been.

Additionally, some things should be shown or not shown depending on whether the auditing code has been able to find them - for example, if it can't find any applications it shouldn't display (blank) information about component usage in applications.

Additionally improves the scope of the auditing code by relaxing some of the restrictions on file extensions and regexes when looking for component instances in ruby. This was prompted by the auditing missing a component in smart answers.

## Why
Code should be robust.

## Visual Changes
None.

### DIFF
--- a/app/models/govuk_publishing_components/audit_comparer.rb
+++ b/app/models/govuk_publishing_components/audit_comparer.rb
@@ -1,12 +1,12 @@
 module GovukPublishingComponents
   class AuditComparer
-    attr_reader :data, :gem_data
+    attr_reader :applications_data, :gem_data
 
     def initialize(gem_data, results)
       if gem_data[:gem_found]
         @gem_data = gem_data
-        @data = sort_results(results)
-        @gem_data[:components_by_application] = get_components_by_application
+        @applications_data = sort_results(results)
+        @gem_data[:components_by_application] = get_components_by_application || []
       end
     end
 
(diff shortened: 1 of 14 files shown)


app/models/govuk_publishing_components/audit_comparer.rb
```diff
@@ -18,51 +20,58 @@ def sort_results(results)
...
-            },
-            {
-              name: "Components in ruby",
-              value: ruby[:components].join(", "),
-            },
-          ],
-          warnings: warnings,
-          warning_count: warnings.length,
-        }
+        if !result[:application_found]
```
injms, line 23: Does this mean the same thing using `unless`? For example: 
```ruby
unless result[:application_found]
```

app/models/govuk_publishing_components/audit_components.rb
```diff
@@ -38,18 +48,16 @@ def initialize(path)
...
 
-  private
-
     def find_files(files, replace)
       files.map { |file| clean_file_name(file.gsub(replace, "")) }.sort
     end
 
     def clean_file_name(name)
-      name.tr("/_-", " ").gsub(".html.erb", "").gsub(".scss", "").gsub(".js", "").gsub("spec", "").gsub(".rb", "").strip
+      name.tr("/_-", " ").gsub(".html.erb", "").gsub(".erb", "").gsub(".scss", "").gsub(".js", "").gsub("spec", "").gsub(".rb", "").strip
```
injms, line 56: Just for readability, could this be:

```suggestion
      name
        .tr("/_-", " ")
        .gsub(".html.erb", "")
        .gsub(".erb", "")
        .gsub(".scss", "")
        .gsub(".js", "")
        .gsub("spec", "")
        .gsub(".rb", "")
        .strip
```

injms, line 56: Earlier there was a change from `*.html.erb` to `*erb` - does this change need to happen here as well?
```suggestion
      name.tr("/_-", " ").gsub(".erb", "").gsub(".erb", "").gsub(".scss", "").gsub(".js", "").gsub("spec", "").gsub(".rb", "").strip
```

app/models/govuk_publishing_components/audit_components.rb
```diff
@@ -38,18 +48,16 @@ def initialize(path)
...
     end
 
     def get_component_name_from_full_path(path)
-      path.gsub("/_", "/").gsub(@templates_full_path, "").gsub(".html.erb", "").tr('\"\'', "")
+      path.gsub("/_", "/").gsub(@templates_full_path, "").gsub(".html.erb", "").gsub(".erb", "").tr('\"\'', "")
```
injms, line 60: Similar to previous suggestion for readability:

```suggestion
      path
        .gsub("/_", "/")
        .gsub(@templates_full_path, "")
        .gsub(".html.erb", "")
        .gsub(".erb", "")
        .tr('\"\'', "")
```

app/models/govuk_publishing_components/audit_comparer.rb
andysellick: Yes, changed, thanks.

app/models/govuk_publishing_components/audit_components.rb
andysellick: Changed, thanks.

spec/component_guide/component_index_spec.rb
```diff
@@ -68,7 +69,7 @@
 @import 'govuk_publishing_components/components/tabs';
 @import 'govuk_publishing_components/components/title';"
 
-    expect(page).to have_selector(".component-doc-h2", text: "Gem components used by this app (13)")
+    expect(page).to have_selector(".component-doc-h2", text: "Gem components used by this app (15)")
```
injms, line 72: Will this expectation break when components are added to or removed from the gem?

app/models/govuk_publishing_components/audit_components.rb
andysellick: Good spot! Actually no. We want to remove the file extension from the filename, but it might have found an `.erb` file or `.html.erb` file, because we're looking for `.erb`, which could be either.

spec/component_guide/component_index_spec.rb
andysellick: It will break if components are added or removed from the dummy gem. This does happen occasionally, but it's fairly easy to figure out.

PR #1605, opened 2020-07-16
Fix JavaScript linting errors

injms:
## What

This was mostly:

 - changing the way objects are accessed
 - swapping double quotes for single quotes
 - objects keys with quotes having quotes remove

And one change of `getAttribute` to `hasAttribute` to check whether an attribute is present without needing a comparison.

## Why

Because [we should lint with Standard](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#linting) - and another branch I'm working on was failing Standard listing.

## Visual Changes

None.

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -30,8 +30,13 @@ namespace :assets do
 end
 
 desc "Run RuboCop linting"
-task lint: :environment do
+task lint_ruby: :environment do
   sh "bundle exec rubocop --format clang"
 end
 
-task default: [:lint, :spec, "app:jasmine:ci"]
+desc "Run Javascript and Sass linting"
+task lint_js_and_sass: :environment do
+  sh "npm run lint"
+end
+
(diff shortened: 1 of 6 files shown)


injms:
> Looks good. I had a look through the component guide and checked the affected components, don't seem to be any problems. I assume all of these syntax changes are compatible on older browsers?

I've checked, and as far as I can tell the changes are all compatbile

```javascript
"jam" === 'jam' // true
```

Accessing properties with the dot notation or the bracket notation is interchangable, as long as the keys are letters, numbers, `$`, and `_` - but doesn't start with a number. This rule was added to Standard in August last year, which could possibly be why it didn't throw any errors until recently.

Putting quotes around the object keys doesn't make a difference either - unless, like accessing object properties, you add spaces are start the key with a number.

And 

```javascript
$link.getAttribute('data-youtube-player-analytics') == "true"
```

gives the same boolean result as 

```javascript
$link.hasAttribute('data-youtube-player-analytics')
```

without needing the double equals and is supported since IE 8.

PR #1610, opened 2020-07-17
Switch from NPM to Yarn

kevindew:
## What

This converts NPM to Yarn for consistency with other GOV.UK Rails based projects.

## Why

In GOV.UK we have a [Rails app convention][convention] to use Yarn over NPM for Rails projects due to the deeper integration of Yarn into Rails. It doesn't reflect any personal preference towards Yarn over NPM.

[convention]: https://docs.publishing.service.gov.uk/manual/conventions-for-rails-applications.html#javascript-package-management


Jenkinsfile
```diff
@@ -6,11 +6,11 @@ node {
...
     rubyLintDiff: false,
     beforeTest: {
-      stage("Install npm dependencies") {
-        sh("npm install")
+      stage("Install yarn dependencies") {
+        sh("yarn")
       }
       stage("Lint Javascript and SCSS") {
-        sh("npm run lint")
+        sh("yarn run lint")
```
injms, line 13: The linting is also part of the default rake task - does including it in the Jenkinsfile mean that the linting will run twice?

kevindew: Oh good point, I'll see if it can be removed.

PR #1611, opened 2020-07-17
Trial switching to Stylelint for (S)CSS linting

kevindew:
Trello: https://trello.com/c/9UQe86W7/107-change-to-a-non-ruby-based-scss-linter

## What

This replaces the abandoned [sass-lint](https://github.com/sasstools/sass-lint) project in favour of using [stylelint](https://stylelint.io/). It uses a pre-release version of [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds) as the underlying ruleset, which we intend to become the de facto (S)CSS linting approach for GDS.

## Why

This is trialling the adoption of a linter that GOV.UK projects can migrate from [scss-lint](https://github.com/sds/scss-lint) which is based on deprecated [Ruby Sass](https://sass-lang.com/ruby-sass). Removing that project will remove Ruby Sass from the GOV.UK stack, allowing adoption of Libsass

I felt it would be useful to try this tool out in a real project to help learn if it is configured correctly for adoption across GDS and GOV.UK. I chose govuk_publishing_components as a place to try this out since a) it's a rare project used by most GOV.UK frontenders; b) compared to some projects it's quite modern in CSS approaches; c) it's already somewhat inconsistent with GOV.UK by using [sass-lint](https://github.com/sasstools/sass-lint) currently.

## Wider motivation

A key motivation of this PR is to try get eyeballs on [stylelint-config-gds](https://github.com/alphagov/stylelint-config-gds) and find out if any of the configuration there is divisive, confusing or simply just something we should debate.

The approach taken for the ruleset of stylelint-config-gds is:

1. Have as few rules as possible, instead leaning on community standards. In a similar vein to how standard for JS resolves the need for lots of rule discussion
2. Be consistent with govuk-frontend styles to offer the least painful path for initial adoption forward
3. Can be applied to plain CSS as well as SCSS, since vanilla CSS has been regaining popularity for years
4. Be sufficiently agreeable that the frontend community of GDS can, or aspire to, adopt the rules

Thus, most of the overrides in stylelint-config-gds are adapting linting rules that [govuk-frontend has chosen](https://github.com/alphagov/govuk-frontend/blob/ce8e22355e32af4d8f3c876779acd26128401842/config/.sass-lint.yml) for consistency. There were a few that weren't compatible (`bem-depth`, `border-zero`, `no-color-literals`) and one I purposely omitted was property ordering as I believe that preference is specific to the govuk-frontend project.

## How and what to review

To review the overall linting rules introduced here you have 3 things to investigate:

- [stylelint-config-gds/scss](https://github.com/alphagov/stylelint-config-gds/blob/39e82c4197f26c303137787f1318599253b991b9/scss.js) this applies rules for GDS projects that only apply to SCSS
- [stylelint-config-gds/css](https://github.com/alphagov/stylelint-config-gds/blob/39e82c4197f26c303137787f1318599253b991b9/css.js) this applies rules for GDS projects that can be applied to  CSS, and contains most of the overriding for govuk-frontend consistency
- [stylelint-config-standard](https://github.com/stylelint/stylelint-config-standard) the foundational project that applies the initial community rules.

Please open issues on [styelint-config-gds](https://github.com/alphagov/stylelint-config-gds/issues/new) for things that should be tweaked. I'm hoping for that we're generally able to accept the community approach to rules and only disable/adapt them when they cause us genuine problems, to avoid getting bogged down in conflicting personal opinions on linting rules. 

To review this PR it is easiest to go through it commit by commit to see the impact of turning on a particular rule and whether this causes any undesirable consequences.

The two rules that seemed to be most impactful were `selector-pseudo-element-colon-notation` and the `max-nesting-depth`. For the former all pseudo-elements were converted to double colons. Which, as I understand, used to be a problem for IE8 but is not a concern now.  The latter rule, `max-nesting-depth` limits to nesting rules to a depth of 2 (which effectively allows one class inside a class and then one media query, this one didn't reflect the approach we've always used for writing SCSS where nesting used to be quite common. However I understand that the widely accepted view is that nesting can be difficult to understand and hard to search so low levels of nesting are preferred (as adopted by the design system), so this felt like something to aspire to.

## Next steps

I'm going to open a draft PR on govuk-frontend applying the rule-set there to try solicit wider input into the rule-set.


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
```diff
@@ -266,10 +268,6 @@
     text-align: left;
   }
 
-  .mc-chart.mc-negative .mc-td.mc-key-cell {
-    padding-right: 25px;
-  }
```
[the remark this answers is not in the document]
kevindew: Ah yes I can see this seems cryptic. The rule is actually already defined in a nested form a few lines above: https://github.com/alphagov/govuk_publishing_components/blob/6439cae0fd7214a36cff7a3453fad3e99da9a3d0/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss#L260-L262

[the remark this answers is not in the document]
andysellick: So this is removing a duplicate rule? Okay, cool, good find.

app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -1,15 +1,13 @@
 .gem-c-action-link {
   display: table;
 
-  &:before {
+  &::before {
```
injms, line 4: The double colon pseudo-element isn't supported in IE 8 - whilst we don't support IE8, I think it'd be worthwhile keeping it as a single colon to give us time to check what effect that this could have so that we can ensure that it doesn't break anything.

```suggestion
  &:before {
```

kevindew: Thanks @injms for taking a look and the feedback. Do you think this is a change all of GDS should have or just this repo?

This rule is an area of discussion to have with the design-system too as they have a mixture of single and double colons usage in their codebase. They, however, use autoprefixer which replaces the double ones with single ones anyway so their IE8 stylesheets are generated with single colons.

If we don't support IE8 do we need to worry about maintaining support for one of its features? I had a quick look in IE8 on gov.uk homepage and can see a few things broken so I'd wonder if this would be an overly-cautious move that blurs the lines of support and non-support.

andysellick: We don't support IE8, but for the sake of an extra colon it doesn't seem worth potentially making IE8 users experience of GOV.UK worse. There's still something like 40,000 IE8 users (at last count, admittedly from nearly a year ago) but that's a lot of people. GOV.UK looks quite broken in IE8, but it should still be usable.

injms: I think that this is a change that all of GDS should have - so can put that into a comment on the linter itself if that helps? 

Changing from single to double is a change that has no effect _apart_ from breaking more things in IE 8. The support for both single and double is the same - and will continue to be so as browsers have to keep supporting both. 

In the spirit of users > devs > technical purity, because this is a change that only breaks things for some of the users stuck on IE8,  and because this doesn't fix or improve anything apart from the code - I think single colon pseudo elements should be the default.

kevindew: Sure thanks for explaining, I'll flip this round in the rules to be for single colon. If that then proves divisive we can discuss further.

kevindew: @injms or @andysellick, I've just opened https://github.com/alphagov/stylelint-config-gds/pull/1 would one of you mind reviewing?

kevindew: I've switched this around now @injms

injms: Super - thanks for making the changes.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -37,28 +37,26 @@ $table-row-even-background-colour: govuk-colour("light-grey", $legacy: "grey-4")
...
-
-      &::after {
-        content: "";
-        position: absolute;
-        top: 5px;
-        right: 0;
-
-        @include govuk-shape-arrow($direction: up, $base: $sort-link-arrow-size-small, $display: block);
-      }
+    .app-table__sort-link &:focus {
```
alex-ju, line 42: I suggest we simplify this:
```suggestion
    .app-table__sort-link:focus {
```

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -37,28 +37,26 @@ $table-row-even-background-colour: govuk-colour("light-grey", $legacy: "grey-4")
...
+      @include govuk-focused-text;
+    }
 
-      &::before {
-        content: "";
-        position: absolute;
-        top: 13px;
-        right: 0;
+    .app-table__sort-link &:after {
```
alex-ju, line 46: I suggest we simplify this:
```suggestion
    .app-table__sort-link:after {
```

kevindew:
Thanks @alex-ju - I've applied changes in b715645

PR #1615, opened 2020-07-21
Fix PhantomJS creeping back into tasks

kevindew:
## What

This changes the approach to loading Gems to be more explicit to resolve surprisingly consequences for JS tests. In the  process of this it also removes some gems that aren't needed and moves others, that are only used in dev, to development dependencies.

## Why

In https://github.com/alphagov/govuk_publishing_components/pull/1598 we added the necessary configuration to replace PhantomJS for JS tests with Chromedriver run through Selenium. Unfortunately due to Bundler/load order complexities this didn't apply in all places, most notably the default `rake` task which is run by CI.

## Visual Changes

None, I hope

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,9 +1,5 @@
-begin
-  require "rspec/core/rake_task"
-  RSpec::Core::RakeTask.new(:spec)
-rescue LoadError
-  puts "Running in production mode"
-end
+require "rubocop/rake_task"
+require "rspec/core/rake_task"
 
 APP_RAKEFILE = File.expand_path("spec/dummy/Rakefile", __dir__)
 
@@ -11,6 +7,9 @@ load "rails/tasks/engine.rake"
 load "rails/tasks/statistics.rake"
 
 require "bundler/gem_tasks"
+
+RuboCop::RakeTask.new
+RSpec::Core::RakeTask.new
 
 namespace :assets do
   desc "Test precompiling assets through dummy application"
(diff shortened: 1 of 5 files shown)


Rakefile
```diff
@@ -29,14 +28,9 @@ namespace :assets do
...
-task lint_ruby: :environment do
-  sh "bundle exec rubocop --format clang"
-end
-
-desc "Run Javascript and Sass linting"
-task lint_js_and_sass: :environment do
+desc "Linting for Ruby, JS and SASS"
+task lint: %i[rubocop environment] do
   sh "yarn run lint"
 end
```
injms, line 34: How does this task lint the Ruby? I can see it running the listing for the JavaScript and Sass via Yarn, but not how Rubocop is used.

spec/dummy/config/application.rb
```diff
@@ -3,17 +3,13 @@
 require "action_controller/railtie"
 require "action_view/railtie"
 require "action_mailer/railtie"
-require "rails/test_unit/railtie"
+# we need to load govuk_test before jasmine_selenium runner so webdrivers is
```
injms, line 6:
```suggestion
# We need to load govuk_test before jasmine_selenium runner so webdrivers is
```

kevindew: Thanks - just changed that.

Rakefile
kevindew: It's on line 32. As rubocop is a rake task we can define it as a dependant task that runs before the code in the lint block runs.

Not sure I explained this particularly well so let me know if it still doesn't make sense.

kevindew:
@injms this should be good again now. Seems to be working properly on Heroku.

Rakefile
injms: Ah - the `%i[rubcocop environment]` is running the task - so if we wanted to add another dependent task it'd be `%[rubocop another_task environment]`?

kevindew: Yup, that's correct

PR #1616, opened 2020-07-21
Add support for custom sizes for radio page headers

DilwoarH:
## What
Add custom page header size support for radio inputs.

## Why
This allows long questions to be displayed as page titles on applications like smart answers.

### Before
Screenshot 2020-07-21 at 13 01 16

### After
Screenshot 2020-07-21 at 13 01 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Make component auditing more resilient ([PR #1604](https://github.com/alphagov/govuk_publishing_components/pull/1604))
 * Add more font sizes to heading component ([PR #1587](https://github.com/alphagov/govuk_publishing_components/pull/1587))
 * Format machine-readable output ([PR #1617](https://github.com/alphagov/govuk_publishing_components/pull/1617))
+* Add support for custom sizes for radio page headers ([PR #1616](https://github.com/alphagov/govuk_publishing_components/pull/1616))
 
 ## 21.59.0
 
--- a/app/views/govuk_publishing_components/components/_radio.html.erb
+++ b/app/views/govuk_publishing_components/components/_radio.html.erb
@@ -6,10 +6,18 @@
   label ||= nil
   heading ||= nil
   heading_caption ||= nil
+  heading_size ||= nil
   small ||= false
   inline ||= false
   is_page_heading ||= false
-  heading_size = 'm' unless ['s', 'm', 'l', 'xl'].include?(heading_size)
+
+  if ['s', 'm', 'l', 'xl'].include?(heading_size)
+    size = heading_size
+  elsif is_page_heading
+    size = 'xl'
+  else
+    size = 'm'
+  end
 
   description ||= nil
   hint ||= nil
@@ -28,7 +36,7 @@
   radio_classes << "govuk-radios--inline" if inline
 
   legend_classes = %w(govuk-fieldset__legend)
-  legend_classes << "govuk-fieldset__legend--#{heading_size}"
+  legend_classes << "govuk-fieldset__legend--#{size}"
 
   aria = "#{hint_id} #{"#{error_id}" if has_error}".strip if hint or has_error
 
@@ -40,9 +48,9 @@
 
     <% if heading.present? %>
       <% if is_page_heading %>
-        <%= tag.legend class: "govuk-fieldset__legend govuk-fieldset__legend--xl gem-c-title" do %>
-          <%= tag.span(heading_caption, class: "govuk-caption-xl") if heading_caption.present? %>
-          <%= tag.h1 heading, class: "gem-c-title__text" %>
+        <%= tag.legend class: legend_classes do %>
+          <%= tag.span(heading_caption, class: "govuk-caption-#{size}") if heading_caption.present? %>
+          <%= tag.h1 heading, class: "gem-c-radio__heading-text govuk-fieldset__heading" %>
         <% end %>
       <% else %>
         <%= tag.legend heading, class: legend_classes %>
--- a/app/views/govuk_publishing_components/components/docs/radio.yml
+++ b/app/views/govuk_publishing_components/components/docs/radio.yml
@@ -183,6 +183,10 @@ examples:
       - value: "blue"
         text: "Blue"
   with_custom_heading_size:
+    description: |
+      This allows the size of the legend to be changed. Valid options are s, m, l, xl, defaulting to m if no option is passed. 
+
+      If the is_page_heading option is true and heading_size is not set, the text size will be xl.
     data:
       name: "radio-group-description"
       heading: "What is your favourite colour?"
--- a/spec/components/radio_spec.rb
+++ b/spec/components/radio_spec.rb
@@ -149,6 +149,52 @@ describe "Radio", type: :view do
     assert_select "legend span.govuk-caption-xl", false
   end
 
+  it "renders radio-group with a custom page heading size" do
+    render_component(
+      name: "favourite-skittle",
+      heading: "What is your favourite skittle?",
+      heading_caption: "Question 3 of 9",
+      heading_size: "l",
+      is_page_heading: true,
+      items: [
+        { label: "Red", value: "red" },
+        { label: "Blue", value: "blue" },
+      ],
+    )
+    assert_select ".govuk-radios"
+    assert_select "legend.govuk-fieldset__legend--l h1.govuk-fieldset__heading", text: "What is your favourite skittle?"
+    assert_select "legend span.govuk-caption-l", "Question 3 of 9"
+  end
+
+  it "renders radio-group with no custom page heading size" do
+    render_component(
+      name: "favourite-skittle",
+      heading: "What is your favourite skittle?",
+      heading_caption: "Question 3 of 9",
+      is_page_heading: true,
+      items: [
+        { label: "Red", value: "red" },
+        { label: "Blue", value: "blue" },
+      ],
+    )
+    assert_select ".govuk-radios"
+    assert_select "legend.govuk-fieldset__legend--xl h1.govuk-fieldset__heading", text: "What is your favourite skittle?"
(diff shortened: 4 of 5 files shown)


app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -40,9 +42,9 @@
 
     <% if heading.present? %>
       <% if is_page_heading %>
-        <%= tag.legend class: "govuk-fieldset__legend govuk-fieldset__legend--xl gem-c-title" do %>
-          <%= tag.span(heading_caption, class: "govuk-caption-xl") if heading_caption.present? %>
-          <%= tag.h1 heading, class: "gem-c-title__text" %>
+        <%= tag.legend class: legend_classes do %>
+          <%= tag.span(heading_caption, class: "govuk-caption-#{heading_size}") if heading_caption.present? %>
+          <%= tag.h1 heading, class: "gem-c-title__text govuk-heading-#{heading_size}" %>
```
36degrees, line 47: To make a legend large, you should use `govuk-fieldset__legend--l` on the legend, and the heading should just have `govuk-fieldset__heading`.

See https://design-system.service.gov.uk/get-started/labels-legends-headings/ for info.

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -9,6 +9,7 @@
   small ||= false
   inline ||= false
   is_page_heading ||= false
+  heading_size = 'xl' if is_page_heading && !heading_size.present?
```
andysellick, line 12: The interplay of these different options (`is_page_heading`, `heading_size`) I'm finding quite confusing. Is there a way of simplifying this? 

Could we have one option for heading (which would take 1, 2, 3 etc. but default to nil, resulting  in no heading) and one option for size (s, m, etc.) and then they wouldn't have to interact?

andysellick: Yeah, that's fine, still feels simpler than the current approach.

app/views/govuk_publishing_components/components/_radio.html.erb
DilwoarH: Done

app/views/govuk_publishing_components/components/_radio.html.erb
DilwoarH: Correct me if I've misunderstood you but the current solution is `heading_size` and `is_page_heading`. All the above line is doing is checking if a heading size has been set when `is_page_heading` is passed in and if it's not then it's set to `xl` by default.a

DilwoarH: I've updated the code a little bit - let me know if this is still confusing

DilwoarH:
> Also need to update the changelog.

Done 👍

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Make component auditing more resilient ([PR #1604](https://github.com/alphagov/govuk_publishing_components/pull/1604))
 * Add more font sizes to heading component ([PR #1587](https://github.com/alphagov/govuk_publishing_components/pull/1587))
+* Add support for custom sizes for radio page headers ([PR #1616](https://github.com/alphagov/govuk_publishing_components/pull/1616))
```
andysellick, line 14: Nitpick - can this match the PR name please?

app/views/govuk_publishing_components/components/_radio.html.erb
andysellick: Sorry, maybe I need more sleep but I'm just finding line 12 hard to read. The heading size is set to xl, unless `is_page_heading` is true and no `heading_size` is specified. Otherwise, on the next line, set heading size to `m` unless a correct heading size has been passed.

If `is_page_heading` is true but no `heading_size` is passed, heading size is set to `m`, is that right? If `is_page_heading` is false but a `heading_size` is passed, it gets set to whatever was passed (if valid).

Could it be rewritten as

```
heading_size = 'm' unless ['s', 'm', 'l', 'xl'].include?(heading_size)
heading_size = 'xl' if is_page_heading
```

DilwoarH: hey that would not work because we want to change the header size.

We only want to default to `xl` if there isn't a `heading_size` passed in

36degrees: Perhaps it's the 'double assignment' (heading_size getting set to 'm' and then immediately changed to 'xl') that's confusing?

Would something like this work? (not tested!)

```ruby
if ['s', 'm', 'l', 'xl'].include?(heading_size)
  size = heading_size
elsif is_page_heading
  size = 'xl'
else
  size = 'm'
end
```

Or, if you want to avoid the extra variable:

```ruby
unless ['s', 'm', 'l', 'xl'].include?(heading_size)
  if is_page_heading
    heading_size = 'xl'
  else
    heading_size = 'm'
  end
end
```

andysellick: I like the first option. It's more lines of code, but much clearer in its intent.

DilwoarH: Thanks @36degrees ! I've updated the code.

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -28,7 +36,8 @@
   radio_classes << "govuk-radios--inline" if inline
 
   legend_classes = %w(govuk-fieldset__legend)
-  legend_classes << "govuk-fieldset__legend--#{heading_size}"
+  legend_classes << "govuk-fieldset__legend--#{size}"
+  legend_classes << "gem-c-title" if is_page_heading
```
andysellick, line 40: Sorry, just noticed, this class `gem-c-title` is from another component. I don't think components should be using each others classes like this - could this be a `govuk-heading` class instead?

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -40,9 +49,9 @@
 
     <% if heading.present? %>
       <% if is_page_heading %>
-        <%= tag.legend class: "govuk-fieldset__legend govuk-fieldset__legend--xl gem-c-title" do %>
-          <%= tag.span(heading_caption, class: "govuk-caption-xl") if heading_caption.present? %>
-          <%= tag.h1 heading, class: "gem-c-title__text" %>
+        <%= tag.legend class: legend_classes do %>
+          <%= tag.span(heading_caption, class: "govuk-caption-#{size}") if heading_caption.present? %>
+          <%= tag.h1 heading, class: "gem-c-title__text govuk-fieldset__heading" %>
```
andysellick, line 54: Not your change (mine in fact from two years ago, sorry), but inline with my previous comment, could this `gem-c-title__text` class be part of this component? You might have to copy the class across.

Since we've now switched to individual component assets for all our applications if the radio component were included without the title component, this part of it would be unstyled.

spec/components/radio_spec.rb
```diff
@@ -149,6 +149,23 @@ def component_name
...
+      items: [
+        { label: "Red", value: "red" },
+        { label: "Blue", value: "blue" },
+      ],
+    )
+    assert_select ".govuk-radios"
+    assert_select "legend.govuk-fieldset__legend--l h1.govuk-fieldset__heading", text: "What is your favourite skittle?"
+    assert_select "legend span.govuk-caption-l", "Question 3 of 9"
+  end
+
```
andysellick, line 168: I think we might need a few more tests, correct me if I'm wrong.

- a test that `size` defaults to `m` if no `heading_size` is passed and `is_page_heading` is not set
- a test that `size` is `xl` if `is_page_heading` is true

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -134,6 +134,19 @@ examples:
         text: "Yes"
       - value: "no"
         text: "No"
+  with_custom_page_header_sizes:
+    data:
```
andysellick, line 138: I think having the `is_page_heading` option here detracts from the point of the example. Maybe add a description to say something like `If heading_size is passed, size is set to that value. If is_page_heading is passed instead, the size defaults to xl`.

app/views/govuk_publishing_components/components/_radio.html.erb
DilwoarH: This line is no longer needed

app/views/govuk_publishing_components/components/_radio.html.erb
DilwoarH: I've added the CSS needed for this

DilwoarH:
@andysellick - all comments have been addressed

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -134,6 +134,25 @@ examples:
         text: "Yes"
       - value: "no"
         text: "No"
+  with_custom_page_heading_sizes:
+    description: |
+      This option allows the question to be a page heading (within a H1 element) and an option to have
```
andysellick, line 139: I think this text is potentially confusing. The size can be set regardless of whether it is a page heading or not. Suggestion:

```
This allows the size of the legend to be changed. Valid options are s, m, l, xl, defaulting to m if no option is passed. 

If the is_page_heading option is true and heading_size is not set, the text size will be xl. 
```

Also the description here states the options are uppercase letters, when in fact if those are passed they are ignored, accepted values are lowercase.

spec/components/radio_spec.rb
andysellick: Thanks for adding some more tests, this looks much better.

app/views/govuk_publishing_components/components/docs/radio.yml
DilwoarH: Thanks - updated

PR #1618, opened 2020-07-21
Add govspeak to radio component description

injms:
## What
Wrapped the radio component's description in a govspeak component so that any markup used in the description is properly displayed. An example in the docs has been added for this use case, and the test has been updated.

## Why

The radio component wrapped the description in a `div` element that has a `govuk-body` class. This is okay if there's only a single paragraph of text for the description; but the correct styles aren't added if markup is used in the description.

This happens on several of the simple smart answers in `frontend`.

Wrapping the description in a govspeak component and a `govuk-body` class means that both strings that doesn't contain any markup and strings that do contain markup will inherit the appropriate styles. Thanks CSS!

Changing the way the component accepts a description wasn't an option as this flaw is already being used in production - the needs to be used for markup _and_ for text.

Regex could've been used to test whether the string started with an HTML element (for example: `/^<[a-z]>/`), and then only if there was markup present - but I thought that was too conservative in what the description parameter could successfully accept.

### From [question 3 of 'Continue to live in the UK if you're an EU, EEA or Swiss citizen'](https://www.gov.uk/staying-uk-eu-citizen/y/yes/no)

| Before | After |
| - | - |
| image | image |

### From [question 1 of 'Tell DVLA you've sold, transferred or bought a vehicle'](https://www.gov.uk/sold-bought-vehicle/y)

| Before | After |
| - | - |
| image | image |

### From [question 4 of 'Check if you need to send a Self Assessment tax return'](https://www.gov.uk/check-if-you-need-tax-return/y/no/less-than-50-000/no)

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/app/views/govuk_publishing_components/components/_radio.html.erb
+++ b/app/views/govuk_publishing_components/components/_radio.html.erb
@@ -11,12 +11,12 @@
   inline ||= false
   is_page_heading ||= false
 
-  if ['s', 'm', 'l', 'xl'].include?(heading_size)
+  if ["s", "m", "l", "xl"].include?(heading_size)
     size = heading_size
   elsif is_page_heading
-    size = 'xl'
+    size = "xl"
   else
-    size = 'm'
+    size = "m"
   end
 
   description ||= nil
@@ -58,7 +58,11 @@
     <% end %>
 
     <% if description.present? %>
-      <%= tag.div description, class: "govuk-body" %>
+      <% output_description = render "govuk_publishing_components/components/govspeak" do %>
+        <%= tag.div description, class: "govuk-body" %>
+      <% end %>
+
+      <%= output_description %>
     <% end %>
 
     <% if hint %>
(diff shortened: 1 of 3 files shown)


sihugh:
Does it need a line in the changelog?

andysellick:
I think we're currently looking at whether this could be achieved in-application by using a rails `capture` block?

injms:
The component's been updated to use `capture` - can it have a re-review @alex-ju  please?

injms:
> Does it need a line in the changelog?

I don't think so, since this is an internal change to the gem.

injms:
> Apologies for not leaving a proper review on this promptly.
> 
> My argument against this change is that by calling the `govspeak` component in `radio` we will create a dependency on a large component that may not be used elsewhere in the project. This can be easily avoided by using the govspeak component to render the description at the application level (and I raised [alphagov/frontend#2437](https://github.com/alphagov/frontend/pull/2437) to address this in `frontend`) then pass the output to the `radio` component.
> 
> I will not block this PR since I won't be around for the next two weeks or so and I hope it will reach a compromise soon.

Your argument makes sense - I'll close this pull request.

We should also make sure that the wider community is aware so we can keep an eye out for unstyled descriptions, and that https://github.com/alphagov/frontend/pull/2437 is the preferred (and quick) way of fixing it. I'll add it to the agenda for the next GOV.UK Frontenders meeting.

PR #1619, opened 2020-07-22
Add running locally details to README

kevindew:
This is to reflect that there wasn't guidance on JS package managers. It
also adds an example of how to run the guide locally and breaks the
documentation into locally and not locally changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,20 +44,27 @@ GOV.UK Publishing Components shares the same standards in terms of browser and a
 
 [See the rubydoc.info documentation](http://www.rubydoc.info/gems/govuk_publishing_components)
 
-Run the documentation locally with:
+
+## Working locally
+
+### Install dependencies
+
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -44,20 +44,27 @@ GOV.UK Publishing Components shares the same standards in terms of browser and a
 
 [See the rubydoc.info documentation](http://www.rubydoc.info/gems/govuk_publishing_components)
 
-Run the documentation locally with:
+
+## Working locally
+
+### Install dependencies
+
+As well as the Ruby, you will also need the [yarn package manager](https://classic.yarnpkg.com/en/docs/install/) to allow installing [NPM](https://npmjs.com) dependencies.
```
alex-ju, line 52: capitalisation in line with branding
```suggestion
As well as the Ruby, you will also need the [yarn package manager](https://classic.yarnpkg.com/en/docs/install/) to allow installing [npm](https://npmjs.com) dependencies.
```

kevindew: Thanks, just amended.

PR #1620, opened 2020-07-22
Replace jQuery in checkboxes.js

alex-ju:
## What
Replace jQuery usage in checkboxes.js with standard functions.

## Why
Part of removing jQuery from GOV.UK.

Quick performance check on nested checkboxes using the script below:
```
document.getElementById('checkboxes-971849fc-0').click() // select all siblings
document.getElementById('checkboxes-971849fc-0-0').click() // unselect a child
document.getElementById('checkboxes-971849fc-1').click() // select all siblings
document.getElementById('checkboxes-971849fc-1-0').click() // unselect a child
```

<table>
<tr><th>Before</th><th>After</th></tr>
<tr><td valign="top">

~7.64 milliseconds

</td><td valign="top">

~1.56 milliseconds

</td></tr>
</table>

## Visual Changes
No visual changes

[Trello card](https://trello.com/c/sghHecNi)

PR #1621, opened 2020-07-22
Fix `get_component_name_from_full_path`

alex-ju:
## What
Jenkins generates working paths based on branch names, sometimes starting with an underscore (e.g. `/var/lib/jenkins/workspace/_replace-jquery-in-checkboxes-js/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb`). In such cases `get_component_name_from_full_path` matches all `/_` occurrences and replaces them with `/` messing the path then failing to read the dummy file. The whole model file should ideally be covered by unit testing to avoid such fails.

## Why
To prevent PRs (such as #1419, #1620 or #1622) from silently failing in CI. Ideally, the functions in models should be covered by unit tests.

## Visual Changes
No visual changes

### DIFF
--- a/app/models/govuk_publishing_components/audit_components.rb
+++ b/app/models/govuk_publishing_components/audit_components.rb
@@ -64,7 +64,7 @@ module GovukPublishingComponents
     end
 
     def get_component_name_from_full_path(path)
-      path.gsub("/_", "/")
+      path.gsub(/.*\K\/_/, "/")
         .gsub(@templates_full_path, "")
         .gsub(".html.erb", "")
         .gsub(".erb", "")


app/models/govuk_publishing_components/audit_components.rb
```diff
@@ -64,7 +64,7 @@ def clean_file_name(name)
     end
 
     def get_component_name_from_full_path(path)
-      path.gsub("/_", "/")
+      path.gsub(/.*\K\/_/, "/")
```
injms, line 67: I'm unfamiliar with what some of this regex does. For future people, can we add in a comment to explain the regex please?

alex-ju: Sure, I've added a more granular explanation to the commit message now. I assumed familiarity from when the auditing code got in https://github.com/alphagov/govuk_publishing_components/pull/1589/files#diff-83be4a6e581e7b1299afb560e4e51015R103.

alex-ju:
are you fine with me merging this, @injms?

PR #1623, opened 2020-07-23
Add support for custom heading size for checkboxes

DilwoarH:
## What
Add custom page header size support for checkboxes.

## Why
This allows long questions to be displayed as page titles on applications like smart answers.

### Before

Screenshot 2020-07-23 at 12 32 53

### After

Screenshot 2020-07-23 at 12 33 00

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -184,6 +184,10 @@ examples:
         - label: "Blue"
           value: "blue"
   with_custom_heading_size:
+    description: |
+      This allows the size of the legend to be changed. Valid options are s, m, l, xl, defaulting to m if no option is passed. 
+
+      If the is_page_heading option is true and heading_size is not set, the text size will be xl.
     data:
       name: "favourite_colour"
       heading: "What is your favourite colour?"
--- a/lib/govuk_publishing_components/presenters/checkboxes_helper.rb
+++ b/lib/govuk_publishing_components/presenters/checkboxes_helper.rb
@@ -14,7 +14,6 @@ module GovukPublishingComponents
                   :id,
                   :hint_text,
                   :description,
-                  :heading_size,
                   :heading_caption,
                   :has_exclusive
 
@@ -35,9 +34,8 @@ module GovukPublishingComponents
 
         @id = options[:id] || "checkboxes-#{SecureRandom.hex(4)}"
         @heading = options[:heading] || nil
-        @heading_size = options[:heading_size]
-        @heading_size = "m" unless %w[s m l xl].include?(@heading_size)
         @heading_caption = options[:heading_caption] || nil
+        @heading_size = options[:heading_size] || nil
         @is_page_heading = options[:is_page_heading]
         @description = options[:description] || nil
         @no_hint_text = options[:no_hint_text]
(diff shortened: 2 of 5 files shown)


lib/govuk_publishing_components/presenters/checkboxes_helper.rb
```diff
@@ -35,14 +36,21 @@ def initialize(options)
...
-        @heading_size = "m" unless %w[s m l xl].include?(@heading_size)
         @heading_caption = options[:heading_caption] || nil
+        @heading_size = options[:heading_size] || nil
         @is_page_heading = options[:is_page_heading]
         @description = options[:description] || nil
         @no_hint_text = options[:no_hint_text]
         @hint_text = options[:hint_text] || "Select all that apply." unless @no_hint_text
         @visually_hide_heading = options[:visually_hide_heading]
+
+        @size = if %w[s m l xl].include?(@heading_size)
```
andysellick, line 47: This could be split out into a separate method.

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -123,6 +123,22 @@ examples:
           value: "green"
         - label: "Blue"
           value: "blue"
+  with_custom_size_for_page_header_and_caption:
+    description: |
+      If a custom heading size is passed in, the page heading and caption size will be changed to set size while still retaining a heading level.
```
andysellick, line 128: I'd suggest we agree on a description based on https://github.com/alphagov/govuk_publishing_components/pull/1616#discussion_r459408223 and make this the same.

app/views/govuk_publishing_components/components/docs/checkboxes.yml
```diff
@@ -123,6 +123,22 @@ examples:
...
+    data:
+      name: "favourite_skittle"
+      heading: "Choose your favourite skittles"
+      heading_caption: "Question 3 of 9"
+      heading_size: l
+      is_page_heading: true
```
andysellick, line 134: I think having the `is_page_heading` option here detracts from the point of the example. The interaction of `is_page_heading` and `heading_size` is a little complicated, I'd suggest it be explained in the description instead.

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
```diff
@@ -16,7 +16,8 @@ class CheckboxesHelper
                   :description,
                   :heading_size,
```
andysellick, line 17: If `heading_size` is only being used to capture a passed option and `@size` is replacing it in the code, I don't think it needs to be in the `attr_reader` list anymore.

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
DilwoarH: Split this up into a separate method

app/views/govuk_publishing_components/components/docs/checkboxes.yml
DilwoarH: Updated to match the radio and removed additional example from docs

lib/govuk_publishing_components/presenters/checkboxes_helper.rb
DilwoarH: We use this in the size method now, should we still remove it?

andysellick: Yes, I think so... prefixing it with the `@` makes it a class level variable, so it can be accessed anywhere in the class (as you're doing). Adding it to the `attr_reader` list means it can be accessed from outside the class, which I don't think you're doing anymore? So I think it can be removed, give it a try.

Expert ruby devs, please be gentle if I've said any of the above wrong...

DilwoarH: I've removed it...

tests pass so I'm merging

PR #1625, opened 2020-07-24
Extend auditing

andysellick:
## What
Extend the auditing code to check applications to find any instances of `gem-c-` classes in stylesheets or code, and raise a warning in the auditing page if found.

## Why
This will help us spot instances of component style overrides (which can break component isolation) or hard coded component markup (which can leave pages broken if the component is changed).

## Visual Changes

Any instances found are added to the existing list of warnings.

Screenshot 2020-07-24 at 08 54 08

All instances are also listed after the warnings.

Screenshot 2020-07-24 at 08 54 22

The code checks all the places that components are currently searched in (stylesheets, templates, javascripts, ruby code) but doesn't raise a warning if the instance is a javascript file, as attaching events to components on an application level seems like it might not cause problems? Happy to discuss this approach though.

### DIFF
--- a/app/views/govuk_publishing_components/audit/show.html.erb
+++ b/app/views/govuk_publishing_components/audit/show.html.erb
@@ -20,7 +20,11 @@
       </li>
     </ul>
     <div class="govuk-tabs__panel" id="applications">
-      <h2 class="govuk-heading-l">Applications</h2>
+      <%= render "govuk_publishing_components/components/heading", {
+        text: "Applications",
+        font_size: "l",
+        margin_bottom: 6
+      } %>
 
       <% if @applications.any? %>
         <details class="govuk-details" data-module="govuk-details">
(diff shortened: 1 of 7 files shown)


alex-ju:
As I haven't got the chance to use the auditing tool in practice, I tagged our fellow front-end developers to understand if they share the view and find the functionality useful.

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -84,6 +84,16 @@
                       </div>
                     <% end %>
                   </dl>
+
+                  <% if application[:gem_style_references].any? %>
+                    <h3 class="govuk-heading-m">Component references</h3>
```
injms, line 89: Could a heading component be used here?

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -84,6 +84,16 @@
...
+                    <p class="govuk-body">This shows instances of `gem-c-` classes found in the application. If a reference is found in a stylesheet or in code a warning is created, as this could be a style override or hard coded component markup.</p>
+                    <ul class="govuk-list govuk-list--bullet">
```
injms, line 91: List component?

PR #1628, opened 2020-07-27
Allow requests for all domains e.g. <app>.dev.gov.uk

alex-ju:
## What
Allow requests for all domains e.g. <app>.dev.gov.uk

## Why
To allow local development and cross-browser/device testing in Browserstack

## Visual Changes
No visual changes

### DIFF
--- a/spec/dummy/config/environments/development.rb
+++ b/spec/dummy/config/environments/development.rb
@@ -12,6 +12,9 @@ Rails.application.configure do
   # Show full error reports.
   config.consider_all_requests_local = true
 
+  # Allow requests for all domains e.g. <app>.dev.gov.uk
+  config.hosts.clear
+
   # Enable/disable caching. By default caching is disabled.
   if Rails.root.join("tmp/caching-dev.txt").exist?
     config.action_controller.perform_caching = true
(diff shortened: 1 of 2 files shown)


config/environments/development.rb
```diff
@@ -0,0 +1,4 @@
+Rails.application.configure do
+  # Allow requests for all domains e.g. <app>.dev.gov.uk
+  config.hosts.clear
+end
```
kevindew, line 4: This should be in the dummy app configuration: https://github.com/alphagov/govuk_publishing_components/blob/master/spec/dummy/config/environments/development.rb 

It's a bit confusing but the code in spec is used every time this app is run in a standalone fashion.

alex-ju: Right.. I've just figured it now.

alex-ju:
I got this failing in CI with

```
1) GovukPublishingComponents::Presenters::HtmlPublicationSchema#structured_data behaves like an article
     Failure/Error: expect(structured_data["articleBody"]).to eq(page.body)

       expected: ""
            got: nil

       (compared using ==)
     # ./spec/lib/govuk_publishing_components/presenters/machine_readable/html_publication_schema_spec.rb:33:in `block (3 levels) in <top (required)>'
```

It seems to be the case with any new PR. Tried to dig into it, but nothing that clicked.

PR #1631, opened 2020-07-31
Version 21.60.1

owenatgov:
Change bar chart numbers and colours to be accessible ([PR #1608](https://github.com/alphagov/govuk_publishing_components/pull/1608))

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (21.60.0)
+    govuk_publishing_components (21.60.1)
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -337,5 +337,8 @@ DEPENDENCIES
   webmock (~> 3.8.3)
   yard
 
+RUBY VERSION
+   ruby 2.6.6p146
+
```
injms, line 342: I don't think we need the Ruby version here, as there's a `.ruby-version` file with this information in it in the repo already.
```suggestion
```

PR #1632, opened 2020-07-31
Remove ability to pass Govspeak unsanitized HTML

injms:
# What

Removes the ability for the Govspeak component to use unsanitized HTML. This changes updates the current non-breaking message to raise an errror whenever non-sanitised HTML is given to the Govspeak component.

This is a **breaking change**.

I've added (another) column to the [frontend apps spreadsheet](https://docs.google.com/spreadsheets/d/1J-0IPyr2ycPuw1IWOIoX0o9Tk9nF35D0DJdsj_ylLQE/edit#gid=0) to highlight which apps will need to be updated, which have already been updated, and which don't use Govspeak at all.

## Why

This was marked as deprecated more than two years ago and should be removed (or the message saying it's deprecated should be removed.)

This means that apps that have:

```erb
<%= render "govuk_publishing_components/components/govspeak", {
  content: some_html_in_a_variable_from_elsewhere.html_safe,
} %>
```

should be changed to:

```erb
<%= render "govuk_publishing_components/components/govspeak", {} do %>
  <%= sanitize(some_html_in_a_variable_from_elsewhere) %>
<% end %>
```

or 

```erb
<%= render "govuk_publishing_components/components/govspeak", {} do %>
  <p>This is some body text with <a href=#>a link</a></p>
<% end %>
```

See https://github.com/alphagov/frontend/pull/2444 for an example of these changes.

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * **BREAKING** Layout header component always displays product name and environment when provided ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
 * Add heading option to panel component ([PR #1741](https://github.com/alphagov/govuk_publishing_components/pull/1741)) MINOR
 * **BREAKING** Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734))
+* **BREAKING** Remove ability to pass Govspeak unsanitized HTML ([PR #1632](https://github.com/alphagov/govuk_publishing_components/pull/1632))
 
 ## 21.69.0
 
--- a/app/views/govuk_publishing_components/components/_govspeak.html.erb
+++ b/app/views/govuk_publishing_components/components/_govspeak.html.erb
@@ -12,22 +12,18 @@
     <% if content.html_safe? %>
       <%= content %>
     <% else %>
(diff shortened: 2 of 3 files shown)


andysellick (approved):

andysellick (approved):
Oh wait nearly forgot... need to update the changelog.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Add heading option to panel component ([PR #1741](https://github.com/alphagov/govuk_publishing_components/pull/1741)) MINOR
 * Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734)) BREAKING
+* BREAKING: Remove ability to pass Govspeak unsanitized HTML ([PR #1632](https://github.com/alphagov/govuk_publishing_components/pull/1632)) BREAKING
```
andysellick, line 14: The established convention is to mark breaking changes with `**BREAKING**` at the start. I don't think this line needs it at the end as well. If you look at the latest version the change on line 13 has been updated to match this.

andysellick (approved):
Once again, I approve this PR 👍 

(just a nitpick over the changelog to address)

PR #1633, opened 2020-08-03
Mark social sharing images as aria-hidden

sihugh:
To fix after feedback from the [gov.uk/transition accessibility check](https://docs.google.com/presentation/d/1yDr23ID0BB8BhCBW4QpWfjPPZ1ti0u6oXCe_GTLj8_I/edit)

The social sharing images are decorative only, so should be hidden from screenreaders.

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

https://trello.com/c/P9rMz4eW/354-round-one-of-accessibility-fixes


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Mark social sharing images as aria-hidden ([PR #1633](https://github.com/alphagov/govuk_publishing_components/pull/1633))
```
injms, line 12: I don't think we need to mention this in the changeling since it's an internal change and it's documented in the commit log.

sihugh: 👍

PR #1634, opened 2020-08-04
Add optional lang parameter to image card component

danacotoran:
## What
Introduce the ability to pass an (optional) lang parameter to specify language on the image card component.

## Why
This is necessary due to the fact that there are instances on the site where a page will be in Welsh, but certain components on the page will display data in English as there is no Welsh translation for those parts of the page. This is a Web Accessibility fail.

A concrete example is [here](https://www.gov.uk/government/organisations/nuclear-decommissioning-authority.cy)  where even though the document is a Welsh translation, the ministers information is all in English.

----------

Goes hand in hand with https://github.com/alphagov/collections/pull/1853

https://trello.com/c/4znXL1Bn

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/image_card.yml
+++ b/app/views/govuk_publishing_components/components/docs/image_card.yml
@@ -9,6 +9,7 @@ accessibility_criteria: |
 
   - include alt text for images when present
   - not have duplicate links for the image and the text
+  - if the contents of the component are in a different language than the rest of the document, include an appropriate `lang` attribute to correctly identify the language used in the component
 shared_accessibility_criteria:
   - link
 examples:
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/image_card.yml
```diff
@@ -204,3 +205,12 @@ examples:
         }
       ]
       extra_links_no_indent: true
+  with_lang:
+    description: |
+      Pass through an appropriate `lang` to set a HTML lang attribute for the component.
```
injms, line 210: Do you think it's worth adding what an appropriate `lang` attribute should be?

```suggestion
      Pass through an appropriate `lang` to set a HTML lang attribute for the component.
      
      The `lang` attribute **must** be set to a [valid BCP47 string](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/lang#Language_tag_syntax). A valid code can be the two or three letter language code - for example, English is `en` or `eng`, Korean is `ko` or `kor` - but if in doubt please check.
```

This was pinched from https://github.com/alphagov/govuk_publishing_components/commit/bfd89eef898d5c67b85dea1b6955a37288c47916#diff-745eea565982ff38d9c748f6a3c85e9eR31.

danacotoran:
Thanks for this @injms I've updated the docs with your suggestion!

You are right, it's best to be specific 👍

PR #1636, opened 2020-08-05
Update all gems

pudiva:


Gemfile.lock
```diff
@@ -337,5 +337,8 @@ DEPENDENCIES
   webmock (~> 3.8.3)
   yard
 
+RUBY VERSION
+   ruby 2.6.6p146
+
```
pixeltrix, line 342: Not sure how this ended up being like this - other Gemfile.lock entries don't appear to have the patch level:

https://github.com/alphagov/service-manual-frontend/blob/master/Gemfile.lock#L372

pudinha: this one didn't even have the `RUBY VERSION` before...

pudinha: Someone added the same: 34b1c188a837a1a2f90e8144b94eba0f93a5bbd1
but then removed it: 717c4ec920ebb7523d297991f79dedbc8cf7a095

pudinha: it was added here: 855fd21d20a651249582f982c59b1ac04dda5136
changed here: 16416b9d48e16bb138994439347e3c0a30b23650
removed here: 47a58202979089e41a3641e4349027248c5a70c4
...
not sure what is going on

pixeltrix: Is there a .bundle/config file in your project directory?

pudinha: nope. no `./.bundle` nor `./bundle`

pudinha:
```
$ find . -iname '*bundle*'
./node_modules/rxjs/bundles
./node_modules/ajv/dist/ajv.bundle.js
./node_modules/ajv/scripts/bundle.js
```

pudinha: I wonder if it has to do with dependabot?

pudinha: Maybe related:
https://stackoverflow.com/questions/37903710/bundler-is-removing-ruby-version-from-gemfile-lock

pudinha: This repo here ignores Gemfile.lock completely:
https://github.com/alphagov/govuk_app_config/blame/master/.gitignore

pudinha: Should gems have Gemfile.lock checked?
https://yehudakatz.com/2010/12/16/clarifying-the-roles-of-the-gemspec-and-gemfile/

injms: The GOV.UK frontend developers discussed this and we think we should keep the lock file in, as this repo is both a gem and an application. The lock file will help us frontend developers when updating the components whilst running the application locally and should help avoid the 'works on my machine' problem.

The ruby version should be in the `.ruby-version` file only - I'd guess that it being added in and out again in the lock file is a mistake when updating things locally and committing them by accident. 

Thank you for taking a look at this - it's very much appreciated 🙏

PR #1639, opened 2020-08-10
[DO NOT MERGE] Feedback component lang

danacotoran:
**EDIT:** needs a bit of looking into, to investigate if the issue would be better fixed at a localised application level vs at a component level;  please don't review/merge this yet

## What

This adds `lang="en"` to the feedback component to correctly indicate the language the component is in. This is subject to change if/when we end up translating this component into other languages.

## Why 
As of right now, the feedback component is not localised.

On some translated pages, most of the content will be in the specified language, however the feedback component is in English and not clearly marked as such. This is a WCAG fail.

Example page: https://www.gov.uk/government/organisations/nuclear-decommissioning-authority.cy 

The content on this page is in Welsh, marked with `lang="cy"`.
The feedback component is in English and sits inside the  `lang="cy"` wrapper. 
In the absence of a relevant Welsh translation, the second best solution is to add `lang="en"` to the feedback component, to indicate language accurately.

### DIFF
--- a/app/views/govuk_publishing_components/components/_feedback.html.erb
+++ b/app/views/govuk_publishing_components/components/_feedback.html.erb
@@ -8,7 +8,7 @@
   path_without_pii = utf_encode(request.fullpath.gsub(email_regex, '[email]'))
 %>
 
-<div class="gem-c-feedback" data-module="feedback">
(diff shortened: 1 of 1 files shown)


danacotoran:
Closing as I opened https://github.com/alphagov/collections/pull/1863 instead

PR #1645, opened 2020-08-14
Remove jquery from step by step component

andysellick:
## What
- remove jquery from the step by step navigation component
- not removing jquery from the test as jasmine has its own jquery, but this has required some changes to make the tests run properly

## Why
We want to remove our dependency on jquery.

## Visual Changes
None.

## IE support

I've made a few changes to the code to try to support older versions of IE. The big change is adding polyfills for `indexOf` and `closest`, without which the component breaks. I've also put in non-polyfill workarounds to avoid using `textContent`, `forEach` and `prepend`, which means the code isn't perhaps as elegant as it could be, but I think this is reasonable.

We don't support IE8 but people still use it. Ideally we'd rewrite the component JS to fail gracefully on older browsers so that the component is left in a usable state (e.g. all steps expanded by default) but that's beyond the scope of this PR.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -1,4 +1,6 @@
-/* eslint-env jquery */
+//= require govuk/vendor/polyfills/Element/prototype/classList.js
+//= require ../vendor/polyfills/closest.js
+//= require ../vendor/polyfills/indexOf.js
 
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -4,429 +4,489 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  'use strict'
+  function Gemstepnav () { }
 
-  Modules.Gemstepnav = function () {
-    var actions = {} // stores text for JS appended elements 'show' and 'hide' on steps, and 'show/hide all' button
-    var rememberShownStep = false
-    var stepNavSize
-    var sessionStoreLink = 'govuk-step-nav-active-link'
-    var activeLinkClass = 'gem-c-step-nav__list-item--active'
-    var activeStepClass = 'gem-c-step-nav__step--active'
-    var activeLinkHref = '#content'
-    var uniqueId
+  Gemstepnav.prototype.start = function ($module) {
+    this.$module = $module[0]
+    this.$module.actions = {} // stores text for JS appended elements 'show' and 'hide' on steps, and 'show/hide all' button
+    this.$module.rememberShownStep = false
+    this.$module.stepNavSize = false
+    this.$module.sessionStoreLink = 'govuk-step-nav-active-link'
+    this.$module.activeLinkClass = 'gem-c-step-nav__list-item--active'
+    this.$module.activeStepClass = 'gem-c-step-nav__step--active'
+    this.$module.activeLinkHref = '#content'
+    this.$module.uniqueId = false
 
-    this.start = function ($element) {
-      // Indicate that js has worked
-      $element.addClass('gem-c-step-nav--active')
+    // Indicate that js has worked
+    this.$module.classList.add('gem-c-step-nav--active')
 
-      // Prevent FOUC, remove class hiding content
-      $element.removeClass('js-hidden')
+    // Prevent FOUC, remove class hiding content
+    this.$module.classList.remove('js-hidden')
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -4,429 +4,483 @@ window.GOVUK = window.GOVUK || {}
...
+    this.$module.activeLinkHref = '#content'
+    this.$module.uniqueId = false
+
+    // Indicate that js has worked
+    this.$module.classList.add('gem-c-step-nav--active')
+
+    // Prevent FOUC, remove class hiding content
+    this.$module.classList.remove('js-hidden')
+
+    this.$module.stepNavSize = this.$module.classList.contains('gem-c-step-nav--large') ? 'Big' : 'Small'
```
kr8n3r, line 26: does bundled gem js have access to govuk-frontent polyfills?

andysellick: That's what I'm currently looking into. The code as it is works fine in most browsers but currently working on some fixes for IE9 and below. IE8 is currently struggling with a lack of support for `indexOf`, I don't know if govuk-frontend supplies a polyfill for that or not. Had some problems with `textContent`, `forEach` and `prepend`, but I've found reasonable workarounds for those.

andysellick: Still not sure whether this has access to govuk-frontend polyfills or not, but since govuk-frontend doesn't support IE8 it doesn't have the polyfills we'd need. I have (perhaps controversially) added some, detailed in the PR description. Happy to discuss.

app/assets/javascripts/component_guide/application.js
```diff
@@ -1,2 +1,3 @@
+//= require_tree ../govuk_publishing_components/vendor/polyfills
```
alex-ju, line 1: Instead of vendoring (and maintaining a polyfill) we could use [the one from govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/vendor/polyfills/Element/prototype/closest.js). I did [a similar thing](https://github.com/alphagov/govuk_publishing_components/pull/1620/files#diff-835c4c4c8070c469c6e70cfe59357344R2) when re-working the checkboxes.js script.

```suggestion
//= require govuk/vendor/polyfills/Element/prototype/closest.js
```

andysellick: Cool, thanks for that. `govuk-frontend` doesn't seem to have a polyfill for `indexOf`, so I think I'll still need to include that.

andysellick: Updated 👍

app/assets/javascripts/component_guide/application.js
```diff
@@ -1,2 +1,4 @@
...
+//= require govuk/vendor/polyfills/Element/prototype/closest.js
```
alex-ju, line 2: This needs to be at the top of `step-by-step-nav.js` otherwise it will only be available in the component guide and not in the apps importing `step-by-step-nav`.

We also need `//= require govuk/vendor/polyfills/Element/prototype/classList.js ` as we're using it in a few places and it's not supported in IE9 and below and has partial support in 10 and 11.

andysellick: I didn't put it in `step-by-step-nav.js` because I didn't want to introduce duplication if other components went a similar way, but that's okay because of gzip?

`classList.js` is interesting - you're right, it isn't supported, but I had no errors relating to that during testing. Is it possible that the govuk-frontend polyfill is already included, which is why it was working? I guess it would be sensible to add it again though, particularly if duplication is handled?

andysellick:
Thanks @alex-ju. I'm happy to discuss the polyfill approach - I chose what seemed sensible, but I'm not 100% sure how our system handles code duplication (see my previous rambling) so would be good to chat about it.

app/assets/javascripts/component_guide/application.js
alex-ju: Yes, we benefit from `govuk-frontend` polyfills in the component guide as we're requiring all components (hence all the polyfills used in `govuk-frontend`). The duplication is not handled especially since `govuk-frontend` code already has the polyfills attached to each component. While not happy with this I am keen to carry on with this model until we move to ES6 modules, have a way to bundle our JavaScripts and benefit from tree-shaking.

andysellick: I think I'm following you, so I've moved the polyfill calls into the component itself.

Oddly, I couldn't get the `govuk-frontend` polyfill for `closest` to work. Tried it in IE8 in XP and Windows 7 to no avail. It fell over in `setActiveStepClass` at line `302`. I've re-added back in another polyfill, which seems to work. We obviously don't want the duplication, but I'm not sure where it's going wrong - the `classList` polyfill seems to be working fine.

alex-ju: 🤔 hmm.. I'll try to have a look at the `closest` polyfill from `govuk-frontend`

PR #1648, opened 2020-08-18
Add brand colour for FCDO

beccapearce:
## What
Add a new brand colour for the FCDO department. This colour is used on the organisation page in dividing lines, icons, link text colour etc. after it is selected in Whitehall.

## Why
The brand colour is used in the new organisation page. The brand colours come from [govuk-frontend](https://github.com/alphagov/govuk-frontend/), however since there may be some time before it is release again, we have
added it to the govuk_publishing_components so we can get it deployed.

[Related pull request in govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/1918#pullrequestreview-469159049)
[Trello card](https://trello.com/c/s0NySL7Q/2098-add-new-brand-colour-for-fcdo)


alex-ju (approved):
Thank you for explaining the reasoning in the commit message.

PR #1649, opened 2020-08-19
Remove jquery from toggle code

andysellick:
## What
Remove jquery from `toggle.js`. Also expands the documentation for the code.

## Why
We want to remove our dependency on jquery.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


andysellick:
I've test this with a local copy of collections (used on [the organisations page](https://www.gov.uk/government/organisations#ministerial_departments)) and it still works as expected.

andysellick:
On a related note, the only documentation for this code seems to be the code itself (with a fairly hefty comment at the start that I've sadly had to add to). Should we think about a better way of documenting our non-component-but-still-generally-available JS somewhere in the component guide?

PR #1650, opened 2020-08-19
Add links to component guide

andysellick:
## What
- make first 'how it looks' heading link to the default example in the component guide for that component
- add 'preview' link for the first example into the same heading, before 'preview all'
- add 'preview' link to the 'how it looks' heading on the default example page

## Why
I've found the lack of this navigation occasionally frustrating for quite a while now. It's not a huge problem, and I know the URLs to type, but I think this makes navigating the component guide fractionally easier.

## Visual Changes
Before | After
------ | ------
Screenshot 2020-08-19 at 11 02 18 | Screenshot 2020-08-19 at 11 02 25
Screenshot 2020-08-19 at 11 03 20 | Screenshot 2020-08-19 at 11 03 27

### DIFF
--- a/app/views/govuk_publishing_components/component_guide/example.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/example.html.erb
@@ -12,7 +12,10 @@
     <div class="component-markdown">
       <%= raw(@component_example.html_description) %>
     </div>
-    <h2 class="component-doc-h2">How it looks</h2>
+    <h2 class="component-doc-h2">
+      How it looks
+      <small>(<a href="<%= component_preview_path(@component_doc.id, @component_example.id) %>" class="govuk-link">preview</a>)</small>
+    </h2>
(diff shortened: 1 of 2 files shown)


maxgds:
Could you update the part of the link that says "preview" to "preview default"? I like the idea but don't think it's very clear what you're getting with the current implementation.

andysellick:
> Could you update the part of the link that says "preview" to "preview default"? I like the idea but don't think it's very clear what you're getting with the current implementation.

I think we should follow the convention of the other examples, which also have 'preview' links e.g.

Screenshot 2020-08-19 at 11 29 31

It's possible that the 'preview all' link isn't now in the best place following this change. Not sure where would be better to put it though.

maxgds (approved):
I'm happy for it to go in, we can tweak it later if it's confusing people.

PR #1654, opened 2020-08-20
Hide transition priority breadcrumb if content is part of Step by Step

hannako:
## What
Hide transition priority breadcrumb if content is part of Step by Step.

## Why
Because having both breadcrumbs is confusing.

### Before
Screenshot 2020-08-20 at 13 24 14

### After
Screenshot 2020-08-20 at 13 26 03


reggieb (approved):
LGTM

PR #1657, opened 2020-08-21
Remove jquery from govspeak

andysellick:
## What
Remove jquery from govspeak, and associated scripts (youtube link enhancement and bar chart enhancement).

Govspeak relies on [magna charta](https://github.com/alphagov/magna-charta), a jquery plugin, originally included as a minified copy of the code. Magna charta has now been archived, so I've copied the code into the gem and rewritten it as a module without jquery. This has included copying the original tests across from magna charta and rewriting them.

Ultimately we may want to replace magna charta altogether, but this change at least gives us a functioning, non-jquery version that we can apply fixes/changes to if required at any point.

Tests still use jquery in places but jasmine has its own jquery so we don't need to worry about that.

## Why
To reduce our dependency on jquery.

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/components/govspeak.js
```diff
@@ -2,25 +2,27 @@ window.GOVUK = window.GOVUK || {}
...
-  Modules.Govspeak = function () {
-    this.start = function ($element) {
-      if (!$element.hasClass('disable-youtube')) {
-        this.embedYoutube($element)
-      }
+  Govspeak.prototype.start = function ($module) {
+    this.$module = $module[0]
 
-      this.createBarcharts($element)
+    if (!this.$module.classList.contains('disable-youtube')) {
```
injms, line 10: I'd suggest using `indexOf` instead of `classList` here - this will have the same end result, no increase in lines of code, slightly less legibility - and gives increased (old) Internet Explorer support:

```suggestion
    if (this.$module.className.indexOf('disable-youtube') === -1) {
```

app/assets/javascripts/govuk_publishing_components/lib/govspeak/barchart-enhancement.js
```diff
@@ -10,16 +10,26 @@ window.GOVUK = window.GOVUK || {};
...
     // causes quite a mess. The not .mc-chart is because it creates a second
     // .js-barchart-table element with .mc-chart and then the
     // .js-barchart-table-init is given when we initialise
-    var $barcharts = this.$element.find('.js-barchart-table:not(.mc-chart):not(.js-barchart-table-init)')
-    $barcharts.each(function () {
-      var $table = $(this)
-      $.magnaCharta($table, { toggleText: 'Change between chart and table' })
-      $table.addClass('js-barchart-table-init')
-    })
+    // we'd use :not selectors but for IE8, previously worked before jquery removal
```
injms, line 17: I _think_ I know what this comment is trying to say - still, I found it quite confusing to understand (but I appreciate that it's the result of layers of work over time).

I think it'd be worth splitting up and putting closer to the code that's doing what this comment is describing, which I think is the `if` statement on line 23.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/barchart-enhancement.js
```diff
@@ -10,16 +10,26 @@ window.GOVUK = window.GOVUK || {};
...
+      var styles = $barchartsOriginal[j].className
+      if (styles.indexOf('mc-chart') === -1 && styles.indexOf('js-barchart-table-init') === -1) {
+        $barcharts.push($barchartsOriginal[j])
+      }
+    }
+
+    for (var i = 0; i < $barcharts.length; i++) {
+      var $table = $barcharts[i]
+      new GOVUK.Modules.MagnaCharta().start($table, { toggleText: 'Change between chart and table' })
+      $table.classList.add('js-barchart-table-init')
```
injms, line 31: Similar to previous comment about better Internet Explorer support:

```suggestion
      $table.className = $table.className + ' js-barchart-table-init'
```

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -0,0 +1,422 @@
+// non-jquery version of magna-charta: https://github.com/alphagov/magna-charta
```
injms, line 1:
```suggestion
// This is a non-jQuery version of [Magna Charta](https://github.com/alphagov/magna-charta).
```

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -0,0 +1,422 @@
...
+      autoOutdent: false,
+      outdentAll: false,
+      toggleAfter: false, // BOOL set TRUE to append the toggle link
+      returnReference: false // for testing purposes
+    }
+
+    for (var k in options) this.options[k] = options[k]
+    this.detectIEVersion()
+
+    // if it's IE7 or less, we just show the plain tables
```
injms, line 23: Leaving aside what browser GOV.UK supports, I think adding a reason helps make this more understandable.
```suggestion
    // Magna Charta doesn't work in IE7 or less - so plain tables are shown to those browsers.
```

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -0,0 +1,422 @@
...
+    this.ENABLED = !(this.ie && this.ie < 8)
```
injms, line 24: With the use of `classList` meaning that only IE 11 + is able to run this JavaScript, should this be changed to only be `ENABLED` when in IE 11 + and not IE? 

(This is complicated by IE10 not supporting conditional comments when in standards mode, so the `detectIEVersion()` function probably won't work.)

injms (changes requested):
I think that this should either:

 - use `classList` and not run at all in IE < 10
 - not use `classList` and run in IE < 10

and also

 - capitalise jQuery correctly 😛

app/assets/javascripts/govuk_publishing_components/lib/govspeak/barchart-enhancement.js
andysellick: Ah, the perils of copying someone else's code into a repository. Yeah, this comment is a bit confusing, rewording.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
andysellick: Thanks for the suggestion, have changed. Again, the perils of adding someone else's code...

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
andysellick: Strangely, this does all work in IE8+, however I think it's a weird coincidence. The step by step component JS now imports a polyfill for classlist, and I think it's being initialised before this (because I've tested it in IE8 and it's working). So I've explicitly added in the polyfill to this file as well.

andysellick:
Thanks @injms. I've gone with `use classList and explicitly include the classList polyfill, so it can run in IE8+`, which wasn't one of your options, but I'm improvising 😁

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
injms: Super!

injms (approved):
🍾

PR #1660, opened 2020-08-24
Fix panel component disjointed heading

danacotoran:
## What
On the panel component, move the optional prepended text and appended text into the same block as the `text` for the heading.

## Why

The prepended text, heading text, and appended text on the panel component should be in the same block.

The issue with the prepended and appended text being outside the heading element is that they provide context for the heading. Screen reader users might be confused by the incomplete heading and not understand it out of context when they are navigating a page according to its headings.

### Example of the issue
The panel text on https://www.gov.uk/bank-holidays for each next bank holiday is disjointed. It should be in the same block level element.
```
  <div>
    The next bank holiday in England and Wales is
  </div>
  <h2>
    31 August
  </h2>
  <div>
      Summer bank holiday
  </div>
```

https://trello.com/c/9eXF6FRF

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Remove jquery from toggle code ([PR #1649](https://github.com/alphagov/govuk_publishing_components/pull/1649))
 * Add role="alert" to cookie banner confirmation ([PR #1658](https://github.com/alphagov/govuk_publishing_components/pull/1658))
 * Remove jquery from step by step component ([PR #1645](https://github.com/alphagov/govuk_publishing_components/pull/1645))
+* Fix panel component disjointed heading ([PR #1660](https://github.com/alphagov/govuk_publishing_components/pull/1660))
 
 ## 21.63.0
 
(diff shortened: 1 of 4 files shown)


owenatgov (approved):
Looks awwwright to me

maxgds:
This is a good step forward on fixing the accessibility but I question whether this should be a h2 at all. It isn't a heading as there is no content tied to it and my guess is that historically it was styled as one because it looked good visually. Given that by its nature what appears here is always going to be in a panel it can't really ever be sentence summary of the subsequent paragraphs of content which is the aim of a heading, and small snippets of text surrounding it are kind of orphaned and meaningless by themselves. I suggest getting a second opinion (@injms ?) in case I'm an outlier on this suggestion, but have a look around for examples of the component in use (I usually do a github search for this) to see if this holds true across the board. If removing it wouldn't leave chunks of content headerless then switch out the markup. Consider a screenreader might be trying to read blocks of content at once, so div, heading, div is going to be very disjointed still. I'd suggest a paragraph with a span around what is currently the h2.

maxgds (changes requested):
Please see the comment I left, I think this might want a bit more tweaking.

maxgds (approved):
As discussed, maybe worth re-examining the use case in Bank Holidays later, but the component seems sound like this, so approving.

PR #1666, opened 2020-08-28
Display preferred super breadcrumb using query string

steveholmes:
## What

The superbreadcrumb makes use of the step_by_step_nav_header. 

The superbreadcrumb is displayed if content is tagged with at least one of the priority taxons defined in `ContentBreadcrumbsBasedOnPriority`.

This change allows the superbreadcrumb to be selected using a query string 'priority-taxon' instead of relying on a simple priority order. If the query string is not defined or not matched, it will use the original logic above.

## Why
Previously if a piece of content is tagged to more than one hub page the superbreadcrumb shown to a user is defined by a hierarchy of hubs, and isn't responsive to the hub that a use has navigated from.

This means that a user might go to a piece of content from one hub, but see the superbreadcrumb for a different hub, which is confusing, and doesn't provide a helpful journey back to the hub they've come from

## Visual Changes
None

## To Test/Verify locally
3 repos are required
this repo with the code changes in this PR
`govuk-content-schemas` has sample content `examples/publication/frontend/tagged_to_two_coronavirus_taxons.json `
`government-frontend` to display the sample content using this repo as a gem

### govuk-content-schemas
As of writing the changes are in a branch awaiting merging `add_tagged_to_two_taxons_superbreadcrumbs`
If not yet merged then this branch will need to be checked out.
Follow setup instruction and run using `bundle exec ruby app.rb`
this should now be running on port 4567

### government-frontend
Edit the `Gemfile` of `government-frontend` so that it picks up the local gem for `govuk_publishing_components`
`gem "govuk_publishing_components", path: "../govuk_publishing_components"`
Follow setup instructions.
To run using the local content store (govuk-content-schemas), run using
`PLEK_SERVICE_CONTENT_STORE_URI=http://localhost:4567/api ./startup.sh --dummy`

### Access content
The sample content should now be accessible at
`http://localhost:3090/examples/publication/tagged_to_two_coronavirus_taxons`
and displaying "Work and financial support during coronavirus" super breadcrumb

To change the supercrumbcrumb using the changes in this PR
`http://localhost:3090/examples/publication/tagged_to_two_coronavirus_taxons?priority-taxon=65666cdf-b177-4d79-9687-b9c32805e450`

The sample content should now be displaying "Support for businesses and self-employed people during coronavirus" super breadcrumb

https://trello.com/c/Kdnp9TPC/761-spike-conditional-super-breadcrumb-based-on-which-hub-a-user-has-come-from

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Superbreadcrumb can have a priority taxon selected via query param ([PR #1666](https://github.com/alphagov/govuk_publishing_components/pull/1666))
+
 ## 21.63.3
 
 * Remove jquery from govspeak ([PR #1657](https://github.com/alphagov/govuk_publishing_components/pull/1657))
--- a/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_priority.rb
+++ b/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_priority.rb
@@ -12,14 +12,15 @@ module GovukPublishingComponents
       }.freeze
 
       # Returns the highest priority taxon that has a content_id matching those in PRIORITY_TAXONS
-      def self.call(content_item)
-        new(content_item).breadcrumbs
+      def self.call(content_item, query_parameters = nil)
+        new(content_item, query_parameters).breadcrumbs
       end
 
-      attr_reader :content_item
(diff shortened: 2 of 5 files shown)


leenagupte:
@steveholmes Thanks for the detailed local testing instructions. They work like a charm! 🥇

alex-ju (approved):
Tested it locally and it works as expected; code looks good. I'm happy for this to get in, Steve!

PR #1668, opened 2020-09-03
Bump version to 21.63.3

owenatgov:
* Remove jquery from govspeak ([PR #1657](https://github.com/alphagov/govuk_publishing_components/pull/1657))
* Add h2 tag into legend by default ([PR #1665](https://github.com/alphagov/govuk_publishing_components/pull/1665)) (I forgot to add this to the changelog in the PR itself which is why it wasn't here before)


injms (approved):
:arrow_up:

PR #1670, opened 2020-09-04
Remove jQuery from select.js

andysellick:
## What
Remove jQuery from select.js

- select.js is used to trigger analytics events on select elements
- it seems to be only used by finder-frontend
- it has an option for extra options to be added to the tracking via a data attribute on the select option elements
- this option does not appear to be in use by finder-frontend, and there was no test for it, so I've added one

## Why
To reduce our dependency on jQuery.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


alex-ju (approved):
This looks good to me, Andy! Reviewed the code and had a play with it to check events. May need a rebase for the changelog.

PR #1671, opened 2020-09-04
Updates step by step nav header

injms:
## What

Changed the nav header to use a heading element instead of a div element.

## Why

This failed WCAG 2.1 SC 1.3.1 as the component acted as a heading, but wasn't marked up as a heading.

## Visual Changes

None.


owenatgov:
Changelog?

owenatgov (approved):
Besides my earlier comment, looks good to me 👍

PR #1676, opened 2020-09-09
Search label background colour fix for no JS

danacotoran:
## What

Ensure search label only has a white background when it is absolutely
positioned over the input field. 

## Why

The search label is not absolutely positioned with JS disabled which leads to the below issue:

### No JS homepage before
Screenshot 2020-09-09 at 15 03 49

### No JS homepage after
Screenshot 2020-09-09 at 15 06 04

### Before (JS disabled)
Screenshot 2020-09-09 at 18 13 12

### After (JS disabled)
Screenshot 2020-09-09 at 18 13 38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Search label background colour fix for no JS ([PR #1676](https://github.com/alphagov/govuk_publishing_components/pull/1676))
+
 ## 21.65.0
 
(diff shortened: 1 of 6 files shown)


owenatgov (approved):

andysellick (changes requested):
This change seems to have broken one of the options, this label is now invisible.

Screenshot 2020-09-09 at 15 53 15

danacotoran:
@andysellick I think that the purpose of the example in the screenshot is to exemplify usage of the search component where the input field is without a border. Do we know that having a search field without a border always implies that it is over a black background (and therefore requires light colour label text when JS is off?)

PR #1680, opened 2020-09-10
Fix search box label colour

andysellick:
## What
Fixes display of search label on blue background when JS is disabled, and the label is shown outside of the search box.

Previous work: https://github.com/alphagov/govuk_publishing_components/pull/1676 (thanks to @danacotoran)

## Why
Search label was unreadable.

## Visual Changes
Before:

Screenshot 2020-09-10 at 12 07 15

After:

Screenshot 2020-09-10 at 12 07 21

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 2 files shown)


danacotoran (approved):
👍

PR #1682, opened 2020-09-10
Mobile search toggle button

danacotoran:
## What
- Modify search toggle button to use button markup instead of an anchor tag, to more closely match the presentation and function of this element.
- Hide element when Javascript is disabled as the element is not functional without JS.
- Modify element text to more accurately represent what the element does ("Show/Hide search" depending on the state)

## Why
The button that opens/closes the layout header search form on mobile is not actually a button, but an anchor tag. Therefore, it does not behave as a button (for instance, it does not activate when using the space key)

This is semantically incorrect and cannot be easily understood by screen reader users, or used by voice control users.


owenatgov (changes requested):
Besides the js question, One issue came out of testing:

The mobile search button isn't hidden on desktop view it looks like:
Screenshot 2020-09-11 at 17 57 06

danacotoran:
Good spotting there @owenatgov! That issue should have been fixed now.

owenatgov (approved):
Looks great!

PR #1683, opened 2020-09-10
Remove jQuery from toggle-input-class JS

andysellick:
## What
Remove jQuery dependence from toggle-input-class JS script (used by the search component).

## Why
To reduce our dependency on jQuery.

## Visual Changes
None.

### DIFF
--- a/spec/javascripts/components/toggle-input-class-on-focus-spec.js
+++ b/spec/javascripts/components/toggle-input-class-on-focus-spec.js
@@ -22,12 +22,12 @@ describe('A toggle class module', function () {
     })
 
     it('applies the focus style on focus and removes it on blur', function () {
-      var searchInput = element.find('.js-class-toggle')
-      expect(searchInput.is('.focus')).toBe(false)
-      searchInput.triggerHandler('focus')
-      expect(searchInput.is('.focus')).toBe(true)
-      searchInput.triggerHandler('blur')
-      expect(searchInput.is('.focus')).toBe(false)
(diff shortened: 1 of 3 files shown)


spec/javascripts/components/toggle-input-class-on-focus-spec.js
```diff
@@ -14,20 +14,20 @@ describe('A toggle class module', function () {
...
     })
 
     it('applies the focus style on focus and removes it on blur', function () {
-      var searchInput = element.find('.js-class-toggle')
-      expect(searchInput.is('.focus')).toBe(false)
-      searchInput.triggerHandler('focus')
-      expect(searchInput.is('.focus')).toBe(true)
-      searchInput.triggerHandler('blur')
-      expect(searchInput.is('.focus')).toBe(false)
+      var searchInput = document.getElementById('test')
```
alex-ju, line 25: non-blocker: why the aditional `id` and not `document.querySelector('.js-class-toggle')` to match the same selector with the script?

alex-ju (approved):
Nice one! Only one comment regarding the reliance on `id` for testing – I may be missing something.

spec/javascripts/components/toggle-input-class-on-focus-spec.js
andysellick: Good question. No real reason, have changed, thanks. I think it maybe initially failed and I switched it to this to be certain the vanilla JS was playing correctly with the events being triggered in the test, and forgot to switch it back.

PR #1684, opened 2020-09-11
Feedback form errors frontend workaround

danacotoran:
## Why
This aims to address issues with WCAG SC: 3.3.1 and 3.3.3 for the "No" and "Is there anything wrong with this page?" feedback forms. 
When submitting an empty form, the erroring fields are not identified. [3.3.1]
When submitting an empty form, the error message is very vague. [3.3.3]

There was a previous attempt to fix this issue here https://github.com/alphagov/feedback/pull/758

Here is what happens when there is an attempt to submit the form without any data.
The generic "something went wrong" message is displayed – this is unhelpful to the user as they cannot identify what went wrong:
image

## What

When invalid data is submitted, the server responds with a `422` status.
However instead of the intended JSON response, it returns a rendered html page. The frontend defaults to the generic error message which is very vague and fails WCAG.

This is tricky to fix from the backend side right now as there is clobbering by nginx on all `422` requests. This implements a frontend workaround for the time being. If a JSON response is not present, and the status code is `422`, the frontend generates the error messages that would have otherwise been sent over from the backend. The error messages were copied from here https://github.com/alphagov/feedback/pull/758 

https://trello.com/c/2D0Dyn7k

## Visual Changes
Before: invalid email address, generic error message:
image

Before: invalid email address, more relevant error message:
Screenshot 2020-09-11 at 12 49 22

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -8,19 +8,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   Modules.Feedback = function () {
     this.start = function ($element) {
+      this.element = $element[0]
+      this.somethingIsWrongForm = this.element.querySelector('#something-is-wrong')
+      this.surveyForm = this.element.querySelector('#page-is-not-useful')
       this.$prompt = $element.find('.js-prompt')
       this.$fields = $element.find('.gem-c-feedback__form-field')
       this.$forms = $element.find('.js-feedback-form')
       this.$toggleForms = $element.find('.js-toggle-form')
       this.$closeForms = $element.find('.js-close-form')
+      this.activeForm = false
       this.$activeForm = false
       this.$pageIsUsefulButton = $element.find('.js-page-is-useful')
       this.$pageIsNotUsefulButton = $element.find('.js-page-is-not-useful')
       this.$somethingIsWrongButton = $element.find('.js-something-is-wrong')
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -103,12 +103,23 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+          '</h2>',
+          '<p>Please tell us what you were doing or what went wrong</p>'
+        ].join(''))
       }
 
       function setHiddenValuesNotUsefulForm (gaClientId) {
         var currentPathName = window.location.pathname.replace(/[^\s=?&]+(?:@|%40)[^\s=?&]+/, '[email]')
         var finalPathName = encodeURI(currentPathName)
-
+        that.$surveyForm.data('invalidInfoError', [
```
injms, line 117: Appreciate that we want to fix this quickly - how much work would it be to not use jQuery for the changes in this pull request?

danacotoran: Short answer yeah it is possible to not use jQuery but it [ends up more verbose](https://github.com/alphagov/govuk_publishing_components/pull/1684/commits/7b93e353e7106d82575b1644add14643197c9266) because everything in this module is reliant on jQuery (and now I'm itching to convert the whole thing to vanilla js but I can't do it in this PR because it's not within the scope of this work 😬)

injms (approved):
Super!

PR #1691, opened 2020-09-15
Fix magna charta span width bug

owenatgov:
## What
Amends the js for magna charta (charts for govspeak) to fix a bug on charts.

## Why
[A user has pointed out a bug on barcharts](https://www.gov.uk/government/publications/uk-house-price-index-summary-march-2020/uk-house-price-index-summary-march-2020#housing-transaction-distributions). Some investigation found that the `mc-bar-outdented` class isn't being applying where it needs to be because the `spanWidth` variable was returning NaN (because the computed style width was returning auto instead of an actual px value).

This amends the js to use `offsetWidth` instead, returning the class addition to it's intended functionality.

### Before
Screenshot 2020-09-15 at 11 12 13

### After
Screenshot 2020-09-15 at 11 12 36

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


chao-xian (approved):
Nice work debugging this, and an elegant solution.

PR #1699, opened 2020-09-18
Replace feedback component links with buttons

owenatgov:
## What
Replaces links with `role="button"` on the feedback component for actual buttons. Based on mockups by a designer.

## Why
Controls are presented as links to seeing users but as buttons to screen reader users and they function like buttons. This can cause screen readers users to struggle to find information and can make it more difficult for voice control users to activate a link/button.

### Before
**Extra small**
Screenshot 2020-09-18 at 17 48 53

**Mobile**
Screenshot 2020-09-18 at 17 49 43

**Tablet**
Screenshot 2020-09-18 at 17 50 01

**Desktop**
Screenshot 2020-09-18 at 17 50 10

### After
**Extra small**
Screenshot 2020-09-18 at 17 50 31

**Mobile**
Screenshot 2020-09-18 at 17 50 48

**Tablet**
Screenshot 2020-09-18 at 17 51 11

**Desktop**
Screenshot 2020-09-18 at 17 51 26

**Card:** https://trello.com/c/N1VFz9Fv/395-feedback-component-buttons-that-look-like-links

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -173,7 +173,7 @@
 
 .gem-c-feedback__option-list-item:first-child {
   @include govuk-media-query($from: mobile) {
-    margin-right: govuk-spacing(4); 
+    margin-right: govuk-spacing(4);
   }
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -11,7 +11,6 @@
 // hide without js
 // show with js, unless also has the js-hidden class
 .gem-c-feedback__js-show,
-.gem-c-feedback__form,
 .gem-c-feedback__prompt-success,
 .gem-c-feedback__prompt-questions,
 .gem-c-feedback__error-summary {
@@ -26,6 +25,24 @@
   }
 }
 
+// maintain table display for prompt and prompt-questions
+.js-enabled .gem-c-feedback__prompt {
+  @include govuk-media-query($from: tablet) {
+    display: table;
+  }
+}
+
+.js-enabled .gem-c-feedback__prompt-questions {
+  @include govuk-media-query($from: tablet) {
+    display: table-cell;
+  }
+}
+
+// show the default feedback form without js
+.js-enabled .gem-c-feedback__form.js-hidden {
+  display: none;
+}
+
 .gem-c-feedback__prompt-questions {
   text-align: center;
   border-bottom: 1px solid govuk-colour("white");
@@ -34,7 +51,7 @@
 
   @include govuk-media-query($from: tablet) {
     width: 50%;
-    float: left;
+    display: table-cell;
     text-align: left;
     border-bottom: 0;
   }
@@ -45,6 +62,8 @@
 
   @include govuk-media-query($from: tablet) {
     text-align: right;
+    vertical-align: bottom;
+    float: none;
   }
 }
 
@@ -53,6 +72,12 @@
   background-color: govuk-colour("blue");
   color: govuk-colour("white");
   outline: 0;
+
+  @include govuk-media-query($from: tablet) {
+    @include govuk-font(16, $weight: bold);
+    display: table;
+    width: 100%;
+  }
 }
 
 .gem-c-feedback__prompt-question,
@@ -65,13 +90,9 @@
 }
 
 .gem-c-feedback__prompt-question {
+  vertical-align: text-top;
   display: inline-block;
-
-  // There's a global h3 rule in some layouts that interferes with this component
-  margin: 0;
-
-  margin-left: govuk-spacing(4);
-  margin-right: govuk-spacing(4);
+  margin: govuk-spacing(2) govuk-spacing(4);
 
   &:focus {
     outline: 0;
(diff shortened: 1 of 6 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -65,27 +69,38 @@
...
+  margin-top: govuk-spacing(2);
+  margin-bottom: govuk-spacing(2);
 
   &:focus {
     outline: 0;
   }
 
+  @include govuk-media-query($from: mobile) {
+    display: inline-block;
+  }
```
andysellick, line 88: I know this is a draft PR, but I think this change doesn't make any difference? It's saying from mobile onwards make it inline-block, but line 68 previously said that, it just wasn't wrapped in a media query. Unless you're going to do something specific below 320px wide?

owenatgov: I don't disagree, however I was testing against small screen sizes on chrome and below 320, without this, it looks pretty rubbish. The buttons are squashed together. I normally wouldn't worry about below 320 but chrome now has a built in view for the galaxy fold, which is apparently only 280px. I can't back this up with any know-how about if this phone is even worth considering. This is something for me to check with the frontend community probably.

owenatgov: Matt Hobbs sent me this https://docs.google.com/spreadsheets/d/1g2cZ_J-gAHeXtXVFWwj9TjYOSqEDVKPCfLEWzDR_Oq4/edit#gid=1399476044 Top 500 stats for screen sizes over the past month. I don't see any screen widths of lower than 320px. With this in mind I'm happy to just move this into the default for the class. It would still be technically broken below around 300px but we don't really have a need to support below that. What do you reckon?

andysellick: I'm slightly confused - before this change, below 320px, the Yes/No buttons move onto separate lines. Don't they still do that even after being restyled to look like buttons?

owenatgov: I re-tested this and for this element specifically it doesn't make a difference 🤦  Will amend and push.

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
```diff
@@ -6,62 +6,29 @@
...
-          class: 'gem-c-feedback__prompt-link js-toggle-form js-page-is-not-useful',
-          data: {
-            'track-category' => 'yesNoFeedbackForm',
-            'track-action' => 'ffNoClick'
-          },
-          'aria-controls': 'page-is-not-useful',
-          'aria-expanded': false,
-          role: 'button',
-        } do %>
+        <button class="govuk-button govuk-button--secondary gem-c-feedback__prompt-link js-toggle-form js-page-is-not-useful" data-track-category="yesNoFeedbackForm" data-track-action="ffNoClick" aria-controls="page-is-not-useful" aria-expanded="false">
```
andysellick, line 20: This all looks sound. I think my only problem is that due to the changes the Yes/No buttons collapse onto a new line much earlier than they used to (see below). This isn't the end of the world, so I'm not certain whether it should hold this up or not.

Screenshot 2020-09-22 at 17 21 53

I've roughly prototyped a possible fix, using `display: table` on the subwrapper, and `display: table-cell` and `vertical-align: bottom` on the two wrapping half-divs. Would look like this:

Screenshot 2020-09-22 at 17 24 58

Happy to discuss.

andysellick (commented):
One small thing.

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
owenatgov: It's a bit late now and you ain't wrong, so I'd rather it be good than not. I'll have a play around tonight and see if I can make this a little tidier. I've got my sights on that big ugly space above `Is this page useful` as well.

andysellick (approved):

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
andysellick: Great, and happy to help if you fancy pairing on it when things calm down a little. In the meantime, I approve this message.

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -65,13 +69,16 @@
...
+  vertical-align: text-top;
   display: inline-block;
 
   // There's a global h3 rule in some layouts that interferes with this component
   margin: 0;
 
   margin-left: govuk-spacing(4);
   margin-right: govuk-spacing(4);
+  margin-top: govuk-spacing(2);
+  margin-bottom: govuk-spacing(2);
```
injms, line 81: Can these margins be put onto one line?
```suggestion
  margin: govuk-spacing(2) govuk-spacing(4);
```

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
owenatgov: Fixed the design for this awkward area between tablet and desktop. Let me know what you think, happy to chat for real.

andysellick: Niiiiice. It's slightly wonky on IE10/11, but I think forgivable. Certainly much better than it was 👍 

Screenshot 2020-09-23 at 12 01 58

injms (approved):

PR #1701, opened 2020-09-21
Update step by step show/hide text to be more verbose for screen readers

owenatgov:
## What
Removes the attributes from the text `show` (or `hide` if the step is open) [on the step by step nav component](https://components.publishing.service.gov.uk/component-guide/step_by_step_nav) that would hide it from assistive technologies and updates the surrounding markup to be more verbose for screen reader users. What is dictated to screen reader users before and after:

- Before: `Heading level 2, Do a thing` or, with show/hide hiding attributes removed, `heading level 2, Do a thing show`
- After: `Heading level 2, Do a thing, show this section`

## Why
A WCAG fail has been raised around the step by step nav component. Because show/hide are hidden from assistive tech, voice control users aren't able to dictate something like "click show" because their software wont be able to find it. This exposes the show/hide text to users and amends the surrounding markup so that what gets read out to screen reader users makes sense.

No visual changes.

**Card:** https://trello.com/c/auI7Wf4Y/373-step-nav-full-button-text-not-available-to-at

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -80,9 +80,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       if (!thisel.querySelectorAll('.js-toggle-link').length) {
         var span = document.createElement('span')
-        span.className = 'gem-c-step-nav__toggle-link js-toggle-link'
-        span.setAttribute('aria-hidden', true)
-        span.setAttribute('hidden', 'hidden')
+        var showHideSpan = document.createElement('span')
+        var commaSpan = document.createElement('span')
+        var thisSectionSpan = document.createElement('span')
+
+        showHideSpan.className = 'gem-c-step-nav__toggle-link js-toggle-link'
+        commaSpan.className = 'govuk-visually-hidden'
+        thisSectionSpan.className = 'govuk-visually-hidden'
+
+        commaSpan.innerHTML = ', '
+        thisSectionSpan.innerHTML = ' this section'
+
+        span.appendChild(commaSpan)
+        span.appendChild(showHideSpan)
+        span.appendChild(thisSectionSpan)
+
         thisel.querySelectorAll('.js-step-title-button')[0].appendChild(span)
       }
     }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+
+## Unreleased
+
+* Update step by step show/hide text to be more verbose for screen readers ([PR #1701](https://github.com/alphagov/govuk_publishing_components/pull/1701)) FIX
+
 ## 21.67.1
 
 * Fix IE11 height bug on image card ([PR #1713](https://github.com/alphagov/govuk_publishing_components/pull/1713))


injms (approved):
🗣️ 👍

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -80,9 +80,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-        span.setAttribute('aria-hidden', true)
-        span.setAttribute('hidden', 'hidden')
+        var showHideSpan = document.createElement('span')
+        var commaSpan = document.createElement('span')
+        var thisSectionSpan = document.createElement('span')
+
+        span.classList.add('gem-c-step-nav__toggle-link')
+        showHideSpan.classList.add('js-toggle-link')
+        commaSpan.classList.add('govuk-visually-hidden')
+        thisSectionSpan.classList.add('govuk-visually-hidden')
```
andysellick, line 90: For this line (and the three before it), you could just do `thisSectionSpan.className = 'govuk-visually-hidden'`

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -80,9 +80,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+        commaSpan.innerHTML = ', '
```
andysellick, line 92: Voiceover pronounces this comma, which is a bit odd (it doesn't do it for commas in a normal sentence). I think it's because it's in a span by itself, which isn't something we can get around. I tried other punctuation but it pronounces that too. Raising as something to ask Anika about.

andysellick (approved):
Looks good, couple of tiny things. We should get Anika to test the review app thoroughly before merging. Also you might want to update the changelog?

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
owenatgov: I'm not able to replicate this locally with voiceover but this might be a version thing, which obviously isn't an excuse. Will raise with Anika as you say.

injms: A way around this could be to add an `aria-label` attribute with what you'd like the screenreader to say. Testing that on VoiceOver on Safari it read things out without reading the comma.

owenatgov: @injms Are you thinking an empty span with an `aria-label`? Or an `aria-label` on the original span? I haven't tested this yet so will muck around.

injms: I _think_ adding it to the original span, so then the visually hidden classes wouldn't be needed.

For example: 

```html
<span aria-label="Show this section">Show</span>
```

owenatgov: From Anika:
>you mean an aria-label specifically for "show this section"? I don't know if that would fix it. It might be worth a try. Although, that would also mean Google will see "Do a thing, show" instead of "Do a thing, show this section". And I think the latter would be better for the Google case?

I played around and came up with this:

```
<span>
<span class="govuk-visually-hidden" aria-label=", ">, </span>
<span class="gem-c-step-nav__toggle-link js-toggle-link">show</span>
<span class="govuk-visually-hidden" aria-label=" this section"> this section</span>
</span>
```

I don't know if this is overkill but it works on my local. I don't know if this is an anti-pattern or not, but I wonder if this solves Andy's problem and maybe the NVDA problem that Anika raised on the card. If we think this is a bad idea, based on Anika's comments on the card, I reckon we should press on with the current setup as, to quote Anika, it's better than it was.

PR #1706, opened 2020-09-22
Add language attribute to feed box

andysellick:
## What
Add `lang="en"` to the element that can appear when the feed link is configured to toggle a box for copying and pasting the link.

Screenshot 2020-09-22 at 17 07 48

Added a new element for this to wrap the input and label only, because for some reason the text of the feed link is repeated inside the existing wrapping element, which is passed to the component and cannot be guaranteed to be English, as opposed to the rest of the text, which comes from the component and is.

## Why
Needed on pages like https://www.gov.uk/world/taiwan/news.zh-tw

## Visual Changes
None.

Trello card: https://trello.com/c/T0DDhI5x/391-update-language-attribute-in-subscription-links-component-of-world-organisation-embassy-pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,12 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add language attribute to feed box ([PR #1706](https://github.com/alphagov/govuk_publishing_components/pull/1706))
+
 ## 21.66.4
+
 * Reorder image card elements ([PR #1695](https://github.com/alphagov/govuk_publishing_components/pull/1695)) FIX
 
 ## 21.66.3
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Something maybe not for right now: subscription links include a locale option http://localhost:3212/component-guide/subscription-links/with_a_different_language Maybe some crossover could be handled with that and this?

andysellick:
> Something maybe not for right now: subscription links include a locale option http://localhost:3212/component-guide/subscription-links/with_a_different_language Maybe some crossover could be handled with that and this?

I saw, but the text of the box itself is still rendered entirely by the component, in English. At some point we could allow the component to accept all of that text from outside and a specific lang attribute to match it, but as we don't have translations for that yet I didn't want to go down that road.

PR #1707, opened 2020-09-22
Revert "Add lang option to organisation logo text"

andysellick:
Reverts alphagov/govuk_publishing_components#1700

Turns out, I don't think we need this. I was able to modify the original pages to use the correct language for the heading (in this PR https://github.com/alphagov/whitehall/pull/5819), and haven't implemented the feature created in this original PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,6 @@
 
 ## 21.66.3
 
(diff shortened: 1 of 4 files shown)


injms (approved):
If it's not needed then is a good decision 👍

PR #1710, opened 2020-09-24
Extend contents list component

danacotoran:
## What
Introduce the ability to pass through `lang` and custom title to the
Contents list component.

If a title lang is not supplied, a `lang` attribute will not be added.
Likewise, if a custom title is not specified, the component will default
to using "Contents" as a title.

## Why

Supports this piece of work:  https://github.com/alphagov/whitehall/pull/5831 
The need for this arose because in some instances (for example on world
location pages), when viewing a translation of the page, this title of
this component will be in English and not marked as such.
If a translation exists, we should use the translated string as the
title (e.g Contenu if the page is in French). Otherwise we can use the
English fallback, as long as `lang="en"` is specified.

https://trello.com/c/D2lTyikn

### DIFF
--- a/spec/components/contents_list_spec.rb
+++ b/spec/components/contents_list_spec.rb
@@ -137,6 +137,16 @@ describe "Contents list", type: :view do
     assert_select ".gem-c-contents-list__title", text: "Cynnwys"
   end
 
+  it "applies a custom title, if one is supplied" do
+    render_component(contents: contents_list, title: "On this page")
+    assert_select ".gem-c-contents-list__title", text: "On this page"
+  end
+
+  it "adds a lang attribute to the title, if a title_lang is supplied" do
+    render_component(contents: contents_list, title_lang: "fr")
+    assert_select ".gem-c-contents-list__title[lang=\"fr\"]"
+  end
+
   it "hides the title" do
     render_component(contents: nested_contents_list, hide_title: true)
     assert_select ".gem-c-contents-list__title", false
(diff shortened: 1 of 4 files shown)


spec/components/contents_list_spec.rb
```diff
@@ -137,6 +137,16 @@ def assert_tracking_link(name, value, total = 1)
     assert_select ".gem-c-contents-list__title", text: "Cynnwys"
   end
 
+  it "applies a custom title, if one is supplied" do
+    render_component(contents: contents_list, title: "On this page")
+    assert_select ".gem-c-contents-list__title", text: "On this page"
+  end
```
andysellick, line 143: This is probably an edge case, but with translations it's always worth checking... what happens if an empty string is passed for the `title` option? (I would expect it to default back to 'Contents' but looking at the code I can't be certain).

andysellick (approved):
Looks good, just one question and you need to update the changelog.

spec/components/contents_list_spec.rb
danacotoran: @andysellick it is an edge case indeed but still that was a good hunch as you spotted an issue. 
I tested and if the `title` option is an empty string, it would indeed use the empty string as the title, which was not the intended behaviour.
I amended things slightly, so now it should fall back to 'Contents' if the `title` parameter is an empty string.

andysellick: Great, nice job.

andysellick (approved):
I repeat my resounding approval of this change.

PR #1713, opened 2020-09-28
Fix IE11 height bug on image card

maxgds:
## What
The recent changes to the source order on the image card introduced a bug affecting IE11. For some reason the image wrapper grows beyond the size of the image it contains. Setting it to any value percentage fixes it, so a 100% height is applied here. It has no knock on effects on other browsers that I can see.  

## Why
Bug reported through Zendesk: https://govuk.zendesk.com/agent/tickets/4250243

### Before
Screenshot 2020-09-28 at 14 24 09

### After
Screenshot 2020-09-28 at 14 22 57

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
💅

PR #1715, opened 2020-09-28
Quick patch to fix Puma 5.0.1 issue

kevindew:
Puma 5.0.1 was released an hour ago and introduces an API
incompatibility with Capybara [1], which causes the tests to fail for
this project in CI. As this project is a Gem the contents of the
Gemfile.lock are not used to install the gem on CI [2] which means new
versions of dependencies can break builds without us intervening.

This applies a quick fix to get these tests working again while the Puma
situation is resolved. Should this be more than a simple mistake we'll
need to re-release govuk_test to cater for it.

[1]: https://github.com/puma/puma/issues/2388
[2]: https://github.com/alphagov/govuk-jenkinslib/blob/cdd76bbc0fb59e7a13ecc9dd9e6bb4518d6dfc97/vars/govuk.groovy#L654

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -177,7 +177,7 @@ GEM
       byebug (~> 11.0)
       pry (~> 0.13.0)
     public_suffix (4.0.6)
-    puma (5.0.0)
+    puma (4.3.6)
       nio4r (~> 2.0)
     rack (2.2.3)
     rack-test (1.1.0)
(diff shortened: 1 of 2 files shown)


benthorner (approved):

kevindew:
The issues behind this have been fixed in [Puma 5.0.2](https://github.com/puma/puma/releases/tag/v5.0.2) so this isn't needed anymore.

Issue #1709, opened 2020-09-23
Button-ified feedback component looks wonky on IE11


owenatgov:
Recently, the feedback component was updated to use visual buttons instead of links with `role="button"` to be compliant with WCAG 2.1 AA. PR: https://github.com/alphagov/govuk_publishing_components/pull/1699

Something that was raised as a non-blocking problem was that, in IE11, the design change looked a tiny bit off:

<img width="1096" alt="94004462-a72db880-fd94-11ea-975a-353f121dd728" src="https://user-images.githubusercontent.com/64783893/94024514-6db67680-fdaf-11ea-99dc-8fe8289df944.png">

This issue to capture the above so it doesn't get forgotten.

owenatgov:
Fixed by PR #1719

PR #1719, opened 2020-10-01
Set feedback buttons to be transparent

owenatgov:
## What
This PR:

- changes the styling of the buttons on the feedback component to be transparent with a white background
- changes the text on the "something wrong" button from `Is there anything wrong with this page?` to `There is something wrong with this page`
- [fixes a display issue on IE11](https://github.com/alphagov/govuk_publishing_components/issues/1709)

## Why
The original button re-design to the feedback component was designed and developed quite quickly so that it can be pushed through in time for the accessibility deadline. It has been pointed out that on some pages, notably pages with CTAs or forms, that these quite bright buttons are a bit distracting. This softens the buttons so that they still look like buttons but are less intrusive.

The text change on the "something wrong" button was an additional suggestion post deployment. As the interaction is now much more clearly a button, a question doesn't make as much sense as a descriptor for a user interaction. The new text is a bit more indicative of a user making a decision with their button press.

### Before
Screenshot 2020-10-01 at 09 54 09

### After
Screenshot 2020-10-01 at 09 53 57

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -90,7 +90,7 @@
 }
 
 .gem-c-feedback__prompt-question {
-  vertical-align: text-top;
+  vertical-align: top;
   display: inline-block;
   margin: govuk-spacing(2) govuk-spacing(4);
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -121,7 +121,7 @@
   &:hover {
     // backup style for browsers that don't support rgba
     background: govuk-colour("black");
-    background: rgba(govuk-colour("black"), 0.2);
+    background: rgba(govuk-colour("black"), .2);
   }
 
   @include govuk-media-query($from: mobile) {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -113,8 +113,16 @@
 
 .gem-c-feedback__prompt-link {
   @include govuk-font(19);
-  box-shadow: 0 2px 0 govuk-colour("black");
+  background: transparent;
+  box-shadow: 0 2px 0 govuk-colour("white");
+  border: 1px govuk-colour("white") solid;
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
```diff
@@ -6,18 +6,18 @@
   <div class="gem-c-feedback__prompt-questions js-prompt-questions">
     <h2 class="gem-c-feedback__prompt-question"><%= t("components.feedback.is_this_page_useful", default: "Is this page useful?") %></h2>
     <!-- Maybe button exists only to try and capture clicks by bots -->
-    <button class="govuk-button govuk-button--secondary gem-c-feedback__prompt-link" data-track-category="yesNoFeedbackForm" data-track-action="ffMaybeClick" aria-expanded="false" style="display: none" hidden="hidden" aria-hidden="true">
+    <button class="govuk-button gem-c-feedback__prompt-link" data-track-category="yesNoFeedbackForm" data-track-action="ffMaybeClick" aria-expanded="false" style="display: none" hidden="hidden" aria-hidden="true">
```
injms, line 9: Could we remove the classes on this honeypot button since it shouldn't be visible or usable?

```suggestion
    <button data-track-category="yesNoFeedbackForm" data-track-action="ffMaybeClick" aria-expanded="false" style="display: none" hidden="hidden" aria-hidden="true">
```

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -113,9 +113,15 @@
...
 .gem-c-feedback__prompt-link {
   @include govuk-font(19);
-  box-shadow: 0 2px 0 govuk-colour("black");
+  background: transparent;
+  box-shadow: 0 2px 0 govuk-colour("white");
+  border: 1px govuk-colour("white") solid;
   min-width: 100%;
 
+  &:hover {
+    background: rgba(govuk-colour("black"), .2);
```
injms, line 122: We should probably add a fallback hex colour just to be nice to browsers that don't support `rgba`. Not a blocker, since IE11 does support `rgba`.

injms (approved):
A couple of non-blocking comments. This looks good to go 🚢

owenatgov:
@injms I've actioned your comments, good spots. If you're happy with it I'll merge 👍

injms (approved):

PR #1726, opened 2020-10-08
[DO NOT MERGE] Suggested hover style for action link

chris-gds:
## What?
Draft PR - conversation starter:
There was a request from Design to update a missing hover state on the [action link](https://components.publishing.service.gov.uk/component-guide/action_link) on [corona virus landing page.](https://www.gov.uk/coronavirus)
Ticket: https://trello.com/c/mBN0xr04

## Why?
Updating the [corona virus](https://www.gov.uk/coronavirus) landing page action link with a unique hover style in complete isolation is not ideal. 
Users will have have a mix of hover styles even within the same page. For example, this also impacts [Bread Crumbs](https://components.publishing.service.gov.uk/component-guide/contextual_breadcrumbs) which is also on this page.

This prompted a wider discussion, white text (while on a coloured background) + hover styles do not exists within the Component Guide and there is [already an issue ](https://github.com/alphagov/govuk-frontend/issues/1417) raised about ensuring consistent hover styles across gov.uk

As a result Design actioned a [wider exploration piece](https://www.figma.com/file/BJBSVtwDgdZPOfQnf3kimJ/Exploration?node-id=67%3A2) to discover a colour that met AA standards and was a perceivable change within multiple contexts. The result was the solution `#DDDCDB`

Draft PR to prompt Frontend discussion for new hex (or possible variation of existing hex) to be absorbed into `govuk-frontend` or the Design System

## Visual Changes

Oct-08-2020 18-24-00

## Anything Else?

There's a couple possible implementation approaches

- Close ticket, mark do not fix (issue has already been raised)
- Update landing page insolation
- Update action link in isolation
- Update within all `publishing_components` add a new colour var to `gov-frontend`
- Add to Design System `$govuk-link-hover-white-text-colour: #DDDCDB`

PR #1728, opened 2020-10-12
Hover style for action link

chris-gds:
## What?

- Adding a (potentially global) component hover style colour for white text on a dark background - starting with action link.

There was a request from Design to update a missing hover state on the [action link](https://components.publishing.service.gov.uk/component-guide/action_link) on [corona virus landing page.](https://www.gov.uk/coronavirus)
Ticket: https://trello.com/c/mBN0xr04

## Why?
Updating the [corona virus](https://www.gov.uk/coronavirus) landing page action link with a unique hover style in complete isolation is not ideal. 
Users will have have a mix of hover styles even within the same page. For example, this also impacts [Bread Crumbs](https://components.publishing.service.gov.uk/component-guide/contextual_breadcrumbs) which is also on this page.

This prompted a wider discussion, white text (while on a coloured background) + hover styles do not exists within the Component Guide and there is [already an issue ](https://github.com/alphagov/govuk-frontend/issues/1417) raised about ensuring consistent hover styles across gov.uk

As a result Design actioned a [wider exploration piece](https://www.figma.com/file/BJBSVtwDgdZPOfQnf3kimJ/Exploration?node-id=67%3A2) to discover a colour that met AA standards and was a perceivable change within multiple contexts. The result was the solution `#DDDCDB`

## Visual Changes

Oct-08-2020 18-24-00

## Anything Else?

- Updating landing page in insolation is deemed too granular, approach to update the action link gem, prompt conversation to update wider `publishing_components` will also start parallel conversation with the Design System team and double check any blind spots with accessibility.

[Draft PR to prompt Frontend discussion](https://github.com/alphagov/govuk_publishing_components/pull/1726) 

_"Contrast accessibility requirements technically pass, although it's fairly close. Not sure what accessibility requirements for hover states are, but presumably they're required."_

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
@@ -1,3 +1,5 @@
+$gem-hover-dark-background: #dddcdb;
+
 .gem-c-action-link {
   display: table;
 
@@ -171,10 +173,18 @@
       color: govuk-colour("white");
     }
 
+    &:hover {
+      color: $gem-hover-dark-background;
+    }
+
     &:focus,
     &:active {
       text-decoration: none;
       color: govuk-colour("black");
+
+      &:hover {
+        color: govuk-colour("black");
+      }
     }
   }
 


app/assets/stylesheets/govuk_publishing_components/components/helpers/_variables.scss
```diff
@@ -25,3 +25,5 @@ $gem-focus-colour: $govuk-focus-colour;
 
 $gem-error-colour: govuk-colour("red");
 $gem-success-colour: govuk-colour("green");
+
+$gem-hover-dark-background: #dddcdb;
```
andysellick, line 29: If this is only used in the action link component right now I'd put it at the top of the action link sass - no need to make it more visible until we're using it elsewhere.

chris-gds: Makes sense, moved and updated

andysellick (approved):

PR #1729, opened 2020-10-12
Update to CHANGELOG

chris-gds:
* Adding hover styling for action link ([PR #1728](https://github.com/alphagov/govuk_publishing_components/pull/1728))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Adding hover styling for action link ([PR #1728](https://github.com/alphagov/govuk_publishing_components/pull/1728))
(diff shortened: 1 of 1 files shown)


andysellick (changes requested):
This won't update the version - it only updates the changelog. There's some other changes inbound that have just been merged, so I'd suggest modifying this PR to be to only add your change to the changelog, then raise a new PR for updating the gem. In that case you'll want to modify the changelog heading to be `Unreleased` rather than a specific version number.

See https://github.com/alphagov/govuk_publishing_components/blob/master/docs/publishing-to-rubygems.md

andysellick (approved):

chris-gds:
branch name inaccurate

PR #1736, opened 2020-10-14
Layout header component always displays product name and environment when provided

andysellick:
## What
- the product name wasn't shown in the header if the environment is 'public'
- product name defaulted to 'Publishing' even if not explicitly passed to the component
- now changed so the product name is shown regardless of the environment, but only shown if it is explicitly passed to the component

**Note that this is potentially a breaking change for the following reasons**

This change means that the component guide changes so most of the examples now don't have a product name, only the example showing the product name has one. This change may have some consequences for applications currently using this component. Specifically:

- [coronavirus vulnerable people service](https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/092e527b41aa656a822154c442c9822976ea5de5/app/views/layouts/application.html.erb#L44) are currently using it
- [collections publisher](https://github.com/alphagov/collections-publisher/blob/4895f0006085953180c63fcde09c3507cbee256d/app/views/layouts/admin_layout.html.erb#L12) is also using it (but interestingly actively blanking the product name in order not to show it, which feels like a workaround)

Thanks to @kevindew for hunting these down ☝️ 

Additionally, the [layout for admin component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb#L1) repeats the behaviour of this component, by setting the product name to 'Publishing' by default, which seems unnecessary to have in two places.

## Why
We want to use the header component in GOV.UK accounts, and the design requires the public environment with the product name shown.

## Visual Changes

The default option for the component was previously to show the product name, even though one was not explicitly passed. Now, no product name is shown unless it is explicitly passed.

- Default [before](https://components.publishing.service.gov.uk/component-guide/layout_header/default/preview) and [after](https://govuk-publis-show-produ-pxdfg7.herokuapp.com/component-guide/layout_header/default/preview)

Before | After
------- | -------
Screenshot 2020-10-15 at 09 35 42 | Screenshot 2020-10-15 at 09 36 01
Screenshot 2020-10-15 at 09 36 14 | Screenshot 2020-10-15 at 09 36 14

Trello card: https://trello.com/c/A7f6AkF3/309-accounts-home-page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ## Unreleased
 
+* **BREAKING** Layout header component always displays product name and environment when provided ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
 * Add heading option to panel component ([PR #1741](https://github.com/alphagov/govuk_publishing_components/pull/1741)) MINOR
-* Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734)) BREAKING
+* **BREAKING** Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734))
 
 ## 21.69.0
 
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -47,7 +47,6 @@
 
     <% unless omit_header %>
       <%= render "govuk_publishing_components/components/layout_header", {
-        environment: "public",
         search: show_search,
         # layout-header will always have border-bottom, unless the layout is full width
         remove_bottom_border: full_width,
--- a/app/views/govuk_publishing_components/components/_layout_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_header.html.erb
@@ -1,6 +1,6 @@
 <%
-  product_name ||= "Publishing"
-  public_environment = environment.eql?("public")
+  product_name ||= nil
+  environment ||= nil
   full_width ||= false
   search ||= false
   search_left ||= false
@@ -8,8 +8,9 @@
   remove_bottom_border ||= false
   search_left ||= false
   width_class = full_width ? "govuk-header__container--full-width" : "govuk-width-container"
+
   header_classes = %w(gem-c-layout-header govuk-header)
-  header_classes << "gem-c-layout-header--#{environment}"
+  header_classes << "gem-c-layout-header--#{environment}" if environment
   header_classes << "gem-c-layout-header--no-bottom-border" if remove_bottom_border
   header_classes << "gem-c-layout-header--search-left" if search_left
 %>
@@ -19,7 +20,7 @@
     <% if search_left %>
       <div class="govuk-grid-row govuk-!-margin-left-0 govuk-!-margin-right-0">
         <div class="gem-c-layout-header__logo govuk-grid-column-one-third-from-desktop">
-          <%= render "govuk_publishing_components/components/layout_header/header_logo", public_environment: public_environment, environment: environment, product_name: product_name %>
+          <%= render "govuk_publishing_components/components/layout_header/header_logo", environment: environment, product_name: product_name %>
         </div>
       </div>
       <div class="govuk-grid-row govuk-!-margin-left-0 govuk-!-margin-right-0">
@@ -33,7 +34,7 @@
     <% else %>
       <div class="govuk-grid-row govuk-!-margin-left-0 govuk-!-margin-right-0">
         <div class="gem-c-layout-header__logo govuk-grid-column-two-thirds">
-          <%= render "govuk_publishing_components/components/layout_header/header_logo", public_environment: public_environment, environment: environment, product_name: product_name %>
+          <%= render "govuk_publishing_components/components/layout_header/header_logo", environment: environment, product_name: product_name %>
         </div>
         <div class="govuk-header__content gem-c-header__content">
           <%= render "govuk_publishing_components/components/layout_header/navigation_items", navigation_items: navigation_items  %>
--- a/app/views/govuk_publishing_components/components/docs/layout_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/layout_header.yml
@@ -44,7 +44,6 @@ examples:
     description: This supports pages where the search appears on the left with multiple navigation links on the right, such as the [How government works](https://www.gov.uk/government/how-government-works) page
     data:
       search_left: true
-      environment: public
       navigation_items:
       - text: Departments
         href: "item-1"
@@ -71,10 +70,8 @@ examples:
     description: This is useful for pages where a large full-width banner is the first thing to appear on the page, for example, the [GOV.UK homepage](https://www.gov.uk)
     data:
       remove_bottom_border: true
-      environment: 'public'
   with_search_bar:
     data:
-      environment: 'public'
       search: true
 
 accessibility_criteria: |
--- a/app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
@@ -11,10 +11,12 @@
         GOV.UK
       </span>
     </span>
-    <% unless public_environment %>
+    <% if product_name %>
       <span class="govuk-header__product-name gem-c-header__product-name">
         <%= product_name %>
       </span>
+    <% end %>
+    <% if environment %>
       <span class="gem-c-environment-tag govuk-tag gem-c-environment-tag--<%= environment %>">
         <%= environment %>
       </span>
--- a/spec/components/layout_header_spec.rb
+++ b/spec/components/layout_header_spec.rb
@@ -11,16 +11,24 @@ describe "Layout header", type: :view do
     assert_select ".govuk-header"
   end
 
-  it "renders the header with environment modifier class" do
+  it "renders the header without environment tag if no environment is given" do
+    render_component({})
+
+    assert_select ".gem-c-environment-tag", 0
+  end
+
+  it "renders the header with environment tag and environment modifier class" do
     render_component(environment: "staging")
 
     assert_select ".govuk-header.gem-c-layout-header--staging"
+    assert_select ".gem-c-environment-tag", text: "staging"
   end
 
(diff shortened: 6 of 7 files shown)


app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -28,6 +28,12 @@ examples:
     data:
       environment: production
       product_name: Product
+  with_product_name_on_public:
+    description: By default when the environment is 'public', the product name is not shown. This option overrides that.
+    data:
+      environment: public
+      product_name: Product
+      show_product_name: true
```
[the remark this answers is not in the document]
andysellick: That's a fair comment. My concern was that I not break anything. Initially I changed the logic by simply moving the `unless` line to wrap only the 'environment' bit, but that made it suddenly appear in the '[with search left and navigation](https://components.publishing.service.gov.uk/component-guide/layout_header/with_left_search_and_navigation)' option, and I was worried that would make it appear in places where it wasn't wanted.

My knowledge of the use of this component is pretty limited though, so if going simpler as described and not introducing a new option is better, I'm happy to amend.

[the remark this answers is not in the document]
kevindew: Sure, might be worth checking with @alex-ju as he introduced the change before. I can understand your concern about not breaking things but good to see if we can avoid making the API convoluted, that may mean having to make a breaking change, but that's probably better for future maintenance and iteration.

I can see this is used on the Coronavirus services, e.g: https://github.com/alphagov/govuk-coronavirus-vulnerable-people-form/blob/092e527b41aa656a822154c442c9822976ea5de5/app/views/layouts/application.html.erb#L44 that would be affected.

I assume the initial change was made because the arguments don't cope with product_name missing and will insert "Publishing" as a name. This default ends up feeling increasingly weird as this component is used in non admin app contexts. I think only Content Publisher actually makes use of this default - and collections publisher actively tries to defeat it: https://github.com/alphagov/collections-publisher/blob/4895f0006085953180c63fcde09c3507cbee256d/app/views/layouts/admin_layout.html.erb#L12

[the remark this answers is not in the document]
andysellick: There's definitely some oddness to how this works. The thing that made me cautious in the first place was that when I first did this I changed it to simply show the product name if it was passed, but that made a product name appear in [this example](https://components.publishing.service.gov.uk/component-guide/layout_header/with_left_search_and_navigation) - even though it's not actually set, so it must be picking it up from somewhere else.

Additionally, In the `_layout_for_admin` component, the product name gets [set by default](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb#L1) to `Publishing`, which is exactly what the `_layout_header` component then [does when it's included inside the layout](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_layout_header.html.erb#L2).

I did talk to @alex-ju about this and he couldn't think of a reason not to simply show the product name if it gets set, but then I looked at the code and got worried I'd be breaking something.

This has been helpful. I'm going to change this PR. I'll simplify the logic so if a product name is passed, it gets shown.  This will mean some changes, but I'll check through the other examples you listed and make sure it's all sensible.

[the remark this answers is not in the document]
andysellick: @kevindew have modified and reworded the description of this PR to be more helpful. Do you think it's worth modifying the layout component for admin to not automatically set a product name?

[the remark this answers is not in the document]
kevindew: That's great. Yes I think applying the same approach to the admin layout makes sense too.

Would you be able to also make environment not be magical too?

As I'm something of the instigator of the breaking part of this I'm happy to do the work fixing repos that are affected by this.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Allow product name in layout header component when public ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
```
kevindew, line 12: We should flag this as a breaking change that there is no longer a default.

andysellick: Added, thanks.

kevindew:
Thanks @andysellick - I think the only thing outstanding is the magical environment behaviour in https://github.com/alphagov/govuk_publishing_components/pull/1736#discussion_r505459881

injms:
If this is going into a major release, can we add #1632 as a second breaking change? I'm currently going through the GOV.UK frontend apps fixing them make sure that they don't break when that's merged in.

alex-ju:
I like this change as it feels less magic and convoluted and returns the header component logic to the [initial state](https://github.com/alphagov/govuk_publishing_components/pull/408/files#diff-dcd30667ef936a3b5eca7af9c3abcb8f152fabe8e90961453ead4ca6ebaca561R21-R25) that matches the Design System model.

There's still a bit of twisted logic with the `public_environment` which has not much to do with the environment (dev, integration, staging, production) but more to where the header component is being used (public layout, admin layout) – not part of this change, but something for us to think about before rolling the component out across the applications and since we're preparing a breaking change we should think about bundling them together (as Ian mentioned in regards to #1632).

app/views/govuk_publishing_components/components/docs/layout_header.yml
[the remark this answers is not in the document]
andysellick: Okay, I _think_ I've understood the code and implemented the changes correctly, can you take a look please @kevindew?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -47,7 +48,7 @@
 
     <% unless omit_header %>
       <%= render "govuk_publishing_components/components/layout_header", {
-        environment: "public",
+        environment: environment,
```
kevindew, line 51: Do we anticipate anyone needing this argument in a public setting? I'm expecting not so I'm wondering if we want to just remove this argument until a need arises.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -1,15 +1,17 @@
 <%
-  product_name ||= "Publishing"
-  public_environment = environment.eql?("public")
+  product_name ||= false
+  environment ||= false
```
kevindew, line 3: Minor I feel like nil is better representative than false for these. False seems to apply they could be true which would be wrong, whereas nil encapsulates a lack of value.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -1,15 +1,17 @@
...
+  environment_branding = !environment ? "public" : environment
```
kevindew, line 4: Does this branding have an impact if there isn't an environment? I also notice we don't have a class defined for `gem-c-layout-header--public` so I'm wondering if instead of this line we want to change line 14 to be `header_classes << "gem-c-layout-header--#{environment}" if environment`

kevindew (commented):
Looks great @andysellick thanks for your work on this, it seems much more logical. I've just got a few minor comments but it looks pretty much ready.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
andysellick: Good point, have removed, thanks.

app/views/govuk_publishing_components/components/_layout_header.html.erb
andysellick: I'd wondered about this! Thanks for the clarity. I felt that `nil` might have been better but couldn't verbalise it properly. Change.

app/views/govuk_publishing_components/components/_layout_header.html.erb
andysellick: Great suggestion, have implemented, thanks.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* **BREAKING** Allow product name in layout header component when public ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
 * Add heading option to panel component ([PR #1741](https://github.com/alphagov/govuk_publishing_components/pull/1741)) MINOR
 * Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734)) BREAKING
```
kevindew, line 14: Do we want to adjust this to be consistent with the `**BREAKING**` formatting used in your change note?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* **BREAKING** Allow product name in layout header component when public ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
```
kevindew, line 12: I wonder if this doesn't quite reflect the breaking nature of this change - which are what I'd want to get my attention when upgrading. Perhaps: Layout header component always displays product name and environment when provided ?

kevindew (approved):
Super code looks great, just added a minor point on the changelog but shouldn't block you merging.

CHANGELOG.md
andysellick: Think I follow, have updated.

CHANGELOG.md
andysellick: Good suggestion, thanks. Was a bit of a hangover from the original intent of the PR, have changed.

CHANGELOG.md
kevindew: Sorry I wasn't specific enough - I'd go the other way and append with `**BREAKING**`e.g `* **BREAKING** Force contents list title to always be Contents or regional equivalent ([PR #1734](https://github.com/alphagov/govuk_publishing_components/pull/1734))` since this is consistent with existing breaking changes: eg: [1](https://github.com/alphagov/govuk_publishing_components/blame/3d5d4013fde6c858b9e4e1566a34e256305b046a/CHANGELOG.md#L647) [2](https://github.com/alphagov/govuk_publishing_components/blame/3d5d4013fde6c858b9e4e1566a34e256305b046a/CHANGELOG.md#L695)

andysellick: Ah yes, of course. Done, thanks.

kevindew (approved):
Awesome 👍

PR #1739, opened 2020-10-16
Update action link component to support simple light arrow

alex-ju:
## What
Update the action link component to support simple light arrow – similar to simple icon, but to be used on dark backgrounds.

## Why
Required by design to be used in the global banner.
[Trello card](https://trello.com/c/vNG432Ms)

## Visual Changes
https://govuk-publis-update-act-mpsfl6.herokuapp.com/component-guide/action_link/simple_light_arrow


sihugh (approved):
Looks very similar to the `simple` one 👍

PR #1742, opened 2020-10-19
Fix print lint attributes from being overwritten

injms:
## What

Fix to prevent the `data` attributes on the print link from being overwritten.

## Why

The `data-` attributes weren't being correctly added to the component - this was because the data attribute in the template was being overwritten by setting a new `data` parameter rather than adding to the existing data parameter.

## Visual Changes

None.


andysellick (commented):
Looks good, but does this need a test?

injms:
Closing, as fixed in https://github.com/alphagov/govuk_publishing_components/pull/1753

PR #1745, opened 2020-10-21
Add analytics from static

andysellick:
## What
Add a copy of the analytics code from `static` into the components gem. This is basically the same code as in static (originally in `govuk_frontend_toolkit`) but with tweaks to make the tests pass.

It's also not in use yet, so hasn't been included as part of the main JS of the gem, so as to not add any unnecessary page weight to applications on GOV.UK. `static` will remain the source of analytics code for GOV.UK for now, until we're ready to migrate this fully.

Note that this version of the analytics code still contains two skipped tests, as detailed in https://github.com/alphagov/static/issues/1812

## Why
There are a couple of reasons for this change.

The first is that we need analytics code for the new accounts manager, which is a standalone application that doesn't use `static` and therefore has no analytics code of its own. A number of options were considered for this, including writing custom code (too complex and unreliable, and a duplication of effort), and using `govuk_frontend_toolkit` (well, no, because it's deprecated).

This option allows us to get the analytics code (including all the functionality we use on GOV.UK) and make small changes (for example, the cookie consent model on accounts will work differently) while continuing to improve and expand it, without effecting GOV.UK's analytics.

The second reason is that when the analytics code was originally migrated from `govuk_frontend_toolkit` to `static` (in this PR https://github.com/alphagov/static/pull/1795), the intent was that it go here, but technical barriers prevented this at the time.

## Visual Changes
None.

## Changes from the code in static

Using this list to track how this code has changed since moving from `static`.

- analytics code is no longer automatically initialised, we leave the application to do that (this might change if it proves unnecessary)
- have removed the GOV.UK specific list of cross linked domains, instead allowing the code to receive an array of them, if nothing is passed, cross domain is not initialised (GOV.UK has a specific list, accounts will not require this)

## Failing tests

There are some intermittent test failures, however they only fail when run with the rest of the tests - they always pass when run individually, which suggests the problem is related to the other tests not cleaning up properly. Since we're not using this functionality in the accounts manager I have disabled these tests. 

These tests must be fixed before we can proceed with fully migrating the analytics code from static.

- `external-link-tracker.spec.js`, specifically `GOVUK.analyticsPlugins.externalLinkTracker does not duplicate the url info if a custom dimension is not provided`. The error is `Error: Expected spy setDimension not to have been called.`, and `GOVUK.analyticsPlugins.externalLinkTracker > listens to click events on only external links`
- `ecommerce.spec.js`, specifically `Ecommerce reporter for results pages > tracks clicks with different event labels`, `Ecommerce reporter for results pages > tracks clicks with product variants` and `Ecommerce reporter for results pages > tracks clicks on search results`


andysellick:
Something I've left out, but will add, is the documentation for analytics from static. Will put that in a new commit shortly.

injms (approved):
🕵️ 👍

PR #1747, opened 2020-10-22
Add heading level option to input component

andysellick:
## What
- allows the label for the input to be wrapped in a heading tag, for situations where the input is the first or only thing on a page, and no other suitable heading is appropriate
- based on https://design-system.service.gov.uk/patterns/question-pages/

## Why
Needed for a page in accounts where the label/input is the only thing on the page.

## Visual Changes

Screenshot 2020-10-23 at 19 52 47

Trello card: https://trello.com/c/pI5hMyeb/329-create-account-flow-frontend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add heading option to input component ([PR #1747](https://github.com/alphagov/govuk_publishing_components/pull/1747)) MINOR
 * Add analytics from static ([PR #1745](https://github.com/alphagov/govuk_publishing_components/pull/1745)) MINOR
 * **BREAKING** Layout header component always displays product name and environment when provided ([PR #1736](https://github.com/alphagov/govuk_publishing_components/pull/1736))
 * Add heading option to panel component ([PR #1741](https://github.com/alphagov/govuk_publishing_components/pull/1741)) MINOR
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -1,4 +1,6 @@
 <%
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
   id ||= "input-#{SecureRandom.hex(4)}"
   value ||= nil
   type ||= "text"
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -53,10 +56,20 @@
...
-    }.merge(label.symbolize_keys) %>
+    <% label_markup = capture do %>
+      <%= render "govuk_publishing_components/components/label", {
+        html_for: id,
+        heading_size: heading_size
+      }.merge(label.symbolize_keys) %>
+    <% end %>
+
+    <% if heading_level %>
+      <%= content_tag(shared_helper.get_heading_level, class: "govuk-label-wrapper") do %>
```
alex-ju, line 67: 💯 for using `govuk-label-wrapper` and checking the output to contain it

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add heading option to input component ([PR #1747](https://github.com/alphagov/govuk_publishing_components/pull/1747)) MINOR
```
alex-ju, line 12: I suggest using 'heading level option' since we already have the heading size option.

```suggestion
* Add heading level option to input component ([PR #1747](https://github.com/alphagov/govuk_publishing_components/pull/1747)) MINOR
```

app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -132,6 +132,13 @@ examples:
         text: "What is your name?"
       name: "name"
       heading_size: "l"
+  with_label_as_heading:
```
alex-ju, line 135: Non-blocker: Whereas in the test it makes sense to check the `heading_level ` in solitude I'm rather worried that presenting this example as separated from the `heading_size` one would make developers chose one or the other instead of using them as a mix. Do you think merging the two examples would bring the code closer to a usual usage?

alex-ju (approved):

app/views/govuk_publishing_components/components/docs/input.yml
injms: +1 on this - I think that we need to be careful to keep headings looking like headings, even if the heading level and the size are separate in the parameters.

owenatgov: Very good shout. Agree that we want headings that look like headings to be headings. I actually can't think of a situation where you wouldn't want this to be the case. I'm gonna try to combine these into a single attribute to force heading levels, maybe with a fallback if a dev explicitly sets it.

PR #1749, opened 2020-10-23
Fix active state on action link

alex-ju:
## What
Fix the active state on the action link component by simplifying the styling and rely on default anchor styling for the active state (as we do for any basic links across GOV.UK and within the Design System).

## Why
Causes issues when used in the `light_text` flag.

The component used in the site-wide banner in Safari 14.0:

Screenshot 2020-10-23 at 17 04 21

Thank you @conordelahunty for reporting this!

## Visual Changes
<table>
<tr><th>Before `:active`</th><th>After `:active`</th></tr>
<tr><td valign="top">

Screenshot 2020-10-23 at 17 02 49

</td><td valign="top">

Screenshot 2020-10-23 at 17 03 25

</td></tr>
</table>

[Preview link](https://govuk-publis-fix-active-tkxbuu.herokuapp.com/component-guide/action_link/with_light_text)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add track click code from static ([PR #1751](https://github.com/alphagov/govuk_publishing_components/pull/1751))
+* Fix active state on action link ([PR #1749](https://github.com/alphagov/govuk_publishing_components/pull/1749))
 * Add heading option to input component ([PR #1747](https://github.com/alphagov/govuk_publishing_components/pull/1747)) MINOR
 * Add analytics from static ([PR #1745](https://github.com/alphagov/govuk_publishing_components/pull/1745)) MINOR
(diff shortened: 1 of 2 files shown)


maxgds:
Is there any chance this could be a breaking change? If somebody is relying on black and hasn't set anything elsewhere might they wind up with unexpected colours and poor contrast?

alex-ju:
> Is there any chance this could be a breaking change? If somebody is relying on black and hasn't set anything elsewhere might they wind up with unexpected colours and poor contrast?

Mmm, not really. The `light_text` option is meant to be used on darker backgrounds (dark blue `#0843a1`, dark grey `#272828`, and more recently on black `#0b0c0c`). In any of these scenarios, we wouldn't rely on the active state to be black `#272828` as this would result in low contrast (or no contrast as it happens on the homepage).

If the component is not used with the `light_text` flag, the link colour for the `:active` state will remain `#0b0c0c`

maxgds (approved):

PR #1752, opened 2020-10-26
Fix orange outline on global site search

danacotoran:
## What
The global header was [recently updated](https://github.com/alphagov/static/pull/2293) to use the search component instead of a hard coded search field.

This resulted in an issue where the input outline is now orange due to more specific styles from `govuk_template` conflicting with the search component styles.

This updates the component styles to be more specific in order to override the CSS from `govuk_template`.

## Why
The site search input focus style has an orange outline due to `govuk_template` styles being more specific than the search component styles.
The outline should match the focus style of the search button.

### Before
Screenshot 2020-10-26 at 14 11 39

### After
Screenshot 2020-10-26 at 14 58 44

NOTE: as of the time this PR is being created the bug is not live, we caught the issue on staging.

https://trello.com/c/RbD3om74

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -36,7 +36,6 @@ $large-input-size: 50px;
 
 .gem-c-search__input[type="search"] { // overly specific to prevent some overrides from outside
   @include govuk-font($size: 19, $line-height: (28 / 19));
-
   padding: 6px;
   margin: 0;
   width: 100%;
@@ -68,6 +67,13 @@ $large-input-size: 50px;
       // `border-width`.
       border-width: $govuk-border-width-form-element * 2;
     }
+  }
+}
+
+@include govuk-compatibility(govuk_template) {
+  //  ultra specific rule overrides focus styling from govuk_template
+  #global-header .gem-c-search__input[type="search"]:focus { // stylelint-disable selector-max-id
+    @extend .gem-c-search__input[type="search"]:focus; // stylelint-disable scss/at-extend-no-missing-placeholder
   }
 }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add custom margin support for print link ([PR #1753](https://github.com/alphagov/govuk_publishing_components/pull/1753))
+* Fix orange outline on global site search ([PR #1752](https://github.com/alphagov/govuk_publishing_components/pull/1752))
 
 ## 23.0.0
 


app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -34,7 +34,9 @@ $large-input-size: 50px;
   }
 }
 
-.gem-c-search__input[type="search"] { // overly specific to prevent some overrides from outside
+// stylelint-disable selector-max-id
+.gem-c-search__input[type="search"],
+#global-header .gem-c-search__input[type="search"] { // overly specific to prevent some overrides from outside
```
alex-ju, line 39: Awful we need this but until we get rid of `govuk_template` styles we don't have a real alternative.

alex-ju (approved):
> as of the time this PR is being created the bug is not live, we caught the issue on staging

We seem to have it in production already 😬 
Screenshot 2020-10-26 at 15 51 06

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
danacotoran: Yes unfortunately we can't remove the conflicting styles from the `govuk_template` repo itself so till we get rid of that dependency, this will have to do 😬

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -1,6 +1,21 @@
 $input-size: 40px;
 $large-input-size: 50px;
 
+%search-input-focus {
```
36degrees, line 4: I'd be tempted to avoid the placeholder class and just extend the original selector (focus n' all) – even though it means disabling another linting rule.

```scss
.gem-c-search__input[type="search"] {

  // ...

  &:focus {
    outline: $govuk-focus-width solid $govuk-focus-colour;
    // Ensure outline appears outside of the element
    outline-offset: 0;
    // Double the border by adding its width again. Use `box-shadow` for this // instead of changing `border-width`
    // Also, `outline` cannot be utilised here as it is already used for the yellow focus state.
    box-shadow: inset 0 0 0 $govuk-border-width-form-element;

    @include govuk-if-ie8 {
      // IE8 doesn't support `box-shadow` so double the border with
      // `border-width`.
      border-width: $govuk-border-width-form-element * 2;
    }
  }
}

// Override overly specific focus styles from govuk_template
@include govuk-compatibility(govuk_template) {
  #global-header .gem-c-search__input[type="search"]:focus { // stylelint-disable selector-max-id
    @extend .gem-c-search__input[type="search"]:focus; // stylelint-disable scss/at-extend-no-missing-placeholder
  }
}
```

That way the 'nice' code is untouched, and the 'hacky' override for template is all neatly contained in one place – and when (one day, when template is retired) we come to clean up the code we just need to remove the block wrapped in `govuk-compatibility` .

danacotoran: Thank you for this comment @36degrees!
Good point there: we want to contain this neatly to make it easier for future us. 

I've updated this based on your suggestion. 👍

alex-ju (approved):
Thank you for the review @36degrees, and for the work on this @danacotoran!

PR #1762, opened 2020-11-03
Change how analytics variables are passed

andysellick:
## What
Change how variables for the ga property and cross domain tracking configuration are passed to the `analyticsInit` function.

## Why
A previous change passed these variables to the `analyticsInit` function so that different applications could use different configurations. Turned out that this approach didn't work with the cookie banner, because it also calls the same function as part of its internal code, and can't access those variables.

Passing these variables from an application into multiple places is too complicated, so the solution is to define the variables as attached to the GOVUK object, so they can be referenced from anywhere in the code.

This requires some tweaks to the tests to make them pass correctly. There are still some issues with the tests that will be addressed prior to the big static move, ecommerce spec is creating errors `Cannot read property 'gaClientId' of undefined` that are only visible in the console if tests are run headless with GA tracking shown, doesn't actually cause a test failure but needs addressing.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change how analytics variables are passed ([PR #1762](https://github.com/alphagov/govuk_publishing_components/pull/1762))
+
 ## 23.2.1
 
 * Fix layout header component in IE ([PR #1760](https://github.com/alphagov/govuk_publishing_components/pull/1760))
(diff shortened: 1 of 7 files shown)


alex-ju (approved):
The change makes sense and it's always great to see tests being updated 😄. Do you think there's a chance you can add an item in the team backlog to come back and briefly document this global-variable-based approach in our docs, for future reference?

andysellick:
Thanks @alex-ju. I've created this card in our roadmap https://trello.com/c/epZ12grR/110-migrate-analytics-code-into-the-gem to include your note, and also because I realised we aren't tracking this piece of work properly yet. Feel free to add anything else I might have missed.

PR #1764, opened 2020-11-04
Fix text overflow bug in Firefox

andysellick:
## What

Fix a layout bug in Firefox that was causing single column list items to column wrap into the next column. Wasn't occurring in any other browsers, because apparently only Firefox doesn't support the `break-inside` property, so this is a bit of a workaround, but including it for other browsers just in case.

Tested in Firefox, Chrome, Edge, everything back to IE8.

## Why

The layout footer component is now being used live, in Accounts 😱 

## Visual Changes

Before: 

Screenshot 2020-11-04 at 14 26 37

After:

Screenshot 2020-11-04 at 14 26 49

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix text overflow bug in Firefox ([PR #1764](https://github.com/alphagov/govuk_publishing_components/pull/1764))
 * Tidy component filenames ([PR #1848](https://github.com/alphagov/govuk_publishing_components/pull/1848))
 * Add print styling for magna charta component ([PR #1867](https://github.com/alphagov/govuk_publishing_components/pull/1867))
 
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
The fix seems to cover browsers well. I suggest linking this to an issue in `govuk-frontend` to keep track of it.

36degrees:
`overflow: hidden` causes the focus state to be clipped to the box for `.govuk-footer__list-item`:

Screenshot 2020-11-06 at 16 52 33

andysellick:
After almost completely forgetting about this PR, I've applied a far less complex fix. I've tested it in a ton of browsers and it seems to have solved the problem properly this time, but happy for further discussion.

danacotoran (approved):
This is a simple solution and it fixes the issue 💯 

The only thing I think I should note is that changing the list items' `display` property to `inline-block` does slightly affect the overall spacing on the footer. 

https://user-images.githubusercontent.com/7116819/104608976-6e368e00-567a-11eb-9927-12ad1259c1a2.mov

I don't think it's a significant enough issue to hold this back so I'm happy to approve.

PR #1767, opened 2020-11-05
Get analytics code ready for use in production

andysellick:
## What
Make changes to the analytics code in the gem, ready for use on live GOV.UK, to replace the existing copy of the code in `static`.

- improve how analytics code is called in the tests and modify previous tests that mocked the analytics objects (causing the new analytics tests to fail)
- re-enable and fix disabled tests in external link tracker and ecommerce analytics tests
- copy across changes in analytics code in static made since the analytics code was copied into the gem

## Why
We want to get this analytics code ready for use in production.

## Visual Changes
None.

Trello cards:

- https://trello.com/c/p3UtkNtE/420-resolve-the-two-copies-of-analytics-code
- https://trello.com/c/epZ12grR/110-migrate-analytics-code-into-the-gem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Get analytics code ready for use in production ([PR #1767](https://github.com/alphagov/govuk_publishing_components/pull/1767))
+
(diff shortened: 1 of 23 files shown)


injms (approved):
Looks good to me - and nice find on fixing the buggy tests 👍

PR #1769, opened 2020-11-06
Align JavaScript modules format in components

alex-ju:
## What
Convert component scripts to follow the [established module pattern](https://github.com/alphagov/govuk_publishing_components/blob/f6a9f48c4ac7efc549ffa799199e40e220384426/docs/javascript-modules.md#module-structure) and add missing tests where needed.

While at it I realised that the `initial-focus` script, used by success and error alert components falls between the cracks and it's not being suggested for imports by our current audit tool. Hence I move it to lib so that's available to all apps until we find a better approach for sharing this script (we may even have this component from `govuk-frontend` rather soon and we won't have to worry about this).

## Why
Align JavaScript file structure and patterns.

## Visual Changes
No visual changes.


andysellick (approved):
👍

PR #1774, opened 2020-11-10
[WIP] Add column list component

owenatgov:
## What
Adds the column list component to the components list.

## Why
The component has been created to account for 2 similar patterns within govuk:

- [The gov.uk homepage](https://www.gov.uk/)
- [The data.gov.uk homepage](https://data.gov.uk/)

This will allow for consistent markup across both instances of the pattern.

Some additional benefits to this component are that on the gov.uk homepage implementation, the 3 columns are 3 separate unordered lists. This may be confusing for screen reader users as this topic block is for all intents and purposes is a single list of topics. As well as this, there is the option to extend this component to act as a "layout" manager for other existing components.

### Browser compatibility note
There is a point of contention around the use of the `columns` attribute to manage the layout of this component as the implementation has been shown to be sketchy on safari and firefox presently. There are 2 issues currently open on govuk-frontend regarding this in the context of the footer component: [issue 1](https://github.com/alphagov/govuk-frontend/issues/1889), [issue 2](https://github.com/alphagov/govuk-frontend/issues/1888). I have tested this component against multiple browsers and ~haven't encountered any issues~ will post the issues I find in the comments of this PR, however as part of the review for this component, please take the time to test across multiple browsers, in particular against accessibility criteria like screen reader and keyboard navigation.

## What it looks like
Screenshot 2020-11-10 at 16 33 20

See the component docs page on the PR deploy for more details.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
@@ -1,0 +1,20 @@
+.gem-c-column-list {
+  columns: 1;
+
+  @include govuk-media-query($from: tablet) {
+    columns: 2;
+  }
+
+  @include govuk-media-query($from: desktop) {
+    columns: 3;
+  }
+}
+
+.gem-c-column-list__item {
+  display: inline-block;
+  width: 100%;
+}
+
+.gem-c-column-list__heading {
+  margin-bottom: govuk-spacing(1);
+}
--- a/app/views/govuk_publishing_components/components/_column_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_column_list.html.erb
@@ -1,7 +1,6 @@
 <%
   items ||= []
 %>
-
 <% if items.any? %>
   <ul class="govuk-list gem-c-column-list">
     <% items.each do |item| %>
--- a/app/views/govuk_publishing_components/components/_column_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_column_list.html.erb
@@ -1,0 +1,7 @@
+<%
+  items ||= []
+%>
+
+<% if items.any? %>
+
+<% end %>
--- a/app/views/govuk_publishing_components/components/_column_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_column_list.html.erb
@@ -3,5 +3,14 @@
 %>
 
 <% if items.any? %>
-
+  <ul class="govuk-list gem-c-column-list">
+    <% items.each do |item| %>
+      <li class="gem-c-column-list__item">
+        <h3 class="govuk-heading-s gem-c-column-list__heading">
+          <%= link_to item[:heading], item[:href], class: "govuk-link" %>
+        </h3>
+        <p class="govuk-body"><%= item[:text] %></p>
+      </li>
+    <% end %>
+  </ul>
 <% end %>
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -34,6 +34,7 @@
 @import "components/file-upload";
 @import "components/govspeak-html-publication";
 @import "components/govspeak";
+@import "components/column-list";
 @import "components/heading";
 @import "components/highlight-boxes";
 @import "components/hint";
--- a/app/views/govuk_publishing_components/components/docs/column_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/column_list.yml
@@ -1,0 +1,218 @@
+name: Column list
+description: The column list component can be used to display a column of content items.
+body: |
+  The markup for this component makes use of the unordered list `columns` attribute. This allows it to be ingested like a typical list by screen readers but allowing sighted users to see an organised column layout.
+
+  The original intent of this component was to account for the topics collections on the homepages of both [GOV.UK](https://gov.uk) and [data.gov.uk](https://data.gov.uk).
+accessibility_criteria: |
+  The column list component must:
+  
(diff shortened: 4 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
```diff
@@ -0,0 +1,20 @@
+.gem-c-column-list {
+  columns: 1;
+
+  @include govuk-media-query($from: tablet) {
+    columns: 2;
```
maxgds, line 5: Is that what the data.gov.uk homepage does? Because it seems like gov.uk gives 3 columns to tablet without a problem and we probably don't want to change that.

app/views/govuk_publishing_components/components/_column_list.html.erb
```diff
@@ -0,0 +1,21 @@
+<%
+  title ||= nil
+  title_hidden ||= false
+  items ||= []
+%>
+<% if items.any? %>
+  <% if title.present? %>
+    <h2 class="<%= title_hidden ? "govuk-visually-hidden" : "govuk-heading-m" %>"><%= title %></h2>
```
maxgds, line 8: I was neutral on this when we discussed it previously but I'm now thinking this belongs outside the component. A page with hidden headings is probably going to have more than one and we don't make it a part of any other component. Additionally you might want a heading that isn't hidden, so you'd have to extend this component or deliberately not feed your data into it so you could present a heading separately. It just seems like extra overhead that doesn't contribute to what the component should be focusing on - making a list lay out horizontally.

app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
```diff
@@ -0,0 +1,20 @@
...
+  }
+
+  @include govuk-media-query($from: desktop) {
+    columns: 3;
```
maxgds, line 9: I was musing on whether this column number might benefit from being configurable and whether it might be useful for pages such as https://www.gov.uk/government/ministers#ministers-by-department - and also on whether the list items should be able to contain images. It doesn't seem appropriate though as those examples are all just flowing left to right rather than splitting into columns.

maxgds (commented):
Haven't had time for a full on review yet, but a few comments and musings added.

app/views/govuk_publishing_components/components/_column_list.html.erb
```diff
@@ -0,0 +1,21 @@
...
+  <% end %>
+
+  <ul class="govuk-list gem-c-column-list">
```
maxgds, line 11: Although we don't want to display visible numbers, it seems these lists are generally ordered deliberately. You specifically want the user to read down the column, not across the row because it's been ordered, so is ol more appropriate here?

app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
owenatgov: data.gov.uk is 3 columns all the way down, however this, in my opinion, was an oversight on my part when I built that implementation. I disagree that 3 columns on mobile isn't a problem because, whilst it just about looks ok on data.gov.uk, it's a bit of a squeeze. If that content ever gets longer it may become unreadable. On the gov.uk homepage it becomes a single column on mobile so I'll get rid of this column 2 on tablet setup.

app/views/govuk_publishing_components/components/_column_list.html.erb
owenatgov: Good shout, the title makes the component too specific. I'll scoop it out.

app/assets/stylesheets/govuk_publishing_components/components/_column-list.scss
owenatgov: Of all the possible future extensions this one makes the most sense. The example as you say it's more row based than column based but it's probably still something worth exploring.

app/views/govuk_publishing_components/components/_column_list.html.erb
owenatgov: I see your point however I don't know if this stacks for screen reader users. I remember Anika raising a couple of WCAG fails around ol's that should be ul's because there's nothing in the content to indicate order like a number or a date. Right now based on this criteria, the 2 places where this would be used don't indicate order in their content so I feel like this could open us up to a WCAG fail. I do wonder if this component becomes more wide spread if it's worth considering which tag we use for the list depending on the situation but for now I don't know if calling this an ordered list makes the most sense.

maxgds: I disagree here - they both list their contents in a predictable order: alphabetical. Not a hill I plan to die on, but worth a mention :)

owenatgov: Ahaaaa... Shows how perceptive I am. As a final act of childish defiance, I'm going to ask on the accessibility channel and get a view from the hivemind.

owenatgov: After asking the accessibility channel, here are few select responses:

>I’d say in most cases, a list of items in alphabetical order would not be an ordered list because you could reorder the list without materially changing the semantic meaning

>Your list is ‘ordered’ but only for ease of browsing, not because it means something to have it ordered (like a recipe has meaning to the order and doesn't work without it)

This therefore still seems like an unordered list to me because it's not necessary to read the first topic before the 2nd to understand the list fully. I feel like the the alphabetisation is purely an enhancement. This could change drastically depending on future uses of this component of course.

owenatgov:
Closing for now as I haven't had a chance to work on this for some time and it's clogging up the PR list. With impending work from the explore team it may also be irrelevant soon.

PR #1780, opened 2020-11-16
Fix GitHub usage link not showing for all components

alex-ju:
## What
For some unknown (and potentially unintended) reason we only show the GitHub usage link if the component has a description definition in the associated yaml. This results in some components missing the link. I fixed this by moving the link outside the conditional statement.

## Why
To have the GitHub usage link visible on all components.

## Visual Changes
<table>
<tr><td valign="top">

Screenshot 2020-11-16 at 12 46 53

</td><td valign="top">

Screenshot 2020-11-16 at 12 46 38

</td></tr>
</table>


owenatgov (approved):
Good spot 💪

PR #1781, opened 2020-11-16
Add heading level to attachment component

danacotoran:
## What
Introduce ability to override default `heading_level` on the attachment
component.

## Why
The component is currently hard coded to use `h2` which isn't very
flexible. This component is often used as a child of a section of a
page, in which case the heading level of the component should be
adjusted to reflect the hierarchy of the content.

## Visual Changes
No visual changes should occur as a result.

https://trello.com/c/KzyicaBh


maxgds (approved):
Handy!

PR #1785, opened 2020-11-19
Update yarn linters

kevindew:
## What

Most notably this PR changes this project from using [standard](https://github.com/standard/standard) to [standardx](https://github.com/standard/standardx) for JS linting. The reason for this change is to allow the editing of rules, which allows us to disable the [no-var](https://eslint.org/docs/rules/no-var) rule.

It also upgrades stylelint-config-gds from a pre-release version to the actual released version and upgrades other yarn packages.

## Why

Across GDS we've been caught off-guard by the switch to requiring `let` or `const` over `var` in our JS applications. Switching to using this approach would mean that the JavaScript we would produce consistent syntax errors in IE versions before 11. Whilst we only provide limited support for these older browsers we're not yet ready for all JS to break in them and have this dictated by our tooling.

I first explored whether it would be appropriate to replace standard with an ESLint config for GDS that used standard's rule set but disabled the no-var rule. This approach however was a bit of a pain in practice as to meet [ESLint conventions](https://eslint.org/docs/developer-guide/shareable-configs#publishing-a-shareable-config) developers had a rather verbose install process with multiple peer dependencies (for example `npm install --save-dev eslint eslint-config-gds eslint-plugin-node eslint-plugin-import eslint-plugin-promise`) which then meant a whole bunch of dependencies to keep up to date, for only one line of config (`"no-var": 0`). I also felt it opened a bunch of questions as to whether other things the linter permitted that wouldn't work in < IE11 should be disabled.

I felt that whilst it's common at GDS that we need to only disable one rule (`no-var`) it would be simplest to suggest using standardx and disable the rule. We may have to rethink this if Standard further pushes approaches that aren't suitable for < IE 11.

## Visual Changes

None

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
@@ -215,8 +215,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       return !isNaN(parseFloat(val))
     },
     stripValue: function (val) {
-      var re = new RegExp('\\,|£|%|[a-z]', 'gi')
-      return val.replace(re, '')
+      return val.replace(/,|£|%|[a-z]/gi, '')
     },
     returnMax: function (values) {
       var max = 0
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
@@ -152,7 +152,7 @@
         }
 
         .mc-value-overflow {
-          text-indent: -99999px !important;
+          text-indent: -99999px !important; // stylelint-disable-line declaration-no-important
         }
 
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -215,8 +215,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       return !isNaN(parseFloat(val))
     },
     stripValue: function (val) {
-      var re = new RegExp('\\,|£|%|[a-z]', 'gi')
-      return val.replace(re, '')
+      return val.replace(/,|£|%|[a-z]/gi, '')
```
alex-ju, line 218: We're not using any dynamic input in this regex to require a constructor, so the conversion shouldn't bite us, but I do share the 'dubious' feeling.

alex-ju (approved):

PR #1787, opened 2020-11-20
Update magna charter experience for screen reader users

owenatgov:
## What
Alters the behaviour for magna charta rendered bar charts for screen reader users.

Currently, screen reader users will only be able to interpret the table portion of the chart, regardless of if it's visible or not, and ignore the toggle button and visualised chart entirely. This PR alters the behaviour to the following:

1. Screen reader users will encounter the button first, which will dictate to them "Change to table and accessible view".
2. If they try to interact with the chart, the message: "This content is not accessible - switch to table" will be announced.
2. When they click the button, "table visible" will be announced.
3. The user can now interact with the table.
4. Clicking the link again, which will now read "Change to chart view" will swap back to chart and announce "chart visible".

This PR also amends the toggle link itself to be a button element as opposed to a link and styles the element to look like a button. This is using a replication of the [print link element](https://components.publishing.service.gov.uk/component-guide/print_link).

## Why
At the foundation, this PR is solving a WCAG failure of SC 4.1.2 because the toggle button cannot be programatically determined aka: it isn't clear what the button is for. Hiding the button and mocking a DOM of just the table for screen reader users is a risk because it impacts other assistive tech users, specifically voice control users or screen reader users who can see from interacting. The content changes along with the announcement that this solution adds communicates the intent of the button in a way that accounts for all users, meeting the success criterion.

The change to the toggle link is firstly to ensure that the toggle link is semantically correct as it behaves like a button not a link and secondly to fix a bug when screen reader users interact with this link where it will sometimes follow the hash previously applied to the link element and send the screen reader's focus to the top of the page, potentially disorienting screen reader users.

## Before
Screenshot 2021-01-07 at 17 29 03
Screenshot 2021-01-07 at 17 29 14

## After
Screenshot 2021-01-07 at 17 29 24
Screenshot 2021-01-07 at 17 29 34

[Card](https://trello.com/c/qWOALsrL/434-role-of-bar-chart-link-cannot-be-programmatically-determined)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
@@ -11,9 +11,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.options = {
       outOf: 65,
       applyOnInit: true,
-      toggleText: 'Toggle between chart and table',
       autoOutdent: false,
       outdentAll: false,
+      chartVisibleText: 'Change to table and accessible view',
+      tableVisibleText: 'Change to chart view',
+      chartAlertText: 'Chart visible',
+      tableAlertText: 'Table visible',
       toggleAfter: false, // BOOL set TRUE to append the toggle link
       returnReference: false // for testing purposes
     }
@@ -27,6 +30,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     // store a reference to the table in the object
     this.$table = $module
 
+    // a container around the graph element so that it can be targeted by screen readers, allowing us to inform screen reader users that the graph isn't accessible
+    this.$graphContainer = document.createElement('div')
+    this.$graphContainer.className = 'mc-chart-container'
+
     // lets make what will become the new graph
     this.$graph = document.createElement('div')
 
@@ -38,6 +45,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     // copy over classes from the table, and add the extra one
     this.$graph.setAttribute('class', this.$table.className)
     this.$graph.classList.add('mc-chart')
+
+    // get the id of the current chart within the page so that it can be used during the generation of the toggleLink
+    this.chartId = this.getChartId($module)
 
     // set the stacked option based on
     // giving the table a class of mc-stacked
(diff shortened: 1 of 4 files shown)


owenatgov:
@chris-gds I've actioned this point on the user journey. Very salient point, the user shouldn't have to go back and forth. This is now ready for review again.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -379,7 +407,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   MagnaCharta.prototype.insert = function () {
-    this.$table.insertAdjacentElement('afterend', this.$graph)
+    var label = document.createElement('span')
+    var labelId = 'mc-chart-not-accessible-' + this.chartId
+
+    label.innerHTML = 'This content is not accessible - switch to table'
+    label.className = 'mc-hidden'
```
chris-gds, line 414: Hey @owenatgov testing on VoiceOver works really well, all makes sense liking the button switch over. 👏🏼

I'm noticing with JAWS this particular `span` copy is not being read out so I can't hear 
`This content is not accessible - switch to table`  

I think it's because this is using `display: none` instead of something like ` govuk-visually-hidden`

owenatgov: Ahaaaaaa that's a good point. I'll amend and test 👍

owenatgov: This is a bit tricky it turns out. If I replace the `mc-hidden` on the label for the graph container with `govuk-visually-hidden` then screen readers (tested with VO) read out that label as a separate element, so the announcements go:

1. "This content is not accessible - please switch to table, group"
2. *down one* "This content is not accessible - please switch to table"
3. *down one* "Out of group, This content is not accessible - switch to table"

I've played around with this, hiding the visually hidden label inside an `aria-hidden` but I can't find a sweet spot. I personally feel like out of these probs, JAWS not reading that this content isn't accessible and VO reading the label again, The former seems like the least damaging. What do you reckon? Should I go back to the drawing board? Have I understood the issue properly?

chris-gds: Further to our conversation balancing all the variables of this element, the impact of this issue highlighted plus zooming out and thinking about magna chart + govspeak in general this is in a really good place 🚀

chris-gds (approved):

PR #1790, opened 2020-11-23
Brexit countdown language and direction fix

injms:
## What

Update the countdown component to check whether a translation is available before setting the language element; and set the direction to `ltr`.

## Why

When used on pages that are in right to left lanuages (such as Arabic) and without a translation available, the countdown displayed the number of days left in reverse. 

It was also tagged as being in the page's language even when there was no translation available, which led to the countdown displaying English but being tagged as being in another language.

The translation check is done for the "components.related_navigation.transition.title" key only as I think that it's a reasonable assumption that if the title isn't translated then the text won't be either. Ideally there would be a check for both, but for the moment this will ensure that it is fixed. If and when further languages are added then this should be reassessed. 

The text direction can be hard-coded to `ltr` because there are currently only two translations available for this text - Welsh and English. Both of these languages are left to right.

## Visual Changes

| Before | After |
| - | - |
| image | image |

### DIFF
--- a/app/views/govuk_publishing_components/components/_transition_countdown.html.erb
+++ b/app/views/govuk_publishing_components/components/_transition_countdown.html.erb
@@ -7,10 +7,11 @@
   text ||= nil
   url ||= nil
   data_attributes ||= {}
-  lang ||= 'en'
+  lang ||= "en"
+  direction ||= "ltr"
   css_classes = %w(gem-c-transition-countdown)
-  css_classes << 'gem-c-transition-countdown--cta' if url
-  css_classes << 'govuk-link' if url
+  css_classes << "gem-c-transition-countdown--cta" if url
+  css_classes << "govuk-link" if url
 %>
 
 <% countdown = capture do %>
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/_transition_countdown.html.erb
```diff
@@ -8,6 +8,7 @@
   url ||= nil
   data_attributes ||= {}
   lang ||= 'en'
+  direction ||= nil
```
maxgds, line 11: I'm in two minds about this. I can see that it's cleaner to have no direction set if there's not an explicit request for one and I can see that it's being set in another component to fix the current live bug, but it feels safer to have ltr as the default - otherwise someone might expose the same problem again by dropping this component into a page somewhere other than the related nav that turns out to run in the opposite direction.

injms: Good point - I've changed this so that the default is `ltr` on the countdown component rather than being set higher up. Ta!

brucebolt (approved):

PR #1793, opened 2020-11-24
Fix to resolve input zoom-text overlap

chris-gds:
## What?

Text bleeds outside the bounding box within the UI search input element - while zooming-text-only at 200% within Firefox, this disrupts the experience for users that have increased their text sizes (a11y)

Using existing helper to resolve. [Ticket 1](https://trello.com/c/xjR2uZnf) & [Ticket 2](https://trello.com/c/4ydEeRoq)

## Why?

As pixels are being used for layout / positioning and not relative units the element doesn't scale correctly text only zoom is used.

## Anything Else

The Design System tackles this / solves this problem by using:

```
      height: govuk-px-to-rem(34px);
    }
```

I believe this will be a short-term fix while most apps go through a migration to adopt the Design System fully.

## Visual Changes

Before

Screenshot 2020-11-23 at 19 33 16

After

Screenshot 2020-11-23 at 19 32 48

This update also impacts other uses of Search, Before:

Screenshot 2020-11-23 at 19 52 19

After 

Screenshot 2020-11-23 at 19 52 37

IE11 testing:

Screenshot 2020-11-24 at 09 08 25

IE8 testing:

Screenshot 2020-11-24 at 09 08 01

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## 23.7.2
 * Fix Brexit countdown in the contextual sidebar to use appropriate lang and text direction ([PR #1790](https://github.com/alphagov/govuk_publishing_components/pull/1790))
+* Fix to resolve input zoom-text overlap ([PR #1793](https://github.com/alphagov/govuk_publishing_components/pull/1793))
 
 ## 23.7.1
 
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
Looks hunky dory to me 👍  Just needs a rebase to master and you're all good.

PR #1794, opened 2020-11-24
Add password reveal component

andysellick:
## What
Adds a new component for a password input including a show/hide button to toggle the display of the password. Uses the input component.

Screenshot 2020-12-03 at 13 30 40

Screenshot 2020-12-03 at 13 30 49

Screenshot 2020-12-03 at 13 30 59

Screenshot 2020-12-03 at 13 31 22

## Why

We have some (limited) research suggesting it's a feature users want for the GOVUK Accounts registration flow.

## Changes

Build notes:

- the feature works by toggling the input type between `password` and `text`, which makes the text visible
- because this is a JS only feature all of the extra elements are added using JS, which means that if JS is disabled the input appears as normal
- the text for all of the content comes from data attributes appended to the DOM, which in turn comes from the translation files, so this can be translated if needed
- the button remains to the right of the input except on mobile, where it stacks, and the full text is visible

Screenshot 2020-12-03 at 13 31 35

Browser compatibility:

- the layout uses flexbox, which is only partially supported in various versions of IE, but attempting to use the `-ms-flexbox` prefixed version produced some unwanted effects, so I've gone with a display table fallback
- tested in all versions of IE and with the exception of some minor layout problems in the older versions works fine, except for IE8, which doesn't allow the `type` of an input to be changed, for some reason

IE8:

Screenshot 2020-12-03 at 13 50 48

IE9:

Screenshot 2020-12-03 at 13 51 14

IE10:

Screenshot 2020-12-03 at 13 51 33

IE11 (which for some reason is squashed against the edges of the browser, please ignore the edges):

Screenshot 2020-12-03 at 13 52 43

Accessibility:

- initially I thought the accessibility for this would be straightforward as non-sighted users wouldn't need this functionality, however I don't think that's true, as a non-sighted user might also want to check what they've typed in a password input
- the show/hide button has an `aria-controls` attribute pointing to the input
- when the button is pressed with a screenreader, I felt there should be some kind of feedback for non-sighted users to indicate that something has happened. Originally I set the input to have an `aria-live="polite"` attribute, but the result of this was that the value of the input is announced, which is a possible security issue. Instead, I've added a visually hidden bit of text that says "Your password is shown/hidden" and put the `aria-live` attribute on that instead. However, if the user shows the password and then returns the focus to the input, the password gets read aloud anyway, but at least with this method they should be aware of what's happening before their password is announced

Security:

- there's been some concern that submitting a form with a password input set to `type="text"` could cause some problems, particularly with the browser maybe remembering the password value on future visits
- input gets `autocomplete="off"` added to prevent this (but can also be set to `new-password` or `current-password` to help browsers understand how to remember passwords)
- since this isn't 100% supported by browsers, the JS also now detects parent form submission and reverts the input to a password type to further protect users
- if this is found to cause a conflict with other JS on a form, the component also accepts the `disable-form-submit-check` option to disable this code

Trello card: https://trello.com/c/aJeQWmOS/418-add-a-show-password-option-to-account-creation-flow-and-sign-in

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/show-password.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/show-password.js
@@ -1,0 +1,63 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ShowPassword () { }
+
+  ShowPassword.prototype.start = function ($module) {
+    this.$module = $module[0]
+    this.input = this.$module.querySelector('.gem-c-input')
+    this.$module.togglePassword = this.togglePassword.bind(this)
+    this.$module.revertToPasswordOnFormSubmit = this.revertToPasswordOnFormSubmit.bind(this)
+    this.input.classList.add('gem-c-input--with-password')
+
+    this.showPasswordText = this.$module.getAttribute('data-show')
+    this.hidePasswordText = this.$module.getAttribute('data-hide')
+    this.shownPassword = this.$module.getAttribute('data-announce-show')
+    this.hiddenPassword = this.$module.getAttribute('data-announce-hide')
+
+    // wrap the input in a new div
+    this.wrapper = document.createElement('div')
+    this.wrapper.classList.add('gem-c-show-password__input-wrapper')
+    this.input.parentNode.insertBefore(this.wrapper, this.input)
+    this.wrapper.appendChild(this.input)
+
+    // create and append the button
+    this.showHide = document.createElement('button')
+    this.showHide.className = 'gem-c-show-password__toggle'
+    this.showHide.setAttribute('aria-controls', this.input.getAttribute('id'))
+    this.showHide.innerHTML = this.showPasswordText
+    this.wrapper.insertBefore(this.showHide, this.input.nextSibling)
+
+    // create and append the status text for screen readers
+    this.statusText = document.createElement('span')
+    this.statusText.classList.add('govuk-visually-hidden')
+    this.statusText.innerHTML = this.hiddenPassword
+    this.statusText.setAttribute('aria-live', 'polite')
+    this.wrapper.insertBefore(this.statusText, this.showHide.nextSibling)
+
+    this.showHide.addEventListener('click', this.$module.togglePassword)
+
+    this.parentForm = this.input.form
+    var disableFormSubmitCheck = this.$module.getAttribute('data-disable-form-submit-check')
+
+    if (this.parentForm && !disableFormSubmitCheck) {
+      this.parentForm.addEventListener('submit', this.$module.revertToPasswordOnFormSubmit)
+    }
+  }
+
+  ShowPassword.prototype.togglePassword = function (event) {
+    event.preventDefault()
+    this.showHide.innerHTML = this.showHide.innerHTML === this.showPasswordText ? this.hidePasswordText : this.showPasswordText
+    this.statusText.innerHTML = this.statusText.innerHTML === this.shownPassword ? this.hiddenPassword : this.shownPassword
+    this.input.setAttribute('type', this.input.getAttribute('type') === 'text' ? 'password' : 'text')
+  }
+
+  ShowPassword.prototype.revertToPasswordOnFormSubmit = function (event) {
+    this.showHide.innerHTML = this.showPasswordText
+    this.statusText.innerHTML = this.hiddenPassword
+    this.input.setAttribute('type', 'password')
+  }
+
+  Modules.ShowPassword = ShowPassword
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
@@ -1,0 +1,61 @@
+.gem-c-show-password__input-wrapper {
+  display: table; // IE fallback
+  display: flex;
+  width: 100%;
+  flex-direction: column;
+
+  @include govuk-media-query($from: mobile) {
+    flex-direction: row;
+  }
+
+  .gem-c-input--with-password {
+    display: table-cell;
+
+    &:focus {
+      z-index: 1;
+    }
+  }
+}
+
+.gem-c-show-password__toggle {
+  @include govuk-font(19);
+  z-index: 0;
+  display: table-cell; // IE fallback
+  padding: govuk-spacing(1) govuk-spacing(3);
+  min-width: 5em; // stops the button width jumping when the text changes
+  color: $govuk-link-colour;
+  text-decoration: underline;
+  background: govuk-colour("white");
+  border: solid 2px $govuk-input-border-colour;
+  white-space: nowrap;
+  cursor: pointer;
+
+  @include govuk-media-query($until: mobile) {
+    padding: govuk-spacing(1);
+    width: 100%;
+    margin-top: -2px;
+    white-space: normal;
+  }
+
+  @include govuk-media-query($from: mobile) {
+    margin-left: -2px;
+  }
+
+  &:hover {
+    color: $govuk-link-hover-colour;
+  }
+
+  &:focus {
+    z-index: 1;
+    background: $govuk-focus-colour;
+    color: $govuk-focus-text-colour;
+    outline: 0;
+  }
+
+  &:active {
+    z-index: 1;
+    background: govuk-colour("white");
+    border-color: $govuk-focus-colour;
+    color: $govuk-link-active-colour;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_show_password.html.erb
+++ b/app/views/govuk_publishing_components/components/_show_password.html.erb
@@ -1,0 +1,40 @@
+<%
+  disable_form_submit_check ||= false
+  label ||= nil
+  value ||= nil
+  id ||= nil
+  name ||= nil
+  autocomplete = "off" unless ['new-password', 'current-password'].include?(autocomplete)
+  describedby ||= nil
+  hint ||= nil
+  error_message ||= nil
+  error_items ||= nil
+
+  if !label
+    raise ArgumentError, "This component requires a label"
+  end
+%>
+<% if label %>
+  <%= tag.div class: "gem-c-show-password",
+    data: {
+      module: "show-password",
+      disable_form_submit_check: disable_form_submit_check,
+      show: t('components.input.show'),
+      hide: t('components.input.hide'),
+      announce_show: t('components.input.announce_show'),
+      announce_hide: t('components.input.announce_hide')
+    } do %>
+      <%= render "govuk_publishing_components/components/input", {
+        label: label,
+        value: value,
+        id: id,
+        name: name,
+        describedby: describedby,
+        hint: hint,
+        error_message: error_message,
+        error_items: error_items,
+        type: "password",
+        autocomplete: autocomplete,
+      } %>
+  <% end %>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/show_password.yml
+++ b/app/views/govuk_publishing_components/components/docs/show_password.yml
@@ -1,0 +1,79 @@
+name: Show password input
+description: A password input field that allows the password to be shown.
+body: |
+  Adds a password reveal button to an input that toggles its type from password to text, revealing the password. Does not appear if JavaScript is disabled.
+
+  Uses a visually hidden bit of text to inform screen reader users of the state of the input rather than announcing the content of the password input when toggled.
+accessibility_criteria: |
+  The component must:
+
+  * indicate to the user that the password has been shown or hidden
+
+  Inputs in the component must:
+
+  * accept focus
+  * be focusable with a keyboard
+  * be usable with a keyboard
+  * be usable with touch
+  * indicate when they have focus
+  * be recognisable as form input elements
+  * have correctly associated labels
+  * be of the appropriate type for their use, e.g. password inputs should be of type 'password'
+
+  Buttons in the component must:
+
+  - accept focus
+  - be focusable with a keyboard
+  - indicate when it has focus
+  - activate when focused and space is pressed
+  - activate when focused and enter is pressed
+govuk_frontend_components:
+  - text-input
+examples:
+  default:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+  disable_form_submit_check:
+    description: The component sets the `autocomplete` attribute on the input to "off" so that browsers don't remember the password. This isn't reliable, so the component reverts the input type back to a password when its parent form is submitted. If this is causing conflicts with other JavaScript on your form, you can set `disable-form-submit-check` to `true` to disable this functionality and implement it yourself.
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      disable_form_submit_check: true
+  set_autocomplete_attribute:
+    description: By default, autocomplete is set to `off`. This can be set to `new-password` or `current-password` to help browsers differentiate between new and current passwords.
+    data:
+      label:
+        text: Please enter your password
+      autocomplete: "current-password"
+      value: this is my password
+  with_attributes:
+    description: The component accepts many but not all of the options that can be passed to a regular input, including `id`, `name` and `describedby` (not shown).
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      id: custom_id
+      name: custom_name
+  with_hint:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      hint: Your password must be at least twelve thousand characters
+  with_error:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      error_message: No it isn't
+  with_error_items:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      error_items:
+      - text: Look I keep telling you
+      - text: This isn't your password
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -54,6 +54,11 @@ en:
       what_wrong: "What went wrong?"
       send_me_survey: "Send me the survey"
       send: "Send"
+    input:
+      show: Show
+      hide: Hide
+      announce_show: Your password is shown
+      announce_hide: Your password is hidden
     organisation_schema:
       all_content_search_description: "Find all content from %{organisation}"
     radio:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add password reveal component ([PR #1794](https://github.com/alphagov/govuk_publishing_components/pull/1794))
 * Add some GOV.UK Accounts specific PII redacts ([PR #1807](https://github.com/alphagov/govuk_publishing_components/pull/1807))
 
 ## 23.7.6
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -60,6 +60,7 @@
 @import "components/search";
 @import "components/select";
 @import "components/share-links";
+@import "components/show-password";
 @import "components/skip-link";
 @import "components/step-by-step-nav-header";
 @import "components/step-by-step-nav-related";
--- a/spec/components/show_password_spec.rb
+++ b/spec/components/show_password_spec.rb
@@ -1,0 +1,51 @@
+require "rails_helper"
+
+describe "Show password", type: :view do
+  def component_name
+    "show_password"
+  end
+
+  it "fails to render when no label is given" do
+    assert_raises do
+      render_component({})
+    end
+  end
+
+  it "renders correctly" do
+    render_component(
+      label: {
+        text: "Your password",
+      },
+    )
+
+    assert_select(".gem-c-show-password[data-module='show-password']")
+    assert_select(".gem-c-show-password[data-show='Show']")
+    assert_select(".gem-c-show-password[data-hide='Hide']")
+    assert_select(".gem-c-show-password[data-announce-show='Your password is shown']")
+    assert_select(".gem-c-show-password[data-announce-hide='Your password is hidden']")
+    assert_select(".gem-c-input[autocomplete='off']")
+    assert_select(".gem-c-input[type='password']")
+  end
+
+  it "renders with autocomplete current password" do
+    render_component(
+      label: {
+        text: "Your password",
+      },
+      autocomplete: "current-password",
+    )
+
+    assert_select(".gem-c-input[autocomplete='current-password']")
+  end
+
+  it "renders with autocomplete off if invalid entry is passed for autocomplete" do
+    render_component(
+      label: {
+        text: "Your password",
+      },
+      autocomplete: "this-is-totally-not-a-thing",
+    )
+
+    assert_select(".gem-c-input[autocomplete='off']")
+  end
+end
--- a/spec/javascripts/components/show-password-spec.js
+++ b/spec/javascripts/components/show-password-spec.js
@@ -1,0 +1,123 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK */
+
+describe('A show password component', function () {
+  var element
+
+  describe('in password reveal mode', function () {
+    beforeEach(function () {
+      element = $(
+        '<div class="gem-c-show-password" data-module="show-password" data-disable-form-submit-check="false" data-show="Show" data-hide="Hide" data-announce-show="Your password is shown" data-announce-hide="Your password is hidden">' +
+          '<div class="govuk-form-group">' +
+            '<label for="input" class="gem-c-label govuk-label">Please enter your password</label>' +
+            '<input name="password" value="this is my password" class="gem-c-input govuk-input" id="input" type="password" autocomplete="off">' +
+          '</div>' +
+        '</div>'
+      )
+      new GOVUK.Modules.ShowPassword().start(element)
+    })
+
+    afterEach(function () {
+      element.remove()
+    })
+
+    it('adds the required elements for password reveal', function () {
+      expect(element.find('.gem-c-show-password__input-wrapper').length).toBe(1)
+      expect(element.find('.gem-c-input.gem-c-input--with-password').length).toBe(1)
+      expect(element.find('.gem-c-show-password__toggle').length).toBe(1)
+      expect(element.find('.gem-c-show-password__toggle').text()).toBe('Show')
+      expect(element.find('.gem-c-show-password__toggle').attr('aria-controls')).toBe('input')
+      expect(element.find('.govuk-visually-hidden').length).toBe(1)
+      expect(element.find('.govuk-visually-hidden').text()).toBe('Your password is hidden')
+    })
+
+    it('reveals the password when clicked', function () {
+      element.find('.gem-c-show-password__toggle').trigger('click')
+
+      expect(element.find('.gem-c-show-password__toggle').text()).toBe('Hide')
+      expect(element.find('input[name="password"]').attr('type')).toBe('text')
+      expect(element.find('.govuk-visually-hidden').text()).toBe('Your password is shown')
+    })
+
+    it('hides the password when clicked again', function () {
+      element.find('.gem-c-show-password__toggle').trigger('click')
+      element.find('.gem-c-show-password__toggle').trigger('click')
+
+      expect(element.find('.gem-c-show-password__toggle').text()).toBe('Show')
+      expect(element.find('input[name="password"]').attr('type')).toBe('password')
+      expect(element.find('.govuk-visually-hidden').text()).toBe('Your password is hidden')
+    })
+  })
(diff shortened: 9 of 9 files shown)


app/assets/javascripts/govuk_publishing_components/components/input.js
```diff
@@ -0,0 +1,48 @@
...
+    showHide.innerHTML = this.showPasswordText
+    this.$module.insertBefore(showHide, this.input.nextSibling)
+
+    this.statusText = document.createElement('span')
+    this.statusText.classList.add('govuk-visually-hidden')
+    this.statusText.innerHTML = this.hiddenPassword
+    this.statusText.setAttribute('aria-live', 'polite')
+    this.$module.insertBefore(this.statusText, showHide.nextSibling)
+
+    var that = this
```
owenatgov, line 30: I'm struggling to wrap my head around this approach. Why do we need to define a separate instance of `this` outside the event listener below? If this is a question of scope and the possibility of multiple show password buttons on one page, wouldn't this module handle that scoping?

spec/javascripts/components/input-spec.js
```diff
@@ -0,0 +1,44 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK */
+
+describe('Input', function () {
+  var element
+
+  beforeEach(function () {
+    element = $(
```
owenatgov, line 8: Is there no way around using jQuery here? Looking in other js tests, they use an approach akin to:

```suggestion
    element = document.createElement('div')
    element.innerHTML =
      '<div class="gem-c-input--reveal-password" data-module="input" data-show="Show" data-hide="Hide" data-password="password" data-announce-show="Your password is shown" data-announce-hide="Your password is hidden">' +
        '<input name="password" class="gem-c-input govuk-input" id="input" type="password">' +
      '</div>'
    
```

edwardhorsford:
@andysellick have you tried styling as a link? I think that's what I'd try for this 'secondary' action.

One thing you probably want to do (and might already be doing, but not mentioned) is to revert to password input on form submission. We want to make sure it continues to be _sent_ as a password field.

It might also be good to investigate how this interacts with password managers.

app/assets/javascripts/govuk_publishing_components/components/input.js
```diff
@@ -0,0 +1,48 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Input () { }
+
+  Input.prototype.start = function ($module) {
+    this.$module = $module[0]
+    this.input = this.$module.querySelectorAll('.gem-c-input')[0]
```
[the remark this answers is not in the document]
andysellick: Good point. I actually thought this as I wrote it, but for some reason didn't follow through. Will change.

app/assets/javascripts/govuk_publishing_components/components/input.js
```diff
@@ -0,0 +1,48 @@
...
+
+    var extraText = " <span class='gem-c-input__password-long-text'>" + this.$module.getAttribute('data-password') + '</span>'
```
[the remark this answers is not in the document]
andysellick: That seems like a good enough reason to not use it?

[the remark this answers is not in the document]
andysellick: It could be done that way, but I think this is clearer and simpler. What would be the advantage of doing it the other way?

spec/javascripts/components/input-spec.js
andysellick: It's okay to use jquery in tests, because the jasmine suite has its own version of jquery, so even after we remove all jquery from our code we can still have it in tests. Makes writing some of our tests a bit easier, too.

andysellick:
> One thing you probably want to do (and might already be doing, but not mentioned) is to revert to password input on form submission. We want to make sure it continues to be _sent_ as a password field.

Why would that be needed? Doesn't the backend just receive form values based on the `name` of the form input regardless of the `type`?

edwardhorsford:
> > One thing you probably want to do (and might already be doing, but not mentioned) is to revert to password input on form submission. We want to make sure it continues to be _sent_ as a password field.
> 
> Why would that be needed? Doesn't the backend just receive form values based on the `name` of the form input regardless of the `type`?

I *think* servers / logging / browsers handle password inputs differently. Worth checking with #security on cross gov slack.

edwardhorsford:
Reading [this old stackoverflow question](https://stackoverflow.com/questions/14788606/is-it-safe-to-use-type-text-for-password-field), it sounds like it might also have the risk that the browser remembers the password (but not in a password manager).

andysellick:
> Reading [this old stackoverflow question](https://stackoverflow.com/questions/14788606/is-it-safe-to-use-type-text-for-password-field), it sounds like it might also have the risk that the browser remembers the password (but not in a password manager).

Thanks @edwardhorsford that's an excellent point. I've modified the code to set `autocomplete="off"` if the password reveal option is used, I think that should remove that problem.

andysellick:
I've tested this in Firefox and it offers to remember the password when the password is shown, once the submit button is pushed on the password form. Chrome seems to behave differently - I can't currently get it to offer to remember the password, with or without the password reveal.

app/assets/javascripts/govuk_publishing_components/components/input.js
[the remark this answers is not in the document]
andysellick: Changed.

app/assets/javascripts/govuk_publishing_components/components/input.js
andysellick: Good point, I was apparently being lazy. Have refactored.

edwardhorsford:
> > Reading [this old stackoverflow question](https://stackoverflow.com/questions/14788606/is-it-safe-to-use-type-text-for-password-field), it sounds like it might also have the risk that the browser remembers the password (but not in a password manager).
> 
> Thanks @edwardhorsford that's an excellent point. I've modified the code to set `autocomplete="off"` if the password reveal option is used, I think that should remove that problem. Still looking into the password manager question.

I still think you're going to want to revert to `type=password` on form submission for safety. I'm pretty sure this has been discussed in #security before.

andysellick:
@edwardhorsford I've asked our developers about the risk of submitting a form containing a password in an input not of type `password` and been told it shouldn't make any difference to the http request that's being sent, or to how the backend receives it. Obviously if there is a problem I'd like to know about it so I can avoid it - do you have any recollection of the specifics of the problem that has previously been discussed?

edwardhorsford:
@andysellick I think at the _least_ it's safer. It's an easy and cheap thing for this js to do, and *ensures* there's no downstream consequences from it not being set to a password field.

Just as an example - setting `autocomplete="off"` is known to be very unreliable - especially in Chrome. Given there's no real downside to reverting to password field on submission, it feels the safer thing to do.

andysellick:
@edwardhorsford unfortunately this is a component of a form, not the form itself. In theory I could write some additional JS to attach to whatever form this is placed inside (to intercept the submit, revert the input to `type=password`, then submit the form) but I'm hesitant to do that because it may have unforeseen consequences (what if a form this gets used in already has JS controlling the submission, would that cause a conflict?) and it feels like a violation of the component model (components should be isolated, not interfering with other elements on the page).

I appreciate that reverting the input type would ensure there's no problems downstream, but I've yet to find evidence that such problems could occur (several developers I've spoken to can't identify a possible problem). As an alternative, I can extend the description of this component to include a recommendation that this approach is used if the show/hide password option is used, does that sound acceptable?

app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -57,6 +57,20 @@ examples:
       name: "labelledby"
       describedby: "wrapper"
       controls: "wrapper"
+  with_password_reveal:
+    description: |
+      Requires that the input type is also set to `password`.
+
+      Adds a password reveal button using JS after the input, which toggles the type of the input from password to text, revealing the password. It also sets the autocomplete on the input to "off", so that browsers don't accidentally remember the password.
+
+      Uses a visually hidden bit of text to inform screen reader users of the state of the input, in preference to announcing the content of the password input when toggled.
```
injms, line 66: Small rewording suggestion - I found the description a little hard to understand, so attempted to make it a bit clearer.

```suggestion
      Adds a password reveal button after the input which toggles the type of the input from password to text revealing the password. It also sets the `autocomplete` attribute on the input to "off" so that browsers don't accidentally remember the password.

      Uses a visually hidden bit of text to inform screen reader users of the state of the input rather than announcing the content of the password input when toggled.
```

config/locales/en.yml
```diff
@@ -54,6 +54,12 @@ en:
       what_wrong: "What went wrong?"
       send_me_survey: "Send me the survey"
       send: "Send"
+    input:
+      show: Show
+      hide: Hide
+      password: password
```
injms, line 60: I like that the translations have been thought about - i18n ftw. 

I don't think that splitting it out into words is best practise - I think the translation would be better as phrases for 'show password' and 'hide password'. In different languages the word 'show' (meaning appear) might need the context of 'password' to be accurate (and visa versa) - and splitting them up individual words doesn't allow that.

```suggestion
      show: Show password
      hide: Hide password
```

andysellick: The reason I split them is because 'password' gets wrapped in a span that gets hidden when the screen is narrower (on mobile but not below, I think) and I didn't want to put markup in the translation file.

app/assets/javascripts/govuk_publishing_components/components/input.js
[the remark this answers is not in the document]
injms: Using `createElement()` is speedier for the browser and uses less memory as there isn't any need to parse a string into a DOM tree. But I suspect for this that the difference is negligible.

app/assets/javascripts/govuk_publishing_components/components/input.js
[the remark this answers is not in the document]
injms: I was going to suggest a slightly different solution - which is to remove the `[0]` and loop through resulting `NodeList`. That way this can handle either single or multiple password inputs on the same page.

app/views/govuk_publishing_components/components/docs/input.yml
andysellick: Thanks, yeah that does read better. Changed.

app/assets/javascripts/govuk_publishing_components/components/input.js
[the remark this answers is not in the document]
andysellick: But because this is a component the JS will still be initialised for each instance of it on the page. I don't want to start trying to do a single initialisation for all where I have to check if it's already been done. And the most we'll have it on any page is twice.

config/locales/en.yml
injms: Hmm, tricky. Hiding the word 'password' then runs into the same translation issue of potentially removing the context for 'Show'. This is why i18n is difficult.

Is there a layout where the word 'password' doesn't need to be hidden?

andysellick: That would have to be a design question, I think. Unfortunately the phrase is quite long and takes up a fair bit of width on the smallest mobiles. Unless we drop it down below the input on mobile, as we do below 320px. I can see if that would be an option.

andysellick:
I've pushed some new changes (separated into another commit for reviewing ease):

- Show/hide password text is no longer split up, so translating should be easier. Button now drops below input on mobile
- component now detects submission event on parent form and reverts the type of the input back to password, just in case

See updated description for full details.

config/locales/en.yml
andysellick: Done, see latest commit.

app/assets/javascripts/govuk_publishing_components/components/input.js
```diff
@@ -0,0 +1,54 @@
...
+    this.statusText.setAttribute('aria-live', 'polite')
+    this.$module.insertBefore(this.statusText, this.showHide.nextSibling)
+
+    this.showHide.addEventListener('click', this.$module.togglePassword)
+
+    this.parentForm = this.input.form
+    var disableFormSubmitCheck = this.$module.getAttribute('data-disable-form-submit-check')
+
+    if (this.parentForm && !disableFormSubmitCheck) {
+      this.parentForm.addEventListener('submit', this.$module.revertToPasswordOnFormSubmit)
```
kevindew, line 36: It looks like this could accidentally add quite a few event handlers to a form. If I understand right this will add a submit event handler for every input component whether or not they are for a password.

Do you know if this definitely solves a problem or is a precaution?

A way around it could be a static function added to the eventListener so it doesn't add this multiple times, or to only apply this when it's a password input.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
```diff
@@ -18,3 +18,37 @@
     padding-left: govuk-spacing(6);
   }
 }
+
+.gem-c-input--reveal-password {
+  // stylelint-disable declaration-block-no-duplicate-properties
+  display: table; // IE fallback
```
kevindew, line 24: Is this for IE <= 9? My instinct was to wonder whether adding extra code was worth this given it'd be mostly for IE8 users, but maybe I've misunderstood.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -119,6 +121,18 @@
     <%= tag.div class: "govuk-input__wrapper" do %>
       <%= input_tag %><%= tag.span suffix, class: "govuk-input__suffix", aria: { hidden: true } %>
     <% end %>
+  <% elsif reveal_password %>
+    <%= tag.div class: "gem-c-input--reveal-password",
```
kevindew, line 125: Do we not want the wrapper here? it feels a bit surprising to not have it.

kevindew (commented):
This looks neat 🥇  and I imagine will be useful in other places.

I added a couple of comments. I do feel like this adds quite a lot of specific functionality for one input scenario that might be better suited in a component of its own right. It feels like if we had a reveal-password-input component we could separate all the customisations of this to it's own place and call the input component through it. Does that seem realistic?

app/assets/javascripts/govuk_publishing_components/components/input.js
andysellick: The JS should only initialise for an input with the password reveal turned on, not for every instance of the input component. The data-module attribute should only be present when the component is set to reveal password mode.

It's unlikely that anyone would use this more than twice in a form (you might have password and confirm password) so I'm not too worried about excessive event handlers.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
andysellick: This is actually for IE <= 11. Pretty much every version of IE has shonky support for flexbox. I experimented with the various hacks you can try to get around it but concluded the myriad of excess CSS wasn't worth the effort when it would fallback to using display table and display reasonably well anyway (see screenshots in description - only IE10 is noticeably a bit rubbish).

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Oh the `govuk-input__wrapper`? That's a good point. Odd. Seems to render fine without it, but I'll take a look, thanks.

andysellick:
@kevindew thanks for the review. Separating this out into a new component hadn't occurred to me, it's an interesting idea. To be honest at this stage of the process I'd rather get this finished than implement such a dramatic change, but it's definitely worth considering for a future iteration.

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Have investigated. It's basically unnecessary - `govuk-input__wrapper` adds some flexbox styles, which sort of help with this layout but don't really.

app/assets/javascripts/govuk_publishing_components/components/input.js
kevindew: Oh is that because only this usage of the component has the data-module attribute of input? I didn't catch that. It seems like that might be a little deceptive if the whole JS for the component is only for a particular modifier. Perhaps that would be clearer if that JS initialised for all instances but exited early when not for password.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
kevindew: I see, maybe my understanding is off but I thought IE 10 and IE 11 understood `display: flex` so this definition of `display: table` would be overwritten in IE > 9 and not have an effect?

app/views/govuk_publishing_components/components/_input.html.erb
kevindew: Sure, thanks for looking into it. If it's unnecessary, but not detrimental, I wonder if it's better to include it as otherwise this is just a special case for this modifier which makes it a bit more special.

One thing that's a bit weird here is that we have this `gem-c-input--reveal-password` class on this div, yet we have `gem-c-input` set on the input (the input seems the wrong place really 🤔 ). I wonder, to match the conventions here, whether this should be `gem-c-input__wrapper--reveal-password`?

kevindew:
> @kevindew thanks for the review. Separating this out into a new component hadn't occurred to me, it's an interesting idea. To be honest at this stage of the process I'd rather get this finished than implement such a dramatic change, but it's definitely worth considering for a future iteration.

Thanks Andy, that makes sense and I can understand it being a frustrating suggestion at this point. I am a bit concerned about this component as it seems a bit sprawling. Maybe it's something to chat about at a frontend session.

app/assets/stylesheets/govuk_publishing_components/components/_input.scss
andysellick: Correct. IE10 and 11 sort of understand `display: flex`, but not properly. I was getting weird layout results.

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Ah yes, probably, good shout. Have changed.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -7,7 +7,9 @@
   describedby ||= nil
   controls ||= nil
   data ||= nil
-  autocomplete ||= nil
+  disable_form_submit_check = true if disable_form_submit_check
+  reveal_password = type == "password" && reveal_password ? true : false
```
injms, line 11: I'd suggest removing the ternary operator and either replace it with a check for `true`:

```suggestion
  reveal_password = type == "password" && reveal_password == true
```

Or remove the ternary operator, which has the same result:
```suggestion
  reveal_password = type == "password" && reveal_password
```

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -7,7 +7,9 @@
...
+  autocomplete = reveal_password ? "off" : autocomplete ? autocomplete : nil
```
injms, line 12: I don't think that this double ternary is easily understandable - can it be split up?
```suggestion
  autocomplete ||= nil
  autocomplete = "off" if reveal_password
```

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Thanks, have changed.

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Thanks, have changed.

andysellick:
I've reworked the code to make this a separate component. Have kept this as a separate commit for now to aid with reviewing, will squash them all before merging. Thanks for everyone's comments so far, I'd really appreciate some re-reviews.

app/assets/javascripts/govuk_publishing_components/components/show-password.js
```diff
@@ -0,0 +1,63 @@
...
+    this.statusText = document.createElement('span')
+    this.statusText.classList.add('govuk-visually-hidden')
+    this.statusText.innerHTML = this.hiddenPassword
+    this.statusText.setAttribute('aria-live', 'polite')
+    this.wrapper.insertBefore(this.statusText, this.showHide.nextSibling)
+
+    this.showHide.addEventListener('click', this.$module.togglePassword)
+
+    this.parentForm = this.input.form
+    var disableFormSubmitCheck = this.$module.getAttribute('data-disable-form-submit-check')
```
kevindew, line 42: Do we expect this attribute to be used in the rollout of this component? I wondered if it was something that we think someone might want and actually a case of [YAGNI](https://martinfowler.com/bliki/Yagni.html) where we could drop the code and test?

app/views/govuk_publishing_components/components/_show_password.html.erb
```diff
@@ -0,0 +1,35 @@
...
+  label ||= nil
+  value ||= nil
+  id ||= nil
+  name ||= nil
+  describedby ||= nil
+  hint ||= nil
+  error_message ||= nil
+  error_items ||= nil
+%>
+<% if label %>
```
kevindew, line 12: It looks like this would show nothing if you don't pass in a label, is that intentional? Might be that we want to raise an error in that case.

app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
```diff
@@ -0,0 +1,54 @@
+.gem-c-show-password__input-wrapper {
+  display: table; // IE fallback
+  display: flex;
```
kevindew, line 3: Oh interesting that stylelint doesn't complain when properties are repeated sequentially - TIL

app/assets/javascripts/govuk_publishing_components/components/show-password.js
```diff
@@ -0,0 +1,63 @@
...
+    this.showPasswordText = this.$module.getAttribute('data-show')
+    this.hidePasswordText = this.$module.getAttribute('data-hide')
+    this.shownPassword = this.$module.getAttribute('data-announce-show')
+    this.hiddenPassword = this.$module.getAttribute('data-announce-hide')
+
+    // wrap the input in a new div
+    this.wrapper = document.createElement('div')
+    this.wrapper.classList.add('gem-c-show-password__input-wrapper')
+    this.input.parentNode.insertBefore(this.wrapper, this.input)
+    this.wrapper.appendChild(this.input)
```
kevindew, line 23: This is neat - I was wondering how you achieved wrapping the input without the hint / error handling being caught up in the wrap.

app/views/govuk_publishing_components/components/docs/show_password.yml
```diff
@@ -0,0 +1,72 @@
...
+      label:
+        text: Please enter your password
+      value: "this is my password"
+      hint: "Your password must be at least twelve thousand characters"
+  with_error:
+    data:
+      label:
+        text: Please enter your password
+      value: "this is my password"
+      error_message: No it isn't
```
kevindew, line 64: Are you concerned at all that the show/reveal button lacks the error border? It didn't seem a big concern for me but thought I'd check.

kevindew (commented):
Looks great @andysellick - thanks for taking the time to refactor it. To my eyes it looks a lot cleaner distinct and I hope you're happier with it too. I've just got a couple of minor comments.

Way out of scope, but it has also got me wondering about the input component overall and it's responsibilities. I wonder if it's ended up rather overloaded being essentially both a form group and input field component. Where instead we might benefit from a simple component based on an input and a form group item one. Where perhaps we'd do something like this:

```
field_id ||= SecureRandom.alpha(10)

<%= render "govuk_publishing_components/components/form_field", {
  label: label,
  field_id: field_id,
  describedby: describedby,
  hint: hint,
  error_message: error_message,
  error_items: error_items,
} do |field_describedby| %>
    value: value,
    id: field_id,
    name: name,
    describedby: field_describedby,
    type: "password",
    autocomplete: "off",
  } %>
<% end %>
```

I might be missing some practicalities though. Just seems that we've got a lot of peripheral things in input component that make it hard to not re-use and then we've got all of that already copied and pasted for both textarea and select components to varying degrees of consistency.

injms (approved):
I think that this is good to go 🚀 🔐 

I agree with @kevindew that it'd be interesting to look at the input component, and I hope we don't go down the path of effectively reinventing HTML by having an form component wrap a label component wrap an input component - I think we'd be in danger of losing the ease of use that comes with this set of components. Nevertheless, it'd be a discussion worth having.

app/assets/javascripts/govuk_publishing_components/components/show-password.js
andysellick: Not for the initial rollout, no. However I am extremely nervous about giving a component control of elements outside of its scope - it's necessary behaviour, but it's somewhat against our component principles and I don't want to assume it won't cause some problems down the line. Given that this addition is only a few extra lines I'd rather keep it in and calm my own fears.

app/views/govuk_publishing_components/components/_show_password.html.erb
andysellick: Have updated to error if no label provided.

app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
andysellick: Yeah - I'm a bit weirded out by that, because it definitely complained the first time I pushed this change, back when it was part of the input component. I'm at a loss to explain this.

app/views/govuk_publishing_components/components/docs/show_password.yml
andysellick: I don't think I'm concerned, no. If an error is present, it's to do with the contents of the input box, not the show/hide password button. However, I have pushed a new commit to address the slightly broken focus/active outline overlap problem.

app/views/govuk_publishing_components/components/docs/show_password.yml
```diff
@@ -0,0 +1,79 @@
+name: Show password input
+description: A password input field that allows the password to be shown.
+body: |
+  Adds a password reveal button to an input that toggles its type from password to text, revealing the password. Does not appear if JS is disabled.
```
injms, line 4: Optinionated nitpick - JavaScript > JS:

```suggestion
  Adds a password reveal button to an input that toggles its type from password to text, revealing the password. Does not appear if JavaScript is disabled.
```

app/views/govuk_publishing_components/components/docs/show_password.yml
```diff
@@ -0,0 +1,79 @@
...
+govuk_frontend_components:
+  - text-input
+examples:
+  default:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+  disable_form_submit_check:
+    description: The component sets the `autocomplete` attribute on the input to "off" so that browsers don't remember the password. This isn't reliable, so the component reverts the input type back to a password when its parent form is submitted. If this is causing conflicts with other JS on your form, you can set `disable-form-submit-check` to `true` to disable this functionality and implement it yourself.
```
injms, line 39:
```suggestion
    description: The component sets the `autocomplete` attribute on the input to "off" so that browsers don't remember the password. This isn't reliable, so the component reverts the input type back to a password when its parent form is submitted. If this is causing conflicts with other JavaScript on your form, you can set `disable-form-submit-check` to `true` to disable this functionality and implement it yourself.
```

app/views/govuk_publishing_components/components/docs/show_password.yml
```diff
@@ -0,0 +1,79 @@
...
+        text: Please enter your password
+      value: this is my password
+      id: custom_id
+      name: custom_name
+  with_hint:
+    data:
+      label:
+        text: Please enter your password
+      value: this is my password
+      hint: Your password must be at least twelve thousand characters
```
injms, line 65: "and contains characters from the Latin, Cyrillic, and Aurebesh alpabets."

injms (approved):
Just had a minor opinion that spelling out JavaScript is better than abbreviating it - feel free to ignore and get this merged in 🕵️

app/views/govuk_publishing_components/components/docs/show_password.yml
andysellick: ⭐  ⚔️

app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
kevindew: Yeah, it turns out the [stylelint-config-recommended rule](https://github.com/stylelint/stylelint-config-recommended/blob/20b9df0e77cd78b699d6265e1b4d614e7d74a332/index.js#L9-L14) has [this condition](https://stylelint.io/user-guide/rules/declaration-block-no-duplicate-properties#ignore-consecutive-duplicates-with-different-values) which will allow immediate duplicate properties as long as the values differ, with the expectation it is used for browser compatibility. Sounds like it'll meet our use case.

app/views/govuk_publishing_components/components/docs/show_password.yml
andysellick: Since I thrive on nitpicks, I can't ignore this one. Changed (and the other one).

PR #1795, opened 2020-11-24
Replaces the random placekitten image placeholder

injms:
## What

Replaced the random kitten placeholder images with non-random images.

The example image used for the large image was also 404-ing, so I replaced it.
## Why

Whilst place kitten is the second best placeholder service available, it supplies random images - which doesn't play well with visual regression tests. Replacing it with a non-random image means that the visual regression tests won't flag a page as having changed.

## Visual Changes

No layout changes - just the images have been changed.

Before; image doesn't load as it violated the content security policy:

image

After:
image


andysellick:
But, but... John Whiskers MP!!! 😢

alex-ju (approved):

PR #1796, opened 2020-11-25
Extend component auditing

andysellick:
## What
Add a few more applications to the auditing (the new accounts system, and DGU).

Also add in a simple check to look for any instances of jQuery in applications.

Screenshot 2020-11-25 at 17 14 31

Screenshot 2020-11-25 at 16 59 09

## Why
Thought this might be useful.

### DIFF
--- a/app/models/govuk_publishing_components/audit_applications.rb
+++ b/app/models/govuk_publishing_components/audit_applications.rb
@@ -7,6 +7,7 @@ module GovukPublishingComponents
       application_found = application_exists(path)
       components_found = []
       @gem_style_references = []
+      @jquery_references = []
 
       if application_found
         templates = Dir["#{path}/app/views/**/*.erb"]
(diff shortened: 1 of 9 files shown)


app/models/govuk_publishing_components/audit_applications.rb
```diff
@@ -76,8 +78,14 @@ def find_components(files, find, type)
       files.each do |file|
         src = File.read(file)
         components_found << find_match(find, src, type)
-        gem_references = find_gem_references(file, src)
-        @gem_style_references << gem_references if gem_references
+
+        if type == "javascripts"
+          jquery_references = find_code_references(file, src, /\$\(/)
```
injms, line 83: Should we include `jQuery()` as well?

Another thought is to look for the ESLint global declaration that stops it complaining about `$` being an undeclared variable:
```javascript
/* global $ */
```

injms: Hmm, having thought about this I can't find any file that exclusively uses `jQuery()` - so I don't think that this is relevant.

injms (approved):
Looks good to go 👍

PR #1799, opened 2020-11-27
Update to responsive bullet SVG

chris-gds:
## What?

Updating the current list bullet numbered SVG into a responsive version so it scales correctly via zoom-text-only / improves a11y considerations 

## Why?

_"Step numbers don't get bigger with bigger font size"_ - [Ticket] (https://trello.com/c/RurRepSL)
Fail of WCAG SC 1.4.4

## Visual Changes

[Example page under _FOI request_](https://www.gov.uk/government/organisations/companies-house)

Before > After (100% - Chrome)

Screenshot 2020-11-27 at 18 24 37

Before > After (@ 200% - Firefox)

Screenshot 2020-11-27 at 18 46 41

Before > After (100% - Firefox)

Screenshot 2020-11-27 at 17 57 35

Before > After (@ 200% - Firefox)
Screenshot 2020-11-27 at 17 58 05

IE11 

Screenshot 2020-11-27 at 18 01 01

## Anything else?

Vertically aligns better for FF and uses "NTA" font when available.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix to update the current list bullet numbered SVG into a responsive version so it scales correctly via zoom-text-only / improves a11y considerations ([PR #1799](https://github.com/alphagov/govuk_publishing_components/pull/1799))
+
 ## 23.7.3
 
 * Fix to resolve input zoom-text overlap ([PR #1793](https://github.com/alphagov/govuk_publishing_components/pull/1793))
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
@@ -12,10 +12,10 @@
 
     @for $i from 1 through 30 {
       &:nth-child(#{$i}) {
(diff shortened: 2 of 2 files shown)


danacotoran:
Looks much better than before, also good call regarding using the NTA font if available.
However, I noticed this issue testing IE11 on Browserstack:

### Before: circular but text is misaligned
Screenshot 2020-11-30 at 16 03 53

### After: text centred but surrounding circle is cut off at the edges 
Screenshot 2020-11-30 at 16 03 42
It doesn't look massively broken, but also does not look as good as it does on other modern browsers. 
If it's massively hard to fix, then it might not be worth the effort to fix it though.

chris-gds:
> It doesn't look massively broken, but also does not look as good as it does on other modern browsers.
> If it's massively hard to fix, then it might not be worth the effort to fix it though.

It does look more pronounced in the side-by-side 🤔     
I've timeboxed some effort this morning without arriving at simple solution.  I'm wondering if the continued time investment is warranted, eg IE11 would expected to get a less than ideal experience.  

I was also basing this on another SVG element nearby - the "download link" that is a bit different also in IE11

image

Screenshot 2020-12-01 at 09 46 48

I'm going to suggest I'll raise an issue on this one as an enhancement so I doesn't get lost however suggest for this PR that's OK for now - what do you think?

danacotoran:
Yeah as I said, don't worry about it if it's hard to fix @chris-gds. 
Definitely don't spend any more time on it, it looks better than it did before even as it is.

danacotoran (approved):

PR #1800, opened 2020-12-01
Amend share links columns spacing

danacotoran:
## What

This changes the share links columns to use `em` values instead of `px` which means that the width of the link containers increase proportionately to the font size. This should make for a better experience for users who use the text zoom setting in their browser.

## Why

When increasing the font size to 200%, some of the social media links
are overlapping, making them unreadable for users who use this setting.

https://trello.com/c/1nrlxFEB

### No zoom before
Screenshot 2020-12-01 at 10 15 21

### No zoom after
Screenshot 2020-12-01 at 10 15 34

### 200% text zoom before
Screenshot 2020-12-01 at 10 16 00

### 200% text zoom after
Screenshot 2020-12-01 at 10 16 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


injms (approved):
👍

PR #1804, opened 2020-12-02
Update subscription links default email link text

1pretz1:
Update subscription links default text

This has been an agreed content change for emails links as we move away
from the terminology 'alert'.

Also removes the `default` argument as I'm struggling to see why we need
it in this instance. Using `default` also meant that a misconfigured locale
path was being covered up which has been corrected.

## Before

Screenshot 2020-12-02 at 11 35 03

## After

Screenshot 2020-12-02 at 11 34 13

Trello:
https://trello.com/c/ssSnzp2L/640-remove-alert-from-all-the-sign-up-links

### DIFF
--- a/lib/govuk_publishing_components/presenters/subscription_links_helper.rb
+++ b/lib/govuk_publishing_components/presenters/subscription_links_helper.rb
@@ -14,13 +14,13 @@ module GovukPublishingComponents
       def email_signup_link_text
         return @local_assigns[:email_signup_link_text] if @local_assigns[:email_signup_link_text]
 
-        I18n.t("govuk_component.subscription_links.email_signup_link_text", default: "Get email alerts")
+        I18n.t("components.subscription_links.email_signup_link_text")
       end
 
       def feed_link_text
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/subscription_links_helper.rb
```diff
@@ -14,13 +14,13 @@ def initialize(local_assigns)
       def email_signup_link_text
         return @local_assigns[:email_signup_link_text] if @local_assigns[:email_signup_link_text]
 
-        I18n.t("govuk_component.subscription_links.email_signup_link_text", default: "Get email alerts")
+        I18n.t("components.subscription_links.email_signup_link_text")
```
benthorner, line 17: Note to self: confirmed these locales are available in our apps due to [[1](https://github.com/alphagov/govuk_publishing_components/blob/9a5c519fb53c628987f371413758fa794b3b7d9c/lib/govuk_publishing_components.rb#L46)], so agree the default should be unnecessary.

benthorner (approved):
Looks good - makes sense to avoid the default translation, since this was masking the fact that the key was incorrect 💯.

PR #1805, opened 2020-12-02
Version 23.7.5

1pretz1:
Trello:
https://trello.com/c/ssSnzp2L/640-remove-alert-from-all-the-sign-up-links

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (23.7.4)
+    govuk_publishing_components (23.7.5)
       govuk_app_config
       kramdown
       plek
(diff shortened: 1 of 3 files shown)


Gemfile.lock
```diff
@@ -106,7 +106,7 @@ GEM
       rest-client (~> 2.0)
     globalid (0.4.2)
       activesupport (>= 4.2.0)
-    govuk_app_config (2.7.0)
+    govuk_app_config (2.7.1)
```
andysellick, line 109: Does this need bumping?

andysellick (approved):

Gemfile.lock
1pretz1: Thanks for the review @andysellick! 

Doesn't need bumping as part of this change but may as well as it's only a minor patch and was included when I bundled.

PR #1807, opened 2020-12-03
Add some GOV.UK Accounts specific PII redacts

andysellick:
## What
Add some new redactions to remove PII from URLs. This is targeted specifically at parameters in URLs in GOV.UK Accounts, so the matching is relatively simple.

## Why
The things that are appearing in URLs could be linked to specific users so we need to remove these values.

## Visual Changes
None.

Trello card: https://trello.com/c/bjcoDa2N/462-remove-or-redact-potential-pii-from-urls

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add some GOV.UK Accounts specific PII redacts ([PR #1807](https://github.com/alphagov/govuk_publishing_components/pull/1807))
+
 ## 23.7.6
 
 * Amend share links columns spacing ([PR #1800](https://github.com/alphagov/govuk_publishing_components/pull/1800))
--- a/app/assets/javascripts/govuk_publishing_components/analytics/pii.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/pii.js
@@ -8,6 +8,11 @@
   var POSTCODE_PATTERN = /[A-PR-UWYZ][A-HJ-Z]?[0-9][0-9A-HJKMNPR-Y]?(?:[\s+]|%20)*[0-9][ABD-HJLNPQ-Z]{2}/gi
   var DATE_PATTERN = /\d{4}(-?)\d{2}(-?)\d{2}/g
 
+  // specific URL parameters to be redacted from accounts URLs
+  var RESET_PASSWORD_TOKEN_PATTERN = /reset_password_token=[a-zA-Z0-9-]+/g
+  var UNLOCK_TOKEN_PATTERN = /unlock_token=[a-zA-Z0-9-]+/g
+  var STATE_PATTERN = /state=.[^&]+/g
+
   function shouldStripDates () {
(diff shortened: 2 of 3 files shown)


kevindew:
Rather than adding in accounts specific parameters could we use a data/meta attribute in the HTML or JS to specify which parameters shouldn't be redacted in analytics and then the control of the parameters could be done within an individual application rather than these scripts needing to know about the app.

andysellick:
> Rather than adding in accounts specific parameters could we use a data/meta attribute in the HTML or JS to specify which parameters shouldn't be redacted in analytics and then the control of the parameters could be done within an individual application rather than these scripts needing to know about the app.

So the regex would be in a meta attribute in the page, which would be picked up by the JS and applied to the URL? Do we do this already - and if so, where?

That sounds cleaner right now although I'm wondering about where these URL parameters will ultimately appear. Eventually we'll have a sign in link from every page on GOV.UK at which point this redaction should be pretty much universal?

kr8n3r:
in static https://github.com/alphagov/static/blob/8b08269e47403f5951d956f2c63d71fb8b565110/app/assets/javascripts/analytics/pii.js#L9

kevindew:
> So the regex would be in a meta attribute in the page, which would be picked up by the JS and applied to the URL? Do we do this already - and if so, where?

We have the approach already where you specify with `meta[name="govuk:static-analytics:strip-dates"]` to specify stripping dates.

I'd not suggest going down the whole route of putting full regexes in for the JS - I'd see about listing parameter names for example: 

```
<meta name="govuk:static-analytics:pii-query-string-parameters" content="param-1,param-2">
```

> That sounds cleaner right now although I'm wondering about where these URL parameters will ultimately appear. Eventually we'll have a sign in link from every page on GOV.UK at which point this redaction should be pretty much universal?

Sure, that sounds quite a distant eventually though? But, yes if the intention is for them to be ubiquitous there's probably some thought that should be put into them. They seem quite generic in name (for example state is a very common parameter). Perhaps if ubiquity is the expectation they should be prefixed in a similar way to `utm_` parameters. If you were to do that you'd have the advantage that you could run your strip against a prefix rather than needing to list each individual parameter (and need to update this gem any time accounts added or retired a parameter)

huwd:
@kevindew / @kr8n3r - I think it's a really good point and if we made that change now on the ground floor we'd have a cleaner more knowable codebase in the long run.

That said we're already getting some of these Params into GA. There's only a 1 hour gap where they could be possibly identifying but we'd really like to cut that risk off now.

Would you be up for us merging this now, then immediately adding a refactor card to our board and getting our PMs/DMs to prioritise dev time into an approach described as above?

... I realise the "I'll do it later" defence doesn't have the most glorious history of follow through...

huwd (approved):

kevindew:
No qualms about resolving it quickly after.

Would suggest ideally in short term this is monkey patched in app rather than us doing a gem release for a short term fix in a different app, but if that's not realistic then go for it.

andysellick:
Thanks both. Just need to update the changelog and then I'll get this merged.

kr8n3r:
I don't work on gov.uk so carry as you were 👍

PR #1811, opened 2020-12-07
Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link

barrucadu:
## What
Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link.

## Why
GOV.UK uses cross-domain analytics (only if the user has consented to cookies on both domains, of course) to give a joined up view of how users interact with government. So naturally this has to work with GOV.UK Accounts too, or we’ll have an ever-growing blind spot in this joined up view.

Cross-domain analytics works by sticking a `_ga` query param onto cross-domain links, which the analytics code at the link destination picks up. The problem is that with accounts, most of the cross-domain stuff is in the form of OAuth journeys: we have cross-domain *redirects*. So we have non-cross-domain links which may do a cross-domain redirect; for example, navigating from an unauthenticated part of a service to an authenticated part, the link is a normal internal link, but it’ll redirect a logged-out user to the accounts domain.

So that these cross-domain redirects don't become an analytics blind spot, we need to do two things:

1. Add the `_ga` param to internal links and forms which *may* redirect (this PR)
2. Ensure that any such redirects preserve the param

## Visual Changes
N/A

---

[Trello card](https://trello.com/c/ISYymNsg/426-fix-cross-domain-tracking-on-redirects)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,11 +7,17 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link ([PR #1811](https://github.com/alphagov/govuk_publishing_components/pull/1811))
+
 ## 23.9.2
 
 * Fix for search input misalignment ([PR #1823](https://github.com/alphagov/govuk_publishing_components/pull/1823))
 * Add type="button" to Show/Hide password button ([PR #1826](https://github.com/alphagov/govuk_publishing_components/pull/1826))
 * Amend Show/Hide password button CSS ([PR #1828](https://github.com/alphagov/govuk_publishing_components/pull/1828))
+
+>>>>>>> 13252807 (Add CHANGELOG entry)
 
 ## 23.9.1
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
@@ -1,0 +1,34 @@
+;(function (global) {
+  'use strict'
+
+  var GOVUK = global.GOVUK || {}
+  GOVUK.Modules = GOVUK.Modules || {}
+
+  GOVUK.Modules.ExplicitCrossDomainLinks = function () {
+    this.start = function ($module) {
+      var element = $module[0]
+
+      if (!global.ga) { return }
+
+      global.ga(function () {
+        var trackers = global.ga.getAll()
+
+        if (!trackers.length) { return }
+
+        var linker = new global.gaplugins.Linker(trackers[0])
+
+        var attrAction = element.getAttribute('action')
+        if (attrAction) {
+          element.setAttribute('action', linker.decorate(attrAction))
+        }
+
+        var attrHref = element.getAttribute('href')
+        if (attrHref) {
+          element.href = linker.decorate(attrHref)
+        }
+      })
+    }
+  }
+
+  global.GOVUK = GOVUK
+})(window)
--- a/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
@@ -1,0 +1,81 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK */
+
+describe('Explicit cross-domain linker', function () {
+  'use strict'
+
+  var explicitCrossDomainLinks
+  var element
+  var linker
+  var trackers
+
+  beforeEach(function () {
+    window.ga = function (callback) {
+      window.ga = {
+        getAll: function () {
+          return trackers
+        }
+      }
+
+      callback()
+    }
+
+    window.gaplugins = {
+      Linker: function () {}
+    }
+
+    linker = {
+      decorate: function () {}
+    }
+
+    explicitCrossDomainLinks = new GOVUK.Modules.ExplicitCrossDomainLinks()
+
+    spyOn(window.gaplugins, 'Linker').and.returnValue(linker)
+    spyOn(linker, 'decorate').and.returnValue('/somewhere?_ga=abc123')
+  })
+
+  afterEach(function () {
+    element.remove()
+    delete window.ga
+    delete window.gaplugins
+  })
+
+  describe('links', function () {
+    beforeEach(function () {
+      element = $('<a href="/somewhere">')
+    })
+
+    it('leaves the link href if unchanged there are no trackers in ga', function () {
+      trackers = []
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('href')).toEqual('/somewhere')
+    })
+
+    it('modifies the link href to append ids from ga to the destination url', function () {
+      trackers = [{ ga_mock: 'foobar' }]
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('href')).toEqual('/somewhere?_ga=abc123')
+    })
+  })
+
+  describe('forms', function () {
+    beforeEach(function () {
+      element = $('<form method="POST" action="/somewhere">' +
+               '<input type="hidden" name="key" value="value" />' +
+               '<button type="submit">Create a GOV.UK account</button>' +
+             '</form>')
+    })
+
+    it('leaves the form action if unchanged there are no trackers in ga', function () {
+      trackers = []
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('action')).toEqual('/somewhere')
+    })
+
+    it('modifies the form action to append ids from ga to the destination url', function () {
+      trackers = [{ ga_mock: 'foobar' }]
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('action')).toEqual('/somewhere?_ga=abc123')
+    })
+  })
+})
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## unreleased
+
+* Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link ([PR #1811](https://github.com/alphagov/govuk_publishing_components/pull/1811))
```
andysellick, line 12: Nitpick: this should match the PR title.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -0,0 +1,43 @@
+/* global GOVUK, $, ga */
+
+(function () {
+  'use strict'
+  window.GOVUK = window.GOVUK || {}
+
+  var ExplicitCrossDomainLinks = function () {
+    this.decorateForm = function (element, trackers) {
+      if (trackers.length) {
+        var existing = element.attr('action')
```
andysellick, line 10: We're trying to remove our dependency on jQuery so all new JS code is supposed to be written without it. Having said that, that's a bit of an ask. If you like you can stick a card in the backlog to get this de-jQuery-ed as a piece of tech debt, and I can look at it later.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -0,0 +1,43 @@
...
+      if (trackers.length) {
+        var existing = element.attr('href')
+        var linker = new window.gaplugins.Linker(trackers[0])
+        var tracked = linker.decorate(existing)
+        element.attr('href', tracked)
+      }
+    }
+  }
+
+  ExplicitCrossDomainLinks.start = function () {
```
andysellick, line 27: This doesn't quite match our normal structure for a GOVUK modules bit of code, but I'm aware I might be missing out on some previous discussions around this. Was it that there were some problems making this into a proper GOVUK module so this was the best solution for now?

andysellick, line 27: If this isn't a GOVUK module, how is this code initialised? I see that it gets included in `analytics.js` but what calls the `start` function to make it run?

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
andysellick: (related - tests don't need to have jQuery removed as jasmine has its own jQuery for running tests, independent of the publically-facing frontend code)

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
andysellick: The answer to this may impact the structure of the test file, as I see you're explicitly calling `decorateLink` (etc.) on the code snippets.

andysellick (commented):
A few questions, but looks like a great start, thanks for moving it into the gem.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
barrucadu: Right, so I don't actually know how the module system works, so I put this together based on the ecommerce code, but probably did that wrong.

barrucadu: Is there some documentation on how to write a new module?

barrucadu: Aha: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/javascript-modules.md

I'll have a read of that and rewrite the code

barrucadu:
@andysellick I think this is properly set up as a module now

spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
```diff
@@ -0,0 +1,74 @@
...
+    it('leaves the link href if unchanged there are no trackers in ga', function () {
+      spyOn(window.ga, 'getAll').and.returnValue([])
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('href')).toEqual('/somewhere')
+    })
+
+    it('modifies the link href to append ids from ga to the destination url', function () {
+      spyOn(window.ga, 'getAll').and.returnValue([{ ga_mock: 'foobar' }])
+      explicitCrossDomainLinks.start(element)
+      expect(element.attr('href')).toEqual('/somewhere?_ga=abc123')
```
andysellick, line 50: Is it worth testing that the ga param is appended correctly if there are already params on the href attribute?

andysellick (commented):
Looks really good! I think you even managed to remove the jQuery (?) good job.

spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
barrucadu: The test's stubbing the value to return (spyOn(linker, 'decorate').and.returnValue('/somewhere?_ga=abc123') above), not executing the actual ga linker code, so adding another test for a different url wouldn't test anything different. All this test is doing really is checking that linker.decorate gets called.

andysellick: Oh I see. Never mind then.

barrucadu:
There's just one thing I want to check, in https://github.com/alphagov/govuk-account-manager-prototype/pull/478/files there are some elements using multiple modules, eg:

```
<a class="govuk-link" href="<%= service[:url] %>" data-module="gem-track-click explicit-cross-domain-links" data-track-category="account-create" data-track-action="confirm-email" data-track-label="<%= service[:name] %>">
```

Will that work?

barrucadu:
Hmm, it doesn't work.  There's only one module per element: https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/modules.js#L31

andysellick:
Yes, I'm afraid you can only have one module per element. But you can wrap that link in a span and put the `gem-track-click` module on it, that module looks down through the DOM to find the element to track.

andysellick:
Or move it to the nearest parent element, that should work too.

barrucadu:
> But you can wrap that link in a span and put the gem-track-click module on it

That's much better than the solution I just hacked together, thanks

barrucadu:
Ok, I've updated the account manager PR to fix the modules

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -0,0 +1,32 @@
...
+      var linker = new global.gaplugins.Linker(trackers[0])
+
+      var attrAction = this.$module.getAttribute('action')
+      if (attrAction) {
+        this.$module.setAttribute('action', linker.decorate(attrAction))
+      }
+
+      var attrHref = this.$module.getAttribute('href')
+      if (attrHref) {
+        this.$module.setAttribute('href', linker.decorate(attrHref))
```
owenatgov, line 26: I'm _pretty sure_ you don't need `setAttribute` for this. I could be totally wrong so if you test this and it doesn't work or you've already tried this then please ignore. If it does work you might be able to do the same for the `action` attribute above but I'm not sure how this would play with older browsers.

```suggestion
        this.$module.href = linker.decorate(attrHref)
```

owenatgov (commented):
Looking good 💪  One little query about attribute setting but no big issues with the code that I can see. It would be useful to see this working somewhere for a quick test if that's possible.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
barrucadu: Great, it works for `href`.  How could I make sure `this.$module.action` is fine for older browsers?

barrucadu:
Good call on deploying it somewhere to try it, it seems `window.ga` isn't always initialised before I try to call `window.ga.getAll()` - and in fact the GA docs explicitly warn against doing that: https://developers.google.com/analytics/devguides/collection/analyticsjs/ga-object-methods-reference

Trying to restructure it now to use that callback approach.

barrucadu:
I don't understand the CI failure I'm getting:

> explicit-cross-domain-links.spec.js:15:31: Expected an assignment or function call and instead saw an expression.

What's wrong with

```
      window.ga = {
        getAll: function () {
          return trackers
        }
      }
```

?

But with this module deployed to integration as part of finder-frontend (https://github.com/alphagov/finder-frontend/compare/msw/accounts-cross-domain-demo) looks like it works:

Screenshot 2020-12-08 at 14 21 58
Screenshot 2020-12-08 at 14 22 08

barrucadu:
Well, the test is somehow passing now 🤷‍♂️

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
owenatgov: The best way to do this is to just check on older browsers. If you don't have access to browserstack you can get it by going via Matt Hobbs _or_ I could test this. Having said that, if you're under pressure to get this out, I'm not precious about this so we can ignore it.

owenatgov (approved):

PR #1814, opened 2020-12-09
Add support for secondary solid buttons

DilwoarH:
# What
Solid secondary buttons styles from GOVUK design systems

# Why
User research has shown that solid secondary buttons worked well

Screenshot 2020-12-09 at 11 09 08

### DIFF
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -17,6 +17,7 @@ module GovukPublishingComponents
                   :start,
                   :secondary,
                   :secondary_quiet,
+                  :secondary_solid,
                   :destructive,
                   :name,
                   :value,
@@ -41,6 +42,7 @@ module GovukPublishingComponents
         @start = local_assigns[:start]
         @secondary = local_assigns[:secondary]
         @secondary_quiet = local_assigns[:secondary_quiet]
+        @secondary_solid = local_assigns[:secondary_solid]
         @destructive = local_assigns[:destructive]
         @name = local_assigns[:name]
         @value = local_assigns[:value]
@@ -77,6 +79,7 @@ module GovukPublishingComponents
         css_classes << "govuk-button--start" if start
         css_classes << "gem-c-button--secondary" if secondary
         css_classes << "gem-c-button--secondary-quiet" if secondary_quiet
+        css_classes << "govuk-button--secondary" if secondary_solid
         css_classes << "govuk-button--warning" if destructive
         css_classes << "gem-c-button--bottom-margin" if margin_bottom && !info_text
         css_classes << "gem-c-button--inline" if inline_layout
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add support for secondary solid buttons ([PR #1814](https://github.com/alphagov/govuk_publishing_components/pull/1814))
+
 ## 23.8.0
 
 * Add password reveal component ([PR #1794](https://github.com/alphagov/govuk_publishing_components/pull/1794))
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -52,6 +52,10 @@ examples:
     data:
       text: "Secondary quiet button"
       secondary_quiet: true
+  secondary_solid_button:
+    data:
+      text: "Secondary solid button"
+      secondary_solid: true
   destructive_button:
     data:
       text: "Destructive button"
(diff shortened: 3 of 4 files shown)


DilwoarH:
required by https://github.com/alphagov/collections/pull/2133

app/views/govuk_publishing_components/components/_button.html.erb
```diff
@@ -2,6 +2,9 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   local_assigns[:classes] = shared_helper.classes
   button = GovukPublishingComponents::Presenters::ButtonHelper.new(local_assigns)
+
+          #require 'pry'; binding.pry
```
alex-ju, line 6: we may want to remove the debugger

alex-ju (commented):
Code looks good (aside from the comment on debugger), but I'm a bit wary that we're adding a new design without explaining when we should be using the solid version and not the secondary. Is there any particular reasoning provided by the designer that we can add to the guidance? Also, I'm aware @andysellick had a meeting with the Design System and GOV.UK's Head of Design – we might be able to capture that decision when adding this new version to the library.

app/views/govuk_publishing_components/components/_button.html.erb
DilwoarH: ops

leenagupte:
@DilwoarH @alex-ju Shouldn't the style for the secondary button be replaced completely if it's going to match the design system rather than add another option? https://design-system.service.gov.uk/components/button/

DilwoarH:
> @DilwoarH @alex-ju Shouldn't the style for the secondary button be replaced completely if it's going to match the design system rather than add another option? https://design-system.service.gov.uk/components/button/

We got a bit of naming work to do to assign it with the Design system. The secondary button that we currently have is used within the admin interfaces so they would need to be migrated over

alex-ju:
> @leenagupte: Shouldn't the style for the secondary button be replaced completely if it's going to match the design system rather than add another option? https://design-system.service.gov.uk/components/button/

### Context
The current designs for secondary buttons (named 'secondary' and 'secondary quiet') in Publishing Components are used in a few applications (including Content Published, Collections Publisher) and were built before there was a 'secondary' variation in the Design System. When the Design System rolled out the secondary variation, designers in GOV.UK at the time felt that we shouldn't replace them straight away, as they tested well in research and the new ones were not coming with a solution for what we call 'secondary quiet'. Currently, the Design System is looking to address this 'secondary quiet' use-case, but we don't know how long will it take and how the new proposal will be received by designers in GOV.UK. A classic case of design debt I would say.

### Options
- If we do want to override the Publishing Components' secondary style with the Design System one, we should make sure that the affected products are ok with it (the least, check this with Product Manager) as it is, in my humble opinion, a visual breaking change.
- If we decide to create a new variation, we should also think of how will we converge this new 'secondary solid' with 'secondary' in the near future.

alex-ju (approved):
Technically correct. I will leave the decision to @DilwoarH and @leenagupte on whether or not to proceed with this approach or replace the current 'secondary' as explained in the comment above.

DilwoarH:
> Code looks good (aside from the comment on debugger), but I'm a bit wary that we're adding a new design without explaining when we should be using the solid version and not the secondary. Is there any particular reasoning provided by the designer that we can add to the guidance? Also, I'm aware @andysellick had a meeting with the Design System and GOV.UK's Head of Design – we might be able to capture that decision when adding this new version to the library.

> Technically correct. I will leave the decision to @DilwoarH and @leenagupte on whether or not to proceed with this approach or replace the current 'secondary' as explained in the comment above.

I think we (as in the community) need to sit together and decide how to proceed.

alex-ju:
> I think we (as in the community) need to sit together and decide how to proceed.

Sounds like a good approach. Given it's a piece of design debt we'd probably need designers in the room too.

leenagupte:
@alex-ju @DilwoarH I'm just happy that the change is being made in the gem. 💯

lib/govuk_publishing_components/presenters/button_helper.rb
```diff
@@ -77,6 +79,7 @@ def css_classes
         css_classes << "govuk-button--start" if start
         css_classes << "gem-c-button--secondary" if secondary
         css_classes << "gem-c-button--secondary-quiet" if secondary_quiet
+        css_classes << "govuk-button--secondary" if secondary_solid
```
leenagupte, line 82: Sorry if I'm being dense, but where is the new style being added? Won't this just add the existing green text version?

alex-ju: This comes from the Design System (`govuk-frontend` via npm): https://github.com/alphagov/govuk-frontend/blob/cf5be781776d7bb24394dd9211f672588ae081f6/src/govuk/components/button/_index.scss#L172-L202

PR #1818, opened 2020-12-10
Add NHS logo

chris-gds:
## What?

Adding NHS logo to gem.

## Why?

As part of [this PR](#toAdd) there was an update to convert bespoke org logo use to use the component `organisation_logo`

## Visual Changes

image

## Anything else?

Ideally this should be an SVG, current org implementation uses inline SVG which is problematic / doesn't render on IE8. 
Raised as [issue](#link)


alex-ju:
While I'm aware this is a draft PR, it's probably worth flagging that we don't usually consider assets as components but we do place them in [a shared space](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/images/govuk_publishing_components) then make them available to all applications using the [rails assets pipeline](https://github.com/alphagov/govuk_publishing_components/blob/5a8bb725c73eb4d2541266671173737ee6dc6bd1/config/initializers/assets.rb#L4).

chris-gds:
> While I'm aware this is a draft PR, it's probably worth flagging that we don't usually consider assets as components but we do place them in [a shared space](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/images/govuk_publishing_components) then make them available to all applications using the [rails assets pipeline](https://github.com/alphagov/govuk_publishing_components/blob/5a8bb725c73eb4d2541266671173737ee6dc6bd1/config/initializers/assets.rb#L4).

👋🏼  Interesting, thanks for taking a look - so would you typically _want_ to add this asset to the gem via the locations you've mentioned? Or would you prefer this type of logo get added to a location like:

`https://assets.publishing.service.gov.uk/government/uploads/system/uploads/organisation/logo/69/HMLR_logo.png`

alex-ju:
> 👋🏼 Interesting, thanks for taking a look - so would you typically _want_ to add this asset to the gem via the locations you've mentioned? Or would you prefer this type of logo get added to a location like:

If multiple apps shared the asset it makes sense for it to sit in this gem, otherwise in the app. I don't have much context on this specific use case as the PR we link to in the _Why_ section seems to be invalid.

chris-gds:
Closing draft PR - from conversation I don't think going into the gem is the correct step. Adding locallying within app [eg PR](https://github.com/alphagov/collections/pull/2148/)

PR #1823, opened 2020-12-11
Fix for search input misalignment

chris-gds:
## What?

Fix to resolve visual mis-alignment on certain uses of the search gem component

## Why

As part of [this PR](https://github.com/alphagov/govuk_publishing_components/pull/1793) this introduced an issue, I noticed there were other instances of the search that were being impacted by this change. This PR is a fix to resolve noticed issues.

## Visuals?

**Before** (left) **After** (right)
[Gem mobile use](https://components.publishing.service.gov.uk/component-guide/search) - fix
Screenshot 2020-12-11 at 13 59 09

Gem Desktop (use slight misalignment)
Screenshot 2020-12-11 at 13 57 53

Various other uses no visual change

Screenshot 2020-12-11 at 13 56 38

Screenshot 2020-12-11 at 13 56 24

Screenshot 2020-12-11 at 13 56 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix for search input misalignment ([PR #1823](https://github.com/alphagov/govuk_publishing_components/pull/1823))
+
 ## 23.9.1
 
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Works like a ⭐ dream ⭐ It looks like you have some linting failures but once those are fixed then I'm happy for this to go in.

PR #1830, opened 2020-12-16
Bump to 23.10.0

barrucadu:
* Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link ([PR #1811](https://github.com/alphagov/govuk_publishing_components/pull/1811))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 23.10.0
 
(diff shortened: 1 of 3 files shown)


injms (changes requested):
Looks good, just needs the file version in `Gemfile.lock` needs to also be bumped - for example: https://github.com/alphagov/govuk_publishing_components/pull/1827/files

brucebolt (commented):
Does the version also need increasing in `Gemfile.lock`?

injms (approved):

PR #1831, opened 2020-12-16
Use ActionController lazily

ethanmills:
This is an attempt to fix the following warning:
DEPRECATION WARNING: Initialization Being able to do this is deprecated. 
Autoloading during initialization is going
to be an error condition in future versions of Rails.
autoloaded the constant ApplicationHelper.

## What
Wrap the use of `ActionController` in `ActiveSupport.on_load(:action_controller)`

## Why
This line is causing a deprecation message in Rails. The full message is:

> DEPRECATION WARNING: Initialization autoloaded the constant ApplicationHelper.
> 
> Being able to do this is deprecated. Autoloading during initialization is going
> to be an error condition in future versions of Rails.
> 
> Reloading does not reboot the application, and therefore code executed during
> initialization does not run again. So, if you reload ApplicationHelper, for example,
> the expected changes won't be reflected in that stale Module object.
> 
> This autoloaded constant has been unloaded.
> 
> In order to autoload safely at boot time, please wrap your code in a reloader
> callback this way:
> 
>     Rails.application.reloader.to_prepare do
>       # Autoload classes and modules needed at boot time here.
>     end
> 
> That block runs when the application boots, and every time there is a reload.
> For historical reasons, it may run twice, so it has to be idempotent.
> 
> Check the "Autoloading and Reloading Constants" guide to learn more about how
> Rails autoloads and reloads.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 23.10.1
 
 * Fixes youtube title bug ([PR #1835](https://github.com/alphagov/govuk_publishing_components/pull/1835))
 
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (23.10.0)
+    govuk_publishing_components (23.10.1)
       govuk_app_config
       kramdown
       plek
--- a/lib/govuk_publishing_components/version.rb
+++ b/lib/govuk_publishing_components/version.rb
@@ -1,3 +1,3 @@
 module GovukPublishingComponents
-  VERSION = "23.10.0".freeze
+  VERSION = "23.10.1".freeze
 end


kevindew:
Hey @ethanmills, awesome stuff looking into this and raising this PR 🎉 

This seems to make sense to me, doing this would defer the execution of: 
https://github.com/alphagov/govuk_publishing_components/blob/8d80defa0bb1b531efedcf0ef0e04070787ab9d0/lib/govuk_publishing_components.rb#L41-L43

to not run until (at latest) this line is executed: 

https://github.com/alphagov/govuk_publishing_components/blob/8d80defa0bb1b531efedcf0ef0e04070787ab9d0/lib/govuk_publishing_components.rb#L52

And I'd guess the choice of using `ActiveSupport.on_load(:action_controller)` is preferable, in this case, to `Rails.application.reloader.to_prepare` because this is a Rails engine rather than an application and may run distinctly?

Do you know how we can replicate this deprecation warning? I've not seen it which makes it a little tricky to verify this does solve the problem? (I wondered if it was something that appeared with Rails 6.1 but couldn't see it flagged in one of our 6.1 apps that uses this gem)

ethanmills:
@kevindew we're based off of [this](https://github.com/DFE-Digital/govuk-rails-boilerplate), and I see the warning both when running tests and when running the app in development. I believe it first occurred in Rails 6.0

kevindew:
> @kevindew we're based off of [this](https://github.com/DFE-Digital/govuk-rails-boilerplate), and I see the warning both when running tests and when running the app in development. I believe it first occurred in Rails 6.0

Thanks, but I'm afraid that doesn't bring me any closer to replicating as that repo doesn't seem to reference this one. I've checked another project of ours running on Rails 6.0.3 and also haven't seen the warning so I'm wondering what is different.

ethanmills:
Sorry, I wasn't very clear. The deprecation warning appearing in the boilerplate app as soon as `gem "govuk_publishing_components"` is added to the gemfile. I'll create a minimal repro case later today

kevindew:
> Sorry, I wasn't very clear. The deprecation warning appearing in the boilerplate app as soon as `gem "govuk_publishing_components"` is added to the gemfile. I'll create a minimal repro case later today

Great - thank you 👍

ethanmills:
@kevindew Here you go: https://github.com/ethanmills/publishing_components_repro 

Instructions in the readme. It's about as minimal as I could make it with rails

kevindew:
> @kevindew Here you go: https://github.com/ethanmills/publishing_components_repro
> 
> Instructions in the readme. It's about as minimal as I could make it with rails

Super thanks. 

I was able to replicate the problem there. I now see the same in our apps. I hadn't realised that was something that isn't output during tests but just written to the log file. This seems a good fix, thank you 👍

kevindew:
@ethanmills sorry for closing this. I tried to rebase, move the comment and add a changelog so I could merge it and ended up syncing your branch with master. That was a rather fatal blow that automatically closed this and turned off my access to fix it 😆 . I've opened up https://github.com/alphagov/govuk_publishing_components/pull/1837 to get this merged in.

ethanmills:
Thanks for sorting this @kevindew

PR #1832, opened 2020-12-16
Update show password component tests

danacotoran:
## What

Update `show password` component spec to check that the type of the password visibility toggle button is "button".

Following on from this PR where I implemented the change but I forgot to update the test:
https://github.com/alphagov/govuk_publishing_components/pull/1826


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## 23.10.0
 
 * Add a data attribute to indicate a link or form should be decorated as if it were a cross-domain link ([PR #1811](https://github.com/alphagov/govuk_publishing_components/pull/1811))
+* Update show password component tests ([PR #1832](https://github.com/alphagov/govuk_publishing_components/pull/1832))
```
alex-ju, line 13: We may not need a changelog entry for this, as it's not affecting the API, it's not getting released anyway and doesn't provide useful information when updating the gem.

alex-ju (approved):

CHANGELOG.md
danacotoran: good point, might be overkill 😄

PR #1835, opened 2020-12-17
Fixes youtube title bug

DilwoarH:
# What

Replaces unstable method of setting the title to a stable method

# Why

The previous way was unstable and youtube changes these variables on a regular basis

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -105,8 +105,10 @@
         events: {
           onReady: function (event) {
             // update iframe title attribute once video is ready
+            // this is done to let screen reader users know that they are focused on a video
+            // https://github.com/alphagov/govuk_publishing_components/pull/908#discussion_r302913995
             var videoTitle = options.title
-            event.target.f.title = videoTitle + ' (video)'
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -105,8 +105,10 @@
         events: {
           onReady: function (event) {
             // update iframe title attribute once video is ready
+            // this is done to let screen reader users know that they are focused on a video
+            // https://github.com/alphagov/govuk_publishing_components/pull/908#discussion_r302913995
             var videoTitle = options.title
-            event.target.f.title = videoTitle + ' (video)'
+            event.target.getIframe().title = videoTitle + ' (video)'
```
alex-ju, line 111: Do we need to worry about the target not being a player object? for example, test that `getIframe` exists before accessing the title and such

alex-ju (approved):
Fixes the big. Wondering if we can improve further by making it more resilient.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
DilwoarH: Because it's inside the ready function - the iframe would always exist from what I understand

PR #1845, opened 2020-12-31
Replace countdown component with sidebar partial

benthorner:
https://trello.com/c/zshkcMQl/743-deploy-31-12-1045pm-update-brexit-transition-sidebar-section-3
https://trello.com/c/8w2ti59g/756-deploy-31-12-1045pm-replace-brexit-branded-sidebar-element-4

Originally we added code to make the transition countdown vanish
automatically, but since then we've also agreed to update content
for the component, but the new content only makes sense *after*
the countdown has vanished. In addition, while the original vanish
time was 11:59PM, most of the other updates are timed for 11PM.

This takes quite a drastic approach to simplifying the countdown
element, by removing it entirely, and merging the remainder of the
component into the contextual sidebar - it's only consumer. If we
don't want to make this change now, an alternative is to release
two new versions: one with an adjusted vanish time, and another with
the updated text (this will still need to be shipped manually).

Please see the commits for more details.

## Before

Screenshot 2020-12-31 at 14 06 39
Screenshot 2020-12-31 at 14 06 49
Screenshot 2020-12-31 at 14 07 01

## After

Screenshot 2020-12-31 at 14 05 09
Screenshot 2020-12-31 at 14 05 20
Screenshot 2020-12-31 at 14 08 25


sihugh (approved):
Tested locally with government-frontend and looks good to me. Thanks!

PR #1848, opened 2021-01-05
Tidy component filenames

andysellick:
## What / why

Couple of bits of tidying:

- rename the JS file for the `layout header` component to match component file naming conventions, was `header.js`. Before this change both the auditing and suggested JS tools were getting confused, as they assumed the JS file would be called `layout-header.js`. Only a few publishing applications, static and accounts are using this
- remove the sass file for the transition countdown component, as this was removed last week 
- rename a couple of component templates to use underscores, not hyphens (any app that includes these components e.g. `finder-frontend` will need to be updated to match the new filename)

Rather than cause a breaking change by renaming named assets and templates that applications rely on, temporary copies of the old files have been included in this PR. This means that the next version of the gem will not be a breaking change, and we don't have to panic rush to update applications to use the correct files. Thanks to @danacotoran for the suggestion.

## Visual Changes
None.


alex-ju (commented):
Thank you for cleaning up the CSS leftovers from https://github.com/alphagov/govuk_publishing_components/pull/1845.

Since this is going to be a breaking change, do we want to bundle other renames as well? to avoid multiple breaking releases for the same type of change. For example [the subscription links component](https://components.publishing.service.gov.uk/component-guide/subscription-links) uses a hyphen in the name (we have `subscription-links` instead of `subscription_links`) – we may have some others too, haven't investigated this. Similarly, we have `radio.js` instead of `radios.js`

andysellick:
Yep, subscription links and translation nav both had hyphens in the template name, which sadly, it turns out, was my fault 🙄 

Radio component consistently uses `radio` for filenames, not `radios`, so I think I'll leave it as it is.

app/assets/javascripts/govuk_publishing_components/components/header.js
```diff
@@ -0,0 +1,5 @@
+// This component relies on JavaScript from GOV.UK Frontend
+// = require govuk/components/header/header.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.Header = window.GOVUKFrontend
```
alex-ju: If we want to move forward with a backwards-compatible approach, I suggest we use symlinks so we avoid having duplicate files (e.g. `ln -s ~/govuk/govuk_publishing_components/app/assets/javascripts/govuk_publishing_components/components/layout-header.js ~/govuk/govuk_publishing_components/app/assets/javascripts/govuk_publishing_components/components/header.js`)

andysellick: That seems like a better approach but can we guarantee that path is correct in all environments? Small thing, but those paths don't match my local system, so it wouldn't work for me locally - although I'm much more concerned about production environments. I think a copy might be safer - this is only a temporary solution after all.

alex-ju: OK. Can we add a deprecation warning at the top of these files then? – might prevent people from changing them.

alex-ju (approved):

app/assets/javascripts/govuk_publishing_components/components/header.js
injms: Symbolic links can be relative - and using relative links would mean that it doesn't reference a path outside of the govuk_publishing_components folder.

andysellick: Will add a warning, good idea.

alex-ju: Indeed @injms – my example was rudimentary, I was hoping for a relative reference in the code

PR #1855, opened 2021-01-06
Version 23.12.0

andysellick:
## 23.12.0

* Get analytics code ready for use in production ([PR #1767](https://github.com/alphagov/govuk_publishing_components/pull/1767))


alex-ju (approved):
I see this as a minor as it ports this new feature (https://github.com/alphagov/static/pull/2367) otherwise it would've been a patch, I guess.

PR #1863, opened 2021-01-11
Hide native browser show password

andysellick:
## What / why

- Edge and IE11 now include their own 'show password' option in password inputs
- while this doesn't interfere with our custom functionality to do this, it is a little odd to have both of them there
- fortunately it can be removed using this bit of CSS

Background: https://textslashplain.com/2020/07/29/revealing-passwords/

## Visual Changes

Edge before:

Screenshot 2021-01-11 at 14 49 22

IE11 before:

Screenshot 2021-01-11 at 16 16 35

After is basically the same thing but without the little eye icon, e.g.

Screenshot 2021-01-11 at 16 20 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


Nooshu (approved):
Looks good to me @andysellick! 

I believe this functionality in IE may even predate IE11. IE10 by the look of it.

PR #1865, opened 2021-01-13
Add custom aria label option for layout header nav

danacotoran:
## What

Introduce the ability to pass a custom `aria-label` to the layout header navigation. 

Set the aria label to "Top level" by default – this was previously set on the `nav` element's child `ul`, however it makes more sense to add the `aria-label` to the `nav` itself, for screen reader users who navigate pages by navigation landmarks.

## Why

On the GOVUK Account there are a few pages with more than one `nav` element on the page. When there are multiple navigation landmarks of the same kind on a page, they should be labelled in order to help assistive tech users differentiate between them as per [WCAG SC 2.4.1: Bypass Blocks](https://www.w3.org/WAI/WCAG21/Understanding/bypass-blocks).

This change will allow us to be more descriptive in the way we label landmarks, and tailor these labels for different scenarios as needed.

## Visual Changes
No visual changes should occur as a result of this update

https://trello.com/c/1cnffruq

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/layout_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/layout_header.yml
@@ -60,6 +60,27 @@ examples:
       - text: News and communications
         href: "item-6"
         active: true
+  with_custom_navigation_aria_label:
+    description: The navigation has `aria-label="Top level"` by default. This option is here for when the `aria-label` needs to be more descriptive than that.
+    data:
+      search_left: true
+      navigation_aria_label: "Departments and policy"
+      navigation_items:
+      - text: Departments
+        href: "item-1"
+      - text: Worldwide
+        href: "item-2"
+      - text: How government works
+        href: "item-3"
+      - text: Get involved
+        href: "item-4"
+      - text: Consultations
+        href: "item-4"
+      - text: Statistics
+        href: "item-5"
+      - text: News and communications
+        href: "item-6"
+        active: true
   full_width:
     description: |
       This is difficult to preview because the preview windows are constrained, but the header will stretch to the size of its container.
--- a/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
@@ -1,7 +1,7 @@
 <% if navigation_items.any? %>
   <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-  <nav class="gem-c-header__nav">
-    <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
+  <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
+    <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end">
       <% navigation_items.each_with_index do |item, index| %>
         <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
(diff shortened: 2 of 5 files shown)


app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,7 +1,8 @@
+<%  navigation_items_label ||= "Top level" %>
```
andysellick, line 1: This default value for `navigation_items_label` could be here or in the `_layout_header` template. What's the benefit of having it here?

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,7 +1,8 @@
...
 <% if navigation_items.any? %>
   <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-  <nav class="gem-c-header__nav">
-    <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end" aria-label="Top Level Navigation">
```
andysellick, line 4: The default value for the aria-label is changing from `Top level navigation` to `Top level` - is that okay? Just want to check.

app/views/govuk_publishing_components/components/docs/layout_header.yml
```diff
@@ -60,6 +60,27 @@ examples:
       - text: News and communications
         href: "item-6"
         active: true
+  with_custom_navigation_aria_label:
+    description: The navigation has `aria-label="Top level"` by default. This option is here for when the `aria-label` needs to be more descriptive than that.
+    data:
+      search_left: true
+      navigation_items_label: "Departments and policy"
```
andysellick, line 67: At the risk of making this option name too verbose, should it include the word `aria`?

andysellick (commented):
Looks good, just a few questions.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
danacotoran: Fair question! I thought it might come up. 
The reason is that when a `nav` has `aria-label`, AT will read it out as **"YOUR_ARIA_LABEL_TEXT navigation"**.  Therefore when I originally left the default value as `Top level navigation`, I noticed that VoiceOver was reading it out as "Top level navigation navigation".
Testing with NVDA a similar redundancy occurred –  when the screen reader reaches the landmark, it states that it is a nav, so having the word "navigation" in the `aria-label` text results in repetition.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
danacotoran: Hm I think I added this here to provide a fallback in case `navigation_items_label` is ever not passed to `_navigation_items.html.erb`. But in hindsight I can probably get around this by adding a conditional around the `aria-label` where it's not added unless `navigation_items_label` exists. I will update the PR with that adjustment.

app/views/govuk_publishing_components/components/docs/layout_header.yml
danacotoran: I think we can include the word `aria` without making it too verbose. I'm thinking it can be changed to `navigation_aria_label` since the word "items" there doesn't really provide much value, IMO.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
andysellick: Good answer.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
danacotoran: Have updated so that the default value for `navigation_aria_label` is now set in `_layout_header.html.erb`.

andysellick (approved):

PR #1867, opened 2021-01-14
Add print styling for magna charta component

owenatgov:
## What
Adds `display: none` to print styling for specific elements within the magna charta bar chart component.

## Why
To address a bug raised for print styles where bar charts will duplicate data on print due to the chart element and the original table both being visible.

### Before
Screenshot 2021-01-14 at 10 03 22

### After
Screenshot 2021-01-14 at 10 01 31

[Card](https://trello.com/c/a1wOFE76/2006-markdown-charts-dont-print)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add print styling for magna charta component ([PR #1867](https://github.com/alphagov/govuk_publishing_components/pull/1867))
+
 ## 23.12.2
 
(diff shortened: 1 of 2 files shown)


injms:
Whilst the _perfect_ option would be to have printable charts, this is a good improvement - and the perfect shouldn't get in the way of the good shipping. 

There's a whole [issue about print stylesheets in the design system](https://github.com/alphagov/govuk-frontend/issues/573) that's worth a read as well.

injms (approved):

owenatgov:
@injms That's a good point. I've made a note to look into this as a little objective to see if I can style it to look pretty and also have a go at applying what the user has picked at the point of pressing print eg: if there are 2 charts and the user switches one to table, the print styles produce 1 table and 1 chart. Until then, it's merge time.

PR #1869, opened 2021-01-14
Change GOVUK Modules scope to document

DilwoarH:
## What

Change the scope to document rather than just body

## Why

This is done to support module within the head ie. on meta tags.

This also brings the scope in line with GOVUK Design system where document is used rather than the body.


alex-ju (commented):
Do you think we can capture this change in [the test](https://github.com/alphagov/govuk_publishing_components/blob/master/spec/javascripts/govuk_publishing_components/modules.spec.js#L9-L16) as well?

DilwoarH:
> Do you think we can capture this change in [the test](https://github.com/alphagov/govuk_publishing_components/blob/master/spec/javascripts/govuk_publishing_components/modules.spec.js#L9-L16) as well?

Hey @alex-ju  - I've added a few tests

alex-ju (approved):

PR #1873, opened 2021-01-18
Remove chevron banner component

alex-ju:
## What
Remove the [chevron banner](https://components.publishing.service.gov.uk/component-guide/chevron_banner) component.

## Why
We don't use it and we don't anticipate to be using it in future 🤞

## Notes
This is technically a breaking change, so I expect this to be merged and released no early than [24.0.0](https://github.com/alphagov/govuk_publishing_components/milestone/4).

## Visual Changes
n/a


andysellick (approved):

andysellick (commented):
Nearly forgot - this needs to include a changelog entry.

alex-ju:
Thanks, Andy! Added the changelog entry not to forget about it, but this will require a rebase anyway when we get closer to 24.0.0

PR #1874, opened 2021-01-18
Remove unneeded list stylesheet

injms:
## What

Removes the unnecessary list stylesheet. Updates the audit to cope with the list component not having a stylesheet.

## Why

The list styles are imported with the core parts of GOV.UK Frontend into the [component support stylesheet][1], so the standalone list stylesheet isn't required.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/c25fba341c13e270ae44e8d0add1d02d8e2c0404/app/assets/stylesheets/govuk_publishing_components/component_support.scss#L4

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,7 +18,7 @@
 * BREAKING: Retire gem-level SCSS variables ([PR #1881](https://github.com/alphagov/govuk_publishing_components/pull/1881))
 * BREAKING: Remove duplicate `lib/auto-track-event` script ([PR #1894](https://github.com/alphagov/govuk_publishing_components/pull/1894))
 * BREAKING: Remove copies of files ([PR #1878](https://github.com/alphagov/govuk_publishing_components/pull/1878))
-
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Good stuff. For reference, the following applications are using this component, and may or may not be already attempting to import this now deleted stylesheet.

```
collections, collections-publisher, content-publisher, datagovuk_find, finder-frontend, frontend, government-frontend, govuk-account-manager-prototype, govuk-coronavirus-vulnerable-people-form, whitehall
```

injms:
Checked the list of applications given by @andysellick - none of them imported the list stylesheet.

 - [x] collections
 - [x] collections-publisher
 - [x] content-publisher
 - [x] datagovuk_find
 - [x] finder-frontend
 - [x] frontend
 - [x] government-frontend
 - [x] govuk-account-manager-prototype
 - [x] govuk-coronavirus-vulnerable-people-form
 - [x] whitehall

alex-ju:
Nice! I think we can get this in now so we can release a major version.

PR #1875, opened 2021-01-18
Update design for step by step navigation to improve accessibility

owenatgov:
## What
Updates the design of the [step by step navigation component](https://components.publishing.service.gov.uk/component-guide/step_by_step_nav), specifically:

- Add an up/down chevron icon for improved visual affordance
- Move the show/hide all button at the top of the component over to the left so that it's easier for zoomed screen users to use
- Alter spacing

## Why
This is part of an effort to align the different types of accordion across govuk to follow a single style and reduce confusion for users and meet [WCAG SC 3.2.4](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-consistent-functionality.html).

### Before
Screenshot 2021-01-18 at 16 51 55

### After
Screenshot 2021-01-18 at 18 44 01

[Card](https://trello.com/c/S4Yr0CgY/606-update-step-by-step-accordion)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -333,34 +333,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var that = this
 
     this.$module.showOrHideAllButton.addEventListener('click', function (event) {
-      var shouldshowAll
-      var showHideTexts = that.$module.querySelectorAll('.js-toggle-link')
       var textContent = this.textContent || this.innerText
+      var shouldShowAll = textContent === that.$module.actions.showAllText
 
-      if (textContent === that.$module.actions.showAllText) {
-        that.$module.showOrHideAllButton.innerHTML = that.$module.upChevronSvg + that.$module.actions.hideAllText
-        for (var i = 0; i < showHideTexts.length; i++) {
-          showHideTexts[i].innerHTML = that.$module.upChevronSvg + that.$module.actions.hideText
-        }
-        shouldshowAll = true
+      // Fire GA click tracking
+      stepNavTracker.trackClick('pageElementInteraction', (shouldShowAll ? 'stepNavAllShown' : 'stepNavAllHidden'), {
+        label: (shouldShowAll ? that.$module.actions.showAllText : that.$module.actions.hideAllText) + ': ' + that.$module.stepNavSize
+      })
 
-        stepNavTracker.trackClick('pageElementInteraction', 'stepNavAllShown', {
-          label: that.$module.actions.showAllText + ': ' + that.$module.stepNavSize
-        })
-      } else {
-        that.$module.showOrHideAllButton.innerHTML = that.$module.downChevronSvg + that.$module.actions.showAllText
-        for (var j = 0; j < showHideTexts.length; j++) {
-          showHideTexts[j].innerHTML = that.$module.downChevronSvg + that.$module.actions.showText
-        }
-        shouldshowAll = false
-
-        stepNavTracker.trackClick('pageElementInteraction', 'stepNavAllHidden', {
-          label: that.$module.actions.hideAllText + ': ' + that.$module.stepNavSize
-        })
-      }
-
-      that.setAllStepsShownState(shouldshowAll)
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -325,20 +337,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var showHideTexts = that.$module.querySelectorAll('.js-toggle-link')
       var textContent = this.textContent || this.innerText
 
+      console.log(textContent)
```
alex-ju, line 340:
```suggestion

```

andysellick (commented):
Looks good. One question - isn't there a test for clicking on the show/hide links and does it need updating? (I'm guessing not since the tests are passing)

alex-ju:
> Looks good. One question - isn't there a test for clicking on the show/hide links and does it need updating? (I'm guessing not since the tests are passing)

The JS tests are using the 'locally defined' markup which we might need to update fro things like `data-show-all-text` and potentially testing the newly introduced JS logic

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -114,7 +126,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var data = []
 
     for (var i = 0; i < this.$module.steps.length; i++) {
-      var stepView = new this.StepView(this.$module.steps[i], this.$module.actions.showText, this.$module.actions.hideText)
+      var stepView = new this.StepView(this.$module.steps[i], this.$module.actions.showText, this.$module.actions.hideText, this.$module.upChevronSvg, this.$module.downChevronSvg)
```
alex-ju, line 129: These calls are starting to look shabby – I suggest we refactor to reduce it to `StepView(this.$module)` and call elements down the tree within the `StepView` function definition as needed.

owenatgov:
I'll check the tests out and make sure they're still robust 👍

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
owenatgov: 🤦

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
owenatgov: I've gone for a similar suggestion but kept the initial element parameter as the way this is passed differs from declaration to declaration and amending the code to cleanly meet just passing in `this.$module` each time would add extra verbosity. Good idea all the same, nice and tidy.

alex-ju (approved):
Code looks good! Assuming cross-browser testing I'm happy for this change to get in.

app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
```diff
@@ -40,6 +40,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$module.sessionStoreLink = this.$module.sessionStoreLink + '_' + this.$module.uniqueId
     }
 
+    this.$module.upChevronSvg = '<svg width="20" height="20" viewBox="0 0 20 20" fill="none" xmlns="http://www.w3.org/2000/svg" class="gem-c-step-nav__chevron">' +
```
chris-gds, line 43: More of a FYI comment than anything else, I've created and referenced these as image assets right now. 

There might be a cross-over point where we can share assets.

owenatgov: I've chosen this just based on what I've observed in the rest of them gem but I'm not precious about which implementation we use at the moment. Both would work cross browser. I think down the line we should agree on an implementation and align accordingly. For me, performance and any potential accessibility implications are going to be the smoking guns here. So far I haven't given this a deep screen reader assessment ie: navigating into the show button and trying to navigate the accessibility tree, but I can try it out.

I personally think lets get this into the gem now and update as and when. Happy to chat about it though.

owenatgov:
I've browser tested this and run it through VO, NVDA and JAWS and I'm happy with it. Along with the review above, I'm going to make the bold move to merge in.

PR #1878, opened 2021-01-19
Remove copies of files

andysellick:
## What
- removing duplicate files
- some file copies were added to prevent a breaking change from occurring when the originals were renamed to match component conventions
- e.g. `_subscription-links.html.erb` is now `_subscription_links.html.erb`, but we kept the original to stop applications from breaking when they upgraded

Original commit to add these duplicates is here: https://github.com/alphagov/govuk_publishing_components/pull/1848/commits/d339809fab2a84af0ae9929414da03e0430770fc

## Why
We've now either merged or have PRs in progress for all applications that reference these components, so once they've been merged it should be safe to create a new version of the gem with this change in. However, even if it doesn't actually break any applications (if I've made all the other changes correctly) I've still marked this as a breaking change because I think it's important to highlight this to anyone reviewing dependabot PRs, just in case.

## Visual Changes
None.


andysellick:
I've pushed a change to `CHANGELOG` but it isn't showing up in the PR yet. Apparently github might be having issues, can someone check this has worked and update if not before merging? Thanks.

alex-ju (approved):

PR #1883, opened 2021-01-20
Update document list component with description text modifier and amended design

owenatgov:
## What
A few updates to [the document list component](https://components.publishing.service.gov.uk/component-guide/document_list):

1. Adds a `full_size_description` modifier to `link` objects that increases the size of description text to the standard 19.
2. Amends the design of list items.
3. Remove some unnecessary margin from the `subtext` element

## Why
This is prompted by work within the govuk accessibility team whilst assessing if we should replace the [list of links on the contacts page](https://www.gov.uk/contact) with the document list component. We've decided to proceed with this on the condition that we make a couple of design amends for the new context, the principal change being allowing for full size description text.

The design change to list items is something that has been planned within the design community for some time but hasn't been given the attention it needs. This design has been agreed upon and reviewed within said community.

The subtext change is just to reduce style verbosity a bit. The visual change, in my opinion, is so insignificant here that I won't be including it in the visual changes.

### Before
Screenshot 2021-01-20 at 17 25 49

### After (with description modifier AND new design)
Screenshot 2021-01-20 at 17 41 49

[Card](https://trello.com/c/Hx76hGvU/554-update-feedback-app-to-use-components)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Update document list component with description text modifier and amended design ([PR #1883](https://github.com/alphagov/govuk_publishing_components/pull/1883))
+
 ## 23.13.1
 
 * Bump govuk-frontend from 3.9.1 to 3.10.2 ([PR #1838](https://github.com/alphagov/govuk_publishing_components/pull/1838))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -7,14 +7,10 @@
 
 .gem-c-document-list__item {
   overflow: hidden;
-  margin-bottom: govuk-spacing(4);
-  padding-bottom: govuk-spacing(4);
-  border-bottom: 1px solid $govuk-border-colour;
+  margin-bottom: govuk-spacing(5);
+  padding-top: govuk-spacing(2);
+  border-top: 1px solid $govuk-border-colour;
   list-style: none;
-
-  &:last-child {
-    border-bottom: 0;
-  }
 }
 
 .gem-c-document-list__item-title {
@@ -52,7 +48,7 @@
 }
 
 .gem-c-document-list__subtext {
-  margin: govuk-spacing(1) 0 0 0;
+  margin: 0;
 }
 
 .gem-c-document-list__item-description,
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -60,6 +60,10 @@
   @include govuk-font($size: 16, $line-height: 1.5);
 }
 
+.gem-c-document-list__item-description--full-size {
+  @include govuk-font($size: 19);
+}
+
 .gem-c-document-list__item-metadata {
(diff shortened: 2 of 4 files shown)


alex-ju:
Are we fine with the visual impact of this change across GOV.UK (frontend apps & publishing apps)? I'm only asking as I see that we now prefer asymmetric spacing on such elements, which is different from how we used to define such components. Is there something that we want to formalize across components (e.g. 2:5 top/bottom ratio). I'm concerned about the vertical rhythm on a page and how this change affects it, but I'm not a designer so happy to obey.

owenatgov:
@alex-ju I've passed your comment onto the designer for a more informed answer. I can try to answer your point on spacing now re: those 2 components.

- The accordion is currently going through a resdesign to align it with our step by step, which the design system team have agreed to adopt. The designer for the accordion work is the same one leading on this PR and the new accordion implements a similar 2:5 ratio.
- The summary list is a blind spot. I'm happy to run this past the designer and get their view. As this is part and parcel from the design system I reckon discussing this with them is a good idea s well.

alex-ju:
> * The accordion is currently going through a resdesign to align it with our step by step, which the design system team have agreed to adopt. The designer for the accordion work is the same one leading on this PR and the new accordion implements a similar 2:5 ratio.

Ah, that's good to know. Thanks for sharing the info! Will proceed with code review then.

alex-ju (approved):
Code looks good. Not sure about the `full_size_description` variable name, but I can't come up with a better suggestion.

miaallers:
> Are we fine with the visual impact of this change across GOV.UK (frontend apps & publishing apps)? I'm only asking as I see that we now prefer asymmetric spacing on such elements, which is different from how we used to define such components. Is there something that we want to formalize across components (e.g. 2:5 top/bottom ratio; looking for consistency with [accordion](https://components.publishing.service.gov.uk/component-guide/accordion) or [summary-list](https://components.publishing.service.gov.uk/component-guide/summary_list)). My concerned is around the vertical rhythm on a page and how this change affects it, but I'm not a designer so happy to obey.

This change to the vertical rhythm is why we have implemented the asymmetric spacing. The spacing does the following:
- Having a smaller space between the title and the border reinforces the relationship between those two things. They are clearly part of the same element
- Having a larger space after each list element reinforces this again

The current equidistant spacing makes it difficult to know where a list item begins and ends. Instead of clear 'chunks' of content that the users eye can skip over, we have a long list of undifferentiated things.

With regards to formalising this across components, this would be a literal dream come true and I think would bring a great deal of clarity to the overall design of GOV.UK. Could we do this??

alex-ju:
Thank you for providing the reasoning behind this change.

owenatgov:
I get the impression that everyone involved is happy so I'm going to hit merge. @alex-ju I agree on the attribute name, it's a bit of a crumby choice but I too can't think of anything better. Right now it's only being used in one place so if we are hit by variable name inspiration we can use this to act quickly.

DilwoarH:
Hey @owenatgov @alex-ju

I think this broke the step by step lists:

Screenshot 2021-02-05 at 10 16 09

owenatgov:
@DilwoarH Do you have al ink to that page? I'd have thought that that shouldn't be a docs list component since it appears to be in a table.

alex-ju:
> Are we fine with the visual impact of this change across GOV.UK (frontend apps & publishing apps)?

Thanks for flagging this, Dilwoar. This is the reason I asked if it's been tested with Publishing apps as well.

PR #1884, opened 2021-01-20
Update design of accordion component

chris-gds:
## What?

Porting the [Design System's Accordion](https://design-system.service.gov.uk/components/accordion/) component into the [gem](https://components.publishing.service.gov.uk/component-guide) and updating the Design to improve a11y and usability.

## Why?

GOV.UK currently uses 3 different accordions on different parts of the site.  Some of them are not as accessible as they could be and they are not consistent in the way they open and close (the language they use). [This is a fail of WCAG SC 3.2.4.](https://www.w3.org/TR/UNDERSTANDING-WCAG20/consistent-behavior-consistent-functionality.html)
This update is part of on-going work to improve and create a consistent design that _might_ eventually feed into the Design System depending on further research / outcome of performance. ([Ticket](https://trello.com/c/RqBAMKhP))

## Visual Changes

Before (left) After (right)

Screenshot 2021-02-08 at 09 54 33

Screenshot 2021-01-28 at 19 10 41

Interactions demo

design-updates-accordion

## Assistive Technology Testing

<details>
  <summary>Testing Videos, Notes & Discussion</summary>
  

https://user-images.githubusercontent.com/71266765/107202332-feee5880-69f1-11eb-8255-9f33e8f2fd8b.mov

(user adjusts colour reference videos)
https://user-images.githubusercontent.com/71266765/106749352-52d9f580-661e-11eb-85fc-1f78fb368f8e.mov

https://user-images.githubusercontent.com/71266765/106749366-55d4e600-661e-11eb-97b5-c174e851b923.mov

https://user-images.githubusercontent.com/71266765/106749370-566d7c80-661e-11eb-99d2-809ed422d51a.mov

https://user-images.githubusercontent.com/71266765/106749373-57061300-661e-11eb-8a4a-60cc9cbc231e.mov

https://user-images.githubusercontent.com/71266765/106749377-57061300-661e-11eb-8d4f-48d52ada308a.mov

IE11
image

- The heading reads the "Show / Hide this section" copy that is potentially redundant or confusing information to the user. This ties into comments from the original Accordion below:

_The ‘Open all’ button reads out as ‘Open all sections’ for screen readers. This is potentially confusing for users as the visual content is different to what screen readers read out._

After discussion with team it was noted that this was based on the tested step-by-step and while not perfect is not an WCAG fail.

- The change in focus states was queried if this would impact WCAG 3.2.4 as it's not consistent with other elements.

The team decided this was not a WCAG concern.

300% zoom 

Screenshot 2021-02-04 at 14 21 08

Possible Improvements

- The Chevron could offer a visual cue on hover (after team discussion it was decided the current the affordance is fine)

- Once the Accordion is opened, some work was done to auto-focus the inner Content once expanded for AT, I removed it as part of this PR as out of scope.

-  Buttons receiving the “hand cursor” has been noted as not strictly best practice. Needs discussion.  

</details>

## Anything else

- PR was raised as Draft as on-going Design iterations were on-going in parallel

 - [Closed PR on collections](https://github.com/alphagov/collections/pull/2257) to ensure Design updates cascade through to uses on other repos.

---

- [Step-by-step update](https://github.com/alphagov/govuk_publishing_components/pull/1875) also implements this new Design in parallel (to resolve WCAG 3.2.4)

 - Porting over from Design System `common.js` polyfill for `nodeListForEach` 
Conversation [about altering this referenced here](https://github.com/alphagov/govuk-frontend/issues/1017): 

- [Further conversations about this element within the Design System are happening here](https://github.com/alphagov/govuk-design-system-backlog/issues/1)

- There is also a [different prototype here](https://blue-badge.cloudapps.digital/can-i-get-a-blue-badge)

- [Issue raised as a reminder](https://github.com/alphagov/govuk_publishing_components/issues/1897) to **remove** this Gem Isolation after Design system updates to this element to stay inline with current architecture  

- Gem Doc's were updated to explain what Experimental means


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,282 @@
-// This component relies on JavaScript from GOV.UK Frontend
```
owenatgov, line 1: Not a criticism but a note: We need to remember to revert this back to a reference to the govuk frontend script once the design system team eventually adopt our accordion, as has been agreed. Is it worth adding a comment here to this effect?

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
+  }
+
+  .gem-c-accordion__open-all {
+
+    position: relative;
+    z-index: 1;
+    border-width: 0;
+    color: $govuk-link-colour;
+    background: none;
+    cursor: pointer;
```
owenatgov, line 36: Is this necessary as the assumption is that this class will always only be applied to a button? There might be some browser implications which I'm not aware of.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
-}
 
-.govuk-accordion--condensed {
-  .govuk-accordion__section-button {
-    @include govuk-media-query($from: tablet) {
-      @include govuk-font($size: 19, $weight: bold);
+  // Focus state to change chervon icon to black
+  .gem-c-accordion__open-all:focus {
+    .gem-c-accordion-nav__chevron--stroke {
+      stroke: govuk-colour("black");
```
owenatgov, line 55: This is a really good idea! I'll implement this into my step by step stuff. Question though, does this work across browsers? I have this vague memory of SVG styling being dodgy on IE.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
+    // Safe area on the right to avoid clashing with icon
+    padding-right: 40px;
+    border-top: $gem-c-accordion-bottom-border-width solid $govuk-border-colour;
+    color: $govuk-link-colour;
+    cursor: pointer;
+  }
+
+  // For devices that can't hover such as touch devices,
+  // remove hover state as it can be stuck in that state (iOS).
+  @media (hover: none) {
```
owenatgov, line 79: Ooooooooo...

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
+  // Align the arrow icon within the button
+  .gem-c-accordion-nav__chevron {
+    margin-right: govuk-spacing(1);
+    vertical-align: text-top;
+    display: inline-block;
+  }
+
+  // Buttons within the headers don’t need default styling
+  .gem-c-accordion__section-button {
+
```
owenatgov, line 95: Gap!

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
+    @include govuk-typography-common;
+    margin-top: 0;
```
owenatgov, line 97: It doesn't look like `margin-right` has been set anywhere here so you could just do `margin: 0` for all these margin declarations. Unless there's something I'm not clear on for why these are being defined this way?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -25,23 +25,27 @@
...
-        section_classes << 'govuk-accordion__section--expanded' if item[:expanded]
+        section_classes = %w(gem-c-accordion__section)
+        section_classes << 'gem-c-accordion__section--expanded' if item[:expanded]
 
-        summary_classes = %w(govuk-accordion__section-summary govuk-body)
+        summary_classes = %w(gem-c-accordion__section-summary govuk-body)
       %>
-      <%= tag.div(class: section_classes) do %>
-        <div class="govuk-accordion__section-header">
+      <%= tag.section(class: section_classes) do %>
```
owenatgov, line 33: nice 👍

owenatgov (commented):
Couple of comments for the draft, looking good so far!

Few extra bits of feedback:

- You've included the js for the design system accordion but [not the test](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/components/accordion/accordion.test.js). Could we add this in please?
- My step by step design updates are imminent and this should be applied to the accordion. I think this just translates to extra spacing and reduced border widths. I'm aware you're working with Mia on this 👍

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
owenatgov: It works across browsers! Great stuff.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: 🤔 Good point.  I'm think as this will be a piece of work to convert the gem back to this Design System I'm thinking this will be implied from other components use.   I'm not sure about a `TODO` of this nature with this, might be better placed within a card?

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
...
+}
+
+.gem-c-accordion__section-header {
+  padding-top: govuk-spacing(3);
+  padding-bottom: govuk-spacing(3);
+}
+
+.js-enabled {
+  .gem-c-accordion {
+
```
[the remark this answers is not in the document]
chris-gds: Thanks - I'm not sure why `styelint` ignores this gap and not others. I'll remove.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: Design System port not mine - but a f'sure, I had a similar "Oooo" reaction  😁

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: Thanks, so initially I had used svg background assets & pseudo elements for the icons however I decided to adopt an implementation more inline with step-by-step with the images so I could inline for this reason...

I started to think about various other techniques with inlining svgs within the gem, there might be an existing helper or pattern that exists for this as it's usually _really_ common.  

It would ideally be best to share 1 icon across the accordion + step-by-step and simply use CSS (rotate etc) to reduce the need for assets.  However this the scope of this, it felt isolated per component (also, I felt a way about images within JS)

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,282 @@
...
+//  = require ../vendor/polyfills/indexOf.js
+//  = require ../vendor/polyfills/common.js
+
 window.GOVUK = window.GOVUK || {}
-window.GOVUK.Modules = window.GOVUK.Modules || {}
-window.GOVUK.Modules.Accordion = window.GOVUKFrontend
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function GemCAccordion () { }
```
andysellick, line 10: This name seems a bit clunky - we only use the `gem-c` prefix for CSS classes. I think `GemAccordion` is fine - I assume `Accordion` is already taken?

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,231 @@
 @import "govuk/components/accordion/accordion";
```
andysellick, line 1: Looks like we've got no instances of `govuk-frontend` accordion styles in the markup anymore. Can this import be removed?

app/assets/javascripts/govuk_publishing_components/components/accordion.js
owenatgov: [Could we raise it as an issue?](https://github.com/alphagov/govuk_publishing_components/issues)

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
owenatgov: My instinct on this is for now™️ to keep them separate and componentised, then once we've launched this, the step by step and figured out the manuals accordion we can work on a combined solution that suits all our needs. I appreciate that making things good from the start is probably the right thing to do however as the implementation for the wider accordion solution hasn't quite finished baking yet, I'd rather launch stuff than not.

chris-gds: Agreed. Let's iterate on these a bit later.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: Ah so this was ported over and is inline with other button patterns on govuk, but! this might be a bigger topic of _should_ a button switch to a hand cursor as it's not strictly a link.

Ref [Link 1](https://medium.com/simple-human/buttons-shouldnt-have-a-hand-cursor-b11e99ca374b), [Link 2](https://ux.stackexchange.com/questions/105024/why-dont-button-html-elements-have-a-css-cursor-pointer-by-default), [Link 3](https://twitter.com/feather/status/559756555441434624?s=20)

I've left this for the moment, I'd rather they alll be the same rather than only these button behave different but it might be one worth raising as an issue and discussing wider

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: Sound good, will do.

injms:
A couple of visual things:

 * when focused and hovered, the heading text appears to be purple/dark blue instead of black whilst the show/hide text is black; I think that the heading text should be black for the best contrast.
 * the yellow background takes up all the space, but the screenshots on the pull request description show yellow only taking a portion of the accordion. Which is correct?
 * the grey border is still present during the focus state, sandwiched between the yellow and the black. I think that this should be removed when focused.

Screenshot 2021-01-28 at 11 15 56

chris-gds:
> A couple of visual things:
> 
> * when focused and hovered, the heading text appears to be purple/dark blue instead of black whilst the show/hide text is black; I think that the heading text should be black for the best contrast.
> * the yellow background takes up all the space, but the screenshots on the pull request description show yellow only taking a portion of the accordion. Which is correct?
> * the grey border is still present during the focus state, sandwiched between the yellow and the black. I think that this should be removed when focused.
> 
> Screenshot 2021-01-28 at 11 15 56

+1 to all of these, the Design has been updated since this was initially raised so will all be addressed.

chris-gds:
There were some considerations noted during Design iterations that are also in sync within [step-by-step](https://github.com/alphagov/govuk_publishing_components/pull/1893)

@injms:
> My worry with the clickable area being so big is that the blank space on the right gives no indication that it's a link and is far enough away from a link to not be thought of as interactive.

Screenshot 2021-01-28 at 15 12 12

@miaallers:

> It's in line with SBS as @owenatgov points out, and this wasn't flagged as an issue during testing either with DAC or in other usability testing. Obviously we can never be 100% certain but the component has been in use for some time now


I noted:  The spacing change between the text and the icon. The icon now doesn't jump but the difference in spacing (which is expected, should be fine but a ponder)

@miaallers 
>  It's not perfect, but nothing is perfect. I think it's fine

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: Added: https://github.com/alphagov/govuk_publishing_components/issues/1897

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: Ah - makes sense, `GemAccordion` used instead.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,268 @@
...
+
+  // Section headers have a pointer cursor as an additional affordance
+  .gem-c-accordion__section-header {
+    padding: govuk-spacing(2) 0 govuk-spacing(5);
+    position: relative;
+    border-top: $gem-c-accordion-bottom-border-width solid $govuk-border-colour;
+    color: $govuk-link-colour;
+  }
+
+  .gem-c-accordion__section-header--focused {
```
owenatgov, line 86: It just occurred to me that you're applying this as a class via js. Can this not be simply applied to a `:focus` state? I'm aware the focus won't be necessary if the js isn't on but if the js ever breaks post markup generation then the focus state will also be broken. I remember I had problems with this but was able to cobble it together with some creative spacing changes, check out my step by step PR or I can talk you through it.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,283 @@
...
+    this.moduleId = this.$module.getAttribute('id')
+    this.sections = this.$module.querySelectorAll('.gem-c-accordion__section')
+    this.openAllButton = ''
+    this.browserSupportsSessionStorage = helper.checkForSessionStorage()
+    this.controlsClass = 'gem-c-accordion__controls'
+    this.openAllClass = 'gem-c-accordion__open-all'
+    this.openAllTextClass = 'gem-c-accordion__open-all-text'
+    this.openAllExpandedClass = 'gem-c-accordion__open-all--expanded'
+    this.sectionHeaderClass = 'gem-c-accordion__section-header'
+    this.sectionHeaderFocusedClass = 'gem-c-accordion__section-header--focused'
```
owenatgov, line 23: See my below comment in the css re: this.

owenatgov (commented):
Few more comments. Tested cross browser and looking good. We're making good headway.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: So, this is due to the focus state visuals changing to the parent element.  The button is being focused semantically but visually the outer element is receiving the styles.  This think this could be refactored however I'm a little hesitate as the Design could be iterated on.

owenatgov: I hear you on the design iteration point, we've both chosen implementations that aren't perfect but do the job. I would argue that mine has less user risk even if it's a little messier and trickier to unpick upon change. I wonder if there's value in getting another dev to step in and take a look at our choices and see what they reckon.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,283 @@
...
+    // Get a count of all Accordion sections that are expanded
+    var expandedSectionCount = this.$module.querySelectorAll('.' + this.sectionExpandedClass).length
+    var areAllSectionsOpen = sectionsCount === expandedSectionCount
+
+    return areAllSectionsOpen
+  }
+
+  // Update "Show all sections" button
+  GemAccordion.prototype.updateOpenAllButton = function (expanded) {
+    var newButtonText = expanded ? '<span class="' + this.openAllTextClass + '">Close all sections</span>' + this.upChevronSvg : '<span class="' + this.openAllTextClass + '">Show all sections</span>' + this.downChevronSvg
```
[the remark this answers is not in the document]
chris-gds: Agreed, i'll sort this.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: >we've both chosen implementations that aren't perfect but do the job.

This, I'm mindful that if the end goal is for this visual change to inform the singular point of truth: the Design System _and_ once there is more critical user & a11y clinic feedback there could be more significant iterations.  

In this context I'm thinking it might be more desirable to sort of perfect a _little_ later when we know "yes this is as good as it can be" type thing..

> I wonder if there's value in getting another dev to step in and take a look at our choices and see what they reckon.

I think so, that way if there are any glaring issues I / we can sort.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,283 @@
...
+  // Set individual header attributes
+  GemAccordion.prototype.initHeaderAttributes = function (headerWrapper, index) {
+    var module = this
+    var span = headerWrapper.querySelector('.' + this.sectionButtonClass)
+    var heading = headerWrapper.querySelector('.' + this.sectionHeadingClass)
+    var summary = headerWrapper.querySelector('.' + this.sectionSummaryClass)
+
+    // Copy existing span element to an actual button element, for improved accessibility.
+    var button = document.createElement('button')
+    button.setAttribute('id', this.moduleId + '-heading-' + (index + 1))
```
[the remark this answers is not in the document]
chris-gds: As this existed when ported, wondering if [desired convention.](https://gds-way.cloudapps.digital/manuals/programming-languages/js.html#javascript-coding-style) - leaving for now

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
owenatgov: With that in mind let's leave it for now. It's a low risk and a more refined solution is probably useful. Feel free to resolve.

owenatgov:
We're looking really good! A couple of little nitpicks from testing:

### Blockers
Spacing is a bit wonky on Firefox on the condensed accordion and on mobile
Screenshot 2021-02-02 at 11 55 39
Screenshot 2021-02-02 at 11 56 09

### Potential problems
On voiceover, there's a weird flow after you open an accordion. I haven't dug into the markup but I suspect that this is to do with the hide text not being removed upon click. See video below.
https://user-images.githubusercontent.com/64783893/106597949-fd82e300-654e-11eb-90d1-c0a1f3aa6aaf.mov

### Non-blockers
The focus state on voiceover where it gives stuff the little border is out of line with the focus state
Screenshot 2021-02-02 at 11 57 45

Icons look a bit squashed on mobile for safari and firefox
Screenshot 2021-02-02 at 11 54 52

chris-gds:
> ### Blockers
> Spacing is a bit wonky on Firefox on the condensed accordion and on mobile

As per our convo this was a little tricky to replicate, this can be viewed by:

Condensed layout on Desktop + macOS + Firefox + User Sets minimum font-size to 16px

Screenshot 2021-02-02 at 15 30 42

Otherwise it's OK (below) I tend to lean to this being an edge case but will tweak.

Screenshot 2021-02-02 at 14 23 38

> ### Potential problems
> On voiceover, there's a weird flow after you open an accordion. I haven't dug into the markup but I suspect that this is to do with the hide text not being removed upon click. See video below.

I think you might be referring to "Hide this section" being heard twice?

> ### Non-blockers
> The focus state on voiceover where it gives stuff the little border is out of line with the focus state

I really want to have these focus states reviewed at the a11y clinic I'm a bit concerned users at 300% or zoomed in more might get confused about these large yellow blocks.

> Icons look a bit squashed on mobile for safari and firefox

Can resolve this one.

chris-gds:
Converting PR back into Draft to gain more feedback.

While the visual updates address the initial WCAG issue of 3.2.4, I am not confident that other Accordion concerns raised around WCAG 1.3.1 have been resolved.

chris-gds:
Discussion outcome notes Added to PR.

owenatgov:
This is looking pretty good. I tested it just now and I'm still getting the issue I raised in my last comment where "show this section"/"hide this section" is heard twice. Digging into the markup, I can see the reason is that the actual section under each heading (`gem-c-accordion__section-content`) has an `aria-labelledby` of its associated heading. Removing it seems to fix this.

I'm interested to know if this aria label is necessary. The step by step for example doesn't include it. Is this one for the DS team?

chris-gds:
Riight I understand your comment now.  I remember requesting comments for this issue the other day (about hearing it twice). The general outcome from the team was AT users are used to filtering _some_ noise.

So it’s needed as it defines / link the relationship of the inner Content to its respective Heading. It's currently [on the DS version](https://design-system.service.gov.uk/components/accordion/default/index.html) and you see a [similar pattern here](https://www.w3.org/TR/wai-aria-practices-1.1/examples/accordion/accordion.html). It's been introduced with the new "Show this section"

I do agree, _I'd_ prefer to resolve, I'll have a think with this one.

chris-gds:
@owenatgov I resolved a few issues in this last update.
- I've altered the `aria` label so it pulls in just the header copy, this resolves the issue of the button being heard twice
- The structure was refactored, focus on the wider element now behaves better (this fixes the mismatch of AT alignment)
- I've also converted the icons into CSS and adjusted, which makes this a lot cleaner.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,25 +1,298 @@
...
+    }
+
+    .gem-c-accordion__section-button {
+      @include govuk-typography-responsive($size: 19, $important: true);
+      padding-top: govuk-spacing(1);
+      padding-bottom: govuk-spacing(5);
+    }
+
+    // Reduce Chevron size
+    .gem-c-accordion-nav__chevron {
```
owenatgov, line 267: I love this!!

owenatgov (approved):
I'm a big fan of the last couple of changes you've made, very nice. I reckon you could squash your commits together a bit before merging though. Also maybe wait to merge or check in govuk frontenders as I know @alex-ju is putting out a breaking release today and as this is quite delicate we may want to wait a bit.

Once this is out I'd like to confer on differences between this and the step by step and see if we can bring in some of your technical choices eg: the css chevron, possibly moving towards bringing these close together, as is the dream.

Farewell accordion country 😃

chris-gds:
Thanks @owenatgov

> squash your commits together a bit before merging though.

Will do, I thought to myself the differences might be a bit easier to review this way to start - then the squash™

> I know @alex-ju is putting out a breaking release today and as this is quite delicate we may want to wait a bit.

Ah OK, will do - i'll ping and hold off for the moment.

> Once this is out I'd like to confer on differences between this and the step by step and see if we can bring in some of your technical choices eg: the css chevron, possibly moving towards bringing these close together, as is the dream.

Yup this works, whoop, leaving Accordion country, now boarding 😃

PR #1894, opened 2021-01-28
Remove duplicate auto-track-event script

alex-ju:
## What
Remove duplicate `lib/auto-track-event.js` script

## Why
After the analytics migration from `static` to this gem we ended up with an identical copy of this script that can be removed to slim down the JS bundle. We'll preserve the [`analytics/auto-track-event.js` version](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/auto-track-event.js) which is made [available to all frontend applications via `static`](https://github.com/alphagov/static/commit/0405ca1f1d424b2acd1571ec1297f6d7b3d3f7dc)

## Visual Changes
No visual changes


andysellick (approved):
Good spot 👍

PR #1898, opened 2021-02-01
Fix footer classes

andysellick:
## What / why
The main `class` declaration for the footer component was missing a closing `"`. Then I decided to make it more readable as well.

## Visual Changes
None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_footer.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_footer.html.erb
@@ -2,8 +2,10 @@
   meta ||= []
   navigation ||= []
   with_border ||= false
+  classes = %w(gem-c-layout-footer govuk-footer)
+  classes << "gem-c-layout-footer--border" if with_border
 %>
-<footer class="gem-c-layout-footer govuk-footer<%= " gem-c-layout-footer--border" if with_border %>" role="contentinfo">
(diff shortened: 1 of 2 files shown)


AlanGabbianelli (approved):

danacotoran (approved):
Thanks for fixing this

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -2,8 +2,10 @@
   meta ||= []
   navigation ||= []
   with_border ||= false
+  classes = %w(gem-c-layout-footer govuk-footer)
+  classes << "gem-c-layout-footer--border" if with_border
 %>
-<footer class="gem-c-layout-footer govuk-footer <%= "gem-c-layout-footer--border" if with_border %> role="contentinfo">
+<footer class="<%= classes.join(' ') %>" role="contentinfo">
```
alex-ju, line 8: Non-blocker, suggestion: if we use the tag helper we get the `join` for 'free' and the attribute sanitised.
```rb
<%= tag.footer class: css_classes, role: "contentinfo" do %>
   ...
<% end %>

andysellick: THAT was the thing I meant to include as part of this change, thanks @alex-ju. Added now 🤦

PR #1899, opened 2021-02-01
Fix deprecation warnings when running tests

andysellick:
## What / why
Running the tests for the gem locally some deprecation warnings were being thrown up. They were:

- `action_view.raise_on_missing_translations` is deprecated
- rendering a template with a `.` in it is deprecated

The first one was an easy fix, just swapping out the old for the new. The second one was a little more complicated. The attachment component has several partials that are `.svg` files but are rendered like partials. I renamed these files to `.html.erb` and removed the file extension from the `render` call, which removed the deprecation warning but suddenly caused one of the component audit tests to fail. 

This was because the formerly `.svg` files were now being recognised by the auditor as potential components that should be included when suggesting Sass/JS for applications, even though they were in a subdirectory for partials and not the main directory location for component templates. The fix for this was to amend the regex that matches this to remove the `/` character when matching components. So if this was a template...

```
<%= render "govuk_publishing_components/components/component" %>
<%= render "govuk_publishing_components/components/partials/partial" %>
```

Previously the auditor was identifying both lines as components, when only the first line is a true component. The question I'm left with is why the regex originally included the `/` character when matching... it seems a bit odd. I've checked it locally with collections and it seems to still produce the right suggested Sass/JS output, but perhaps some more thorough checking would be useful.

I think this only really happened with components within components but I think one of the publishing apps might pull in a partial from the header component directly for some reason, I'm hoping someone can shed some light on this.

The specific deprecation warnings were:

```
DEPRECATION WARNING: action_view.raise_on_missing_translations is deprecated and will be removed in Rails 6.2. 
Set i18n.raise_on_missing_translations instead. Note that this new setting also affects how missing translations are handled in controllers. 
(called from block (2 levels) in <top (required)> at /Users/andysellick/workspace/govuk/govuk_publishing_components/spec/component_guide/component_audit_spec.rb:5)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_spreadsheet.svg 
(called from _app_views_layouts_application_html_erb__3522064166250592246_71200 at /Users/andysellick/workspace/govuk/govuk_publishing_components/spec/dummy/app/views/layouts/application.html.erb:16)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_spreadsheet.svg 
(called from _app_views_layouts_application_html_erb__3522064166250592246_71200 at /Users/andysellick/workspace/govuk/govuk_publishing_components/spec/dummy/app/views/layouts/application.html.erb:16)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_generic.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:48)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_generic.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:48)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_document.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:44)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_document.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:44)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_spreadsheet.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:46)

DEPRECATION WARNING: Rendering actions with '.' in the name is deprecated: 
govuk_publishing_components/components/attachment/_thumbnail_spreadsheet.svg 
(called from block (3 levels) in ___sers_andysellick_workspace_govuk_govuk_publishing_components_app_views_govuk_publishing_components_components__attachment_html_erb___2448659267357681147_74020 at /Users/andysellick/workspace/govuk/govuk_publishing_components/app/views/govuk_publishing_components/components/_attachment.html.erb:46)
```

## Visual Changes
None.


app/views/govuk_publishing_components/components/_attachment.html.erb
```diff
@@ -41,11 +41,11 @@
                 "aria-hidden": true,
                 data: data_attributes do %>
       <% if attachment.document? %>
-        <%= render "govuk_publishing_components/components/attachment/thumbnail_document.svg" %>
+        <%= render "govuk_publishing_components/components/attachment/thumbnail_document" %>
```
alex-ju, line 44: Out of curiosity, do you know if by renaming the files from `thumbnail_document.svg` to `_thumbnail_document.svg.erb` we would achieve the same result?

alex-ju (approved):

app/views/govuk_publishing_components/components/_attachment.html.erb
andysellick: That hadn't occurred to me, interesting question. I don't know, no, but I think the change to the auditing to only match `/component/name/` and not `/component/partial/partial` is probably good anyway.

andysellick: Might delay merging this to investigate if I can find time.

Issue #1889, opened 2021-01-25
The feedback component looks broken when JavaScript is disabled


alex-ju:
The feedback component presents a 'Thank you...' message and a 'There is something wrong with this page' button (that doesn't do anything) when JavaScript is disabled. The layout is also different from the JavaScript-enabled version, presenting a narrow button. In such scenario (JavaScript disabled) I wonder if there's any value in presenting any of the elements in the blue area.

<img width="957" alt="Screenshot 2021-01-25 at 10 19 57" src="https://user-images.githubusercontent.com/788096/105692881-e6ad1280-5ef6-11eb-8acb-687ab0c25c5e.png">

Issue #1825, opened 2020-12-14
Click area of buttons in feedback component span outside the element's visible boundaries


alex-ju:
The click area of buttons in [feedback component](https://components.publishing.service.gov.uk/component-guide/feedback) span outside the element's visible boundaries.

I suspect this is a remnant from when these buttons were actually links and there was no boundary on the element.

<img width="1436" alt="Screenshot 2020-12-14 at 10 49 20" src="https://user-images.githubusercontent.com/788096/102073119-c15abe00-3dfa-11eb-9162-74ca74f04a13.png">

PR #1900, opened 2021-02-01
Fix bugs on the feedback component

owenatgov:
## What
Alters some minor styling for the feedback component.

## Why
This is to address https://github.com/alphagov/govuk_publishing_components/issues/1825 and https://github.com/alphagov/govuk_publishing_components/issues/1889. The first issue is addressed by removing a now unnecessary `after` styling rule. The second by moving the js hiding styles to the bottom of the stylesheet to prevent them being undermined by style rules that cascade later in the stylesheet.

### Before - 1825
Screenshot 2021-02-01 at 14 40 01

### After - 1825
Screenshot 2021-02-01 at 14 39 37

### Before - 1889
Screenshot 2021-02-01 at 14 40 51

### After - 1889
Screenshot 2021-02-01 at 14 40 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix bugs on the feedback component ([PR #1900](https://github.com/alphagov/govuk_publishing_components/pull/1900))
+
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
Thank you for addressing these issues, @owenatgov!

PR #1901, opened 2021-02-01
Update document list documentation

owenatgov:
## What
Updates the document list documentation to specify what a document is in the context of this component plus a few other little touchups.

## Why
Whilst working on govuk accessibility there was some confusion about if this should be used for a particular use case on the grounds that nobody was clear what a "document" was eg: there was some thought that it should be used for downloadable ODTs/literal documents. I raised this in a govuk frontenders catchup and folks agreed it was fine as it is and is being used suitably already but maybe a way to mitigate confusion would be to update the docs to be clear about what a document actually is.

The ordered list line has been removed as this is no longer presented in an ordered list.

Docs only so no changelog update.

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/document_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/document_list.yml
@@ -1,16 +1,16 @@
 name: Document list
 description: An ordered list of documents including a document type, when updated and a link.
 body: |
-  Outputs a list to documents, based on an array of document data. This may include:
+  Outputs a list to documents, based on an array of document data. A "document" in this context can be an asset (such as a ODT or other downloadable document) or a web page.
+  
+  The component can display:
 
-  * document title
-  * link to the document
-  * last updated date object
-  * document type
+  * a document title
+  * a link to the document
+  * a last updated date object
+  * a document type
 
   Tracking can be added to the links by supplying optional data attributes for each.
-
-  Documents are presented in an ordered list as the component expects that the ordering of the documents is relevant.
 accessibility_criteria: |
   The component must:
 


app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -1,16 +1,16 @@
 name: Document list
 description: An ordered list of documents including a document type, when updated and a link.
 body: |
-  Outputs a list to documents, based on an array of document data. This may include:
+  Outputs a list to documents, based on an array of document data. A "document" in this context is considered any page on the web, so the document list can be considered a list of references (links, titles etc) to web pages.
+  
+  This component may include:
```
andysellick, line 6: I think this would be better written as `This may include` as it's referring to the list of things, rather than the component?

app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -1,16 +1,16 @@
 name: Document list
 description: An ordered list of documents including a document type, when updated and a link.
 body: |
-  Outputs a list to documents, based on an array of document data. This may include:
+  Outputs a list to documents, based on an array of document data. A "document" in this context is considered any page on the web, so the document list can be considered a list of references (links, titles etc) to web pages.
```
andysellick, line 4: Is it "any page on the web" or any page on GOV.UK? Can I instead suggest:

`A "document" in this context can be an asset (such as a PDF or other downloadable document) or a web page.`

app/views/govuk_publishing_components/components/docs/document_list.yml
owenatgov: I read this list as elements that can be part of the component. We've just been talking about something else (what a document is) so I thought it was helpful to qualify what this list is, otherwise people will think this list expands on what a document is. Have I gone astray?

app/views/govuk_publishing_components/components/docs/document_list.yml
owenatgov: I'm going to amend this to `ODT` instead of `PDF` as a micro effort towards throwing PDFs away since they're inaccessible.

app/views/govuk_publishing_components/components/docs/document_list.yml
andysellick: That's a good point, hadn't thought of that. How about `The component can display:`?

andysellick (approved):

PR #1902, opened 2021-02-02
Update layout header component

danacotoran:
## What

Add data attribute support on the layout header navigation links. 

The reason for this is because we are intending to use the `layout_header` component for the transition checker nav whose links are cross-domain links and need data attributes for tracking.

## Why

During development the markup for the account nav (appears on the transition page, and transition checker) was copied over from the layout header component instead of using the component itself. 

Screenshot 2021-02-02 at 18 30 08

The component audits are now complaining that we are including the assets for `layout_header` in `static`, when we are not using the component.

To fix this problem and reduce repetition, we are removing the hard coded account navigation from static, and using the layout header component in its place: alphagov/static#2408

## Visual Changes

https://trello.com/c/9Kjy1hjb

### DIFF
--- a/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
@@ -1,16 +1,21 @@
 <% if navigation_items.any? %>
   <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-  <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
+  <%= tag.nav class: "gem-c-header__nav", aria: { label: navigation_aria_label } do %>
     <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end">
       <% navigation_items.each_with_index do |item, index| %>
-        <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
-          <%= "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu] %>">
+        <%
+          li_classes = %w(govuk-header__navigation-item)
+          li_classes << "govuk-header__navigation-item--active" if item[:active]
+          li_classes << "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu]
+        %>
+        <%= tag.li class: li_classes do %>
           <%= link_to(
             item[:text],
             item[:href],
             class: 'govuk-header__link',
+            data: item[:data],
           ) %>
-        </li>
+        <% end %>
       <% end %>
     </ul>
   <% end %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Fix bugs on the feedback component ([PR #1900](https://github.com/alphagov/govuk_publishing_components/pull/1900))
 * Add data attributes to layout footer component ([PR #1904](https://github.com/alphagov/govuk_publishing_components/pull/1904))
 * Add option to remove border from document list items ([PR #1907](https://github.com/alphagov/govuk_publishing_components/pull/1907))
+* Update layout header component ([PR #1902](https://github.com/alphagov/govuk_publishing_components/pull/1902))
 
 ## 23.14.0
 
--- a/app/views/govuk_publishing_components/components/docs/layout_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/layout_header.yml
@@ -81,6 +81,17 @@ examples:
       - text: News and communications
         href: "item-6"
         active: true
+  with_navigation_link_data_attributes:
+    description: Supports adding data attributes i.e for tracking
+    data:
+      navigation_items:
+      - text: Departments
+        href: "item-1"
+        data:
+          module: "a custom attribute"
+          something_else: "some other custom attribute"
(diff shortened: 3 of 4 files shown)


app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,17 +1,24 @@
 <% if navigation_items.any? %>
-  <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-  <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
+  <div class="govuk-header__navigation-container">
+    <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
+    <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
```
alex-ju, line 4: Non-blocker, stylistic: aware this line wasn't introducted in this PR, but we don't need the rounded brackets here and I don't see the value in this short conditional statement `navigation_aria_label ? navigation_aria_label : nil`, I think we can contract it.
```suggestion
    <%= tag.nav class: "gem-c-header__nav", aria: { label: navigation_aria_label } do %>
```

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,17 +1,24 @@
...
     <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end">
       <% navigation_items.each_with_index do |item, index| %>
-        <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
-          <%= "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu] %>">
+        <%
+          li_classes = %w(govuk-header__navigation-item)
+          li_classes << "govuk-header__navigation-item--active" if item[:active]
+          li_classes << "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu]
+        %>
+        <%= tag.li(class: li_classes) do %>
```
alex-ju, line 12: Non-blocker, stylistic
```suggestion
        <%= tag.li class: li_classes do %>
```

alex-ju (approved):

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,17 +1,24 @@
 <% if navigation_items.any? %>
-  <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
-  <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
+  <div class="govuk-header__navigation-container">
```
andysellick, line 2: What's the need for this element? I assume the styles are coming from `govuk-frontend` - was this an omission that you're correcting?

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,17 +1,24 @@
...
+    <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
+    <%= tag.nav(class: "gem-c-header__nav", 'aria-label': navigation_aria_label ? navigation_aria_label : nil ) do %>
     <ul id="navigation" class="govuk-header__navigation govuk-header__navigation--end">
       <% navigation_items.each_with_index do |item, index| %>
-        <li class="govuk-header__navigation-item <%= "govuk-header__navigation-item--active" if item[:active] %>
-          <%= "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu] %>">
+        <%
+          li_classes = %w(govuk-header__navigation-item)
+          li_classes << "govuk-header__navigation-item--active" if item[:active]
+          li_classes << "govuk-header__navigation-item--collapsed-menu-only" if item[:show_only_in_collapsed_menu]
```
andysellick, line 10: Nice cleaning 👍

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,17 +1,24 @@
...
           <%= link_to(
             item[:text],
             item[:href],
             class: 'govuk-header__link',
+            data: item[:data],
           ) %>
-        </li>
+        <% end %>
       <% end %>
     </ul>
```
andysellick, line 21: Nitpick: does this whole `ul` block need indenting?

andysellick (approved):
Looks good, only some minor comments.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
danacotoran: Good shout. No need for this element! I originally thought it was necessary to add an extra container around the `nav` in order to achieve what I needed but I can remove it now.

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
danacotoran: Thanks for the suggestion, it looks much cleaner.

PR #1903, opened 2021-02-03
Add rel attribute option to document list

owenatgov:
## What
Add a `rel` option to the document list component which gets applied to the markup as an attribute on the document link.

## Why
We use the `rel` attribute in govuk content to dictate links to external sites, principally for use by search engines. This has emerged as a requirement for the document list component where it's being used to link to external sites. [Read more about what the rel attribute is and how it can be used here](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel).

No visual changes.

### DIFF
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -29,12 +29,21 @@
           item_classes = "gem-c-document-list__item-title #{brand_helper.color_class} #{title_with_context_class if item[:link][:context]}"
 
           if item[:link][:path]
+            rel = [
+              "external",
+              "nofollow",
+              "noopener",
+              "noreferrer",
+              "opener",
+            ].include?(item[:link][:rel]) ? item[:link][:rel] : nil
+
             link_to(
               item[:link][:text],
               item[:link][:path],
               data: item[:link][:data_attributes],
               class: "#{item_classes} gem-c-document-list__item-link",
-              lang: item[:link][:locale].presence
+              lang: item[:link][:locale].presence,
+              rel: rel,
             )
           else
             content_tag(
--- a/app/views/govuk_publishing_components/components/docs/document_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/document_list.yml
@@ -305,3 +305,21 @@ examples:
           path: '/become-an-apprentice'
           description: 'Becoming an apprentice - what to expect, apprenticeship levels, pay and training, making an application, complaining about an apprenticeship.'
           full_size_description: true
+  with_rel_link_attribute:
+    description: |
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/docs/document_list.yml
```diff
@@ -288,3 +288,15 @@ examples:
           path: '/become-an-apprentice'
           description: 'Becoming an apprentice - what to expect, apprenticeship levels, pay and training, making an application, complaining about an apprenticeship.'
           full_size_description: true
+  with_rel_link_attribute:
+    description: |
+      The [rel attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) is an option attribute to dictate the relationship between the document being linked to and the current page. This is predominantly used by search engines to help index journeys through a website.
+      
+      On GOV.UK, this is typically used to dictate and external website being linked to with `rel="external"`.
```
andysellick, line 295: Should this be `to dictate an external website`

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -33,7 +33,8 @@
               item[:link][:path],
               data: item[:link][:data_attributes],
               class: "#{item_classes} gem-c-document-list__item-link",
-              lang: item[:link][:locale].presence
+              lang: item[:link][:locale].presence,
+              rel: item[:link][:rel].presence,
```
andysellick, line 37: Are we happy to allow anything to be passed in here? We might be, I'm uncertain.

andysellick (commented):
Looks good. You could squash all the commits into one though.

app/views/govuk_publishing_components/components/_document_list.html.erb
owenatgov: That's a good point. I don't think there are any use cases for anything other than `external` right now. Would it be valuable to replace this with just an external attribute? Eg: `external: true`. which would render the `rel` attribute.

andysellick: I don't know what else we might be using but I looked up the list and there's a few that might be useful, e.g. nofollow.

I was thinking something like this https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_fieldset.html.erb#L8

owenatgov: That's probably a bit more extensible. So here's a proposal: I implement that fieldset solution and do one of the following:

1. Just include `external` for now
2. Include everything in that table that the `a` tag specifies as an annotation

andysellick: Sounds good.

alex-ju (approved):

PR #1904, opened 2021-02-03
Add data attributes to layout footer component

danacotoran:
## What 
The `layout_footer` currently used in Accounts does not have the same data attributes for tracking as the footer on the main site. 

Add data attributes for click tracking to the layout footer component to match the attributes of the hard coded footer in static.
I hard coded `data-track-category` to `footerClicked` as it the value of this attribute should be the same for every link in the footer.

## Why
So that we can have click event tracking on the GOVUK Account footer. 

Goes hand in hand with this [pull request](https://github.com/alphagov/govuk-account-manager-prototype/pull/665).

------

https://trello.com/c/9A6UPqi8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix bugs on the feedback component ([PR #1900](https://github.com/alphagov/govuk_publishing_components/pull/1900))
+* Add data attributes to layout footer component ([PR #1904](https://github.com/alphagov/govuk_publishing_components/pull/1904))
 
 ## 23.14.0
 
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
I'm personally reluctant to embedding tracking in the component, but it seems like we use `data-module="track-click"` in various components all over the library so I can't say no to one more. We also don't have a framework for dealing with tracking yet, so this is probably a good change to have consistent tracking on the footer component for the moment.

danacotoran:
@alex-ju  thanks for approving. I'm with you on that, I don't love the idea of having it hardcoded in the component but when I looked it up it seemed to be embedded in components everywhere so I thought I'd stick with the established way of doing things.

PR #1905, opened 2021-02-04
Adds Reorderable lists component

DilwoarH:
## What

A list of items that can be reordered

List items can be reordered by drag and drop or by using the up/down buttons. On small viewports the drag and drop feature is disabled to prevent being triggered when scrolling on touch devices.

This component uses SortableJS - a JavaScript library for drag and drop interactions. When JavaScript is disabled a set of inputs will be shown allowing users to provide an order index for each item.

## Why

This is used by publishing tools to allow reordering of lists such as steps in step by step, entries in timelines and order of visual presentation of attachments.

## Disclaimer

This is a port/extension of the work that @alex-ju has done on the content publisher app.

## Demo

https://govuk-publis-add-reorde-g116bw.herokuapp.com/component-guide/reorderable_list

## Visual Changes

screenshot-localhost_3212-2021 03 03-11_22_34

### No JS view

Screenshot 2021-02-05 at 18 41 03

https://trello.com/c/SD4pLfen/1086-add-reorder-component-to-govukpublishingcomponents

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Remove lists from summary action links ([PR #1956](https://github.com/alphagov/govuk_publishing_components/pull/1956))
 * Fix GOV.UK Frontend deprecation warning for component-guide print stylesheet ([PR #1961](https://github.com/alphagov/govuk_publishing_components/pull/1961))
 * Update search box button ([PR #1957](https://github.com/alphagov/govuk_publishing_components/pull/1957))
+* Adds Reorderable lists component ([PR #1905](https://github.com/alphagov/govuk_publishing_components/pull/1905))
 
 ## 24.4.1
 
--- a/app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js
@@ -1,0 +1,108 @@
+//= require sortablejs/Sortable.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ReorderableList () { }
+
+  ReorderableList.prototype.start = function ($module) {
+    this.$module = $module[0]
+    this.$upButtons = this.$module.querySelectorAll('.js-reorderable-list-up')
+    this.$downButtons = this.$module.querySelectorAll('.js-reorderable-list-down')
+
+    this.sortable = window.Sortable.create(this.$module, { // eslint-disable-line new-cap
+      chosenClass: 'gem-c-reorderable-list__item--chosen',
+      dragClass: 'gem-c-reorderable-list__item--drag',
+      onSort: function () {
+        this.updateOrderIndexes()
+        this.triggerEvent(this.$module, 'reorder-drag')
+      }.bind(this)
+    })
+
+    if (typeof window.matchMedia === 'function') {
+      this.setupResponsiveChecks()
+    } else {
+      this.sortable.option('disabled', true)
+    }
+
+    var boundOnClickUpButton = this.onClickUpButton.bind(this)
+    this.$upButtons.forEach(function (button) {
+      button.addEventListener('click', boundOnClickUpButton)
+    })
+
+    var boundOnClickDownButton = this.onClickDownButton.bind(this)
+    this.$downButtons.forEach(function (button) {
+      button.addEventListener('click', boundOnClickDownButton)
+    })
+  }
+
+  ReorderableList.prototype.setupResponsiveChecks = function () {
+    var tabletBreakpoint = '40.0625em' // ~640px
+    this.mediaQueryList = window.matchMedia('(min-width: ' + tabletBreakpoint + ')')
+    this.mediaQueryList.addListener(this.checkMode.bind(this))
+    this.checkMode()
+  }
+
+  ReorderableList.prototype.checkMode = function () {
+    this.sortable.option('disabled', !this.mediaQueryList.matches)
+  }
+
+  ReorderableList.prototype.onClickUpButton = function (e) {
+    var item = e.target.closest('.gem-c-reorderable-list__item')
+    var previousItem = item.previousElementSibling
+    if (item && previousItem) {
+      item.parentNode.insertBefore(item, previousItem)
+      this.updateOrderIndexes()
+    }
+    // if triggered by keyboard preserve focus on button
+    if (e.detail === 0) {
+      if (item !== item.parentNode.firstElementChild) {
+        e.target.focus()
+      } else {
+        e.target.nextElementSibling.focus()
+      }
+    }
+    this.triggerEvent(e.target, 'reorder-move-up')
+  }
+
+  ReorderableList.prototype.onClickDownButton = function (e) {
+    var item = e.target.closest('.gem-c-reorderable-list__item')
+    var nextItem = item.nextElementSibling
+    if (item && nextItem) {
+      item.parentNode.insertBefore(item, nextItem.nextElementSibling)
+      this.updateOrderIndexes()
+    }
+    // if triggered by keyboard preserve focus on button
+    if (e.detail === 0) {
+      if (item !== item.parentNode.lastElementChild) {
+        e.target.focus()
+      } else {
+        e.target.previousElementSibling.focus()
+      }
+    }
+    this.triggerEvent(e.target, 'reorder-move-down')
+  }
+
+  ReorderableList.prototype.updateOrderIndexes = function () {
+    var $orderInputs = this.$module.querySelectorAll('.gem-c-reorderable-list__actions input')
+    $orderInputs.forEach(function (input, index) {
+      input.setAttribute('value', index + 1)
+    })
+  }
+
+  ReorderableList.prototype.triggerEvent = function (element, eventName) {
+    var params = { bubbles: true, cancelable: true }
+    var event
+
+    if (typeof window.CustomEvent === 'function') {
+      event = new window.CustomEvent(eventName, params)
+    } else {
+      event = document.createEvent('CustomEvent')
+      event.initCustomEvent(eventName, params.bubbles, params.cancelable)
+    }
+
+    element.dispatchEvent(event)
+  }
+
+  Modules.ReorderableList = ReorderableList
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/govuk_publishing_components/components/_reorderable-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_reorderable-list.scss
@@ -1,0 +1,117 @@
+.gem-c-reorderable-list {
+  @include govuk-font(19, bold);
+
+  list-style-type: none;
+  margin-bottom: govuk-spacing(6);
+  margin-top: 0;
+  padding-left: 0;
+  position: relative;
+
+  .govuk-form-group {
+    margin-bottom: 0;
+  }
+}
+
+.gem-c-reorderable-list__item {
+  margin-bottom: govuk-spacing(3);
+  border: 1px solid $govuk-border-colour;
+  padding: govuk-spacing(3);
+}
+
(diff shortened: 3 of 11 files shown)


kevindew:
Nice one putting this together Dilwoar.

I do still think there needs to be more consolidation between the two implementations rather than having options for all. Nikin is available to have a meeting tomorrow to work towards this consolidation.

@alex-ju do you have any recollections for why Content Publisher made particular design decisions and or opinions on any parts where the collections-publisher designer is an improvement? as I understand the difference's we've got are:

1. Green and grey secondary buttons (Collections Publisher) vs grey secondary buttons (Content Publisher) - seems fine and non-divisive to go with just grey secondary buttons
2. Draggable elements in boxes (Collections Publisher) vs border bottom line (Content Publisher) - this seems the most divisive with a few people stating that boxes are easier to understand for drag and drop and probably the biggest decision
3. Whether to have numbers or not - these seem to be confusing and they seem reasonable to be implemented as an option, though I do wonder whether they're needed at all?

alex-ju:
> Nice one putting this together Dilwoar.
> 
> I do still think there needs to be more consolidation between the two implementations rather than having options for all. Nikin is available to have a meeting tomorrow to work towards this consolidation.
> 
> @alex-ju do you have any recollections for why Content Publisher made particular design decisions and or opinions on any parts where the collections-publisher designer is an improvement? as I understand the difference's we've got are:
> 
> 1. Green and grey secondary buttons (Collections Publisher) vs grey secondary buttons (Content Publisher) - seems fine and non-divisive to go with just grey secondary buttons
> 2. Draggable elements in boxes (Collections Publisher) vs border bottom line (Content Publisher) - this seems the most divisive with a few people stating that boxes are easier to understand for drag and drop and probably the biggest decision
> 3. Whether to have numbers or not - these seem to be confusing and they seem reasonable to be implemented as an option, though I do wonder whether they're needed at all?

Hi both,

I don't have strong opinions on this, but I can provide a bit of background being involved with both:
 - Collections Publisher implementation was a rough improvement of an existing implementation in a rush to replace the old publishing interface for step-by-steps; this was tested internally with a limited number of users.
 - Content Publisher implementation was developed afterwards, was build from scratch and iterated a few times; this was tested with external stakeholders in a round of research.

Some of my design-related recollections of this during the Content Publisher times, in relation to the above points:
 - grey secondary buttons were used not to stand out as we found that users were less inclined to take the button-based approach to reorder
 - a boxes-based approach was considered but we realised it involved more vertical spacing – which became a problem on documents with lots of attachments as users needed to combine drag-and-drop with scroll and things end up messy
 - users didn't explicitly complain about the numbers, but some colleagues were confused by them after starting the reorder

I share the view that having as fewer options as possible would be ideal.

DilwoarH:
@kevindew - this is now ready to review

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Add spacing to cookie banner confirmation message ([PR #1936](https://github.com/alphagov/govuk_publishing_components/pull/1936))
 * Fix Sass warning for extending a compound selector ([PR #1933](https://github.com/alphagov/govuk_publishing_components/pull/1933))
 * Add default aria label for navigation items (PR [#1946](https://github.com/alphagov/govuk_publishing_components/pull/1946))
+* Adds Reorder-able lists component ([PR #1905](https://github.com/alphagov/govuk_publishing_components/pull/1905))
```
kevindew, line 15: We refer to this as "reorderable list" component in the component documentation so this should match that naming.

app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js
```diff
@@ -0,0 +1,108 @@
+//= require sortablejs/Sortable.js
```
kevindew, line 1: As this file comes from the node_modules directory we'll need to include that in packaging of the gem: https://github.com/alphagov/govuk_publishing_components/blob/fba876a49e6512e339b4c27719cd6f1b017db364/govuk_publishing_components.gemspec#L17

app/views/govuk_publishing_components/components/_reorderable_list.html.erb
```diff
@@ -0,0 +1,44 @@
...
+    <%= tag.li class: "gem-c-reorderable-list__item" do %>
+      <%= tag.div class: "gem-c-reorderable-list__wrapper" do %>
+        <%= tag.div class: "gem-c-reorderable-list__content" do %>
+          <%= tag.p item[:title], class: "gem-c-reorderable-list__title" %>
+          <%= tag.p(item[:description], class: "gem-c-reorderable-list__description") if item[:description].present? %>
+        <% end %>
+        <%= tag.div class: "gem-c-reorderable-list__actions" do %>
+          <%= render "govuk_publishing_components/components/input", {
+            label: {
+              text: sanitize("Position<span class='govuk-visually-hidden'> for #{item[:title]}</span>")
```
kevindew, line 19: I recall that we did the work that we shouldn't need sanitize here in collections-publisher: https://github.com/alphagov/collections-publisher/blob/ca14ae8058f4115fe60eaefba8e4102a2c8ae19e/app/views/components/_reorderable_list.html.erb#L17-L19

Though I saw it crept back in a couple of lines later 🙈 . I imagine we can still get rid of it with using capture.

spec/javascripts/components/reorderable-list-spec.js
```diff
@@ -0,0 +1,206 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK, spyOnEvent */
```
kevindew, line 2: I think this can be simplified to `/* global spyOnEvent */` as we did in collections publisher.

app/views/govuk_publishing_components/components/docs/reorderable_list.yml
```diff
@@ -0,0 +1,76 @@
+name: Reorderable list
+description: A list of items that can be reordered
+body: |
+  List items can be reordered by drag and drop or by using the up/down buttons.
+  On small viewports the drag and drop feature is disabled to prevent being triggered
+  when scrolling on touch devices.
+
+  This component uses SortableJS - a JavaScript library for drag and drop interactions.
+  When JavaScript is disabled a set of inputs will be shown allowing users to provide
+  an order index for each item.
```
kevindew, line 10: Do you think we should document the structure of the data submitted by this component? I'm wondering if that would be a tricky thing about knowing how to use this component in an app.

I guess we could say something like:

When this component is embedded into a form and that form is submit you will receive a parameter of ordering (which can be customised with the `input_name` option). This is submitted as a collection which Rails will interpret as hash of item id => ordering position.  For example, for two items with id "a", "b" that are ordered accordingly, you'd receive a submission of `ordering[a]=1&ordering[b]=2`.

kevindew (commented):
This looks good. I've not looked too closely on the code at this point, however I did spot a couple of initial things that we'd altered in collections-publisher already so I wanted to check whether you've included all the changes we made there?

spec/javascripts/components/reorderable-list-spec.js
DilwoarH: done

app/views/govuk_publishing_components/components/_reorderable_list.html.erb
DilwoarH: done

app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js
DilwoarH: done

CHANGELOG.md
DilwoarH: done

spec/javascripts/components/reorderable-list-spec.js
DilwoarH: I reverted this as it fails tests:

```
$ standardx 'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'
standardx: Use JavaScript Standard Style (tweaked by standardx) (https://github.com/standard/standardx)
  /var/lib/jenkins/workspace/onents_add-reorderable-component/spec/javascripts/components/reorderable-list-spec.js:56:29: 'GOVUK' is not defined.
  /var/lib/jenkins/workspace/onents_add-reorderable-component/spec/javascripts/components/reorderable-list-spec.js:89:29: 'GOVUK' is not defined.
  /var/lib/jenkins/workspace/onents_add-reorderable-component/spec/javascripts/components/reorderable-list-spec.js:120:29: 'GOVUK' is not defined.
  /var/lib/jenkins/workspace/onents_add-reorderable-component/spec/javascripts/components/reorderable-list-spec.js:151:29: 'GOVUK' is not defined.
  /var/lib/jenkins/workspace/onents_add-reorderable-component/spec/javascripts/components/reorderable-list-spec.js:181:29: 'GOVUK' is not defined.
error Command failed with exit code 1.
```

It's cleaner to have it there than to do `window.GOVUK`

app/views/govuk_publishing_components/components/docs/reorderable_list.yml
DilwoarH: Done

DilwoarH:
@kevindew - ready to review

app/views/govuk_publishing_components/components/docs/reorderable_list.yml
```diff
@@ -0,0 +1,92 @@
...
+  of `ordering[a]=1&ordering[b]=2` and rails will translate that into params as:
+  
+  ```
+    {
+      "ordering" => {
+        "a" => "1",
+        "b" => "2"
+      }
+    }
+  ```
```
kevindew, line 26:
```suggestion
  When this component is embedded into a form and that form is submit you will receive a
  parameter of `ordering` (which can be customised with the `input_name` option).
  This will contain item ids and ordering positions in a hash.

  For example, for two items with id "a" and "b" that are ordered accordingly,
  you'd receive a submission of `ordering[a]=1&ordering[b]=2`, which Rails can
  translate to `"ordering" => { "a" => "1", "b" => "2" }`.
```

A few minor modifications to this.

I noticed that using  \`\`\` doesn't work properly in these docs too (pops out inline) Since the code is simple we should probably just put on a single line rather than using an alternative syntax to get a pre (for example `~~~` works).

kevindew (commented):
Ok looks great - just a few small suggestions and then this is good for ✅

app/assets/stylesheets/govuk_publishing_components/components/_reorderable-list.scss
```diff
@@ -0,0 +1,120 @@
...
+
+  @include govuk-media-query($from: desktop) {
+    display: inline-flex;
+    width: 100%;
+  }
+}
+
+.gem-c-reorderable-list__content,
+.gem-c-reorderable-list__actions {
+  margin-bottom: govuk-spacing(2);
```
alex-ju, line 46: I suspect we don't need this additional spacing anymore. I was previously being used for vertical spacing as there was no padding on the item itself. With the current design the padding in `gem-c-reorderable-list__item` ensures equal spacing. We might, however, need space between the content and the action on mobile, so I'll make a suggestion there.

app/assets/stylesheets/govuk_publishing_components/components/_reorderable-list.scss
```diff
@@ -0,0 +1,120 @@
...
+}
+
+.gem-c-reorderable-list__content {
+  @include govuk-media-query($from: desktop) {
```
alex-ju, line 50: This suggestion is for content to have some space on small screens, when actions are placed under content.

```suggestion
  margin-bottom: govuk-spacing(2);
  @include govuk-media-query($from: desktop) {
    margin-bottom: 0;
```

alex-ju (commented):
Looks good overall. I left a style suggestion, as I think with the new design we can keep the item padding equal. Happy to approve this after @kevindew's last few suggestions are addressed.

DilwoarH:
@kevindew @alex-ju all comments addressed

alex-ju (approved):
🔢

PR #1907, opened 2021-02-05
Add option to remove border from document list items

owenatgov:
## What
Adds an optional attribute to the document list component that removes the border visual element from document list items.

## Why
The document list component is currently being used in collections publisher and content publisher and the recent visual change was missed in these apps and [caused some visual disparity](https://github.com/alphagov/govuk_publishing_components/pull/1883). This removes the border element which is the source of said visual discrepancy. This decision is weighted by the fact that there is a plan to eventually [extract the docs list item as a standalone component](https://github.com/alphagov/govuk_publishing_components/compare/extract-document-item-component) meaning there is an appetite for this to not use the border in certain interfaces, such as the publishing apps.

### Before
Screenshot 2021-02-05 at 12 09 22

### After
Screenshot 2021-02-05 at 12 17 41

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Fix bugs on the feedback component ([PR #1900](https://github.com/alphagov/govuk_publishing_components/pull/1900))
 * Add data attributes to layout footer component ([PR #1904](https://github.com/alphagov/govuk_publishing_components/pull/1904))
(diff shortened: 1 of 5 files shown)


DilwoarH:
@alex-ju - how does it look with multiple steps?

DilwoarH (approved):
Amazing work Alex 🥇

DilwoarH:
There's a typo in you commit message and I think you need to rebase too

PR #1910, opened 2021-02-08
Rescope Brexit CTA on contextual sidebar

alex-ju:
## What
Instead of hiding the Brexit CTA on all answer pages, we were meant to only hide it on answer pages that have a start button, which unfortunately requires hardcoding their document IDs.

## Why
A misinterpretation of product requirements in #1888 led to a considerable decrease in engagement rate on answer pages (without a start button) which were driving significant traffic to `/transition`.

## Visual Changes
Example: https://www.gov.uk/call-charges
Details: answer page without start button, tagged to corporate-information

<table>
<tr><td valign="top">

www gov uk_call-charges

</td><td valign="top">

localhost_3090_call-charges

</td></tr>
</table>

[Trello card](https://trello.com/c/XmaPJL11)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Rescope Brexit CTA on contextual sidebar ([PR #1910](https://github.com/alphagov/govuk_publishing_components/pull/1910))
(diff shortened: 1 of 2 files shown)


sihugh (approved):
Might be worth us looking into exposing a "has_start_button" property on some formats.

hannako (approved):

PR #1912, opened 2021-02-09
Update CHANGELOG.md

chris-gds:
## What
Updating `CHANGELOG.md`


maxgds (approved):
Try to keep this coupled with the actual change in future, sorry for overlooking it when I reviewed.

chris-gds:
> Try to keep this coupled with the actual change in future, sorry for overlooking it when I reviewed.

Ah OK will do, I thought the convention might be to split this out (just looking at the history of this these types of updates)

maxgds:
> Ah OK will do, I wrongly thought the convention might be to split this out

We aim for ideally same commit, definitely same PR.

PR #1913, opened 2021-02-09
Update data-module on layout header and footer

danacotoran:
## What
Change value of `data-module` attribute to "gem-track-click" on the layout header and layout footer components.

## Why
The reason behind this change is that click tracking is not working on these elements on GOVUK Accounts. 
That is because Accounts does not use `static` and therefore also does not use the [track-click script](https://github.com/alphagov/static/blob/master/app/assets/javascripts/modules/track-click.js) which  enables click tracking on modules with the `data-module="track-click"` attribute.

Changing the value of this attribute to "gem-track-click" means that the header and footer will work with the [component gem's track-click script](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/lib/track-click.js) instead.

https://trello.com/c/9A6UPqi8

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ## Unreleased
 
+* Update `data-module` on layout header and footer ([PR #1913](https://github.com/alphagov/govuk_publishing_components/pull/1913))
 * Update design of accordion component ([PR #1884](https://github.com/alphagov/govuk_publishing_components/pull/1884))
-  
+
 ## 24.0.0
 
 * Bump govuk-frontend from 3.10.2 to 3.11.0 ([PR #1911](https://github.com/alphagov/govuk_publishing_components/pull/1911))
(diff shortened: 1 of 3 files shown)


alex-ju (changes requested):
We already have this in `lib` and planning to remove it from static very soon https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/lib/track-click.js

danacotoran:
Yeah you're right @alex-ju. 
Andy suggested that we should copy the old `track-click` module from static as he thought that the gem's version would not work as it's a re-written version of the old one. 
However, I tested this locally by changing the `data-module` attribute to "gem-track-click", and that does seem to do the trick.

alex-ju:
Thanks for looking into this, @danacotoran! That's great news! This is how we actually ended up with the current version in `lib` ([by copying it from static](https://github.com/alphagov/govuk_publishing_components/pull/1751)), so I'd really like to  avoid doing this again.

alex-ju (approved):
Don't forget to add a changelog entry.

We'll have to do this [for every component](https://github.com/alphagov/govuk_publishing_components/search?p=3&q=track-click&type=code), unfortunately, before being able to remove the copy in `static`.

PR #1915, opened 2021-02-09
Allow emergency banner and global bar in public layout component

injms:
## What
Updates and tweaks to public layout to prepare it for use.

Added slots for the emergency banner and the global bar - these slots are just spaces that raw HTML will be added to the page. Whilst both the banner and the bar are [componentised in Static][1], moving those components to the components gem is outside of the scope of this pull request.

The blue border from the header component has been switched off, as the emergency banner needs to go between the black header and the blue bar - the header component didn't offer the ability to do this, so the blue bar is now an element in the public layout component. This was an easier path than editing the widely used header component.

The global bar has the blue header border built into it, which can clash with the fixed width blue border. A CSS tweak / hack has been added to stop a double blue border from appearing.

Minor tweaks to:
 - update header toggle button to reflect changes made in the Design System.
 - remove some duplicate margin rules
 - wrap some padding in the `govuk_layout` compatibility mode flag

## Why
Some extra functionality is needed; and there are some changes that need to be made to make it work nicely with other apps because of Slimmer. 

## Visual Changes

Public layout component before:
image

Public layout component after:

image

Example with national emergency banner and global bar:

image

image

Example with JavaScript turned off:

Global bar doesn't appear when JavaScript is not turned on.

Screenshot 2021-03-18 at 14 17 52

Double blue bar before:

image

Double blue bar after:
image

[1]: https://github.com/alphagov/static/pull/2433

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -78,12 +78,14 @@
   }
 }
 
-.gem-c-layout-header__search.govuk-grid-column-one-third-from-desktop,
-.gem-c-layout-header__logo.govuk-grid-column-one-third-from-desktop,
-.gem-c-layout-header__search.govuk-grid-column-one-third,
-.gem-c-layout-header__logo.govuk-grid-column-two-thirds {
-  padding-right: 0;
-  padding-left: 0;
+@include govuk-compatibility(govuk_template) {
+  .gem-c-layout-header__search.govuk-grid-column-one-third-from-desktop,
+  .gem-c-layout-header__logo.govuk-grid-column-one-third-from-desktop,
+  .gem-c-layout-header__search.govuk-grid-column-one-third,
+  .gem-c-layout-header__logo.govuk-grid-column-two-thirds {
+    padding-right: 0;
+    padding-left: 0;
+  }
 }
 
 .gem-c-layout-header__logo,
@@ -104,7 +106,6 @@
 
 .gem-c-header__content.govuk-header__content {
   width: auto;
-  padding-left: 0;
 
   @include govuk-media-query($from: desktop) {
     float: right;
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -4,6 +4,7 @@
   global_bar ||= nil
   html_lang ||= "en"
   layout_helper = GovukPublishingComponents::Presenters::PublicLayoutHelper.new(local_assigns)
+  navigation_items ||= []
   omit_header ||= false
   show_search = local_assigns.include?(:show_search) ? local_assigns[:show_search] : true
   title ||= "GOV.UK - The best place to find government services and information"
@@ -70,6 +71,7 @@
     <% unless omit_header %>
       <%= render "govuk_publishing_components/components/layout_header", {
         search: show_search,
+        navigation_items: navigation_items,
 
         # The (blue) bottom border needs to be underneath the emergency banner -
         # so it has been turned off and added in manually.
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -1,10 +1,12 @@
 <%
-  title ||= "GOV.UK - The best place to find government services and information"
-  html_lang ||= "en"
+  emergency_banner ||= nil
   full_width ||= false
+  global_bar ||= nil
+  html_lang ||= "en"
+  layout_helper = GovukPublishingComponents::Presenters::PublicLayoutHelper.new(local_assigns)
   omit_header ||= false
   show_search = local_assigns.include?(:show_search) ? local_assigns[:show_search] : true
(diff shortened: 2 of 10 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -48,13 +50,24 @@
...
+
+    <%= raw(emergency_banner) %>
+
+    <% unless full_width %>
+      <div class="gem-c-layout-for-public__blue-bar govuk-width-container"></div>
+    <% end %>
+
+    <%= raw(global_bar) %>
+
+    <div id="wrapper" class="<%= "govuk-width-container" unless full_width %>">
```
alex-ju, line 96: It's unfortunate that we'll have to carry this id over, but it is used all over our tests, including e2e.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -7,6 +7,26 @@
...
+#
+# The global banner is shown with JavaScript, so users without JavaScript won't
+# see it. So the constrained width blue bar can't be turned off as then it'll be
+# off for everyone.
+#
+# This booleon adds a CSS class that shifts the banners up by the blue bar's
+# height, making the two blue bars overlap and appear as one. The class is added
+# when a) there's content for the emergency or global banner *and* b) when using
+# the contrained width layout.
+  blue_bar_dedupe = !full_width && global_bar.present?
```
alex-ju, line 30: As discussed, I think we can afford this bit of depth to be able to move on with the main objective. The global bar design is likely to change soon and I'd rather try to address it when iterating on it.

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -28,18 +27,22 @@
         <div class="govuk-grid-column-full govuk-grid-column-one-third-from-desktop gem-c-layout-header__search">
           <%= render "govuk_publishing_components/components/layout_header/search" %>
         </div>
-        <div class="govuk-header__content gem-c-header__content govuk-grid-column-full">
-          <%= render "govuk_publishing_components/components/layout_header/navigation_items", navigation_items: navigation_items, navigation_aria_label: navigation_aria_label  %>
-        </div>
+        <% if navigation_items.any? %>
```
alex-ju, line 30: Not necessary within the scope of this PR, but do we need any temporary selector for slimmer to be able to render the login for accounts?

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -28,18 +27,22 @@
...
+          <div class="govuk-header__content gem-c-header__content govuk-grid-column-full">
+            <%= render "govuk_publishing_components/components/layout_header/navigation_items", navigation_items: navigation_items, navigation_aria_label: navigation_aria_label  %>
+          </div>
+        <% end %>
       </div>
     <% else %>
-      <div class="govuk-grid-row govuk-!-margin-left-0 govuk-!-margin-right-0">
```
alex-ju, line 36: 🔥

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -78,12 +78,14 @@
   }
 }
 
-.gem-c-layout-header__search.govuk-grid-column-one-third-from-desktop,
-.gem-c-layout-header__logo.govuk-grid-column-one-third-from-desktop,
-.gem-c-layout-header__search.govuk-grid-column-one-third,
-.gem-c-layout-header__logo.govuk-grid-column-two-thirds {
-  padding-right: 0;
-  padding-left: 0;
+@include govuk-compatibility(govuk_template) {
```
alex-ju, line 81: 🙌  can't wait to see them go!

alex-ju (approved):
Excellent move! Are there any further changes required to the layout before plugging it into static? Also let me know if you'd like me to check this against a certain branch in static.

Requires a changelog entry; Subjective suggestion; is it more accurate if we phrase it closer to 'allow emergency and global bar..' as they won't be required

app/views/govuk_publishing_components/components/_layout_header.html.erb
injms: Not that I can see in the [accounts_shower.rb](https://github.com/alphagov/slimmer/blob/971af9aff9d26b8e549fc739f05e8afdeca0a9c3/lib/slimmer/processors/accounts_shower.rb) - both the logged in and logged out links need to be [added using the `navigation_items` parameter](https://github.com/alphagov/static/blob/master/app/views/root/_base.html.erb#L35-L59), and then Slimmer chooses which to show and hide.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
injms: It took me an embarrassingly long while to realise that this was the reason that all the tests were failing!

injms:
> Excellent move! Are there any further changes required to the layout before plugging it into static? Also let me know if you'd like me to check this against a certain branch in static.

There are some changes required - and that's a work in progress in https://github.com/alphagov/static/pull/2411.

> Requires a changelog entry; subjective suggestion: is it more accurate if we phrase it closer to 'allow emergency and global bar..' as they won't be required

Ah, I always forget about the changelog. I'll use your suggestion there, ta!

PR #1918, opened 2021-02-10
Update cookies banner to align it with govuk-frontend

maxgds:
## What
Update cookies banner to inherit the Design System cookie banner component

## Why
This brings our cookie banner in line with the design system in terms of look and feel and follows their guidance for compliance. The biggest differences:

- Adds a reject button to the default implementation and plumbs it into the js
- Hides things with the hidden attribute instead of relying on styles (which fails if a user or browser blocks CSS)
- Use the new design system button group for button layout
- Made the hide button on the confirmation message look like a button

### Before
Screenshot 2021-02-12 at 14 40 42

### After
Screenshot 2021-02-12 at 14 40 09

### Before
Screenshot 2021-02-12 at 14 40 57

### After
Screenshot 2021-02-12 at 14 41 13

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -9,10 +9,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.hideCookieMessage = this.hideCookieMessage.bind(this)
     this.$module.showConfirmationMessage = this.showConfirmationMessage.bind(this)
     this.$module.setCookieConsent = this.setCookieConsent.bind(this)
+    this.$module.rejectCookieConsent = this.rejectCookieConsent.bind(this)
 
     this.$module.cookieBanner = document.querySelector('.gem-c-cookie-banner')
     this.$module.cookieBannerConfirmationMessage = this.$module.querySelector('.gem-c-cookie-banner__confirmation')
-
+    this.$module.cookieBannerConfirmationMessageText = this.$module.querySelector('.gem-c-cookie-banner__confirmation-message')
     this.setupCookieMessage()
   }
 
@@ -24,9 +25,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       }
     }
 
-    this.$acceptCookiesLink = this.$module.querySelector('button[data-accept-cookies]')
-    if (this.$acceptCookiesLink) {
-      this.$acceptCookiesLink.addEventListener('click', this.$module.setCookieConsent)
+    this.$acceptCookiesButton = this.$module.querySelector('button[data-accept-cookies]')
+    if (this.$acceptCookiesButton) {
+      this.$acceptCookiesButton.style.display = 'block'
+      this.$acceptCookiesButton.addEventListener('click', this.$module.setCookieConsent)
+    }
+
+    this.$rejectCookiesButton = this.$module.querySelector('button[data-reject-cookies]')
+    if (this.$rejectCookiesButton) {
+      this.$rejectCookiesButton.style.display = 'block'
+      this.$rejectCookiesButton.addEventListener('click', this.$module.rejectCookieConsent)
     }
 
     this.showCookieMessage()
@@ -56,6 +64,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   CookieBanner.prototype.hideCookieMessage = function (event) {
     if (this.$module) {
+      this.$module.hidden = true
       this.$module.style.display = 'none'
       window.GOVUK.cookie('cookies_preferences_set', 'true', { days: 365 })
     }
@@ -66,6 +75,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   CookieBanner.prototype.setCookieConsent = function () {
+    if (this.$acceptCookiesButton.getAttribute('data-cookie-types') === 'all') {
+      this.$module.cookieBannerConfirmationMessageText.insertAdjacentHTML('afterbegin', 'You have accepted additional cookies. ')
+    }
     window.GOVUK.approveAllCookieTypes()
     this.$module.showConfirmationMessage()
     this.$module.cookieBannerConfirmationMessage.focus()
@@ -78,11 +90,19 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
   }
 
-  CookieBanner.prototype.showConfirmationMessage = function () {
-    this.$cookieBannerMainContent = document.querySelector('.gem-c-cookie-banner__wrapper')
+  CookieBanner.prototype.rejectCookieConsent = function () {
+    this.$module.cookieBannerConfirmationMessageText.insertAdjacentHTML('afterbegin', 'You have rejected additional cookies. ')
+    this.$module.showConfirmationMessage()
+    this.$module.cookieBannerConfirmationMessage.focus()
+    window.GOVUK.cookie('cookies_preferences_set', 'true', { days: 365 })
+  }
 
-    this.$cookieBannerMainContent.style.display = 'none'
+  CookieBanner.prototype.showConfirmationMessage = function () {
+    this.$cookieBannerMainContent = document.querySelector('.js-banner-wrapper')
+
+    this.$cookieBannerMainContent.hidden = true
     this.$module.cookieBannerConfirmationMessage.style.display = 'block'
+    this.$module.cookieBannerConfirmationMessage.hidden = false
   }
 
   CookieBanner.prototype.isInCookiesPage = function () {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
@@ -1,3 +1,4 @@
+@import "govuk/components/cookie-banner/cookie-banner";
 $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
 
 .js-enabled {
@@ -7,48 +8,12 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
 }
 
 .gem-c-cookie-banner {
-  @include govuk-font($size: 16);
-  padding: govuk-spacing(2) 0;
   background-color: $govuk-cookie-banner-background;
 }
 
-.gem-c-cookie-banner--services {
+// can't be used without js so implement there
+.gem-c-cookie-banner .gem-c-button {
   display: none;
-}
-
-.gem-c-cookie-banner__message {
-  display: inline-block;
-  padding-bottom: govuk-spacing(2);
-
-  @include govuk-font($size: 16);
-  @include govuk-media-query($from: desktop) {
-    padding-right: govuk-spacing(4);
-  }
-}
-
-.gem-c-cookie-banner__button {
-  &.govuk-grid-column-one-half-from-desktop {
-    padding: 0;
-  }
-
-  .govuk-button {
-    @include govuk-media-query($from: desktop) {
-      width: 90%;
-    }
-
-    @include govuk-media-query($until: desktop) {
-      margin-bottom: govuk-spacing(4);
-    }
-  }
-}
-
-// Only show accept button if users have js and can accept
-.gem-c-cookie-banner__button-accept {
-  display: none;
-}
-
-.js-enabled .gem-c-cookie-banner__button-accept {
-  display: inline-block;
 }
 
 .gem-c-cookie-banner__confirmation {
@@ -81,51 +46,11 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
   }
 }
 
-.gem-c-cookie-banner__hide-button {
-  @include govuk-font($size: 19);
-  outline: 0;
-  border: 0;
-  background: none;
-  text-decoration: underline;
-  color: $govuk-link-colour;
-  padding: govuk-spacing(0);
-  margin-top: govuk-spacing(2);
-
-  &:hover {
-    color: $govuk-link-hover-colour;
-    cursor: pointer;
-  }
-
-  &:focus {
-    @include govuk-focused-text;
-  }
-
-  @include govuk-media-query($from: desktop) {
-    margin-top: govuk-spacing(0);
-    position: absolute;
-    right: govuk-spacing(1);
-  }
-}
-
-.gem-c-cookie-banner__buttons--flex {
-  display: flex;
-  flex-wrap: wrap;
-  align-items: baseline;
-
-  .govuk-button,
-  .gem-c-cookie-banner__link {
-    flex-grow: 1;
-    flex-basis: 10rem;
-    margin-right: govuk-spacing(3);
-    margin-bottom: govuk-spacing(3);
-  }
-}
-
 // Override the styles from govuk_template
 // stylelint-disable selector-max-id
 .gem-c-cookie-banner#global-cookie-message {
   background-color: $govuk-cookie-banner-background;
-  padding: govuk-spacing(4) 0;
+  padding: 0;
   box-sizing: border-box;
 
   .gem-c-cookie-banner__message,
@@ -135,18 +60,13 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
     @include govuk-font($size: 19);
   }
 
-  .gem-c-cookie-banner__message {
-    margin-bottom: 0;
-  }
-
   p {
     @include govuk-font($size: 19);
     margin: 0 0 govuk-spacing(2) 0;
   }
 
-  .gem-c-cookie-banner__confirmation-message {
-    @include govuk-media-query($from: desktop) {
-      margin-bottom: 0;
-    }
+  .gem-c-cookie-banner__message,
+  .gem-c-cookie-banner__confirmation {
+    margin-bottom: - govuk-spacing(2);
   }
 }
--- a/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
@@ -1,72 +1,83 @@
 <%
   id ||= 'global-cookie-message'
-  title ||= "Tell us whether you accept cookies"
-  text ||= raw("We use <a class='govuk-link' href='/help/cookies'>cookies to collect information</a> about how you use GOV.UK. We use this information to make the website work as well as possible and improve government services.")
+  title ||= "Cookies on GOV.UK"
+  text ||= ["We use some essential cookies to make this website work.", "We’d like to set additional cookies to understand how you use GOV.UK, remember your settings and improve government services.", "We also use cookies set by other sites to help us deliver content from their services."]
+  if text.kind_of?(Array)
+    newtext = ""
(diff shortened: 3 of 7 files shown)


app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -66,6 +72,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   CookieBanner.prototype.setCookieConsent = function () {
+    if (this.$acceptCookiesLink.getAttribute('data-cookie-types') === 'all') {
+      this.$module.cookieBannerConfirmationMessageText.insertAdjacentHTML('afterbegin', 'You have accepted additional cookies. ')
```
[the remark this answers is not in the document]
andysellick: (same for the 'rejected' text below)

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,80 @@
 <%
   id ||= 'global-cookie-message'
-  title ||= "Tell us whether you accept cookies"
-  text ||= raw("We use <a class='govuk-link' href='/help/cookies'>cookies to collect information</a> about how you use GOV.UK. We use this information to make the website work as well as possible and improve government services.")
+  title ||= "Cookies on GOV.UK"
+
+  text ||= raw("<p>We use some essential cookies to make this website work.</p>
+                <p>We’d like to set additional cookies so we and our partners can remember your settings, understand how you use GOV.UK and make improvements.</p>
```
andysellick: I'm wondering if there's a way to stop this being a breaking change. Could it detect if it's passed a single string (the old way) and wrap that in a para and render it, but otherwise expect an array of strings, which it would then each wrap in a para and render?

This is a bit messy, but it seems to work.

```
  text ||= ["We use some essential cookies to make this website work", "We'd like to set additional cookies to understand how you use GOV.UK, remember your settings and improve government services.", "We also use cookies set by other sites to help us deliver content from their services."]

  if text.kind_of?(Array)
    newtext = ""
    text.each do |t|
      newtext += "<p>#{t}</p>"
    end

    text = newtext
  else
    text = "<p>#{text}</p>"
  end

  text = raw(text)
```

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,80 @@
...
 %>
-
 <div id="<%= id %>" class="<%= css_classes.join(' ') %>" data-module="cookie-banner" role="region" aria-label="cookie banner" data-nosnippet>
-  <div class="gem-c-cookie-banner__wrapper govuk-width-container">
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <div class="gem-c-cookie-banner__message">
-          <h2 class="govuk-heading-m"><%= title %></h2>
-          <p class="govuk-body"><%= text %></p>
+  <div class="govuk-cookie-banner gem-c-cookie-banner__wrapper" role="region" aria-label="<%= title %>">
```
andysellick, line 12: I can't see `gem-c-cookie-banner__wrapper` in the code anywhere, is this needed?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,80 @@
...
+    <div class="gem-c-cookie-banner__message govuk-cookie-banner__message govuk-width-container govuk-body">
+
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
+          <h2 class="gem-c-cookie-banner__heading govuk-cookie-banner__heading govuk-heading-m"><%= title %></h2>
```
andysellick, line 17: I can't see `gem-c-cookie-banner__heading` in the code, is this needed?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,80 @@
...
+
+          <div class="gem-c-cookie-banner__content govuk-cookie-banner__content">
```
andysellick, line 19: I can't see `gem-c-cookie-banner__content` in the code, is this needed?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
andysellick: This would also allow us to insert the `govuk-body` class on the `p` elements, and remove lines 71-74 of the stylesheet, where we go against our coding standards and style a `p` element without a class.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,80 @@
...
         <% end %>
-      </div>
     </div>
   </div>
 
   <div class="gem-c-cookie-banner__confirmation govuk-width-container" tabindex="-1">
     <p class="gem-c-cookie-banner__confirmation-message" role="alert"><%= confirmation_message %></p>
     <button class="gem-c-cookie-banner__hide-button" data-hide-cookie-banner="true" data-module="track-click" data-track-category="cookieBanner" data-track-action="Hide cookie banner">Hide</button>
   </div>
+
```
andysellick, line 78: Nitpick: do we need an empty line here?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -8,18 +8,16 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
...
 .gem-c-cookie-banner {
-  @include govuk-font($size: 16);
-  padding: govuk-spacing(2) 0;
   background-color: $govuk-cookie-banner-background;
 }
 
-// Only show accept button if users have js and can accept
-.gem-c-cookie-banner__button-accept {
+// Only show buttons if users have js and can accept
+.gem-c-button.js-visible {
```
andysellick, line 15: Can you explain the logic here a bit? I'm confused why we have JS visible rules for individual elements of the cookie banner, when the cookie banner itself isn't shown when JS is disabled.

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -8,18 +8,16 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
...
   display: none;
 }
 
-.js-enabled .gem-c-cookie-banner__button-accept {
-  display: inline-block;
+.js-enabled .gem-c-button.js-visible {
```
andysellick, line 19: This seems to be applying styling using a `js-` prefixed class, which is specifically for JS hooks only and not to be used for styling. Is there an alternative way this could be implemented? Could it be shown/hidden in the JS as required?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
maxgds: The banner is shown when JS is disabled, but the buttons aren't. I don't know why that decision was made but it's the way it's worked for a long time. It links to the cookies page which tells you you need js to set your preferences. 

So the flow is: show the banner by default. If js is on, hide it with styling based on the js-enabled class, while the more specific js re-displays it if necessary based on appropriate cookie reading logic.

andysellick: That's really interesting, thanks. I think it would be worth documenting this behaviour somewhere, perhaps in the component guide page?

Testing this in the preview app the whole component disappears if JS is disabled. Is that expected?

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
maxgds: > This seems to be applying styling using a `js-` prefixed class, which is specifically for JS hooks only and not to be used for styling. Is there an alternative way this could be implemented? Could it be shown/hidden in the JS as required?

I felt this was a reasonable way to do it as it is using js to apply the js-enabled class that this relies on. I can do this in the js code but that seemed like more overhead. I'll look into it.

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
maxgds: No... I've got it running on local static/government-frontend and that doesn't happen. :(

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -29,6 +30,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.$acceptCookiesLink.addEventListener('click', this.$module.setCookieConsent)
     }
 
+    this.$rejectCookiesLink = this.$module.querySelector('button[data-reject-cookies]')
```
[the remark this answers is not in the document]
maxgds: Fixing for accept and reject.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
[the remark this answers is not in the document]
maxgds: Will revisit later, but for the purpose of getting this in front of users asap agreed to skip for now.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: Done. PR and changelog updated accordingly.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: This one was needed as it's the hook for hiding the main banner when we show the accepted/rejected confirmation.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: removed

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: removed

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
maxgds: Whatever's going wrong is also happening on the live version, needs further investigation, but won't block this.

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
maxgds: Fixed.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: Sadly the overrides are still needed because govuk-template has a selector with an ID that overrides this and shrinks all the text.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -9,7 +9,7 @@
     end
     text = newtext
   else
-    text = "<p class='govuk-body'>#{text}</p>"
+    text = "<p class='govuk-body govuk-body-l'>#{text}</p>"
```
andysellick, line 12: Why is `govuk-body-l` applied here and not on line 8?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
andysellick: Could it be changed to a `js-` class instead?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -20,12 +20,12 @@
...
 %>
 <div id="<%= id %>" class="<%= css_classes.join(' ') %>" data-module="cookie-banner" role="region" aria-label="cookie banner" data-nosnippet>
-  <div class="govuk-cookie-banner" role="region" aria-label="<%= title %>">
+  <div class="govuk-cookie-banner gem-c-cookie-banner__wrapper" role="region" aria-label="<%= title %>">
     <div class="gem-c-cookie-banner__message govuk-cookie-banner__message govuk-width-container govuk-body">
       <div class="govuk-grid-row">
         <div class="govuk-grid-column-two-thirds">
           <h2 class="govuk-cookie-banner__heading govuk-heading-m"><%= title %></h2>
-          <div class="govuk-cookie-banner__content">
+          <div class="govuk-cookie-banner__content" >
```
andysellick, line 28: Nitpick: seems to be an extra space in this line.

app/views/govuk_publishing_components/components/docs/cookie_banner.yml
```diff
@@ -16,7 +18,7 @@ examples:
   with_custom_content:
     data:
       title: "Can we store analytics cookies on your device?"
-      text: "<p>This is some custom text in my cookie banner which lets users know what we're using cookies for. I can also include a link to the <a class='govuk-link' href='/cookies'>cookies page</a></p>"
+      text: "This is some custom text in my cookie banner which lets users know what we're using cookies for. I can also include a link to the <a class='govuk-link' href='/cookies'>cookies page</a>"
```
andysellick: Probably worth adding another example like this one to show passing multiple strings for the text as an array, and including a test for that (if there isn't one already).

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
maxgds: fixed

app/views/govuk_publishing_components/components/docs/cookie_banner.yml
maxgds: added

app/views/govuk_publishing_components/components/docs/cookie_banner.yml
```diff
@@ -18,14 +20,21 @@ examples:
...
+
   with_custom_cookie_preferences_href:
     data:
       cookie_preferences_href: "/cookies"
   in_services_asking_for_analytics_only:
     description: Use this type of banner in your service if it only uses cookies for analytics.
     data:
       title: Can we store analytics cookies on your device?
-      text: Analytics cookies help us understand how our website is being used.
+      text: "Analytics cookies help us understand how our website is being used."
```
andysellick, line 37: Why did quote marks get added to this line? Are they needed?

spec/components/cookie_banner_spec.rb
```diff
@@ -51,15 +57,24 @@ def component_name
...
   end
 
+  it "renders with multi paragraph content" do
+    render_component(
+      title: "Can we store analytics cookies on your device?",
+      text: ["Paragraph 1", "Paragraph 2"],
+      confirmation_message: "You’ve accepted all cookies.",
+    )
+    assert_select "p.govuk-body", text: "Paragraph 1"
+    assert_select "p.govuk-body", text: "Paragraph 1"
```
andysellick, line 72: Looks like a copy/paste oversight here - this line should say `text: Paragraph 2`

andysellick (commented):
Nearly there, just a couple of tiny things.

alex-ju (commented):
Thank you for taking this on, Max! Looks really good and it's great to see a lot of SCSS removed. I finished reviewing the visual bits and left a few minor suggestions which will hopefully make it identical to the DS version. Will finish the functional checks and get back quickly.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,83 @@
 <%
   id ||= 'global-cookie-message'
-  title ||= "Tell us whether you accept cookies"
-  text ||= raw("We use <a class='govuk-link' href='/help/cookies'>cookies to collect information</a> about how you use GOV.UK. We use this information to make the website work as well as possible and improve government services.")
+  title ||= "Cookies on GOV.UK"
+  text ||= ["We use some essential cookies to make this website work", "We'd like to set additional cookies to understand how you use GOV.UK, remember your settings and improve government services.", "We also use cookies set by other sites to help us deliver content from their services."]
```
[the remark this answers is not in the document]
maxgds: Had already spotted the . Have now fixed the apostrophe too.

spec/components/cookie_banner_spec.rb
maxgds: Yup. Fixed.

app/views/govuk_publishing_components/components/docs/cookie_banner.yml
maxgds: Ah, added them to accommodate the <p> tags earlier, I think. Fixed.

app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -135,18 +60,12 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
...
     @include govuk-font($size: 19);
     margin: 0 0 govuk-spacing(2) 0;
   }
 
-  .gem-c-cookie-banner__confirmation-message {
-    @include govuk-media-query($from: desktop) {
-      margin-bottom: 0;
-    }
+  .gem-c-cookie-banner__message {
+    margin-bottom: 0;
```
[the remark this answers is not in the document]
maxgds: 👍

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,72 +1,83 @@
...
-
 <div id="<%= id %>" class="<%= css_classes.join(' ') %>" data-module="cookie-banner" role="region" aria-label="cookie banner" data-nosnippet>
-  <div class="gem-c-cookie-banner__wrapper govuk-width-container">
-    <div class="govuk-grid-row">
-      <div class="govuk-grid-column-two-thirds">
-        <div class="gem-c-cookie-banner__message">
-          <h2 class="govuk-heading-m"><%= title %></h2>
-          <p class="govuk-body"><%= text %></p>
+  <div class="govuk-cookie-banner js-banner-wrapper" role="region" aria-label="<%= title %>">
+    <div class="gem-c-cookie-banner__message govuk-cookie-banner__message govuk-width-container govuk-body">
```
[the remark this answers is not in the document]
maxgds: No, that was overlooked from when I was trying to avoid the id-related overrides. Will remove.

andysellick (commented):
Small thing I just noticed in the component guide - in the last example, the only button visible is 'yes' - shouldn't 'no' also be visible?

Screenshot 2021-02-16 at 08 40 18

maxgds:
> Small thing I just noticed in the component guide - in the last example, the only button visible is 'yes' - shouldn't 'no' also be visible?

Fixed - the it had the wrong data tag so the js wasn't displaying it. Test updated too.

alex-ju (approved):
Unless Andy has any other questions or spots anything else, I would be happy to get this in. One last suggestion, maybe a bit more detail on the changelog (e.g. Update cookies banner to align it with `govuk-frontend`) could help when looking back.

andysellick (approved):
Looks good 👍  well done.

PR #1922, opened 2021-02-11
Update `govuk-frontend` base SCSS imports

alex-ju:
## What
Clean up `govuk_frontend_support.scss`

## Why
It's [recommended to import base instead of settings, tools, helpers](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#import-settings-tools-and-helpers-css-in-one-line) since `govuk-frontend` 3.7.0

The [FCDO colour is already available](https://github.com/alphagov/govuk-frontend/blob/master/CHANGELOG.md#fixes-5) since `govuk-frontend` 3.9.0 

### Note
In `govuk-frontend` 4.0.0 we [won't be able to import `core` and `overrides` without `base`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/component_support.scss#L4-L7), which means that our [`component_support.scss`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/component_support.scss) must import [`govuk_frontend_support.scss`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss) by default [instead of doing it separately/optionally at the app level](https://github.com/alphagov/collections/blob/master/app/assets/stylesheets/application.scss#L7-L8).

## Visual Changes
No visual changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Update `govuk-frontend` base SCSS imports ([PR #1922](https://github.com/alphagov/govuk_publishing_components/pull/1922))
+
 ## 24.1.0
 
 * Update `data-module` on layout header and footer ([PR #1913](https://github.com/alphagov/govuk_publishing_components/pull/1913))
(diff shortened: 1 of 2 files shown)


andysellick:
Nice change. I'm hoping that `base` adds no CSS weight?

alex-ju:
> Nice change. I'm hoping that `base` adds no CSS weight?

[`base`](https://github.com/alphagov/govuk-frontend/blob/master/src/govuk/_base.scss) is virtually the same with `settings`, `tools`, `helpers` which is why there was no rush with this change, but 4.0.0 is getting close.

andysellick (approved):
👍

PR #1923, opened 2021-02-11
Remove redundant import in accordion component

alex-ju:
## What
Remove redundant import in the accordion component that was accidentally introduced in #1884. 

We do however need it for the print stylesheet, as we have two [applications requiring the `components/print/accordion` but not the `component_support`](https://github.com/alphagov/collections/blob/master/app/assets/stylesheets/print.scss#L2). This could potentially be fixed in future by removing `component_support` from `print/_accordion` and updating the suggested print imports to include `component_support`.

## Why
This import resulted in a considerable increase in our render-blocking resources.

## Visual Changes
No visual changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Update `govuk-frontend` base SCSS imports ([PR #1922](https://github.com/alphagov/govuk_publishing_components/pull/1922))
+* Remove redundant import in accordion component ([PR #1923](https://github.com/alphagov/govuk_publishing_components/pull/1923))
 
 ## 24.1.0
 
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Looks sensible. Can you record here specifically what it is that this component needs from `component_support`?

alex-ju:
> Looks sensible. Can you record here specifically what it is that this component needs from `component_support`?

All the `govuk-frontend` mixins used in [accordion](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss) 🙂

chris-gds:
Ahh thanks for this @alex-ju! Do we _want_ to use the `govuk-frontend` mixins in this way? 🤔

alex-ju:
> Ahh thanks for this @alex-ju! Do we _want_ to use the `govuk-frontend` mixins in this way? 🤔

Yes, we want to use `govuk-frontend` mixins as much as possible as it helps with building components in a more consistent way. The only different thing with this component is that the print stylesheet requires the component stylesheet (as the Design System doesn't ship print stylesheets at the moment).

PR #1893, opened 2021-01-27
Further step by step nav sidebar design updates

owenatgov:
## What
Makes several design updates to the step by step:

- Alter vertical padding for step titles
- Adjust circle size, logic font size and number/logic positioning
- Reduce border width across component
- Add a black background to the "active" step circle
- Move the chevron icon to the right
- Add a customisation of the focus state for step titles

## Why
The root of this work comes from efforts by the govuk accessibility team to ensure that accordions across govuk are consistent and accessible.

This PR is an iteration on [the recently deployed update to the step by step component](https://github.com/alphagov/govuk_publishing_components/pull/1875). These design changes are both responses to feedback following launch and further improvements based on further conversations between our designer and the original designer. This is also going to feed into [the accordion redesign](https://github.com/alphagov/govuk_publishing_components/pull/1884) happening alongside this.

[Card](https://trello.com/c/ebTxqizo/611-design-tweaks-for-the-step-by-step-accordion)

### Full page - before
Screenshot 2021-01-27 at 10 57 07

### Full page - default
Screenshot 2021-01-27 at 10 53 47

### Full page - focus states
Screenshot 2021-01-28 at 09 53 28
Screenshot 2021-01-28 at 09 53 57
Screenshot 2021-01-28 at 09 54 13

### Full page - no JS
Screenshot 2021-01-27 at 10 58 33

### Contextual sidebar - before
Screenshot 2021-01-27 at 10 59 57

### Contextual sidebar - default
Screenshot 2021-01-28 at 09 55 42

### Contextual sidebar - focus states
Screenshot 2021-01-28 at 09 56 11
Screenshot 2021-01-28 at 09 56 22
Screenshot 2021-01-28 at 09 56 36
Screenshot 2021-01-28 at 09 56 46

### Contextual sidebar - no JS
Screenshot 2021-01-27 at 11 02 05

PR #1925, opened 2021-02-12
Fix toggle click tracking on step-by-steps

alex-ju:
## What
In #1893 we added an SVG to the step by step navigation continuing to rely on the same logic for tracking, but the way the toggle click tracking script is designed expects the element triggering the click event to be an HTMLElement, not a path element, which causes a hard error and fails to send data to GA. With this change, we avoid capturing events at the SVG level, which means they will bubble and be captured by the parent HTMLElement.

## Why

- step-by-step-nav script fails on pages when users click the SVG element
- we are losing tracking data since 9 Feb

## Visual Changes
No visual changes

Example page: https://www.gov.uk/apply-transit-visa

<table>
<tr><td valign="top">

before-track

</td><td valign="top">

after-track

</td></tr>
<tr><td valign="top">

before-js

</td><td valign="top">

No JavaScript errors

</td></tr>
</table>


owenatgov (approved):
Great spot, great solution, great!

PR #1927, opened 2021-02-15
Fix auditing screens

andysellick:
## What
- recent changes to the accordion had broken the accordions on the component auditing screens
- this wasn't the fault of the component, the markup for the auditing accordions was hard coded
- unsure why this was (even though I built it) but have replaced with the component (apparently successfully)

## Why
It was broken.

## Visual Changes
The only visual change is the accordion now uses the updated accordion design, but this is already known.

### DIFF
--- a/app/views/govuk_publishing_components/audit/show.html.erb
+++ b/app/views/govuk_publishing_components/audit/show.html.erb
@@ -39,98 +39,105 @@
           </div>
         </details>
 
-        <div class="gem-c-accordion" data-module="gem-c-accordion" id="accordion-with-summary-sections">
-          <% @applications.each_with_index do |application, index| %>
-            <div class="gem-c-accordion__section ">
-              <div class="gem-c-accordion__section-header">
-                <h2 class="gem-c-accordion__section-heading">
-                  <span class="gem-c-accordion__section-button" id="accordion-with-summary-sections-heading-<%= index %>">
-                    <%= application[:name] %>
-                  </span>
-                </h2>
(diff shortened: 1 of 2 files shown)


app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -39,98 +39,105 @@
...
-                  <% application[:warnings].each do |warning| %>
-                    <p class="govuk-body">
-                      <strong class="govuk-tag">Warn</strong>
-                      <strong><%= warning[:component] %></strong> - <%= warning[:message] %>
-                    </p>
-                  <% end %>
+        <% application_items = [] %>
+        <% @applications.each_with_index do |application, index| %>
+          <%
+            summary = '<strong class="govuk-tag govuk-tag--red">Application not found</strong>'
```
injms, line 45: I don't think you use `index` anywhere - if that's the case, this could be rejigged to use `.map` directly to `application_items`

eg:

```suggestion
        <% application_items = @applications.map do |application| %>
```

andysellick: I don't need the `index` but I'd need to change quite a bit more to make this work as inside this loop it adds things onto the `application_items` array, for output by the component. It's probably doable but more of an optimisation than I want to get into right now.

injms: Okaydokey - approved since using map or each doesn't make a difference what's being output.

injms (approved):

PR #1936, opened 2021-02-17
Add spacing to cookie banner confirmation message

maxgds:
## What
Add spacing to cookie banner confirmation message

## Why
It currently sits too close to the top of the browser

### Before
Screenshot 2021-02-17 at 11 29 16

### After
Screenshot 2021-02-17 at 11 28 53

Screenshot 2021-02-17 at 11 35 28

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
@@ -46,11 +46,18 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
   }
 }
 
+// Override style from design system so we can have consistent
+// padding on the the banner and the confirmation
+
+.govuk-cookie-banner {
+  padding-top: 0;
+}
+
 // Override the styles from govuk_template
 // stylelint-disable selector-max-id
 .gem-c-cookie-banner#global-cookie-message {
   background-color: $govuk-cookie-banner-background;
-  padding: 0;
+  padding: govuk-spacing(3) 0 0 0;
   box-sizing: border-box;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_cookie-banner.scss
```diff
@@ -39,6 +39,7 @@ $govuk-cookie-banner-background: govuk-colour("light-grey", "grey-4");
 
 .gem-c-cookie-banner__confirmation-message {
   margin-right: govuk-spacing(4);
+  padding-top: govuk-spacing(3);
```
andysellick, line 42: I think it would be better to have the padding on the overall component parent rather than add specific padding to child elements that does the same thing - although it looks like the problem here is the DS styles rather than yours.

I would remove the top padding from `govuk-cookie-banner` and put it on `gem-c-cookie-banner` instead, which would pad that and the confirmation message as well.

maxgds: It's messy with the required overrides, but I accept your point about applying it to the parent consistently - have updated.

andysellick: Thanks, that seems better to me. Unfortunately the spacing isn't consistent, due to a child of the confirmation message having padding of its own...

Screenshot 2021-02-17 at 17 08 12

...but to be honest at this point if these are all DS styles then I think you've already done enough to try to make this consistent and any further discussion should happen at a different level, so I'm going to approve since you've already solved the problem you set out to solve.

andysellick (approved):

PR #1937, opened 2021-02-17
Add anchor navigation feature to accordions

owenatgov:
## What
Applies an optional anchor navigation feature to accordions. This allows anchor links within accordions to:

- link to another accordion heading on the same page and automatically open that accordion
- link to an id anchor within an accordion on the same page and automatically open the containing accordion
- link to either of the above on different pages with the same setup (accordions with this feature flag) and open the associated accordion as above.

Additionally fixes a couple of typos in the accordions js and removes a line specifying data_attributes as `nil`. The latter is unnecessary and generates an extra `null` attribute on every item on the docs.

## Why
On manuals, there was previously a function to link between individual sections in manuals for ease of content. This was done by linking via an id anchor and the old bespoke manuals accordion js would work this out and open the accordion. This feature was overlooked in the recent update to manual accordions, leaving a lot of broken links and potentially confusing user journeys across manuals. There's a question to be answered about how accessible this functionality is, however for the moment this PR repairs the bug without having to undo the accordion changes and re-introducing a WCAG 2.3.4 failure.

No visual changes.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -11,8 +11,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GemAccordion.prototype.start = function ($module) {
     this.$module = $module[0]
+    this.sectionClass = 'gem-c-accordion__section'
     this.moduleId = this.$module.getAttribute('id')
-    this.sections = this.$module.querySelectorAll('.gem-c-accordion__section')
+    this.sections = this.$module.querySelectorAll('.' + this.sectionClass)
     this.openAllButton = ''
     this.browserSupportsSessionStorage = helper.checkForSessionStorage()
     this.controlsClass = 'gem-c-accordion__controls'
@@ -35,6 +36,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.initControls()
     this.initSectionHeaders()
+
+    // Feature flag for anchor tag navigation used on manuals
+    if (this.$module.getAttribute('data-anchor-navigation')) {
+      this.openByAnchorOnLoad()
+      this.addEventListenersForAnchors()
+    }
 
     // See if "Show all sections" button text should be updated
     var areAllSectionsOpen = this.checkIfAllSectionsOpen()
@@ -59,7 +66,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     accordionControls.appendChild(this.openAllButton)
     this.$module.insertBefore(accordionControls, this.$module.firstChild)
 
-    // Build addtional wrapper for open all toggle text, place icon after wrapped text.
+    // Build additional wrapper for open all toggle text, place icon after wrapped text.
     var wrapperOpenAllText = document.createElement('span')
     wrapperOpenAllText.classList.add(this.openAllTextClass)
     this.openAllButton.insertBefore(wrapperOpenAllText, this.openAllButton.childNodes[0] || null)
@@ -74,6 +81,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     nodeListForEach(this.sections, function (section, i) {
       // Set header attributes
       var header = section.querySelector('.' + this.sectionHeaderClass)
+
       this.initHeaderAttributes(header, i)
       this.setExpanded(this.isExpanded(section), section)
 
@@ -106,12 +114,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     srPause.classList.add('govuk-visually-hidden')
     srPause.innerHTML = ', '
 
-    // Build addtional copy for assistive technology
-    var srAddtionalCopy = document.createElement('span')
-    srAddtionalCopy.classList.add('govuk-visually-hidden')
-    srAddtionalCopy.innerHTML = ' this section'
+    // Build additional copy for assistive technology
+    var srAdditionalCopy = document.createElement('span')
+    srAdditionalCopy.classList.add('govuk-visually-hidden')
+    srAdditionalCopy.innerHTML = ' this section'
 
-    // Build addtional wrapper for toggle text, place icon after wrapped text.
+    // Build additional wrapper for toggle text, place icon after wrapped text.
     var wrapperShowHideIcon = document.createElement('span')
     var icon = document.createElement('span')
     icon.classList.add(this.upChevonIconClass)
@@ -138,7 +146,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     button.appendChild(showIcons)
-    button.appendChild(srAddtionalCopy)
+    button.appendChild(srAdditionalCopy)
   }
 
   // When section toggled, set and store state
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -274,6 +274,32 @@ examples:
             text: "How people read"
           content:
             html: "<p class='govuk-body'>This is the content for How people read.</p>"
+  with_id_on_headings:
+    description: |
+      Some apps require custom ids per accordion section heading for linking between those specific sections across multiple pages. An example of this is on manuals pages where multiple manuals will each include large sets of accordions and will reference between specific sections for ease of access to that content.
+
+      This attribute is paired with functionality to automatically open an accordion section on page load where the URL hash is the same as the id of that accordion. [Live example](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#associations)
+
+      Unlike with the accordion-wide custom id attribute, this id isn't stored in `localStorage` so does not need to be unique across your domain, but should still be unique to the page.
```
chris-gds, line 283: This is great to mention this **unique** aspect. I'm wondering if this should be thought about when it comes to the API work.  

I'm thinking about the easy of duplicate IDs happening..

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -106,12 +111,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-    var srAddtionalCopy = document.createElement('span')
-    srAddtionalCopy.classList.add('govuk-visually-hidden')
-    srAddtionalCopy.innerHTML = ' this section'
+    // Build additional copy for assistive technology
+    var srAdditionalCopy = document.createElement('span')
+    srAdditionalCopy.classList.add('govuk-visually-hidden')
+    srAdditionalCopy.innerHTML = ' this section'
 
-    // Build addtional wrapper for toggle text, place icon after wrapped text.
+    // Build additional wrapper for toggle text, place icon after wrapped text.
```
chris-gds, line 119: Ahh good spot!

app/views/govuk_publishing_components/components/docs/accordion.yml
andysellick: Rails has a thing for generating random values that are unique called `SecureRandom`. I don't think it'd be useful here (looks like if no id is passed no id is created) but thought it worth mentioning.

spec/components/accordion_spec.rb
```diff
@@ -224,6 +224,26 @@ def component_name
...
+  it "adds id to heading when attribute passed" do
+    test_data = {
+      id: "condensed-layout",
+      condensed: true,
+      items: [{
+        heading: { text: "Heading 1", id: "heading-with-id" },
+        summary: { text: "Summary 1." },
+        content: { html: "<p>Content 1.</p>" },
+      },
+              {
```
andysellick, line 236: Has something gone wrong with the indentation here?

spec/components/accordion_spec.rb
```diff
@@ -224,6 +224,26 @@ def component_name
...
+                heading: { text: "Heading 2" },
+                summary: { text: "Summary 2." },
+                content: { html: "<p>Content 2.</p>" },
+              }],
+    }
+    render_component(test_data)
+    assert_select ".gem-c-accordion__section-heading", count: 2
+    assert_select ".gem-c-accordion__section-heading#heading-with-id", count: 1
```
andysellick, line 244: Is it worth testing that the other headings don't have IDs?

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -274,6 +274,32 @@ examples:
...
+    data:
+      items:
+        - heading:
+            text: "Writing well for the web"
```
andysellick, line 287: For reference I don't believe you need to put quotes around simple strings like this in yml.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -106,12 +111,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
     srPause.innerHTML = ', '
 
-    // Build addtional copy for assistive technology
-    var srAddtionalCopy = document.createElement('span')
-    srAddtionalCopy.classList.add('govuk-visually-hidden')
-    srAddtionalCopy.innerHTML = ' this section'
+    // Build additional copy for assistive technology
+    var srAdditionalCopy = document.createElement('span')
+    srAdditionalCopy.classList.add('govuk-visually-hidden')
+    srAdditionalCopy.innerHTML = ' this section'
```
andysellick, line 117: Ideally this text would come from the `en.yml` file so it could be translated. The way to do that is to include it in the template via a data attribute on an element, then read it in using the JS.

andysellick (commented):
Drive by review, couple of comments/questions. Also I'd suggest squashing your two commits together as they're part of the same change.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: This _is_ worth thinking about / doing.  

...I'm now wondering if this type of translation functionally exists on the Design System and if not what that might mean for the end goal to remove this from them gem 🤔

app/views/govuk_publishing_components/components/docs/accordion.yml
chris-gds: Ah thanks @andysellick, good to know, To give _some_ context in a publishing app an ID might be passed to link elsewhere - eg

`## Add associations (tagging) 
{:#associations}`

app/views/govuk_publishing_components/components/docs/accordion.yml
owenatgov: If you're ok with it I'll leave this how it is for now as this seems to be the pattern for the rest of the doc, however I'll keep this in mind in future. I can issue it if you want?

app/assets/javascripts/govuk_publishing_components/components/accordion.js
owenatgov: Do you mind if I issue this? This PR is to fix a fairly critical bug and I'd like to get it out sooner rather than later.

app/views/govuk_publishing_components/components/docs/accordion.yml
owenatgov: Remember that there's a lot we still don't know. Manuals publisher might have validation for duplicate ids. This wouldn't surprise me to be honest, it's a feature in Whitehall of all places! I don't think policing data here is a valuable use of time.

app/views/govuk_publishing_components/components/docs/accordion.yml
andysellick: Yes, that would be great, thanks.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
andysellick: Yes, that would be great, thanks.

owenatgov:
@andysellick Done https://github.com/alphagov/govuk_publishing_components/issues/1941

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -283,5 +290,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+  GemAccordion.prototype.openForAnchor = function (hash) {
+    var target = document.querySelector(hash)
+    var section = this.getContainingSection(target)
+
+    this.setExpanded(true, section)
+  }
+
+  // Loop through the given ids ancestors until the parent section class is found
+  GemAccordion.prototype.getContainingSection = function (target) {
+    while (!target.classList.contains('gem-c-accordion__section')) {
```
chris-gds, line 322: Non-blocker, I think this could be swapped to ref a var.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -5,22 +5,22 @@
   id ||= "default-id-#{SecureRandom.hex(4)}"
   items ||= []
   condensed ||= false
+  anchor_navigation ||= false
```
chris-gds, line 8: The flag is a nice halfway.

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -274,6 +274,35 @@ examples:
             text: "How people read"
           content:
             html: "<p class='govuk-body'>This is the content for How people read.</p>"
+  with_the_anchor_link_navigation:
+    description: |
+      Some apps require custom ids per accordion section heading for linking between those specific sections, sometimes across multiple pages. An example of this is on manuals pages where multiple manuals will each include large sets of accordions and will reference between specific sections for ease of access to that content. [Live example](https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#associations).
```
chris-gds, line 279: Further to our chat -  I think accordions on `manuals-frontend` needs more consideration in general. We need to work out how we can do this feature, work out if the API changes can be done.

However for this issue - it's a good fix knowing what we know.

chris-gds (approved):

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -283,5 +290,41 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+  GemAccordion.prototype.openByAnchorOnLoad = function () {
+    if (window.location.hash && this.$module.querySelector(window.location.hash)) {
+      this.openForAnchor(window.location.hash)
+    }
+  }
+
+  // Add event listeners for links to open accordion sections when navigated to using said anchor links on the page
+  // Adding an event listener to all anchor link a tags in an accordion is risky but we circumvent this risk partially by only being a layer of accordion behaviour instead of any sort of change to link behaviour
+  GemAccordion.prototype.addEventListenersForAnchors = function () {
+    var links = this.$module.querySelectorAll('.gem-c-accordion__section-content a[href*="#"]')
```
chris-gds, line 303: Same with this could be swapped for `this.sectionInnerContent`

PR #1946, opened 2021-02-19
Add default aria label for navigation items

owenatgov:
## What
Adds a default value for `navigation_aria_label` to the `navigation_items` sub-component.

## Why
Currently, `navigation_aria_label` is passed directly to `navigation_items` without any defaulting under the assumption that it's only used by `layout_header`. This however breaks [anything that's only using the separately exposed navigation_items](https://github.com/alphagov/datagovuk_find/blob/24a0ef70958720370a42e7daa7c7b1437fc4dada/app/views/layouts/_proposition_header.html.erb). Adding a default here allows the sub-component to render without needed to pass a custom aria label if it isn't required.

No visual changes.

### DIFF
--- a/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
@@ -1,3 +1,5 @@
+<% navigation_aria_label ||= "Top level" %>
+
 <% if navigation_items.any? %>
   <button role="button" class="govuk-header__menu-button gem-c-header__menu-button govuk-js-header-toggle" aria-controls="navigation" aria-label="Show or hide Top Level Navigation">Menu</button>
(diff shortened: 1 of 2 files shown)


ChrisBAshton (approved):
LGTM 👍

app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
```diff
@@ -1,3 +1,5 @@
+<% navigation_aria_label ||= "Top level" %>
```
chris-gds, line 1: Ace!
Curious about the phrasing of "Top Level" being the Content ideal

andysellick, line 1: I'm obviously late to this party, but it would be great if this text could be pulled from the translations file.

PR #1948, opened 2021-02-23
Update rubocop-govuk to 4.0.0.pre.1

kevindew:
Trello: https://trello.com/c/pqXbKZiC/215-test-rubocop-govuk-400pre1-prior-to-400-release

This bumps to the 4.0.0 pre release of rubocop-govuk which is part of trialling the new rubocop-govuk release. It seems to have gone into this app without too many problems which is good news.

All changes here are to resolve linter issues, the pin to 4.0.0.pre.1 is intended to be removed once the final release of rubocop-govuk 4.0.0 is out.

This ruby-version for this repo is changed to be the minimum this gem supports (2.6) to reflect both best practice for Ruby gem development and a cop that expects the minimum ruby version to be set and linted against.


alex-ju (approved):
Thanks for raising this, Kevin! I quite like the freshly enforced rules.

spec/features/contextual_navigation_spec.rb
```diff
@@ -259,7 +259,7 @@ def given_there_are_three_related_to_step_nav
   def given_there_are_three_related_to_step_nav_and_one_secondary_to_step_nav
     content_store_has_random_item(links: {
       related_to_step_navs: 3.times.map { random_step_nav_item("step_by_step_nav") },
-      secondary_to_step_navs: 1.times.map { random_step_nav_item("step_by_step_nav") },
+      secondary_to_step_navs: [random_step_nav_item("step_by_step_nav")],
```
ChrisBAshton, line 262: This is an improvement!

ChrisBAshton (approved):
LGTM - though I wonder what downsides the last commit has 🤔 we've explained that we usually set a more modern version than the minimum, but not why that's the case.

Would it be better to set the minimum version in the Gemfile with `ruby "~> 2.6.6"`, and leave the `.ruby-version` as the recommended 'developer' version of Ruby?

kevindew:
> LGTM - though I wonder what downsides the last commit has 🤔 we've explained that we usually set a more modern version than the minimum, but not why that's the case.

I think the only reason we set a more modern version in the `.ruby-version` is because it gets caught up in initiatives to upgrade all GOV.UK versions of Ruby (I think we even have a script to apply it). Aside from that scenario it's a relatively unusual approach for general work as it provides a lot more risks than working with the minimum version does.

> Would it be better to set the minimum version in the Gemfile with `ruby "~> 2.6.6"`, and leave the `.ruby-version` as the recommended 'developer' version of Ruby?

If you have a `required_ruby_version` and `.ruby-version` mismatch, rubocop will complain: 

> Gemspec/RequiredRubyVersion: required_ruby_version (2.6, declared in govuk_publishing_components.gemspec) and TargetRubyVersion (2.7, which may be specified in .rubocop.yml) should be equal.

which seems reasonable given you want the linting rules applied based on the version of Ruby you support. You can, if needed, tell rubocop which Ruby version to lint for in `.rubocop.yml` but this seems just another thing to maintain and I'm not convinced there is a good argument to do so.

I'm assuming your suggested constraint was meant to be `~> 2.6`, as `~> 2.6.6` would actually only allow you to install `>= 2.6.6, < 2.7`. `~> 2.6` is basically just a pessimistic version of what we have now as it means `>= 2.6, < 3` and it seems unnecessary to be pessimistic. 

I'd imagine most of the arguments that we want a newer minimum Ruby version for gems should be justified by the gem dropping support for that Ruby version.

ChrisBAshton:
Thanks @kevindew , that all sounds good to me 👍

Issue #1661, opened 2020-08-25
Components audit doesn't work on the main components website


alex-ju:
Visiting https://components.publishing.service.gov.uk/component-guide/audit show 'application not found' for every app.

<img width="1439" alt="Screenshot 2020-08-25 at 12 07 13" src="https://user-images.githubusercontent.com/788096/91167441-e00e3b00-e6cb-11ea-9f69-a77af1c845b6.png">

andysellick:
I knew about this, but it's good to record it. Unfortunately the way the audits page works we can't get the data for the applications in here. Some of the stuff in the 'components' tab should still work, but maybe we can make this page better.

I could probably add a flag to say if all of the applications aren't found, replace the content of the 'applications' tab with a better error message.

PR #1949, opened 2021-02-24
Add auditing to applications

andysellick:
## What
Make component auditing work in the component guide for applications, and add the list of component audit warnings to the main component guide page when viewed inside an application.

Fixes https://github.com/alphagov/govuk_publishing_components/issues/1661

## Why
To improve visibility and awareness of issues raised by the auditing.

## Visual Changes
Audit summary on the main page (adds the warning count and links to the audit details beneath the main title).

Screenshot 2021-02-24 at 16 46 51

Auditing on a component guide inside an application (only shows the results from this application, no need to put inside an accordion).

Screenshot 2021-02-24 at 16 46 21

Auditing (for components only) on the live component guide.

Screenshot 2021-02-25 at 15 26 39

## Speed Changes

I was concerned about the impact running the auditing would have on the main component guide page's loading time. I've tested this using `collections` and `frontend` and the difference seems negligible. `collections` takes just over 10 seconds for both this version and master, `frontend` has more of an impact with loading times of around 9 and a half seconds increasing by one second with these changes.

I think that's an acceptable change but there's probably more that can be done to optimise this code if people find it's slowing things too much, happy to discuss.


app/assets/javascripts/component_guide/application.js
```diff
@@ -1,2 +1,11 @@
...
+//= require jquery/dist/jquery
 //= require_tree ./vendor
 //= require_tree .
+//= require ../govuk_publishing_components/modules
+//= require ../govuk_publishing_components/components/accordion
+//= require ../govuk_publishing_components/components/tabs
+
+$(document).ready(function () {
+  GOVUK.modules.start()
+})
```
alex-ju, line 11:
```suggestion
//= require ../govuk_publishing_components/dependencies
//= require ../govuk_publishing_components/components/accordion
//= require ../govuk_publishing_components/components/tabs
```

Instead of importing jQuery and modules.js separately then initialising the modules script, we can follow our [generic guidance](https://github.com/alphagov/govuk_publishing_components/blob/master/docs/install-and-use.md#import-javascript-for-individual-components) and call dependencies, which covers all these. Also not sure we _need_ to call `vendor` explicitly – that was meant to provide polyfills to be imported at the component level.

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -2,284 +2,291 @@
...
-        </a>
-      </li>
-    </ul>
-    <div class="govuk-tabs__panel" id="applications">
-      <%= render "govuk_publishing_components/components/heading", {
-        text: "Applications",
-        font_size: "l",
-        margin_bottom: 6
-      } %>
+<div class="govuk-tabs" data-module="govuk-tabs">
```
alex-ju: Is there a reason for not using the component here?

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -2,284 +2,291 @@
...
-              How to use this information
-            </span>
-          </summary>
-          <div class="govuk-details__text">
-            <p class="govuk-body">This page shows information about component use on GOV.UK. This information has been cross referenced with the components in the gem to produce warnings where e.g. a print stylesheet for a component exists but has not been included in an application.</p>
-            <p class="govuk-body">Warnings should be investigated, although there may be a reason why the application has been configured as it is. Note that 'code' can refer to templates or ruby code.</p>
-          </div>
-        </details>
+    <% if @applications.any? %>
+      <details class="govuk-details" data-module="govuk-details">
```
alex-ju: Similarly here, any reason for not using the component?

alex-ju (commented):
I don't feel confident reviewing the Rails code, but I left a few suggestions on the frontend side.

app/assets/javascripts/component_guide/application.js
andysellick: I knew there must be a cleaner way of doing this, thanks, will update.

andysellick:
For anyone reviewing this - I've broken the code a bit doing a rebase and unfortunately pushed, so bear with me while I correct it.

app/views/govuk_publishing_components/audit/show.html.erb
andysellick: There was a reason for using not-the-component when the auditing was first built, I can't quite recall why, but that's no longer the case so have updated, also the details one. Shout if there's any others I've missed, thanks.

app/views/govuk_publishing_components/audit/show.html.erb
andysellick: Fixed, thanks.

injms (approved):
Looks good - tested on frontend and collections and it works nicely 👍

PR #1951, opened 2021-03-01
Add custom heading level option to radio component

danacotoran:
If the radio component is passed a heading, and the heading is not a "Page heading", then that heading is always a `h2`.

This is not always appropriate; there are instances where more flexibility is necessary in order to maintain a correct hierarchical heading structure. 

For example, on the Account information consent page:
Screenshot 2021-03-01 at 16 20 45

This introduces the shared helper's heading level function to the radio component. This way we can ensure the default 
 remains `h2`, while also having the ability to pass a custom heading level where necessary.

Relevant card: https://trello.com/c/lNkwruEi/647-fix-accessibility-issues-raised-by-dac


app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -54,7 +55,7 @@
         <% end %>
       <% else %>
         <%= tag.legend class: legend_classes do %>
-          <%= tag.h2 heading, class: "govuk-fieldset__heading" %>
+          <%= content_tag(shared_helper.get_heading_level, heading, class: "govuk-fieldset__heading" ) %>
```
andysellick, line 58: Nitpick: there's a space inside the closing `)` but none after the opening one. I'd remove it.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -198,6 +198,21 @@ examples:
         text: "Green"
       - value: "blue"
         text: "Blue"
+  with_custom_heading_level:
+    description: |
+      This allows the heading level to be changed.
+      Heading level will default to `h2` if nothing is passed.
```
andysellick, line 204: This is on a new line but yml renders it on the same line as the line above, unless there's a line space. Did you mean for this to be on a separate line?

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -184,7 +184,7 @@ examples:
         text: "Blue"
   with_custom_heading_size:
     description: |
-      This allows the size of the legend to be changed. Valid options are s, m, l, xl, defaulting to m if no option is passed. 
+      This allows the size of the legend to be changed. Valid options are s, m, l, xl, defaulting to m if no option is passed.
 
       If the is_page_heading option is true and heading_size is not set, the text size will be xl.
```
andysellick, line 189: Since we've now got this option, the `is_page_heading` option is a bit redundant/confusing. I think it would be better if these options were consolidated together, so if a heading size is passed, it gets rendered as that heading, otherwise it's not a heading. Then there'd be one option for setting the heading level, and one option for setting the heading size.

This is outside the scope of this PR but maybe worth raising an issue about?

andysellick (approved):
Looks good, +1 for using the shared helper. I've approved but noted a few minor things.

app/views/govuk_publishing_components/components/docs/radio.yml
danacotoran: Totally agreed, have raised an issue to further consolidate the component https://github.com/alphagov/govuk_publishing_components/issues/1953

app/views/govuk_publishing_components/components/docs/radio.yml
danacotoran: Good shout and nope it doesn't have to be a separate line so I've left it as one line

PR #1957, opened 2021-03-04
Update search box button

injms:
## What

Adds the ability to set the visually hidden text in the search box button component.

Updates layout heading to use the search box component with the button text "Search GOV.UK".

## Why

Duplicate buttons on a page - for example, two buttons for different forms that both have the text "Search" - can lead to confusion for people who use assistive tech. Screen readers can collate all of the forms and buttons on a page, so duplicate text would make it harder to understand which button did what. Voice control relies on unique text, so saying "Click search" may end up with the wrong search button being clicked. 

## Visual Changes

No visual changes; the visually hidden text will be customisable:
image

And the layout header text has been updated:
Screenshot 2021-03-04 at 14 38 39

### DIFF
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -2,9 +2,9 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   aria_controls ||= nil
-  button_text ||= "Search"
+  button_text ||= t("components.search_box.search_button")
   id ||= "search-main-" + SecureRandom.hex(4)
-  label_text ||= "Search on GOV.UK"
+  label_text ||= t("components.search_box.label")
   name ||= "q"
   no_border ||= false
   size ||= ""
@@ -35,7 +35,7 @@
       class: "gem-c-search__item gem-c-search__input js-class-toggle",
       id: id,
       name: name,
-      title: "Search",
+      title: t("components.search_box.input_title"),
       type: "search",
       value: value,
     ) %>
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/layout_header/_search.html.erb
```diff
@@ -1,8 +1,17 @@
-<button class="search-toggle js-header-toggle" data-search-toggle-for="search">Show search</button>
-<form id="search" class="gem-c-layout-header__search-form govuk-clearfix" action="/search" method="get" role="search">
+<button class="search-toggle js-header-toggle" data-search-toggle-for="search">
+  Show search
```
owenatgov, line 2: Another translation request for this and line 12.

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -1,8 +1,16 @@
 <%
-  size ||= ""
-  no_border ||= false
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
-  classes = %w(gem-c-search)
+
+  aria_controls ||= nil
+  button_text ||= "Search"
```
owenatgov, line 5: Could this and line 7 be moved to translations? Just for best practice.

owenatgov (commented):
Drive by review. Looking tidy overall 👍  I can give this a proper test later on today.

app/views/govuk_publishing_components/components/layout_header/_search.html.erb
injms: Yep, I should've done that. Fixed!

app/views/govuk_publishing_components/components/_search.html.erb
injms: Also done - thanks!

alex-ju (approved):
🙌

Issue #1941, opened 2021-02-18
Structure of accordion and accordion docs content needs tidying up


owenatgov:
This issue encompasses 2 problems uncovered in https://github.com/alphagov/govuk_publishing_components/pull/1937

1. There is content in the accordion js, specifically the grammatical screen reader additions for the show/hide buttons, which is isolated to that js. It could be added to a translation file and instead passed to the component in the markup, reducing the risk of the wrong language being shown to users. This is also true for the step by step component.
2. Throughout the docs for the accordion, there are several single lines of content surrounded by quotes. This isn't necessary in YML.

andysellick:
The step by step component should already be getting its text from data attributes in the template, from the translation file - unless it's changed recently.

owenatgov:
The "this section" stuff isn't, that was added hastily during the run-up to WCAG-geddon last September.

owenatgov:
This has now been solved by the linked PR

PR #1958, opened 2021-03-04
Tidy up untranslated content and formatting on accordion docs

owenatgov:
## What
User facing: Moves some loose content into our translation file.

Docs facing: Tidies up some of the accordion documentation, specifically removing unnecessary quote marks on single line content.

## Why
Principally best practice changes. Serving our content via translation files allows us to easily serve alternative content by locale in the event that it's required.

Addresses https://github.com/alphagov/govuk_publishing_components/issues/1941 and @andysellick's comment on https://github.com/alphagov/govuk_publishing_components/pull/1946

No visual changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Tidy up untranslated content and formatting on accordion docs ([PR #1958](https://github.com/alphagov/govuk_publishing_components/pull/1958)) PATCH
+
 ## 24.9.1
 
 * Change to the exceptions list for Brexit branding ([PR #2011](https://github.com/alphagov/govuk_publishing_components/pull/2011))
+
 
 ## 24.9.0
 
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -31,7 +31,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.upChevonIconClass = 'gem-c-accordion-nav__chevron'
     this.downChevonIconClass = 'gem-c-accordion-nav__chevron--down'
 
-    // Indicate that js has worked
+    // Translated component content and language attribute pulled from data attributes
+    this.$module.actions = {}
+    this.$module.actions.locale = this.$module.getAttribute('data-locale')
+    this.$module.actions.showText = this.$module.getAttribute('data-show-text')
+    this.$module.actions.hideText = this.$module.getAttribute('data-hide-text')
+    this.$module.actions.showAllText = this.$module.getAttribute('data-show-all-text')
+    this.$module.actions.hideAllText = this.$module.getAttribute('data-hide-all-text')
+    this.$module.actions.thisSectionVisuallyHidden = this.$module.getAttribute('data-this-section-visually-hidden')
+
+    // Indicate that JavaScript has worked
     this.$module.classList.add('gem-c-accordion--active')
 
     this.initControls()
@@ -117,7 +126,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     // Build additional copy for assistive technology
     var srAdditionalCopy = document.createElement('span')
     srAdditionalCopy.classList.add('govuk-visually-hidden')
-    srAdditionalCopy.innerHTML = ' this section'
+    srAdditionalCopy.innerHTML = this.$module.actions.thisSectionVisuallyHidden
+
+    if (this.$module.actions.locale) {
+      srAdditionalCopy.lang = this.filterLocale('this_section_visually_hidden')
+    }
 
     // Build additional wrapper for toggle text, place icon after wrapped text.
     var wrapperShowHideIcon = document.createElement('span')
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -31,6 +31,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.upChevonIconClass = 'gem-c-accordion-nav__chevron'
     this.downChevonIconClass = 'gem-c-accordion-nav__chevron--down'
 
+    // component content pulled from translated data attributes
```
injms, line 34: Just to be consistent with your other comments:

```suggestion
    // Component content pulled from translated data attributes
```

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -31,6 +31,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    this.$module.actions = {}
+    this.$module.actions.showText = this.$module.getAttribute('data-show-text')
+    this.$module.actions.hideText = this.$module.getAttribute('data-hide-text')
+    this.$module.actions.showAllText = this.$module.getAttribute('data-show-all-text')
+    this.$module.actions.hideAllText = this.$module.getAttribute('data-hide-all-text')
+    this.$module.actions.thisSectionVisuallyHidden = this.$module.getAttribute('data-this-section-visually-hidden')
+
     // Indicate that js has worked
```
injms, line 42: Whilst we're here...

```suggestion
    // Indicate that JavaScript has worked
```

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -40,21 +40,21 @@ examples:
     data:
       items:
         - heading:
-            text: "Writing well for the web"
+            text: Writing well for the web
```
injms, line 43: Out of interest, why the all removal of the quotes? I thought we used them to avoid YAML's liberal use of things that could be booleans / Norway = false problem.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -14,6 +14,11 @@
   data_attributes ||= {}
   data_attributes[:module] = 'gem-accordion'
   data_attributes[:anchor_navigation] = anchor_navigation
+  data_attributes[:show_text] = t("components.accordion.show")
+  data_attributes[:hide_text] = t("components.accordion.hide")
+  data_attributes[:show_all_text] = t("components.accordion.show_all")
+  data_attributes[:hide_all_text] = t("components.accordion.hide_all")
+  data_attributes[:this_section_visually_hidden] = t("components.accordion.this_section_visually_hidden")
```
injms, line 21: The i18n `t` method will fall back to the default language when a key isn't present in the current language - so that could mean that a page in Spanish with the accordion might have the controls in English if there's no translations available. 

This mixing of languages is fine, as long as the `lang` attribute is used. Is this something that should be added?

app/views/govuk_publishing_components/components/docs/accordion.yml
owenatgov: This was based on recommendations by Andy that yml doesn't need quotes around single line text. If there's a standard that this goes against or we just think it's a bad idea I'm not precious about it.

app/views/govuk_publishing_components/components/_accordion.html.erb
owenatgov: Good point. I'll drop in some `lang` magic to account for this.

app/views/govuk_publishing_components/components/docs/accordion.yml
injms: Andy's right, the quotes aren't needed in most cases - the main exception to this is that YAML tends to be quite enthusiastic about turning unquoted strings into booleans. There isn't a GDS Way standard, so I tend to follow the file's convention to avoid bike shedding.

owenatgov: If you're ok with it I'm gonna leave this how it is. These strings don't seem to be incurring yml's wrath right now and it also fixes some general formatting in the doc itself. It's not a complex job to undo it if it's a problem and I'm not precious about this if you'd rather I change it now.

injms: Of course, that's absolutely fine - I was just wondering if there was some documentation somewhere or a discussion that I'd missed.

app/views/govuk_publishing_components/components/_accordion.html.erb
injms: That works for now, since the translation is only available in English. There's now a risk that someone could now add translations for another language without realising that the `lang` attribute is hardcoded though.

injms: Since https://github.com/alphagov/govuk_publishing_components/pull/1992 was merged in and released, I think it might be worth thinking about how we can get this working without hardcoding the `lang` attribute. Happy to pair on this if you'd like?

owenatgov: Agreed. Lemme have a crack at it and I'll get in touch with you if I run into problems.

owenatgov: Updated. Lemme know your thonks.

owenatgov: Following a slack chat I've now updated this again to instead use `t_locale` plus some fancy data attribute management. Lemme know what you think.

injms (approved):
Super 👍

CHANGELOG.md
```diff
@@ -7,9 +7,14 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+<<<<<<< HEAD
```
alex-ju, line 10: We have some rebase artefacts that we'd need to remove before merging.

owenatgov: How did that get in there? 🙃  Will sort out.

PR #1963, opened 2021-03-08
Swap `sanitize` for `raw` in list component

injms:
## What

The list component used the `sanitize` method. This has been swapped for the `raw` method. 

Tests to ensure that attributes on markup passed into component are unaltered have been added.

## Why

The sanitize method strips out elements and attributes that are not in an allowlist from a string, and then allows the string to be added as markup rather than being added as a string. This could strip out `data-` and other useful attributes given to the list component.

 The `raw` method doesn't alter the markup - so any string passed into the component will be used verbatim. Since the strings being given to the list component come from within an application, we can presume these to be trustworthy.

## Visual Changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_list.html.erb
@@ -1,12 +1,12 @@
 <%
-  id ||= nil
+  aria_label ||= nil
   extra_spacing ||= nil
+  id ||= nil
+  items ||= []
   list_type ||= "unordered"
   visible_counters ||= nil
-  items ||= []
-  aria_label ||= nil
 
-  classes = %w(gem-c-list govuk-list)
+  classes = %w[gem-c-list govuk-list]
   classes << "govuk-list--bullet" if visible_counters && list_type === "unordered"
   classes << "govuk-list--number" if visible_counters && list_type === "number"
   classes << "govuk-list--spaced" if extra_spacing
@@ -20,7 +20,7 @@
 <% if items.any? %>
   <%= content_tag list_tag, class: classes, id: id, "aria-label": aria_label do %>
     <% items.each do |item| %>
-      <li><%= sanitize(item) %></li>
+      <li><%= raw(item) %></li>
     <% end %>
   <% end %>
 <% end %>
--- a/spec/components/list_spec.rb
+++ b/spec/components/list_spec.rb
@@ -87,7 +87,7 @@ describe "List", type: :view do
     assert_select "ul#this-is-a-test-id li:nth-child(2)", text: "Another test item"
   end
 
-  it "adds links within a list" do
+  it "adds markup within the list items" do
     render_component(
       items: [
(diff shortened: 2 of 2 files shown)


app/views/govuk_publishing_components/components/_list.html.erb
```diff
@@ -1,12 +1,12 @@
 <%
-  id ||= nil
+  aria_label ||= nil
```
andysellick, line 2: Not blocking, just curious why you moved these lines around?

andysellick (approved):

andysellick:
Also forgot to mention, need to update the changelog.

kevindew (approved):
This looks good to me and looks like it'll fix the problem 👍 

In general we should avoid `raw` as it does present XSS risks (it's preferable that the client calling the component marks a string as `.html_safe`). However I can't see a way around that in this component without serious BC breaks and well, for this component, there seems a real expectation that any sanitisation needs to be done before calling this component.

app/views/govuk_publishing_components/components/_list.html.erb
injms: So they're alphabetical - I find it makes it easier to find a specific variable. And alphabetical because it's simple and unambiguous.

injms:
@kevindew I thought that `raw()` was preferred over `.html_safe`, and that they basically did the same thing? (Only difference is that `raw()` just has the slight advantage because it does `.to_s` and then `.html_safe` on anything passed to it.)

kevindew:
> @kevindew I thought that `raw()` was preferred over `.html_safe`, and that they basically did the same thing? (My understanding is that the only difference is that `raw()` just has the slight advantage because it does `.to_s` and then `.html_safe` on anything passed to it.)

They serve slightly different purposes, so I don't think it's quite a case of one is preferred over the other as it's contextual - and in the context of this PR I think `.raw` is a better choice than `.html_safe`. You're right that they do very similar things and I think the difference between them is just better communication of purpose.

Generally you use `.html_safe` as a means to specify that data can be rendered later as HTML. It's probably better explained with an example, as I imagine we're on slightly different wavelengths

```
# hypothetical_component.html.erb
<h1><%= raw title %></h1>
```

Allows someone to do:

```
<%= render "hypothetical_component", title: "My title with a <a href='/'>a link</a>" %>
```

But risks an XSS issue with

```
<%= render "hypothetical_component", title: "A title for #{params[:unsafe_user_input]}" %>
```

Whereas a safer approach moves those concerns to the caller:

```
# hypothetical_component.html.erb
<h1><%= title %></h1>
```

Where someone has to say when they expect html escaping to be removed:

```
<%= render "hypothetical_component", title: "My title with a <a href='/'>a link</a>".html_safe %>
```

And by default protects against:

```
<%= render "hypothetical_component", title: "A title for #{params[:unsafe_user_input]}" %>
```

However, like I said, all of this seems impractical for this component as it'd break a lot of use cases.

PR #1967, opened 2021-03-10
Add title option to summary list links

andysellick:
## What
- link titles for 'change' and 'delete' links by default are automatically generated based on the text 'Change/Delete [thing]' but this causes problems when the thing and the link text don't naturally follow
- adding in the option to pass explicit title attributes for the links

## Why
We've got a situation in Accounts where the text doesn't work with the automatically generated titles, so adding an option to pass explicit ones seems useful.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


danacotoran (approved):
👍

PR #1971, opened 2021-03-11
Summary list changes

andysellick:
## What
Does the following:

- adds an option to remove the visually hidden span the component automatically adds into link text, containing the row or overall name
- adds an option for a wider layout for the field name

## Why
The first change is because I have a situation where the text doesn't work for the default behaviour. The text would be `Change GOV.UK can send you feedback emails` - which doesn't make sense. Instead I'd like to be able to pass custom text without the component automatically appending visually hidden text that doesn't make sense.

The second change is because again the text I need to put in the component doesn't look great with the current layout.

## Visual Changes
No change for the first change, new layout option looks like this.

Screenshot 2021-03-11 at 12 18 55

Trello card: https://trello.com/c/lNkwruEi/647-fix-accessibility-issues-raised-by-dac

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Summary list changes ([PR #1971](https://github.com/alphagov/govuk_publishing_components/pull/1971))
+
 ## 24.5.0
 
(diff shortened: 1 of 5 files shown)


danacotoran:
@andysellick The wide title layout changes look good to me, however I am still confused as to why there is a need for both the custom `title`, and the custom hidden text option on the links. 

I might be missing something obvious here, it would be good to have a conversation about it when you're around.

danacotoran (approved):
LGTM :)

PR #1973, opened 2021-03-15
Remove title attribute from summary list

andysellick:
## What
Remove the title attributes from the links in the summary list component. These are automatically generated based on the link text and the thing they refer to, or passed to the component directly (a change I recently added).

## Why
We don't usually use the title attribute for links on GOV.UK, instead relying on visually hidden text within the link text. What's strange is that this component already does this, so the title attribute seems an unnecessary addition, particularly when the original component in the Design System doesn't have title attributes on links - https://design-system.service.gov.uk/components/summary-list/.

There's also evidence that screen readers treat it in various ways. Specifically https://www.deque.com/blog/text-links-practices-screen-readers/ highlights this, and this article https://www.powermapper.com/tests/screen-readers/labelling/a-title/ shows that it can cause problems in 17 screen reader / browser combinations.

## Visual Changes
None.

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/summary_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/summary_list.yml
@@ -71,12 +71,12 @@ examples:
 
       The text of the link can be customised using the `link_text` option. Take care that the provided `link_text` still makes sense to screen readers when combined with the title. For instance, `Reorder` link text and `Items` title becomes `Reorder Items`, which reads fine, but link text of `Summary` would read as `Summary Items`, which does not make sense.
 
-      To override this behaviour, use `link_text_no_enhance` on the link in question, as shown. This will remove the visually hidden text, allowing you to add your own. The component will render a `span` element passed in a capture block.
+      To override this behaviour, use `link_text_no_enhance` on the link in question, as shown. This will remove the visually hidden text, allowing you to add your own. The component will render a `span` element with a class of `govuk-visually-hidden` included in a passed [capture block](https://apidock.com/rails/ActionView/Helpers/CaptureHelper/capture) as the `link_text` parameter.
     data:
       <<: *default-example-data
       edit:
         href: "edit-title-summary-body"
-        link_text: Edit this document
+        link_text: Edit this document <span class="govuk-visually-hidden">my hidden text for screenreaders</span>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/summary_list.yml
```diff
@@ -73,13 +72,10 @@ examples:
       The text of the link can be customised using the `link_text` option. Take care that the provided `link_text` still makes sense to screen readers when combined with the title. For instance, `Reorder` link text and `Items` title becomes `Reorder Items`, which reads fine, but link text of `Summary` would read as `Summary Items`, which does not make sense.
```
danacotoran, line 72: Can't put a comment on L68 so I'll just leave it here.

Would this variant be better renamed to `with_no_hidden_text`, or something like that? 

Or alternatively could the example be adjusted to better represent what is detailed in the description? The example showcases what the `link_text_no_enhance` parameter does, but doesn't show how to replace the default visually hidden text with your custom one.

alex-ju (approved):
I did a bit of investigation in an attempt to shed a bit of light on how we ended up with this. Content Publisher (then later Collection Publisher), by design, needed a `summary` component. At the time this was only available as a pattern in the prototype kit. The component was ported from the prototype kit into Content Publisher, where [`title` referred to the section title](https://github.com/alphagov/content-publisher/commit/0378faf5e2fbffb49388c83bb6b91f7be24edfb6#diff-3406468b5d06b8ac1daaec1581f64ae93f7976040df447ad3b957e4abf6920cfR2) – it seems that over time this attribute morphed into being used as an HTML `title` attribute – hence the redundant element.

Happy for this to be removed as we now have the visually hidden element.

danacotoran (commented):
Just one comment, I think it looks good otherwise.

app/views/govuk_publishing_components/components/docs/summary_list.yml
andysellick: I think this is a bit out of scope for this PR but it seems important so I'm adding it in. I think clarifying the description is better than renaming the example, so I've given it a shot. Hopefully this is better.

danacotoran (approved):
Thanks for updating the doc with my suggestion 👍

PR #1974, opened 2021-03-15
Rework accordion anchor link nav to allow for colon anchors

owenatgov:
## What
Rework the anchor link navigation functionality of the accordion component to remove the hash from the id when navigating and use `getElementById` instead of `querySelector ` when searching for those anchor ids.

## Why
There is a pre-existing feature in manuals frontend that allows for the navigation between footnotes. These anchor links are marked up using the format of `fnref:[footnote number]`. [Example of this in use](https://www.gov.uk/guidance/drivers-hours-goods-vehicles/1-eu-and-aetr-rules-on-drivers-hours). `querySelector` doesn't like that colon in id queries because it [interferes with existing CSS syntax](https://bugzilla.mozilla.org/show_bug.cgi?id=883044). This change allows it to work and fixes this bug in manuals frontend.

No visual changes

[Card](https://trello.com/c/pdwRd2nA/683-fix-footnotes-on-manuals-accordions)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Summary list changes ([PR #1971](https://github.com/alphagov/govuk_publishing_components/pull/1971))
+* Rework accordion anchor link nav to allow for colon anchors ([PR #1974](https://github.com/alphagov/govuk_publishing_components/pull/1974))
 
 ## 24.5.0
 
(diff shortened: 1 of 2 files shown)


danacotoran:
This is outside the scope of this change, but I wonder if we should have tests for all this javascripty stuff in the accordion.

There is an `accordion_spec.rb` which is all fine and dandy, but would it be good to have tests for the more "stateful" JS-driven bits too? There's been a lot of good work on this component recently, however I wonder if our testing of certain aspects of this component could be a bit more robust.

owenatgov:
@danacotoran Totally agree. I'm currently putting together a js spec separate to this PR. My broad plan is to get this into a release sooner so that it fixes a bug and then write js tests for the functionality against this once it's merged. Happy to negotiate this.

danacotoran:
@owenatgov ah thanks, it's good you're already aware of this and making steps towards it 👍 
(just to clarify I wasn't expecting you to do it as part of this PR, it's definitely outside the scope of this work, was just raising it)

danacotoran (approved):
I have tested this locally and it seems to work as intended. It doesn't break anything else that I can see 👍

PR #1978, opened 2021-03-16
Fix bug in summary list link text

andysellick:
## What
- `link_text_no_enhance` for row links was being checked against the overall links for the component, so passing the parameter for the row when not passing it for the overall links had no effect
- tests were not specific enough to spot this
- now corrected

## Why
It weren't working.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


danacotoran (approved):
Thanks for fixing this and making the tests more specific. And sorry I missed it in the previous review 🤦🏻‍♀️ 

Might be good to amend the second commit to reflect that it's updating the changelog.

andysellick:
Whoops, meant to squash those commits. Done, thanks!

PR #1980, opened 2021-03-17
Add spacing to header menu

injms:
## What

The menu appeared differently on smaller screens when using with left side search and when used on it's own without search - this tweak makes sure that the spacing is the same for both of these uses of the menu.

## Why

So that the menu is consistent.

## Visual Changes

Menu when used with left side search:

image

Before:
image

After: 
image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -54,7 +54,6 @@
   .gem-c-header__content {
     position: relative;
     width: 100%;
-    padding-bottom: govuk-spacing(2);
   }
 
   @include govuk-media-query($from: desktop) {
@@ -103,8 +102,9 @@
 }
 
 .gem-c-header__content.govuk-header__content {
-  width: auto;
   padding-left: 0;
+  padding-bottom: govuk-spacing(2);
+  width: auto;
 
   @include govuk-media-query($from: desktop) {
     float: right;


danacotoran:
@injms good eyes for detail, I've never noticed this before 👏 

I'm going to nitpick and say it's not very obvious from your PR description and from the screenshots what's actually changed. It would be good to specify exactly what has changed in terms of spacing in the description (the main difference I noticed between your screenshots was that one was using the wrong font so that confused me for a second).

Also while we're at it, I was wondering if it would be good to make the left-right padding consistent as well. I noticed the variant with search has some extra padding to the right, for whatever reason: 
Screenshot 2021-03-17 at 14 42 29

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -103,8 +102,9 @@
 }
 
 .gem-c-header__content.govuk-header__content {
-  width: auto;
   padding-left: 0;
+  padding-bottom: govuk-spacing(2);
```
danacotoran, line 106: Unfortunately this seems to be affecting vertical spacing on the header (even on the variants without the menu). It's adding some extra space below the logo: 
Before | After
------------ | -------------
Screenshot 2021-03-17 at 14 58 57 | Screenshot 2021-03-17 at 14 58 36

injms: Ah, bother - this isn't the quick fix I was hoping it was.

injms:
This isn't the quick fix I was hoping, and I don't have the time or headspace at the moment to give this a deeper look.

PR #1982, opened 2021-03-18
Fix component preview

andysellick:
## What
Adds some padding to the simple component preview, to make component previewing on 'preview' and 'preview all' screens better.

## Why
I started work on a problem with the search component and found that on the 'preview all' screen the component on a dark background doesn't show the dark/blue background. Turns out that's because on those screens there's no padding, and the component completely covered the dark/blue background.

## Visual Changes

Yes, it indents the previews a little bit, but it does that on the main page already and I think being able to see the preview properly in context is worth it.

Before | After
------ | ------
Screenshot 2021-03-18 at 11 11 13 | Screenshot 2021-03-18 at 11 11 20

### DIFF
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -148,7 +148,7 @@ $gem-guide-border-width: 1px;
 
 .component-guide-preview--simple {
   border: 0;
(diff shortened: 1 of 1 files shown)


injms (approved):
Love an improvement in one line!

PR #1985, opened 2021-03-18
Add GOVUK Frontend Details module to GOVUK Modules and amend modules.js `start` function

danacotoran:
# What
- Amend JS modules initialiser code to account for the case where a GOVUK JS module has a GOVUK Frontend counterpart module which also needs to be initialised
- Add GOVUK Frontend's Details JS module to our own GOVUK Modules object so that the module can be initialised with the other modules

# Why
As part of the Accounts team's DAC report it was flagged that the `details` element was not functional in Internet Explorer and assistive tech was not able to interact with this component as expected.

Some investigation revealed that:
1. this is well known issue where all versions of Internet Explorer and Edge (pre-chromium) don't fully support details/summary: https://caniuse.com/details

2. on the Design System's version of the details component, the issue would not occur, which would suggest that our components gem version is not properly importing the fix from GOVUK Frontend.

After some digging I realised that while we are requiring `govuk-frontend`'s `details.js` script in our own component javascript, it is never initialised, therefore we were not applying their polyfill for IE.

These changes add GOVUK Frontend's Details JS module to our own GOVUK Modules object so that the module can be initialised with the other modules, which should automatically resolve our IE issue.

This also changes the GOVUK modules start function. The way the module start function works currently is that it will either init a GOV.UK Publishing & Legacy Module, or if that doesn't exist, it will init the GOV.UK Frontend equivalent if it exists.
This does not work well in situations where we need both, for instance in the case of the `details` component, where we have a `GovukDetails.js `, but also a `Details.js` (from GOV.UK Frontend) which both need to run for the details component to work correctly.

### Before – details component does nothing on IE

https://user-images.githubusercontent.com/7116819/111684484-ac892e80-881e-11eb-848b-2048adde7d77.mov

### After – details component works as expected on IE

https://user-images.githubusercontent.com/7116819/111684469-a85d1100-881e-11eb-9d1e-8f8960bd3519.mov

https://trello.com/c/lNkwruEi/647-fix-accessibility-issues-raised-by-dac

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/details.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/details.js
@@ -1,7 +1,8 @@
 /* eslint-env jquery */
 // = require govuk/components/details/details.js
 window.GOVUK = window.GOVUK || {}
-window.GOVUK.Modules = window.GOVUK.Modules || {};
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.Details = window.GOVUKFrontend;
 
 (function (Modules) {
   function GovukDetails () { }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add GOVUK Frontend Details module to GOVUK Modules and amend modules.js start function ([PR #1985](https://github.com/alphagov/govuk_publishing_components/pull/1985))
 * Rescope Brexit CTA to en/cy locale only ([PR #1984](https://github.com/alphagov/govuk_publishing_components/pull/1984))
 * Add js tests for accordion component ([PR #1977](https://github.com/alphagov/govuk_publishing_components/pull/1977))
 * Fix search component label background ([PR #1983](https://github.com/alphagov/govuk_publishing_components/pull/1983))
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -40,7 +40,9 @@
           module = new GOVUK.Modules[moduleName]()
           module.start(element)
           element.data('module-started', true)
-        } else if ( // GOV.UK Frontend Modules
+        }
+
+        if ( // GOV.UK Frontend Modules
           typeof GOVUK.Modules[frontendModuleName] === 'function' &&
           GOVUK.Modules[frontendModuleName].prototype.init &&
           !started
--- a/spec/javascripts/govuk_publishing_components/modules.spec.js
+++ b/spec/javascripts/govuk_publishing_components/modules.spec.js
@@ -84,13 +84,29 @@ describe('GOVUK Modules', function () {
       GOVUK.Modules.GovukTestAlertModule = GovukTestAlertModule
 
       // GOV.UK Frontend Modules
-      function TestAlertModule (element) {
+      function TestAlertFrontendModule (element) {
         this.element = element
       }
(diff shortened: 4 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -1,16 +1,19 @@
 /* eslint-env jquery */
 // = require govuk/components/details/details.js
 window.GOVUK = window.GOVUK || {}
-window.GOVUK.Modules = window.GOVUK.Modules || {};
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.Details = window.GOVUKFrontend;
```
alex-ju, line 5: Good find! It seems like this was [missed when the details script was brought over from `govuk-frontend`](https://github.com/alphagov/govuk_publishing_components/commit/6112261619d47d8f930e58387b66023deb657cc7#diff-a235b89aac8e8310ee228b267a5837df7801e2c6510b4f9772077a88abbf56c0R5). This line should be enough to fix the problem; the initialisation will be automatically done via [`modules.js`](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/modules.js#L48).

danacotoran: > This line should be enough to fix the problem

@alex-ju I would have thought so too, in fact that was the first thing I tried, however it didn't seem to be enough. I've done  some digging and I think what is happening here is:
- the details component has `data-module="govuk-details"`, therefore in the [modules.js start function](https://github.com/alphagov/govuk_publishing_components/blob/176d844f0085ccb8bcf6bc808987e26cdcd17530/app/assets/javascripts/govuk_publishing_components/modules.js#L31), `moduleName` is set as `GovukDetails` and `frontendModuleName` is `Details`
- the [first condition](https://github.com/alphagov/govuk_publishing_components/blob/176d844f0085ccb8bcf6bc808987e26cdcd17530/app/assets/javascripts/govuk_publishing_components/modules.js#L35) in the if statement evaluates to `true` because `GOVUK.Modules[GovukDetails]` exists and it has an `init` function –  so the code at L40-L42 runs
- since the condition before it was true, the `else if` below does not run. Therefore the GOVUK Frontend `Details` module does not get initialised even though it exists and is "initialisable" 😄

danacotoran: Basically I think there’s a bigger issue with how we init modules. What I did is really a workaround. Not sure if we should try and go with the workaround for now and try to resolve the core issue later. Let me know what you think, anyway.

alex-ju: Oh, that is bad! I was wondering how other scripts work (for example checkboxes) and I was surprised to learn that we use [an extra `data-module` attribute](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_checkboxes.html.erb#L26) at a lower level in the markup tree 🙈. Yes please, splitting the conditional statement in `modules.js` sounds like a good idea.

alex-ju: If you have other things you're currently working on let me know and I can pick this up. However works best for you.

danacotoran: Thanks for offering @alex-ju, I've just pushed the latest changes to this branch. 
Will be updating the PR desc to reflect what's changed.

alex-ju (approved):
Thank you for taking the time to address the core issue here!

PR #2002, opened 2021-03-30
Remove `@extend` from component Sass

injms:
## What

Removes the use of `@extend` from the 
* subscription links
* step by step
* related navigation
* contextual sidebar
* Magna Charta
* highlight boxes
* translation navigation
* share links
* print link
* previous and next link
* image card

components' stylesheets and moves the extended classes into the markup.

## Why

To enable the component to be used when only importing `frontend_support`.

The subscription links, step by step, related navigation, contextual sidebar, and Magna Charta components all `@extend` parts of GOV.UK Frontend, so they need 

```scss
@import "govuk_publishing_components/component_support";
```
in order to work. This increases the size of the stylesheets being served. For example:

```scss
@import "govuk_publishing_components/component_support";
@import "govuk_publishing_components/components/title";
```
gives a stylesheet that is 155KB in size.

Using only `govuk_frontend_support` means a much smaller stylesheet:

```scss
@import "govuk_publishing_components/govuk_frontend_support";
@import "govuk_publishing_components/components/title";
```

This results in a stylesheet that's 1.3KB in size.

This is helpful for GOV.UK frontend apps that use Static - otherwise there shouldn't be any noticeable difference.

Static uses `component_support` for the gem layout - this imports all of the GOV.UK Frontend core, objects, utilities, and  override classes. The classes that are compiled are then in Static's stylesheet, which is then available to each frontend app that uses Static because Static provides the head, header, and footer for each page.

Using `component_support` in a frontend application would re-import these classes again, adding unnecessary CSS to the page without adding a thing. 

## Visual Changes

Whilst I was here, I fixed a minor problem with the focus state of the Brexit link in the contextual sidebar:

Before:

image

After:

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contextual-sidebar.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contextual-sidebar.scss
@@ -6,7 +6,6 @@ $transition-campaign-dark-blue: #1e1348;
 }
 
 .gem-c-contextual-sidebar__brexit-heading {
-  @extend %govuk-heading-s;
   margin-top: govuk-spacing(3);
   margin-bottom: govuk-spacing(2);
 }
@@ -24,7 +23,6 @@ $transition-campaign-dark-blue: #1e1348;
   }
 
   .gem-c-contextual-sidebar__brexit-text {
-    @extend %govuk-link;
     @include govuk-font(16);
 
     margin-top: 0;
@@ -38,6 +36,8 @@ $transition-campaign-dark-blue: #1e1348;
 }
 
 .gem-c-contextual-sidebar__brexit-cta:focus {
+  box-shadow: 0 $govuk-focus-width $govuk-focus-text-colour;
+
   .gem-c-contextual-sidebar__brexit-text {
     text-decoration: none;
   }
--- a/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
@@ -15,3 +15,5 @@ $gem-secondary-button-hover-background-colour: govuk-colour("light-grey", $legac
 
 $gem-quiet-button-colour: govuk-colour("dark-grey", $legacy: "grey-1");
 $gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
+
+$gem-hover-dark-background: #dddcdb;
--- a/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
@@ -2,3 +2,16 @@
 // It provides govuk-frontend but adds no weight to the compiled CSS
 @import "components/helpers/govuk-frontend-settings";
 @import "govuk/base";
+
+@import 'components/mixins/margins';
+// TODO: remove this focus override when govuk_template is no longer used
+@import 'components/mixins/govuk-template-link-focus-override';
+@import 'components/mixins/css3';
+
+$gem-secondary-button-colour: #00823b;
+$gem-secondary-button-hover-colour: darken($gem-secondary-button-colour, 5%);
+$gem-secondary-button-background-colour: govuk-colour("white");
+$gem-secondary-button-hover-background-colour: govuk-colour("light-grey", $legacy: "grey-4");
+
+$gem-quiet-button-colour: govuk-colour("dark-grey", $legacy: "grey-1");
+$gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
@@ -11,7 +11,7 @@
       <span class="gem-c-step-nav-related__pretitle"><%= pretitle %></span>
       <% if links.length == 1 && !always_display_as_list %>
           <a href="<%= links[0][:href] %>"
-            class="gem-c-step-nav-related__link govuk-link"
+            class="govuk-link"
             data-track-category="stepNavPartOfClicked"
             data-track-action="<%= pretitle %>"
             data-track-label="<%= links[0][:href] %>"
@@ -27,7 +27,7 @@
           <% links.each do |link| %>
             <li class="gem-c-step-nav-related__link-item">
               <a href="<%= link[:href] %>"
-                class="gem-c-step-nav-related__link govuk-link"
+                class="govuk-link"
                 data-track-category="stepNavPartOfClicked"
                 data-track-action="<%= pretitle %>"
                 data-track-label="<%= link[:href] %>"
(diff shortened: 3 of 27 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_contextual-sidebar.scss
```diff
@@ -38,6 +36,8 @@ $transition-campaign-dark-blue: #1e1348;
 }
 
 .gem-c-contextual-sidebar__brexit-cta:focus {
+  box-shadow: 0 $govuk-focus-width $govuk-focus-text-colour;
```
andysellick, line 39: You can do `@include govuk-focused-text;` although I'm not 100% sure that would work here, just mentioning for reference in case it's useful.

app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
```diff
@@ -4,14 +4,14 @@
...
-    class="gem-c-step-nav-related <%= "gem-c-step-nav-related--singular" if links.length == 1 %>" 
+  <div
+    class="gem-c-step-nav-related <%= "gem-c-step-nav-related--singular" if links.length == 1 %>"
     data-module="gem-track-click">
     <h2 class="gem-c-step-nav-related__heading">
       <span class="gem-c-step-nav-related__pretitle"><%= pretitle %></span>
       <% if links.length == 1 && !always_display_as_list %>
           <a href="<%= links[0][:href] %>"
-            class="gem-c-step-nav-related__link"
+            class="gem-c-step-nav-related__link govuk-link"
```
andysellick, line 14: If the only thing the `gem-c-step-nav-related__link` class did was extend `govuk-link`, is it even necessary anymore? Could the class on this element just be `govuk-link`?

andysellick (commented):
This looks great, just a couple of comments. We should definitely be moving away from using `extend` in favour of just putting the relevant class into the markup, but I'm uncertain this will actually save us CSS size in reality - won't we still need to include `component_support` for most if not all of our applications?

I think now is probably a good time to revisit the individual asset import model, to make sure we're doing it in the most efficient way possible (and also because we built it so long ago my memory on how it all fits together is a bit rusty - at the very least, we should document it a bit better). Thanks for raising this, let's talk about it.

app/assets/stylesheets/govuk_publishing_components/components/_contextual-sidebar.scss
injms: I haven't got around to adding screenshots to show the difference as this is still in draft. I'm almost doing the same as `govuk-focused-text` - this doesn't add the the slight yellow border at the top that appears above the red border.

app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
injms: Very probably. I'll check once I've got this working with [alphagov/frontend](https://www.github.com/alphagov/frontend), then I'll optimise.

injms:
> I'm uncertain this will actually save us CSS size in reality - won't we still need to include `component_support` for most if not all of our applications?

Short answer, yes but no.

Longer answer - it depends on the template that the app uses. 

This work is to prepare for the gem layout in https://github.com/alphagov/static/pull/2411. These changes include adding `component_support` to Static's `application.css` - this means that Static supplies the GOV.UK Frontend CSS which is then part of CSS's glorious global namespace. 

As those styles are coming from the stylesheet that Static provides, it seems pointless (and occasionally frustrating) to re-include them in the app again.

I'm testing this in [alphagov/frontend][custom-frontend]. As our convention is to use nice flat BEM, there are a few funky side effects and a lot of repeated styles. For example:

```html
<p class="emergency-banner__description govuk-body">Emergency banner description</p>
```
 has styles from `govuk-body` and styles from `emergency-banner__description` (thanks to alphagov/static), which is then overwritten by `govuk-body` (thanks to alphagov/frontend):

image

So the aim of this pull request is to prevent the repeated CSS caused by the double `@import` of `component_support`.

This is backwards compatible as the aim is for either both `component_support` and `govuk_frontend_support`, or only `govuk_frontend_support` to be used. There should always be `component_support` supplied to the page from somewhere - whether in Static in a separate stylesheet, or in the app itself if Static is deprecated. 

Once this is being used, then we can remove bespoke CSS isn't needed - for example, [these lines from the homepage styles][homepage-styles] are redundant when using the gem layout.

> I think now is probably a good time to revisit the individual asset import model, to make sure we're doing it in the most efficient way possible (and also because we built it so long ago my memory on how it all fits together is a bit rusty - at the very least, we should document it a bit better). Thanks for raising this, let's talk about it.

This is purposefully a draft pull request to try and show what I'm trying to achieve as I realise it's a bit convoluted - lots of things relying on other things that rely on yet more things. If there's way that's better or that people prefer then super!

[static]: https://github.com/alphagov/static
[custom-frontend]: https://github.com/alphagov/frontend
[homepage-styles]: https://github.com/alphagov/frontend/blob/master/app/assets/stylesheets/views/_homepage.scss#L1-L27

app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb
injms: I've removed the `gem-c-step-nav-related__link` from this component. This was quite tied into the tests, but aside from that I couldn't find a reason to keep the class there.

app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
```diff
@@ -2,3 +2,18 @@
...
+
+$gem-secondary-button-colour: #00823b;
+$gem-secondary-button-hover-colour: darken($gem-secondary-button-colour, 5%);
+$gem-secondary-button-background-colour: govuk-colour("white");
+$gem-secondary-button-hover-background-colour: govuk-colour("light-grey", $legacy: "grey-4");
+
+$gem-quiet-button-colour: govuk-colour("dark-grey", $legacy: "grey-1");
+$gem-quiet-button-hover-colour: darken($gem-quiet-button-colour, 5%);
+
+$gem-hover-dark-background: #dddcdb;
```
andysellick, line 19: It feels like this line (and the button stuff above it) would be more appropriate in `component_support.scss` but I'm guessing there's a reason it's in here?

andysellick (approved):
Based on today's discussion I'm going to approve this, also I totally forgot I reviewed it already and didn't follow up after my comments were addressed, sorry! Just one thing, but I don't think it's blocking.

andysellick:
Also does this need a CHANGELOG entry? I think we should be tracking this stuff.

app/assets/stylesheets/govuk_publishing_components/govuk_frontend_support.scss
injms: I placed this in `govuk_frontend_support.scss` because it doesn't add anything to the compiled stylesheet, and can be used either on its own or alongside `component_support.scss`. It gives flexibility, since `govuk_frontend_support.scss` should always be there, and `component_support.scss` is optional.

I think that these two files should be renamed - but that's a breaking change, and I didn't think it necessary here.
If these files were to be renamed, I'd suggest naming them to reflect that one doesn't add anything to the compiled CSS, and one that does.

andysellick: Fair enough, sounds sensible. Yes, I also agree now that we should rename them, because they're a bit confusing. But I remember struggling to come up with good names when I created them!

PR #1929, opened 2021-02-16
Update success notice component

maxgds:
## What
Updates the [success alert component](https://components.publishing.service.gov.uk/component-guide/success_alert) to be inline with the design system [success variant of the notification banner component](https://design-system.service.gov.uk/components/notification-banner/).

## Why
This is a final piece of work to make sure that our cookie banner and general cookie strategy is inline with design system guidance. The design system [cookie page](https://design-system.service.gov.uk/patterns/cookies-page/) uses their notifications banner so this is just to make sure we're consistent.

[Card](https://trello.com/c/B3zI99d8/654-make-sure-cookies-page-is-in-line-with-design-system-guidance)

[Relevant page](https://www.gov.uk/help/cookies)

### Before
Screenshot 2021-03-25 at 15 44 38

### After
Screenshot 2021-03-25 at 15 55 20

PR #2004, opened 2021-04-06
Tidy up success alert markup

owenatgov:
## What
- Adds a hashed ID to the success alert heading ID
- Add the option to set a custom ID for the title and the `aria-labelledby`
- Removes an unnecessary `div` from the success alert markup
- Adds some guidance advising devs to pass sanitised/raw'd markup to the `description` attribute

## Why
In response to comments made by Alex post-merge on https://github.com/alphagov/govuk_publishing_components/pull/1929

No visual changes

### DIFF
--- a/app/views/govuk_publishing_components/components/_success_alert.html.erb
+++ b/app/views/govuk_publishing_components/components/_success_alert.html.erb
@@ -1,23 +1,24 @@
-<% description ||= nil %>
+<%
+  description ||= nil
+  title_id ||= "govuk-notification-banner-title-#{SecureRandom.hex(4)}"
+%>
 
 <%= tag.div class: "gem-c-success-alert govuk-notification-banner govuk-notification-banner--success",
   role: "alert",
   tabindex: "-1",
   aria: {
-    labelledby: "govuk-notification-banner-title",
+    labelledby: title_id,
   },
   data: {
     module: "initial-focus",
   } do %>
   <div class="govuk-notification-banner__header">
-    <h2 class="govuk-notification-banner__title" id="govuk-notification-banner-title">
-      <%= t("govuk_component.success_alert.success", default: "Success") %>
-    </h2>
+    <%= tag.h2 t("govuk_component.success_alert.success", default: "Success"), class: "govuk-notification-banner__title", id: title_id %>
   </div>
   <div class="govuk-notification-banner__content">
     <% if description.present? %>
       <%= tag.h3 message, class: "govuk-notification-banner__heading" %>
-      <%= tag.div description %>
+      <%= description %>
     <% else %>
       <%= tag.p message, class: "govuk-body gem-c-success-alert__message" %>
     <% end %>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_success_alert.html.erb
```diff
@@ -1,4 +1,7 @@
-<% description ||= nil %>
+<%
+  description ||= nil
+  id = "govuk-notification-banner-title-#{SecureRandom.hex(4)}"
```
alex-ju, line 3:
```suggestion
  id ||= "govuk-notification-banner-title-#{SecureRandom.hex(4)}"
```
we can allow custom ids as well. more importantly, we'd need `labelledby` on [L10](https://github.com/alphagov/govuk_publishing_components/pull/2004/files#diff-81b823e894a91648587dfbb3e5a12dfb1046735370fe56f4286dc39a96b1b39aR10) to use this `id`

owenatgov: So, lemme clarify the 2 asks here. The first is to add an `id` option to the component API. Very doable. The 2nd I'm not quite clear on. Where does the `aria-labelledby` need to go and what should it be? Am I completely misunderstanding?

alex-ju: I could've been more clear. The point of having an `id` on the 'title' element is for [`aria-labelledby` to point to it](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_success_alert.html.erb#L7). If we're going to make this `id` dynamic, we'd need to update the `aria-labelledby` reference to point to this custom `id`.

alex-ju: we may want to call it `title_id` to reflect where it will be applied in the API – we tend to use `id` when the attribute is applied to the top-level element/component-level

owenatgov: Ahaaaaaaa gotcha. Thanks Alex. Will action.

alex-ju (approved):

chris-gds:
Taking a re-look as requested, one thing I did notice as part of [this PR](https://github.com/alphagov/govuk_publishing_components/pull/1929) is a difference in green link style that exists on the [Design System](https://design-system.service.gov.uk/patterns/cookies-page/cookies-updated/index.html) but not on the gem version:

image

owenatgov:
@chris-gds I noticed this too and I'm not really sure how to handle it. The style exists in a class that's to be applied to the links (`govuk-notification-banner__link`) not via any sort of parent class so I can't really apply it to the gem. The way I've handled this is via [this PR](https://github.com/alphagov/frontend/pull/2680) where for now I'm just adding that class in the component call itself.

Going forward I'm not sure what's best. I don't want to mutate links as they're coming in as this could create some really unpredictable risk. I also don't want to replicate the styling myself since that's repeating existing work which could change in future releases. I'm not really sure _why_ they chose to set this link colour to green. I'm gonna run my eye over the docs again and then ask the design system team about it.

PR #2007, opened 2021-04-07
Image card link size option

andysellick:
## What

Adds an option to the image card to change the link size if needed. Uses a copy of the code from the heading helper, put into the shared helper, to do this.

(I'd change the heading component to use this new shared helper code but it's slightly different, I'll come back to it later)

Note that the link size is automatically the correct size for the 'large' version of the image card component, but can be overridden if needed.

## Why
The new GOV.UK roadmap page is going to use this component and it needs the links to be a different size.

## Visual Changes

No changes to the default state of the component, but here it is with the next option applied.

Screenshot 2021-04-07 at 08 47 26

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -63,7 +63,6 @@
 }
 
 .gem-c-image-card__title-link {
-  @extend %govuk-link;
   text-decoration: none;
 
   &:hover {
@@ -135,7 +134,6 @@
   }
 
   .gem-c-image-card__list-item-link {
-    @extend %govuk-link;
     line-height: 1.35em;
   }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -59,7 +59,6 @@
 }
 
 .gem-c-image-card__title {
-  @include govuk-font(19, $weight: bold);
   margin: 0;
 }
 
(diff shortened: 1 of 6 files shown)


lib/govuk_publishing_components/presenters/shared_helper.rb
```diff
@@ -31,6 +31,23 @@ def get_heading_level
         "span"
       end
 
+      def get_heading_size(option, fallback)
+        govuk_class = "govuk-heading-"
+
+        case option
```
owenatgov, line 37: Not critical but you could condense this down

```suggestion
        if ["xl", "l", "m", "s"].include? option
            "#{govuk_class}#{option}"
        else
            "#{govuk_class}#{fallback}"
        end
```

owenatgov (commented):
One little comment about code tidiness. Not critical so happy to negotiate.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
```diff
@@ -59,12 +59,10 @@
 }
 
 .gem-c-image-card__title {
-  @include govuk-font(19, $weight: bold);
```
injms, line 62: This could be left in if you'd like - unlike using `@extend` with GOV.UK Frontend, the mixins are weightless so are included in the `govuk_frontend_support` Sass file.

andysellick: Good to know. The reason I took it out though is it's redundant - the govuk-heading class now gets appended onto the element, which provides all these styles (in fact, if I leave it in, it conflicts with the govuk-heading class, so better to remove it).

lib/govuk_publishing_components/presenters/shared_helper.rb
andysellick: Updated.

owenatgov (approved):
Happy with this 👍

PR #2010, opened 2021-04-09
Extend modules start

andysellick:
## What

Allow cookie consent dependent modules to be initialised immediately when the user consents to cookies, so the page doesn't have to be reloaded first.

- update the cookie banner to call `GOVUK.modules.start()` when cookies are accepted
- extend the `modules.start` function to check the `data-module-started` for a specific value, `delay`
- this allows cookie consent dependent modules to exist on a page and be initialised when the user clicks to accept cookies, without the page needing to be reloaded
- assumes that modules that do this check for cookie consent first before doing anything
- update documentation to explain this
- modify cookie banner test slightly to more accurately simulate running this in a real world environment

See the updated doc file in this PR for more details of how this works in practice.

## Why
We have a new feature on the bank holidays page where users can save their location, but this is done with cookies and so we don't show this feature unless cookies are accepted. Currently if you accept cookies on that page the feature doesn't appear until the page is reloaded.

This feature is an experiment, so it may not be continued, but I suspect that as we move towards a more personalised GOV.UK we will increasingly need the ability to initialise code like this on a page only after cookie consent is given.

## Visual Changes
None.

Trello card: https://trello.com/c/QweAgiPc/693-bank-holidays-experiment-things-to-clean-up

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Extend modules start ([PR #2010](https://github.com/alphagov/govuk_publishing_components/pull/2010))
+
 ## 24.9.0
 
 * Image card link size option ([PR #2007](https://github.com/alphagov/govuk_publishing_components/pull/2007))
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -88,6 +88,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     if (window.GOVUK.globalBarInit) {
       window.GOVUK.globalBarInit.init()
     }
+    window.GOVUK.modules.start()
   }
 
   CookieBanner.prototype.rejectCookieConsent = function () {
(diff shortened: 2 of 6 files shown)


alex-ju (changes requested):
Introducing a new 'delayed' state may not be the right approach for this as it seems brittle and may cause unexpected event chains. I suggest emitting an event when cookies are accepted and setting up an event listener in the script that depends on cookies being accepted.

andysellick:
> Introducing a new 'delayed' state may not be the right approach for this as it seems brittle and may cause unexpected event chains. I suggest emitting an event when cookies are accepted and setting up an event listener in the script that depends on cookies being accepted.

Yeah, I thought this approach might be a little controversial. Your idea is interesting - how would you see that working in practice? What kind of event?

alex-ju:
> > Introducing a new 'delayed' state may not be the right approach for this as it seems brittle and may cause unexpected event chains. I suggest emitting an event when cookies are accepted and setting up an event listener in the script that depends on cookies being accepted.
> 
> Yeah, I thought this approach might be a little controversial. Your idea is interesting - how would you see that working in practice? What kind of event?

If you're looking for a reference, [reorderable-list component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js#L93-L105) does something similar, dispatching a custom event (I would personally prefix them to prevent event clashing) then [another script](https://github.com/alphagov/content-publisher/blob/master/app/assets/javascripts/modules/gtm-reorder-listener.js#L12-L14) picks-up these events and do their stuff.

andysellick:
@alex-ju I've come up with an alternative based on your suggestion - https://github.com/alphagov/govuk_publishing_components/pull/2041

PR #2013, opened 2021-04-13
Remove grid overrides in layout-header 

alex-ju:
## What
Fix layout-header by removing overrides

## Why
The new layout header component is currently only used (via static) in [collections](https://www.gov.uk/transition) and [finder-frontend](https://www.gov.uk/transition-check/questions) – part of the Accounts journey. We currently have an issue with the alignment of this component with the main container and it seems like removing these overrides fixes the issue.

## Visual Changes
<table>
<tr><td valign="top">

before

</td><td valign="top">

after

</td></tr>
<tr><td valign="top">

mobile-before-closed

</td><td valign="top">

mobile-after-closed

</td></tr>
<tr><td valign="top">

mobile-before-open

</td><td valign="top">

mobile-after-open

</td></tr>
</table>


injms (approved):
Good catch - thanks for fixing!

Issue #2018, opened 2021-04-15
Headings in related navigation have incorrectly applied govuk-link class


36degrees:
For example:

```html
<h3 id="related-nav-topics-9e06e2ed" class="gem-c-related-navigation__sub-heading govuk-link" data-track-count="sidebarRelatedItemSection">Explore the topic</h3>
```

This is more obvious in [the review app for #2016](https://govuk-publis-link-test--a7hzmf.herokuapp.com/component-guide/related_navigation), where the headings get an underline and hover state.

Possibly related to the changes in https://github.com/alphagov/govuk_publishing_components/commit/8f5a840f5794ffe7de9c73c0f731a35c0ccdbc51.

PR #2020, opened 2021-04-19
Remove govuk-link class from headings

andysellick:
## What
Remove the `govuk-link` class from headings in the component. This was being inserted because of a recent change to stop using Sass's `extend` feature. Unfortunately in this case the method for generating classes that had been given the `govuk-link` class was being used for both links and headings, where inserting the `govuk-link` class by default wasn't correct.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/2018

## Visual Changes
Fixes a very small visual bug with the headings, returning to the intended visual appearance. See the linked issue for details.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


injms (approved):
Thanks for fixing 👍

PR #2025, opened 2021-04-21
Add data attributes to show password component

andysellick:
## What
Allow data attributes to be passed to the show password component. This basically passes them to the input component, which the show password component wraps (interestingly the ability to pass data attributes to the input component doesn't appear to be documented, but you can do it, leading to a possible discussion about components not only wrapping other components but relying on aspects of their API).

However, the complexity is that I don't want data attributes on the input, I want them on the show/hide button. This is generated by JS, so I've added a bit to the script to check for any data attributes that start with `data-button-` and move them onto the button. It's a bit convoluted, but it works, and I'm not sure how else to do it without hard coding something.

## Why
We want to add tracking to the show/hide button, so we can see how many people are using it. This will be done by wrapping the component in an element with the `gem-track-click` data-module, then passing relevant attributes to the component.

## Visual Changes
None.

Trello card: https://trello.com/c/HRCea3NT/721-add-tracking-to-the-show-password-component

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/show-password.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/show-password.js
@@ -41,6 +41,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.wrapper.insertBefore(this.statusText, this.showHide.nextSibling)
 
     this.showHide.addEventListener('click', this.$module.togglePassword)
+    this.moveDataAttributesToButton()
 
     this.parentForm = this.input.form
     var disableFormSubmitCheck = this.$module.getAttribute('data-disable-form-submit-check')
@@ -63,6 +64,17 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.showHide.setAttribute('aria-label', this.showPasswordFullText)
     this.statusText.innerHTML = this.hiddenPassword
     this.input.setAttribute('type', 'password')
+  }
+
+  ShowPassword.prototype.moveDataAttributesToButton = function () {
+    var attrs = this.input.attributes
+    for (var i = attrs.length; i >= 0; i--) {
+      var attr = attrs[i]
+      if (attr && /^data-button/.test(attr.name)) {
+        this.showHide.setAttribute(attr.name.replace('-button', ''), attr.value)
+        this.input.removeAttribute(attr.name)
+      }
+    }
   }
 
   Modules.ShowPassword = ShowPassword
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Unreleased
 
+* Add data attributes to show password component ([PR #2025](https://github.com/alphagov/govuk_publishing_components/pull/2025))
 * Update documentation about component conventions and visual regression testing ([PR #2015](https://github.com/alphagov/govuk_publishing_components/pull/2015))
 * Remove `@extend` from notice component styles ([PR #2030])(https://github.com/alphagov/govuk_publishing_components/pull/2030)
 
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/show-password.js
```diff
@@ -65,5 +66,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.input.setAttribute('type', 'password')
   }
 
+  ShowPassword.prototype.moveDataAttributesToButton = function () {
+    for (var i = 0; i < this.input.attributes.length; i++) {
```
danacotoran, line 70: I wonder if we can compress this a little bit, feels quite verbose with the repetition of `this.input.attributes`. Could we store that as a variable that we can reference instead?


Additional idea I just had: how about instead of passing attributes for the button in a `data-button-somethingcustom="something"` format, we try using a single `data-button` attribute containing as many `name:value` pairs as needed. So what I'm saying is instead of: 
```
        button_do_this: do_this
        button_do_that: do_that
```
would it be possible to do something like:

```
  data_button:
        do_this: do_this
        do_that: do_that
```

Then instead of looping over all the attributes on the input element, identifying those which start with `data-button-` and processing them accordingly, the script could directly do `this.input. getAttribute('data-button')`, then parse the value which would be an object (probably, haven't actually checked)...

Hopefully that makes some sense, let me know what you think. 😅

danacotoran (changes requested):
The solution does not seem to work when there is more than one `data-button-something` on the input – in the component guide example, only `data-button-do-this` is applied to the button while `data-button-do-that` stays on the input field. 

Additionally, the JS test spec does not check for more than one `data-button-` attribute, if it should be possible to pass more than one data attributes to the button in this manner, could the test be updated to reflect that accordingly?

app/assets/javascripts/govuk_publishing_components/components/show-password.js
andysellick: Good spot on the bug. A weird thing was happening - every time I removed an attribute, the list of attributes would shrink, so iterating to the next one would actually be the next one plus 1. The solution is a bit more verbose, but it works.

I liked the idea of putting the button attributes together but I'm not sure it solves a problem - it's just a different way of doing it. Either way I end up implementing a mechanism for handling two sets of data attributes (one for the input, one for the button) in Ruby or JS or both.

app/assets/javascripts/govuk_publishing_components/components/show-password.js
```diff
@@ -65,5 +66,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.input.setAttribute('type', 'password')
   }
 
+  ShowPassword.prototype.moveDataAttributesToButton = function () {
+    var attrs = []
```
danacotoran, line 70: I would suggest that we can achieve the intended result using only one `for` loop here, by doing something like:

    
    var inputAttrs = this.input.attributes
    for (var j = inputAttrs.length; j >= 0; j--) {
      var attr = inputAttrs[j]
      if (attr && /^data-button/.test(attr.name)) {
        this.showHide.setAttribute(inputAttrs[j].name.replace('-button', ''), inputAttrs[j].value)
        this.input.removeAttribute(inputAttrs[j].name)
      }
    }

danacotoran (commented):
Achieves the intended result, however I have one refactoring suggestion which I think would make this code more terse and easier to read.
Also, just curious, why did you use `test` instead of `match` to determine if the data attribute should be moved to the button?

app/assets/javascripts/govuk_publishing_components/components/show-password.js
andysellick: Ah, of course - going backwards through them means the current index doesn't change when one is deleted. Brilliant, thanks.

danacotoran (approved):
👏🏼

PR #2034, opened 2021-04-27
Swap out Sass `em()` function for `govuk-em()`

injms:
## What

The [`em()` function][1] converts from pixels to em and is identical to the [`govuk-em()` function][2]. Swapping the functions means no change in functionality. The `govuk-em()` function is included in the `govuk_frontend_support` Sass.

Added a `@warn` message to say that the `em()` function could be removed in a future major release.

## Why

Both functions are identical and can be swapped by changing the function name with no difference in output. This reduces the amount of helpers we have to maintain.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/037ce27f3e05973c60d1522bcd113677cc6e7122/app/assets/stylesheets/govuk_publishing_components/components/helpers/_px-to-em.scss
[2]: https://github.com/alphagov/govuk-frontend/blob/c8d2deddc2cf7fa6d116f2af320b322f23856689/package/govuk/tools/_px-to-em.scss

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,8 @@
 
 * Add data attributes to show password component ([PR #2025](https://github.com/alphagov/govuk_publishing_components/pull/2025))
 * Update documentation about component conventions and visual regression testing ([PR #2015](https://github.com/alphagov/govuk_publishing_components/pull/2015))
-* Remove `@extend` from notice component styles ([PR #2030])(https://github.com/alphagov/govuk_publishing_components/pull/2030)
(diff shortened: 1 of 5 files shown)


alex-ju (approved):
🙌

alex-ju:
Not sure what's the best way to keep track of things we want to deprecate. Maybe an issue if you have the time.

injms:
> Not sure what's the best way to keep track of things we want to deprecate. Maybe an issue if you have the time.

Noted this down as an issue - https://github.com/alphagov/govuk_publishing_components/issues/2035 - and wanted to be vague about when this could be removed as we probably don't want to do a major release for just this.

alex-ju (approved):

PR #2036, opened 2021-04-27
Resolve IE11 accordion gem doc duplication

chris-gds:
## What?

Updating to use polyfill version for `forEach` for IE11

## Why?

The Accordion on the gem docs was being visually duplicated.

## Visual Changes

Before
image

After 

Screenshot 2021-04-28 at 09 03 02

## Anything else?

_"`querySelectorAll()` is a method which return a `NodeList`. And on [Internet Explorer](
https://caniuse.com/mdn-api_nodelist_foreach), `foreach()` only works on Array objects. (It works with NodeList with ES6, not supported by IE11)."_ [ref](https://stackoverflow.com/questions/55741747/javascript-foreach-loops-not-working-on-ie11)

[Related PR](https://github.com/alphagov/govuk_publishing_components/pull/1937)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 
 ## Unreleased
 
+* Resolve IE11 accordion gem doc duplication ([PR #2036](https://github.com/alphagov/govuk_publishing_components/pull/2036))
 * Add data attributes to show password component ([PR #2025](https://github.com/alphagov/govuk_publishing_components/pull/2025))
 * Update documentation about component conventions and visual regression testing ([PR #2015](https://github.com/alphagov/govuk_publishing_components/pull/2015))
 * Remove `@extend` from notice component styles ([PR #2030](https://github.com/alphagov/govuk_publishing_components/pull/2030))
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
LOL whoops 😂  Amazing work debugging this Chris. Would it be possible to add some notes in the PR Why going into more detail about why this was happening and why the fix solves it?

maxgds:
> LOL whoops 😂 Amazing work debugging this Chris. Would it be possible to add some notes in the PR Why going into more detail about why this was happening and why the fix solves it?

Yeah, I'm really puzzled. I have been testing this on browserstack and it looks solid as a fix but I'm baffled if that is the fix as to why the accordions on the live site weren't affected but in the component guide they were. Do you have any insight, Chris?

chris-gds:
@owenatgov course - I'll update here and in the PR notes.

_"`querySelectorAll()` is a method which return a `NodeList`. And on [Internet Explorer](
https://caniuse.com/mdn-api_nodelist_foreach), `foreach()` only works on Array objects. (It works with NodeList with ES6, not supported by IE11)."_ [ref](https://stackoverflow.com/questions/55741747/javascript-foreach-loops-not-working-on-ie11)

Essentially the Design System version had a polyfill in place that was ported over, this just needed to be updated to use it. I remember facing a similar issue with this when initial work started (missed spotting this at review for 1937 🙈)

_why the accordions on the live site weren't affected but in the component guide they were._

It's a good question, the issues flags on [this page](https://www.gov.uk/guidance/content-design/content-types) within IE however the resulting duplication doesn't look present, I would need to spend some time investigating my assumption is the differences in what's loaded on live vs the gem docs.

PR #2043, opened 2021-05-04
Update postcode regex for PII stripping

sihugh:
## What

Fixing the postcode regex to not match UUIDs

## Why

The postcode regex that we use often matches UUIDs that are in page URL query strings. These UUIDs are used to show different types of navigation on a page, so are significant for analysis terms.

I've attempted to fix this by adding word boundaries around the regex. Matching a sequence of characters that look like postcode format in a longer string seems a bit eager.

The example provided by the performance analyst was:

https://www.gov.uk/search/services?parent=%2Ftransition&topic=d6c2de5d-ef90-45d1-82d4-5f2438369eea

is being stripped to:

https://www.gov.uk/search/services?parent=/transition&topic=[postcode]5d-ef90-45d1-82d4-5f2438369eea

which is not useful.

https://trello.com/c/fxEriQnM/1489-redact-pii-from-ga-in-a-less-aggressive-manner

## Visual Changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   of the commit log.
 
 
+## Unreleased
+
+* Update postcode regex for PII stripping
+
 ## 24.10.0
 
 * Resolve IE11 accordion gem doc duplication ([PR #2036](https://github.com/alphagov/govuk_publishing_components/pull/2036))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 
 ## Unreleased
 
-* Update postcode regex for PII stripping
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,10 @@
   of the commit log.
 
 
+## Unreleased
+
+* Update postcode regex for PII stripping
```
alex-ju, line 13:
```suggestion
* Update postcode regex for PII stripping ([PR #2043](https://github.com/alphagov/govuk_publishing_components/pull/2043))
```

alex-ju (approved):
Oh, word boundaries were all we needed – I was expecting a regex overhaul when I read the card title 😅 

One nitpick on the changelog entry, otherwise looks great!

PR #2044, opened 2021-05-04
Add legacy `govspeak` class alongside gem govspeak

injms:
## What

Adds the `govspeak` class alongside the `gem-c-govspeak` class - this makes sure that the `govspeak` has the same output as `gem-c-govspeak` for minimal additional CSS.

## Why

There are many many many places where the older `govspeak` class is used on elements that wrap around content, rather than `gem-c-govspeak`. This class comes as part of the content from the content store (eg [in `details.body` in the cabinet office content][1].

Adding the legacy class as a fix is easier and safer that changing the myriad of places where this is added - around [7K places across alphagov][2]).

This is not the ideal fix - that would be to change Whitehall et al to output `<div class="gem-c-govspeak">` -  this will do for now, and will help ensure that there are no unstyled elements whilst this change occurs.

[1]: https://www.gov.uk/api/content/government/organisations/cabinet-office
[2]: https://github.com/search?q=org%3Aalphagov+%27%3Cdiv+class%3D%22govspeak%22%3E%27&type=Code&ref=advsearch&l=&l=

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unpublished
 * Remove unused i18n keys ([PR #2038](https://github.com/alphagov/govuk_publishing_components/pull/2038))
+* Add legacy `govspeak` class alongside gem-c-govspeak ([PR #2044](https://github.com/alphagov/govuk_publishing_components/pull/2044))
 
 ## Unreleased
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_govspeak.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_govspeak.scss
@@ -21,6 +21,7 @@
 @import "govspeak/typography";
 @import "govspeak/warning-callout";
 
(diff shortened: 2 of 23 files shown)


maxgds (changes requested):
Whilst I appreciate the comment "Legacy class name" it doesn't describe to me why it needs to be there. Could we expand on that in the comment, mentioning that it's sometimes contained in content items?

maxgds (changes requested):
Whilst I appreciate the comment "Legacy class name" it doesn't describe to me why it needs to be there. Could we expand on that in the comment, mentioning that it's sometimes contained in content items?

injms:
> Whilst I appreciate the comment "Legacy class name" it doesn't describe to me why it needs to be there. Could we expand on that in the comment, mentioning that it's sometimes contained in content items?

Good point - I've added a (hopefully) better comment explaining why it's unfortunately needed.

maxgds (approved):

PR #2045, opened 2021-05-04
Improve a11y of breadcrumbs

chris-gds:
## What 

Adding `<nav>` element labeled `Breadcrumb` identifies the structure as a breadcrumb trail and makes it a navigation landmark so that it is easy to locate.

`aria-current` applied to the last link in the set to indicate that it represents the current page.

## Why

Improving a11y of breadcrumbs, best practice, related to 1.3.1

_Breadcrumbs are not labelled in any way, making it harder for screen reader users and users who change styles to understand what they are and harder for screen reader users to navigate to them_

## Visual Changes
No visual changes

## Anything else?

https://www.w3.org/TR/wai-aria-practices-1.1/examples/breadcrumb/index.html

Draft PR:
- [ ] CHANGELOG
- [ ] JAWS
- [ ] VoiceOver
- [ ] NVDA
- [ ] Dragon?

### DIFF
--- a/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
@@ -16,17 +16,21 @@
   <%= raw JSON.pretty_generate(breadcrumb_presenter.structured_data) %>
 </script>
 
-<div class="<%= classes.join(" ") %>" data-module="gem-track-click">
+<nav class="<%= classes.join(" ") %>" data-module="gem-track-click" aria-label="Breadcrumb">
   <ol class="govuk-breadcrumbs__list">
     <% breadcrumbs.each_with_index do |crumb, index| %>
       <% breadcrumb = GovukPublishingComponents::Presenters::Breadcrumb.new(crumb, index) %>
         <li class="govuk-breadcrumbs__list-item">
         <% if breadcrumb.is_link? %>
+          <% if index == breadcrumbs.size - 1 %>
+            <% last_link = index ? "page" : "" %>
+          <% end %>
           <%= link_to(
             breadcrumb[:title],
             breadcrumb.path,
             data: breadcrumb.tracking_data(breadcrumbs.length),
             class: "govuk-breadcrumbs__link",
+            "aria-current": last_link,
           ) %>
         <% else %>
           <%= breadcrumb[:title] %>
@@ -34,4 +38,4 @@
         </li>
     <% end %>
   </ol>
-</div>
+</nav>


alex-ju (commented):
Have we discussed the change with the Design System team? As this component is inherited from `govuk-frontend` I suggest we propose the change at source and once agreed with them we roll it out on our library. Otherwise, we risk diverging from source.

edwardhorsford:
This came up recently on the cross gov slack - [link here](https://ukgovernmentdigital.slack.com/archives/C6DMEH5R6/p1619789439348200?thread_ts=1619788745.347800&cid=C6DMEH5R6) whilst it lasts.

In short, the markup in the design system is intentional - I believe to keep noise of extra landmarks down.

Some of the links in the thread:
https://github.com/alphagov/govuk-design-system-backlog/issues/33#issuecomment-480219573
https://github.com/alphagov/govuk_elements/pull/182
https://github.com/alphagov/static/pull/238
https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/

chris-gds:
@edwardhorsford interesting - thank you for all these extra resources & insight (Draft PR's coming in useful) - I'll review.

@alex-ju the suggestion was to action the reverse approach, I'll query again to confirm as I understand this viewpoint

chris-gds:
Option 1 - I can update the pattern on the GOV.UK (within the gem) however this then diverges from the Design System. Which isn’t ideal (and has been flagged.

Option 2 - is contribute to the Design System to allow this to cascade into GOV.UK. Professional opinion can be evidenced however our team does not have access to users to perform user testing so this most likely won’t be accepted as we are unable properly validate work with user research.

This is a bit disappointing as I’m unclear if we can _ever_ contribute properly to the Design System as the burden is on us to action user testing. Which leaves me with:

Option 3. Do nothing.
Accept this type of work isn’t a good use of our time as we aren’t structured correctly as an organisation and to focus on other priorities.

Currently asking opinions from others.

chris-gds:
Updating with some learnings:

I believe Gov are redefining best practice for this particular use of this element based on previous research and testings.

-  `<nav>` - there is history and compelling evidence [not to use this](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/) 

- `aria-label="Breadcrumb"` this has been [removed intentionally](https://github.com/alphagov/static/pull/737) as it's [not a valid label.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles) 
Suggested `aria-labelledby` could be a good solution in this context.

-  The `aria-current` enhancement seems to be a welcomed improvement.

Contributing directly is under discussion to understand if this pathway is possible.

36degrees:
> `<nav>` - there is history and compelling evidence [not to use this](https://insidegovuk.blog.gov.uk/2013/07/03/rethinking-navigation/)

Worth noting that research is from 2013 – it was definitely a conscious choice not to mark the breadcrumbs up as a navigation landmark at the time, but we shouldn't assume that it still holds true today. However, we should probably avoid changing it unless we know from research (either our own, or someone else's) that it's helpful to users.

> `aria-label="Breadcrumb"` this has been [removed intentionally](https://github.com/alphagov/static/pull/737) as it's [not a valid label.](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles)

These are two different things – the breadcrumb **role** was removed, because it's not a valid role. That's not the same thing as "breadcrumb" being the label for the landmark. If the breadcrumbs are marked up as a navigation landmark, that landmark absolutely should have a label. If they are not, they do not need one.

Worth noting that adding an `aria-label` does introduce a little extra complexity in that "Breadcrumbs" will need to be localised into different languages (and, as `aria-label` is not visible on the screen, that's something that could be easily missed).

chris-gds:
👋🏼  @36degrees 

> Worth noting that research is from 2013 – it was definitely a conscious choice not to mark the breadcrumbs up as a navigation landmark at the time, but we shouldn't assume that it still holds true today. However, we should probably avoid changing it unless we know from research (either our own, or someone else's) that it's helpful to users.

Agreed, it's the most recent understanding so while we shouldn't update we should still validate.

> These are two different things – the breadcrumb role was removed, because it's not a valid role.  That's not the same thing as "breadcrumb" being the label for the landmark. 

You're right, I've blurred these two thinking they're referring to the same thing 🙈 (I'll revise my earlier comment).

> If the breadcrumbs are marked up as a navigation landmark, that landmark absolutely should have a label. If they are not, they do not need one.

Are you saying that you don't think _any_ label is needed this context? `aria-labelledby` or otherwise (keep as is)?
([Noticed some other comments here](https://github.com/alphagov/govuk_elements/issues/412#issuecomment-284025323))

> Worth noting that adding an aria-label does introduce a little extra complexity in that "Breadcrumbs" will need to be localised into different languages (and, as aria-label is not visible on the screen, that's something that could be easily missed).

🤔  The localisation aspect definitely widens the scope of this and is worth highlighting. I'm aware there is not a consistent localisation pattern on GOV.UK (but localisation solutions do exists) however I'm unclear if this has been resolved consistently within the Design System.

app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
```diff
@@ -16,22 +16,26 @@
...
         <% if breadcrumb.is_link? %>
+          <% if index == breadcrumbs.size - 1 %>
+            <% last_link = index ? "page" : "" %>
+          <% end %>
           <%= link_to(
             breadcrumb[:title],
             breadcrumb.path,
             data: breadcrumb.tracking_data(breadcrumbs.length),
             class: "govuk-breadcrumbs__link",
+            "aria-current": last_link,
```
maxgds, line 33: Just a note here in case this PR does wind up getting carried forward: this is not quite right for our implementation of breadcrumbs. We don't show the current page in the list, as per the design system. The design system can add a current page (though it's not covered as an example in their docs) but it's not implemented here, we only show parent items. When the current page does exist in Design System breadcrumbs it's a list item but not a link.

selfthinker:
One other solution to consider is adding a (hidden) heading before the breadcrumbs, so that screen reader users who navigate by heading can easily find it and understand what it is.

It is currently only visually clear that they are breadcrumbs, I would argue those visuals are missing from the markup as it's "just" a list of links. The visuals that identify them as breadcrumbs are the chevrons and the positioning. Just imagine we'd use standard ordered list styling there instead of breadcrumb styling and you'd see how it's instantly less recognisable.

I think we should do **something** to programmatically at least give a hint as to what this list of links is. There are multiple ways to do this, putting it in a `nav` is just one of them, adding a hidden heading is another, but there will be more.

jon-kirwan:
> One other solution to consider is adding a (hidden) heading before the breadcrumbs, so that screen reader users who navigate by heading can easily find it and understand what it is.
> 
> It is currently only visually clear that they are breadcrumbs, I would argue those visuals are missing from the markup as it's "just" a list of links. The visuals that identify them as breadcrumbs are the chevrons and the positioning. Just imagine we'd use standard ordered list styling there instead of breadcrumb styling and you'd see how it's instantly less recognisable.
> 
> I think we should do **something** to programmatically at least give a hint as to what this list of links is. There are multiple ways to do this, putting it in a `nav` is just one of them, adding a hidden heading is another, but there will be more.
> 
> A heading or other form of label doesn't need to be hidden. Just starting the breadcrumbs with the visible words "You are here:" (or similar) would make it easier to understand for other groups of people as well.

Thanks for adding the suggestions @selfthinker. I'm now closing the pull request but have opened a new one [here](https://github.com/alphagov/govuk-frontend/issues/4451) and have included references to your comments in the description

PR #2046, opened 2021-05-05
Update design of metadata component

owenatgov:
## What
Update's the design of the [metadata block component](https://components.publishing.service.gov.uk/component-guide/metadata) along with some general tidying. The specifics:

- Restyles the data term and definition on desktop so that there is no longer a gap between the two
- Removes the colon next to the "published" and "updated at" terms
- Creates translation keys which previously didn't exist
- Give the associated files a bit of a spring clean

## Why
This design existed in government frontend but was [accidentally overwritten](https://github.com/alphagov/government-frontend/pull/2026) by work by govuk accessibility during our component update work. This change benefits screen magnification users as it will be easier for them to reach information when zoomed in.

A more detailed note from our designer on why it was changed in the first place and why we want to re-update it and apply it everywhere:

>The change was made originally because the two column layout was inconsistent across pages — each instance of the metadata component had its own spacing between the columns that was dependent on the length of the content in the first column.
We decided this looked messy, didn't serve the user from page to page (elements jumping around isn't great for accessibility and predicting where content is).
Additionally, the gap between the columns would be baffling to screen magnification users.
This was done, if I recall correctly, as part of the template consolidation work (or an outgrowth of it). It was also done as part of a wider piece of work to tackle the two column layout used for contents lists on Mainstream Guides. Everything was put into a single column to improve usability and accessibility.

## Visual Changes
| Before | After |
| --- | --- |
| Screenshot 2021-05-05 at 10 32 22 | Screenshot 2021-05-05 at 15 13 00 |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
@@ -7,6 +7,11 @@
   a {
     @include govuk-link-common;
     @include govuk-link-style-default;
+    font-weight: bold;
+  }
+
+  .gem-c-metadata__definition-link {
+    font-weight: normal;
   }
 }
 
@@ -29,17 +34,19 @@
   }
 }
 
+.gem-c-metadata__term,
+.gem-c-metadata__definition {
+  line-height: 1.4;
+}
+
 .gem-c-metadata__term {
   margin-top: .5em;
-  line-height: 1.4;
 
   @include govuk-media-query($from: tablet) {
     box-sizing: border-box;
     float: left;
     clear: left;
-    width: 30%;
-    max-width: 11em;
-    padding-right: govuk-spacing(2);
+    padding-right: govuk-spacing(1);
     margin-top: 0;
   }
 }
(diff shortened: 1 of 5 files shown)


nnagewad:
Looks awesome. 👍🏽

maxgds:
I'm concerned about hiding the commas. In my terrible wrangling of VoiceOver I'm not seeing any difference between live and this, but I wonder if it might make things confusing for some. My concern is that rather than hearing "department of work and pensions comma home office comma hmrc" might you wind up with a word soup where the user has to try to work out where the delimiters are and getting it wrong? "department of work" and "pension home" "office".

Blogs posts like this one tend to say leave things be if they sound a bit clunky as trying to fix them might make the issue worse: https://accessibility.blog.gov.uk/2017/02/08/advice-for-creating-content-that-works-well-with-screen-readers/
Extract:
So the best answer seems to be: don't write content that works specifically for screen readers, write content that works well for everyone. Use correct punctuation, spelling and grammar, use standard conventions for acronyms and abbreviations, and use words that are appropriate for your audience.

Maybe one to take to the capability team?

owenatgov:
@maxgds The reason I did this originally was on voice over, the dictation was reading each link and each comma separately. See video's for example of the problem vs the solution in VO (excuse messy recording):

https://user-images.githubusercontent.com/64783893/117165859-c4604600-adbd-11eb-9ab7-f7050c1ff3cc.mov

https://user-images.githubusercontent.com/64783893/117165888-caeebd80-adbd-11eb-8b7e-09603754fb7d.mov

However, testing more extensively in other screen readers it does read stuff out as one continuous sentence which is confusing:

https://user-images.githubusercontent.com/64783893/117166068-f40f4e00-adbd-11eb-8dea-5f34b5c8bdeb.mov

My gut feeling from extensive testing and your reasoning is that even though this problem persists on VO it isn't a problem in NVDA so maybe I should just revert it? What do you think?

owenatgov:
@maxgds Not to worry, makes sense to me. Removed and updated the PR description

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -1,3 +1,5 @@
+$space-between-data: govuk-spacing(1);
```
maxgds, line 1: I'm not sure what value is added by defining this when it's only used a couple of times. I'm not going to block merging based on it though.

maxgds (approved):

owenatgov:
I'm hold off on merging this for a bit because there's a separate piece of work to inform content designers about this change.

PR #2050, opened 2021-05-06
Fix travel advice pages to use parent breadcrumbs

sihugh:
Travel advice pages used to use parent breadcrumbs, but when we tried to simplify it, we missed this rule.


sihugh:
Just pushed the correct changelog entry 🙄 .  Thanks @alex-ju

Issue #1953, opened 2021-03-02
Clean up redundancy within radio component


danacotoran:
Recently the ability to pass a custom heading level was introduced to the radio component https://github.com/alphagov/govuk_publishing_components/pull/1951 

Since we've now got this option, the `is_page_heading` option is a bit redundant/confusing. 
The `is_page_heading` parameter simply dictates that the heading be a `h1` of size `xl`. 
It is possible to retire `is_page_heading` and use `heading_level` and `heading_size` to achieve the same result.

PR #1637, opened 2020-08-05
Add space below radio component when page heading

injms:
## What

Adds extra margin between the question and the first item - which could be a description, hint text, or the radio input - if the component is being used as a page heading.

## Why

To give the heading more space, as requested by @miaallers.

## Visual Changes

| Before | After|
|-|-|
| image | image |
| image | image |

PR #2051, opened 2021-05-06
Retire is_page_heading parameter (part 1)

danacotoran:
Better description: _coming soon!_

------
## What

This is step 1 towards removing the `is_page_heading` parameter from the radio component. 

## Why

Due to a need for more flexibility, the ability to pass a custom heading level was introduced to the radio component some time back #1951
The `is_page_heading` parameter is a specialised flag that dictates that the heading should be a `h1` of size `xl`.
This used to solve a problem when the component did not use to support `heading_level`, however now that we can use `heading_level` and `heading_size` to achieve the same result, the `is_page_heading` parameter becomes redundant.

The reason we have to do this in steps is that the component is referenced with the `is_page_heading` parameter in several applications.
Due to the way the logic in this component is structured, it would be a breaking change if we simply went ahead with replacing `is_page_heading: true` with `heading_level: 1` and `heading_size: 'xl'`

## Visual Changes
N/A

------

Related issue: #1953


app/assets/stylesheets/govuk_publishing_components/components/_radio.scss
```diff
@@ -1,5 +1,6 @@
 @import "govuk/components/radios/radios";
 
-.gem-c-radio__heading-text {
+.gem-c-radio__heading-text,
+.govuk-fieldset__legend h1 {
```
alex-ju, line 4: I'm uneasy whenever I see styles applied to tags. If we really need this gem override on the `govuk-fieldset__heading` can we instead make sure `gem-c-radio__heading-text` is added to the heading (regardless of the level)?

alex-ju: Looking at [the original implementation](https://design-system.service.gov.uk/components/radios/) we don't seem to need this custom margin.

danacotoran: To my eyes, I would agree that the extra margin isn't needed there @alex-ju but it seems like it was added as requested by one of our designers  https://github.com/alphagov/govuk_publishing_components/pull/1637 
I think it was added to improve readability.

I'm going to think about a better way of doing it, you're right, probably shouldn't apply the styles to the tag.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix travel advice pages to use parent breadcrumbs ([PR #2050](https://github.com/alphagov/govuk_publishing_components/pull/2050))
+* Retire is_page_heading parameter (step 1) ([PR #2051](https://github.com/alphagov/govuk_publishing_components/pull/2051))
```
alex-ju, line 13: The actual removal of the `is_page_heading` feature will be part of a breaking change release, so I'd rather rephrase this to what it currently does (e.g. Refactor heading logic in radio component)

alex-ju (approved):
Looks good! One suggestion for the changelog and I'm not sure we earn much by removing the vowels in `heading_lvl`, but since this is not exposed in the component API, I'll let you decide on a final name.

CHANGELOG.md
danacotoran: Thanks! 🙏

PR #2059, opened 2021-05-11
Adding ID to list item

chris-gds:
## What?

Extending the list component to allow a user to pass an ID to a list item, should no custom ID be provide one will be generated.

## Why?

Whilst [converting](https://github.com/alphagov/whitehall/pull/6091) an element to use the [list component](http://localhost:3212/component-guide/list) (on Whitehall) this resulted in a testing error due to a missing ID. Current iteration of this gem component does not provide an option to generate an ID or provide a custom one. 

This extension aims to resolve this and provide these options for future list component use.

## Visual Changes
None however screenshots for Context attached

Screenshot 2021-05-11 at 13 54 35

Screenshot 2021-05-11 at 13 54 45

## Anything else?

Raising a draft for feedback if _should_ do this.

### DIFF
--- a/app/views/govuk_publishing_components/components/_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_list.html.erb
@@ -2,6 +2,8 @@
   aria_label ||= nil
   extra_spacing ||= nil
   id ||= nil
+  id_li ||= nil
+  list_id ||= false
   items ||= []
   list_type ||= "unordered"
   visible_counters ||= nil
@@ -10,6 +12,9 @@
   classes << "govuk-list--bullet" if visible_counters && list_type === "unordered"
   classes << "govuk-list--number" if visible_counters && list_type === "number"
   classes << "govuk-list--spaced" if extra_spacing
+
+  # Default list ID is "list-item-[]" unless a custom ID is defined.
+  list_id = list_id & id_li.present? ? id_li : "list-item-#{SecureRandom.hex(4)}"
 
   # Default list type is unordered list.
   list_tag = "ul"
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_list.html.erb
```diff
@@ -20,7 +25,9 @@
 <% if items.any? %>
   <%= content_tag list_tag, class: classes, id: id, "aria-label": aria_label do %>
     <% items.each do |item| %>
-      <li><%= raw(item) %></li>
+      <%= tag.li(id: list_id) do %>
+        <%= raw(item) %>
+      <% end %>
```
injms, line 30: Am I reading this right - would this put the same `id` on each of the list items in the list?

injms: Just checked, and the same `id` is added to different list elements.

injms (changes requested):
There shouldn't be multiple `id`s that are the same on the page - each list item needs a unique `id` attribute, or no `id` attribute.

chris-gds:
> There shouldn't be multiple `id`s that are the same on the page - each list item needs a unique `id` attribute, or no `id` attribute.

Agreed. In this context this PR needs to be adjusted for a unique ID attribute.  
However as this requirement is currently only required in Whitehall (I've not seen other apps need this feature) currently investigating why a unique ID is being used and if it can be removed to avoid this need for this update.

andysellick:
@maxgds this PR's been sitting around for a while - is this feature still a requirement or can this PR be closed?

maxgds:
I think we found a workaround for this - certainly this work isn't on our radar at the moment. I'll close this but bear in in mind if the issue resurfaces.

PR #2060, opened 2021-05-12
Rewrite README to be consistent with other repos

benthorner:
https://trello.com/c/Psf7EzPB/204-move-away-from-startupsh

This follows the pattern established elsewhere [1][2], but with
some deviations to account for this being a library, as well as
a runnable application. Please see the commits for more details.

[1]: https://github.com/alphagov/collections/pull/2362
[2]: https://github.com/alphagov/collections/pull/2375


README.md
```diff
@@ -46,22 +45,11 @@ GOV.UK Publishing Components shares the same standards in terms of browser and a
...
-### Install dependencies
+You can use the [GOV.UK Docker environment](https://github.com/alphagov/govuk-docker) or the local `startup.sh` script to run the app. Read the [guidance on local frontend development](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) to find out more about each approach, before you get started.
 
-As well as the Ruby, you will also need the [Yarn package manager](https://classic.yarnpkg.com/en/docs/install/) to allow installing [npm](https://npmjs.com) dependencies.
-
-Install the application's dependencies with:
-
-```sh
-bundle install
-yarn install
```
andysellick, line 60: Has this bit been deleted entirely or moved somewhere else? Isn't it important?

README.md
```diff
@@ -80,16 +68,6 @@ bin/rake app:jasmine
...
 
-### Viewing the component guide
-
-View the component guide in your browser with:
-
-```sh
-./startup.sh
-```
-
-The guide should be available at http://localhost:3212
```
andysellick, line 92: Isn't it helpful to know how to access the locally running instance of this?

docs/run-component-guide.md
```diff
@@ -1,26 +0,0 @@
...
-
-```
-cd ~/govuk/govuk-docker
-make govuk_publishing_components
-
-cd ~/govuk/govuk_publishing_components
-govuk-docker-up
-```
-
-Then visit [http://govuk-publishing-components.dev.gov.uk/component-guide/](http://govuk-publishing-components.dev.gov.uk/component-guide/).
```
andysellick, line 26: Is this URL listed somewhere else? If not I'm not sure someone new to this would be able to figure it out - or is it listed in the output of the command?

README.md
benthorner: This is covered by GOV.UK Docker. Since a lot of frontend repos require both of these commands, we should list them in [the guidance for local frontend dev](https://docs.publishing.service.gov.uk/manual/local-frontend-development.html) if we think they're not obvious (when following the non-GOV.UK Docker path). Since they're not listed there right now, my assumption was that people either use GOV.UK Docker, or know enough about frontend dev to run them by themselves (but I might be wrong there).

README.md
benthorner: This is printed out when you run the script, so I think it's OK to remove it - this matches the approach we've taken in other frontend repos.

docs/run-component-guide.md
benthorner: Good question, I was being a bit blasé here. This is actually the root URL i.e. going to `govuk-publishing-components.dev.gov.uk` redirects to `/component-guide`. So we don't need to mention it explicitly.

README.md
alex-ju: I see no reference of yarn in the mentioned documentation, hence there's a great value in having them here, especially for a new started who may be less familiar with it.

README.md
```diff
@@ -1,103 +1,47 @@
...
-```
-Javascript is tested using Jasmine and the [Jasmine gem](https://github.com/pivotal/jasmine-gem). Tests can be run either in the browser or on the command line via the dummy app’s tasks:
 
-```sh
-# browser
+# run JS tests in browser
 bin/rake app:jasmine
 
-# command line
+# run JS tests on command line
```
alex-ju, line 35: We seem to phrase these rather inconsistently in various applications (for example in `collections`, used as a reference for this PR we have 'To run JavaScript tests (only):'

README.md
```diff
@@ -1,103 +1,47 @@
...
-There are 2 types of helper classes in this app:
-
-- [AppHelpers](lib/govuk_publishing_components/app_helpers). Are exposed to the applications using this gem. They should be documented using RDoc.
-- [Component Presenters](lib/govuk_publishing_components/presenters). Anything in these classes is only for use within the components. They should be marked `@private`.
+Components should be added to this gem if they are required in more than one application, otherwise they should be added to that application. All components should meet the same standards of browser and assistive technology support described in [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support).
 
-## Relationship with GOV.UK Design System
+- 👉 [Install and use this gem](docs/install-and-use.md)
+- 👉 [Use a component in your application](docs/use-components.md)
+- 👉 [Develop a component](docs/develop-component.md)
```
alex-ju, line 18: What is the reasoning for splitting the documentation links into two?

alex-ju (changes requested):
While I have a great appreciation for consistency, I have concerns about most of these changes as they can make it harder for newcomers to understand and make the best of this repo. Can you share with us any anecdotal evidence (ideally some guerrilla research) used when shaping up this new format for the README files? Have we involved/consulted any technical writer while doing so?

README.md
owenatgov: I personally can't think of a reason why it would be beneficial to remove this line over making it crystal clear where to go to get the component guide.

benthorner: Hey @owenatgov, thanks for taking a look. I'm a bit confused: can you explain what you mean a bit more? What scenario are you thinking about in your comment?

owenatgov: No worries! So a lot of this is a bit conjecture-y but: new devs learning to use our tooling will arrive at the docs first if they get confused. The script in question in this case does make it clear where to go but terminal environments aren't the most usable places and what'll probably end up happening is that any devs who miss that line in the terminal will ask where to go, realise through some discussion that it's in the terminal window and feel silly. We don't really want to be perpetuating imposter syndrome (even in a small way)

I feel this way about quite a few changes in this PR which Andy and Alex have already outlined. I understand why you've made them but similarly, in my opinion, consistency and "tidiness" gain us very little in terms of documenting our process over spelling things out clearly to devs. There is zero harm (again IMO) in making our docs "dumb".

benthorner: Thanks for explaining @owenatgov, that makes sense to me. Even if someone does see the output, it's also possible they may not understand what it means / whether it's related to their goal.

I do think we need to be careful with putting too much "stuff" in our READMEs:

- Reading a long document is off-putting, and can equally contribute to imposter syndrome, if someone feels they have a mountain to learn before they can do anything. One line is obviously fine, and this one is also relevant to the main goal of a README, which is to provide the minimum info to get started with the repo (vs. doing everything ever).

- Maintaining a long document is also off-putting. It's easier for false information to persist this way. In general, we avoid reviewing READMEs on GOV.UK, and most of them contained out-of-date or misleading info as a consequence. So we should think hard about what's necessary, and whether we can centralise the knowledge.

Although this thread is about one line, it's actually the whole section that I'm suggesting we delete. The reason for this is that the content is misleading, because it only applies to one way of running this repo: if you're using GOV.UK Docker, these instructions won't work. What I've done instead is to point at [the central dev doc about this](https://github.com/alphagov/govuk_publishing_components/pull/2060/commits/aa7860921635ddb7233b9af8abea6b5842eba66b#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R50):

- The doc explains both alternatives for running this and other frontend repos.
- The doc also covers how to view the index page, which for this app, is the component guide.

We [agreed this approach for Collections](https://github.com/alphagov/collections/pull/2375), with the aim of rolling it out across other frontend repos. I went to some effort to ask frontend devs to give their thoughts at that stage, so I'm surprised to be encountering comments now. Equally I want to make sure we're making a change we can all agree with. 

That was quite a long comment, sorry! I've been pondering how best to proceed with this PR, so it's more a collection of my thoughts. Some questions I'd like you to consider are:

- Is the central doc on frontend development sufficient to replace this section?
- If it's not, what do we need to have here? Do we need this in the other frontend repos as well?
- How can we support people who don't use the startup.sh scripts at the same time?

owenatgov: You make some compelling point here. I one little reflection here which you might find beneficial:

>We agreed this approach for Collections, with the aim of rolling it out across other frontend repos. I went to some effort to ask frontend devs to give their thoughts at that stage, so I'm surprised to be encountering comments now.

A hypothesis about this specifically is that I had no idea you could run this repo using govuk-docker. In my mind, and I think the minds of a lot of other FEs on govuk, this was a separate because it's a gem and the docs element lives in a different world to our frontend apps. It sort of does but also that's no reason not to use govuk-docker really. I therefore compartmentalise it differently and made the (incorrect) assumption that this was the only way to run it. This probably speaks to your point that these docs are in fact misleading.

I saw you asking for FE input re: dockerising everything and didn't get involved as I was already using govuk-docker for everything... except this.

As for the original comment, I frankly don't know enough about how rails does serving to make a call here eg: I'm not sure if the rest of our apps, if run using the `startup.sh` method, would use `localhost:3212` or something else. I think I'd defer to @alex-ju's comments on this one ie: if they've been agreed and there's been some sort of investigation then I retract my comment.

benthorner (commented):
Me and @alex-ju have been through the diff for this PR and agreed on a few changes to make before it goes in, as well as areas where the original change is OK - I'll resolve comments for those shortly.

Thanks again for your time @alex-ju ⭐.

README.md
benthorner: I've removed the original commit and replaced it with a new one to do this: 727ec0cdac4b8bd1b717564b340b7688fb3f3a59

docs/develop-component.md
```diff
@@ -10,6 +10,38 @@ A component must:
 * [meet the component principles](/docs/component_principles.md)
 * [follow component conventions](/docs/component_conventions.md)
 
+## Generate a new component
```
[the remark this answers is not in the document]
benthorner: Dropped this commit and replaced with: eaf64195dae6906cc2eac10a970b2b22aa07be57

README.md
```diff
@@ -13,19 +13,9 @@ A Ruby gem to document and distribute components for GOV.UK applications. It pro
 
 Components should be added to this gem if they are required in more than one application, otherwise they should be added to that application. All components should meet the same standards of browser and assistive technology support described in [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend#browser-and-assistive-technology-support).
 
-## How to
-
-- [Install and use this gem](docs/install-and-use.md)
-- [Use a component in your application](docs/use-components.md)
-- [Generate a new component](docs/generate-a-new-component.md)
-- [Develop a component](docs/develop-component.md)
```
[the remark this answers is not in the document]
benthorner: Done in this commit: eaf64195dae6906cc2eac10a970b2b22aa07be57

README.md
benthorner: I've had a look to see which READMEs we have this in, which are just: [content-publisher](https://github.com/alphagov/content-publisher#running-the-test-suite), [collections](https://github.com/alphagov/collections#running-the-test-suite) and [whitehall](https://github.com/alphagov/whitehall#running-the-test-suite). I know there are other apps with JS tests, but it's beyond the scope of the work we're doing to fix that. As long as the README is at least correct, someone else can add guidance if they feel the need in future.

Back to your point here, the more popular form is text + code (vs. comment + code), so I've added another commit to switch to that form (2784152f2c7c5e7b09a5b9c51092e10c6a93d60f), as well as a PR to make content-publisher consistent (https://github.com/alphagov/content-publisher/pull/2355), if you'd like to review that as well.

There's actually a bigger problem here, which is that the browser usage of Jasmine doesn't work in GOV.UK Docker. Luckily, @injms has got a PR to try and address that (https://github.com/alphagov/govuk-docker/pull/498) for this repo, so I don't think it needs to block this change. (We'll want to do a similar thing for other apps, too.)

benthorner:
@alex-ju I think I've now addressed all the comments.

Would you like to take another look?

README.md
```diff
@@ -1,103 +1,52 @@
...
+
+- [Install and use this gem](docs/install-and-use.md)
+- [Use a component in your application](docs/use-components.md)
+- [Generate a new component](docs/generate-a-new-component.md)
+- [Testing a component](docs/testing-components.md)
+- [Move a component from an application to the gem](docs/moving-components-upstream-into-this-gem.md)
+- [Publish/release a new version of the gem](docs/publishing-to-rubygems.md)
+- [Keep this gem in sync with the Design System](docs/upgrade-govuk-frontend.md)
+- [Code documentation on rubydoc.info](http://www.rubydoc.info/gems/govuk_publishing_components)
+* [Component conventions](docs/component_conventions.md)
```
alex-ju, line 48:
```suggestion
- [Component conventions](docs/component_conventions.md)
```

Annoyingly, in GFM if we use a different symbol for lists (in this case `*` instead of `-`) the list will be split into two.

README.md
```diff
@@ -1,103 +1,52 @@
...
+* [Component principles](docs/component_principles.md)
```
alex-ju, line 49:
```suggestion
- [Component principles](docs/component_principles.md)
```

alex-ju (approved):
Looks good! Thanks for taking the time to work on this!
One small issue with list markers, otherwise good to go!

benthorner:
Oops, thanks for spotting @alex-ju. I've amended the second last commit to use `-`.

PR #2061, opened 2021-05-12
Remove `is_page_heading` parameter from radio component

danacotoran:
# What

This removes all mentions of the `is_page_heading` option from the radio component.

# Why
Recently, the ability to pass a custom heading level to the radio component was introduced: https://github.com/alphagov/govuk_publishing_components/pull/1951 

Since specifying a custom heading level is now an option, the `is_page_heading` parameter became redundant. This parameter simply dictates that the radio component legend heading should be a `h1` of size `xl` by default.
It is possible to achieve the same result using `heading_level` and/or `heading_size` instead.

Whenever we need this component to have a h1 in the future, we can simply use the `heading_level` option (in combination with `heading_size`, if necessary).

This issue is recorded here https://github.com/alphagov/govuk_publishing_components/issues/1953 

# Visual changes
There should be no visual changes as a result of this update, as work has already been done to update all references of the radio component to use `heading_level`/`heading_size` instead of `is_page_heading: true` –  for instance [here](https://github.com/alphagov/finder-frontend/pull/2503/files)

There are still some references to the `is_page_heading` option in some older, read-only repositories which are using older versions of the gem ([Example](https://github.com/alphagov/govuk-search-relevance-tool/blob/9874bfbfbd39ee2a4b7de8b64a436ffd0cde5098/app/views/brexit_checker/_question_single_choice.html.erb#L7)). In the hypothetical situation where one of these repos is unarchived and updated to the latest version of `govuk_publishing_components`, this would be a breaking change in those repos.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,8 +10,9 @@
 ## Unreleased
 
 * BREAKING: Remove ability to link contextual text on titles ([PR #2192](https://github.com/alphagov/govuk_publishing_components/pull/2192))
-
 * Delete empty print stylesheets ([PR #2225](https://github.com/alphagov/govuk_publishing_components/pull/2225))
+* BREAKING: Remove is_page_heading parameter from radio component ([PR #2061](https://github.com/alphagov/govuk_publishing_components/pull/2061))
+
 
 ## 24.21.1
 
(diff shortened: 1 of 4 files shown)


owenatgov:
Nitpick before I dive into review: I can see that you've been dashing around our codebase removing instances of this _buuuuut_ should this be labelled as a breaking change as we're mutating our API?

owenatgov (approved):
Besides my nitpick this looks perfectly fine to me ✨

danacotoran:
@owenatgov good point Owen. 
I prefixed the changelog entry with "BREAKING" to highlight that this is a breaking change as per our conventions.

PR #2062, opened 2021-05-12
Remove jQuery from the feedback component

andysellick:
## What
Remove jQuery code from the feedback component's JavaScript, and replace with vanilla JS.

There are two things that might be a problem with this PR:

**Ajax**

Firstly, this component used jQuery's ajax functionality. We've only got a few scripts that do this, and there are a few options for what to replace it with (`XMLHttpRequest`, [`fetch`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch)). We should have a consistent approach for this.

I've chosen `XMLHttpRequest` because I'm more familiar with it, but `fetch` might be better. I don't want this conversation to block this PR - I think it's more important we remove jQuery, and if we have to, we can come back later and fix up the consistency (as I said, it's only a few scripts that do this).

**Internet Explorer**

Doing ajax requests and form serialisation is really hard in IE, particularly old IE but also even IE11. After going through numerous hoops and writing a lot of extra code I've finally come to the decision that the simplest option is to wrap the form submission JS in a check that excludes IE from using it. 

This means that IE users will still get the normal functionality of the feedback component, but when they submit a response to either of the two forms it will not be handled by JS and be passed as a normal form submission to the backend (which is what happens now if you have JS disabled in your browser).

Consequently, I've wrapped the tests that check form submission in the same if statement that the component code uses to check for IE11, as there's no point running those tests on IE11 (they immediately try to do a proper form submission and break the test suite).

I accept that this might be a bit more controversial so happy to discuss.

## Why
We're trying to remove our dependency on jQuery.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -1,5 +1,4 @@
-/* eslint-env jquery */
-
+/* global XMLHttpRequest, FormData */
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -11,53 +10,56 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
       setInitialAriaAttributes()
       setHiddenValues()
 
-      this.$toggleForms.on('click', function (e) {
-        e.preventDefault()
-        toggleForm($(e.target).attr('aria-controls'))
-        trackEvent(getTrackEventParams($(this)))
-        updateAriaAttributes($(this))
-      })
+      for (var j = 0; j < this.toggleForms.length; j++) {
```
kr8n3r, line 31: can you instead attach one click event listener and check for target element?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -11,53 +10,56 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-      this.$surveyForm = $(this.surveyForm)
-      this.$surveyWrapper = $element.find('#survey-wrapper')
+      this.pageIsUsefulButton = this.element.querySelector('.js-page-is-useful')
+      this.pageIsNotUsefulButton = this.element.querySelector('.js-page-is-not-useful')
+      this.somethingIsWrongButton = this.element.querySelector('.js-something-is-wrong')
+      this.promptQuestions = this.element.querySelectorAll('.js-prompt-questions')
+      this.promptSuccessMessage = this.element.querySelector('.js-prompt-success')
+      this.surveyWrapper = this.element.querySelector('#survey-wrapper')
 
       var that = this
```
injms, line 25: I wonder if the `that` variable could be better named to indicate that it's the `this` for the whole module?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -68,45 +70,71 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-          setInitialAriaAttributes()
-          that.$activeForm.toggleClass(jshiddenClass)
-        }).fail(function (xhr) {
-          showError(xhr)
-          enableSubmitFormButton($form)
-        })
-      })
+      // much of the JS needed to support sending the form contents via this script is
+      // unsupported by IE, even IE11. This check causes IE to not intercept form submits
+      // and let them happen normally, which is handled already by the backend
```
injms, line 75: I think that this is a good approach to let IE11 fall back to the non-JavaScript version of the component.

It's good to document this in the code - and it'd be good to think about how to let the component guide show the browser support for each component, much like the accessibility criteria.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -68,45 +70,71 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        that.somethingIsWrongForm.appendChild(javascriptEnabled)
+
+        var referrer = document.createElement('input')
+        referrer.setAttribute('type', 'hidden')
+        referrer.setAttribute('name', 'referrer')
+        referrer.setAttribute('value', document.referrer || 'unknown')
+        that.somethingIsWrongForm.appendChild(referrer)
         that.somethingIsWrongForm.invalidInfoError = [
           '<h2>',
           '  Sorry, we’re unable to send your message as you haven’t given us any information.',
```
injms, line 140: Could the double space be removed?

```suggestion
          'Sorry, we’re unable to send your message as you haven’t given us any information.',
```

app/assets/javascripts/govuk_publishing_components/components/feedback.js
andysellick: I think this is a convention we use elsewhere, but do you have a suggestion?

app/assets/javascripts/govuk_publishing_components/components/feedback.js
andysellick: That is a good idea. I think doing that properly is outside the scope of this PR, but I'll add a note to the YML file to make this clear.

injms: Super - ta!

app/assets/javascripts/govuk_publishing_components/components/feedback.js
andysellick: Good point, have tidied up a bit.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
injms: In this instance, would `thisModule` describe what `that` is? I've seen `var that = this` used, but there doesn't seem to be consistency about what `that` is - sometimes it's used for the whole module and sometimes a function inside the module. 

Not a blocker, just a thought.

andysellick: Yeah, I like that. Changed.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -11,53 +10,56 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+          var refocusClass = '.js-' + formToToggle
+          thisModule.element.querySelector(refocusClass).focus()
+        })
+      }
 
-      this.$pageIsUsefulButton.on('click', function (e) {
+      this.pageIsUsefulButton.addEventListener('click', function (e) {
         e.preventDefault()
-        trackEvent(getTrackEventParams(that.$pageIsUsefulButton))
+        trackEvent(getTrackEventParams(thisModule.pageIsUsefulButton))
```
injms, line 57: Outside the scope of removing jQuery, so just a quesiton - does the use of `trackEvent` mean that the 'Yes' button doesn't do anything if a user doesn't accept website usage cookies or leaves the cookie banner alone?

andysellick: It looks that way. The only thing that button seems to do is fire an event at Google analytics. In theory that shouldn't be visible unless cookies have been consented to, but as you said, outside the scope of removing jQuery.

injms (approved):

PR #1164, opened 2019-10-14
Add component print styles

andysellick:
## What
We seem to be missing some print stylesheets for some components, where print styles would be sensible. This PR adds some.

## Why
People print pages on GOV.UK.

## Visual Changes
Screenshots are a bit pointless because most of these changes involve hiding things, but here's a summary.

- [attachment](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/attachment/preview): hide the icon (not sure about this one, should we hide the entire component?)
- [back link](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/back_link/preview): hide completely
- [button](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/button/preview): make all buttons look the same (and like buttons, see below)
- [textarea](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/textarea/preview): general tidy (see below)
- [layout footer](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/layout_footer/preview): hide everything but the OGL and copyright lines
- [layout header](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/layout_header/preview): minor change, hide the 'menu' button
- [metadata](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/metadata/preview): hide toggle and show all updates links
- [search component](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/search/preview): hide completely when printed
- [skip link](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/skip_link/preview): hide it
- [subscription links](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/subscription-links/preview): hide them
- [translation nav](https://govuk-publishing-compo-pr-1164.herokuapp.com/component-guide/translation-nav/preview): hide them

Things I've not added print styles for:

- I was originally going to hide all form elements (particularly the file upload component) but then realised people might actually want to print a form so they can see what fields they need so decided it was better to leave them all visible
- I wanted to add print styles for the details component (the contents of a collapsed one are not shown in print) but apparently this isn't as simple as it might sound, so I've not done that for now
- in theory if a modal dialog is open and someone prints the page it should be visible but as I don't think this is being used on public GOV.UK I don't think it's important right now
- I assumed next and previous navigation wouldn't be needed when printed but it can contain information about the following and preceding pages, so I've left it visible
[...]

PR #2063, opened 2021-05-12
Hide print link component from print version

danacotoran:
It does not make sense for the print link to be visible in the printed document. 
This adds a [design system override class](https://design-system.service.gov.uk/styles/layout/#override-how-elements-display) to hide it on the printed version of the page.

Fixes https://github.com/alphagov/govuk_publishing_components/issues/1770

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Hide print link component from print version ([PR #2063](https://github.com/alphagov/govuk_publishing_components/pull/2063))
+
 ## 24.10.3
 
 * Remove phase banner restrictions ([PR #2057](https://github.com/alphagov/govuk_publishing_components/pull/2057))
--- a/app/views/govuk_publishing_components/components/_print_link.html.erb
+++ b/app/views/govuk_publishing_components/components/_print_link.html.erb
@@ -13,7 +13,7 @@
     margin_bottom: margin_bottom
   })
 
-  wrapper_classes = %w(gem-c-print-link)
+  wrapper_classes = %w(gem-c-print-link govuk-!-display-none-print)
   wrapper_classes << "gem-c-print-link--show-without-js" unless require_js
   wrapper_classes << (shared_helper.get_margin_top)
   wrapper_classes << (shared_helper.get_margin_bottom)


alex-ju (commented):
Previously, we seem to have taken [a different approach on this](https://github.com/alphagov/govuk_publishing_components/pull/1164). What would be the advantages of using the `govuk-frontend` override class instead?

danacotoran:
@alex-ju that's a good question – I think the advantage is explained quite well in [the commit that introduces this class to `govuk-frontend` ](https://github.com/alphagov/govuk-frontend/commit/b9aec5d9a55bc0b3ee83d7b6c2e01ebc8eb55623):
> This is explicit and preferable to listing all the selectors for things that shouldn't be printed. These lists grow unwieldy over time

Using the class instead of explicit `display: none` rules can help us reduce the size of our print stylesheet. 

This reminds me around a year ago I set out to remove `display: none` rules from component print stylesheets in favour of using the `govuk-!-display-none-print` class  https://github.com/alphagov/govuk_publishing_components/pull/1561/. 
Can't quite remember why I never came to finish that task, but I digress 😅

alex-ju:
> @alex-ju that's a good question – I think the advantage is explained quite well in [the commit that introduces this class to `govuk-frontend` ](https://github.com/alphagov/govuk-frontend/commit/b9aec5d9a55bc0b3ee83d7b6c2e01ebc8eb55623):
> 
> > This is explicit and preferable to listing all the selectors for things that shouldn't be printed. These lists grow unwieldy over time
> 
> Using the class instead of explicit `display: none` rules can help us reduce the size of our print stylesheet.
> 
> This reminds me around a year ago I set out to remove `display: none` rules from component print stylesheets in favour of using the `govuk-!-display-none-print` class #1561.
> Can't quite remember why I never came to finish that task, but I digress 😅

This change is certainly welcomed and it seems to be aligned with the work you've done in #1561.

This is very much related to a wider conversation about class composition but also about the way we use override classes from `govuk-frontend`, but since we're currently using override classes widely and we embraced class composition for styling links and headings, I assume the community is fine with the approach.

Do you think you'd have the time to rebase #1561 (add cherry-pick this commit) so we have a consistent way of doing this at the library level?

danacotoran:
> Do you think you'd have the time to rebase #1561 (add cherry-pick this commit) so we have a consistent way of doing this at the library level?

Your suggestion sounds sensible @alex-ju. I'll try to do that today. 
I just have a nagging feeling that there was _some_ good reason that I abandoned that PR, and it's evading me right now. 😄  It will probably come back when I go to rebase that branch.

PR #2066, opened 2021-05-13
Add [Withdrawn] to the og:title tag of withdrawn content

benjamineskola:
## What

For content that has been withdrawn, prefixes the `og:title` tag with '[Withdrawn]'.

https://trello.com/c/7OxshPDC/2498-2-add-withdrawn-to-ogtitle-tag

## Why

> The prefix is currently added to the title metatag, but not og:title. The latter is used when content is shared on social media etc, so in those cases it's not clear to users that content has been withdrawn.

### DIFF
--- a/app/views/govuk_publishing_components/components/_machine_readable_metadata.html.erb
+++ b/app/views/govuk_publishing_components/components/_machine_readable_metadata.html.erb
@@ -17,7 +17,7 @@
 <meta property="og:site_name" content="GOV.UK" />
 <meta property="og:type" content="article" />
 <meta property="og:url" content="<%= page.canonical_url %>" />
-<meta property="og:title" content="<%= page.title %>" />
+<meta property="og:title" content="<% if page.withdrawn? %>[Withdrawn] <% end %><%= page.title %>" />
 <meta property="og:description" content="<%= strip_tags(page.description) %>" />
 
 <% if page.has_image? %>
--- a/spec/components/machine_readable_metadata_spec.rb
+++ b/spec/components/machine_readable_metadata_spec.rb
@@ -84,6 +84,28 @@ describe "Machine readable metadata", type: :view do
     assert_equal escaped_html, JSON.parse(json_linked_data)["foo"]["bar"]
   end
 
+  it "marks the page as withdrawn in the oc:title tag where appropriate" do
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_machine_readable_metadata.html.erb
```diff
@@ -17,7 +17,7 @@
 <meta property="og:site_name" content="GOV.UK" />
 <meta property="og:type" content="article" />
 <meta property="og:url" content="<%= page.canonical_url %>" />
-<meta property="og:title" content="<%= page.title %>" />
+<meta property="og:title" content="<% if page.content_item['withdrawn_notice'].present? %>[Withdrawn] <% end %><%= page.title %>" />
```
thomasleese, line 20: I think it's worth adding this logic into [the presenter](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/machine_readable/page.rb). Perhaps a method like `page.withdrawn?`?

benjamineskola: I think it would be good to be able to use the same logic in the [meta tags presenter](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/meta_tags.rb#L35) too, but then there are scoping issues — I'm not sure if there's somewhere I could put it where it's in scope for both?

spec/components/machine_readable_metadata_spec.rb
```diff
@@ -84,6 +84,27 @@ def component_name
...
+
+    render_component(content_item: example, schema: :article)
+
+    assert_select "meta[property='og:title'][content='[Withdrawn] #{example['title']}']"
+  end
+
+  it "does not mark the page as withdrawn in the oc:title tag unless it has been withdrawn" do
+    example = GovukSchemas::RandomExample.for_schema(frontend_schema: "case_study") do |item|
+      item
+    end
```
thomasleese, line 101:
```suggestion
    example = GovukSchemas::RandomExample.for_schema(frontend_schema: "case_study")
```

I think the block is optional, if there are no modifications to the `item`.

thomasleese (approved):

PR #2074, opened 2021-05-18
Update link styles for show password component

owenatgov:
## What
Update the button styling for the [show password component](https://components.publishing.service.gov.uk/component-guide/show_password) to use the latest link styles on the button text.

## Why
Part of ongoing work by the govuk frontend community to use the latest link styles from the design system. The show password component uses text styling within it's main button that simulates the style of a link so we're applying the link styles here for consistency.

| Before | After |
| --- | --- |
| Screenshot 2021-05-18 at 15 22 32 | Screenshot 2021-05-18 at 15 22 58 |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead. ([PR #1561](https://github.com/alphagov/govuk_publishing_components/pull/1561))
+* Update link styles for show password component ([PR #2074](https://github.com/alphagov/govuk_publishing_components/pull/2074))
 
 ## 24.10.3
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_show-password.scss
@@ -49,6 +49,7 @@
 
   &:hover {
     color: $govuk-link-hover-colour;
+    @include govuk-link-hover-decoration;
   }
 
   &:focus {


danacotoran:
Okay @owenatgov so... I have thoughts. Or more like, questions. As I am not sure of the answer really.  

Firstly: are we also applying the new link styles to things that aren't links, but have some overlap in styling with the links? As from a semantic point of view, even though it looks a little bit like a link, this Show/Hide password element is fundamentally a button. 

Secondly: as a result of this change, I can notice bit of a shift in the focus state of the button which is difficult to explain so I'll use screen recordings to my aid. 

### Before the new link styles

https://user-images.githubusercontent.com/7116819/118683631-7275e200-b7f9-11eb-86a2-cc6d7d66b42c.mov


### After the new link styles

https://user-images.githubusercontent.com/7116819/118683644-74d83c00-b7f9-11eb-97b9-bb6bfc333205.mov

owenatgov:
@danacotoran
Answering first:

1. In short I agree that it's not necessarily a good thing to apply govuk link styles to things that aren't links. My reasoning for doing this is that this element includes blue underline text as part of its design which looks like a link (to me) and this is a case of trying to make our patterns consistent. Thinking about this in more detail I don't really understand why we have this design of an underlined "link" looking thing inside a button.
2. Hrmmmmmmm that new focus state is a bit weird. I would guess what's happening here is a clash between the applied focus state of the show/hide button and the focus state applied by `govuk-link`.

I'm on the fence here. You're right, this is a button, not a link, so it's a bit sus to use an explicit link class. However I'm still a bit perplexed by this design. We have a few patterns that have link-like styles eg: prev/next nav, accordions etc and I think that this is going to require a design call on if we're happy leaving them alone or if we want to plonk the hover states on there.

I'm gonna check if there's a mixin/class in govuk frontend that might isolate the hover styles without impacting the focus state as an immediate next step. I appreciate that response was a bit wooly but did you have any further thoughts?

danacotoran (approved):
Thanks for the elaborate response Owen. This fixed the focus issue I spotted.

I'm with you on the sentiment that the design of this component is slightly perplexing. 
For consistency across GOVUK, it does make sense to apply the new styles to something that looks like a link, and behaves like a link, even if that thing is not technically a link.

PR #2079, opened 2021-05-19
Fix IE11 initCustomEvent error

danacotoran:
Some work was done to allow cookie dependent modules to start immediately when the user consents to cookies. https://github.com/alphagov/govuk_publishing_components/pull/2041 
To make this possible, some functionality was introduced to trigger a custom cookie consent event when consent is given. It seems like some of the code that was introduced causes a JS error on IE11 specifically. 
image

The error appears to be related to the `initCustomEvent` method. Namely, it seems like IE11 does not like when this method is called with less than the required 4 arguments, and consequently throws an error.
This passes `null` as a 4th argument, which should appease IE11 without causing issues in other browsers.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead. ([PR #1561](https://github.com/alphagov/govuk_publishing_components/pull/1561))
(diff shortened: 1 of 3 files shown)


alex-ju (approved):
Good spot @danacotoran! It seems like we're making the same mistake in the [reorderable-list component](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/reorderable-list.js#L101) would you mind updating it there too?

danacotoran:
Thanks @alex-ju! 
Shall I make the change in `reorderable-list.js` as part of this PR or do I think I should open a new one?

PR #2080, opened 2021-05-20
Update accordion link hover to new styles

injms:
## What

Update the accordion hover states so it uses the new hover states from [GOV.UK Frontend v3.12.0](https://github.com/alphagov/govuk-frontend/releases/tag/v3.12.0).

This couldn't be automatically updated alongside by bumping because the accordion needs the hover states on text within the button, this required a number of bespoke changes.

Slight tweak to hover of the 'Show all sections' link - the hover state visual change is now triggered by hovering the whole button, not just the text.

## Why

To bring the accordion component up to date, alongside all of the other components that can't be automatically updated.

## Visual Changes

| State | Before | After | 
| - | - | - |
| Show all link hover | image | image |
| Show all link focus and hover | image | image |
| Show section hover | image | image |
| Show section hover and focus | image | image |


owenatgov (approved):
Nicely done 👍  This is extra useful as I can steal it for the step by step nav update 👀

PR #2081, opened 2021-05-20
Add govuk-link classes to government navigation links

owenatgov:
## What
Adds the following classes to links on the [government navigation](https://components.publishing.service.gov.uk/component-guide/government_navigation) component. The classes it adds are:

- `govuk-link`
- `govuk-link--no-underline`
- `govuk-link--inverse`

## Why
Part of ongoing work by the govuk frontend community to ensure we are using the new link styles from the altest release of govuk frontend.

This is not a full implementation of the new link styles due to clashes with styles in the header from govuk-toolkit, our legacy implementation of what is now govuk frontend. There's no value in trying to fix this now as it would require either an update to the archived govuk-toolkit repo or a set of convoluted override styles between the gem and static. The decision to not implement this fully is further confounded by:

- Work by the govuk accessibility team to remove the static core layout, therefore the govuk-toolkit implementation
- Work by the govuk explore team which could make this component redundant

These classes at the very least mean that when these 2 workstreams are complete, we'll be using the correct classes and it does subtly apply a small portion of the new link styles (distance between the text and the underline on hover).

| Before | After |
| --- | --- |
| Screenshot 2021-05-20 at 10 44 38 | Screenshot 2021-05-20 at 10 44 45 |

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/government_navigation.yml
+++ b/app/views/govuk_publishing_components/components/docs/government_navigation.yml
@@ -2,6 +2,8 @@ name: Government navigation
 description: Navigation placed in the header by Slimmer and used by pages which sit
   under the /government path. This is a markup only component and is not available
   for preview here. It can be passed a string to mark a link as being active.
+body: | 
+  Please note: because the markup on this component is currently tied to styles that aren't present in the components gem, they will appear "broken". This is because they are intended to be used in the [header component](/component-guide/layout_header/with_left_search_and_navigation) as white links against a black background. You can see a styled example of this component in use on the ['Counter-Fraud Standards and Profession' page](https://www.gov.uk/government/groups/counter-fraud-standards-and-profession).
 accessibility_criteria: |
   The government navigation component must:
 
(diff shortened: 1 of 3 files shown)


injms (approved):

owenatgov:
@injms Good spot! Added a note to the docs.

app/views/govuk_publishing_components/components/docs/government_navigation.yml
```diff
@@ -2,6 +2,8 @@ name: Government navigation
 description: Navigation placed in the header by Slimmer and used by pages which sit
   under the /government path. This is a markup only component and is not available
   for preview here. It can be passed a string to mark a link as being active.
+body: | 
+  Please note: because the markup on this component is currently tied to styles that aren't tied to this components gem, they will appear "broken". This is because they are intended to be used in the GOV.UK header as white links against a black background. You can see a styled example of this component in use [here](https://www.gov.uk/government/groups/counter-fraud-standards-and-profession).
```
injms, line 6:
```suggestion
  Please note: because the markup on this component is currently tied to styles that aren't present in the components gem, they will appear "broken". This is because they are intended to be used in the [header component](/component-guide/layout_header/with_left_search_and_navigation) as white links against a black background. You can see a styled example of this component in use on the ['Counter-Fraud Standards and Profession' page](https://www.gov.uk/government/groups/counter-fraud-standards-and-profession). 
```

Issue #2086, opened 2021-05-20
Accordion anchor link navigation breaks with invalid anchor


andysellick:
I'm not sure I fully understand the exact problem here but raising for further investigation.

The accordion has an option to allow a [section to open based on the URL hash](https://components.publishing.service.gov.uk/component-guide/accordion/with_the_anchor_link_navigation). This can break the entire accordion JS if the hash does not match an accordion section, but does match another element.

I found this in the components audit pages (running locally) where accordions are inside tabs. Reloading the page having clicked a tab left the URL hash as `#applications`, which resulted in the below.

![Screenshot 2021-05-20 at 17 17 27](https://user-images.githubusercontent.com/861310/119013836-48eebf00-b98f-11eb-8856-a8b71c2025a4.png)

This is the error in the console:

![Screenshot 2021-05-20 at 17 17 34](https://user-images.githubusercontent.com/861310/119013865-5015cd00-b98f-11eb-8d5e-4e9c41d6d54c.png)

According to the documentation this option can be disabled if the `anchor_navigation` flag is false, but this didn't change the problem (this flag defaults to false, and I wasn't setting it).

Thanks to @owenatgov for helping debug this.

andysellick:
@owenatgov I'm trying to work out what else needs fixing on this. I thought there might be a problem if the anchor link matched both an accordion section and a tab (for example) but that shouldn't happen because IDs are unique. Can you clarify what else needs fixing?

MartinJJones:
@andysellick following on from the frontend triage meeting yesterday.

An issue exists in the component given the steps below:

- `anchor_navigation` flag is passed as `true`
- an `id` is passed in that exists in the document but is not related to the accordion, a heading for example
- we then get the same error as before in the getContainingSection method
![Screenshot 2022-08-19 at 14 04 51](https://user-images.githubusercontent.com/28779939/185624558-379c4035-8aec-470b-a65a-157ab83859d5.png)

Broken example - id in the url is for a regular heading element separate to the accordion
https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#section-title

A possible fix could be to get the id from the module instead of the document:
https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/accordion.js#L64

There may be other fixes and refactoring required when testing further.

PR #2087, opened 2021-05-20
Fix accordion anchor link navigation bug

andysellick:
## What
If the URL of a page with an accordion on it has a location hash in the URL that doesn't match an actual accordion section, an error can occur as described in issue https://github.com/alphagov/govuk_publishing_components/issues/2086

This partly fixes this problem by checking for an actual value for the `data-anchor-navigation` attribute, however this might not be a solution to the whole problem.

## Why
Noticed in the component auditing pages (locally) the accordion was breaking if tabs were changed and the page reloaded.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Changelog? Other than that, grrrrreat stuff!

PR #2088, opened 2021-05-20
New link styles for prev next navigation

owenatgov:
## What
Amends the hover state for the [previous and next navigation component](https://components.publishing.service.gov.uk/component-guide/previous_and_next_navigation), specifically applies the new `govuk-link` hover decoration to the label element and to the title element when a label isn't present.

## Why
Part of ongoing work by the govuk frontend community to ensure we are using the new link styles from the latest govuk frontend release.

This change specifically is an accessibility enhancement as, besides the slight background change, there is no clear indicator that the state of the component has changed.

| State | Before | After |
| --- | --- | --- |
| With label | Screenshot 2021-05-20 at 17 55 46 | Screenshot 2021-05-20 at 17 55 56 |
| Without label | Screenshot 2021-05-20 at 17 56 07 | Screenshot 2021-05-20 at 17 56 16 |


andysellick (approved):
👍

PR #2090, opened 2021-05-21
Make jasmine tests pass in Internet Explorer

andysellick:
## What
Make (most of) our jasmine tests for JavaScript pass in Internet Explorer, currently aiming at IE10+.

## Why
We're removing jQuery across GOV.UK and we need confidence that the replacement code will work in all the browsers we support, especially when JS often has different ways (new/old) of doing the same thing. It would be good to be able to run our tests in all browsers and know they should pass.

Currently there are 188 failing tests. This PR fixes all but 13 of them, so there's still some work to do, but the changes are getting too big to review and I want to get some of these fixes out pronto.

Screenshot 2021-05-25 at 10 12 37

A lot of the failing tests seem to be due to syntax issues within the tests themselves, rather than our actual JS, so adding a few lines to replace a `forEach` loop with a `for` loop doesn't feel like much of a price to pay for working tests.

## Visual Changes
None.

Bonus side effects:

- magna charta graphs (govspeak) now appear to work in IE10
- analytics code (cookie banner) is quite broken in IE10, this change might fix some of that: https://github.com/alphagov/govuk_publishing_components/issues/2103

Trello card: https://trello.com/c/CAkmuuTK/424-make-js-tests-pass-in-all-browsers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,10 +6,14 @@
 - Don't include changes that are purely internal. The CHANGELOG should be a
   useful summary for people upgrading their application, not a replication
   of the commit log.
-  
+
+## Unreleased
+
+* Make jasmine tests pass in Internet Explorer ([PR #2090](https://github.com/alphagov/govuk_publishing_components/pull/2090))
+
 ## 24.13.3
 
-* Add event tracking to header elements ([#2110](https://github.com/alphagov/govuk_publishing_components/pull/2110)) 
+* Add event tracking to header elements ([PR #2110](https://github.com/alphagov/govuk_publishing_components/pull/2110)) 
 
 ## Unreleased
 * Allow a custom link to be used on the header logo ([PR #2114](https://github.com/alphagov/govuk_publishing_components/pull/2114))
--- a/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
@@ -33,7 +33,9 @@ window.GOVUK.Modules.Checkboxes = window.GOVUKFrontend;
     if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
       // Where checkboxes are manipulated externally in finders, `suppressAnalytics`
       // is passed to prevent duplicate GA events.
-      if (!event.detail || (event.detail && event.detail.suppressAnalytics !== true)) {
+      // use Oliver Steele's Nested Object Access Pattern https://hackernoon.com/accessing-nested-objects-in-javascript-f02f1bd6387f
+      var allowAnalytics = ((event || {}).detail || {}).suppressAnalytics !== true
+      if (allowAnalytics) {
         var $checkbox = event.target
         var category = $checkbox.getAttribute('data-track-category')
         if (category) {
@@ -100,11 +102,11 @@ window.GOVUK.Modules.Checkboxes = window.GOVUKFrontend;
     var $exclusiveOption = $checkboxes.querySelector('input[type=checkbox][data-exclusive]')
     var $nonExclusiveOptions = $checkboxes.querySelectorAll('input[type=checkbox]:not([data-exclusive])')
 
-    if ($currentCheckbox.dataset.exclusive === 'true' && $currentCheckbox.checked === true) {
+    if ($currentCheckbox.getAttribute('data-exclusive') === 'true' && $currentCheckbox.checked === true) {
       for (var i = 0; i < $nonExclusiveOptions.length; i++) {
         $nonExclusiveOptions[i].checked = false
       }
-    } else if ($currentCheckbox.dataset.exclusive !== 'true' && $currentCheckbox.checked === true) {
+    } else if ($currentCheckbox.getAttribute('data-exclusive') !== 'true' && $currentCheckbox.checked === true) {
       if ($exclusiveOption) {
         $exclusiveOption.checked = false
       }
--- a/app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js
@@ -15,7 +15,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ContextualGuidance.prototype.handleFocus = function (event) {
     this.hideAllGuidance()
-    if (!event.target.dataset.contextualGuidanceHideOnly) {
+    if (!event.target.getAttribute('data-contextual-guidance-hide-only')) {
       this.$guidance.style.display = 'block'
     }
   }
--- a/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
@@ -2,15 +2,17 @@
   'use strict'
   window.GOVUK = window.GOVUK || {}
 
-  window.GOVUK.triggerEvent = function (element, eventName) {
-    var params = { bubbles: true, cancelable: true }
+  window.GOVUK.triggerEvent = function (element, eventName, parameters) {
+    var params = parameters || {}
+    params.bubbles = true
+    params.cancelable = true
     var event
 
     if (typeof window.CustomEvent === 'function') {
       event = new window.CustomEvent(eventName, params)
     } else {
       event = document.createEvent('CustomEvent')
-      event.initCustomEvent(eventName, params.bubbles, params.cancelable, null)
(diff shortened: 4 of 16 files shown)


spec/javascripts/components/accordion-spec.js
```diff
@@ -103,18 +103,21 @@ describe('An accordion component', function () {
...
         expect(section).toHaveClass('gem-c-accordion__section--expanded')
         expect(section.querySelector('.gem-c-accordion__toggle-text').innerText).toEqual('Hide')
-      })
+      }
 
       topLevelControl.click()
       expect(topLevelControl.innerText).toEqual('Show all sections')
 
-      element.querySelectorAll('.gem-c-accordion__section').forEach(function (section) {
+      for (s = 0; s < sections.length; s++) {
```
injms, line 116: Instead of reassigning `s`, could this another variable?

```suggestion
      for (var x = 0; x < sections.length; x++) {
```

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
```diff
@@ -33,7 +33,8 @@ window.GOVUK.Modules.Checkboxes = window.GOVUKFrontend;
     if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
       // Where checkboxes are manipulated externally in finders, `suppressAnalytics`
       // is passed to prevent duplicate GA events.
-      if (!event.detail || (event.detail && event.detail.suppressAnalytics !== true)) {
+      var allowAnalytics = ((event || {}).detail || {}).suppressAnalytics !== true
```
injms, line 36: I'm finding this hard to parse - would it be better to split this over several lines? Or into several variables?

app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
```diff
@@ -2,15 +2,17 @@
...
+    params.bubbles = true
+    params.cancelable = true
     var event
 
     if (typeof window.CustomEvent === 'function') {
       event = new window.CustomEvent(eventName, params)
     } else {
       event = document.createEvent('CustomEvent')
-      event.initCustomEvent(eventName, params.bubbles, params.cancelable, null)
+      event.initCustomEvent(eventName, params.bubbles, params.cancelable, ((params || {}).detail))
```
alex-ju, line 15: Not sure I get this, if `params` is `parameters || {}` why would we need to check if it exists on this line again? could it be `params.detail` as we'll get an empty object on L6 in the case where no parameters are passed?

alex-ju (commented):
I didn't know about the Jasmine's `callFake ` not being supported in IE. One comment on top of Ian's otherwise looks good!

injms:
This is super work, thank you! 

One thing we need to think about is how we make sure that we don't slide back into tests failing for Internet Explorer 10+ again - maybe using a service like BrowserStack's Automate - but that's a task for another day.

app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js
```diff
@@ -15,7 +15,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   ContextualGuidance.prototype.handleFocus = function (event) {
     this.hideAllGuidance()
-    if (!event.target.dataset.contextualGuidanceHideOnly) {
+    if (!event.target.getAttribute('contextualGuidanceHideOnly')) {
```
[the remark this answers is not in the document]
andysellick: Ooh, yes, potentially. Good spot. I'll check.

spec/javascripts/components/accordion-spec.js
andysellick: It could, yes. What's the value of that? Wouldn't creating a new variable within this scope use (fractionally) more memory?

andysellick: Oh but this is within the test so not a big deal if it does, I guess.

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
andysellick: I'll have a look, see what I can do.

app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js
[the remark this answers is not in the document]
andysellick: I can't see where this attribute appears - there doesn't seem to be a test for it, and the only instance I can find of `contextualGuidanceHideOnly` is in this file. @alex-ju is this something in the publishing applications? Is it needed?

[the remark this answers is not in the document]
alex-ju: so, the equivalent of `dataset.contextualGuidanceHideOnly` is getAttribute('data-contextual-guidance-hide-only')` (attributes get camelcased when added to the dataset object). it used to be a feature in `content-publisher`, but must have ben removed as I can't see any instance of it atm.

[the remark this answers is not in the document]
andysellick: Thanks both. I've updated the code and added a test for this. Although I'm uncertain if I understand this functionality - it seems like this option allows us to hide the contextual guidance for certain elements, so that they never appear again? Why is that useful? Or have I misunderstood?

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
andysellick: This is a very clever technique I found online for finding whether a nested attribute exists in an object. It looks complicated at first, but it's actually quite simple. Here's an example.

Say you want to check for the existence of a nested attribute `object.first.second`, but `first` might not exist. If you try `if (object.first.second)` the code will error, because `first` doesn't exist. Using this technique, at each level of the evaluation, the code returns either the attribute it finds (e.g. `first`) or an empty object. That then gets appended to the next level down, which means an error can't occur, we get to the end of the expression and either find what we're looking for, or we find a series of empty nested objects doesn't contain what we're looking for.

Breaking this up into more lines of code was what I was trying to avoid as it gets really verbose. I was nesting multiple if statements (`if (event) { if (event.detail) { if (event.detail.suppressAnalytics) { }}}`).

I've added a comment in the code to explain this and provided a link to more details.

app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
andysellick: Yes, absolutely right, that check is unnecessary. Must have been left over from an earlier iteration when I was working this out. Have changed.

app/assets/javascripts/govuk_publishing_components/components/contextual-guidance.js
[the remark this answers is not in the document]
alex-ju: It was more of an exception to the rule. So the contextual guidance was designed to be visible when a certain input element is focused, and be hidden when the input is 'blurred'. This exception was set in place for a certain case where we only wanted to make sure the contextual guidance is hidden on blur, without showing it on focus. The behaviour was documented in the commits in `content-publisher`, before the component got moved to the library – I guess one of the downsides of developing a component at the app-level.

As the feature is not being used anymore, I don't see value in keeping it around – arguably not too relevant to this PR.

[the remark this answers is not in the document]
andysellick: Great. I'll raise a separate PR once this is merged to take it out, thanks for the context.

app/assets/javascripts/govuk_publishing_components/components/checkboxes.js
injms: Thanks - the link will definitely help future me who'll completely forget about this and go "huh" when i read this again 😄  👍 

Can't wait to be able to use optional chaining!

CHANGELOG.md
```diff
@@ -6,10 +6,14 @@
 - Don't include changes that are purely internal. The CHANGELOG should be a
   useful summary for people upgrading their application, not a replication
   of the commit log.
-  
+
+## Unreleased
+
+* Make jasmine tests pass in Internet Explorer ([PR #2090](https://github.com/alphagov/govuk_publishing_components/pull/2090))
```
injms, line 12: Do we need to add this to changelog, as it's a purely internal change?

injms (approved):

CHANGELOG.md
andysellick: I think so, because I believe it also fixes some bugs in IE10 (documented in the description).

PR #2092, opened 2021-05-24
New link styles for step by step nav

owenatgov:
## What
Apples the latest link styles from govuk frontend to the [step by step navigation component](https://components.publishing.service.gov.uk/component-guide/step_by_step_nav).

## Why
Part of ongoing work by the govuk frontend community to ensure that we are using the new link styles from the latest release of govuk frontend.

This change also includes a minor tweak to the "show all" button on this component and the accordion component focus states so that the thick underline from the new hover state doesn't interfere visually with the small focus state.

## Visual Changes
| State | Before | After |
| --- | --- | --- |
| All sections button hover | Screenshot 2021-05-24 at 11 50 04 | Screenshot 2021-05-24 at 11 50 20 |
| Single section button hover | Screenshot 2021-05-24 at 11 50 26 | Screenshot 2021-05-24 at 11 50 33 |
| All sections button focus and hover | Screenshot 2021-05-24 at 11 50 52 | Screenshot 2021-05-26 at 17 27 36 |
| Links default | Screenshot 2021-05-24 at 15 54 04 | Screenshot 2021-05-24 at 15 54 11 |
| Links standard link hover | Screenshot 2021-05-24 at 15 54 18 | Screenshot 2021-05-24 at 15 54 26 |
| Links active link hover | Screenshot 2021-05-24 at 15 54 34 | Screenshot 2021-05-24 at 15 54 46 |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
@@ -113,10 +113,11 @@ $top-border: solid 1px govuk-colour("mid-grey", $legacy: "grey-3");
   position: relative;
   z-index: 1; // this and relative position stops focus outline underlap with border of accordion
   margin: .5em 0 1.5em;
-  padding: 0 0 govuk-spacing(1);
+  padding: 0 0 govuk-spacing(2);
 
   &:hover .gem-c-step-nav__button-text {
-    text-decoration: underline;
+    @include govuk-link-decoration;
+    @include govuk-link-hover-decoration;
   }
 
   .gem-c-step-nav--large & {
@@ -338,7 +339,8 @@ $top-border: solid 1px govuk-colour("mid-grey", $legacy: "grey-3");
     }
 
     .gem-c-step-nav__button-text {
-      text-decoration: underline;
+      @include govuk-link-decoration;
+      @include govuk-link-hover-decoration;
     }
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -21,6 +21,7 @@
 * Update links styles for accordion component ([PR #2080](https://github.com/alphagov/govuk_publishing_components/pull/2080))
 * New link styles for prev next navigation ([PR #2088](https://github.com/alphagov/govuk_publishing_components/pull/2088))
 * Update organisation logos hover state ([PR #2089](https://github.com/alphagov/govuk_publishing_components/pull/2089))
+* New link styles for step by step nav ([PR #2092](https://github.com/alphagov/govuk_publishing_components/pull/2092))
 
 ## 24.10.3
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
@@ -36,7 +36,7 @@ $gem-c-accordion-bottom-border-width: 1px;
     -webkit-appearance: none;
     cursor: pointer;
     margin-bottom: govuk-spacing(4);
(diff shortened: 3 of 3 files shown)


samdub:
This is looking good for the accordion open/close, but would be good to see how the new style is applied to the links within each section. 

Specifically:
1. How links to tasks with the accordion are styled and spaced
2. Step by step has a special link style for the page that you are on, currently black and underlined. This is visually differentiate links to other pages and indicate it's the page you're on. A reckon: it might be nice to drop the underline (except on hover) for this link to add further visual differentiation from the surrounding links. 

@miaallers is probably the best person to review this from a design point of view.

Screenshot 2021-05-24 at 12 06 04

owenatgov:
@samdub I've added screenshots for what the active links look like. In short these styles were inherited as part of the link change automatically so don't require further work 🙌 

On the active link, that's a reasonable reckon. Right now the black colour in the only visual indicator. My concern is that without the underline, folks won't know it's a link in context, even if we change the colour from black to blue. Maybe we could make it bold? I can chat to Mia about this in detail to find a reasonable solution.

owenatgov:
I had a chat separately with Sam and Mia about this and they're happy with the active link styles as they are currently. This is therefore ready for dev review!

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -125,6 +126,10 @@ $top-border: solid 1px govuk-colour("mid-grey", $legacy: "grey-3");
 
   &:focus {
     @include step-nav-chevron-focus-state;
+
+    .gem-c-step-nav__button-text {
+      text-decoration: none;
+    }
```
chris-gds, line 132: 👋🏼   

You've mentioned 

> This change also includes a minor tweak to focus with hover so that the thick underline doesn't interfere visually with the small focus state.

I'm in two minds about this inconsistency for a button + focus + hover. However I understand what you mean about it being crowded.  I'd be tempted to increase the padding to enlarge the focus area to factor in the larger underline.

May-26-2021 13-02-31
vs
https://components.publishing.service.gov.uk/component-guide/step_by_step_nav

IE11 for ref
May-26-2021 12-55-14

----

Not related but i've noticed with the numbers on the side a hover is present but only while no "actively" selected and / black

andysellick (approved):
Looks good, although I think Chris's comment should be responded to before merging.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
owenatgov: What do you think of this?

| Default | Hover + focus |
| --- | --- |
| Screenshot 2021-05-26 at 16 30 11 | Screenshot 2021-05-26 at 16 29 57 |

I was concerned that it'd be too spacious but it's not too bad and is a little less crowded. The use case of focused _and_ underlined is also very slim IMO. If you're happy with this I can also apply this to the accordion and update the PR description.

chris-gds: >The use case of focused and underlined is also very slim IMO.

Agreed, I'm also leaning towards the consistency aspect of the other buttons + focus + hover as well. 

> If you're happy with this I can also apply this to the accordion and update the PR description. 

It's probably also worth noting, reviewing the Design System, I think the inverse is probably more true with elements _not_ getting any hover change once focused.  (I know the general focus state in this context is being looked at in parallel)

So for the context of this link update keeping as is would be all good rn - thanks @owenatgov.

andysellick: Personally I don't like the focus also having the underline. But I'm one person. Can we consult a designer on this?

PR #2094, opened 2021-05-24
Release 24.11.0

benjamineskola:
* Add `[Withdrawn]` to the beginning of the `og:title` meta tag for withdrawn pages ([PR #2066](https://github.com/alphagov/govuk_publishing_components/pull/2066))
* Fix accordion anchor link navigation bug ([PR #2087](https://github.com/alphagov/govuk_publishing_components/pull/2087))
* Remove jQuery from the feedback component ([PR #2062](https://github.com/alphagov/govuk_publishing_components/pull/2062))
* Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead. ([PR #1561](https://github.com/alphagov/govuk_publishing_components/pull/1561))
* Fix IE11 `initCustomEvent` error ([PR #2079](https://github.com/alphagov/govuk_publishing_components/pull/2079))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 24.11.0
 
+* Add `[Withdrawn]` to the beginning of the `og:title` meta tag for withdrawn pages ([PR #2066](https://github.com/alphagov/govuk_publishing_components/pull/2066))
 * Fix accordion anchor link navigation bug ([PR #2087](https://github.com/alphagov/govuk_publishing_components/pull/2087))
 * Remove jQuery from the feedback component ([PR #2062](https://github.com/alphagov/govuk_publishing_components/pull/2062))
 * Remove `display: none` rules from component print stylesheets, and use the `govuk-!-display-none-print` class instead. ([PR #1561](https://github.com/alphagov/govuk_publishing_components/pull/1561))
@@ -17,7 +18,6 @@
 ## 24.10.3
 
 * Remove phase banner restrictions ([PR #2057](https://github.com/alphagov/govuk_publishing_components/pull/2057))
-* Add `[Withdrawn]` to the beginning of the `og:title` meta tag for withdrawn pages ([PR #2066](https://github.com/alphagov/govuk_publishing_components/pull/2066))
 
 ## 24.10.2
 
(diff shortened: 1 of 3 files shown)


thomasleese:
I think this release should be more than a patch bump given the number of things that have changed.

thomasleese (approved):

benjamineskola:
@thomasleese Is the branch name significant? If Jenkins is going to be using it I think I'll need to re-open the PR.

thomasleese:
The branch name shouldn't matter.

owenatgov:
Couple o' thinks:

- Does jenkins use the branch name or the PR title? If it uses the branch name then yes you'll have to rename the branch otherwise the release is technically going to be misleading (a minor change, not a fix). Sorry 😬 
- How critical is this? We've got some link style work that's nearly done. We want to avoid release noise if we can so if this can wait a few days that'd be really cool.

thomasleese:
> * Does jenkins use the branch name or the PR title? If it uses the branch name then yes you'll have to rename the branch otherwise the release is technically going to be misleading (a minor change, not a fix). Sorry 😬

The branch name will appear in the merge commit added by GitHub, but the release is generated from the version number, not the branch name.

> * How critical is this? We've got some link style work that's nearly done. We want to avoid release noise if we can so if this can wait a few days that'd be really cool.

@nicholsj Do you have an opinion on this?

thomasleese:
> @nicholsj Do you have an opinion on this?

I spoke to Jonathan, and we don't mind waiting a few days.

Although I wonder, is it really so bad to have lots of releases?

owenatgov:
@thomasleese Cheers Tom. Let me know if we end up taking too long. Don't wanna block things being good :)

Re: many releases, this was a discussion by the govuk frontend community that we'd like to avoid a lot gem releases in a short space of time if we can help it as we've noticed that less-loved frontend apps might "skip" a few releases, meaning that those dependabot PRs are sometimes secretly piping in a lot of new features which can lead to less control over what goes out in a given release. Admittedly that's off the top of my head. If you wanted to discuss this in more detail with the FEs and get a better idea of the decision I can air drop in a few of the devs who were involved in the decision.

In this case I maybe wouldn't mind splitting the releases as the link style work we're doing means there'd b a looooot of lines in the changelog.

owenatgov:
👋  Just an update that we haven't quite finished the link style work I mentioned just yet and I'll be going on AL after today. I'd recommend merging this in and we can create a separate release for the link style work.

PR #2095, opened 2021-05-24
Use the`init` API to initialise modules

alex-ju:
## What
Use the `init` API to initialise modules.

## Why

- To converge to a single JavaScript pattern across GDS (aligning the current scripts with the ones we currently have in `govuk-frontend`)
- To be able to port components upstream (in `govuk-frontend`) without needing to refactor the code
- To initialise [JavaScript modules without requiring jQuery](https://trello.com/c/sghHecNi/91-remove-jquery-dependency)

## Modules in GOV.UK
JavaScript modules in GOV.UK are actually design patterns for keeping pieces of code independent of other components. We currently have 3 design patterns for writing JavaScript in GOV.UK applications.

### GOV.UK Frontend Toolkit Modules
Mostly found in legacy scripts, either migrated from Toolkit, Elements, or written in the same era.
```
  var that = this
  that.start = function ($element) {
    //
  }
}
```

### GOV.UK Publishing Modules
Found across GOV.UK applications, on relatively new scripts (~2-3 years old). Similar to the `govuk-frontend` approach described below, it relies on the classic prototype pattern but with a `start` function that takes a jQuery object that initialises on.
```
function TestModule () { }
TestModule.prototype.start = function ($element) {
  //
}
```

### GOV.UK Frontend Modules
Used in `govuk-frontend` scripts. Relies on the classic prototype pattern with an  `init` function. The element passed to the constructor function is a standard HTMLElement.
```
function TestModule (element) {
  this.element = element
}
TestModule.prototype.init = function () {
  //
}
```

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -7,15 +7,12 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function GemAccordion () { }
-
-  GemAccordion.prototype.start = function ($module) {
-    this.$module = $module[0]
+  function GemAccordion ($module) {
+    this.$module = $module
     this.sectionClass = 'gem-c-accordion__section'
     this.moduleId = this.$module.getAttribute('id')
     this.sections = this.$module.querySelectorAll('.' + this.sectionClass)
     this.openAllButton = ''
-    this.browserSupportsSessionStorage = helper.checkForSessionStorage()
     this.controlsClass = 'gem-c-accordion__controls'
     this.openAllClass = 'gem-c-accordion__open-all'
     this.openAllTextClass = 'gem-c-accordion__open-all-text'
@@ -39,6 +36,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.actions.showAllText = this.$module.getAttribute('data-show-all-text')
     this.$module.actions.hideAllText = this.$module.getAttribute('data-hide-all-text')
     this.$module.actions.thisSectionVisuallyHidden = this.$module.getAttribute('data-this-section-visually-hidden')
+  }
+
+  GemAccordion.prototype.init = function () {
+    this.browserSupportsSessionStorage = helper.checkForSessionStorage()
 
     // Indicate that JavaScript has worked
     this.$module.classList.add('gem-c-accordion--active')
--- a/app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
@@ -2,14 +2,16 @@ window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  function ModalDialogue () { }
-
-  ModalDialogue.prototype.start = function ($module) {
-    this.$module = $module[0]
+  function ModalDialogue ($module) {
+    this.$module = $module
     this.$dialogBox = this.$module.querySelector('.gem-c-modal-dialogue__box')
     this.$closeButton = this.$module.querySelector('.gem-c-modal-dialogue__close-button')
     this.$html = document.querySelector('html')
     this.$body = document.querySelector('body')
+  }
+
+  ModalDialogue.prototype.init = function () {
+    if (!this.$dialogBox || !this.$closeButton) return
 
     this.$module.resize = this.handleResize.bind(this)
     this.$module.open = this.handleOpen.bind(this)
--- a/spec/javascripts/components/copy-to-clipboard-spec.js
+++ b/spec/javascripts/components/copy-to-clipboard-spec.js
@@ -7,8 +7,8 @@ describe('Copy to clipboard component', function () {
   var FIXTURE
 
   function loadCopyToClipboard () {
-    var copyToClipboard = new GOVUK.Modules.CopyToClipboard()
-    copyToClipboard.start($('[data-module=copy-to-clipboard]'))
+    var element = document.querySelector('[data-module=copy-to-clipboard]')
+    new GOVUK.Modules.CopyToClipboard(element).init()
   }
 
   beforeEach(function () {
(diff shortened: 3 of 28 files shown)


spec/javascripts/components/copy-to-clipboard-spec.js
```diff
@@ -25,7 +25,8 @@ describe('Copy to clipboard component', function () {
 
     loadCopyToClipboard()
 
-    $('.govuk-button').trigger('click')
+    var copyButton = document.querySelector('button')
+    copyButton.dispatchEvent(new window.Event('click'))
```
andysellick, line 29: Is it worth using the `GOVUK.triggerEvent` script here? Might make it more cross browser compatible.

(there's a few other places like this as well)

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
```diff
@@ -2,14 +2,16 @@ window.GOVUK = window.GOVUK || {}
...
+  function ModalDialogue ($module) {
+    this.$module = $module
     this.$dialogBox = this.$module.querySelector('.gem-c-modal-dialogue__box')
     this.$closeButton = this.$module.querySelector('.gem-c-modal-dialogue__close-button')
     this.$html = document.querySelector('html')
     this.$body = document.querySelector('body')
+  }
+
+  ModalDialogue.prototype.init = function () {
+    if (!this.$dialogBox || !this.$closeButton) return
```
andysellick, line 14: I notice you've added this check in a few scripts. Can you explain why this is needed now?

spec/javascripts/components/reorderable-list-spec.js
```diff
@@ -117,9 +117,9 @@ describe('Reorderable list component', function () {
     var reorderableList
     var sortable
     beforeEach(function () {
-      reorderableList = new GOVUK.Modules.ReorderableList()
-      reorderableList.start($(element))
-      spyOnEvent($(element), 'reorder-drag')
+      reorderableList = new GOVUK.Modules.ReorderableList(element)
+      reorderableList.init()
```
andysellick, line 121: In other places the initialisation of the module is done in a single line e.g. `new GOVUK.Modules.ReorderableList(element).init()`. I can see that in some places (in this file, for example) that this is done so that a spy can be initialised, but here it looks like it could be combined. Could this be made more consistent?

spec/javascripts/components/show-password-spec.js
```diff
@@ -14,7 +14,7 @@ describe('A show password component', function () {
           '</div>' +
         '</div>'
       )
-      new GOVUK.Modules.ShowPassword().start(element)
+      new GOVUK.Modules.ShowPassword(element[0]).init()
```
andysellick, line 17: I'd not normally push to remove jQuery from the tests as well, but you've done it in other places and it would feel cleaner if module was being passed `element` instead of `element[0]`. Non-blocking.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -39,6 +36,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.actions.showAllText = this.$module.getAttribute('data-show-all-text')
     this.$module.actions.hideAllText = this.$module.getAttribute('data-hide-all-text')
     this.$module.actions.thisSectionVisuallyHidden = this.$module.getAttribute('data-this-section-visually-hidden')
+  }
```
andysellick, line 39: A lot of things get initialised here before the `init` function. Is that bad, or is that okay? Genuinely uncertain.

andysellick (commented):
Great stuff here Alex 👏  some minor comments and questions, but nothing major.

app/assets/javascripts/govuk_publishing_components/components/modal-dialogue.js
alex-ju: To make the scripts a bit more resilient. If the markup is not as expected (e.g. the script is called, but the component markup not, or has been altered by other scripts) we prevent the script from failing and potentially dragging the entire JavaScript execution off with it.

spec/javascripts/components/reorderable-list-spec.js
alex-ju: Yes, I've tried to only separate it for Spys – this might be accidental, will double-check and update.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
alex-ju: My view is that the constructor should contain all static attributes (hence the queries and getters moved here) then the `init` should handle behavior (event listeners and/or subsequent function calls).

spec/javascripts/components/show-password-spec.js
alex-ju: I can have a go, but given the amount of jQuery used in tests I'd rather give it another go at removing jQuery from tests separately, not to stall this PR.

spec/javascripts/components/reorderable-list-spec.js
alex-ju: Ah, we need the instance on [L123](https://github.com/alphagov/govuk_publishing_components/pull/2095/files/a466858544bc561805e4dbc2bb2c63cb8c9bd8af..35aeb30eabe389109b48fc9fb8705b979de6326a#diff-d528e7ff346f0e72e2a2ebfe3b55d754ebc8f60603c04eb3c4aff0349938e73eR123) to check for a property, that's why I had to keep it separately.

spec/javascripts/components/show-password-spec.js
andysellick: Fair point, non-blocking.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
andysellick: And presumably there's no situation where the constructor gets called but the `init` doesn't?

alex-ju: With the GOV.UK modules system there isn't. But when not used with the modules system (for example if we want to move the code to `govuk-frontend`) it gives the flexibility of initialising the script at any point in time/with any mechanishm and without querying the attributes again.

andysellick: Great, thanks for the explanation.

spec/javascripts/components/copy-to-clipboard-spec.js
alex-ju: Have updated to use the new `triggerEvent` utility

andysellick (approved):

PR #2109, opened 2021-06-02
Reset organisation logo margin

injms:
Setting the organisation's margin to zero as the component inherited the browser's default margins when used as a heading.

Also tidied up the component template to use `content_tag` instead of interpolating the tags, and added a nicer way of checking for an appropriate heading level.

## What

 - Set the organisation logo component's margins to zero
 - Swapped to using `content_tag` instead of interpolating the element type
 - Switched to a nice and more readable way of checking that the heading level is appropriate

## Why

If used as a heading the organisation logo would inherit the default margin set by the browser - setting it to zero avoids this inheritance.

Previously the outer element was either a `div` or heading element - and this was interpolated (for example `<<% element_type %>>` and `</<% element_type %>>`) - using `content_tag` instead makes this component more readable.

## Visual Changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_organisation_logo.html.erb
+++ b/app/views/govuk_publishing_components/components/_organisation_logo.html.erb
@@ -6,16 +6,13 @@
   heading_level ||= false
   inline ||= false
 
-  # Check if `heading_level` is a number; if so, check if it's an appropriate
-  # number.
-  use_heading = heading_level.is_a?(Integer) ?
-                  (heading_level >= 1 && heading_level <= 6) :
-                  false
+  # Check if `heading_level` is an appropriate number:
+  use_heading = [*1..6].include?(heading_level)
 
-  # Set the wrapping element to be a heading or a `div`
-  wrapping_element = use_heading ? "h#{heading_level}" : "div"
+  # Set the wrapping element to be a heading or a `div`:
+  wrapping_element = (use_heading ? "h#{heading_level}" : "div").to_sym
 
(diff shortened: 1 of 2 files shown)


app/views/govuk_publishing_components/components/_organisation_logo.html.erb
```diff
@@ -6,16 +6,13 @@
   heading_level ||= false
   inline ||= false
 
-  # Check if `heading_level` is a number; if so, check if it's an appropriate
-  # number.
-  use_heading = heading_level.is_a?(Integer) ?
-                  (heading_level >= 1 && heading_level <= 6) :
-                  false
+  # Check if `heading_level` is an appropriate number:
+  use_heading = [*1..6].include?(heading_level)
```
owenatgov, line 10: Very posh 👍

app/views/govuk_publishing_components/components/_organisation_logo.html.erb
```diff
@@ -24,10 +21,13 @@
   ]
   container_classes << "gem-c-organisation-logo__container--inline" if inline
 %>
-<<%= wrapping_element %>
-  class="<%= wrapper_classes.join(" ") %>"
-  <%= "data-module=gem-track-click" if organisation[:data_attributes] %>
->
+
+<%= content_tag(wrapping_element, {
```
owenatgov, line 25: Much tidier. It took me a moment to understand what was going on pre-change.

owenatgov (approved):
Just needs a changelog update and then I'm happy for this to go in!

PR #2112, opened 2021-06-03
Rename /transition to /brexit

sihugh:
We renamed the taxon today to reflect the language that people use. There is a redirect in place so the links will all still work, but the scroll tracking is currently broken because it's fixed to the URL not the content ID.

https://trello.com/c/wxR5o06H/1539-implement-brexit-landing-page-url-change

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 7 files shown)


alex-ju (approved):

alex-ju:
We have it [here too](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/public_layout_helper.rb#L27), for the new layout.

PR #2114, opened 2021-06-04
Allow a custom link to be used on the header logo

injms:
## What

Adds a `logo_link` parameter to the header and the public layout components to allow the link to be changed. The link will default to the previous behaviour of being a forward slash.

## Why

The ability to change the logo link was required to match the behaviour of the current `core_layout` in static.

## Visual Changes

None.


chao-xian (approved):
LGTM

PR #2115, opened 2021-06-04
Add alternate Brexit callout text and link

sihugh:
Probably worth viewing per commit because one of them just adds a tonne of empty locale file entries.

## What
Add alternate Brexit callout text and link to the Brexit callout sidebar element. (No changes to layout or styles)

We're only adding this behaviour to English hub pages at the moment because we don't have translations to other languages yet.

## Why

This text will be shown on the new Brexit hub pages to reduce the chance of circular journeys. Without this, people may land on the campaign page, navigate to a hub page, and then navigate _back_ to the campaign page by clicking on the call to action.

### This will be visible on the two tagged pages

Screenshot 2021-06-04 at 12 38 35

This change means that people on a hub page will be directed to the checker instead.

https://trello.com/c/JcdFcuZN/1557-brexit-sidebar-nav-box-for-new-business-and-citizen-hub-pages


alex-ju (approved):
Looks good! 🙌  I learned about the i18n tasks with this PR – a bit strange that we need all these empty keys in there, but at least there's an automated way to generate them.

PR #2117, opened 2021-06-07
Group tracking scripts in `govuk_publishing_components/analytics`

alex-ju:
## What
Group tracking scripts in `govuk_publishing_components/analytics`

## Why
To follow convention and avoid importing `track-click` script twice (once via `lib` at the applications level and once in [`static`](https://github.com/alphagov/static/blob/main/app/assets/javascripts/header-footer-only.js#L5))

## Visual Changes
No visual changes.

## Notes
This is a breaking change for `static` as it's the only application explicitly referencing `govuk_publishing_components/lib/track-click`. To update, [L5](https://github.com/alphagov/static/blob/main/app/assets/javascripts/header-footer-only.js#L5) needs to be removed, relying on [L2](https://github.com/alphagov/static/blob/main/app/assets/javascripts/header-footer-only.js#L2) which import the analytics bundle.

The `track-select-change` was previously made available to public-facing frontend apps via `lib`, now the script will be available via `static`. To avoid any data loss we shall be updating `static` first to the latest release, to make sure the script is available in all apps, then the applications.


andysellick (approved):
👏  for anything that starts to clean up the analytics scripts 😁

PR #2135, opened 2021-06-15
Add real user metrics using Speedcurve's LUX

injms:
## What

Adds the real user metrics to the public layout component by adding the LUX scripts. This includes:

 - Adding tests to ensure both LUX scripts are present in the public layout component _only_ when a user has allowed usage tracking cookies
 - Adding `lux` and `lux-polyfill` to the assets precompile list
 - Updating the LUX script to ensure that the image beacon is used
 - Updating the LUX script to add GDS's Speedcurve user ID number - usually this is appended to the script `src` with a query string, but that doesn't work in this instance as Rails adds a fingerprint to the filename that prevents the script from finding itself
 - Prettifying `lux-polyfill` to make it slightly easier to see what's going on; this is compressed when going through the asset pipeline
 - Adding documentation explaining RUM and how to update the LUX scripts

## Why

Real user metrics will give a much greater insight into the performance of GOV.UK for everyone. This will show in greater detail what parts of GOV.UK can be improved without using synthetic tests and what steps can be taken to improve the experience for everyone.

This is a sibling pull request to https://github.com/alphagov/static/pull/2445 - that covers the `core_layout` template; this pull request ensures that the `gem_layout` in static has feature parity.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
+++ b/app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
@@ -1,0 +1,36 @@
+(function() {
+  var marker = document.getElementsByTagName('script')[0]
+  var scripts = [
+    '<%= path_to_javascript("govuk_publishing_components/vendor/lux/lux-polyfill") %>',
+    '<%= path_to_javascript("govuk_publishing_components/vendor/lux/lux") %>'
+  ]
+  var parsedCookie = (function() {
+    try {
+      var cookies = document.cookie.split(';')
+
+      for (var i = 0; i < cookies.length; i++) {
+        var cookie = cookies[i].replace(/\s/g, '') // Remove all whitespace.
+        var cookieName = 'cookies_policy='
+
+        if (cookie.indexOf(cookieName) === 0) {
+          return JSON.parse(cookie.substring(cookieName.length))
+        }
+      }
+    } catch (error) {
+      console.error(error)
+    }
+
+    return {}
+  })()
+
+  if (parsedCookie.usage === true) {
+    for (var i = 0; i < scripts.length; i++) {
+      var script = document.createElement('script')
+      script.src = scripts[i]
+      script.async = true
+      script.defer = true
+
+      marker.parentNode.insertBefore(script, marker)
+    }
+  }
+})()
--- a/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-polyfill.js
+++ b/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-polyfill.js
@@ -1,0 +1,152 @@
+/* eslint-disable */
+
+LUX = (function () {
+  var a =
+    "undefined" !== typeof LUX && "undefined" !== typeof LUX.gaMarks
+      ? LUX.gaMarks
+      : [];
+  var d =
+    "undefined" !== typeof LUX && "undefined" !== typeof LUX.gaMeasures
+      ? LUX.gaMeasures
+      : [];
+  var j = "LUX_start";
+  var k = window.performance;
+  var l =
+    "undefined" !== typeof LUX && LUX.ns
+      ? LUX.ns
+      : Date.now
+      ? Date.now()
+      : +new Date();
+  if (k && k.timing && k.timing.navigationStart) {
+    l = k.timing.navigationStart;
+  }
+  function f() {
+    if (k && k.now) {
+      return k.now();
+    }
+    var o = Date.now ? Date.now() : +new Date();
+    return o - l;
+  }
+  function b(n) {
+    if (k) {
+      if (k.mark) {
+        return k.mark(n);
+      } else {
+        if (k.webkitMark) {
+          return k.webkitMark(n);
+        }
+      }
+    }
+    a.push({ name: n, entryType: "mark", startTime: f(), duration: 0 });
+    return;
+  }
+  function m(p, t, n) {
+    if ("undefined" === typeof t && h(j)) {
+      t = j;
+    }
+    if (k) {
+      if (k.measure) {
+        if (t) {
+          if (n) {
+            return k.measure(p, t, n);
+          } else {
+            return k.measure(p, t);
+          }
+        } else {
+          return k.measure(p);
+        }
+      } else {
+        if (k.webkitMeasure) {
+          return k.webkitMeasure(p, t, n);
+        }
+      }
+    }
+    var r = 0,
+      o = f();
+    if (t) {
+      var s = h(t);
+      if (s) {
+        r = s.startTime;
+      } else {
+        if (k && k.timing && k.timing[t]) {
+          r = k.timing[t] - k.timing.navigationStart;
+        } else {
+          return;
+        }
+      }
+    }
+    if (n) {
+      var q = h(n);
+      if (q) {
+        o = q.startTime;
+      } else {
+        if (k && k.timing && k.timing[n]) {
+          o = k.timing[n] - k.timing.navigationStart;
+        } else {
+          return;
+        }
+      }
+    }
+    d.push({ name: p, entryType: "measure", startTime: r, duration: o - r });
+    return;
+  }
+  function h(n) {
+    return c(n, g());
+  }
+  function c(p, o) {
+    for (i = o.length - 1; i >= 0; i--) {
+      var n = o[i];
+      if (p === n.name) {
+        return n;
+      }
+    }
+    return undefined;
+  }
+  function g() {
+    if (k) {
+      if (k.getEntriesByType) {
+        return k.getEntriesByType("mark");
+      } else {
+        if (k.webkitGetEntriesByType) {
+          return k.webkitGetEntriesByType("mark");
+        }
+      }
+    }
+    return a;
+  }
+  return { mark: b, measure: m, gaMarks: a, gaMeasures: d };
+})();
+LUX.ns = Date.now ? Date.now() : +new Date();
+LUX.ac = [];
+LUX.cmd = function (a) {
+  LUX.ac.push(a);
+};
+LUX.init = function () {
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
```diff
@@ -0,0 +1,36 @@
+(function() {
+  var marker = document.getElementsByTagName('script')[0]
+  var scripts = [
+    '<%= path_to_javascript("govuk_publishing_components/vendor/lux/lux-polyfill") %>',
+    '<%= path_to_javascript("govuk_publishing_components/vendor/lux/lux") %>'
+  ]
+  var parsedCookie = (function() {
```
andysellick, line 7: Can we use any of the existing cookie functions code for any of this or does it not offer this functionality?

app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
```diff
@@ -0,0 +1,36 @@
...
+      for (var i = 0; i < cookies.length; i++) {
+        var cookie = cookies[i].replace(/\s/g, '') // Remove all whitespace.
+        var cookieName = 'cookies_policy='
+
+        if (cookie.indexOf(cookieName) === 0) {
+          return JSON.parse(cookie.substring(cookieName.length))
+        }
+      }
+    } catch (error) {
+      console.error(error)
```
andysellick, line 20: Probably out of scope for this PR, but I wonder if there's something better we can do in the event of an error, like firing a specific GA event? Would let us have some awareness of problems occurring.

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-polyfill.js
```diff
@@ -0,0 +1,152 @@
+/* eslint-disable */
```
andysellick, line 1: Is it worth putting a comment at the top of this file (and the other Lux file) to explain where it's from and how to get the latest version? I can imagine having a problem at some point in the future and someone trying to fix this, not knowing that it's not something we've written (I guess it is in `vendor` though).

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux.js
```diff
@@ -0,0 +1,835 @@
...
+LUX.customerid = 47044334;
+
+// Turn on the image-based beacon, rather than have a remote JavaScript file
+// fetched and executed.
+LUX.beaconMode = "simple";
+
+// Setting debug to `true` shows what happening as it happens. Running
+// `LUX.getDebug()` in the browser's console will show the history of what's
+// happened.
+LUX.debug = false;
```
andysellick, line 835: This adds quite a lot of JS. Do we know exactly what the file size impact of this change will be?

docs/real-user-metrics.md
```diff
@@ -0,0 +1,60 @@
...
+
+The benefit of RUM is that it shows how pages are performing in real situations, rather than on synthetic tests. This means that we don't have to guess at what conditions might be - RUM provides insight into what conditions actually are.
+
+## Loading
+
+The LUX scripts should only be loaded when a user has opted into usage tracking. This is done using the `rum-loader` script, which checks if the `usage` cookie is `true` and then loads the required scripts. This is already part of the [Public Layout component].
+
+## Custom changes when updating LUX
+
+The scripts for the real user metrics are loaded from our servers - this allows us to know what is contained within the scripts and reduce the risk of unwelcome things being added to GOV.UK
```
andysellick, line 13: Nitpick: full stop at the end of the line.

docs/real-user-metrics.md
```diff
@@ -0,0 +1,60 @@
...
+
+Because of this, the scripts are audited before being updated and then two extra lines are added to make sure it works correctly. These two lines set the customer ID and turn on beacon mode.
```
andysellick, line 15: Can we say anything more specific about this audit and what is involved?

docs/real-user-metrics.md
```diff
@@ -0,0 +1,60 @@
...
+
+### Customer ID
+
+When loading `lux.js` from Speedcurve's CDN, the customer ID is appended to the end of the URI as a query string. The script looks for a script in the DOM with a source of `lux.js`, and from that extracts the customer ID.
```
andysellick, line 19: Just to clarify - this is GOV.UK's customer ID, not something that identifies a user? I think this might be worth clarifying.

docs/real-user-metrics.md
```diff
@@ -0,0 +1,60 @@
...
+
+```javascript
+LUX.getDebug()
+```
+
+[Performance API]: https://developer.mozilla.org/en-US/docs/Web/API/Performance_API
+
+[LUX]: https://speedcurve.com/features/lux/
+
+[Public Layout component]: components.publishing.service.gov.uk/component-guide/layout_for_public
```
andysellick, line 58: I think this needs `https://` on the front or github assumes it's a relative link.

spec/components/layout_for_public_spec.rb
```diff
@@ -96,4 +96,59 @@ def component_name
...
+    Capybara.current_driver = Capybara.javascript_driver
+
+    visit "/public"
+
+    assert page.has_selector?("html > head > script[src*='rum-loader']", visible: :all)
+    assert page.has_no_selector?("html > head > script[src*='lux/lux']", visible: :all)
+    assert page.has_no_selector?("html > head > script[src*='lux/lux-polyfill']", visible: :all)
+  end
+
+  it "does not contain real user metrics scripts on the page where cookies are accepted" do
```
andysellick, line 115: Looks like this is testing that the scripts aren't immediately loaded when the user accepts cookies. Just wanted to highlight some recent changes to the gem that allow for JS like this to be executed as soon as cookies are accepted, if that could be useful here: https://github.com/alphagov/govuk_publishing_components/blob/master/docs/javascript-modules.md#modules-and-cookie-consent

docs/real-user-metrics.md
injms: Yes, this is the ID of us as a customer of Speedcurve - not the ID of the user visiting GOV.UK. I'll amend to emphasise this.

app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
injms: This needs to run as quickly as possible without waiting for anything else to load - parsing the cookie here means no dependencies to wait for. 

The cookie helper functions are bundled into `application.js`, which is loaded just before the closing body tag - whilst we could move the specific cookie helper that reads cookies out of that bundle, it just felt as if it was asking for trouble further down the line.

app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
injms: We can and I'll do some digging and see if this is feasible - but, as you say, out of scope for the moment. I think frontend error reporting might be a too big an issue to be solved here!

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-polyfill.js
injms: This script shouldn't change too often if at all - it's the equivalent of the Google Analytic's snippet that goes in the pages you want to track. 

The `lux.js` might change regularly - and has a version variable near the [top of the script](https://github.com/alphagov/govuk_publishing_components/blob/51fd6a286b2745f70cb061d9821aa7294e9206ac/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux.js#L17).

I'll add a comment at the top of both scripts explaining what they're for, and pointing at the docs.

docs/real-user-metrics.md
injms: Fixed.

docs/real-user-metrics.md
injms: Fixed - thank you!

spec/components/layout_for_public_spec.rb
injms: This is on purpose - I couldn't tell what would happen to the metrics if a user accepts cookies a while after loading a page. I then opted to only run this on the page visited after a user has accepted cookies rather than possibly get spurious data.

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux.js
injms: So the file sizes are:

| File | `rum-loader` | `lux-polyfill` | `lux` |
| - | - | - | - |
| Uncompressed | 1.5kb | 3.5kb | 33.5kb |
| Minified | 1.1kb | 2.3kb | 21.8kb | 

So the total minified is 25.2kb (and could be smaller when gzipped, but I can't find a way of turning that on locally).

All of which is loaded using the `async` attribute - so it won't block any loading or rendering of the page.

docs/real-user-metrics.md
injms: I'll ask - and update the docs when I hear back. This shouldn't be a blocker for this pull request.

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux.js
andysellick: Good numbers, and good to see you've got the stats. Worth adding to the docs?

docs/real-user-metrics.md
andysellick: Fair enough. Just so long as we don't forget.

spec/components/layout_for_public_spec.rb
andysellick: Okay, that sounds like a sensible approach.

andysellick (approved):
Looks good.

PR #2141, opened 2021-06-15
Update header tracking to cover for menu and search button toggles

chao-xian:
It wasn't clear that the toggle was affecting both buttons, and previously the GA values were the same for both when they should be distinct.

## What
Fixes to header menu tracking events.

## Why
As part of gathering some baseline data ahead of updating the header design.

[Trello](https://trello.com/c/CwPwNE60/321-implement-analytics-tracking-on-the-existing-headers)

## Visual Changes
None

### DIFF
--- a/app/views/govuk_publishing_components/components/layout_header/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_search.html.erb
@@ -18,5 +18,10 @@
     id: "site-search-text",
     margin_bottom: 0,
     no_border: true,
+    data_attributes: {
+      track_category: "headerClicked",
+      track_action: "searchSubmitted",
+      track_label: "none",
+    },
   } %>
 </form>


app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
```diff
@@ -14,31 +20,42 @@
...
-          if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
-            window.GOVUK.analytics.trackEvent('headerClicked', 'menuOpened', { label: 'none' })
+          if (isTargetSearchToggle) {
+            sendClick('searchOpened')
+          } else {
+            sendClick('menuOpened')
           }
         }
         if (sourceClass.indexOf('js-visible') !== -1) {
           this.setAttribute('class', sourceClass.replace(/(^|\s)js-visible(\s|$)/, ''))
```
injms, line 45: On a header with both a menu and search, I'm seeing both `targetClass` and `sourceClass` being true when I click on the search toggle - and both false when clicking on the menu toggle. Going to dig a bit and see if I can find out why.

chao-xian:
This was superseded by https://github.com/alphagov/govuk_publishing_components/pull/2143

PR #2146, opened 2021-06-17
Release v24.15.0

injms:
Includes:

 * Set LUX's sampling rate to 1% ([PR 2145](https://github.com/alphagov/govuk_publishing_components/pull/2145))


andysellick (approved):
Does this need a minor release? Seems like a very small change.

injms:
> Does this need a minor release? Seems like a very small change.

I think it does as it changes the way LUX works; and it's not a patch to fix something that's broken.

PR #2154, opened 2021-06-21
Update LUX to activate when usage cookies are accepted

injms:
## What

Updates LUX loading strategy to measure (but not report) the page experience of users that haven't yet accepted cookies.

LUX sets a cookie so that users metrics can be seen from page to page - [this requires LUX to set a cookie][1]. To ensure that this wasn't tracking users that opted out, both the LUX JavaScript "snippet" and the reporting JavaScript were loaded on the page after a user had accepted cookies.

The LUX snippet is loaded as a JavaScript file as GOV.UK content security policy doesn't allow inline JavaScript. This is loaded first, and measures the user's experience - and then does nothing else.

The `rum-loader.js` checks a users preferences, and will only load `lux-reporter.js` if a user has opted in. If either the `cookie-consent` event has been fired or the usage tracking cookie was set, the `lux-reporter.js` script is loaded. 

The `lux-reporter.js` script then sends the measurements off to be analysed; since this JavaScript sets cookies, this file is only loaded if a user has allowed usage cookies.

Other changes worth noting:
 - LUX "snippet" is in the `lux-measurer.js` file
 - The self hosted `lux.js` (usually loaded from Speedcurve's CDN) is in the `lux-reporter.js` file
 - Tests added and updated to check for reporting when a user has clicked on the cookie banner

Will help with https://github.com/alphagov/static/issues/2516

[1]: https://support.speedcurve.com/en/articles/4468508-using-lux-with-cookie-consent-banners

## Why

This change ensures that the user's privacy preferences are respected _and_ the crucial first page load on a cold cache can be accurately reported. 

## Visual Changes

None


app/assets/javascripts/govuk_publishing_components/rum-loader.js.erb
```diff
@@ -23,14 +20,22 @@
...
+    script.async = true
+    script.defer = true
+
+    marker.parentNode.insertBefore(script, marker)
+  }
+
+  if (parsedCookie.usage === true) {
+    insertScript(scriptSrc)
+  } else {
+    window.addEventListener('cookie-consent', function() {
```
alex-ju, line 37: 💯

alex-ju (approved):
Excellent work on this, Ian!

PR #2157, opened 2021-06-22
Add tracking on search submit to public layout

chao-xian:
## What
Previous data tracking events added to the search button weren't included in the new public layout template as they didn't need to until frontend applications began using the newer template in production. Well they do now!

## Why

This is part of work to baseline data of usage of navigation elements before a redesign.

Screenshot 2021-06-22 at 13 15 23

## Visual Changes
None

### DIFF
--- a/app/views/govuk_publishing_components/components/layout_header/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_search.html.erb
@@ -18,5 +18,10 @@
     id: "site-search-text",
     margin_bottom: 0,
     no_border: true,
+    data_attributes: {
+      track_category: "headerClicked",
+      track_action: "searchSubmitted",
+      track_label: "none",
+    },
   } %>
 </form>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add tracking on search submit to public layout ([PR #2157](https://github.com/alphagov/govuk_publishing_components/pull/2157))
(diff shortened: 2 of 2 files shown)


app/views/govuk_publishing_components/components/layout_header/_search.html.erb
```diff
@@ -18,5 +18,10 @@
     id: "site-search-text",
     margin_bottom: 0,
     no_border: true,
+    data_attributes: {
+      track_category: "headerClicked",
+      track_action: "searchSubmitted",
+      track_label: "none",
```
DilwoarH, line 24: will this work without a module?

injms: Agreed, I think it needs:

```ruby
data_module: "gem-track-click",
```
to make sure that these attributes are used.

injms: Wait, scrub that - the `data-module` is added automatically in the search component itself: https://github.com/alphagov/govuk_publishing_components/blob/a341111172fb799fa79a99b6cc65425f26d98b77/app/views/govuk_publishing_components/components/_search.html.erb#L14

chao-xian: Default is set to https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_search.html.erb#L14

chao-xian: It's making me think though - is it a good idea to have a default module over there?

injms (approved):

PR #2160, opened 2021-06-23
Amend layout_for_public component

danacotoran:
Adjust `layout_for_public` component to accept an optional `product_name` parameter, as well as an option to hide the feedback component. 

Both of these features are needed for the account homepage which is being moved over from `govuk-account-manager-prototype` to `frontend`: https://github.com/alphagov/frontend/compare/msw/account-home-page.

The account manager uses a different site header from the rest of GOVUK. Some main differences are:

- the account related navigation links in the header
- the “Account” product name next to the home link/logo
- the home link directs the user back to the account homepage (as opposed to other pages on GOVUK where the home link will link back to the GOVUK homepage)
- account pages also don't use the side-wide feedback component

This was not an issue on `govuk-account-manager-prototype` as it has its own layout and does not use `static` or `slimmer`, so we simply called the components needed with the necessary parameters on our application file. 

The `frontend` app on the other hand does use `static`, and the default layout used is vastly different from the layout required by accounts pages. 
A possible solution to this problem is the introduction of a [special layout template for accounts](https://github.com/alphagov/static/pull/2527) in `static` which is to be requested only by the account controller. These changes are here to support the introduction of the new template.

-----
https://trello.com/c/lIDQNATi/813-implement-the-govuk-account-home-page

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Allow custom font size on action links ([PR #2162](https://github.com/alphagov/govuk_publishing_components/pull/2162))
 * Add scroll tracking to /guidance/import-and-export-goods-using-preference-agreements ([PR #2165](https://github.com/alphagov/govuk_publishing_components/pull/2165))
+* Amend layout_for_public component ([PR #2160](https://github.com/alphagov/govuk_publishing_components/pull/2160))
 
 ## 24.15.3
 
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -1,7 +1,9 @@
 <%
+  omit_feedback_form ||= false
   emergency_banner ||= nil
   full_width ||= false
   global_bar ||= nil
+  product_name ||= nil
   html_lang ||= "en"
   layout_helper = GovukPublishingComponents::Presenters::PublicLayoutHelper.new(local_assigns)
   logo_link ||= "/"
@@ -76,6 +78,7 @@
         search: show_search,
         logo_link: logo_link,
         navigation_items: navigation_items,
+        product_name: product_name,
 
         # The (blue) bottom border needs to be underneath the emergency banner -
         # so it has been turned off and added in manually.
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -1,7 +1,9 @@
 <%
+  hide_feedback_form ||= false
```
injms, line 2: Could this be named `omit_feedback_form` to keep it consistent with `omit_header`?

spec/components/layout_for_public_spec.rb
```diff
@@ -47,6 +47,18 @@ def component_name
     assert_select ".gem-c-layout-for-public .gem-c-layout-header", false
   end
 
+  it "can omit the feedback component" do
+    render_component(hide_feedback_form: true)
+
+    assert_select ".gem-c-layout-for-public .gem-c-feedback", false
+  end
```
injms, line 54: Could a test be added for when `hide_feedback_form` is false and/or not present?

spec/components/layout_for_public_spec.rb
```diff
@@ -47,6 +47,18 @@ def component_name
...
+
+  it "can add a product name in the header" do
+    render_component(product_name: "Account")
+
+    assert_select ".gem-c-layout-for-public .gem-c-header__product-name", text: "Account"
+  end
```
injms, line 60: This is the same as a test for the header component - do you think it's needed for the public layout component as well?

https://github.com/alphagov/govuk_publishing_components/blob/1dee4ae94c9187455bd5b5cf714e4bf5c0ef219f/spec/components/layout_header_spec.rb#L46-L50

danacotoran: Yeah I think my intention with this test is to verify that `product_name` is correctly passed through from `layout_for_public` to `layout_header`.

spec/components/layout_for_public_spec.rb
danacotoran: Done!

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Good point!

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Allow custom font size on action links ([PR #2162](https://github.com/alphagov/govuk_publishing_components/pull/2162))
+* Amend layout_for_public component ([PR #2160](https://github.com/alphagov/govuk_publishing_components/pull/2160))
```
andysellick, line 13: You might as well merge this into the only other commit, since it relates directly to it.

andysellick (approved):
Looks good 👍

PR #2170, opened 2021-06-28
Allow multiple JavaScript modules on a certain element

alex-ju:
## What
Allow multiple JavaScript modules on a certain element.

## Why
To allow components using scripts from both `govuk-frontend` and `govuk_publishing_components` to work as intended. This fixes a bug where the checkboxes and details components only load the `govuk-frontend` scripts as a result of [converging to a single module system](https://github.com/alphagov/govuk_publishing_components/pull/2095).

This PR proposes a uniform notation between the `data-module` values (e.g. "govuk-checkboxes" or "gem-checkboxes") and the JavaScript module names (`GovukCheckboxes` and `GemCheckboxes`) with the possibility to apply multiple modules on a certain element. This makes the modules system a bit less 'magical' in the sense that its behaviour is more predictable instead of trying to automatically initialise multiple modules based on a single `data-module` declaration.

## Visual Changes
No visual changes.

### DIFF
--- a/app/views/govuk_publishing_components/components/_details.html.erb
+++ b/app/views/govuk_publishing_components/components/_details.html.erb
@@ -7,7 +7,7 @@
   css_classes << (shared_helper.get_margin_bottom)
 
   data_attributes ||= {}
-  data_attributes[:module] = 'govuk-details'
+  data_attributes[:module] = 'govuk-details gem-details'
 %>
 <%= tag.details class: css_classes, data: data_attributes, open: open do %>
   <summary class="govuk-details__summary" data-details-track-click>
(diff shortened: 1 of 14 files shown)


app/views/govuk_publishing_components/components/_details.html.erb
```diff
@@ -7,7 +7,7 @@
   css_classes << (shared_helper.get_margin_bottom)
 
   data_attributes ||= {}
-  data_attributes[:module] = 'govuk-details'
+  data_attributes[:module] = 'govuk-details gem-details'
```
theseanything, line 10: Nit: Should this be plural "modules", to indicate you can specify more than one?

theseanything (approved):
Nice clean up of the module names!!

app/views/govuk_publishing_components/components/_details.html.erb
alex-ju: That's a good point, I'm not 100% confident of the full implications of renaming `data-module` to `data-modules`. Would it be ok with you if I record this as an issue to investigate further?

theseanything: Sure thing! Not a blocker for this PR.

PR #2175, opened 2021-07-01
Add margin bottom to radio component

andysellick:
## What
Adds the margin_bottom option to the radio component, using the shared helper.

Also expands the documentation for the shared helper, because I had to look through code to remember how to do this.

## Why
Need this option for the radio component on the coronavirus landing page.

## Visual Changes

Screenshot 2021-07-01 at 09 08 37

Trello card: https://trello.com/c/eE62ythU/447-coronavirus-timeline-frontend

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add margin bottom to radio component ([PR #2175](https://github.com/alphagov/govuk_publishing_components/pull/2175))
+
 ## 24.17.0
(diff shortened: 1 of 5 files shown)


alex-ju (approved):
In-line with how we do margins on other components. The shared helper docs might be a bit too prescriptive (concerned about it possibly being out of date in future).

andysellick:
> The shared helper docs might be a bit too prescriptive (concerned about it possibly being out of date in future).

That's fair, although I think we generally don't do enough to maintain our documentation right now. I'm happy to include more detail like this and own it, in the hope that it encourages more use of the helper.

PR #2191, opened 2021-07-06
Switch environment app helper to use GOVUK_ENVIRONMENT_NAME

karlbaker02:
This PR switches the `Environment` app helper to use the `GOVUK_ENVIRONMENT_NAME` when getting the current environment for use in the layout header component. The reason for this change is that we're removing the `ERRBIT_ENVIRONMENT_NAME` environment variable in favour of `GOVUK_ENVIRONMENT_NAME`, with the former being an overhang from the days when GOV.UK used Errbit. The `GOVUK_ENVIRONMENTS` variable is refactored from a hash to an array, as we no longer have a weird edge case where the name of the integration environment is different to the acceptance environment that the method returns; there will be a performance impact with this as we're no longer doing a constant-time lookup, but given the number of environments this should be negligible.

It also refactors the associated tests and introduces coverage for the Heroku environment.

Trello: https://trello.com/c/0Z2F9lKU

### DIFF
--- a/spec/lib/govuk_publishing_components/app_helpers/environment_spec.rb
+++ b/spec/lib/govuk_publishing_components/app_helpers/environment_spec.rb
@@ -7,16 +7,20 @@ RSpec.describe GovukPublishingComponents::AppHelpers::Environment do
       expect(environment_name_for("something-else")).to eql("development")
     end
 
-    it "returns integration as well" do
-      expect(environment_name_for("integration-blue-aws")).to eql("integration")
+    it "returns a listed environment when specified" do
+      expect(environment_name_for("integration")).to eql("integration")
+    end
+
+    it "returns example when the Heroku environment variable is set" do
+      expect(environment_name_for("heroku", environment_variable_name: "HEROKU")).to eql("example")
     end
   end
 
-  def environment_name_for(actual_environment_name)
-    previous = ENV["ERRBIT_ENVIRONMENT_NAME"]
-    ENV["ERRBIT_ENVIRONMENT_NAME"] = actual_environment_name
+  def environment_name_for(actual_environment_name, environment_variable_name: "GOVUK_ENVIRONMENT_NAME")
+    previous = ENV[environment_variable_name]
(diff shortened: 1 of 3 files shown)


spec/lib/govuk_publishing_components/app_helpers/environment_spec.rb
```diff
@@ -7,16 +7,20 @@
...
 
-    it "returns integration as well" do
-      expect(environment_name_for("integration-blue-aws")).to eql("integration")
+    it "returns a listed environment when specified" do
+      expect(environment_name_for("integration")).to eql("integration")
+    end
+
+    it "returns example when the Heroku environment variable is set" do
+      expect(environment_name_for("heroku", environment_variable_name: "HEROKU")).to eql("example")
     end
```
kevindew, line 16: Nice catch 👍

kevindew (approved):
Looks great.

alex-ju:
This seems to me like a breaking change, so I'll label it accordingly for us to bundle it in the next major release.

karlbaker02:
Thanks @alex-ju. Just one question on the breaking change, which is that the new environment variable has been deployed through Puppet to all environments, so should be accessible by all apps which would be affected. Are there any other apps I perhaps don't know about which would be affected by this? I'm also happy to slightly tweak this to make it a non breaking-change.

alex-ju:
> Thanks @alex-ju. Just one question on the breaking change, which is that the new environment variable has been deployed through Puppet to all environments, so should be accessible by all apps which would be affected. Are there any other apps I perhaps don't know about which would be affected by this? I'm also happy to slightly tweak this to make it a non breaking-change.

Nope, the variable's existence on different environments is my main concern. If you're confident you've covered that we could ship it as a patch.

karlbaker02:
Cool, in that case then there shouldn't be anything adversely affected by this (I'm currently deploying apps which don't use this gem but the deprecated govuk_admin_template gem, so if there is an issue with missing env vars I will soon come across it).

PR #2192, opened 2021-07-06
Remove ability to link context text on titles

maxgds:
## What
This removes the ability to add a link to the contextual text that sometimes appears above a title. 

## Why
In many cases where this was used there were already breadcrumbs serving the same purpose and meaning duplicate links. Additionally, the link was not obvious unless hovered meaning there was no realistic was for most users to know it was there. 

## Visual Changes

THIS IS A BREAKING CHANGE. If an app updates to this version with attempts to link the context then the context hash will be exposed:

Screenshot 2021-07-06 at 15 02 58

The following PRs cover removing all the examples I'm aware of and this version of the components should not be used until they are merged:
https://github.com/alphagov/government-frontend/pull/2117
https://github.com/alphagov/whitehall/pull/6120
https://github.com/alphagov/collections/pull/2463

### DIFF
--- a/app/views/govuk_publishing_components/components/_title.html.erb
+++ b/app/views/govuk_publishing_components/components/_title.html.erb
@@ -3,9 +3,6 @@
 
   context ||= false
   context_locale ||= false
-  context_text = context.is_a?(Hash) ? context[:text] : context
-  context_href = context.is_a?(Hash) ? context[:href] : false
-  context_data = context.is_a?(Hash) ? context[:data] : false
 
   inverse ||= false
   local_assigns[:margin_top] ||= 8
@@ -24,7 +21,7 @@
 <%= content_tag(:div, class: classes) do %>
   <% if context %>
     <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
-      <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
+      <%= context %>
     </span>
   <% end %>
   <h1 class="<%= heading_classes.join(" ") %>">
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@
 
 ## Unreleased
 
+* BREAKING: Remove ability to link contextual text on titles ([PR #2192](https://github.com/alphagov/govuk_publishing_components/pull/2192))
+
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_title.html.erb
```diff
@@ -24,7 +21,7 @@
 <%= content_tag(:div, class: classes) do %>
   <% if context %>
     <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
-      <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
+      <%= context %>
```
injms, line 24: To prevent a breaking change, could this be `context[:text]` with a warning saying that this will be deprecated in the future?

```suggestion
      <%= context.is_a?(Hash) ? context[:text] : context %>
```

maxgds: I considered that but it felt like leaving junk around to clear up later. I think I've captured all the places it's used and I'd rather be done with it in one go. I can be convinced if we have prior examples of handling change that way?

injms: That is a very good point. The prior example that comes to mind [is around Govspeak sanitisation][1] - but that warning message hung around for _years_, which I think makes a strong case that your method is better.

[1]: https://github.com/alphagov/govuk_publishing_components/pull/1632

injms (approved):
Looks good - looking forward to non-confusing light grey links being eradicated.

We could bundle this with #2117 and #2061 when it needs to be released.

owenatgov:
There are 2 docs bits that reference the contextual link functionality which would be good to get removed here:

- https://components.publishing.service.gov.uk/component-guide/title/with_context_link
- https://components.publishing.service.gov.uk/component-guide/title/in_html_publication_with_context_link

Also changelog 🌳

maxgds:
> There are 2 docs bits that reference the contextual link functionality which would be good to get removed here:
> 
> * https://components.publishing.service.gov.uk/component-guide/title/with_context_link
> * https://components.publishing.service.gov.uk/component-guide/title/in_html_publication_with_context_link
> 
> Also changelog 🌳

You're quite right - I reverted the docs to get my example screenshot of it broken, and totally failed to check in the changelog.

maxgds:
I've converted this to a draft as I'll be away a few days. I think I've caught all the things it would break but would rather do spot checks as we merge dependabots to be sure, so I'll pick it up again when I'm back.

PR #2195, opened 2021-07-07
Change tracking on details component

andysellick:
## What
- move the tracking attributes onto the summary element rather than the whole component
- this prevents collisions from occurring if we want to track links or other elements that appear inside the details element, using the gem-track-click script

## Why
I recently needed to track radio buttons inside a details component and the tracking was either not firing or firing twice, the second track being from the details component. This is partly because the `gem-track-click` tracking script attempts to traverse up the DOM if it can't find what it's trying to track, which was then colliding with the tracking on the details component... it was all very confusing. 

The solution is to move the tracking from the details component to the summary element within, which is actually the thing being tracked anyway.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/details.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/details.js
@@ -7,24 +7,26 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
 (function (Modules) {
   function GemDetails ($module) {
     this.$module = $module
-    this.customTrackLabel = this.$module.getAttribute('data-track-label')
+    this.$summary = this.$module.querySelector('.govuk-details__summary')
+
+    this.customTrackLabel = this.$summary.getAttribute('data-track-label')
     this.detailsClick = this.$module.querySelector('[data-details-track-click]')
   }
 
   GemDetails.prototype.init = function () {
     if (this.customTrackLabel) { // If a custom label has been provided, we can simply call the tracking module
       var trackDetails = new window.GOVUK.Modules.GemTrackClick()
-      trackDetails.start($(this.$module))
+      trackDetails.start($(this.$summary))
     } else if (this.detailsClick) { // If no custom label is set, we use the open/close status as the label
       this.detailsClick.addEventListener('click', function (event) {
-        this.trackDefault(this.$module)
+        this.trackDefault(this.$summary)
       }.bind(this))
     }
   }
 
   GemDetails.prototype.trackDefault = function (element) {
     if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
-      var componentStatus = (element.getAttribute('open') == null) ? 'open' : 'closed'
+      var componentStatus = (this.$module.getAttribute('open') == null) ? 'open' : 'closed'
       var trackCategory = element.getAttribute('data-track-category')
       var trackAction = element.getAttribute('data-track-action')
       var trackOptions = element.getAttribute('data-track-options')
--- a/app/views/govuk_publishing_components/components/_details.html.erb
+++ b/app/views/govuk_publishing_components/components/_details.html.erb
@@ -6,15 +6,18 @@
   css_classes = %w(gem-c-details govuk-details)
   css_classes << (shared_helper.get_margin_bottom)
 
+  details_data_attributes = {}
+  details_data_attributes[:module] = 'govuk-details gem-details'
+
   data_attributes ||= {}
-  data_attributes[:module] = 'govuk-details gem-details'
+  data_attributes[:details_track_click] = ''
 %>
-<%= tag.details class: css_classes, data: data_attributes, open: open do %>
-  <summary class="govuk-details__summary" data-details-track-click>
+<%= tag.details class: css_classes, data: details_data_attributes, open: open do %>
+  <%= tag.summary class: "govuk-details__summary", data: data_attributes do %>
     <span class="govuk-details__summary-text">
       <%= title %>
     </span>
-  </summary>
+  <% end %>
   <div class="govuk-details__text">
     <%= yield %>
   </div>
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -6,25 +6,31 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
 
 (function (Modules) {
   function GemDetails ($module) {
+    console.log('details')
```
injms, line 9: Do this logging still need to be here?
```suggestion
```

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -6,25 +6,31 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
...
-        this.trackDefault(this.$module)
+        this.trackDefault(this.$summary)
       }.bind(this))
     }
   }
 
   GemDetails.prototype.trackDefault = function (element) {
     if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
-      var componentStatus = (element.getAttribute('open') == null) ? 'open' : 'closed'
+      var componentStatus = (this.$module.getAttribute('open') == null) ? 'open' : 'closed'
```
injms, line 33: Welp, this appears to make no sense at first look but actually _does_ make sense because the `open` attribute is blank so returns `null` when not present and `""` when present... 🤯

app/views/govuk_publishing_components/components/_details.html.erb
```diff
@@ -6,15 +6,18 @@
...
+  details_data_attributes[:module] = 'govuk-details gem-details'
+
   data_attributes ||= {}
-  data_attributes[:module] = 'govuk-details gem-details'
+  data_attributes[:details_track_click] = ''
 %>
-<%= tag.details class: css_classes, data: data_attributes, open: open do %>
-  <summary class="govuk-details__summary" data-details-track-click>
+<%= tag.details class: css_classes, data: details_data_attributes, open: open do %>
+  <%= tag.summary class: "govuk-details__summary", data: data_attributes do %>
```
injms, line 16: If we've got `details_data_attributes`, might it be worth renaming `data_attributes` to `summary_data_attributes`?
```suggestion
  <%= tag.summary class: "govuk-details__summary", data: summary_data_attributes do %>
```

injms:
Looks good - just had a couple of questions, and there's some logging to remove - and then this should be ready to go.

app/assets/javascripts/govuk_publishing_components/components/details.js
andysellick: Ugh, sorry, head cold. Removed.

app/assets/javascripts/govuk_publishing_components/components/details.js
andysellick: Yeah I didn't write this! However, things I learned today... the details element appends the 'open' attribute to itself by itself. Took me a few minutes to figure out how that was happening...

app/views/govuk_publishing_components/components/_details.html.erb
andysellick: I think I kept it like that because `data_attributes` is something that can be passed to the component, and changing it would mean a breaking change to the interface. But happy to be corrected if I've misunderstood.

injms: Oh yeah, that's correct 🤦 Best kept as `data_attributes` then.

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -7,24 +7,29 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
 (function (Modules) {
   function GemDetails ($module) {
     this.$module = $module
-    this.customTrackLabel = this.$module.getAttribute('data-track-label')
+    this.$summary = this.$module.querySelector('.govuk-details__summary')
+
+    this.customTrackLabel = this.$summary.getAttribute('data-track-label')
     this.detailsClick = this.$module.querySelector('[data-details-track-click]')
+    console.log(this.customTrackLabel, this.detailsClick)
```
injms, line 14:
```suggestion
```

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -7,24 +7,29 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
...
   }
 
   GemDetails.prototype.init = function () {
     if (this.customTrackLabel) { // If a custom label has been provided, we can simply call the tracking module
+      console.log('customTrackLabel')
```
injms, line 19:
```suggestion
```

app/assets/javascripts/govuk_publishing_components/components/details.js
```diff
@@ -7,24 +7,29 @@ window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
...
       var trackDetails = new window.GOVUK.Modules.GemTrackClick()
-      trackDetails.start($(this.$module))
+      trackDetails.start($(this.$summary))
     } else if (this.detailsClick) { // If no custom label is set, we use the open/close status as the label
+      console.log('detailsClick')
```
injms, line 23:
```suggestion
```

app/assets/javascripts/govuk_publishing_components/components/details.js
andysellick: Wait. So in removing one console.log, I managed to add three more? Wow I was not well yesterday afternoon. I am going to bed.

(fixed - hopefully)

injms: They were there - I just didn't highlight them, sorry!

injms (approved):

PR #2196, opened 2021-07-07
Intervention component

maxf:
## What

A new component called "intervention" which will be used for showing personalised content (based on previous browsing). The component's content is hardcoded for now, as this is an MVP which will be used only in the Start a Business area of gov.uk 

## Why

The data labs' Personalisation Without an Account project includes showing users recommended content, based on their behaviour on www.gov.uk

## Visual Changes

Visual design for new component
sab

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Switch environment app helper to use GOVUK_ENVIRONMENT_NAME ([PR #2191](https://github.com/alphagov/govuk_publishing_components/pull/2191)) PATCH
+* Add intervention component ([PR #2196](https://github.com/alphagov/govuk_publishing_components/pull/2196))
 * Fix misaligned crown in heading component properly ([PR #2206](https://github.com/alphagov/govuk_publishing_components/pull/2206)) PATCH
 
 ## 24.18.5
--- a/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
@@ -1,0 +1,16 @@
+.gem-c-intervention {
+  @include govuk-text-colour;
+  @include govuk-responsive-padding(4);
+  @include govuk-responsive-margin(4, "bottom");
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+}
+
+.gem-c-intervention__title {
+  @include govuk-font(24, $weight: bold);
+  margin-top: 0;
+  @include govuk-responsive-margin(4, "bottom");
+}
+
+.gem-c-intervention__paragraph {
+  @include govuk-font(19);
+}
--- a/app/views/govuk_publishing_components/component_guide/index.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/index.html.erb
@@ -105,5 +105,5 @@
 </ul>
 
 <div class="component-markdown">
-  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/develop-component.md">creating a new one</a>.</p>
+  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/generate-a-new-component.md">creating a new one</a>.</p>
 </div>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -1,0 +1,15 @@
+<%= tag.section class: "gem-c-intervention", role: "region" do %>
+  <h2 class="gem-c-intervention__title">
+    <a class="govuk-link" href="/next-steps-for-your-business">
+      <%= t("components.intervention.title") %>
+    </a>
+  </h2>
+
+  <p class="gem-c-intervention__paragraph">
+    <%= t("components.intervention.description") %>
+  </p>
+
+  <p class="gem-c-intervention__paragraph">
+    <%= t("components.intervention.dismiss_html") %>
+  </p>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/intervention.yml
+++ b/app/views/govuk_publishing_components/components/docs/intervention.yml
@@ -1,0 +1,24 @@
+name: Intervention
+description: An area that contains personalised content to the user
+body: |
+  The intervention is used to show personalised content. For instance, if the user has visited multiple
+  pages in the same area of the site, we might want let them know that there are other pages on GOV.UK
+  that would be useful to them. This component would be used to add this personalised content and would
+  indicate to the user that this is not normally part of the page, but has been added for them specifically.
+
+  Right now the contents of the component are static, as the MVP of personalised content is only for Start a Business.
+  Since many pages will use this component with the same text, we hard-code it here for now.
+
+  The dismiss link will reload the page but the `hide-intervention` query string parameter will cause the
+  backed not to show the intervention again. Some progressive enhancement will be added in later to avoid
+  reloading the page if JavaScript is available.
+accessibility_criteria: |
+  The intervention component must:
+
+  - have a border colour contrast ratio of more than 4.5:1 with its background to be visually distinct
+
+  - always render headings with associated description content, so there are no isolated heading elements inside the component
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -72,6 +72,10 @@ en:
       news_and_communications: News and communications
       statistics: Statistics
       worldwide: Worldwide
+    intervention:
+      title: Check the next steps for your limited company
+      description: You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.
+      dismiss_html: <a class="govuk-link" href="?hide-intervention=true">Hide this suggestion</a> if it’s not relevant to you
     layout_footer:
       copyright_html: <a class="govuk-footer__link govuk-footer__copyright-logo" href="http://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/copyright-and-re-use/crown-copyright/">© Crown copyright</a>
       licence_html: All content is available under the <a class="govuk-footer__link" href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence v3.0</a>, except where otherwise stated
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -42,6 +42,7 @@ $govuk-new-link-styles: true;
 @import "components/image-card";
 @import "components/input";
 @import "components/inset-text";
+@import "components/intervention";
 @import "components/inverse-header";
 @import "components/label";
 @import "components/layout-footer";
--- a/spec/components/intervention_spec.rb
+++ b/spec/components/intervention_spec.rb
@@ -1,0 +1,14 @@
+require "rails_helper"
+
+describe "Intervention", type: :view do
+  def component_name
+    "intervention"
+  end
+
+  it "renders the component" do
+    render_component({})
+    assert_select ".gem-c-intervention__title", text: "Check the next steps for your limited company"
+    assert_select ".gem-c-intervention__paragraph", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
+    assert_select ".gem-c-intervention__paragraph", text: "Hide this suggestion if it’s not relevant to you"
+  end
+end


app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -0,0 +1,18 @@
+name: Intervention
+description: An area that contains personalised content to the user
+body: |
+  The intervention is used to show personalised content. For instance, if the user has visited multiple
+  pages in the same area of the site, we might want let them know that there are other pages on gov.uk
```
injms, line 5:
```suggestion
  pages in the same area of the site, we might want let them know that there are other pages on GOV.UK
```

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -105,5 +105,5 @@
 </ul>
 
 <div class="component-markdown">
-  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/develop-component.md">creating a new one</a>.</p>
+  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="/alphagov/govuk_publishing_components/blob/master/docs/generate-a-new-component.md">creating a new one</a>.</p>
```
injms, line 108: Nice catch 👍

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
...
+  @include govuk-responsive-margin(8, "bottom");
+
+  clear: both;
+  background-color: govuk-colour("light-grey", $legacy: "grey-4");
+
+  .govuk-govspeak {
+    p:last-child {
+      margin-bottom: 0;
+    }
+  }
```
injms, line 13: I don't think that we should override the `govuk-govspeak` styles like this - it'd be better to add another class (example:`gem-c-intervention-govspeak`) and then use that to remove the margin. This way it's clearer where the styles are coming from.

Another thought is that I can't find where `govuk-govspeak` is used in the component - so is this still needed?

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  .govuk-body:last-child {
+    margin-bottom: 0;
+  }
```
injms, line 17: Similar to my other comment, I don't think that we should use `govuk-` prefixed classes to alter styles.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
...
+
+  &:last-child {
+    margin-bottom: 0;
+  }
+
+  a {
+    @include govuk-link-common;
+    @include govuk-link-style-default;
+    @include govuk-link-print-friendly;
+  }
```
injms, line 33: If the content is hardcoded at the moment, is it worth removing the `a` styles here and using `govuk-link` in the HTML?

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
+<%
+  local_assigns[:margin_bottom] ||= 8
+  local_assigns[:margin_bottom] = 8 if local_assigns[:margin_bottom] > 9
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
+  css_classes = %w[gem-c-intervention]
+  css_classes << (shared_helper.get_margin_bottom)
+
+  aria_attributes = {label: 'Intervention'}
```
injms, line 10: I don't think that the name "Intervention" is that useful - in VoiceOver it reads out "Intervention, Region".

I'd suggest removing the `aria-label` attribute completely; or using `aria-labelledby` to point to the heading inside the region; or replacing the content of `aria-label` with something that better explains the region such as "Suggested pages".

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
...
+
+  variant = request.headers["HTTP_GOVUK_ABTEST_STARTABUSINESSSEGMENT"]
+  is_sab_page = request.headers["HTTP_GOVUK_ABTEST_ISSTARTABUSINESSPAGE"] == "true"
+  should_show_intervention = variant == "B" && is_sab_page
```
injms, line 14: Having the A/B test logic inside the component feels a bit icky - could this logic be moved into the application with the component itself completely placed inside the `if` statement inside the application's view?

For example, in the application the logic could be:
```erb
  <%= render("govuk_publishing_components/components/intervention") if should_show_intervention %>
```

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
+.gem-c-intervention {
+  @include govuk-text-colour;
+  @include govuk-responsive-padding(4);
+  @include govuk-responsive-margin(8, "bottom");
+
+  clear: both;
```
andysellick, line 6: I can't currently see this rendered so the answer may be obvious, by why is `clear` needed on this element?

app/views/govuk_publishing_components/component_guide/index.html.erb
andysellick: Yes, but the link doesn't work, it still needs `https://github.com/` at the start.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
+<%
+  local_assigns[:margin_bottom] ||= 8
+  local_assigns[:margin_bottom] = 8 if local_assigns[:margin_bottom] > 9
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
+  css_classes = %w[gem-c-intervention]
+  css_classes << (shared_helper.get_margin_bottom)
```
andysellick, line 8: If this component uses the shared helper to provide an option for different bottom margins, that should be documented in the yml file.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  aria_attributes = {label: 'Intervention'}
+
+  variant = request.headers["HTTP_GOVUK_ABTEST_STARTABUSINESSSEGMENT"]
+  is_sab_page = request.headers["HTTP_GOVUK_ABTEST_ISSTARTABUSINESSPAGE"] == "true"
+  should_show_intervention = variant == "B" && is_sab_page
+%>
+<% if should_show_intervention %>
+  <%= tag.section class: css_classes, aria: aria_attributes, role: "region" do %>
+    <h2 class="gem-c-intervention__title">
+      <a href=\"https://www.gov.uk/next-steps-for-your-business\">Check the next steps for your limited company</a>
```
andysellick, line 19: Why are the quote marks around the href attribute escaped?

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
...
+    </h2>
+
+    <p class="gem-c-intervention__description"> You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.</p>
+  <% end %>
+<% end %>
```
andysellick, line 24: Nitpick: we normally leave a newline at the end of files.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
...
+    margin-bottom: 0;
+  }
+}
+
+.gem-c-intervention__title {
+  @include govuk-font(24, $weight: bold);
+  margin-top: 0;
+  @include govuk-responsive-margin(4, "bottom");
+
+  &:last-child {
```
andysellick, line 25: There's only one instance of this element in the component, so is the `last-child` selector needed?

app/views/govuk_publishing_components/components/_intervention.html.erb
andysellick, line 10: Is 'intervention' a word that users will understand? I'm asking because I'm not sure I understand it in this context.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  is_sab_page = request.headers["HTTP_GOVUK_ABTEST_ISSTARTABUSINESSPAGE"] == "true"
+  should_show_intervention = variant == "B" && is_sab_page
+%>
+<% if should_show_intervention %>
+  <%= tag.section class: css_classes, aria: aria_attributes, role: "region" do %>
+    <h2 class="gem-c-intervention__title">
+      <a href=\"https://www.gov.uk/next-steps-for-your-business\">Check the next steps for your limited company</a>
+    </h2>
+
+    <p class="gem-c-intervention__description"> You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.</p>
```
andysellick, line 22: Nitpick: unnecessary space at the start of the text.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,24 @@
...
+  <% end %>
```
andysellick, line 23: We normally put readable text into a translation file (so that it can be translated, and more easily updated by non-developers).

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,39 @@
+.gem-c-intervention {
+  @include govuk-text-colour;
+  @include govuk-responsive-padding(4);
+  @include govuk-responsive-margin(8, "bottom");
```
andysellick, line 4: I think you don't need to define margin here, because you're using the shared helper and it looks like it will default to returning the class for this much margin anyway?

andysellick (commented):
Had a quick look. Should it be possible to view the component in the preview app? My headers extension has gone awry on me.

app/views/govuk_publishing_components/components/_intervention.html.erb
oscarwyatt: We're running the A/B test across three different apps (and probably in multiple views of those different apps) so it seemed simpler to have the logic in one place, even if it is a bit unconventional - saves having three implementations and three sets of tests and the associated increase in maintenance/potential for problems

app/views/govuk_publishing_components/components/_intervention.html.erb
maxf: Good point. I'll remove the aria-label for now since the explanation is in the text already.

maxf: probably not, that's why I'm removing aria-label (see above)

app/views/govuk_publishing_components/components/_intervention.html.erb
injms: I'm going to disagree here. It's simpler to have the business logic in the controller of the application - all logic is in that one place in the application and not spread out over views, controllers, gems, and so on. Trying to trace exactly where a decision to show or hide a component becomes a lot harder if the logic sits outside of the application, and especially if it sits outside of Rails expected conventions.

I agree it's easier in the short term to drop this into the components gem, but I think it'll create more problems that it'll solve. The components gem is supposed to be for components - if you really need the logic in one place, then it shouldn't be in the components gem.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: None of the `govuk-` components are used at this point, so removing in the css

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: No, it's not obvious. I copied from `notice` but in hindsight I can't see why it's needed here. Removing.

app/views/govuk_publishing_components/components/_intervention.html.erb
maxf: Left over from copied element. Removing.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: removed

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: No longer using the shared helper

chao-xian:
@oscarwyatt @maxf Explore are running AB tests across 5 frontend apps and have put the logic into the apps, as is the standard as per documented in the AB test gem https://github.com/alphagov/govuk_ab_testing#usage

chao-xian (dismissed):
We shouldn't be putting logic for AB test in a component. Coupling logic of an AB test into a component does not align with component isolation.

app/views/govuk_publishing_components/components/_intervention.html.erb
maxf: Thanks, fair enough. We've now removed application logic from the component.

config/locales/en.yml
```diff
@@ -72,6 +72,10 @@ en:
       news_and_communications: News and communications
       statistics: Statistics
       worldwide: Worldwide
+    intervention:
+      title: Check the next steps for your limited company
+      description: You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.
+      dismiss_html: <a href="?hide-intervention=true">Hide this suggestion</a> if it’s not relevant to you
```
injms, line 78: I think this needs a `govuk-link` class to be styled correctly - it's appearing the browser default blue link at the moment.
```suggestion
      dismiss_html: <a class="govuk-link" href="?hide-intervention=true">Hide this suggestion</a> if it’s not relevant to you
```

app/views/govuk_publishing_components/component_guide/index.html.erb
chao-xian: As Andy mentioned, the link is still broken.

app/views/govuk_publishing_components/component_guide/index.html.erb
```diff
@@ -105,5 +105,5 @@
 </ul>
 
 <div class="component-markdown">
-  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="/alphagov/govuk_publishing_components/blob/master/docs/generate-a-new-component.md">creating a new one</a>.</p>
+  <p class="govuk-body">If you cannot find a suitable component consider extending an existing component or <a href="https://github.com/alphagov/govuk_publishing_components/blob/master/docs/generate-a-new-component.md">creating a new one</a>.</p>
```
chao-xian, line 108: Ah I see you fixed the documentation in this commit. A rebase to tidy this up into the previous commit?

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= tag.section class: "gem-c-intervention", role: "region" do %>
+  <h2 class="gem-c-intervention__title">
+    <a href="https://www.gov.uk/next-steps-for-your-business">
```
chao-xian, line 3: `.govuk-link` as Ian suggested?

CHANGELOG.md
```diff
@@ -11,6 +11,27 @@
 
 * Add intervention component ([PR #2196](https://github.com/alphagov/govuk_publishing_components/pull/2196))
 
+* Add scroll tracking to welsh brexit child taxon pages ([PR #2201](https://github.com/alphagov/govuk_publishing_components/pull/2201))
+
+## 24.18.4
```
chao-xian, line 16: We don't keep merge commits on main branches except for merging of PRs - this can be rebased.

chao-xian (commented):
Thanks for re-working this @maxf - just a couple of really minor comments.

CHANGELOG.md
```diff
@@ -30,6 +30,8 @@
 * Update LUX to activate when usage cookies are accepted ([PR #2154](https://github.com/alphagov/govuk_publishing_components/pull/2154)) PATCH
 * Add custom dimension to the page view data for pages tagged to brexit taxons. And remove the custom dimension from superbreadcrumb clicks. ([PR #2186](https://github.com/alphagov/govuk_publishing_components/pull/2186))
 
+* Add intervention component ([PR #2196](https://github.com/alphagov/govuk_publishing_components/pull/2196))
```
andysellick, line 33: This should be under the 'Unreleased' heading, after line 10.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,24 @@
+.gem-c-intervention {
+  @include govuk-text-colour;
+  @include govuk-responsive-padding(4);
+  @include govuk-responsive-margin(4, "bottom");
+
```
andysellick, line 5: Nitpick: is this empty line needed?

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -0,0 +1,24 @@
...
+  a {
+    @include govuk-link-common;
+    @include govuk-link-style-default;
+    @include govuk-link-print-friendly;
+  }
+}
+
+.gem-c-intervention__description,
+.gem-c-intervention__dismiss {
```
andysellick, line 22: If these two classes are styled exactly the same, could they be merged into a single class?

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -0,0 +1,18 @@
...
+description: An area that contains personalised content to the user
+body: |
+  The intervention is used to show personalised content. For instance, if the user has visited multiple
+  pages in the same area of the site, we might want let them know that there are other pages on GOV.UK
+  that would be useful to them. This component would be used to add this personalised content and would
+  indicate to the user that this is not normally part of the page, but has been added for them specifically.
+
+  Right now the contents of the component are static, as the MVP of personalised content is only for Start a Business.
+  Since many pages will use this component with the same text, we hard-code it here for now.
+accessibility_criteria: |
```
andysellick, line 11: The component also contains a link, so you should add the shared accessibility criteria for links, add the following (as in https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/docs/accordion.yml):

```
  - link
```

config/locales/en.yml
```diff
@@ -72,6 +72,10 @@ en:
       news_and_communications: News and communications
       statistics: Statistics
       worldwide: Worldwide
+    intervention:
+      title: Check the next steps for your limited company
+      description: You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.
+      dismiss_html: <a class="govuk-link" href="?hide-intervention=true">Hide this suggestion</a> if it’s not relevant to you
```
andysellick, line 78: I'm wondering how this link will be handled - there's no explanation in the documentation.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -0,0 +1,15 @@
+<%= tag.section class: "gem-c-intervention", role: "region" do %>
+  <h2 class="gem-c-intervention__title">
+    <a class="govuk-link" href="https://www.gov.uk/next-steps-for-your-business">
```
andysellick, line 3: Does this need to be an absolute link?

andysellick (commented):
A few more quick comments.

CHANGELOG.md
maxf: well spotted. Stupid git (not you, just git)

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: nope. I just picked the wrong component to copy from ;)

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: I'm sure they'll diverge eventually but makes sense not to anticipate things too much

config/locales/en.yml
maxf: Added:

    The dismiss link will reload the page but the `hide-intervention` query string parameter will cause the
    backed not to show the intervention again. Some progressive enhancement will be added in later to avoid
    reloading the page if JavaScript is available.

app/views/govuk_publishing_components/components/_intervention.html.erb
maxf: No it doesn't. Fixed.

injms:
I think you need to remove commit 04df82ecd03d967950c08d5db3a2d3cd46982150, and then rebase this branch against the master branch rather than merging master into this branch.

maxf:
@injms Done (note to self: don't trust editing files in github directly)

injms (approved):
Super! Happy to approve this component for the A/B test.

I think that if the test proves successful then either using another component, such as the [notice], [inset text], or [panel] components, or making a variation of these components would be a good direction to consider.

[notice]: https://components.publishing.service.gov.uk/component-guide/notice
[inset text]: https://components.publishing.service.gov.uk/component-guide/inset_text
[panel]: https://components.publishing.service.gov.uk/component-guide/panel

PR #2197, opened 2021-07-07
Update custom dimension tracking for brexit superbreadcrumbs

hannako:
## What
- Update CD 111 to track the audience on page views of all pages tagged to a brexit (priority) taxon. 
- Add CD 112 to track the audience on page views for pages tagged to a brexit (priority) taxon AND showing the brexit superbreadcrumb.

## Why
Request from PA

## Visual Changes
None

## Meta tags for pages with brexit superbreadcrumb

- both brexit-superbreadcrumb and brexit-audience are set. 

Screenshot 2021-07-07 at 16 49 13

## Met tags for pages tagged to a brexit priority taxon without a superbreadcrumb

- only the brexit-audience is set.

Screenshot 2021-07-07 at 16 52 22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add Welsh translations for topics, transition and tabs ([PR #2193](https://github.com/alphagov/govuk_publishing_components/pull/2193))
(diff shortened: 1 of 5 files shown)


alex-ju (approved):
Assuming the dimension indexes are the right ones (I never knew what's behind these numbers and if we keep track of them in any way) the change looks good!

PR #2199, opened 2021-07-07
Expand trigger event script options

andysellick:
## What
Expands the trigger event script to allow the following:

- allow a keyCode to be passed (yes, this is deprecated, but it works right now)
- allow 'bubbles' and 'cancelable' to be set, previously defaulted to true

## Why
The first thing is needed for some tests in finder-frontend (e.g. https://github.com/alphagov/finder-frontend/blob/main/spec/javascripts/live_search_spec.js#L291). The second thing seems like a logical addition.

## Visual Changes
None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/lib/trigger-event-spec.js
+++ b/spec/javascripts/govuk_publishing_components/lib/trigger-event-spec.js
@@ -8,6 +8,9 @@ describe('The trigger event code', function () {
       if (((event || {}).detail || {}).test === true) {
         obj.secondCalledFunction()
       }
+      if (event.keyCode === 13) {
+        obj.secondCalledFunction()
+      }
     },
     secondCalledFunction: function () {}
   }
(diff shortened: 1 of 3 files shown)


spec/javascripts/govuk_publishing_components/lib/trigger-event-spec.js
```diff
@@ -8,6 +8,9 @@ describe('The trigger event code', function () {
       if (((event || {}).detail || {}).test === true) {
         obj.secondCalledFunction()
       }
+      if (event.keyCode === 13) {
```
chao-xian, line 17: What does 13 signify? Can we assign the value to a variable name that describes that please?

andysellick: For the purposes of the test it doesn't matter what gets passed - the test passes a keycode to the triggerEvent script, then checks that the same keycode comes out the other end in the called function.

If you're still interested though it's the keycode for Enter.

chao-xian: Yeah I know, and the tests explain what's going on well but this bit made me think ([Don't Make Me Think](https://www.interaction-design.org/literature/article/don-t-make-me-think-key-learning-points-for-ux-design-for-the-web)). It just makes it easier to read. You could add a comment?

```suggestion
      // For example if user presses Enter
      if (event.keyCode === 13) {
```

```suggestion
      enterKeyCode = 13
      if (event.keyCode === enterKeyCode) {
```

andysellick: I thought about it, but I think it detracts from the point of the test. I could have put anything in here, it's not significant.

chao-xian: It's nbd as it was a suggestion and not a blocker anyway.

chao-xian (approved):
LGTM but could you also point to the tests in finder-frontend as examples please.

PR #2200, opened 2021-07-07
Give working examples for the checkbox name

leenagupte:
## What
Give working examples for the checkbox name

## Why
Checkboxes in Rails have a bit of a configuration quirk. If you don't add `[]` to the name, only the last selected value is passed to the controller, rather than an array of all of the selected values.

The name needs to be backed by indicating that the checkbox param is an array in the controller.

For example, if the checkbox is called `nationality[]`, then you need to do the permit as:

```
params.permit(nationality: [])
```
## Visual Changes
N/A

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/checkboxes.yml
+++ b/app/views/govuk_publishing_components/components/docs/checkboxes.yml
@@ -1,7 +1,7 @@
 name: Form checkboxes
 description: Let users select one or more options with checkboxes.
 body: |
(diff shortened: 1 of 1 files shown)


hannako (approved):
This is a strange one! Is it related to [this](https://www.rubydoc.info/docs/rails/4.0.0/ActionView%2FHelpers%2FFormBuilder:check_box) gotcha?

leenagupte:
> This is a strange one! Is it related to [this](https://www.rubydoc.info/docs/rails/4.0.0/ActionView%2FHelpers%2FFormBuilder:check_box) gotcha?

It sounds like it. Good find!

PR #2201, opened 2021-07-08
Add scroll tracking for brexit child taxon pages in welsh

hannako:
## What
Add scroll tracking to the welsh versions of the brexit child taxon pages.

## Why
Request from PA

## Visual Changes
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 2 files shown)


alex-ju (approved):
We could've made this more generic, to pick up URLs with and without `.cy` at the end, but that also means that we may end up tracking other Welsh pages that we might not want to, so happy for us to go with this and review scroll tracking on translated pages more holistically.

PR #2203, opened 2021-07-09
Fix arrow size on Brexit variation of action link

alex-ju:
## What
Adjust the size and positioning of the Brexit variation of action link

## Why
We'd like the Brexit arrow to scale down (along with the text) on narrow screens

## Visual Changes
<table>
<tr><td valign="top">

Screenshot 2021-07-09 at 12 53 48

</td><td valign="top">

Screenshot 2021-07-09 at 12 50 12

</td></tr>
</table>


hannako (approved):
I haven't been able to test this out locally by pointing collections at your branch. But the change looks sensible so 👍

PR #2207, opened 2021-07-13
Add inverse option for govspeak

andysellick:
## What
Adds an inverse option for govspeak. This is fairly limited right now, so is only guaranteed to work for headings, paragraphs and links.

## Why
We're using govspeak in the heading section of the coronavirus landing page, which is blue, and we want to add links.

## Visual Changes
Screenshot 2021-07-13 at 11 02 36

https://trello.com/c/o7kekKgI/471-not-to-do-until-we-have-approval-link-styling-within-coronavirus-header-markdown


app/assets/stylesheets/govuk_publishing_components/components/_govspeak.scss
```diff
@@ -28,3 +28,12 @@
...
 }
+
+.gem-c-govspeak--inverse {
+  color: govuk-colour('white');
+
+  a {
+    @include govuk-link-common;
+    @include govuk-link-style-inverse;
+  }
+}
```
alex-ju, line 39: Would this be better placed in [_typography](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss)?

spec/components/govspeak_spec.rb
```diff
@@ -12,6 +12,15 @@ def component_name
     assert_select ".gem-c-govspeak h1", text: "content"
   end
 
+  it "renders inverse content correctly" do
+    render_component(
+      inverse: true,
+      content: "<h2>inverse</h2>".html_safe,
```
injms, line 18: I believe that `raw()` is preferred over `.html_safe` - but I can't find the documentation to say so and why. They're effectively doing the same thing in this instance, so not a big deal.

```suggestion
      content: raw("<h2>inverse</h2>")
```

andysellick: This was based on the other tests in the file.

app/assets/stylesheets/govuk_publishing_components/components/_govspeak.scss
andysellick: Fair point, have moved. I've not nested it in the existing structure of this file as I'm concerned about the volume of outputted CSS, but added it in at the end as a standalone item.

PR #2211, opened 2021-07-14
Fix new link styles on document list

andysellick:
## What
The document list component was recently updated to use the new `govuk-frontend` link styles, but I spotted a couple of problems.

- fixes With only link variant, by removing overflow hidden (was clipping bottom border off focussed link)
- fixes With parts variant, by adding extra styles in and improving spacing beneath link (hover was wrong and hover + focus had a double underline)
- remove general link style in favour of using govuk-link class in template

## Why
Improving consistency for our link styles.

## Visual Changes
Before | After
------ | ------
Screenshot 2021-07-14 at 12 01 10 | Screenshot 2021-07-14 at 12 01 35
Screenshot 2021-07-14 at 12 03 52 | Screenshot 2021-07-14 at 12 03 59

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix new link styles on document list ([PR #2211](https://github.com/alphagov/govuk_publishing_components/pull/2211))
 * Add Layout super navigation header component ([PR #2179](https://github.com/alphagov/govuk_publishing_components/pull/2179))
 
 ## 24.20.0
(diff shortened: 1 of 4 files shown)


owenatgov (approved):
Top spot!

kevindew:
@andysellick just an FYI that this change caused Smokey to fail for finder-frontend. I've put this up to fix it: https://github.com/alphagov/smokey/pull/843

Not too sure what the ideal solution is to this sort of problem - I'm assuming we don't consider the selectors within a component to be part of a public API, so it doesn't seem right that Smokey depends on that selector structure. However I don't think there is a means for the application (finder-frontend) to specify it's own selector that could be used for this.

andysellick:
@kevindew thanks for the heads up. Is there a way we can run Smokey on changes in the gem ahead of it being a new release and getting rolled out to an application?

PR #2212, opened 2021-07-14
Add meta tag to all content tagged to Brexit

hannako:
## What
Improve the tracking of pages tagged to a brexit priority taxon. 

## Why
If a page is tagged to a priority brexit taxon, we want it to contain a meta tag for the brexit audience type (`govuk:brexit-audience`), and send the value of that meta tag to GA as custom dimension 112. Currently this will not happen if the page is also tagged to a higher priority taxon (eg coronavirus child taxons).

## Source of bug
The [brexit_audience](https://github.com/alphagov/govuk_publishing_components/blob/bb7d37240a898b38ab3e70d669c6187459e9f3e1/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_priority.rb#L46-L52) method is used by the meta_tag presenter to [decide](https://github.com/alphagov/govuk_publishing_components/blob/bb7d37240a898b38ab3e70d669c6187459e9f3e1/lib/govuk_publishing_components/presenters/meta_tags.rb#L126) if the meta tag should be added. And brexit_audience is calling `priority_taxon` which uses the order of taxons in the PRIORITY_TAXON hash. 

## Visual Changes
None

## Before (no brexit-audience meta tag is set)

Screenshot 2021-07-14 at 23 37 00

## After (contains brexit-audience meta tag (Custom dimension 112)

Screenshot 2021-07-14 at 23 35 22

### DIFF
--- a/spec/components/contextual_breadcrumbs_spec.rb
+++ b/spec/components/contextual_breadcrumbs_spec.rb
@@ -42,11 +42,26 @@ describe "ContextualBreadcrumbs", type: :view do
     content_item
   end
 
+  def tag_to_priority_coronavirus(content_item)
+    content_item["links"]["taxons"][1]["title"] = "Coronavirus Education"
+    content_item["links"]["taxons"][1]["content_id"] = "272308f4-05c8-4d0d-abc7-b7c2e3ccd249"
+    content_item
+  end
+
   it "renders the brexit-superbreadcrumb meta tag on content tagged to brexit" do
     content_item = example_document_for("guide", "guide")
     content_item = tag_to_brexit(content_item)
     render_component(content_item: content_item)
     assert_select "meta[name='govuk:brexit-superbreadcrumb'][content='Brexitbusiness']"
+  end
+
+  it "does not render the brexit-superbreadcrumb meta tag on content tagged to brexit AND a priority coronavirus taxon" do
+    content_item = example_document_for("guide", "guide")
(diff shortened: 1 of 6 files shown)


spec/components/contextual_breadcrumbs_spec.rb
```diff
@@ -42,13 +42,28 @@ def tag_to_brexit(content_item)
...
     assert_select "meta[name='govuk:brexit-superbreadcrumb'][content='Brexitbusiness']"
   end
 
+  it "does not render the brexit-superbreadcrumb meta tag on content tagged to brexit AND a priority coronavirus taxon" do
+    content_item = example_document_for("guide", "guide")
+    content_item = tag_to_brexit(content_item)
+    content_item["links"]["taxons"] << {}
+    content_item = tag_to_priority_coronavirus(content_item)
+    render_component(content_item: content_item)
+    assert_select "meta[name='govuk:brexit-superbreadcrumb'][content='Brexitbusiness']", count: 0
```
sihugh, line 64: I got a bit confused here until I realised they were two different metatags 🤦

sihugh (approved):
🍞

spec/components/contextual_breadcrumbs_spec.rb
hannako: Yeah it IS confusing! Thank you for the review 😅

PR #2213, opened 2021-07-15
Correcting changelog

andysellick:
## What / why

- a build fail with the release of 24.20.0 meant that PR 2179 was actually included in that release, but the changelog didn't say so, this corrects that
-


injms (approved):
Thanks for sorting this out; the new component isn't used anywhere yet, so it shouldn't affect anything.

Issue #2082, opened 2021-05-20
Notice, success and error alert components are not aligned


alex-ju:
The [notice](https://components.publishing.service.gov.uk/component-guide/notice), [success](https://components.publishing.service.gov.uk/component-guide/success_alert) and [error alert](https://components.publishing.service.gov.uk/component-guide/error_alert) components (and the [brexit notice](https://government-frontend.herokuapp.com/component-guide/brexit-notice) – hopefully, to be removed soon, so probably less to worry about) would ideally be one component with modifiers applied to generate variations. This would help with having a consistent API and with maintenance.

The `success alert` component was updated as part of rolling out the new [cookies page](https://design-system.service.gov.uk/patterns/cookies-page/) pattern as detailed in the Design System. However, the other ones are still using the old design.

andysellick:
This is an old issue, but we've looked into it. Short version is that the effort required to do this seems to outweigh the benefit. Long version follows.

The main problem is the effort of changing this. The components in question are used throughout our applications (success alert is used in 15 applications), with instances sometimes reaching tens or more per application. This would require a lot of time to change, including in a number of applications that are particularly difficult to run locally (e.g. `email-alert-frontend`). The code for the components is relatively straightforward although there are notable differences in the three - success and error are quite similar (success has an extra heading along the top) while notice is somewhat more complicated.

While merging all three seemed impractical, another suggestion was that we try to follow the Design System more. They only have two components - error and notification - with success being a variant of notification. Unfortunately success is the most used of the components, so just removing one component doesn't dramatically reduce the deployment complexity. There's also some benefit in having them separate as they're more discoverable - I hadn't noticed the success variant and initially assumed the component didn't exist in the Design System.

While we aim to have a consistent API and reduce maintenance this doesn't seem like a good option, so I'm going to close this issue, acknowledging that the situation isn't perfect. If in future it is reopened or raised again we can discuss it again, but this is our decision for now.

PR #2214, opened 2021-07-16
Update notice component to better match DS notification banner

danacotoran:
Between the [notice](https://components.publishing.service.gov.uk/component-guide/notice), [success](https://components.publishing.service.gov.uk/component-guide/success_alert) and [error alert](https://components.publishing.service.gov.uk/component-guide/error_alert) components, we've got a mix of styles and approaches. 
Ideally these would all be one component, with modifiers applied to generate variations. 
This is a step towards that goal.

The GOV.UK version of the notification component currently deviates from the Design System version. 
This adopts Design System classes and styles for this component, only keeping few CSS overrides that are necessary for our usage of this component. 

Instead of applying the "Important" banner style across the board, it has been suggested by the GOVUK design community to make it opt-in, so that it can be applied only where it suits the intent of the banner.

Additionally, the title text is customisable so any term can be used in case that the word "Important" doesn't quite fit with the information the notice is trying to convey.

-----

#2082 

Design System notification banner https://design-system.service.gov.uk/components/notification-banner/ 
GOV.UK notice component https://components.publishing.service.gov.uk/component-guide/notice

### DIFF
--- a/app/views/govuk_publishing_components/components/_notice.html.erb
+++ b/app/views/govuk_publishing_components/components/_notice.html.erb
@@ -8,29 +8,35 @@
   local_assigns[:margin_bottom] ||= 8
   local_assigns[:margin_bottom] = 8 if local_assigns[:margin_bottom] > 9
 
+  banner_title ||= t("components.notice.banner_title")
+  banner_title_id ||= "govuk-notification-banner-title-#{SecureRandom.hex(4)}"
+  show_banner_title ||= false
+  heading_level = show_banner_title ? "h3" : "h2"
+
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
-  css_classes = %w[gem-c-notice]
+  css_classes = %w[govuk-notification-banner gem-c-notice]
   css_classes << (shared_helper.get_margin_bottom)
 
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/_notice.html.erb
```diff
@@ -8,29 +8,35 @@
   local_assigns[:margin_bottom] ||= 8
   local_assigns[:margin_bottom] = 8 if local_assigns[:margin_bottom] > 9
 
+  banner_title ||= "Important"
```
owenatgov, line 11: Could we translationify this?

danacotoran: Have moved the string to the locale file

danacotoran:
> One little nitpick but otherwise this looks 👍 Something else worth noting is that I wasn't able to test this locally as it's currently experiencing the weird heroku issues we had earlier this week. A rebase should sort it.

Sorry about that @owenatgov, I guess this branch was not up to date! I've done a rebase so I think you should be able to check this locally now.

owenatgov (approved):
Fabulous! Just needs a rebase then this is good ot go in.

PR #2217, opened 2021-07-19
Add scroll tracking to travel advice pages

andysellick:
## What
Adds scroll tracking to the following pages:

- https://www.gov.uk/guidance/red-amber-and-green-list-rules-for-entering-england
- https://www.gov.uk/guidance/travel-abroad-from-england-during-coronavirus-covid-19
- https://www.gov.uk/guidance/travel-to-england-from-another-country-during-coronavirus-covid-19
- https://www.gov.uk/find-travel-test-provider

## Why
We want to know if users are scrolling on these pages.

## Visual Changes
None.

Trello card: https://trello.com/c/BSAZSZnI/479-add-scroll-tracking-to-travel-pages

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
@@ -233,6 +233,50 @@
       ['Percent', 60],
       ['Percent', 80],
       ['Percent', 100]
+    ],
+    '/guidance/red-amber-and-green-list-rules-for-entering-england': [
+      ['Heading', 'Stay up-to-date'],
+      ['Heading', 'Red list of countries and territories'],
+      ['Heading', 'Amber list of countries and territories'],
+      ['Heading', 'Amber list rules if you are fully UK vaccinated'],
+      ['Heading', 'Amber list rules if you are not fully UK vaccinated'],
+      ['Heading', 'Green list countries and territories'],
+      ['Heading', 'Green list rules'],
+      ['Heading', 'Ireland, the UK, the Channel Islands and the Isle of Man'],
+      ['Heading', 'Transit stops in amber or red list countries'],
+      ['Heading', 'Job and medical exemptions'],
+      ['Heading', 'Travelling abroad from England'],
+      ['Heading', 'Demonstrating your COVID-19 vaccination status when travelling abroad'],
+      ['Heading', 'Foreign, Commonwealth & Development Office travel advice']
+    ],
+    '/guidance/travel-abroad-from-england-during-coronavirus-covid-19': [
+      ['Heading', 'Before you travel abroad'],
+      ['Heading', '1. Check the rules for the country you’re going to'],
+      ['Heading', '2. Find out what you need to do when you return'],
+      ['Heading', '3. Book any COVID-19 tests or quarantine hotel packages you need for your return'],
+      ['Heading', 'Before you return'],
+      ['Heading', 'After you return'],
+      ['Heading', 'Travelling with children'],
+      ['Heading', 'Ireland, the UK, the Channel Islands and the Isle of Man'],
+      ['Heading', 'Exemptions for work, medical or compassionate reasons']
+    ],
+    '/guidance/travel-to-england-from-another-country-during-coronavirus-covid-19': [
+      ['Heading', 'Before you travel'],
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
```diff
@@ -233,6 +233,50 @@
...
+      ['Heading', 'Travelling with children'],
+      ['Heading', 'Ireland, the UK, the Channel Islands and the Isle of Man'],
+      ['Heading', 'Exemptions for work, medical or compassionate reasons']
+    ],
+    '/find-travel-test-provider': [
+      ['Percent', 20],
+      ['Percent', 40],
+      ['Percent', 60],
+      ['Percent', 80],
+      ['Percent', 100]
```
injms, line 279: This percentage tracking used so often - is it worth assigning it to a variable and then referencing that variable instead of repeating it over and over and over again?

andysellick: Honestly I'd not seen this code until today and I think the whole thing needs a total overhaul. The idea that adding scroll tracking to a single page means adding weight to our overall JS is not in any way good. Also page headings are hard coded here, which means if they change, presumably this stops working?

My next task is to look at an alternative way of doing this. But yes, that's a good optimisation in the meantime, I'll look at it.

andysellick: Added.

app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
```diff
@@ -5,21 +5,17 @@
 
   window.GOVUK = window.GOVUK || {}
 
+  var percentages = [
```
gclssvglx, line 8: Can this be a `const` Andy, or we not using those yet?

app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
```diff
@@ -5,21 +5,17 @@
...
+    ['Percent', 20],
+    ['Percent', 40],
+    ['Percent', 60],
+    ['Percent', 80],
+    ['Percent', 100]
+  ]
+
   var CONFIG = {
```
gclssvglx, line 16: As above - can this be a `const`?

app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
```diff
@@ -233,6 +233,50 @@
...
+      ['Heading', 'Amber list rules if you are fully UK vaccinated'],
+      ['Heading', 'Amber list rules if you are not fully UK vaccinated'],
+      ['Heading', 'Green list countries and territories'],
+      ['Heading', 'Green list rules'],
+      ['Heading', 'Ireland, the UK, the Channel Islands and the Isle of Man'],
+      ['Heading', 'Transit stops in amber or red list countries'],
+      ['Heading', 'Job and medical exemptions'],
+      ['Heading', 'Travelling abroad from England'],
+      ['Heading', 'Demonstrating your COVID-19 vaccination status when travelling abroad'],
+      ['Heading', 'Foreign, Commonwealth & Development Office travel advice']
```
gclssvglx, line 180: I assume headings work in the same way as percentages, i.e. points on the page that the tracking identifies as either hitting or going past?  Also, I assume they have to be in order?  I see what you mean, this is a tad yucky!

gclssvglx (approved):
Thanks @andysellick this looks great!  Lovely clean up on the percentages - makes the code so much nicer.  

No blockers, but one question...

injms (approved):

app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
andysellick: Afraid not. `const` isn't supported by IE10 and below, and only partly supported in IE11 https://caniuse.com/const

app/assets/javascripts/govuk_publishing_components/analytics/scroll-tracker.js
andysellick: I assume so, yes, yes, and yes 😁

Issue #2035, opened 2021-04-27
Deprecate Sass `em()` function


injms:
The [`em()` function][1] should be deprecated as it can be swapped for the identical [`govuk-em()` function][2] from GOV.UK Frontend.

There is a `@warn` message to let people know that this function could be removed in a future major release of the GOV.UK Publishing Components gem.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/037ce27f3e05973c60d1522bcd113677cc6e7122/app/assets/stylesheets/govuk_publishing_components/components/helpers/_px-to-em.scss
[2]: https://github.com/alphagov/govuk-frontend/blob/c8d2deddc2cf7fa6d116f2af320b322f23856689/package/govuk/tools/_px-to-em.scss

PR #2220, opened 2021-07-20
Remove `em()` Sass mixin

alex-ju:
## What
Remove `em()` Sass mixin

## Why
This was deprecated for a while now and it's no longer used in GOV.UK applications.
Fixes https://github.com/alphagov/govuk_publishing_components/issues/2035

## Visual Changes
No visual changes


andysellick (approved):
Looks good. If this isn't in use anymore does this still count as a breaking change?

(also need to update the changelog)

alex-ju:
> Looks good. If this isn't in use anymore does this still count as a breaking change?

It is technically a breaking change and I haven't looked anywhere outside our core frontend apps so I thought we could bundle it with [others in the pipeline](https://github.com/alphagov/govuk_publishing_components/milestone/5).

andysellick:
> It is technically a breaking change and I haven't looked anywhere outside our core frontend apps so I thought we could bundle it with [others in the pipeline](https://github.com/alphagov/govuk_publishing_components/milestone/5).

Good shout. I've just noticed it's also a breaking change because static explicitly imports this file: https://github.com/alphagov/static/blob/main/app/assets/stylesheets/header-footer-only.scss#L6

alex-ju:
> Looks good. If this isn't in use anymore does this still count as a breaking change?
> 
> (also need to update the changelog)

It oddly imports it without using it.. 🤷 
Will remove it from `static` beforehand to make dependabot happy.

PR #2224, opened 2021-07-21
Intervention: add tracking code and design tweaks

maxf:
## What

* Intervention component: adjust the design and add GA code

## Why

We need to track when the user sees the component and what action they take upon seeing it

### Before

before

### After

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,7 +12,7 @@
 * BREAKING: Remove ability to link contextual text on titles ([PR #2192](https://github.com/alphagov/govuk_publishing_components/pull/2192))
 * Delete empty print stylesheets ([PR #2225](https://github.com/alphagov/govuk_publishing_components/pull/2225))
 * BREAKING: Remove is_page_heading parameter from radio component ([PR #2061](https://github.com/alphagov/govuk_publishing_components/pull/2061))
-
+* Intervention: add tracking code and design tweaks ([PR #2224](https://github.com/alphagov/govuk_publishing_components/pull/2224))
 
 ## 24.21.1
 
--- a/app/assets/javascripts/govuk_publishing_components/components/intervention.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/intervention.js
@@ -1,0 +1,16 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Intervention ($module) {
+    this.$module = $module
+  }
+
+  Intervention.prototype.init = function () {
+    if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+      window.GOVUK.analytics.trackEvent('interventionBanner', 'interventionShown')
+    }
+  }
+
+  Modules.Intervention = Intervention
+})(window.GOVUK.Modules)
--- a/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
@@ -1,16 +1,18 @@
 .gem-c-intervention {
   @include govuk-text-colour;
-  @include govuk-responsive-padding(4);
-  @include govuk-responsive-margin(4, "bottom");
+  @include govuk-responsive-padding(6, "left");
+  @include govuk-responsive-padding(6, "right");
+  @include govuk-responsive-padding(4, "top");
+  @include govuk-responsive-padding(4, "bottom");
+  @include govuk-responsive-margin(6, "bottom");
   background-color: govuk-colour("light-grey", $legacy: "grey-4");
-}
 
-.gem-c-intervention__title {
-  @include govuk-font(24, $weight: bold);
-  margin-top: 0;
-  @include govuk-responsive-margin(4, "bottom");
-}
+  .gem-c-intervention__title {
+    @include govuk-responsive-margin(0);
+    @include govuk-responsive-padding(4, "bottom");
+  }
 
-.gem-c-intervention__paragraph {
-  @include govuk-font(19);
+  .govuk-body:last-of-type {
+    @include govuk-responsive-margin(0, "bottom");
+  }
 }
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -1,15 +1,33 @@
-<%= tag.section class: "gem-c-intervention", role: "region" do %>
-  <h2 class="gem-c-intervention__title">
-    <a class="govuk-link" href="/next-steps-for-your-business">
-      <%= t("components.intervention.title") %>
-    </a>
+<%
+  data_attributes ||= {}
+  data_attributes[:module] = 'intervention'
+%>
+
+<%= tag.section class: "gem-c-intervention", role: "region", data: data_attributes do %>
+  <h2 class="gem-c-intervention__title govuk-heading-s">
+    <a class="govuk-link" href="/next-steps-for-your-business"
+      data-module="gem-track-click"
+      data-track-category="interventionBanner"
+      data-track-action="interventionClicked"
+      data-track-dimension="<%= t("components.intervention.title") %>"
+      data-track-dimension-index="29"
+      data-track-label="/next-steps-for-your-business">
+      <%= t("components.intervention.title") %></a>
   </h2>
 
-  <p class="gem-c-intervention__paragraph">
+  <p class="govuk-body">
     <%= t("components.intervention.description") %>
   </p>
 
-  <p class="gem-c-intervention__paragraph">
-    <%= t("components.intervention.dismiss_html") %>
+  <p class="govuk-body">
+    <a class="govuk-link" href="?hide-intervention=true"
+      data-module="gem-track-click"
+      data-track-category="interventionBanner"
+      data-track-action="interventionDismissed"
+      data-track-label="?hide-intervention=true"
+      data-track-dimension="<%= t("components.intervention.dismiss_link_text") %>"
+      data-track-dimension-index="29">
+      <%= t("components.intervention.dismiss_link_text") %></a>
+    <%= t("components.intervention.dismiss_post_link") %>
   </p>
 <% end %>
--- a/spec/components/intervention_spec.rb
+++ b/spec/components/intervention_spec.rb
@@ -8,7 +8,7 @@ describe "Intervention", type: :view do
   it "renders the component" do
     render_component({})
     assert_select ".gem-c-intervention__title", text: "Check the next steps for your limited company"
-    assert_select ".gem-c-intervention__paragraph", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
-    assert_select ".gem-c-intervention__paragraph", text: "Hide this suggestion if it’s not relevant to you"
+    assert_select ".govuk-body", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
+    assert_select ".govuk-body a", text: "Hide this suggestion"
   end
 end
--- a/spec/javascripts/components/intervention-spec.js
+++ b/spec/javascripts/components/intervention-spec.js
@@ -1,0 +1,23 @@
+/* eslint-env jasmine */
+/* global GOVUK */
+
+describe('The intervention component', function () {
+  var FIXTURE = "<section class='gem-c-intervention' data-module='intervention'></section>"
+
+  beforeEach(function () {
+    spyOn(GOVUK.analytics, 'trackEvent')
+    window.setFixtures(FIXTURE)
+    var element = document.querySelector('[data-module="gem-intervention"]')
+    new GOVUK.Modules.Intervention(element).init()
+  })
+
+  afterEach(function () {
+    if (GOVUK.analytics.trackEvent.calls) {
+      GOVUK.analytics.trackEvent.calls.reset()
+    }
+  })
+
+  it('triggers a tracking event when the component is displayed', function () {
+    expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith('interventionBanner', 'interventionShown')
+  })
+})
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -75,7 +75,8 @@ en:
     intervention:
       title: Check the next steps for your limited company
       description: You might be interested in this because you’ve been browsing guidance relevant to starting a limited company.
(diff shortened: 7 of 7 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -1,16 +1,26 @@
...
+  @include govuk-responsive-padding(6, "right");
+  @include govuk-responsive-padding(4, "top");
+  @include govuk-responsive-padding(4, "bottom");
+  @include govuk-responsive-margin(6, "bottom");
   background-color: govuk-colour("light-grey", $legacy: "grey-4");
 }
 
 .gem-c-intervention__title {
-  @include govuk-font(24, $weight: bold);
+  @include govuk-font(19, $weight: bold);
```
danacotoran, line 12: Not sure we need custom CSS for `.gem-c-intervention_title`. Could we not achieve this using [govuk-heading-s](https://design-system.service.gov.uk/styles/typography/headings/index.html)?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,17 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.GovukDetails = window.GOVUKFrontend;
```
danacotoran, line 3: I'm not sure you meant to include this line!

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+
+## Unrelased
```
danacotoran, line 11: Seems like there's a small typo here: `Unrelased` should be `Unreleased`

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -1,16 +1,26 @@
...
   margin-top: 0;
-  @include govuk-responsive-margin(4, "bottom");
+  @include govuk-responsive-margin(0);
+  @include govuk-responsive-padding(4, "bottom");
 }
 
 .gem-c-intervention__paragraph {
   @include govuk-font(19);
+  @include govuk-responsive-margin(0);
```
danacotoran, line 20: Similar comment as above, these styles seem similar (or the same) as the standard [govuk-body](url) paragraph style, so I wonder if we can just apply that class to the relevant elements?

danacotoran (commented):
Haven't tested the tracking side of it but I just left a few general comments with things I noticed on first glance!

spec/components/intervention_spec.rb
```diff
@@ -9,6 +9,12 @@ def component_name
     render_component({})
     assert_select ".gem-c-intervention__title", text: "Check the next steps for your limited company"
     assert_select ".gem-c-intervention__paragraph", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
-    assert_select ".gem-c-intervention__paragraph", text: "Hide this suggestion if it’s not relevant to you"
+    assert_select ".gem-c-intervention__paragraph a", text: "Hide this suggestion"
```
danacotoran, line 12: This of course checks that the correct tracking attributes are present, which is fine, but I wonder if there should also be a javascript test to verify that the Analytics events are actually firing as expected. 

I think there are some examples you can have a look at [here](https://github.com/alphagov/govuk_publishing_components/tree/master/spec/javascripts/components).

danacotoran: Actually to be fair you don't have much custom tracking here (just the event that's expected to fire on page load and the link click events I suppose) so it _might_ be overkill to add JS test coverage?  not actually sure.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -1,15 +1,24 @@
...
     <%= t("components.intervention.description") %>
   </p>
 
   <p class="gem-c-intervention__paragraph">
-    <%= t("components.intervention.dismiss_html") %>
+    <a class="govuk-link" href="?hide-intervention=true"
+      data-track-category="interventionBanner"
+      data-track-action="interventionDismissed"
+      data-track-label="?hide-intervention=true"
+      data-track-options="{'dimension29':'<%= t("components.intervention.dismiss_link_text") %>'}">
```
danacotoran, line 20: Looking at other components (for example [here](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_step_by_step_nav_related.html.erb#L18)) it seems like there is an option to use the `data-track-dimension` and `data-track-dimension-index` attribute to set custom dimensions which seems a bit cleaner to me, if we're only setting one dimension. 
There's [some additional processing](https://github.com/alphagov/govuk_publishing_components/blob/117837e0079eda0c58a7895dd027279952da8b1a/app/assets/javascripts/govuk_publishing_components/lib/track-click.js#L49) that happens when there are extra options which I think can be avoided in this particular instance.

danacotoran (changes requested):
Just a couple more issues I spotted checking this in a browser.

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
...
+
+* Intervention component: adjust the design and add GA code ([PR #2224](https://github.com/alphagov/govuk_publishing_components/pull/2224))
```
andysellick, line 13: Nitpick: Changelog entries should match the title of the PR.

spec/components/intervention_spec.rb
andysellick: Every component that has a JS file also needs to have a JS test file. It should be fairly straightforward - spyon the GA tracking function, initialise the component, check that the GA tracking function was called.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: Indeed. Done.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
maxf: I've simplified the whole markup by using `govuk-` attributes and keeping only the necessary CSS

spec/components/intervention_spec.rb
maxf: Added

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -1,15 +1,24 @@
 <%= tag.section class: "gem-c-intervention", role: "region" do %>
   <h2 class="gem-c-intervention__title">
-    <a class="govuk-link" href="/next-steps-for-your-business">
-      <%= t("components.intervention.title") %>
-    </a>
+    <a class="govuk-link" href="/next-steps-for-your-business"
```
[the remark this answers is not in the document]
maxf: Added

app/views/govuk_publishing_components/components/_intervention.html.erb
maxf: I've made the change

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,8 @@
+/* eslint-env jasmine, jquery */
+/* global GOVUK */
+
+describe('Intervention component', function () {
+  it('triggers a tracking event when the component is displayed', function () {
+    spyOn(GOVUK.analytics, 'trackEvent')
```
andysellick, line 6: This test doesn't test anything. You'll need to add a chunk of markup, initialise the component code on it and do and `expect` line to check that the spy has been called. Have a look at https://github.com/alphagov/govuk_publishing_components/blob/master/spec/javascripts/components/details-spec.js for an example.

andysellick (changes requested):

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
+/* eslint-env jasmine, jquery */
```
andysellick, line 1: I'm not sure this file uses any jQuery, is `jquery` needed here?

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
...
+/* global GOVUK */
+
+describe('Intervention component', function () {
```
andysellick, line 4: Nitpick: I'd name this `describe` `An intervention component` or `The intervention component`.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
...
+  var FIXTURE
+
+  function loadInterventionComponent () {
```
andysellick, line 7: Having this as a separate function seems unnecessary - it's two lines that could be put into the `beforeEach` function quite easily.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
...
+    var element = document.querySelector('[data-module="gem-intervention"]')
+    new GOVUK.Modules.Intervention(element).init()
+  }
+
+  beforeEach(function () {
+    spyOn(GOVUK.analytics, 'trackEvent')
+    spyOn(GOVUK.Modules, 'GemTrackClick').and.callFake(function () { this.start = function () {} })
+
+    FIXTURE = `
```
andysellick, line 16: For readability of the `beforeEach` function, I'd move this out of here to the top of the file.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
...
+     <section class="gem-c-intervention" role="region" data-module="intervention">
+        <h2 class="gem-c-intervention__title govuk-heading-s">
```
andysellick, line 18: For the purposes of this test all of the markup here doesn't seem necessary, as it doesn't do anything, am I right? Since all we're testing is the component loading and firing a tracking event, could most of the HTML here be removed?

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,59 @@
...
+  var FIXTURE
+
+  function loadInterventionComponent () {
+    var element = document.querySelector('[data-module="gem-intervention"]')
+    new GOVUK.Modules.Intervention(element).init()
+  }
+
+  beforeEach(function () {
+    spyOn(GOVUK.analytics, 'trackEvent')
+    spyOn(GOVUK.Modules, 'GemTrackClick').and.callFake(function () { this.start = function () {} })
```
andysellick, line 14: Is this line needed? This spy doesn't appear to be referenced again, and all we need to test is that the analytics `trackEvent` function was called - it doesn't matter if it doesn't do anything or not.

spec/components/intervention_spec.rb
```diff
@@ -8,7 +8,13 @@ def component_name
...
-    assert_select ".gem-c-intervention__paragraph", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
-    assert_select ".gem-c-intervention__paragraph", text: "Hide this suggestion if it’s not relevant to you"
+    assert_select ".govuk-body", text: "You might be interested in this because you’ve been browsing guidance relevant to starting a limited company."
+    assert_select ".govuk-body a", text: "Hide this suggestion"
+  end
+
+  it "renders the tracking attributes" do
+    render_component({})
+    assert_select ".gem-c-intervention__title a[data-track-category='interventionBanner'][data-track-action='interventionClicked'][data-track-label='/next-steps-for-your-business'][data-track-dimension-index='29'][data-track-dimension='Check the next steps for your limited company']"
+    assert_select ".govuk-body a[data-track-category='interventionBanner'][data-track-action='interventionDismissed'][data-track-label='?hide-intervention=true'][data-track-dimension-index='29'][data-track-dimension='Hide this suggestion']"
```
andysellick, line 18: I'm not sure a test is needed to check that the hard coded attributes in the component template are what they're supposed to be. If they could be altered by passing options, then yes, definitely needs a test, but the component doesn't allow that.

spec/javascripts/components/intervention-spec.js
maxf: Makes sense. I thought that was the convention (looking at other components)

spec/components/intervention_spec.rb
maxf: Removed.
But arguably, if that test isn't needed then the first tests isn't, either, since it also tests static content.
It won't matter on further iterations when the text is passed as an option anyway,

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,37 @@
+/* eslint-env jasmine */
+/* global GOVUK */
+
+describe('The intervention component', function () {
+  const FIXTURE = `
```
andysellick, line 5: We can't use `const` in our public code because it's not supported by IE10 and not fully supported by IE11. That doesn't matter so much here because the tests aren't part of the public code, however, the tests still need to pass in IE10.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,37 @@
...
+    <section class="gem-c-intervention" role="region" data-module="intervention">
+      <h2>
+        <a href="/next-steps-for-your-business">
+          Check the next steps for your limited company
+        </a>
+      </h2>
+
+      <p class="govuk-body"></p>
+      <p class="govuk-body"></p>
```
andysellick, line 14: I really don't think you need any markup in this component fixture at all - it won't make any difference to the outcome of the test.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,37 @@
...
+
+  afterEach(function () {
+    if (GOVUK.analytics.trackEvent.calls) {
+      GOVUK.analytics.trackEvent.calls.reset()
+    }
+  })
+
+  it('triggers a tracking event when the component is displayed', function () {
+    loadInterventionComponent()
+    expect(GOVUK.analytics.trackEvent.calls.count()).toEqual(1)
```
andysellick, line 35: This is good, but you can also make this test more specific e.g.

`expect(GOVUK.analytics.trackEvent).toHaveBeenCalledWith('interventionBanner', 'interventionShown')`

spec/javascripts/components/intervention-spec.js
maxf: Fixed.
How do we check that the tests pass in IE10 though?

andysellick: You can run the jasmine tests through a local browser, using `bundle exec rake app:jasmine` and going to localhost:8888 in IE10 in browserstack. Ideally we'd have that as part of the build pipeline somehow, still working on that.

andysellick (approved):

danacotoran (approved):
Happy to approve this now, thanks!

Issue #2065, opened 2021-05-13
Remove references to now empty print stylesheets from applications' print files


danacotoran:
In https://github.com/alphagov/govuk_publishing_components/pull/1561, `display: none` rules were removed from several component print stylesheets, and the `govuk-!-display-none-print` class was used instead.
As a consequence of this, some of the component print stylesheets which previously used to specify that the component should be hidden in print view, are now empty files ([Example](https://github.com/alphagov/govuk_publishing_components/blob/9213c4fc722900eda91c722cbb1af69ecacdafb7/app/assets/stylesheets/govuk_publishing_components/components/print/_back-link.scss)).
These empty files could not be removed immediately due to being referenced in application print stylesheets ([Example](https://github.com/alphagov/govuk-account-manager-prototype/blob/7858b55afcddc2f855d909abd051838d95f21f13/app/assets/stylesheets/print.scss#L2))

The purpose of this task is to hunt down and remove imports of these now empty print stylesheets. So that we can eventually delete the empty stylesheets altogether.

danacotoran:
All merged

PR #2225, opened 2021-07-23
Delete empty print stylesheets

danacotoran:
Not long ago, there was some refactoring to remove `display: none` rules from component stylesheets in favour of using the `govuk-!-display-none-print` class instead: alphagov/govuk_publishing_components#1561

As a consequence of that refactoring process, some of the component print stylesheets which previously used to contain CSS are now empty files  ([Example](https://github.com/alphagov/govuk_publishing_components/blob/9213c4fc722900eda91c722cbb1af69ecacdafb7/app/assets/stylesheets/govuk_publishing_components/components/print/_back-link.scss)). These empty files could not be removed immediately as they were referenced in many applications' print stylesheets  ([Example](https://github.com/alphagov/govuk-account-manager-prototype/blob/7858b55afcddc2f855d909abd051838d95f21f13/app/assets/stylesheets/print.scss#L2))

~**Step 1** in the process of deleting these empty files is to hunt down references of these files in application print stylesheets and remove them. This step will be finalised once all the PRs linked to [this issue](https://github.com/alphagov/govuk_publishing_components/issues/2065) have been completed.~ **Done**

**Step 2** is removing these files from the components gem altogether, which is what this PR is doing. This step can only be completed once we have completed step 1.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Delete empty print stylesheets ([PR #2225](https://github.com/alphagov/govuk_publishing_components/pull/2225))
+
 ## 24.21.1
 
 * Fix issue where metatag component throws an error when priority_brexit_taxon isn't available. ([PR #2227](https://github.com/alphagov/govuk_publishing_components/pull/2227))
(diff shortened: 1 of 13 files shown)


andysellick (approved):
I'd merge the two commits in this PR into one, since they are related.

PR #2226, opened 2021-07-23
Add option to nest context within the h1 for title component

jon-kirwan:
## What
Move the caption inside of the h1.

## Why
Currently, the context is missed when page titles are read by screen readers. This way when users navigate through headings with a screenreader will hear the title with context announced.

No visual changes.

https://trello.com/c/gPFqiwZa/799-update-headings-for-smart-answers-outcome

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Convert a tags to buttons on load in header-navigation js ([PR #2235](https://github.com/alphagov/govuk_publishing_components/pull/2235))
 * Allow label sizes for search component label ([PR #2236](https://github.com/alphagov/govuk_publishing_components/pull/2236)) MINOR
 * Update notice component to better match DS notification banner ([PR #2214](https://github.com/alphagov/govuk_publishing_components/pull/2214))
+* Add option to nest context within the h1 for title component ([PR #2226](https://github.com/alphagov/govuk_publishing_components/pull/2226))
 
 ## 25.0.0
 
--- a/app/views/govuk_publishing_components/components/_title.html.erb
+++ b/app/views/govuk_publishing_components/components/_title.html.erb
@@ -3,6 +3,7 @@
 
   context ||= false
   context_locale ||= false
+  context_inside ||= false
 
   inverse ||= false
   local_assigns[:margin_top] ||= 8
@@ -18,13 +19,22 @@
   heading_classes = %w[gem-c-title__text]
   heading_classes << (average_title_length.present? ? 'govuk-heading-l' : 'govuk-heading-xl')
 %>
+
+<% @context_block = capture do %>
+  <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
+    <%= context %>
+  </span>
+<% end %>
+
 <%= content_tag(:div, class: classes) do %>
-  <% if context %>
-    <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
-      <%= context %>
-    </span>
+  <% if context && !context_inside %>
+    <%= @context_block %>
   <% end %>
+  
   <h1 class="<%= heading_classes.join(" ") %>">
+    <% if context && context_inside %>
+      <%= @context_block %>
+    <% end %>
     <%= title %>
   </h1>
 <% end %>
--- a/app/views/govuk_publishing_components/components/docs/title.yml
+++ b/app/views/govuk_publishing_components/components/docs/title.yml
@@ -24,6 +24,13 @@ examples:
     data:
       context: Publication
       title: My page title
+  with_context_inside:
+    description: |
+      If the context should be considered part of the page heading, you can nest the context within the <code><h1></code>.
+    data:
+      context: Publication
+      title: My page title
+      context_inside: true
   with_context_language_labelled:
     description: |
       Sometimes this component appears on a page that has been translated. The title will naturally be supplied in the required language but the context string may fall back to the default. In these instances we need to label the language so the page remains semantic and screenreaders can handle the switch.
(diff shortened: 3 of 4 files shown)


app/views/govuk_publishing_components/components/_title.html.erb
```diff
@@ -22,12 +22,12 @@
...
 <%= content_tag(:div, class: classes) do %>
-  <% if context %>
+  <h1 class="<%= heading_classes.join(" ") %>">
+    <% if context %>
     <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
       <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
     </span>
-  <% end %>
-  <h1 class="<%= heading_classes.join(" ") %>">
+    <% end %>
```
andysellick, line 30: Nitpick: the span inside this `if` statement should be indented.

andysellick (approved):
Looks good. One small thing - the PR description doesn't explicitly state which component this change refers to, I had to figure it out from the files changed.

maxgds (commented):
I feel like this ability to put the context within the H1 should be an option rather than the default. The design system says "If the caption should be considered part of the page heading, you can also nest the caption within the h1." and I'm not sure that we want to retroactively update all instances of it that are on the site already - but I'm not sure that we don't, either.  I'm going to open a discussion on the govuk frontenders Slack channel.

jon-kirwan:
Thanks all. @andysellick. I've updated the PR to indicate the title component has been updated and have amended indentation in the template too.

@maxgds - i've now added this as an option instead so it can be implemented like so -

```ruby
<%= render "govuk_publishing_components/components/title", {
  context: "Publication",
  title: "My page title",
  context_inside: true
} %>
```

I've also updated the docs and tests.

I'm still quite new to Ruby so would appreciate any feedback on a better way to reference the context (which is currently repeated for both context options) -

```ruby
<span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
  <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
</span>
```

maxgds:
Looking good. I can see there are going to be some conflicts with the code I updated yesterday which removes the context href, so you're probably going to need a rebase against master and might have to untangle that a bit, sorry.

In terms of the repetitive context, you could use a content_for block to define it once in the template, see docs here:
https://apidock.com/rails/ActionView/Helpers/CaptureHelper/content_for

I'd also suggest a tweak to your tests - in each one checking for context I'd add a negative assertion for the other position - so we know it just appears once. Docs for that are here: https://apidock.com/rails/ActionController/Assertions/SelectorAssertions/assert_select

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add option to nest the context within the `<h1>` ([PR #2226](https://github.com/alphagov/govuk_publishing_components/pull/2226))
```
andysellick, line 12: This should match the PR title.

Gemfile.lock
```diff
@@ -74,7 +74,7 @@ GEM
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
     ast (2.4.2)
-    brakeman (5.0.4)
+    brakeman (5.1.1)
```
andysellick, line 77: Changes to dependencies in `Gemfile.lock` shouldn't be included in a PR to change a component unless those dependencies are related to the change - which in this case I think they're not. Can these changes be removed?

app/views/govuk_publishing_components/components/_title.html.erb
```diff
@@ -21,13 +22,20 @@
...
+  <% if context && !context_inside %>
     <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
       <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
     </span>
   <% end %>
+  
   <h1 class="<%= heading_classes.join(" ") %>">
+    <% if context && context_inside %>
+      <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
+        <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
```
andysellick, line 36: It looks like these lines are the same as lines 28-30? In which case you can use the rails `capture` method, something like this:

```
<% context_block = capture do %>
    <span class="govuk-caption-xl gem-c-title__context" <%= "lang=#{context_locale}" if context_locale.present? %>>
      <%= context_href ? link_to(context_text, context_href, class: 'gem-c-title__context-link govuk-link', data: context_data) : context_text %>
    </span>
<% end %>

...

<% if context && !context_inside %>
  <%= context_block %> 
<% end %>

<h1>
  <% if context &&  context_inside %>
    <%= context_block %>
  <% end %>
</h1>

```

app/views/govuk_publishing_components/components/docs/title.yml
```diff
@@ -24,6 +24,13 @@ examples:
     data:
       context: Publication
       title: My page title
+  with_context_inside:
+    description: |
+      If the context should be considered part of the page heading, you can also nest the context within the <code><h1></code>.
```
andysellick, line 29: Nitpick: 'also' is unnecessary, `...you can nest the context within...`

spec/components/title_spec.rb
```diff
@@ -21,6 +21,11 @@ def component_name
     assert_select ".govuk-caption-xl", text: "Format"
   end
 
+  it "title context inside appears" do
```
andysellick, line 24: `title context appears inside` maybe?

jon-kirwan:
Thanks both. All changes above have now been added / amended.

maxgds (approved):
Looks good to me!

PR #2227, opened 2021-07-23
Fix issue where priority_brexit_taxon is nil

theseanything:
This prevents an error from being thrown when `priority_brexit_taxon` is nil because there isn't relevant taxon tagged to the content item


chao-xian (approved):

danacotoran:
Could you add a changelog entry before you merge this please @theseanything

Issue #2103, opened 2021-05-28
Cookie banner problem in IE10


andysellick:
Parts of the cookie banner that should be shown, hidden and updated as users interact with it aren't working properly in Internet Explorer 10. I think some of it is due to reliance on the `hidden` attribute, which [isn't supported in IE10](https://caniuse.com/hidden) but there may be other causes.

https://user-images.githubusercontent.com/861310/119979964-8963c380-bfb3-11eb-8e08-aedf30e8724e.mov

danacotoran:
Should be fixed with the release of https://github.com/alphagov/govuk_publishing_components/pull/2231 so will be closing

danacotoran:
#2231 had to be reverted, so I've reopened this for now

MartinJJones:
Closing this issue as JavaScript no longer runs in IE10

PR #2231, opened 2021-07-26
Fix cookie banner issue (IE10)

danacotoran:
It seems that IE10 does not support the `hidden` attribute. 

This is causing some issues with the cookie banner component: when the confirmation banner is visible (upon rejecting/accepting cookies), the cookie choice panel should disappear, which does not happen in <IE10.

This attempts to fix the issue by using CSS to hide elements accordingly when the `hidden` attribute is present (as opposed to toggling `display` with JS).

Additionally, switch from using property dot notation to set hidden to true/false, to using `setAttribute` because although the former worked in other major browsers, IE10 seemed to ignore it (perhaps due to the distinction between [properties](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/hidden) and [attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden)).

-----

More details and a lovely visual of the bug in action on the [original issue](https://github.com/alphagov/govuk_publishing_components/issues/2103).

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -27,13 +27,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.$acceptCookiesButton = this.$module.querySelector('button[data-accept-cookies]')
     if (this.$acceptCookiesButton) {
-      this.$acceptCookiesButton.style.display = 'block'
       this.$acceptCookiesButton.addEventListener('click', this.$module.setCookieConsent)
     }
 
     this.$rejectCookiesButton = this.$module.querySelector('button[data-reject-cookies]')
     if (this.$rejectCookiesButton) {
-      this.$rejectCookiesButton.style.display = 'block'
       this.$rejectCookiesButton.addEventListener('click', this.$module.rejectCookieConsent)
     }
 
@@ -46,7 +44,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var shouldHaveCookieMessage = (this.$module && window.GOVUK.cookie('cookies_preferences_set') !== 'true')
 
       if (shouldHaveCookieMessage) {
-        this.$module.style.display = 'block'
+        this.$module.removeAttribute('hidden')
 
         // Set the default consent cookie if it isn't already present
         if (!window.GOVUK.cookie('cookies_policy')) {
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -101,6 +101,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   CookieBanner.prototype.showConfirmationMessage = function () {
     this.$cookieBannerMainContent = document.querySelector('.js-banner-wrapper')
 
+    this.$cookieBannerMainContent.style.display = 'none'
```
injms, line 104: Rather than doing this in the JavaScript, would adding:

```css
[hidden] {
  display: none;
}
```
to the CSS work?

36degrees: This should all already be handled by this code in GOV.UK Frontend:

https://github.com/alphagov/govuk-frontend/blob/72e9d6e43cffa7db79477343a2cf931812f9b7c7/src/govuk/components/cookie-banner/_index.scss#L20-L24

I think the issue is that you're showing and hiding the cookie banner by setting `display: block` rather than by toggling the `hidden` attribute. The `display: block;` inline styles is then overriding the CSS.

If you consistently use the `hidden` attribute then this should all 'just work'.

There's a 'test' client-side implementation here if that helps (it's a lot simpler as it doesn't actually read or store cookies):

https://github.com/alphagov/govuk-frontend/blob/72e9d6e43cffa7db79477343a2cf931812f9b7c7/app/views/full-page-examples/cookie-banner-client-side/index.njk#L184-L223

andysellick: According to caniuse `hidden` isn't supported by IE10? https://caniuse.com/hidden

36degrees: IE10 and older have no _native_ support for the attribute, but the `[hidden]` attribute selector works just fine in CSS, which means we can make it work with just:

```css
.govuk-cookie-banner[hidden] {
  display: none;
}
```

The implementation in GOV.UK Frontend was tested all the way back to IE8 as part of https://github.com/alphagov/govuk-frontend/pull/2126.

danacotoran: thanks all, sounds like this needs some reworking!

PR #2235, opened 2021-07-28
Convert a tags to buttons on load in header-navigation js

owenatgov:
## What
Adds a check in `search-navigation.js` which asserts if the given element is an `a` tag. If so, it converts it into a `button`.

## Why
- Using `a` tags by default as interactable elements is poor accessibility as it is a misuse of `a` tags, as they aren't being used like links as intended
- Allows us to more easily specify different no-js and js instances of a given UI eg: a search button that is a link to `/search` for no-js but reveals a search bar on click for js enabled.

No visual changes

[Card](https://trello.com/c/8unZtntm/341-fix-the-search-link-in-the-no-js-mobile-version-of-the-old-menu-bar)


app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
```diff
@@ -7,12 +7,31 @@
...
+        var buttonText = thisEl.getAttribute('data-button-text')
+
+        button.className = thisEl.className
+
+        button.setAttribute('data-search-toggle-for', href ? href.substr(1) : thisEl.getAttribute('data-search-toggle-for'))
+        button.innerText = buttonText || thisEl.innerText
+
+        thisEl.parentNode.replaceChild(button, thisEl)
+        thisEl = button
+      }
```
maxf, line 30: Any other attributes that might have to be copied over in the future (like tracking attributes)?

owenatgov: That's a good shout. Ideally we'd take everything and maybe do some clean-up. Will investigate.

owenatgov: Reckon I've found a solution. Let me know what you think. Being bold and resolving this for now.

maxf:
How about tests? Is there anything that will check that the element is getting properly replaced along with its attributes?

owenatgov:
@maxf Tests for this module are a bit of a complicated problem at the moment. She the comment in the spec in it's present state for more details https://github.com/alphagov/govuk_publishing_components/blob/master/spec/javascripts/components/header-navigation-spec.js#L61-L68 I can add tests for brevity none the less.

maxf (approved):

owenatgov:
Added some test for the purpose of documentation

chao-xian:
Great work Owen! Thanks for doing this. Small thing for future: the `Why` section in the PR description is perfect for the body of the commit message. This follows the convention set out in [this excellent blogpost](https://chris.beams.io/posts/git-commit/#why-not-how) (and is linked to from [GDS Way guidance](https://gds-way.cloudapps.digital/standards/source-code.html#commit-messages)) - so basically it's our house style.

chao-xian:
Sorry - one more: is there a corresponding issue for the skipped tests or card anywhere to address it? Is migrating to GOV.UK.Modules part of fixing tech debt in our JS?

owenatgov:
@chao-xian Looking through the issues in this repo there doesn't appear to be any tracking (@injms please correct me if I'm wrong once you're back tomoz). I can throw one together today!

PR #2236, opened 2021-07-28
Allow label sizes in search component

injms:
## What

The search component doesn't allow for the text size of the label to be adjusted. This allows the `label_size` parameter that will alter the label tet size, as long as it's set to `xl`, `l`, `m`, or `s`.

This doesn't override the current default - if no label size is given, then this
component's label will be the same as before.

Setting `label_size` forces `inline_label` to be `false`.

## Why

The explore header needed a differently sized label.

There was a requirement to set the text size of the label - but without overriding the current default.

## Visual Changes

No changes to existing versions.

`xl` version:
image

`l` version:
image

`m` version:
image

`s` version:
image

### DIFF
--- a/app/views/govuk_publishing_components/components/_radio.html.erb
+++ b/app/views/govuk_publishing_components/components/_radio.html.erb
@@ -15,12 +15,12 @@
   heading_classes = %w(govuk-fieldset__heading)
   heading_classes << "gem-c-radio__heading-text" if heading_level == 'h1'
 
-  if ['s', 'm', 'l', 'xl'].include?(heading_size)
+  if (shared_helper.valid_heading_size?(heading_size))
     size = heading_size
-  elsif heading_level == 'h1'
-    size = 'xl'
+  elsif heading_level == "h1"
+    size = "xl"
   else
-    size = 'm'
+    size = "m"
   end
 
   description ||= nil
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -4,6 +4,7 @@
   aria_controls ||= nil
   button_text ||= t("components.search_box.search_button")
   id ||= "search-main-" + SecureRandom.hex(4)
+  label_size ||= nil
   label_text ||= t("components.search_box.label")
   name ||= "q"
   no_border ||= false
@@ -12,6 +13,13 @@
 
   data_attributes ||= {}
   data_attributes[:module] = 'gem-track-click'
+
+  label_classes = []
+  if (["xl", "l", "m", "s"].include?(label_size))
+    label_classes << "govuk-label govuk-label--#{label_size}"
+  else
+    label_classes << "gem-c-search__label"
+  end
 
   classes = %w[gem-c-search govuk-!-display-none-print]
   classes << (shared_helper.get_margin_top)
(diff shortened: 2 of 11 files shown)


app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -13,6 +14,13 @@
   data_attributes ||= {}
   data_attributes[:module] = 'gem-track-click'
 
+  label_classes = []
+  if (["xl", "l", "m", "s"].include?(label_size))
```
andysellick, line 18: There's a method in the shared_helper that does something very similar to this... it might be worth considering if it could be reused? https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/shared_helper.rb#L34

injms: The most I think that this can be abstracted out is a method to have once place to check if the size is correct, since the sizes are appended to a number of different classes in the fieldset, input, label, radio, and select components. Otherwise we'd have a "shared" helper for each component - one for `govuk-label--X`, one for `govuk-fieldset--X` etc - which seems slightly pointless.

I've added a helper method and updated the components, including the search component, to use it: https://github.com/alphagov/govuk_publishing_components/blob/06f063931cd756367c9fe88ff0cc3477a871fc92/lib/govuk_publishing_components/presenters/shared_helper.rb#L34-L36

andysellick: That looks good.

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -15,12 +15,12 @@
...
-  if ['s', 'm', 'l', 'xl'].include?(heading_size)
+  if (shared_helper.valid_heading_size?(heading_size))
     size = heading_size
-  elsif heading_level == 'h1'
-    size = 'xl'
+  elsif heading_level == "h1"
+    size = "xl"
   else
-    size = 'm'
+    size = "m"
```
andysellick, line 23: Suggestion for brevity (I haven't checked that this works though)

```
size = "m"
size = heading_size if (shared_helper.valid_heading_size?(heading_size))
size = "xl" if heading_level == "h1"
```

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -13,6 +14,13 @@
   data_attributes ||= {}
   data_attributes[:module] = 'gem-track-click'
 
+  label_classes = []
+  if (shared_helper.valid_heading_size?(label_size))
+    label_classes << "govuk-label govuk-label--#{label_size}"
+  else
+    label_classes << "gem-c-search__label"
+  end
```
andysellick, line 22: Sorry, don't want to throw a spanner in the works late in the day but have you checked this option when against a dark background? If we discard the `gem-c-search__label` class for the custom label size then the stuff that's already in place doesn't get used, and the label has black text, which for smaller text sizes is a contrast fail.

You might have to remove the size styling from `gem-c-search__label` and put in a default govuk label size, in order to keep that class by default.

app/views/govuk_publishing_components/components/_radio.html.erb
injms: I think a refactor is outside the scope of this pull request. I don't really want to rewrite things that aren't the search component, aside from adding in the helper.

app/views/govuk_publishing_components/components/_search.html.erb
chao-xian: Thanks for pointing this out @andysellick. We're not of any instance where this happens. If it does crop up we'll fix it of course.

injms: My preferred option would be to completely remove the `gem-c-search__label` styles and switch to using the Design System styles - but that'd be a breaking change and not one that I'm keen to embark on at the moment.

This shouldn't be an issue as all of the current uses of this component don't set at `label_size` parameter, so they'll retain the `gem-c-search__label` styles.

PR #2237, opened 2021-07-29
Simplify the modules code

andysellick:
## What
- should do exactly the same thing, but be slightly easier to read and understand

## Why
Discussions around this subject recently revealed confusion between the two `if` statements in this code, specifically what the difference between them is - it's hard to spot the `!` in the middle of the rest of the statements. Restructuring it like this hopefully makes it clearer what the code is doing.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -30,27 +30,21 @@
         var moduleNames = element.data('module').split(' ')
 
         for (var j = 0, k = moduleNames.length; j < k; j++) {
-          var module
           var moduleName = camelCaseAndCapitalise(moduleNames[j])
           var started = element.data(moduleNames[j] + '-module-started')
 
-          if ( // GOV.UK Publishing & Legacy Modules
-            typeof GOVUK.Modules[moduleName] === 'function' &&
-            !GOVUK.Modules[moduleName].prototype.init &&
-            !started
-          ) {
-            module = new GOVUK.Modules[moduleName]()
-            module.start(element)
-            element.data(moduleNames[j] + '-module-started', true)
-          }
+          if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
+            // GOV.UK Legacy Modules using jQuery
+            if (!GOVUK.Modules[moduleName].prototype.init) {
+              new GOVUK.Modules[moduleName]().start(element)
+              element.data(moduleNames[j] + '-module-started', true)
+            }
 
-          if ( // GOV.UK Frontend Modules
-            typeof GOVUK.Modules[moduleName] === 'function' &&
-            GOVUK.Modules[moduleName].prototype.init &&
-            !started
-          ) {
-            module = new GOVUK.Modules[moduleName](element[0]).init()
-            element.data(moduleNames[j] + '-module-started', true)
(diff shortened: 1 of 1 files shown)


app/assets/javascripts/govuk_publishing_components/modules.js
```diff
@@ -34,23 +34,19 @@
...
-            GOVUK.Modules[moduleName].prototype.init &&
-            !started
-          ) {
-            module = new GOVUK.Modules[moduleName](element[0]).init()
-            element.data(moduleNames[j] + '-module-started', true)
+            // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend Modules
+            if (GOVUK.Modules[moduleName].prototype.init) {
+              module = new GOVUK.Modules[moduleName](element[0]).init()
+              element.data(moduleNames[j] + '-module-started', true)
+            }
```
kevindew, line 49: Just a couple of quick suggestions to make this simpler still:

```js    
              // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend Modules
              new GOVUK.Modules[moduleName](element[0]).init()
            } else {
              // GOV.UK Legacy Modules using jQuery
              new GOVUK.Modules[moduleName]().start(element)
            }
            element.data(moduleNames[j] + '-module-started', true)
```

You'd also be able to remove the [`var module` from line 33](https://github.com/alphagov/govuk_publishing_components/blob/72c9be4f7b562783ae68c6b00ef626e5d207d5b5/app/assets/javascripts/govuk_publishing_components/modules.js#L33) since we don't actually do anything with it.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Simplify the modules code ([PR #2237](https://github.com/alphagov/govuk_publishing_components/pull/2237))
```
kevindew, line 12: A changelog entry isn't necessary as this is an internal change: https://github.com/alphagov/govuk_publishing_components/blob/master/CHANGELOG.md#changelog

kevindew (commented):
Looks great - nice improvement. I've added a couple of suggestions.

app/assets/javascripts/govuk_publishing_components/modules.js
andysellick: I've implemented part of your suggestion, but I'm hesitant about switching from two `if` statements to an `if else` - the original code seemed quite explicit about that. I'm fairly sure there shouldn't be a module that doesn't have either a `start` or an `init` function, but assuming that is something I'd be nervous about.

kevindew: Thanks, there isn't a reason to be hesitant though the code you've got is just an else in disguise. As you have:

`if (!GOVUK.Modules[moduleName].prototype.init) {` and `if (GOVUK.Modules[moduleName].prototype.init) {`

one will definitely run, if one of those conditions resolves to true the other will resolve to false 

The previous code also catered for whether things were started and whether a module existed. Therefore, it was possible for neither of these to run. This is now covered by the conditional on line 36.

andysellick: I checked with @alex-ju and the reason for having two `if` statements is because we now allow more than one module to be declared on an element, and they could be of different types. Having two `if` statements allows them both to run.

alex-ju (approved):

PR #2243, opened 2021-08-05
Add new analytics tags for the super navigation header

owenatgov:
## What
Add new analytics tags for the super navigation header.

## Why
So that we can track user engagement with the new navigation header during the upcoming A/B test.

Note that the following tags have yet to be added as they are waiting on https://github.com/alphagov/govuk_publishing_components/pull/2223 to be finished so that they can be added into the js:

- `menuClosed`
- `menuOpened`
- `topicsClosed`
- `topicsOpened`
- `governmentactivityClosed`
- `governmentactivityOpened`
- `searchClosed`
- `searchOpened`

No visual changes.

[Card](https://trello.com/c/tzaEtEhO/320-implement-analytics-tracking-on-the-new-menu-bar-header)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Add new analytics tags for the super navigation header ([PR #2243](https://github.com/alphagov/govuk_publishing_components/pull/2243))
+
 ## 25.1.0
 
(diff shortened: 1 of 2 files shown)


owenatgov:
Closing this for now as it makes much more sense to branch directly from https://github.com/alphagov/govuk_publishing_components/pull/2223 instead of trying to go around the houses after it gets merged in.

PR #2245, opened 2021-08-05
🚧  [DO NOT MERGE] Add account layout wrapper component

danacotoran:
**Please ignore this PR for now** – exploring `layout_for_public` based alternative as suggested by @alex-ju 

-----

Some account-related components, contents, and CSS are repeated across different apps on GOVUK.

The first instance of this was when we [relocated the account homepage from the govuk-account-manager-app to the frontend app](https://github.com/alphagov/frontend/pull/2841), which meant that some of the page furniture and content was duplicated between those apps.

More recently, the accounts team are working towards the goal of single page notifications which will enable users who have linked their email notifications address and their GOVUK account email address to manage their email subscriptions through their account. 

Part of this work is making the email management pages (rendered by `email-alert-frontend`) look the same as the account manager when the user is authenticated with a linked account. Preview:

Screenshot 2021-08-05 at 18 02 13

That would make it three separate apps where page furniture and content are repeated to achieve the same layout.

The purpose of this PR is to nip this issue in the bud by creating an account layout/wrapper component which can be shared between different apps. 
This will mean that the content, the CSS, the markup for all the different building blocks that make up the account layout will live in the gem which should help us prevent inconsistencies between different applications using this layout.

Example of usage: https://github.com/alphagov/email-alert-frontend/pull/1106/files?diff=split&w=1#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3eb

https://trello.com/c/F0ZDY0RC

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add account layout wrapper component ([PR #2245](https://github.com/alphagov/govuk_publishing_components/pull/2245))
+
 ## 25.1.0
 
 * Convert a tags to buttons on load in header-navigation js ([PR #2235](https://github.com/alphagov/govuk_publishing_components/pull/2235))
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -12,6 +12,7 @@ $govuk-new-link-styles: true;
 
 // components
 @import "components/accordion";
+@import "components/account-layout";
 @import "components/action-link";
 @import "components/attachment";
 @import "components/attachment-link";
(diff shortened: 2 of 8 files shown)


alex-ju:
Could this be a variation of [gem_layout_account](https://github.com/alphagov/static/blob/main/app/views/root/gem_layout_account.html.erb) with some optional subcomponents (e.g. the new side navigation and feedback banner)

danacotoran:
@alex-ju I think I understand what you're suggesting. I considered extending `layout_for_public` with the optional account related subcomponents too which sounded ideal indeed... but then I realised that the `govuk-account-manager-prototype` app itself doesn't use `layout_for_public` and I'm not quite sure why hahhhh. I think it was an issue of the cookie banner deviating from the one on GOVUK. I'm gonna draft this PR and have a go at the alternative solution.

danacotoran:
Closing this as I've gone with a different solution as suggested in the previous comment  https://github.com/alphagov/govuk_publishing_components/pull/2255

PR #2250, opened 2021-08-09
Update super navigation styles to avoid `govuk-layout` clashes

injms:
## What

 - Adds more specific selectors to override the generic `a:link` styles coming from `govuk-layout`.
  - Swaps `rem` for `px` as the base font size for the `core_layout` template is different to the base font size in the `gem_layout` template.

## Why

Frontend applications that use the `core_layout` template were seeing blue links and tiny icons in the header - this fixes those problems so that the different templates can use the super navigation header and have it look consistent.

## Visual Changes

Before:
image

After:
image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -1,15 +1,15 @@
 $search-icon-size: 20px;
 
 @mixin chevron($colour) {
-  border-bottom: .2rem solid govuk-colour($colour);
-  border-right: .2rem solid govuk-colour($colour);
+  border-bottom: 3px solid govuk-colour($colour);
+  border-right: 3px solid govuk-colour($colour);
   content: " ";
   display: inline-block;
-  height: .5rem;
-  margin: 0 .5em 0 .15em;
+  height: 8px;
+  margin: 0 8px 0 2px;
   transform: translateY(-35%) rotate(45deg);
   vertical-align: middle;
-  width: .5rem;
+  width: 8px;
 }
 
 %top-level-navigation-link-base {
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -24,8 +24,9 @@ $search-icon-size: 20px;
   text-decoration: none;
 
   &,
+  &:link,
   &:visited {
-    color: govuk-colour("blue"); // FIXME: contrast against govuk-colour("light-grey") not high enough
+    color: $govuk-link-colour;
```
owenatgov, line 29: Should we keep that comment?

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -128,8 +130,8 @@ $search-icon-size: 20px;
 
     &:before,
     &:after {
-      $width: 1.2rem;
-      $height: .2rem;
+      $width: 20px;
```
owenatgov, line 133: Can we use `govuk-spacing` here?

```suggestion
      $width: govuk-spacing(4);
```

owenatgov (approved):
Some bothersome comments but not blockers necessarily

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
injms: I checked and the contrast of the blue against the light grey is 4.62, so it's okay 👍

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
injms: Sure thing - done: 

https://github.com/alphagov/govuk_publishing_components/blob/9a87c5d73cca8f1070f647452db00abb26b3a254/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss#L133

owenatgov:
We're good we're good we're good. Just needs a quick changelog update and then 🚀

PR #2255, opened 2021-08-09
Extend layout for public with account components

danacotoran:
## What
Update `layout_for_public` component to support GOVUK Account components.

## Why

Some account-related components, contents, and CSS are repeated across different apps on GOVUK.

The first instance of this was when we [relocated the account homepage from the govuk-account-manager-app to the frontend app](https://github.com/alphagov/frontend/pull/2841), which meant that some of the page furniture and content was duplicated between those apps.

More recently, the accounts team are working towards the goal of single page notifications which will enable users who have linked their email notifications address and their GOVUK account email address to manage their email subscriptions through their account. 

Part of this work is making the email management pages (rendered by `email-alert-frontend`) look the same as the account manager when the user is authenticated with a linked account. Preview:

Screenshot 2021-08-05 at 18 02 13

That would make it three separate apps where page furniture and content are repeated to achieve the same layout.

The purpose of this PR is to extend `layout_for_public` so that it can be used as a "shell" for account (or "account-style" pages).

## Visual changes

Existing variations of the component will remain unchanged. A few options have been introduced to support the main GOVUK Account layout variations. The intention is for this to be used with a layout template in `static` (or, in the case of the `govuk-account-manager-prototype` app, be used directly in the application layout)

----

### Related:
- creates the template in `static`:  https://github.com/alphagov/static/pull/2571 
- applies the account layout to select pages on  `email-alert-frontend`: https://github.com/alphagov/email-alert-frontend/pull/1106 

https://trello.com/c/F0ZDY0RC

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
@@ -11,3 +11,55 @@
   margin-top: - govuk-spacing(2);
   position: relative;
 }
+
+.gem-c-layout-for-public-account-menu {
+  margin: 0 0 govuk-spacing(6) 0;
+  padding: 0;
+
+  @include govuk-media-query($from: desktop) {
+    margin-right: govuk-spacing(4);
+  }
+}
+
+// Used to determine width of blue location indicator in navigation mention
+$current-indicator-width: 4px;
+
+.gem-c-layout-for-public-account-menu__item {
+  list-style-type: none;
+  padding: 12px 0;
+
+  @include govuk-media-query($from: tablet) {
+    margin: 0 0 govuk-spacing(4);
+    padding: govuk-spacing(1) 0;
+  }
+}
+
+.gem-c-layout-for-public-account-menu__item--current {
+  margin-left: -(govuk-spacing(2) + $current-indicator-width);
+  padding-left: govuk-spacing(2);
+  border-left: $current-indicator-width solid govuk-colour("blue");
+
+  .gem-c-layout-for-public-account-menu__link {
+    text-decoration: none;
+  }
+}
+
+.gem-c-layout-for-public-account-menu__link {
+  @include govuk-font(19, $weight: "bold");
+
+  &:not(:focus):hover {
+    color: $govuk-link-colour;
+  }
+}
+
+.gem-c-layout-for-public-account-feedback-footer {
+  margin-bottom: govuk-spacing(4);
+  padding: govuk-spacing(3);
+  background: govuk-colour("light-grey");
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(6);
+    margin-top: govuk-spacing(2);
+    margin-bottom: govuk-spacing(8);
+  }
+}
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -14,6 +14,12 @@
   show_search = local_assigns.include?(:show_search) ? local_assigns[:show_search] : true
   title ||= "GOV.UK - The best place to find government services and information"
 
+  show_account_layout ||= false
+  account_nav_location ||= nil
+  omit_account_feedback_footer ||= false
+  omit_account_navigation ||= false
+  omit_account_phase_banner ||= false
+
 # This is a hack - but it's the only way I can find to not have two blue bars on
 # constrained width layouts.
 #
@@ -74,7 +80,7 @@
       href: "#content"
     } %>
 
-    <%= render "govuk_publishing_components/components/cookie_banner" %>
+    <%= render "govuk_publishing_components/components/cookie_banner", layout_helper.cookie_banner_data %>
 
     <% unless omit_header %>
       <% if show_explore_header %>
@@ -107,14 +113,27 @@
       <% end %>
     <% end %>
 
-    <div id="wrapper" class="<%= "govuk-width-container" unless full_width %>">
-      <%= yield :before_content %>
-      <main class="govuk-main-wrapper" id="content">
+    <% if show_account_layout %>
+      <%= render "govuk_publishing_components/components/layout_for_public/account-layout", {
+        omit_account_phase_banner: omit_account_phase_banner,
+        omit_account_navigation: omit_account_navigation,
+        account_nav_location: account_nav_location,
+      } do %>
+        <%= yield :before_content %>
         <%= yield %>
-      </main>
-    </div>
+      <% end %>
+    <% else %>
+      <div id="wrapper" class="<%= "govuk-width-container" unless full_width %>">
+        <%= yield :before_content %>
+        <main class="govuk-main-wrapper" id="content">
+          <%= yield %>
+        </main>
+      </div>
+    <% end %>
 
-    <% unless omit_feedback_form %>
+    <% if show_account_layout && !omit_account_feedback_footer %>
+      <%= render "govuk_publishing_components/components/layout_for_public/account-feedback-footer" %>
+    <% elsif !omit_feedback_form %>
       <div class="govuk-width-container">
         <%= render "govuk_publishing_components/components/feedback" %>
       </div>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 # Unreleased
 
 * Fix cookie banner issue (IE10) ([PR #2231](https://github.com/alphagov/govuk_publishing_components/pull/2231))
+* Extend layout for public with account components ([PR #2255](https://github.com/alphagov/govuk_publishing_components/pull/2255))
 
 # 25.2.3
 
--- a/app/views/govuk_publishing_components/components/docs/layout_for_public.yml
+++ b/app/views/govuk_publishing_components/components/docs/layout_for_public.yml
@@ -61,3 +61,62 @@ examples:
     data:
       emergency_banner: <div class="govuk-!-padding-top-3 govuk-!-padding-bottom-3">This is the emergency banner slot</div>
       global_bar: <div class="govuk-!-padding-top-5 govuk-!-padding-bottom-3">This is the global bar slot</div>
+  with_account_layout_enabled:
+    description: Adds account layout wrapper around the content passed in to the component
+    data:
+      show_account_layout: true
+      block: |
+        <h2 class="govuk-heading-l">This is a title</h2>
+        <p class="govuk-body">This is some body text with <a href="https://example.com">a link</a>.
+  with_account_layout_but_without_account_feedback_footer:
+    description: The account layout renders with an account-specific feedback banner by default. This flag allows this banner to be omitted.
+    data:
+      show_account_layout: true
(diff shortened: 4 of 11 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
```diff
@@ -11,3 +11,55 @@
   margin-top: - govuk-spacing(2);
   position: relative;
 }
+
```
injms, line 14: Would these extra styles that are only for the account things be better placed in their own stylesheet? That way it wouldn't be loaded for applications that don't currently need it.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
       <% end %>
     <% end %>
 
-    <div id="wrapper" class="<%= "govuk-width-container" unless full_width %>">
-      <%= yield :before_content %>
-      <main class="govuk-main-wrapper" id="content">
-        <%= yield %>
-      </main>
-    </div>
+    <% if show_account_layout %>
```
injms, line 116: Could the new markup be placed in a partial that's then included here? Appreciate that this could be a premature optimisation though.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
+            margin_bottom: 4,
+          } %>
+
+          <p class="govuk-body govuk-!-margin-bottom-0">
+            <%= t("components.layout_for_public.account_layout.feedback.banners.footer_intro") %>
+            <a href="<%= "#{Plek.find('account-manager')}/feedback" %>" class="govuk-link"><%= t("components.layout_for_public.account_layout.feedback.banners.footer_link") %></a>
+            <%= t("components.layout_for_public.account_layout.feedback.banners.footer_outro") %>
+          </p>
+          <% end %>
+      <% end unless omit_account_feedback_footer %>
```
injms, line 170: Minor quibble - I think that this `unless` should be at the start of this `do` block so that both of the conditionals are together. For example:
```ruby
<% if show_account_layout >%
  <% unless omit_account_feedback_footer %>
```

injms (approved):
My comments shouldn't be blockers on this - and I think they're indicative of a possible future problem (do components have too many uses?) that we should solve as a community.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: I've placed some of the markup in partials (it might be premature like you said, but it's much better for readability so I agreed it was worth doing!)

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
-      <%= yield :before_content %>
-      <main class="govuk-main-wrapper" id="content">
-        <%= yield %>
-      </main>
-    </div>
+    <% if show_account_layout %>
+      <div class="<%= "govuk-width-container" unless full_width %>">
+        <% message = capture do %>
+          <%= t("components.layout_for_public.account_layout.feedback.banners.phase_intro") %>
+          <a class="govuk-link" href=<%= "#{Plek.find('account-manager')}/feedback" %>><%= t("components.layout_for_public.account_layout.feedback.banners.phase_link") %></a>
```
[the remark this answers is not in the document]
danacotoran: thanks, good suggestion 👍

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
+          <%= t("components.layout_for_public.account_layout.feedback.banners.phase_outro") %>
+        <% end %>
+
+        <%= render "govuk_publishing_components/components/phase_banner", {
+          phase: "alpha",
+          message: message
+        } unless omit_account_phase_banner %>
+
+
```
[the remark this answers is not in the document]
danacotoran: Sorted

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
+            <div id="wrapper">
+              <%= yield :before_content %>
+              <main role="main" id="content">
+                <%= yield %>
+              </main>
+            </div>
+          </div>
+        </div>
+      </div>
+      <% else %>
```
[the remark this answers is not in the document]
danacotoran: all my sloppy indentation should be fixed now... 🤦🏻‍♀️

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
+          <%= render "govuk_publishing_components/components/heading", {
+            text: t("components.layout_for_public.account_layout.feedback.banners.title"),
+            heading_level: 2,
+            font_size: "m",
+            margin_bottom: 4,
+          } %>
+
+          <p class="govuk-body govuk-!-margin-bottom-0">
+            <%= t("components.layout_for_public.account_layout.feedback.banners.footer_intro") %>
+            <a href="<%= "#{Plek.find('account-manager')}/feedback" %>" class="govuk-link"><%= t("components.layout_for_public.account_layout.feedback.banners.footer_link") %></a>
```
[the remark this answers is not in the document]
danacotoran: Done!

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Updated but I didn't quite do as suggested. I went with this instead:
```
<% if show_account_layout && !omit_account_feedback_footer %>
  <%= render "govuk_publishing_components/components/layout_for_public/account-feedback-footer" %>
<% elsif !omit_feedback_form %>
  <div class="govuk-width-container">
      <%= render "govuk_publishing_components/components/feedback" %>
  </div>
<% end %>
```

app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
danacotoran: I'm in two minds about this suggestion because on the one hand it sounds like a good idea to optimise the styles like this. But on the other hand, the only place `layout-for-public` component styles seem to be imported currently is [in the application stylesheet in static](https://github.com/alphagov/static/blob/80e0b13f860ec538ad4c006662b9c6f804d00b2c/app/assets/stylesheets/application.scss#L12). 

So let's say we split out the accounty stuff into a new `layout-for-public-accounty-stuff.scss` stylesheet – I'm not sure how I would conditionally import that stylesheet in `application.scss` only when the account layout template is being used if you know what I mean? Not sure if we have any previous examples where we've already done something like that.

injms: That makes complete sense - you're right, it'd be a separate stylesheet that'd be included anyway.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -107,14 +113,62 @@
...
+        } unless omit_account_phase_banner %>
+
+
+        <div class="govuk-grid-row govuk-main-wrapper">
+          <%= tag.div(class: 'govuk-grid-column-one-third') do %>
+            <%= render "govuk_publishing_components/components/layout_for_public/account-navigation", { page_is: account_nav_location} %>
+          <% end unless omit_account_navigation %>
+
+          <div class="govuk-grid-column-two-thirds">
+            <div id="wrapper">
```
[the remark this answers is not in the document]
injms: Yep - Slimmer needs the `id` so it knows what parts of the page to swap in and out: [source](https://github.com/alphagov/slimmer/blob/5b8428704f143547bdae2c7cb9a585ab77d3181b/lib/slimmer/processors/body_inserter.rb#L3)

If it's not there, then nothing can be selected from the page - so returns `nil`, which hasn't got the `has_attribute?` method and this error is thrown.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
injms: That's more concise and combines the conditionals - nice.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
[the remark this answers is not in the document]
andysellick: Ah, fair enough then.

andysellick (approved):

app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
```diff
@@ -11,3 +11,55 @@
...
+  margin: 0 0 govuk-spacing(6) 0;
+  padding: 0;
+
+  @include govuk-media-query($from: desktop) {
+    margin-right: govuk-spacing(4);
+  }
+}
+
+// Used to determine width of blue location indicator in navigation mention
+$_current-indicator-width: 4px;
```
alex-ju, line 25: non-blocker: why the `_` notation?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -14,6 +14,12 @@
   show_search = local_assigns.include?(:show_search) ? local_assigns[:show_search] : true
   title ||= "GOV.UK - The best place to find government services and information"
 
+  show_account_layout ||= false
+  account_nav_location ||= nil
```
alex-ju, line 18: can we document at least this `account_nav_location` flag (if not all of them) in the component yaml? It's not clear which values are allowed and how they affect the output.

danacotoran: Alex they should all be documented here [here](https://github.com/alphagov/govuk_publishing_components/blob/cf063218fb57ace309bd057aaac166a0b785e657/app/views/govuk_publishing_components/components/docs/layout_for_public.yml#L78) but to be fair I could probably add better descriptions to be clearer about what they do (it's pretty clear if you look at the component previews though, I think)

app/assets/stylesheets/govuk_publishing_components/components/_layout-for-public.scss
danacotoran: Because it was copied from over [here](https://github.com/alphagov/govuk-account-manager-prototype/blob/6706030ab38d0d16a86274e39db9d61211c1849e/app/assets/stylesheets/accounts.scss#L15)  and that had the `_` notation for some reason 😄  
I can update it to be more consistent with our other sass variables.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
danacotoran: Have added more detail to the component docs to make it clearer what each option is there for!

alex-ju: Thank you! None of them blockers for the release, but good reminder for our future selves.

PR #2256, opened 2021-08-09
Fix final issues with tracking on super nav header

owenatgov:
## What
Add tracking to section heading links when there aren't any section children, thus the link is displayed by default, and fix a typo.

## Why
To make tracking work for govuk explore.

No visual changes.


chao-xian (approved):
Final-final-final 😋

PR #2259, opened 2021-08-10
Fix search button on mobile

alex-ju:
## What
The markup for this button was updated in static [1] and in this repo's mockup markup [2][3], but not in the template itself.

[1] https://github.com/alphagov/static/blob/main/app/views/root/_base.html.erb#L18

[2] https://github.com/alphagov/govuk_publishing_components/blob/master/spec/javascripts/components/header-navigation-spec.js#L24-L26
[3] #2235

## Why
To fix an issue where, on mobile, the search input is not shown when clicking the magnifying glass icon.

## Visual Changes
No visual changes.

Preview the [public_layout component in isolation](https://govuk-publis-fix-search-nh7jfb.herokuapp.com/public) for manual testing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+# Unreleased
+
+* Fix search button on mobile ([PR #2259](https://github.com/alphagov/govuk_publishing_components/pull/2259))
+
+
 # 25.2.3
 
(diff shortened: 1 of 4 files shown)


alex-ju:
The immediate fix was [addressed in static](https://github.com/alphagov/static/pull/2572). Converting this into a draft as it can be addressed more thoroughly (e.g. make sure the anchor is visible in a no-js experience).

PR #2260, opened 2021-08-11
GOVUK.Modules.start can accept DOM elements

kevindew:
I've opened this as a draft to share the idea, if you have an opinion on it feel free to comment.

## What

This adjusts `GOVUK.Modules.find` to be able to handle the situation where the input is not a jQuery object and can instead be a DOM element. This has been done so that GOVUK.Modules.start need not only be called with jQuery wrapped elements.
## Why

The reason for this is to support GOV.UK in its journey away from jQuery. Previously we needed to add jQuery back into code so that this function could be called, whereas this change allows a regular DOM element to be injected.

It will allow us to [change this code](https://github.com/alphagov/finder-frontend/blob/67f01543cb9ce5d5331f423061630acc6619d3c0/app/assets/javascripts/live_search.js#L198)

from `GOVUK.modules.start($(this.$resultsWrapper))` to `GOVUK.modules.start(this.$resultsWrapper)`.

In the longer term I expect that we could deprecate calling this function with a jQuery object and then remove it (or replace `start` with an `init` method to match modules).

## Visual Changes

None


JonathanHallam (approved):
Nicely done, thank you!

PR #2262, opened 2021-08-11
Update search toggle tracking

injms:
## What

Updates the event being fired by the search toggle from `menuOpened` and `menuClosed` to `searchOpened` and `searchClosed`.

It'd be nice to have a more elegant way of doing this, but this fixes a problem that needs to be fixed quickly.

## Why

This will differentiate the events being fired by the menu and the search toggles to allow for more accurate analysis of users interactions.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
@@ -47,16 +47,17 @@
         var isSearchToggle = sourceClass.match('search-toggle')
         var showText = this.getAttribute('data-show-text') || 'Show search'
         var hideText = this.getAttribute('data-hide-text') || 'Hide search'
+        var buttonName = this.getAttribute('data-button-name') || 'menu'
 
         if (targetClass.indexOf('js-visible') !== -1) {
           target.setAttribute('class', targetClass.replace(/(^|\s)js-visible(\s|$)/, ''))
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/lib/header-navigation.js
```diff
@@ -51,12 +51,12 @@
         if (targetClass.indexOf('js-visible') !== -1) {
           target.setAttribute('class', targetClass.replace(/(^|\s)js-visible(\s|$)/, ''))
           if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
-            window.GOVUK.analytics.trackEvent('headerClicked', 'menuClosed', { label: 'none' })
```
owenatgov, line 54: This is sadly going to trigger for menu buttons for core layout applications: code is [here](https://github.com/alphagov/static/blob/main/app/views/root/proposition_menu.html.erb#L3). Example of a page using this is [here](https://www.gov.uk/government/how-government-works).

Are we happy with this as a possible bit of chuff?

injms: Right, well that's fun. I've updated this to read from a `data-name` attribute so it shouldn't confuse between the two, and there's a fallback in case there's no `data-name` present on the button.

owenatgov (approved):
Amazing!

PR #2263, opened 2021-08-11
Extend track click script

andysellick:
## What

Modify `GemTrackClick` so that if no `data-track-label` attribute is provided, the text of the link clicked will be used.

- this will allow tracking to be added to a wrapping element containing links managed outside of an editable template

## Why
We want to add tracking to links within an element where the content is managed outside of the template (e.g. a YML file) and the content is subject to change, without having to keep ensuring the links have the right data attributes. This allows us to add the module, action and event attributes to the parent, and the script will automatically use the text of the clicked link as the label.

## Visual Changes
None.

Trello card: https://trello.com/c/Q6kr2nFK/551-add-ga-tracking-codes-to-fcdo-call-out-box-of-covid-info

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 # Unreleased
 
+* Extend track click script ([PR #2263](https://github.com/alphagov/govuk_publishing_components/pull/2263))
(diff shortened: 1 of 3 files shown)


injms (dismissed):
Neat - looks good to me 👍

injms (approved):

PR #2265, opened 2021-08-12
Fix track click link tracking

andysellick:
## What
Modify the track click script so that link tracking without a label explicitly given should use the link href, not the text of the link.

The logic of the code works as follows:

- if `track-links-only` is specified, only clicks on links are tracked
- link href is acquired and stored, if a label is specified, that is used, otherwise it falls back to the href

If normal tracking is happening:

- link href is acquired and stored (but might not return anything as it might not be a link)
- if a label is specified, that is used, otherwise it falls back to the href, which if it doesn't exist (if it's not a link) no label is included in the GA event

## Why
We want to track links in an element by only putting tracking on the element, not the individual links (as they are content controlled and subject to change). Fixes the behaviour introduced in https://github.com/alphagov/govuk_publishing_components/pull/2263 to use the link href, not the link text.

## Visual Changes
None.

Trello card: https://trello.com/c/Q6kr2nFK/551-add-ga-tracking-codes-to-fcdo-call-out-box-of-covid-info

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/track-click.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/track-click.js
@@ -24,24 +24,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GemTrackClick.prototype.handleClick = function (target) {
     var options = { transport: 'beacon' }
-    var linkText
+    var linkUrl
 
     // if clicked element hasn't got the right attributes, look for a parent that matches
     if (!target.hasAttribute('data-track-category') && !target.hasAttribute('data-track-action')) {
-      linkText = target.textContent || target.innerText
+      linkUrl = target.getAttribute('href')
       target = target.closest('[data-track-category][data-track-action]')
     }
 
     if (target) {
       var category = target.getAttribute('data-track-category')
       var action = target.getAttribute('data-track-action')
-      var label = target.getAttribute('data-track-label')
+      var label = target.getAttribute('data-track-label') || linkUrl
       var value = target.getAttribute('data-track-value')
       var dimension = target.getAttribute('data-track-dimension')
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/track-click.js
```diff
@@ -24,24 +24,26 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   GemTrackClick.prototype.handleClick = function (target) {
     var options = { transport: 'beacon' }
-    var linkText
+    var linkUrl
 
     // if clicked element hasn't got the right attributes, look for a parent that matches
     if (!target.hasAttribute('data-track-category') && !target.hasAttribute('data-track-action')) {
-      linkText = target.textContent || target.innerText
+      linkUrl = target.getAttribute('href')
```
alex-ju, line 31: Do we have an idea of the impact of this change (switching from using the label/text to using the href/url? I'm under the impression that `track-click` is rather widely used across GOV.UK

alex-ju: Ah, I see it's only been [recently introduced](https://github.com/alphagov/govuk_publishing_components/pull/2263).

andysellick: It should have no impact, because this is a feature I added this morning, but got wrong - this PR corrects that.

My previous PR https://github.com/alphagov/govuk_publishing_components/pull/2263 added the ability to include the link text as the label if no label was specified. But it should have been the link href, not the text 🤦

injms (approved):

alex-ju (approved):
I suggest we rename the changelog entry (and the PR title) to something like _Fix `track-click` script extension in #2263_, as the current one feels too generic.

PR #2270, opened 2021-08-17
Add lazy loading

chris-gds:
## What

Adding lazy loading option to [image card](https://components.publishing.service.gov.uk/component-guide/image_card) providing a default to `auto`

## Why

As part of [this PR](https://github.com/alphagov/whitehall/pull/6236) realised that `loading="lazy"` was an attribute that exists on the image. [This had a measurable improvement on performance](https://github.com/alphagov/whitehall/pull/4979). 

In order to maintain use of this attribute _and_ use the [image card componen](https://components.publishing.service.gov.uk/component-guide/image_card)t adding the option to the gem

## Visual Changes
No visual changes, however visual representation of code update below:

Screenshot 2021-08-17 at 11 29 37

---

Screenshot 2021-08-17 at 11 29 18

## Anything else

Ref: https://web.dev/browser-level-image-lazy-loading/
[`auto`] has been added as the "default lazy-loading behaviour of the browser, which is the same as not including the attribute"

### DIFF
--- a/spec/components/image_card_spec.rb
+++ b/spec/components/image_card_spec.rb
@@ -123,4 +123,14 @@ describe "ImageCard", type: :view do
     render_component(href: "#", lang: "cy")
     assert_select ".gem-c-image-card[lang='cy']"
   end
+
+  it "applies lazy loading attribute when lazy is specified" do
+    render_component(href: "#", image_src: "/moo.jpg", image_alt: "some meaningful alt text", image_loading: "lazy")
+    assert_select ".gem-c-image-card__image[loading='lazy']"
+  end
+
+  it "checks image loading attribute is 'auto' when value 'image_loading' is not specified" do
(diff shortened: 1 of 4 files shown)


spec/components/image_card_spec.rb
```diff
@@ -123,4 +123,9 @@ def component_name
     render_component(href: "#", lang: "cy")
     assert_select ".gem-c-image-card[lang='cy']"
   end
+
+  it "applies lazy loading attribute when lazy is specified" do
```
alex-ju, line 127: can we update other tests to reflect that `"auto"` is applied by default?

alex-ju (approved):

spec/components/image_card_spec.rb
chris-gds: Course, added

Issue #2221, opened 2021-07-20
Componentise large numbers


chris-gds:
## What 

"Big numbers" was an element that we believe qualifies for being a Component

### Reasons for doing this:

- This element is being used in several places
- I understand this could be used on more high-traffic places in future.
- It meets much of our criteria to Componentise

## Visuals

![image](https://user-images.githubusercontent.com/71266765/126362279-ee2db59d-01eb-440a-b5e2-a3463e6dce7b.png)

## More context

There was a previous a11y issue that was fixed here [add link]
[This was already highlighted in a presentation done by the Design System](https://docs.google.com/presentation/d/1T1HaQmvzPC2x3eHl8rx1Zfej2eiG70dPeD7ixXroCS0/edit#slide=id.gdaa9f1469c_0_135)
It also appears on a prototype here: https://explore-prototype-4.herokuapp.com/ (DM for opening details)
- https://www.gov.uk/government/statistics/uk-house-price-index-scotland-february-2020/uk-house-price-index-scotland-february-2020
- https://www.gov.uk/government/how-government-works

owenatgov:
# API proposal

## High level
We present the big numbers component as part of a list of values with an optional heading. Each value will be a number, can have an associated "label" (not a literal label tag) and can be a link. The latter 2 attributes of the component are optional.

Users can also communicate that a number is a "minimum" to a given value by adding a "+" to the end of the number eg: 400+ government agencies.

## Example API calls
### A heading, labels and links
```
<%= render "govuk_publishing_components/components/big_number", {
    heading: "Respond to consultations",
    items: [
        {
            number: 116,
            label: "Open consultations",
            href: "/government/publications?publication_filter_option=open-consultations",
        },
        {
            number: 553,
            label: "Closed consultations in the past 12 months",
            href: "/government/publications?publication_filter_option=open-consultations",
        },
    ],
} %>
```

[Live example](https://www.gov.uk/government/get-involved)

### No heading and "plus"
```
<%= render "govuk_publishing_components/components/big_number", {
    items: [
        {
            number: 23,
            label: "Ministerial departments",
            href: "/organisations#ministerial_departments",
        },
        {
            number: 400,
            show_plus: true,
            label: "Other agencies",
            href: "/councils",
        },
    ],
} %>
```

[Live example](https://www.gov.uk/)

### No links or labels
```
<%= render "govuk_publishing_components/components/big_number", {
    heading: "Ministerial departments",
    items: [
        {
[...]

PR #2278, opened 2021-08-20
Add big number component

owenatgov:
## What
Adds the big number component to the component library

## Why
We have a couple of fragmented instances of this pattern across govuk which meets our criteria for componentisation.

You can read more on the reasoning, along with a spec for the API and the markup in the comments, here: https://github.com/alphagov/govuk_publishing_components/issues/2221

## Visual Changes (from bespoke examples on production)
| Link | Before | After |
| --- | --- | --- |
| https://www.gov.uk/ | Screenshot 2021-08-20 at 16 23 15 | Screenshot 2021-08-20 at 16 23 26 |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
@@ -1,0 +1,41 @@
+.gem-c-big-number {
+  margin-bottom: govuk-spacing(3);
+}
+
+.gem-c-big-number__value {
+  @include govuk-font($size: 80, $weight: bold);
+}
+
+.gem-c-big-number__label {
+  @include govuk-font($size: 16, $weight: bold, $line-height: 2);
+
+  // This pseudo element is to bypass an issue with NVDA where block level elements are dictated separately.
+  // What's happening here is that the label and the number technically have an inline relationship but appear to have a block relationship thanks to this element
+  &:before {
+    content: "";
+    display: block;
+  }
+}
+
+.gem-c-big-number__link {
+  display: inline-block;
+  text-decoration: none;
+
+  // Add govuk-link hover decoration to main value if no label present but a href attribute is
+  .gem-c-big-number__value--decorated {
+    @include govuk-link-decoration;
+    @include govuk-link-style-no-underline;
+  }
+
+  .gem-c-big-number__label {
+    @include govuk-link-decoration;
+  }
+
+  &:hover,
+  &:active {
+    .gem-c-big-number__label,
+    .gem-c-big-number__value--decorated {
+      @include govuk-link-hover-decoration;
+    }
+  }
+}
--- a/app/views/govuk_publishing_components/components/_big_number.html.erb
+++ b/app/views/govuk_publishing_components/components/_big_number.html.erb
@@ -1,0 +1,35 @@
+<%
+  href ||= nil
+  number ||= nil
+  label ||= nil
+  href ||= nil
+  data_attributes ||= nil
+  classes = ["gem-c-big-number__value"]
+  
+  if label.nil? && href
+    classes << "gem-c-big-number__value--decorated"
+  end
+%>
+<% if number %>
+  <% big_number_value = capture do %>
+    <%= tag.span class: classes do %>
+      <%= number %>
+    <% end %>
+
+    <% unless label.nil? %>
+      <% # add a virtual space here to handle screen readers printing dictations without a space between the number and the label %>
+      <span class="govuk-visually-hidden">&nbsp;</span>
+      <span class="gem-c-big-number__label">
+        <%= label %>
+      </span>
+    <% end %>
+  <% end %>
+  
+  <%= tag.div class: "gem-c-big-number" do %>
+    <% unless href.nil? %>
+      <%= link_to big_number_value, href, class: "govuk-link gem-c-big-number__link", data: data_attributes %>
+    <% else %>
+      <%= big_number_value %>
+    <% end %>
+  <% end %>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/big_number.yml
+++ b/app/views/govuk_publishing_components/components/docs/big_number.yml
@@ -1,0 +1,50 @@
+name: Big number
+description: A big number, with a label if needed, for visualising quantitative values such as the number of government departments or historical prime ministers.
+body: |
+  This component requires at least a non-falsey `number` attribute value passed to it, otherwise it will not render.
+accessibility_criteria: |
+  This component must:
+
+  - communicate number value and label (if present) in a single dictation when read with a screen reader
+  - convey the meaning of the number shown
+shared_accessibility_criteria:
+- link
+examples:
+  default:
+    data:
+      number: 119
+  with_labels:
+    description: Labels for numbers are given inline styling but stacked using pseudo elements with display block. This is to bypass an issue with NVDA where block level elements are dictated separately.
+    data:
+      number: 119
+      label: Open consultations
+  passing_extra_symbols:
+    description: "In some cases, we want to communicate more than just the flat numeric value eg: 400+, 90%, -20, 1M etc. This is why we allow values to be passed as flat strings."
+    data:
+      number: "400+"
+      label: Other agencies and public bodies
+  with_link:
+    data:
+      number: 23
+      label: Ministerial departments
+      href: "/government/organisations#ministerial_departments"
+  with_link_but_no_label:
+    description: Numbers that are links are underlined unless a label is provided, in which case the label is given the underline.
+    data:
+      number: 23
+      href: "/government/organisations#ministerial_departments"
+  with_data_attributes:
+    description: |
+      These data attributes will only be present if a `href` attribute is present.
+      
+      This will also not automatically apply a `gem-track-click` module attribute if the data attributes pertain to click tracking. Remember to apply this outside the component call in a surrounding element, if using.
+    data:
+      number: 23
+      label: Ministerial departments
+      href: "/government/organisations#ministerial_departments"
+      data_attributes:
+        track-category: homepageClicked
+        track-action: departmentsLink
+        track-label: "/government/organisations#ministerial_departments"
+        track-dimension: 23 Ministerial departments
+        track-dimension-index: 29
--- a/spec/components/big_number_spec.rb
+++ b/spec/components/big_number_spec.rb
@@ -1,0 +1,57 @@
+require "rails_helper"
+
+describe "Big number", type: :view do
+  def component_name
+    "big_number"
+  end
+
+  it "renders nothing when nothing is passed to it" do
+    assert_empty render_component({})
+    assert_empty render_component(number: false)
+  end
+
+  it "renders a number value" do
+    render_component(number: 500)
+
+    assert_select ".gem-c-big-number", count: 1
+  end
+
+  it "renders a number label if a label attribute is present" do
+    render_component({
+      number: 500,
+      label: "test runs",
+    })
+
+    assert_select ".gem-c-big-number__label", text: "test runs"
+  end
+
+  it "wraps a number item in a link if a href attribute is present" do
+    render_component({
+      number: 500,
+      label: "test runs",
+      href: "/tests",
+    })
+
+    assert_select "a.gem-c-big-number__link", count: 1, href: "/tests"
+  end
+
+  it "includes the number decorator class link if a href attribute is present but a label is not present" do
+    render_component({
+      number: 500,
+      href: "/tests",
+    })
+
+    assert_select ".gem-c-big-number__value.gem-c-big-number__value--decorated"
+  end
+
+  # The space mentioned in the below test is to handle screen readers printing dictations without a space between the number and the label
+  # We don't want this to get removed accidentally, hence the following test
+  it "ensures that a visually hidden space is included for screen readers when a label is present" do
+    render_component({
+      number: 500,
+      label: "test runs",
+    })
+
+    assert_select ".govuk-visually-hidden", count: 1
+  end
+end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+# Unreleased
+
+* Add big number component ([PR #2278](https://github.com/alphagov/govuk_publishing_components/pull/2278))
+
 ## 27.0.0
 
 * **BREAKING:** Rename extra_links to extra_details (image card) ([PR #2300](https://github.com/alphagov/govuk_publishing_components/pull/2300))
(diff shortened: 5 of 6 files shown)


app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
+name: Big number
+description: A big number, with a label if needed, for visualising quantitative values such as the number of government departments or historical prime ministers.
```
maxf, line 2: It looks like that the component is only meant to display positive integers. So either you should specify that constraint and create tests for when something else is passed, or you should allow more types of numbers (looking at the code it might actually work if you pass any string, in fact)

app/views/govuk_publishing_components/components/big_number/_value.html.erb
```diff
@@ -0,0 +1,23 @@
...
+    classes << "gem-c-big-number__value--decorated"
+  end
+%>
+
+<%= tag.span class: classes do %>
+  <%= number %><%= "+" if show_plus %>
+<% end %>
+
+<% unless label.nil? %>
+  <% # add a virtual space here to handle screen readers printing dictations without a space between the number and the label %>
```
andysellick, line 18: This is obviously an important issue - have you considered adding a test for the presence of this element in this situation? Might make it harder to accidentally remove in future.

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
...
+body: |
+  This component needs at least one item in the `items` array passed to it, otherwise it won't render. If no `number` attribute is passed in the item, it will default to "0".
```
andysellick, line 4: This isn't strictly true... if I wanted to display a number 0 and nothing else, I have to pass 0 as the number attribute, otherwise the component considers the data empty and doesn't render anything.

Maybe this is trying to be too clever... why not just display whatever is passed, and if nothing is passed, don't display anything?

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
...
+accessibility_criteria: |
+  - uses clear language and is positioned in a logical context on the page it's being used on
```
andysellick, line 6: Just a small readability nitpick - I think this list needs to start with something like `The component must`.

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -0,0 +1,19 @@
+<%
+  items ||= []
+%>
+<% if items.any? %>
+  <div class="gem-c-big-number">
+    <% if items.length > 1 %>
```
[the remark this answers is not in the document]
andysellick: Also the component sets the spacing between more than one number. What if that doesn't match the required design? Then we need an option for the layout (horizontal, vertical) and options for the space between them. Only rendering one number seems simpler.

app/views/govuk_publishing_components/components/big_number/_value.html.erb
```diff
@@ -0,0 +1,23 @@
...
+  href ||= nil
+  classes = ["gem-c-big-number__value"]
+  
+  if label.nil? && href
+    classes << "gem-c-big-number__value--decorated"
+  end
+%>
+
+<%= tag.span class: classes do %>
+  <%= number %><%= "+" if show_plus %>
```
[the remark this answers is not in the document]
andysellick: Just checked - it renders any string passed, so this seems unnecessary.

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
...
+    description: |
+      If more than one number is present in the `items` array, it will be rendered as an unordered list as opposed to just the number element alone.
+    data:
+      items:
+        - number: 119
+        - number: 553
+  with_labels:
+    description: |
+      Labels for numbers, if present, are styled in such a way that they are inline but appear to have a block relationship. This is to adhere to established design whilst bypassing a minor issue for NVDA users where the software will identify block level items and dictate them separately. This this case it is less jarring for screen reader users to hear the number and the label as a single dictation.
+    data:
```
andysellick, line 25: This description is a bit wordy. Suggestion:

```
Labels for numbers are given inline styling but stacked using pseudo elements with display block. This is to bypass an issue with NVDA where block level elements are dictated separately.
```

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
...
+      items:
+        - number: 23
+          label: Ministerial departments
+        - number: 400
+          show_plus: true
+          label: Other agencies and public bodies
+  with_link:
+    description: |
+      If a `href` attribute is present on a given number, we wrap that item in an `a` tag. Only the label is underlined and underline hover styles are only applied to that label.
```
andysellick, line 43: This description might be unnecessary - I think this information is conveyed in the preview beneath.

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,75 @@
...
+        - number: 23
+          label: Ministerial departments
+          href: "/government/organisations#ministerial_departments"
+        - number: 400
+          show_plus: true
+          label: Other agencies and public bodies
+          href: "/government/organisations#agencies_and_other_public_bodies"
+  with_link_but_no_label:
+    description: |
+      In the event that a `href` is provided but a `label` isn't, a decorator class is applied to the number itself that mirrors the `govuk-link--no-underline` class.
```
andysellick, line 55: I think this could be written more clearly...

```
Numbers that are links are underlined unless a label is provided, in which case the label is given the underline.
```

andysellick (commented):
Looks good, although I've raised a few thoughts that I think need to be considered.

Also I'd suggest merging these commits into a single one - as individually they don't do anything.

app/views/govuk_publishing_components/components/big_number/_item.html.erb
```diff
@@ -0,0 +1,13 @@
...
+  href ||= nil
+%>
+
+<div class="gem-c-big-number__item">
+  <% unless href.nil? %>
+    <%= link_to href, class: "govuk-link gem-c-big-number__link" do %>
+      <%= render "govuk_publishing_components/components/big_number/value", local_assigns %>
+    <% end %>
+  <% else %>
+      <%= render "govuk_publishing_components/components/big_number/value", local_assigns %>
```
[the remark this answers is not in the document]
owenatgov: That seems reasonable to have it in one place. Do you have an example of a capture in use in another component i could study?

app/views/govuk_publishing_components/components/_big_number.html.erb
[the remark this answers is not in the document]
owenatgov: So my thoughts on this are:

- On govuk presently, we use this component as a stacked list more often than we don't. [Example 1](https://www.gov.uk/government/how-government-works). [Example 2](https://www.gov.uk/government/get-involved). [Example 3](https://www.gov.uk/). I feel that it creates risk by not enforcing a list structure for multiple numbers, creating an (albeit quite minor) accessibility risk and effectively outsourcing the messiness across all our apps. This, whilst involving some extra calculation in the component, keeps the apps simple which is where the risk of things becoming messier is higher, in my opinion.
- I feel that the spacing is adequate both for between numbers and for between it and proceeding elements. I reckon that your suggestion for a custom margin is probably a good idea, which also is a case for only having this be one number that devs call as and when, but also I'm keen not to build in features that aren't necessary _right now_.

I'm on the fence about making this a single component in all honestly. Interested to know your thoughts on the above.

app/views/govuk_publishing_components/components/big_number/_value.html.erb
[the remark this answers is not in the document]
owenatgov: Max F pointed something to this effect out in his comment above. I was thinking about adding some sort of validation on the number input to dissuade wayward devs from passing words to the component, but I'm also wondering if this is overcomplicating things a bit.

app/views/govuk_publishing_components/components/docs/big_number.yml
owenatgov: I probably should've been clearer here. The 2nd sentence is for if, for some reason, other attributes like `label` are passed by the `number` attribute isn't. I suppose I could just check for the number and render nothing if it's not there?

app/views/govuk_publishing_components/components/big_number/_item.html.erb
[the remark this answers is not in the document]
andysellick: https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_button.html.erb#L8 is a good example, but there are plenty of others if you search.

app/views/govuk_publishing_components/components/_big_number.html.erb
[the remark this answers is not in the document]
andysellick: I didn't realise we had that many instances of multiple numbers, but I still think it should be a single thing. I think we can trust developers to do the right thing (and the list styles in question are already included in every application).

Absolutely, we shouldn't build in features that aren't needed right now, I'm thinking ahead to the complexity of unpicking this if e.g. different spacing is required, or an ordered list is required instead of an unordered one. Having it as a single thing makes it much more flexible, and reduces its complexity significantly.

app/views/govuk_publishing_components/components/big_number/_value.html.erb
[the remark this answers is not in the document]
andysellick: I think that would be overcomplicating it. I think it's far easier to let developers pass any string without having to worry about what constitutes valid input - the intent of the component is clear.

(other things we might want to display that would further complicate matters should we decide to validate input include e.g. `90%`, `-20`, `1.5`, `1M`, `1Mb` - let's not go down that road)

app/views/govuk_publishing_components/components/docs/big_number.yml
andysellick: Sounds good.

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -0,0 +1,19 @@
+<%
+  items ||= []
+%>
+<% if items.any? %>
+  <div class="gem-c-big-number">
```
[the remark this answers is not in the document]
owenatgov: Good call. I haven't been able to test this everywhere ass govuk docker has been a mess for me recently but I'll investigate spacing on this when I can.

app/views/govuk_publishing_components/components/big_number/_value.html.erb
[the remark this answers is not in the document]
owenatgov: THat's a very good point. I'll remove.

owenatgov:
@andysellick I think this is ready for a re-review now. I've made a couple of updates to the spec, some of it outside your review, which you can read about here https://github.com/alphagov/govuk_publishing_components/issues/2221#issuecomment-916197999

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,52 @@
+name: Big number
+description: A big number, with a label if needed, for visualising quantitative values such as the number of government departments or historical prime ministers.
+body: |
+  This component requires at least a non-falsey `number` attribute value passed to it, otherwise it will not render.
+accessibility_criteria: |
+  This component must:
+  - uses clear language and is positioned in a logical context on the page it's being used on
+  - communicates number value and label (if present) in a single dictation when reading with screen reader software
```
andysellick, line 8: Nitpick on language here. It currently reads a little strangely.

```
This component must uses clear language and is positioned in a logical context
This component must communicates number value and label
```

Also I think accessibility criteria should define the way the component itself behaves, not how it should be used - effectively it should be a set of rules that the component must adhere to in order to be accessible. Something like:

```
The component must:

- communicate number value and label (if present) in a single dictation when read with a screen reader
- convey the meaning of the number shown
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -0,0 +1,39 @@
...
+.gem-c-big-number__value {
+  @include govuk-font($size: 80, $weight: bold);
+}
+
+.gem-c-big-number__label {
+  @include govuk-font($size: 16, $weight: bold, $line-height: 2);
+
+  &:before {
+    content: "";
+    display: block;
```
andysellick, line 14: The reason for this pseudo element is documented nicely in the YML file but I wonder if it would be worth putting a short comment in here as well? The reason being that if a developer might not think to consult the YML file before removing this bit (if they thought it wasn't useful).

app/views/govuk_publishing_components/components/docs/big_number.yml
```diff
@@ -0,0 +1,52 @@
...
+shared_accessibility_criteria:
+- link
+examples:
+  default:
+    data:
+      number: 119
+  with_labels:
+    description: |
+      Labels for numbers are given inline styling but stacked using pseudo elements with display block. This is to bypass an issue with NVDA where block level elements are dictated separately.
```
andysellick, line 17: Nitpick: this description (and a few below) don't need to be formatted using `|` as it's a single line. You could just write `description: Labels for numbers are given...`

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -0,0 +1,34 @@
...
+
+  <% unless label.nil? %>
+    <% # add a virtual space here to handle screen readers printing dictations without a space between the number and the label %>
+    <span class="govuk-visually-hidden">&nbsp;</span>
+    <span class="gem-c-big-number__label">
+      <%= label %>
+    </span>
+  <% end %>
+<% end %>
+<% if number %>
```
andysellick, line 26: This checks whether or not to render the component, correct? Would it make more sense for it to wrap the whole of the code (i.e. from line 13 onwards) rather than just this part?

spec/components/big_number_spec.rb
```diff
@@ -0,0 +1,58 @@
...
+  it "renders nothing when nothing is passed to it" do
+    assert_empty render_component({})
+    assert_empty render_component(number: false)
+  end
+
+  it "renders a number value" do
+    render_component(number: 500)
+
+    assert_select ".gem-c-big-number", count: 1
+    assert_select ".govuk-list", count: 0
```
andysellick, line 17: What's this check for? I don't see the `govuk-list` class used anywhere in the component.

andysellick (commented):
Nice. I like the changes you've made - makes the component much clearer and more straightforward.

Just found a few things that need a look, but otherwise looks good 👍

app/views/govuk_publishing_components/components/docs/big_number.yml
owenatgov: Your suggestion makes sense to me, updating...

spec/components/big_number_spec.rb
owenatgov: Oops, legacy test. Removing...

andysellick (approved):
Good stuff. I'm also still particularly enjoying the name of this component.

PR #2281, opened 2021-08-23
Fix navigation header layout error that occurs when resizing browser window

injms:
## What
Removes the window resize listener inside the `updateStates` function, which itself was used with a window resize or matchMedia listener.

## Why
When moving from mobile to desktop view, and resizing the window to exactly 769 pixels wide the layout was broken. This was because the listener to change the layouts was added at this breakpoint, but the function wouldn't be fired until the _next_ resize event one pixel later.

Since this was a `resize` listener being fired by another resize or matchMedia listener it could be removed and replaced with firing the function.

Example of the visual glitch:

Screenshot 2021-08-23 at 14 15 28

## Visual Changes

Before at 769 pixels wide:

Screenshot 2021-08-23 at 14 14 28

After at 769 pixels wide:

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,46 +7,50 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-# 25.5.0
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Looks good, works fine. Dreams do come true.

chao-xian:
@owenatgov 
image

PR #2283, opened 2021-08-24
Add element tracking to track click

andysellick:
## What
- allows us to specify an element class within which link clicks will be tracked
- caters for the situation where we want to track specific links within a govspeak block and can't get to the source, except from a high level parent element

This PR also adds some documentation for this script, since it's been having some new features lately.

## Why

We want to track link clicks in the second grey box on pages like https://www.gov.uk/foreign-travel-advice/spain (underneath 'Summary') but that box and all the text below it is managed through a publishing interface, so we can only apply tracking stuff to a parent element. This allows us to provide the classname of the grey box so that only links clicked in that box are tracked.

## Visual Changes
None.

Trello card: https://trello.com/c/JREgQ2iv/579-spike-into-tracking-enhancements

### DIFF
--- a/docs/track-click.md
+++ b/docs/track-click.md
@@ -1,0 +1,89 @@
+# Track click script
+
+The gem includes a script that allows tracking to be added to clicked elements with relative ease. It depends upon the main analytics code to function.
+
+Basic use:
+
+```html
+<a href="/link"
+  data-module="gem-track-click"
+  data-track-category="category"
+  data-track-action="action"
+  data-track-label="Foo">
+  Link
+</a>
+```
+
+Advanced use. Adds a custom dimension of `dimension29` with a value of `dimension-value`, and a `value` of `9`.
+
+```html
+<a href="/link"
+  data-module="gem-track-click"
+  data-track-category="category"
+  data-track-action="action"
+  data-track-label="label"
+  data-track-dimension="dimension-value"
+  data-track-dimension-index="29"
+  data-track-value="9">
+  Link
+</a>
+```
+
+Track with arbitrary JSON:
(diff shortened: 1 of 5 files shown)


docs/track-click.md
```diff
@@ -0,0 +1,60 @@
...
+Advanced use. Adds a custom dimension of `dimension29` with a value of `dimension-value`, and a `value` of `9`.
+
+```html
+<a href="/link" data-module="gem-track-click" data-track-category="category" data-track-action="action" data-track-label="label" data-track-dimension="dimension-value" data-track-dimension-index="29" data-track-value="9">Link</a>
+```
+
+Track with arbitrary JSON:
+
+```html
+<a href='/link' data-module='gem-track-click' data-track-category='category' data-track-action='1' data-track-label='/' data-track-options='{\"dimension28\": \"foo\", \"dimension29\": \"bar\"}'>Link</a>
```
injms, line 20: I don't think you need to escape the double quotes here as `JSON.parse()` doesn't need it and it's inside single quotes:

```suggestion
<a href='/link' data-module='gem-track-click' data-track-category='category' data-track-action='1' data-track-label='/' data-track-options='{"dimension28": "foo", "dimension29": "bar"}'>Link</a>
```

injms (approved):
Minor quibble about escaping `"`, but nothing that should block this from being merged in.

🦖

docs/track-click.md
andysellick: Of course, thanks for spotting! Have corrected. I've also reformatted all the examples so they're not on a single line, made it a bit more readable.

PR #2289, opened 2021-08-31
Add a dependency on govuk_personalisation, and use it to generate account URLs in the account layout

barrucadu:
## What
Change the account parts of the public layout component to use [the `GovukPersonalisation::Urls` module](https://www.rubydoc.info/gems/govuk_personalisation/0.7.0/GovukPersonalisation/Urls), which we have introduced as a consistent way to link to account-related pages across apps.

This is a breaking change, because it introduces a dependency on [the govuk_personalisation gem](https://rubygems.org/gems/govuk_personalisation), which has a dependency on `rails ~> 6`, whereas govuk_publishing_components currently has a dependency on `rails >= 5.0.0.1`, so any Rails 5 users will need to upgrade to Rails 6.  This shouldn't be a problem for GOV.UK, as we are already using Rails 6 for everything.

## Why
We currently link to account-related URLs in a few apps, and have adopted different practices in different places:

- In some places we use hard-coded paths on GOV.UK ([eg, static](https://github.com/alphagov/static/blob/9b0caaa6de54a0e3624f29520e97fec5a4fe29d2/app/views/root/_gem_base.html.erb#L47))
- In some places, we use environment variables with a full URI ([eg, email-alert-frontend](https://github.com/alphagov/email-alert-frontend/blob/88458caef82beb1b0b04b04cc46e0bdf7264fc35/app/controllers/subscriptions_management_controller.rb#L102))
- In some places, we use hard-coded paths on the Account Manager domain ([eg, govuk_publishing_components](https://github.com/alphagov/govuk_publishing_components/blob/3406c13d2564dbbde03e6b5ed3fc97842819c327/app/views/govuk_publishing_components/components/layout_for_public/_account-layout.html.erb#L4))

So we introduced the `GovukPersonalisation::Urls` module to give a Plek-like approach to finding URLs:

- For things on GOV.UK, it'll generate a URL relative to the website root (or the specific frontend app when running in development mode)
- For things off GOV.UK, it'll generate a URL relative to the Account Manager
- For both, an environment variable can be set to override the URL

So we will make our URL generation logic consistent, and also be able to support the migration to Digital Identity away from the Account Manager by overriding URLs in Puppet.

## Visual Changes
None.

---

[Trello card](https://trello.com/c/ZvDGaba4/958-use-di-auth-in-integration-give-feedback-on-their-documentation)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* **BREAKING:** Add a dependency on govuk_personalisation, and use it to generate account URLs in the account layout ([PR #2289](https://github.com/alphagov/govuk_publishing_components/pull/2289))
+
+  You must make the following changes when you migrate to this release:
+  - Upgrade to Rails 6
+
 ## 25.7.0
 
 * Add text list to image card ([PR #2286](https://github.com/alphagov/govuk_publishing_components/pull/2286))
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -12,6 +12,7 @@ updates:
   - dependency-name: gds-sso
   - dependency-name: govspeak
   - dependency-name: govuk_app_config
+  - dependency-name: govuk_personalisation
   - dependency-name: govuk_publishing_components
   - dependency-name: govuk_schemas
   - dependency-name: govuk_sidekiq
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -3,9 +3,10 @@ PATH
   specs:
     govuk_publishing_components (25.7.0)
       govuk_app_config
+      govuk_personalisation (>= 0.7.0)
       kramdown
       plek
-      rails (>= 5.0.0.1)
+      rails (>= 6)
       rouge
       sprockets (< 4)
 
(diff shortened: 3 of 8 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## unreleased
```
danacotoran, line 10: Major nitpick: we'd usually capitalise **Unreleased**

CHANGELOG.md
```diff
@@ -7,6 +7,11 @@
...
+
+* Use `GovukPersonalisation::Urls` to generate account URLs ([PR #2289](https://github.com/alphagov/govuk_publishing_components/pull/2289))
+* **BREAKING:** Add dependency on govuk_personalisation, which increases the minimum supported Rails version to 6 ([PR #2289](https://github.com/alphagov/govuk_publishing_components/pull/2289))
```
danacotoran, line 13: We'd generally just add one `CHANGELOG` entry per PR, and usually have it match the title of the PR. So this would usually be recorded as:
> **BREAKING:** Add a dependency on govuk_personalisation, and use it to generate account URLs in the account layout ([PR #2289](https://github.com/alphagov/govuk_publishing_components/pull/2289))

If it's a breaking change we'd usually make a note of what changes should be made (if any) when one migrates to this version of the gem.

danacotoran (approved):
Looks good to me besides some Changelog nitpicking

CHANGELOG.md
barrucadu: I've added a note, though in practice for GOV.UK apps this won't be a breaking change (as we're already on Rails 6)

PR #2292, opened 2021-09-03
Allow default header navigation on public layout

alex-ju:
## What
This PR updates `layout_for_public` to be able to render the default header navigation.

## Why
A set of applications [1][2][3] require a header navigation (sometimes referred to as 'proposition menu') which is only available in `govuk_template` at the moment. 

## Note
We expect the `layout_header `component to be replaced by the `layout_super_navigation_header` in future, so all this code will be removed, but in the meantime this work is required to be able to progress with removing the dependency on `govuk_template`.

[1] https://github.com/alphagov/whitehall/pull/6240
[2] https://github.com/alphagov/finder-frontend/pull/2616
[3] https://github.com/alphagov/government-frontend/pull/2206

## Visual Changes
[Preview `layout_for_public` with default navigation](https://govuk-publis-allow-defa-xbzgfn.herokuapp.com/component-guide/layout_for_public/navigation_default/preview)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Allow default header navigation on public layout ([PR #2292](https://github.com/alphagov/govuk_publishing_components/pull/2292))
+
 ## 26.0.0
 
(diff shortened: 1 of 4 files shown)


maxgds (approved):

alex-ju:
Converting this to a draft as the current header (and the associated navigation) may not be around for long.

injms:
The [navigation header](https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header/) is now the default header across GOV.UK - is this okay to be closed now?

PR #2294, opened 2021-09-07
Fix global banner wrapper in public layout

alex-ju:
## What
Fix global banner wrapper in public layout

## Why
The `global-banner-wrapper` element is rendered whether or not there is an active global banner as it's a slot for other elements such as the survey banner. When there is no content in this container, because of the negative margin-top it pulls the element below (often the back link or the breadcrumbs) higher up than it should be. By setting a `min-height` we make sure no layout alteration happen in this scenario, while it scales up when content is presented within the container.

## Visual Changes

<table>
<tr><td valign="top">

Screenshot 2021-09-07 at 10 32 36

</td><td valign="top">

Screenshot 2021-09-07 at 10 32 59

</td></tr>
</table>


maxgds (approved):
Nice fix.

PR #2296, opened 2021-09-07
Fix document list children

alex-ju:
## What
Remove class composition in favour of attribute declaration on `gem-c-document-list__children` elements.

## Why
Class composition (in this case `gem-c-document-list__children`  with `govuk-list`) means that when we have the same attribute (let's say `margin-top`) the order of precedence is dictated by where these class definitions are presented in the compiled CSS files – known as document order. Document order makes styling flakey as when swapping a component in or out or changing the order of SCSS imports we may end up with different styling. This happened in https://github.com/alphagov/finder-frontend/pull/2616 when switching to the new `gem_layout` where the document-list SCSS import is followed by [error-summary](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/stylesheets/govuk_publishing_components/components/_error-summary.scss), which [includes the basic list styles](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/error-summary/_index.scss#L1) making `govuk-list` styles to override definitions in `gem-c-document-list__children`.

## Visual Changes
<table>
<tr><td valign="top">

Screenshot 2021-09-07 at 12 19 05

</td><td valign="top">

Screenshot 2021-09-07 at 12 20 59

</td></tr>
</table>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Create devolved nations component ([PR #2280](https://github.com/alphagov/govuk_publishing_components/pull/2280))
+* Fix document list children ([PR #2296](https://github.com/alphagov/govuk_publishing_components/pull/2296))
 
 ## 26.0.0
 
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Personally I'd merge the two commits as they relate directly to each other.

PR #2304, opened 2021-09-09
Fix sticky hover on search button in navigation header

injms:
## What
Fixes the sticky hover state seen on touchscreen devices by unsetting hover states when a touchscreen device is being used.

Card: https://trello.com/c/fZ9bs2G0

## Why
A stick hover state is seen on some touchscreen devices - when a button is selected, the hover state is also triggered. This caused the search button in the navigation header to look hovered when the focus state moved on. Fixing this makes the interface easier to understand, as the states of the buttons are no longer muddled.

### Before

Screenshot 2021-09-09 at 14 12 36

### After

Screenshot 2021-09-09 at 14 10 42

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-# Unreleased
+## Unreleased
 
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
A sneaky solution. Very good 👍

PR #2305, opened 2021-09-09
Add new scroll tracking

andysellick:
## What
Add a new scroll tracker. It works as follows:

- written as a module, so it can be initialised as a meta tag using the `data-module` attribute
- only allows one instance of itself to be run
- defaults to tracking by percentages, currently they're hard coded as we've never used anything other than `20, 40, 60, 80, 100` (checked with a PA on this)
- can switch to tracking headings using an additional attribute of `data-track-type="headings"`. In this configuration it finds all the headings on the page dynamically, and tracks as each one comes into view ('into view' is defined as the whole of the heading is visible in the viewport)
- tracks anything visible on the page when it first loads
- invisible headings (display: none either on themselves or a parent) are not tracked, unless a change in the page causes them to become visible, in which case they will
- headings inside certain elements are not tracked (hard coded as a list of classes, includes the cookie banner, feedback component and footer)
- for both headings and percentages, the code calculates the positions of those things on initialisation and stores the values, to reduce the computational overhead on scroll. It also listens for page resize events and recalculates the positions when that happens. This allows for browser resizing or users changing the orientation of their mobile device
- it also sets an interval to compare the height of the page with the height of the page stored at initialisation, to check for other changes that could alter the height and change the thing positions e.g. hiding the cookie banner, opening or closing an accordion
- it also takes note of which things have been previously tracked (both headings and percentage positions) so they don't get tracked twice on the same page. The mechanism by which it does this is slightly different for each

Newly added feature:

- the script now checks for the presence of a location hash (e.g. `https://www.gov.uk/foreign-travel-advice/spain/coronavirus#finance`) and if there is one and if it matches a valid page element, no tracking is done on initial page load
- this prevents false positives from being tracked i.e. you don't want to track the headings at the top of the page, as the user hasn't actually read them
- what happens is the page loads, the browser jumps down to the anchor, and that causes a scroll event, and the script continues as normal

## Why
I needed to write this because of some shortcomings with the existing tracker. We want to add tracking to the foreign travel advice pages, to track headings (here's the list: https://www.gov.uk/foreign-travel-advice). Problems with the existing tracker:

- to add a page, the URL must be hardcoded into the JS file for the tracker. There are already more than 200 lines of config and this would add another 226 URLs (for the main pages alone, plus all the other pages e.g. `/foreign-travel-advice/spain/coronavirus`) plus all of the headings, which have to be hardcoded... this would have added hundreds of extra lines to this JS file, which gets loaded onto every single page of GOV.UK.
- the existing tracker when tracking headings needs to have those headings hard coded into the JS file, which is brittle because these headings might change. The new tracker identifies the headings on initialisation.

This new tracker doesn't replace the existing scroll tracker, because that would be disruptive and too risky to do all at once, and there may be benefits of both approaches. Specifically, the existing tracker allows scroll tracking to be created on a specific URL, whereas this new scroll tracker will be initialised on any page where the calling template is rendered. For example if you wanted to track foreign travel advice pages (which all use the same template) this new tracker is good if you want to track all of them, but the existing tracker is good if you only wanted to track one of them.

## Visual Changes
None.

Trello card: https://trello.com/c/JREgQ2iv/579-spike-into-tracking-enhancements


kevindew (commented):
This looks great @andysellick, thanks for digging into all this. It looks so much more maintainable than the previous version.

I've not given this an in-depth review, just looking over the approach. 

As this is something that can and should only exist once on a HTML page I feel like it'd benefit from being configured with a meta element in the head rather than a div element in the body. This would feel consistent with the pattern we've taken previously for analytics by the use of meta tags for static analytics: https://github.com/alphagov/govuk_publishing_components/blob/9769f35e95024eda025f3ca11896bfa117f04958/app/assets/javascripts/govuk_publishing_components/analytics/pii.js#L26 - whereas a body element for this implies a type of constraining.

Another thought is whether this module should be within the analytics package of this codebase rather than as a module in lib? I also assume, by it tracking nodes at the start that we'd not expect this to work with a page where the dom is changed dynamically (I guess it wouldn't even make sense in that scenario)

app/assets/javascripts/govuk_publishing_components/lib/auto-scroll-tracker.js
```diff
@@ -0,0 +1,170 @@
...
+(function (Modules) {
+  function AutoScrollTracker ($module) {
+    this.$module = $module
+    this.SCROLL_TIMEOUT_DELAY = 20
+    this.RESIZE_TIMEOUT_DELAY = 100
+    this.PAGEHEIGHT_TIMEOUT_DELAY = 500
+    this.pageHeight = document.querySelector('body').clientHeight
+    this.trackedNodes = []
+    this.config = {
+      ignoreHeadingsInside: ['.govuk-cookie-banner', '.gem-c-feedback', 'footer'],
```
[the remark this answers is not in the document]
andysellick: That might be simpler, yes. I did it this way because I thought maybe at some point we might want an option on the script that would allow other elements to be ignored as well (say if we wanted to ignore the headings in the sidebar on a particular page). But we're not doing that yet, so... I'll have a think. It's probably simpler the way you suggest.

app/assets/javascripts/govuk_publishing_components/lib/auto-scroll-tracker.js
```diff
@@ -0,0 +1,170 @@
...
+    var headingsFound = document.querySelectorAll(headings)
+    if (headingsFound) {
+      for (var i = 0; i < headingsFound.length; i++) {
+        var heading = headingsFound[i]
+        // only track headings that are visible i.e. not inside display: none
+        if (heading.offsetParent !== null && !this.shouldBeIgnored(heading)) {
+          var pos = heading.getBoundingClientRect()
+          headingsDetails.push({
+            node: heading,
+            alreadySeen: heading.getAttribute('data-already-seen'),
```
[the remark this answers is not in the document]
andysellick: Great shout, have changed.

andysellick:
Thanks for taking a look @kevindew.

> As this is something that can and should only exist once on a HTML page I feel like it'd benefit from being configured with a meta element in the head rather than a div element in the body.

Yes, that would be tidier, especially as it only needs to happen once. However I would need to write completely new code to initialise it - at the moment it's using the modules code, and I like the consistency. But I'm happy to think about it some more.

> Another thought is whether this module should be within the analytics package of this codebase rather than as a module in lib? 

Agreed. The reason it's there is because as a module it didn't start when it was inside the analytics dir, so this is linked to your previous comment. If we go that route I'd definitely move it in there.

> I also assume, by it tracking nodes at the start that we'd not expect this to work with a page where the dom is changed dynamically (I guess it wouldn't even make sense in that scenario)

I hadn't considered that (do we have pages where headings change dynamically?) so no, I wouldn't expect it to work in that situation (although it should detect headings that are shown after page load by e.g. an accordion being opened). If there's a need for that then I'd suggest that kind of thinking for version 2, where it would be good to try to use intersection observers and do this all properly (but they're not supported by IE so we're doing this old school).

kevindew:
> Thanks for taking a look @kevindew.
> 
> > As this is something that can and should only exist once on a HTML page I feel like it'd benefit from being configured with a meta element in the head rather than a div element in the body.
> 
> Yes, that would be tidier, especially as it only needs to happen once. However I would need to write completely new code to initialise it - at the moment it's using the modules code, and I like the consistency. But I'm happy to think about it some more.

Sure, what's the existing precedence of initialisation for the analytics code? Are we going to bundle this script as part of analytics.js or expect apps to opt into this file individually? 

Seems like you could apply it with minimal changes if it loads implicitly e.g: 

```
var trackerMeta = document.querySelector("meta[name=govuk:scroll-tracker]")

if (trackerMeta) {
  new AutoScrollTracker(trackerMeta).init()
}
```

I imagine there's also a middle ground where you use the meta element for the module. E.g: `<meta name="govuk:scroll-tracker" content="percentages" data-module="auto-scroll-tracker" />` not too sure if that's a good idea or not 🤔 

> > Another thought is whether this module should be within the analytics package of this codebase rather than as a module in lib?
> 
> Agreed. The reason it's there is because as a module it didn't start when it was inside the analytics dir, so this is linked to your previous comment. If we go that route I'd definitely move it in there.

Gotcha, it does look like we already have a few modules in the analytics directory already: [track-click.js](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/track-click.js), [auto-track-event.js](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/analytics/auto-track-event.js), etc.

By the way, is there a convention as to whether this (as a module) should be prefixed with Gem or not. I couldn't really tell what the convention is but I know there is some, i.e gem-track-click

> > I also assume, by it tracking nodes at the start that we'd not expect this to work with a page where the dom is changed dynamically (I guess it wouldn't even make sense in that scenario)
> 
> I hadn't considered that (do we have pages where headings change dynamically?) so no, I wouldn't expect it to work in that situation (although it should detect headings that are shown after page load by e.g. an accordion being opened). If there's a need for that then I'd suggest that kind of thinking for version 2, where it would be good to try to use intersection observers and do this all properly (but they're not supported by IE so we're doing this old school).

I'm not sure of any - I guess some places where content is dynamically loaded like mainstream browse potentially, definitely rare. I think it just wouldn't make sense though with dynamic headers as a record of having someone scrolled to a header wouldn't make sense without the state of the page at the time.

app/assets/javascripts/govuk_publishing_components/lib/auto-scroll-tracker.js
```diff
@@ -0,0 +1,170 @@
...
+          if (pageHeight !== this.pageHeight) {
+            this.pageHeight = pageHeight
+            this.getWindowDetails()
+          }
+        }.bind(this), this.PAGEHEIGHT_TIMEOUT_DELAY)
+      }
+    }
+  }
+
+  AutoScrollTracker.prototype.onScroll = function onScroll () {
```
injms, line 61: Out of curiosity, what's the difference between
```js
AutoScrollTracker.prototype.onScroll = function onScroll () { ... }
```
and 
```js
AutoScrollTracker.prototype.onScroll = function () { ... }
```
?

andysellick: Ah, good spot. The difference is that the first one isn't anonymous. Or something like that. It's kinda weird. Even though both of them are called `onScroll` as far as referring to it inside the module is concerned, only the first example actually has a function name of `onScroll`.

I meant to take this out - it was part of getting the tests to work properly, but then I found a different way to remove the scroll listener after each test. I'll remove, thanks.

andysellick:
> Another thought is whether this module should be within the analytics package of this codebase rather than as a module in lib?

I've moved it. Turns out the reason it wasn't initialising was because I hadn't included it in the `analytics.js` file 🤦 

> I imagine there's also a middle ground where you use the meta element for the module. E.g: `<meta name="govuk:scroll-tracker" content="percentages" data-module="auto-scroll-tracker" />` not too sure if that's a good idea or not 🤔

I thought this wouldn't work, because I thought modules only initialised within the body, but it turns out this works fine. So yes, this is how we can initialise this module. I've tried to make minimal changes, so it can be initialised like this:

`<meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings"/>`

> By the way, is there a convention as to whether this (as a module) should be prefixed with Gem or not. I couldn't really tell what the convention is but I know there is some, i.e gem-track-click

No. GemTrackClick is a special case, because it was originally in static but we wanted a new version without conflicting with the existing version. Now though, the version in static has been removed (I think) and we've only got this one, so at some point we could remove the `gem` bit, but it's not urgent.

app/assets/javascripts/govuk_publishing_components/lib/auto-scroll-tracker.js
andysellick: Removed.

injms (commented):
This looks good from a code perspective. Only one question - does this need any documentation anywhere?

andysellick:
> This looks good from a code perspective. Only one question - does this need any documentation anywhere?

Good shout, yes. I'll add some.

injms (approved):

injms:
(Might need a rebase to stop the changelog getting muddled up.)

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+
+  AutoScrollTracker.Heading = function (config) {
+    this.config = config
+  }
+
+  AutoScrollTracker.Heading.prototype.getTrackingNodes = function () {
+    var headingsDetails = []
+    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+    var headingsFound = document.querySelectorAll(headings)
+    if (headingsFound) {
```
kevindew, line 127: I think this line is redundant, as you'll always get something that returns true and something you can iterate through from a `querySelectorAll`

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+        }
+      }
+    }
+    return headingsDetails
+  }
+
+  // check heading isn't inside something that needs ignoring e.g. cookie banner, footer
+  AutoScrollTracker.Heading.prototype.shouldBeIgnored = function (heading) {
+    for (var i = 0; i < this.config.ignoreHeadingsInside.length; i++) {
+      if (heading.closest(this.config.ignoreHeadingsInside[i])) {
```
kevindew, line 149: do we have a polyfill for closest? I understand it's not supported by any versions of IE

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+  }
+
+  AutoScrollTracker.Percentage = function (config) {
+    this.config = config
+  }
+
+  AutoScrollTracker.Percentage.prototype.getTrackingNodes = function (trackedNodes) {
+    var body = document.body
+    var html = document.documentElement
+    var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight)
```
kevindew, line 162: Corr, that's a lot of potential heights!

docs/analytics/auto_scroll_tracker.md
```diff
@@ -0,0 +1,52 @@
+# Scroll tracking
+
+The analytics code includes the ability to add scroll tracking to specified pages for users who have consented to cookies.
+
+There are currently two scroll tracking scripts in the gem. The new one has been written to address problems with the old one, but the scroll tracking handled by the old one has yet to be transferred to the new one.
```
kevindew, line 5: Rather than calling it "the new one" which seems to often age badly - it might be worth saying.

There are currently two scroll tracking scripts in the gem. You should use AutoScrollTracker. The previous one, ScrollTracker, has been deprecated by the introduction of AutoScrollTracker and should now be considered legacy.

The new one has been written to address problems with the old one, but the scroll tracking handled by the old one has yet to be transferred to the new one

docs/analytics/auto_scroll_tracker.md
```diff
@@ -0,0 +1,52 @@
...
+Operation:
+
+- Tracking events are only fired once i.e. if a user scrolls up and down a page, duplicate events are not tracked.
+- Tracking events are fired on page load for things that are immediately visible.
+- If the user has followed a jump link, e.g. `https://www.gov.uk/foreign-travel-advice/spain/coronavirus#finance` the script detects this and doesn't fire tracking events until after the browser jumps to the relevant section. It also checks that the hash matches a valid element on the page, and continues as normal if it doesn't.
+- Headings are tracked only when the whole of the heading is visible in the viewport.
+- Headings inside certain elements are ignored (includes the cookie banner, feedback component and footer).
+- Hidden headings are not tracked, unless they become visible (e.g. if inside an accordion that is opened).
+
+## ScrollTracker (the old one)
```
kevindew, line 45:
```suggestion
## ScrollTracker (deprecated, legacy approach)
```

docs/analytics/auto_scroll_tracker.md
```diff
@@ -0,0 +1,52 @@
+# Scroll tracking
+
+The analytics code includes the ability to add scroll tracking to specified pages for users who have consented to cookies.
+
+There are currently two scroll tracking scripts in the gem. The new one has been written to address problems with the old one, but the scroll tracking handled by the old one has yet to be transferred to the new one.
+
+## AutoScrollTracker (the new one)
```
kevindew, line 7:
```suggestion
## AutoScrollTracker
```

docs/analytics/auto_scroll_tracker.md
```diff
@@ -0,0 +1,52 @@
...
+- Tracking events are only fired once i.e. if a user scrolls up and down a page, duplicate events are not tracked.
+- Tracking events are fired on page load for things that are immediately visible.
+- If the user has followed a jump link, e.g. `https://www.gov.uk/foreign-travel-advice/spain/coronavirus#finance` the script detects this and doesn't fire tracking events until after the browser jumps to the relevant section. It also checks that the hash matches a valid element on the page, and continues as normal if it doesn't.
+- Headings are tracked only when the whole of the heading is visible in the viewport.
+- Headings inside certain elements are ignored (includes the cookie banner, feedback component and footer).
+- Hidden headings are not tracked, unless they become visible (e.g. if inside an accordion that is opened).
+
+## ScrollTracker (the old one)
+
+ScrollTracker should not be used for new scroll tracking requirements as it has several limitations.
```
kevindew, line 47: We should add a note that this should not be used for new code.

spec/javascripts/govuk_publishing_components/lib/auto-scroll-tracker-spec.js
```diff
@@ -0,0 +1,234 @@
...
+      document.body.appendChild(el)
+      scrollTracker = new GOVUK.Modules.AutoScrollTracker(el)
+      scrollTracker.init()
+    })
+
+    afterEach(function () {
+      document.body.removeChild(el)
+    })
+
+    it('should send a tracking event on page load for headings that are already visible', function () {
```
kevindew, line 75:
```suggestion
    it('should send a tracking event on initialisation for headings that are already visible', function () {
```

Since it doesn't necessarily need to be page load.

spec/javascripts/govuk_publishing_components/lib/auto-scroll-tracker-spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+    it('should not track a heading more than once', function () {
+      GOVUK.analytics.trackEvent.calls.reset()
+      GOVUK.triggerEvent(document.body, 'resize')
+      jasmine.clock().tick(200)
+
+      expect(GOVUK.analytics.trackEvent.calls.count()).toBe(0)
+    })
+
+    it('should track headings on scroll and ignore already tracked headings', function () {
```
kevindew, line 88: I found this test and the next couple a little confusing by the terminology of "on scroll" when the test isn't triggering a scroll event. I'd imagine just removing "on scroll" from the test description would clear that up (I imagining you're using resize as that doesn't carry the check that scroll does that the user has changed scroll position)

spec/javascripts/govuk_publishing_components/lib/auto-scroll-tracker-spec.js
```diff
@@ -0,0 +1,234 @@
...
+
+    it('should track new positions on scroll and ignore already tracked positions', function () {
+      var height = window.innerHeight
+      setPageHeight(height * 4)
+
+      expect(GOVUK.analytics.trackEvent.calls.count()).toBe(1)
+      expect(GOVUK.analytics.trackEvent.calls.argsFor(0)).toEqual(['ScrollTo', 'Percent', { label: '20', nonInteraction: true }])
+
+      GOVUK.analytics.trackEvent.calls.reset()
+      setPageHeight(10)
```
kevindew, line 172: These adjustments of the page height are neat to this this.

kevindew (commented):
This looks great @andysellick, nice work here. I'm sorry to give you so many comments after an approval - I hope they're mostly helpful in improving this. There seemed to be only a couple of blocking things if you're really keen to get this out.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function AutoScrollTracker ($module) {
+    this.$module = $module
+    this.SCROLL_TIMEOUT_DELAY = 20
+    this.RESIZE_TIMEOUT_DELAY = 100
+    this.PAGEHEIGHT_TIMEOUT_DELAY = 500
```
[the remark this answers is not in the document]
andysellick: No problem, have moved them into the config and camel cased as suggested.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+    if (consentCookie && consentCookie.settings) {
+      this.startModule()
+    } else {
+      this.startModule = this.startModule.bind(this)
+      window.addEventListener('cookie-consent', this.startModule)
+    }
+  }
+
+  AutoScrollTracker.prototype.startModule = function () {
+    if (!window.GOVUK.analyticsVars.scrollTrackerStarted) {
```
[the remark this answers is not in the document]
andysellick: Done, thanks.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+      // if the URL has a hash we want to prevent tracking on initial page load
+      // until the browser jumps down the page, at which point a scroll event
+      // will happen and tracking will continue normally
+      var windowHash = window.location.hash
+      var dontTrackOnLoad = windowHash && document.getElementById(windowHash.substring(1))
+      if (!dontTrackOnLoad) {
+        this.trackVisibleNodes()
+      }
+
+      if (this.trackedNodes) {
```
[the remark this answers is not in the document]
andysellick: Yes, probably. Changed.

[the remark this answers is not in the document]
andysellick: Thanks, didn't know that querySelectorAll returned an empty array on not found.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+          }
+        }.bind(this), this.PAGEHEIGHT_TIMEOUT_DELAY)
+      }
+    }
+  }
+
+  AutoScrollTracker.prototype.onScroll = function () {
+    clearTimeout(this.scrollTimeout)
+    this.scrollTimeout = setTimeout(function () {
+      this.trackVisibleNodes()
```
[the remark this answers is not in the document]
andysellick: Ooh, good spot. It should just be the one call, I think this was an error caused by me being too clever with find and replace when changing this bit.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Thanks, didn't know that querySelectorAll returned an empty array on not found. Changed.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Good point, have added. Also just noticed that I get multiple test failures in IE10, so I'm going to fix that before merging.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Yeah, apparently the only way to be sure...

spec/javascripts/govuk_publishing_components/lib/auto-scroll-tracker-spec.js
andysellick: I've tweaked the test a little, to keep the scroll event. Good point.

docs/analytics/auto_scroll_tracker.md
andysellick: Isn't that what this is saying? Can you suggest some better wording?

docs/analytics/auto_scroll_tracker.md
andysellick: Changed, thanks.

docs/analytics/auto_scroll_tracker.md
andysellick: Changed, thanks.

docs/analytics/auto_scroll_tracker.md
andysellick: Updated, thanks.

spec/javascripts/govuk_publishing_components/lib/auto-scroll-tracker-spec.js
andysellick: Changed, thanks.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: IE10 test failures now addressed.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,184 @@
...
+(function (Modules) {
+  function AutoScrollTracker ($module) {
+    this.$module = $module
+    this.SCROLL_TIMEOUT_DELAY = 20
+    this.RESIZE_TIMEOUT_DELAY = 100
+    this.PAGEHEIGHT_TIMEOUT_DELAY = 500
+    this.pageHeight = document.querySelector('body').clientHeight
+    this.trackedNodes = []
+    this.config = {
+      ignoreHeadingsInside: ['.govuk-cookie-banner', '.gem-c-feedback', 'footer'],
```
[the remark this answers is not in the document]
andysellick: I was a bit torn. I was thinking ahead to allowing this to be an option to be passed to this module, which would allow for greater flexibility in the future, but we don't have a requirement for that now and we shouldn't build things we don't need. So I've changed it as you suggested, although I've left it as an array (of one item) of elements that headings are allowed in, so if we do want to expand it in future it's easier.

docs/analytics/auto_scroll_tracker.md
kevindew: Sorry - I seem to have lost my mind in that comment. Who knows what my brain read.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,191 @@
...
+  }
+
+  AutoScrollTracker.Heading = function (config) {
+    this.config = config
+  }
+
+  AutoScrollTracker.Heading.prototype.getTrackingNodes = function () {
+    var headingsDetails = []
+    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+    var headingsFound = document.querySelectorAll(headings)
```
kevindew, line 129: Just a minor point that we could remove the need for `this.allowHeading` and the closest polyfill by using the allowedHeading selectors (or looking them as elements and doing querySelectorAll on them) - which would save a bunch of DOM traversal (no idea quite how expensive that is on low power devices mind)

E.g.

```
var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
var selectors = this.config.allowHeadingsInside.reduce(function (memo, item) {
  var appended = headings.map(function (heading) { return item + ' ' + heading })
  return memo.concat(appended)
}, [])
var headingsFound = document.querySelectorAll(selectors)
```

kevindew (approved):
This looks great to me, I've just made one non-blocking suggestion

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Thanks @kevindew. I like your suggestion, but I've reworked the details slightly. I've added it in an extra commit (which I'll squash once you've taken a look).

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -0,0 +1,195 @@
...
+    return headingsDetails
+  }
+
+  // check heading is inside allowed elements, generally ignores everything outside of page content
+  AutoScrollTracker.Heading.prototype.findAllowedHeadings = function () {
+    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+    var headingsFound = []
+
+    for (var h = 0; h < this.config.allowHeadingsInside.length; h++) {
+      var inside = document.querySelector(this.config.allowHeadingsInside[h])
```
kevindew, line 152: I think you should adjust this to cater for a selector within `this.config.allowHeadingsInside` not existing - just incase we either have any rogue documents without a main or we expand that selector.

You may want to do it based on:

```
var insideElements = document.querySelectorAll(this.config.allowHeadingsInside)
for (var h = 0; h < insideElements.length; h++) {
  var found = insideElements[h].querySelectorAll(headings)
  // ...
}
```

kevindew (commented):
The allowed headings change looks great, there's just a potential error situation that I think we can refactor away.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Good shout, have adjusted. It's a bit of a nest of loops, but it feels safer.

andysellick:
> With the introduction of this new approach I think it might be a good idea to make a plan for converging/shifting to it in the long term. Raising an issue to document the shortcomings of the old approach and how we should be moving to the new approach might be a good start.

Absolutely. We already have a card for this (shout if you can't access: https://trello.com/c/sdugtMbX/643-migrate-to-new-scroll-tracker) and I've already documented all the existing tracking in a spreadsheet and I'm working with the PA community to identify what we can remove/migrate (have removed three URLs already - https://github.com/alphagov/govuk_publishing_components/pull/2339).

PR #2306, opened 2021-09-10
Define available locales

1pretz1:
Defines the locales actually in use, in the application. This has been
done as a prerequisite for the Rails Translation Manager changes [1].

[1]: https://github.com/alphagov/rails_translation_manager/pull/12

Trello:
https://trello.com/c/Ma1ygjNc/2567-add-automated-tests-to-check-translations-5

### DIFF
--- a/spec/dummy/config/application.rb
+++ b/spec/dummy/config/application.rb
@@ -24,6 +24,72 @@ module Dummy
     # -- all .rb files in that directory are automatically loaded.
     config.i18n.default_locale = :en
     config.i18n.fallbacks = [I18n.default_locale]
+    config.i18n.available_locales = %i[
+      ar
+      az
+      be
+      bg
+      bn
+      cs
+      cy
+      da
+      de
+      dr
+      el
+      en
+      es-419
+      es
+      et
+      fa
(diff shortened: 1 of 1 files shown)


spec/dummy/config/application.rb
```diff
@@ -24,6 +24,72 @@ class Application < Rails::Application
     # -- all .rb files in that directory are automatically loaded.
     config.i18n.default_locale = :en
     config.i18n.fallbacks = [I18n.default_locale]
+    config.i18n.available_locales = %i[
```
injms, line 27: Does this mean that this list will need to be manually updated when a new translation file is added?

injms: Makes sense - and I can just see this being a thing that gets left out if another language is added. Is there any way that this list and the list of translation files could be compared to make sure that no language is missing? Or failing that, could a quick "So you want to add another language" guide be added to the [component gem's docs][gd]?

[gd]: https://github.com/alphagov/govuk_publishing_components/tree/master/docs

ChrisBAshton (approved):

PR #2307, opened 2021-09-13
Add missing `govuk-template` class to public layout

alex-ju:
## What
Add missing [`govuk-template` class](https://design-system.service.gov.uk/styles/page-template/) to public layout. 

## Why
This class set the overall page background colour to the same colour as used by the footer to give the illusion of a long footer, prevents automatic text sizing, as we already cater for small devices and would like the browser to stay on 100% text zoom by default and Force the scrollbar to always display in IE, to prevent horizontal page jumps as content height changes.

## Visual Changes
<table>
<tr><td valign="top">

Screenshot 2021-09-13 at 11 56 16

</td><td valign="top">

Screenshot 2021-09-13 at 11 56 01

</td></tr>
</table>

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -44,8 +44,8 @@
   body_css_classes << "draft" if draft_watermark
 -%>
 <!DOCTYPE html>
-  <!--[if lt IE 9]><html class="lte-ie8" lang="<%= html_lang %>"><![endif]-->
-  <!--[if gt IE 8]><!--><html lang="<%= html_lang %>"><!--<![endif]-->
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -45,7 +45,7 @@
 -%>
 <!DOCTYPE html>
   <!--[if lt IE 9]><html class="lte-ie8" lang="<%= html_lang %>"><![endif]-->
```
injms, line 47: Is the class needed on the old IE version as well?

```suggestion
  <!--[if lt IE 9]><html class="lte-ie8 govuk-template" lang="<%= html_lang %>"><![endif]-->
```

injms (approved):
Good catch - thank you!

PR #2311, opened 2021-09-17
Remove heading and use strong in step nav header

jon-kirwan:
## What
https://trello.com/c/eOOiMuK1/329-step-nav-header-looks-like-a-heading-but-isnt-one

Change markup to -
- Remove heading (h2) and 
- To give extra importance to the text 'Part of', replace `span` with `strong` inside the step nav header 

## Why
The step nav header is not a heading because there is no content directly underneath. Also the document outline is broken (though it is okay to have a `h2` above a `h1`).


danacotoran (approved):

danacotoran (changes requested):
@jon-kirwan post-approval realisation (sorry!!!): can we adjust the component so that `margin-bottom` override classes are applied only if `margin_bottom` is specified? Currently  `govuk-!-margin-bottom-0` is applied by default to the container which was necessary when it was a heading, not sure it's necessary anymore now that we've changed it to a `div`!

jon-kirwan:
> @jon-kirwan post-approval realisation (sorry!!!): can we adjust the component so that `margin-bottom` override classes are applied only if `margin_bottom` is specified? Currently `govuk-!-margin-bottom-0` is applied by default to the container which was necessary when it was a heading, not sure it's necessary anymore now that we've changed it to a `div`!

Okay, I think this should be working now. I've had to tweak a couple of additional things here because I discovered that when no margin_bottom was provided an extra character space was output in the class tag e.g.

`<div class="gem-c-step-nav-header " data-module="gem-track-click">`

So i've now modified things a little to use a content_tag in place of div which seems to do the trick

Also removed a test which I think we no longer need -

```
  it "defaults to no margin bottom" do
    render_component(title: "This is my title")

    assert_select '.gem-c-step-nav-header.govuk-\!-margin-bottom-0'
  end
```

app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
```diff
@@ -21,15 +21,16 @@
 
   local_assigns[:margin_bottom] ||= 0
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
-  classes = "gem-c-step-nav-header #{shared_helper.get_margin_bottom}"
+  classes = %w(gem-c-step-nav-header)
+  classes << shared_helper.get_margin_bottom if local_assigns[:margin_bottom] != 0
```
danacotoran, line 25: I think we could actually remove line 22, and change this line to be `classes << shared_helper.get_margin_bottom if local_assigns[:margin_bottom]`

jon-kirwan: That looks good thanks, i've now updated as above.

danacotoran (commented):
Thanks! 👏  the only thing left: I just spotted you haven't added an entry to `CHANGELOG.md` yet. 
Happy to approve this as soon as that's done.

danacotoran (approved):
thanks 👍

jon-kirwan:
> Thanks! 👏 the only thing left: I just spotted you haven't added an entry to `CHANGELOG.md` yet.
> Happy to approve this as soon as that's done.

🤦 Ah I need to stop doing this. Added now. Thanks again!

PR #2317, opened 2021-09-22
Update publishing docs

andysellick:
## What
Update the guidance for creating a new version of the gem.

## Why
The line I've deleted encouraged readers to upgrade all the packages in the `Gemfile.lock` when versioning the gem, which would introduce potential further changes to that file. If upgraded packages are required, they should be included in the PRs that introduce the change that depends on them, or be handled by dependabot. The only purpose of a PR to version the gem should be to version the gem.

## Visual Changes
None.

### DIFF
--- a/docs/publishing-to-rubygems.md
+++ b/docs/publishing-to-rubygems.md
@@ -19,15 +19,13 @@ Before publishing a new version [check the open and approved pull requests](http
 
 4. Update [`lib/govuk_publishing_components/version.rb`](/lib/govuk_publishing_components/version.rb) version with the new version number.
 
-5. Run `bundle install && yarn install` to ensure you have the latest dependencies installed.
-
-6. Commit changes. These should include updates in the following files:
+5. Commit changes. These should include updates in the following files:
   - [`CHANGELOG.md`](/CHANGELOG.md)
   - [`lib/govuk_publishing_components/version.rb`](/lib/govuk_publishing_components/version.rb)
   - [`Gemfile.lock`](/Gemfile.lock)
 
-7. Create a pull request and copy the changelog text for the current version in the pull request description.
+6. Create a pull request and copy the changelog text for the current version in the pull request description.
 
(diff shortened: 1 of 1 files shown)


maxgds (approved):
I only recently realised this was being recommended and did wonder why!

injms:
Dependabot only updates a subset of the dependencies ([for example](https://github.com/alphagov/govuk_publishing_components/blob/9299d5599938ce947da7d6ce0658ddb13725753d/.github/dependabot.yml#L9-L28)), not all of them - with this in mind, I don't think that this change should be made.

maxgds:
> Dependabot only updates a subset of the dependencies ([for example](https://github.com/alphagov/govuk_publishing_components/blob/9299d5599938ce947da7d6ce0658ddb13725753d/.github/dependabot.yml#L9-L28)), not all of them - with this in mind, I don't think that this change should be made without digging into why this is.

I'm in two minds here. I don't think we're following the advice as it was written, and agree with Andy that a version bumping PR should just be version bumping, not introducing other updates. On the other hand I'm loathe to see security issues leak in from us not updating dependencies.

injms:
> > Dependabot only updates a subset of the dependencies ([for example](https://github.com/alphagov/govuk_publishing_components/blob/9299d5599938ce947da7d6ce0658ddb13725753d/.github/dependabot.yml#L9-L28)), not all of them - with this in mind, I don't think that this change should be made without digging into why this is.
> 
> I'm in two minds here. I don't think we're following the advice as it was written, and agree with Andy that a version bumping PR should just be version bumping, not introducing other updates. On the other hand I'm loathe to see security issues leak in from us not updating dependencies.

I think that bumping dependencies _is_ part of releasing the gem - they're part of the update to this gem. I also think that if bumping the dependencies becomes its own task then no one will do it as we're all so busy - so making it part of releasing the gem is a good idea. However, we (and I include myself in this) haven't been doing that most of the time - so maybe it's just that we're out of practice.

andysellick:
I didn't realise that dependabot only maintained some of the dependencies. In that case, lacking a better option, I'm going to close this.

PR #2320, opened 2021-09-23
Allow custom logo link in navigation header

injms:
## What

Adds the `logo_link` parameter to the navigation header.

## Why

So the navigation header can have a custom link for the crown logo, which make the behaviour of the other header and will stop it from breaking the smoke tests.

## Visual Changes

None.


chao-xian (approved):

owenatgov (commented):
Looks good to me. BUT. Could we add something to the docs for the super nav?

PR #2321, opened 2021-09-23
Allow data attributes to be applied to big numbers when no link is present

owenatgov:
## What
Amends the data attributes functionality on the big number component to allow for attr's to be applied when no link is present. Currently, data attributes are only applied to the `a` tag generated if an `href` attribute is passed. This applies attributes to the `span` containing the number if they are present.

## Why
Part of getting the big numbers onto [the departments page](https://www.gov.uk/government/organisations). This page has a live search which updates the current implementation of the big numbers. Currently, the big number component can't work with this functionality in a robust manner.

No visual changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Allow custom logo link in navigation header ([PR #2320](https://github.com/alphagov/govuk_publishing_components/pull/2320)) MINOR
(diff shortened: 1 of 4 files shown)


injms (commented):
Do you think that this needs a test for this functionality?

injms (approved):
4️⃣ 2️⃣ 👍

PR #2326, opened 2021-09-29
Correctly translate list of nations in Devolved Nations component

brucebolt:
## What
Allows the connector word in the Devolved Nations component to be correctly translated.

## Why
At the moment, we are always using the English word `and` to join lists of three or more nations, but using the language default for two items.

Where there isn't a translation for `Applies to` or the nation names (e.g. in Welsh), we fallback to English for all the string, except the word `and` for two item lists, e.g. `Applies to England a Wales` (where `a` is the Welsh word for `and`).

Following this work, the fallback string will all be in English, i.e. `Applies to England and Wales` and we will be able to have the entire string translated in the future.

## Visual Changes
For Welsh content applicable to two nations, from:

```
Applies to England a Wales
```

to

```
Applies to England and Wales
```

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4723387)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
++ Correctly translate the connector word in the Devolved Nations component ([PR #2326](https://github.com/alphagov/govuk_publishing_components/pull/2326))
+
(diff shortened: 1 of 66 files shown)


hannako (approved):
I didn't know about the Rails to_sentence method, very nifty.

PR #2330, opened 2021-09-30
Add blue arrow option to action link

andysellick:
## What
Adds an option to put a blue arrow on the action link.

## Why
Because there aren't enough options on this component already. No, actually it's because we want this for some redesign changes on the coronavirus page.

## Visual Changes

Screenshot 2021-09-30 at 17 01 10

Trello card: https://trello.com/c/KeJVLjie/649-collections-frontend-implementation-of-the-new-coronavirus-header-design


app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -92,6 +92,18 @@
   }
 }
 
+.gem-c-action-link--blue-arrow {
+  &:before {
+    width: 35px;
+    height: 30px;
+    background: image-url("govuk_publishing_components/action-link-arrow--blue.png");
```
danacotoran, line 99: I think you forgot to add the PNG to `assets/images`.

danacotoran:
It's not within the scope of this PR but I noticed some of the variants basically have the same styling with the only different thing being the image used as the arrow. I wonder if we can consolidate that using SASS mixins to make the component stylesheet less repetitious.

app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
andysellick: 🤦

danacotoran (approved):
➡️ 🚀 !

PR #2333, opened 2021-10-04
Add legend option for radio input

chris-gds:
## What

Adding a `legend` and styled `legend` options to the radio input element so that the [Contact us Page can be updated with a less confusing heading structure.](https://github.com/alphagov/feedback/pull/1300)

## Why

When a radio input is used on a page a `heading` may not be required for accessibility.  A `legend` however would be required.  This adds this option as well as providing visual styling options should the the legend need to have different size or be bold

There is also an update to include a possible error state and clarification of language when as heading is used to indicate this is [nested inside a legend](https://github.com/whatwg/html/pull/5090 )

[There are some on-going accessibility considerations regarding a heading nested inside a legend](https://github.com/alphagov/govuk-frontend/issues/1605)

## Visual Changes

Two additional example options

Two additional example options to use legend and legends that are styled, in this option "m"

## Anything else

This update - when realised - will cascade into this [PR on feedback](https://github.com/alphagov/feedback/pull/1300) to gain the changes.


jon-kirwan:
Looks and works fine for me. The only very minor thing would be to suggest if both commits should be merged in to one (though I understand the two commits address different things so maybe they should be kept separate).

app/views/govuk_publishing_components/components/_radio.html.erb
```diff
@@ -122,7 +122,7 @@
           <% end %>
 
           <% if item[:conditional] %>
-            <%= tag.div item[:conditional], class:"govuk-radios__conditional govuk-radios__conditional--hidden", id: conditional_id %>
+            <%= tag.div item[:conditional], class:"govuk-radios__conditional govuk-radios__conditional--hidden govuk-body", id: conditional_id %>
```
andysellick, line 125: I'd be hesitant to change this as presumably wherever this is used the content passed to the component is already styled by the tags and classes inside it - I don't know what effect wrapping that in a `govuk-body` class would have.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -52,6 +52,29 @@ examples:
         text: "Use Government Gateway"
       - value: "govuk-verify"
         text: "Use GOV.UK Verify"
+  with_legend:
+    description: This adds a `legend` element containing the text
```
andysellick, line 56: I'd suggest putting this next to the heading example, as they're linked and we want people to find them. Might even be worth combining the two into the `with_legend` example, with a description like:

```
This adds a legend element containing the text supplied.

By default, this will render as a H2 inside the legend. This can be changed to a different heading level using the heading_level option.

If a legend without a heading is required, set heading_level to 0.
```

andysellick: I'm also wondering about replacing most of the places where we use `heading` with `legend`, as that's more accurate.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -52,6 +52,29 @@ examples:
...
+    data:
+      name: "radio-group-legend"
+      heading: "What's it to do with?"
+      heading_level: 0
+      items:
+      - value: "yes"
+        text: "Yes"
+      - value: "no"
+        text: "No"
+  with_legend_custom_size:
```
andysellick, line 66: This feels like a duplication of the `with_custom_heading_size` example. I think it might be better to expand on the detail of the existing example, rather than add another.

spec/components/radio_spec.rb
```diff
@@ -148,6 +148,37 @@ def component_name
...
+      heading: "What is your favourite smartie?",
+      heading_level: 0,
+      items: [
+        { label: "Red", value: "red" },
+        { label: "Blue", value: "blue" },
+      ],
+    )
+    assert_select ".govuk-radios"
+    assert_select "legend", "What is your favourite smartie?"
+    assert_select "legend span.govuk-fieldset__heading", true
```
andysellick, line 163: This test should probably also check that `assert_select "legend h2", false`.

spec/components/radio_spec.rb
```diff
@@ -148,6 +148,37 @@ def component_name
...
+  end
+
+  it "renders radio-group with a custom legend size" do
```
andysellick, line 166: I'm not sure this test is necessary - testing the size is already done by the `it "renders radio-group with a custom page heading size"` and similar tests.

app/views/govuk_publishing_components/components/_radio.html.erb
chris-gds: 🤔  Good point, let me check some more instances

app/views/govuk_publishing_components/components/docs/radio.yml
chris-gds: I thought this also, do you think it's worth thinking about changing the element as well:

```
heading: "What's it to do with?"
heading_level: 0
```

```
legend: "What's it to do with?"
heading_level: 0
```

spec/components/radio_spec.rb
chris-gds: I thought this was _slightly_ different as it checks the `legend > span` is being styled?

spec/components/radio_spec.rb
chris-gds: Makes sense

app/views/govuk_publishing_components/components/docs/radio.yml
andysellick: It's probably worth doing, but it would introduce a breaking change and make this whole thing more complicated. Up to you.

spec/components/radio_spec.rb
andysellick: Possibly. Could this be folded into one of the existing tests then?

app/views/govuk_publishing_components/components/docs/radio.yml
chris-gds: My stomach is saying - I'll note but leave this one for the moment for when there's a bit more confidence with this type change.  However I'll update the examples.

chris-gds: > I'm also wondering about replacing most of the places where we use heading with legend, as that's more accurate.

It's semi accurate, it's _nested_ headings in this instance. I think I'm going to mentioned this rather than change them all to legends

spec/components/radio_spec.rb
chris-gds: So with this one, there was no other test that had `heading_level` set to `0` as a result created a `span` I did think this was slightly different and couldn't quite roll this into another one.

app/views/govuk_publishing_components/components/_radio.html.erb
chris-gds: Decided to drop this commit. I need to properly test with all the other instances where this appears across the site but _finding_ them is out of scope of this. Will capture.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -85,8 +85,22 @@ examples:
         text: "Use Government Gateway"
       - value: "govuk-verify"
         text: "Use GOV.UK Verify"
+  with_legend:
+    description: |
+      This adds a `legend` element containing the text supplied. 
+      
+      By default, this will render as a `h2` inside the legend. This can be changed to a different `heading` level using the `heading_level` option to `0`
```
andysellick, line 92: Wording suggestion: `Legend text is automatically wrapped inside a H2. To render the text without it,  heading_level must be set to 0.`

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -95,6 +109,42 @@ examples:
         text: "Yes"
       - value: "no"
         text: "No"
+  with_custom_heading_size:
```
andysellick, line 112: I think there's too many variations on the same thing and it's getting confusing. Suggestion:

- Keep `with_legend`
- Remove `with_heading`
- Rename `with_custom_heading_size` to `with_heading` and use it to demonstrate both the heading_level and heading_size options
- Remove `with_custom_heading_level`
- Create a new example `with_heading_1` that demonstrates the `if heading level is 1 and heading_size is not set` example

I think that covers all the options, but point out if I've missed any.

app/views/govuk_publishing_components/components/docs/radio.yml
```diff
@@ -109,8 +159,9 @@ examples:
         text: "Blue"
   with_page_header_and_caption:
     description: |
-      If a caption text is provided with a page heading, it will be displayed above the heading.
-      A caption can only be used with a page heading. If a heading is not provided the caption will not render.
+      If a caption text is provided with a page nested `heading`, it will be displayed above the nested `heading`.
```
andysellick, line 162: I'm not sure what is meant by a `page nested heading`. Suggestion:

```
A caption can be added using the heading_caption option. Captions will only be displayed if text for the heading option is present.

The pattern is used across GOV.UK to show a high-level section that this page belongs to.
```

spec/components/radio_spec.rb
andysellick: But the options passed to the component for this test exactly match the options in the `renders radio-group with a legend` test on line 151 (unless I've overlooked something). Could the assert_select lines from this test not be moved into that one?

andysellick (changes requested):

PR #2335, opened 2021-10-06
Raise argument error on image card component

alex-ju:
## What
Raise argument error on `image_card` component if no `href` or `extra_details` are provided.

## Why
To prevent this component from silently failing to render.

This is a recommended action following an [incident review](https://docs.google.com/document/d/1DBC7gAhHoNThVqpPxl9DnpCTEfrABthdcSVC0s6RiEw/edit#). There is a wider conversation to have about extending this rule to all components in the library and documenting it.

## Visual Changes
No visual changes

### DIFF
--- a/app/views/govuk_publishing_components/components/_image_card.html.erb
+++ b/app/views/govuk_publishing_components/components/_image_card.html.erb
@@ -14,14 +14,13 @@
   heading_link_classes = %w[
     gem-c-image-card__title-link
     govuk-link
-  ] 
+  ]
   heading_link_classes << brand_helper.color_class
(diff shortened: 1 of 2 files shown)


maxgds:
I'm a bit concerned that this might be a breaking change and that the conversation you mention is important to have. I've a feeling that we sometimes rely on the silent rendering of nothing if all the data isn't supplied and don't have guard clauses to protect against it, though I have no evidence to back this up. What will happen if the data fed to the image card component is compromised on live? Will the page stop rendering due to an error, will the error state appear in the page, or will we just see errors raised in logs?

PR #2337, opened 2021-10-06
Update devolved nations component to accept type property

jon-kirwan:
## What
https://trello.com/c/4SUBn4mb/880-update-devolved-nations-component

Update the component to accept a type property e.g. "Publication"

## Why

Currently, all external links are lablled as guidance.

However, not all content is guidance. We should refer to "publication" as the default but allow a parameter to be passed to define the actual type, eg statistics.  This error was pointed out on basecamp by our stakeholders so we should address it quickly.

## Anything else


app/views/govuk_publishing_components/components/_devolved_nations.html.erb
```diff
@@ -15,7 +17,7 @@
     <% if devolved_nations_helper.nations_with_urls.any? %>
       <%= content_tag :ul, class: "govuk-list govuk-!-margin-top-1 govuk-!-margin-bottom-0" do -%>
         <% devolved_nations_helper.nations_with_urls.each do |k, v| %>
-          <%= content_tag(:li, link_to("Guidance for #{t("components.devolved_nations.#{k}")}", v[:alternative_url], class: "govuk-link")) %>
+          <%= content_tag(:li, link_to("#{type} for #{t("components.devolved_nations.#{k}")}", v[:alternative_url], class: "govuk-link")) %>
```
maxgds, line 20: Can we do anything to clean up the translations here? It seems we only ever have publication or guidance, is that right? Ideally we would be translating the whole phrase Guidance|Publication for |Scotland|Wales|Ireland as you can't guarantee word order in other languages, and also here we're not translating the word "for" at all.

maxgds: All the consultations I found looked like this: https://www.gov.uk/government/consultations/open-access-restriction-at-castle-beeny-farm-2-how-to-comment - basically they're all saying "This applies to England" and offering no alternatives. But I'm by no means certain I'm catching everything.

I think it would make sense to do the translation in the component wholly, it is after all what does the presentation - taking data to make that decision feels cleaner. Do ask around for a second opinion though as I'm not confident that this is definitely the way we should go. We do need to handle the whole string, though, I think, and if the calling app is responsible for the strings you start to question the point of the component, which is why I'm leaning that way.

jon-kirwan: I'll investigate some more and see if anything comes up. I've been trawling through this list here: https://www.gov.uk/search/policy-papers-and-consultations?page=1 but so far nothing.

I think you're right too - i'd wondered about it myself beforehand but since we already had the translations in `government_frontend` decided to leave it there.

I'll get another opinion but does seem a better approach to move it all to the component.

PR #2338, opened 2021-10-06
Change accordion language

chris-gds:
## What

Update accordion use to improve clarity

## Why

Feedback from technical writer to be more inline with Content guidance and improve clarity.


app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -59,7 +59,7 @@ examples:
     description: |
       An `id` is optional as it's automatically generated, but it can be supplied if a specific `id` is required.
 
-      The `id` must be **unique** across the domain of your service - this is because as the open / closed state of individual instances of the accordion persists across page loads using `localStorage`.
+      The `id` must be unique across the domain of your service - this is because the open / closed state of individual instances of the accordion persists across page loads using `localStorage`.
 
       Used as an `id` in the HTML for the accordion as a whole, and also as a prefix for the `id`s of the section contents and the buttons that open them, so that those `id`s can be the target of `aria-labelledby` and `aria-control` attributes.
     data:
```
chris-gds, line 65: > _This sentence is over 25 words, which is our recommended maximum length.
What’s the (grammatical) subject of this sentence?_

app/views/govuk_publishing_components/component_guide/show.html.erb
```diff
@@ -24,7 +24,7 @@
           </div>
         </div>
       <% end %>
-      <p class="govuk-body"><%= link_to "Search for usage of this component on GitHub", @component_doc.github_search_url, class: "govuk-link" %>.</p>
+      <p class="govuk-body"><%= link_to "Search for usage of this component on GitHub.", @component_doc.github_search_url, class: "govuk-link" %>.</p>
```
andysellick, line 27: Usual convention is to not include prior or following punctuation in link text, like the full stop here. Also there's already a full stop at the end of the sentence, so I'm confused why this has been added. Currently looks like this:

Screenshot 2021-10-06 at 16 36 26

andysellick (changes requested):

chris-gds:
>I think we should loop in said technical writer so we can discuss these recommendations and respond to review 

Fair, requesting as a first step

chris-gds:
There's bit quite a few updates to this one, I anticipate a few more to get this to a better place

chris-gds:
Added some of the changes in a separate commit (for ease of review as time has passed) will squash once happy.

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -3,33 +3,34 @@ description: The accordion component lets users show and hide sections of relate
...
-  Section headings must use a button element:
+  Section headings must use a button element so that users:
 
-    * so that sections can be toggled with the space and enter keys
-    * so that sections can't be opened in a new tab or window
+    * can toggle sections with the space and enter keys
+    * cannot open sections in a new tab or window
 
-  When JavaScript is unavailable the component must:
+  When CSS and / or JavaScript is unavailable, the component must:
```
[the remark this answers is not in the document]
EoinShaughnessy: @chris-gds Fair enough! Thanks for explaining.

app/views/govuk_publishing_components/components/docs/accordion.yml
EoinShaughnessy: That's shorter, which is cool, but I wasn't sure whether it was the `id` or the service that uses `localStorage`...

Maybe we could break it up into 2 sentences. Something like:

> The `id` must be unique across the domain of your service. This is because the open or closed state of individual instances of the accordion uses `localStorage` to persist across page loads.

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -183,7 +183,7 @@ examples:
               </ul>'
   with_data_attributes:
     description: |
-      Users can add multiple custom attributes to the accordion to each section. Accepts an hash, so multiple attributes can be added.
+      Adds custom data attributes to each section of the accordion. Accepts an hash, so multiple attributes can be added.
```
andysellick, line 186: I think this should be `Accepts a hash`?

andysellick (commented):
Happy with your response to my comments, seems good overall, one tiny comment.

app/views/govuk_publishing_components/components/docs/accordion.yml
chris-gds: This is the current sentence

`Adds custom data attributes to each section of the accordion. *Accepts an hash*, so multiple attributes can be added.`

Do you mean that this sentence above should perhaps drop or move the first part?

Eg just: `Accepts an hash, so multiple attributes can be added.`

app/views/govuk_publishing_components/components/docs/accordion.yml
chris-gds: Sure, updated to this

image

app/views/govuk_publishing_components/components/docs/accordion.yml
andysellick: No I think it should be `Accepts a hash` and not `Accepts an hash`.

chris-gds: Ahh! Ace, updated

chris-gds:
Hey @andysellick is this OK to go in?

andysellick (approved):
I'm happy if everyone else is.

PR #2353, opened 2021-10-13
Make action link blue icon smaller on mobile

andysellick:
## What
Make the action link blue arrow option icon smaller - should be around 80% smaller on mobile than desktop. Originally there was no size difference.

## Why
To fit with design.

## Visual Changes

Before and after

blue_arrow_example

Desktop, top (unchanged)
Mobile, bottom (now smaller)

Trello card: https://trello.com/c/IdVDZamX/41-adapt-landing-page-desktop-layout-to-show-h2s-in-the-left-hand-column

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


andysellick:
@danacotoran thanks, that's the second time I've done that today, I seem to be having a strange afternoon. Fixed.

PR #2362, opened 2021-10-14
Raise deprecation warnings on unused components

alex-ju:
## What
Raise deprecation warnings on unused components that we plan to retire in 28.0.0

I've been pondering whether or not to raise these using `ActiveSupport`, but thinking that we use `govuk_publishing_components` outside the Rails realm (e.g. developer docs using Middleman) I'm getting skeptical about it. 
Happy to hear suggestions for a more effective approach.

## Why
We plan to retire a few components that are not being used and we don't plan to use them in the near future.

## Visual Changes
No visual changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Raise deprecation warnings on unused components: `admin_analytics`, `government_navigation`, `highlight_boxes` and `taxonomy_list` ([PR #2362](https://github.com/alphagov/govuk_publishing_components/pull/2362))
 * Adjust cookie_consent parameter conditions ([PR #2399](https://github.com/alphagov/govuk_publishing_components/pull/2399))
 
 ## 27.10.1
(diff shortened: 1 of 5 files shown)


kevindew:
This looks good @alex-ju - the `warn` approach looks good and I like the idea.

A couple of thoughts:

- naming the exact version that will be removing them seems a bit delicate (particularly down to patch level) - just saying in a future major version would be enough https://github.com/alphagov/gds-api-adapters/pull/974/files#diff-cf6b6b3424473a4667a44bb8cea6fe3451382b6f11d644b122008aaad3c0a417R7
- do we want to advise anyone to do something if they do use the component? often a deprecation carries a "do this instead" but understand these components may just no longer be used by anything
- if it is that these components just aren't used anymore do we have a purpose of the message? i.e do we plan to search the logs for it. There's a part of me that's wondering whether we expect these warnings to do anything or if we'd have the same effect by just deleting them without a warning (i.e. no-one monitors for the warnings)
- is there a timeframe until v28 or their removal? say if someone commits a breaking change tomorrow do we then remove them for next weeks release?

Sorry - I feel like my questions are a bit on the over-thinking side!

injms (approved):

injms:
My only thought is to add this to the changelog to highlight that deprecations are coming - not essential, but could be useful?

alex-ju:
> My only thought is to add this to the changelog to highlight that deprecations are coming - not essential, but could be useful?

Great suggestion, Ian – will do so

PR #2399, opened 2021-11-01
Adjust cookie_consent parameter conditions

danacotoran:
The `cookie_consent` param should also have a value of 'not-engaged' when `cookies_preferences_set` is `null`.  
The reason for this change can be observed by doing the following:
1. clear your browser cookies for gov.uk –  when you refresh the page the cookie banner should appear
2. resist the immediate urge to reject (or accept) cookies 😄
3. visit https://www.gov.uk/brexit and check the `href` of the sign in link – the query string on the link is `cookie_consent=reject`, when it should probably be `cookie_consent=not-engaged`

This adjusts the explicit cross domain links script to always append `cookie_consent=not-engaged` unless `cookies_preferences_set` is explicitly set to 'true'.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
@@ -47,8 +47,14 @@ describe('Explicit cross-domain linker', function () {
       element = $('<a href="/somewhere">')
     })
 
-    it('modifies the link href to append cookie_consent parameter "not-engaged" if cookie preferences are not set', function () {
+    it('modifies the link href to append cookie_consent parameter "not-engaged" if cookies_preferences_set cookie is "false"', function () {
       GOVUK.cookie('cookies_preferences_set', 'false')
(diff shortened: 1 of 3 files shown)


spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
```diff
@@ -47,7 +47,7 @@ describe('Explicit cross-domain linker', function () {
       element = $('<a href="/somewhere">')
     })
 
-    it('modifies the link href to append cookie_consent parameter "not-engaged" if cookie preferences are not set', function () {
+    it('modifies the link href to append cookie_consent parameter "not-engaged" if cookies_preferences_set cookie is not "true"', function () {
```
barrucadu, line 50: Can we have a test for the new case?  Presumably this catches the case where there is no `cookie_preferences_set` cookie

danacotoran: Thanks, I've updated the spec

barrucadu (approved):

PR #2400, opened 2021-11-01
Include Rails Translation Manager and set correct root

1pretz1:
This PR includes Rails Translation Manager and sets the actual Gem root when called by RTM.

Supersedes https://github.com/alphagov/rails_translation_manager/pull/17.

More info in the commits ->

---

Trello:
https://trello.com/c/k4rrt3UG/2686-update-csv-import-rails-translation-manager-3


ChrisBAshton (approved):
Really nice 👍 I'm pleased the 'hack' is in here rather than in rails_translation_manager.

Issue #2331, opened 2021-09-30
Navigation items misaligned on Layout super nav component


fofr:
https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header

The "Departments" text is 1px misaligned with the two drop downs either side of it:

![Screenshot 2021-09-30 at 17 00 39](https://user-images.githubusercontent.com/319055/135492147-a0b7223c-b2f9-430c-91e1-2cdf3ed7ffe9.png)

There’s also a 1px misalignment on the search box toggle when open and focused vs closed and focused (note the black bar on the X image):
![Screenshot 2021-09-30 at 17 08 09](https://user-images.githubusercontent.com/319055/135492415-f392e565-7fd3-4902-8f59-9fffb8996b91.png)
![Screenshot 2021-09-30 at 17 08 06](https://user-images.githubusercontent.com/319055/135492421-217d5bf0-a804-40c4-9eb8-0befe4abf8ef.png)

injms:
Firstly, thanks for reporting these visual issues.

The misalignment of "Departments" should hopefully be fixed in the coming week or so.

The black line on the button is deliberately placed to differentiate between the cookie banner and open section:

![image](https://user-images.githubusercontent.com/1732331/135508946-d83a0278-af38-481a-af0e-f6948b78a6a9.png)

This small border isn't an ideal solution - it's going to be iterated on to find something that doesn't look so glitchy when the cookie banner isn't present but doesn't make the cookie banner and an open section look connected.

injms:
I'm closing this as the alignment was fixed in https://github.com/alphagov/govuk_publishing_components/pull/2401 - which has now been [released in the gem](https://github.com/alphagov/govuk_publishing_components/pull/2403) and then [bumped in the application that renders the header](https://github.com/alphagov/static/pull/2639).

PR #2401, opened 2021-11-01
Update navigation header focus states (take two)

injms:
## What
Repeat of https://github.com/alphagov/govuk_publishing_components/pull/2319 which was reverted in https://github.com/alphagov/govuk_publishing_components/pull/2395. 

Should fix the reason for the reversion - #2319 was causing the end to end tests to fail, specifically when a page rendered by a frontend application was visited. 

These were failing with a not very verbose 500 error - the theory is that this was caused by a closing `</div>` tag being omitted, which would cause the Nokogiri parser to throw an error in Static, which would then return a 500 error. 

This pull request outputs valid HTML with the correct number of closing tags, so should not cause the same error to persist.

Updates the focus states on the navigation elements to use [`:focus-visible`][mdn] with a fallback for browsers that don't support that pseudo-selector.

Also fixes misalignment reported in #2331.

## Why

Using `:focus` means that the focus state is triggered when a user clicks on a button - this could lead to users being uncertain to what the state of the button is, as the button stays focussed when the cursor moves away from the button.

Using `:focus-visible` means that the focus state is only used when an element is tabbed to - the focus remains until the next item is navigated to.

### No-JavaScript layout

| Mobile before | After |
| - | - |
| image | image

Desktop before (navigation items are showing focus, hover, no-state):

www gov uk_ (1)

Desktop after (navigation items are showing focus, hover, no-state):
govuk-publishing-components dev gov uk_public (3)

### JavaScript available layout

| Screen size | Before | After |
| - | - | - |
| < 360px | image | image |
| > 360px | components publishing service gov uk_public(iPhone X) | govuk-publishing-components dev gov uk_public(iPhone X) |

Desktop before:

components publishing service gov uk_public (1)

Desktop after:

govuk-publishing-components dev gov uk_public (4)

Menu open before:

components publishing service gov uk_public (2)

Menu open after:

govuk-publishing-components dev gov uk_public (6)

[mdn]: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible


app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -314,7 +221,11 @@ $search-icon-size: 20px;
 }
 
 .gem-c-layout-super-navigation-header__navigation-items {
-  .js-module-initialised & {
+  @include govuk-media-query($from: "desktop") {
+    display: inline-block;
+  }
+
+  .js-enabled & {
```
owenatgov, line 228: We'd want to swap this out for `.js-module-initialised` in line with https://github.com/alphagov/govuk_publishing_components/pull/2388

```suggestion
  .js-module-initialised & {
```

Same thing on lines 370, 538 and 859.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -443,230 +535,196 @@ $search-icon-size: 20px;
...
+      border-bottom-color: govuk-colour("light-grey");
+      color: govuk-colour("light-grey");
     }
   }
 }
 
+.gem-c-layout-super-navigation-header__navigation-top-toggle-close-icon {
+  color: $govuk-text-colour;
+  display: none;
+  font-size: 36px;
```
owenatgov, line 703: Not a comment for this PR really but I wonder if we should feedback to the DS team if they should allow for a version of `govuk-font` which doesn't whack in a bunch of media queries.

owenatgov (commented):
One comment. Looking swell otherwise 👍 

I should've caught that missing closing div tag when I reviewed this, sorry 😞  Did you manage to test this against local e2e's?

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
injms: Good catch; swapped the classes back in https://github.com/alphagov/govuk_publishing_components/pull/2401/commits/f1893d5d62368fa84a02ae0b64d3d9a6a76b556c.

injms:
> One comment. Looking swell otherwise 👍
> 
> I should've caught that missing closing div tag when I reviewed this, sorry 😞 Did you manage to test this against local e2e's?

Nah, the `</div>` was there but cunningly hidden inside an `if` statement - and a browser's more relaxed approach to closing tags meant the component behaved as expected. Only by using the old-school view source was the error highlighted; and I ran a couple of pages through the W3C markup validator to double check it was okay this time. It was an error so basic it shouldn't even need checking for.

I haven't managed to run this against local e2e tests yet - mainly because I haven't been able to run the e2e tests locally at all yet. I've run Static locally with this change, and then run `frontend`, `collections`, and `government-frontend` tests locally which all passed okay.

owenatgov (approved):
I'm gonna approve this as we're under less pressure right now to get something out so we have more time to debug if e2e's end up still kicking off.

PR #2407, opened 2021-11-02
Fix font size on super navigation header

alex-ju:
## What
Rem units are not meant to be used in compatibility mode (in conjunction with legacy dependencies such as `govuk_template`) as the `font-size` for the root element is not controlled. Hence we wrap these declarations and only apply them when rem units are set to be used by `govuk-frontend`.

## Why
To fix the font size when the super navigation header is used in compatibility mode (e.g. https://www.gov.uk/search).

## Visual Changes
<table>
<tr><td valign="top">

www gov uk_search (1)

</td><td valign="top">

www gov uk_search

</td></tr>
</table>

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -271,7 +271,9 @@ $chevron-indent-spacing: 7px;
 
   display: inline-block;
   font-size: 19px;
-  font-size: govuk-px-to-rem(19px);
+  @if $govuk-typography-use-rem {
+    font-size: govuk-px-to-rem(19px);
+  }
   font-weight: bold;
   margin: govuk-spacing(3) 0;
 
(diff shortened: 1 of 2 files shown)


injms (approved):
Looks good - I think this needs a changelog entry if this is going to be released on its own.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -312,7 +314,9 @@ $chevron-indent-spacing: 7px;
     @include govuk-media-query($from: "desktop") {
       float: left;
       font-size: 16px;
-      font-size: govuk-px-to-rem(16px);
+      @if $govuk-typography-use-rem {
+        font-size: govuk-px-to-rem(16px);
+      }
```
injms, line 319:
```suggestion
      // stylelint-disable max-nesting-depth
      @if $govuk-typography-use-rem {
        font-size: govuk-px-to-rem(16px);
      }
      // stylelint-enable max-nesting-depth
```

PR #2411, opened 2021-11-03
Extend new scroll tracker functionality

andysellick:
## What
Extend the functionality of the new scroll tracking script, to provide new features. Specifically:

- allows specific URLs to be targeted or excluded from scroll tracking, where a template renders more than one URL
- allows specific URLs to track custom percentage positions or specific headings only

## Why
Attempts to migrate scroll tracking from the old script to the new script have highlighted features that require implementation.

## Visual Changes
None.

Trello card: https://trello.com/c/sdugtMbX/37-migrate-to-new-scroll-tracker

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
@@ -31,9 +31,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       return
     }
 
+    this.trackType = this.$module.getAttribute('data-track-type')
+    var trackHeadings = this.$module.getAttribute('data-track-headings')
+    if (trackHeadings) {
+      try {
+        this.config.trackHeadings = JSON.parse(trackHeadings)
+      } catch (e) {
+        // if there's a problem with the config, don't start the tracker
+        console.error('Scroll tracker configuration error: ' + e.message, window.location)
+        window.GOVUK.analyticsVars.scrollTrackerStarted = false
+        return
+      }
+    }
+
     window.GOVUK.analyticsVars.scrollTrackerStarted = true
 
-    this.trackType = this.$module.getAttribute('data-track-type')
     if (this.trackType === 'headings') {
       this.track = new AutoScrollTracker.Heading(this.config)
     } else {
@@ -145,15 +157,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   // check heading is inside allowed elements, generally ignores everything outside of page content
   AutoScrollTracker.Heading.prototype.findAllowedHeadings = function () {
-    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
     var headingsFound = []
+    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+    var trackHeadings = this.config.trackHeadings
 
+    // this is a loop that only happens once as we currently only have one
+    // allowed element for headings to be in - 'main'
     for (var h = 0; h < this.config.allowHeadingsInside.length; h++) {
       var insideElements = document.querySelectorAll(this.config.allowHeadingsInside[h])
       for (var e = 0; e < insideElements.length; e++) {
         var found = insideElements[e].querySelectorAll(headings)
         for (var f = 0; f < found.length; f++) {
-          headingsFound.push(found[f])
+          if (trackHeadings) {
+            if (trackHeadings.includes(found[f].textContent.trim())) {
+              headingsFound.push(found[f])
+            }
+          } else {
+            headingsFound.push(found[f])
+          }
         }
       }
     }
--- a/docs/analytics/auto_scroll_tracker.md
+++ b/docs/analytics/auto_scroll_tracker.md
@@ -20,11 +20,31 @@ AutoScrollTracker is a GOVUK.Module and can be initialised by adding the relevan
 
 By default, AutoScrollTracker tracks by percentage scrolled. Specifically, it will fire a GA event when the user scrolls to 20%, 40%, 60%, 80% and 100% of the page height.
 
-It can track headings instead of percentages, using this additional option:
+It can track headings instead of percentages, using the `data-track-type` option.
 
 ```html
 <% content_for :extra_head_content do %>
   <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings"/>
+<% end %>
+```
+
+If only some headings need to be tracked, they can be specified using the `data-track-headings` option as shown. This option has been included with the module to support legacy tracking. It is not recommended for future use as headings on pages can change, which would prevent the tracking from working.
+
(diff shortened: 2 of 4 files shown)


alex-ju (approved):

kevindew:
Thanks for opening this Andy. I've not really looked too much into the implementation and wanted to question the rationale a bit. 

To me it seems a bit strange to have a meta tag for a page that basically says "scroll track this page" and then have the data attributes say "but not this page". Almost like it seems a bit heavy on the compute to set-up a module and then have it to be configured to be a no-op. 

It'd seem slightly simpler and more decoupled if the logic for whether a page should be tracked or not should be handled in the app that set's the meta tag. So instead of having code like:

```html
<% content_for :extra_head_content do %>
  <%
    track_details = "{
      'include': [
        {
          'path': '/foreign-travel-advice/benin'
        },
        {
          'path': '/foreign-travel-advice/france'
        },
      ]
    }".squish
  %>
  <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-details="<%= track_details %>"/>
<% end %>
```

instead have:

```html
<% if ["/foreign-travel-advice/benin", "/foreign-travel-advice/france"].include?(content_item.base_path) %>
  <% content_for :extra_head_content do %>
    <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker">
  <% end %>
<% end %>
```

and then this module doesn't have to have its own logic, and all the "will it run" logic is in the app that calls the module.

Perhaps I'm missing scenarios that make this hard though?

And apologies, I know it's no fun receiving a comment like this having done all the implementation and tests.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -203,15 +213,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   // check heading is inside allowed elements, generally ignores everything outside of page content
   AutoScrollTracker.Heading.prototype.findAllowedHeadings = function () {
-    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
     var headingsFound = []
+    var headings = ['h1', 'h2', 'h3', 'h4', 'h5', 'h6']
+    var trackSpecificHeadings = this.config.trackSpecificHeadings
 
+    // this is a loop that only happens once as we currently only have one
+    // allowed element for headings to be in - 'main'
```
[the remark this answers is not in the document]
andysellick: Yeah, me too. I put this in so that we could potentially extend it in future - for example if we needed to track headings outside of the `main` element. I might just take it out and hardwire the `main` element though, will have a think.

andysellick:
@kevindew no need to apologise! This is exactly why I requested your review. I kept thinking while I was writing this that it felt a little brittle and over complicated, and you've provided exactly what I need.

Have updated this PR so the module only now has the option to track specific headings, and updated the doc to include details of your example. Thanks very much 👍

docs/analytics/auto_scroll_tracker.md
```diff
@@ -20,14 +20,34 @@ AutoScrollTracker is a GOVUK.Module and can be initialised by adding the relevan
...
 <% content_for :extra_head_content do %>
   <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings"/>
 <% end %>
 ```
 
+If only some headings need to be tracked, they can be specified using the `data-track-headings` option as shown. This option has been included with the module to support legacy tracking. It is not recommended for future use as headings on pages can change, which would prevent the tracking from working.
+
+```html
+<% content_for :extra_head_content do %>
+  <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings" data-track-headings="['Example heading', 'Another example heading']"/>
```
kevindew, line 35: I imagine the example here might be better illustrated by using Ruby to_json. 
```suggestion
  <meta name="govuk:scroll-tracker" content="" data-module="auto-scroll-tracker" data-track-type="headings" data-track-headings='<%= ["Example heading", "Another example heading"].to_json %>' />
```

as it seems a bit dubious suggesting people handroll their JSON. Better still as:

```
<%= tag.meta name: "govuk-scroll-tracker", content: "", data: { module: "auto-scroll-tracker", "track-type": "headings", "track-headings": ["Example heading", "Another example heading"].to_json } %>
```

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
```diff
@@ -31,9 +31,21 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       return
     }
 
+    this.trackType = this.$module.getAttribute('data-track-type')
+    var trackHeadings = this.$module.getAttribute('data-track-headings')
+    if (trackHeadings) {
+      try {
+        // Valid JSON doesn't use '', but we can use replace to be forgiving
+        this.config.trackHeadings = JSON.parse(trackHeadings.replace(/'/g, '"'))
+      } catch {
```
kevindew, line 40: Should we do anything with the error? it seems like it'd be easy for no-one to know there is a problem here - perhaps console.error?

kevindew (commented):
Looks good - just a couple of comments.

It feels a little fishy to me to do the JSON quote fix - is that something we need to support from the previous version?

docs/analytics/auto_scroll_tracker.md
andysellick: Great, thanks, have updated.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Good idea, I've put a console.error in.

docs/analytics/auto_scroll_tracker.md
andysellick: Also have removed the JSON quote fix - I was trying to make the script more flexible, but if you don't think it's necessary that's fine.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
kevindew: Super. I'd suggest including the error message:

```
} catch (e) {
  console.error('Scroll tracker configuration error: ' + e.message, window.location)
  // ...
}
```

docs/analytics/auto_scroll_tracker.md
kevindew: Super - I hope it'll be ok. It's potentially a risk, but if the script is forgiving then it creates an impression of JSON flexibility which the language doesn't have which may catch people out any other places.

app/assets/javascripts/govuk_publishing_components/analytics/auto-scroll-tracker.js
andysellick: Updated as suggested, thanks.

PR #2428, opened 2021-11-09
Auditing enhancements

andysellick:
## What
Improvements to the components auditing, specifically:

- adds links to github for in the warnings section (for evidence of component classes and jQuery references), to aid navigation
- removes auditing for applications that are no longer in use
- add details for when components in the gem have a helper file
- adds counts to details of components in the gem (e.g. x components have a helper*)
- add links to asset files in github in component details

*Not all of the helpers follow a consistent naming convention, so the helper count isn't accurate yet. This PR renames the attachment helper, but I keep getting merge conflicts on it so I'll update the others in a separate PR rather than having to keep updating this one.

## Why
Understanding our components and how they are used across GOV.UK is important, particularly as our components increase in number and complexity.

## Visual Changes

Showing additions to the component details section (links, counts, helper column added):

Screenshot 2022-03-22 at 10 38 48


app/views/govuk_publishing_components/audit/_component_contents.html.erb
```diff
@@ -0,0 +1,82 @@
...
+      <tr class="govuk-table__row">
+        <th scope="row" class="govuk-table__header">
+          <a href="<%= component[:link] %>" class="govuk-link"><%= component[:name] %></a>
+        </th>
+        <td class="govuk-table__cell">
+          <% if component[:stylesheet] %>
+            <strong class="govuk-tag govuk-tag--green">
+              <a href="<%= component[:stylesheet] %>" class="govuk-link">Yes</a>
+            </strong>
+          <% end %>
```
injms, line 42: If `component[:stylesheet]` isn't present, we're left with an empty table cell which isn't ideal. Could there be a non-strong "No" present?
```suggestion
          <% if component[:stylesheet] %>
            <strong class="govuk-tag govuk-tag--green">
              <a href="<%= component[:stylesheet] %>" class="govuk-link">Yes</a>
            </strong>
          <% else %>
            No
          <% end %>
```

injms (approved):
This looks super! One small question that shouldn't block this from being merged in 👍

app/views/govuk_publishing_components/audit/_component_contents.html.erb
andysellick: That's a fair point, but there's a lot of information in this table and it's currently easier to compare 'yes' with empty than compare 'yes' with 'no'. Thanks for making this non-blocking, maybe if we can get the auditing to a better position it might be good to do a review of the pages and include this question in it.

kevindew:
@andysellick do we consider the classes in `GovukPublishingComponents::Presenters` to be part of the public API of this gem?

I've noticed that Whitehall is failing on this update with errors: `uninitialized constant GovukPublishingComponents::Presenters::Attachment` and wondering if a) they are a public API and this should have been a major change or b) they're a private API and Whitehall probably shouldn't be using them directly.

andysellick:
@kevindew argh sorry, I spotted that one of the publishing apps had this problem but hadn't noticed it in whitehall. I can look at that PR.

To answer your question, I think we consider them part of the public API, but we currently have no mechanism for keeping any kind of track of how this is used (I'm hoping to add something to the auditing). I'd like to have a situation where if a change like this is introduced we can be aware of potential problems that changes like this will cause ahead of time.

kevindew:
> @kevindew argh sorry, I spotted that one of the publishing apps had this problem but hadn't noticed it in whitehall. I can look at that PR.
> 
> To answer your question, I think we consider them part of the public API, but we currently have no mechanism for keeping any kind of track of how this is used (I'm hoping to add something to the auditing). I'd like to have a situation where if a change like this is introduced we can be aware of potential problems that changes like this will cause ahead of time.

That's cool - you don't have to fix it. I was just wondering whether it was encouraged or discouraged usage. 

We could use documentation comments as a way to specify which APIs are public or not - I can't remember off the top of my head but I think there could be some [YARD](https://github.com/lsegal/yard) annotations that could help or could just tag some things with `# Public API`

PR #2432, opened 2021-11-10
Remove Brexit checker callout

sihugh:
## What
Remove the version of the Brexit call to action that links directly to the checker. It's only used on [the Brexit](https://www.gov.uk/guidance/brexit-guidance-for-businesses) [hub pages](https://www.gov.uk/guidance/brexit-guidance-for-individuals-and-families). 

## Why
The Brexit checker is being retired in 3 weeks' time. It was invented for a time when there were lots of high impact changes happening in a short period of time. This is no longer the case, so it's becoming less useful (and less used). 

We don't want to send people to the checker now, because it's probably not going to give them what they need.

### Retiring this version of the Brexit sidebar nav 👇 

Screenshot 2021-11-10 at 09 08 19

### The only one remaining is this version 👇 

Screenshot 2021-11-10 at 09 08 41


alex-ju (approved):
🧹

PR #2435, opened 2021-11-10
Update feedback component to fix accessibility issues

jon-kirwan:
## Note (local testing)
There is an issue whereby the feedback assets (CSS and JS) are currently shipped twice:

- from the application e.g. `government-frontend`
- from `static`.

The assets should only be imported from static. However, at the moment, if we remove the assets at the application level then there are style problems where the feedback component styles are overridden by subsequent imports. An example is documented here: https://github.com/alphagov/whitehall/pull/6240#issuecomment-977888889.

For now, to test locally, it's necessary to:
- update the Gemfile for `gem 'govuk_publishing_components', path: '../govuk_publishing_components'` at **both** the application level (e.g. `government-frontend`) and `static`
- import the textarea component styles `@import 'govuk_publishing_components/components/textarea';` in to `application.scss` for `static`.

## What
https://trello.com/c/MOshcFm1/771-update-feedback-component-to-hide-and-show-content-in-a-more-robust-way
- Stop hiding things with CSS classes and/or "display:none" and hide them with the HTML "hidden" attribute instead. 
- The close buttons should only be revealed at the right time, and the forms themselves should be presented by using JS to switch off the hidden attribute
- When JS is disabled the 'Help us improve GOV.UK' form should display. When JS is enabled the form should be hidden with JS using the HTML attribute "hidden".
- Make the text input fields more usable for longer messages
- Make sure the text input limit is implemented properly and the error message matches it
- Update 'Accessibility acceptance criteria'.

## Why
- Tidy up the feedback component so that it no longer relies on CSS hiding/showing to manage interaction 

## Visual changes
<table role="table">
<thead>
<tr>
<th>Before (inputs)</th>
<th>After (inputs)</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774593-4675c4d6-bac1-40ff-b675-e85f750e889f.png"></a></td>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774583-10718a32-30fc-4be3-9ac6-307a391b7ec3.png"></a></td>
</tr>
</table>

<table role="table">
<thead>
<tr>
<th>Before (JS disabled & CSS enabled)</th>
<th>After (JS disabled & CSS enabled)</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774798-b0fce000-0fba-4a55-b388-db7b2a6e59a6.png"></a></td>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774788-e39890aa-8430-4dba-bd6c-fb1dece21d5d.png"></a></td>
</tr>
</table>

<table role="table">
<thead>
<tr>
<th>Before (JS enabled & CSS disabled)</th>
<th>After (JS enabled & CSS disabled)</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774987-85bb881a-72c9-4fd9-8ee2-ed28ea7b9929.png"></a></td>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141774985-7b4189f6-485e-4c66-9521-63480dd48047.png"></a></td>
</tr>
</table>

<table role="table">
<thead>
<tr>
<th>Before (JS disabled & CSS disabled)</th>
<th>After (JS disabled & CSS disabled)</th>
</tr>
</thead>
<tbody>
<tr>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141775450-5b61667f-0640-4b5b-8f92-de8c73f04c20.png"></a></td>
<td valign="top"><a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/141775444-0899beee-ab6a-482a-a8f0-7d4c2f4cdc59.png"></a></td>
</tr>
</table>

## Anything Else
- `var el = e.target.closest('button')` - just wanted to highlight the use of the `closest()` method here. We seem to have a polyfill included in the bundle for this so I expect it's okay.
- `hidden` can be overridden with display properties e.g. from https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/hidden -
> Note: Changing the value of the CSS display property on an element with the hidden attribute overrides the behaviour. For instance, elements styled display: flex will be displayed despite the hidden attribute's presence.

The best thing seems to be to include something like this `[hidden] { display: none !important; }`. For now, I've scoped it to the feedback component, but I think this should be added to our global styles.

- I have removed occurrences of `aria-hidden` since it should not be added when the HTML hidden attribute is present https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-hidden_attribute

### DIFF
--- a/app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
+++ b/app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
@@ -1,9 +1,10 @@
 <form action="/contact/govuk/problem_reports"
   id="something-is-wrong"
-  class="gem-c-feedback__form js-feedback-form js-hidden"
+  class="gem-c-feedback__form js-feedback-form"
   data-track-category="Onsite Feedback"
   data-track-action="GOV.UK Send Form"
-  method="post">
+  method="post"
+  hidden>
   <button
     class="govuk-button govuk-button--secondary gem-c-feedback__close gem-c-feedback__js-show js-close-form"
     data-track-category="Onsite Feedback"
@@ -15,26 +16,28 @@
 
   <div class="govuk-grid-row">
     <div class="govuk-grid-column-two-thirds">
-      <div class="gem-c-feedback__error-summary gem-c-feedback__js-show js-hidden js-errors" tabindex="-1"></div>
+      <div class="gem-c-feedback__error-summary gem-c-feedback__js-show js-errors" tabindex="-1" hidden></div>
 
       <input type="hidden" name="url" value="<%= url_without_pii %>">
 
       <h3 class="gem-c-feedback__form-heading"><%= t("components.feedback.help_us_improve_govuk") %></h3>
       <p id="feedback_explanation" class="gem-c-feedback__form-paragraph"><%= t("components.feedback.dont_include_personal_info") %></p>
 
-      <%= render "govuk_publishing_components/components/input", {
+      <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("components.feedback.what_doing")
         },
         name: "what_doing",
+        rows: 3,
         describedby: "feedback_explanation"
       } %>
 
-      <%= render "govuk_publishing_components/components/input", {
+      <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("components.feedback.what_wrong")
         },
-        name: "what_wrong"
+        name: "what_wrong",
+        rows: 3
       } %>
 
       <%= render "govuk_publishing_components/components/button", {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,9 @@
 
 ## Unreleased
 
+* **BREAKING:** Update feedback component to fix accessibility issues ([PR #2435](https://github.com/alphagov/govuk_publishing_components/pull/2435))
+  You must make the following changes when you migrate to this release:
+  - Upgrade `static` to the latest version before you upgrade your application
 * Remove duplicate link in menu header ([PR #2547](https://github.com/alphagov/govuk_publishing_components/pull/2547))
 
 ## 27.20.0
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -25,10 +25,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.setInitialAriaAttributes()
     this.setHiddenValues()
 
+    this.prompt.hidden = false
+    for (var k = 0; k < this.promptQuestions.length; k++) {
+      this.promptQuestions[k].hidden = false
+    }
+    this.surveyForm.hidden = true
+
     for (var j = 0; j < this.toggleForms.length; j++) {
       this.toggleForms[j].addEventListener('click', function (e) {
         e.preventDefault()
-        var el = e.target
+        var el = e.target.closest('button')
         this.toggleForm(el.getAttribute('aria-controls'))
         this.trackEvent(this.getTrackEventParams(el))
         this.updateAriaAttributes(el)
@@ -36,6 +42,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }
 
     for (var i = 0; i < this.closeForms.length; i++) {
+      this.closeForms[i].hidden = false
       this.closeForms[i].addEventListener('click', function (e) {
         e.preventDefault()
         var el = e.target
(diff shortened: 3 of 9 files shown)


app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
```diff
@@ -15,26 +16,28 @@
...
       <h3 class="gem-c-feedback__form-heading"><%= t("components.feedback.help_us_improve_govuk") %></h3>
       <p id="feedback_explanation" class="gem-c-feedback__form-paragraph"><%= t("components.feedback.dont_include_personal_info") %></p>
 
-      <%= render "govuk_publishing_components/components/input", {
+      <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("components.feedback.what_doing")
         },
         name: "what_doing",
+        rows: 2,
```
andysellick, line 31: If we're going to turn these inputs into textareas, is it worth making them more than just one extra line taller? I'm often frustrated by tiny textareas on websites, seems like if we're making it bigger we should make it properly bigger.

jon-kirwan: Yep agreed. That sounds good to me. Both components now updated to have `rows: 3`.

chris-gds:
For ref - placed in situ within other apps the behaviour is not as expected, investigating (red border is to confirm local version) 

https://user-images.githubusercontent.com/71266765/142614865-f512246a-953a-4ced-9bc5-b70c07a9db21.mov

https://user-images.githubusercontent.com/71266765/142614894-52b7b00e-6c99-4d7c-8334-dcc5fbce9e8c.mov

jon-kirwan:
> For ref - placed in situ within other apps the behaviour is not as expected, investigating (red border is to confirm local version)
> 
>  Screen.Recording.2021-11-19.at.11.10.26.mov 
>  Screen.Recording.2021-11-17.at.17.21.31.mov

I've now added some comments in PR description for this issue under **'Note (regards local testing and release)'**

alex-ju:
Excellent PR description, it made it easier to understand the issue and the proposed changes. As a suggestion, from a reviewer's perspective, a bit more granularity in terms of commits could be helpful, but it's still digestible. Code looks good to me and I appreciate the `[hidden] { display: none !important; }` could be something that sits at a higher level, but in the current component model we may have to sacrifice some code repetition in the benefit of component isolation (as in making sure the component doesn't rely on external styling).

An important step for this work is the rollout plan, as the component styles are imported twice (in `static` and at the application level), as you mentioned. Upgrading `static` first and the apps later means that we will serve the no-js experience to everyone for a short time. Updating the apps first and static afterward would create an issue where pressing 'Report a problem' would make the feedback component disappear, as Chris mentioned.

alex-ju (approved):

jon-kirwan:
> Excellent PR description, it made it easier to understand the issue and the proposed changes. As a suggestion, from a reviewer's perspective, a bit more granularity in terms of commits could be helpful, but it's still digestible. Code looks good to me and I appreciate the `[hidden] { display: none !important; }` could be something that sits at a higher level, but in the current component model we may have to sacrifice some code repetition in the benefit of component isolation (as in making sure the component doesn't rely on external styling).
> 
> An important step for this work is the rollout plan, as the component styles are imported twice (in `static` and at the application level), as you mentioned. Upgrading `static` first and the apps later means that we will serve the no-js experience to everyone for a short time. Updating the apps first and static afterward would create an issue where pressing 'Report a problem' would make the feedback component disappear, as Chris mentioned.

Thanks for the review, Alex 👍

Regards the release plan, I just want to double check that it's okay to provide a no-js experience, for a short period of time, whilst we get all the dependabot releases merged. In this scenario, the "Page is not useful" form is displayed and the component toggles are hidden (as if a user had clicked 'no' from the component). If that's okay, I think that sounds like a good plan.

I'll also need to create a separate PR to add the textarea component sass import to `static`

Initially, I'd been thinking of something [like this](https://trello.com/c/MOshcFm1/771-update-feedback-component-to-hide-and-show-content-in-a-more-robust-way) under **Activity** but this could be tricky to try and synchronise all these changes.

PR #2437, opened 2021-11-11
Add notification button to the metadata component

danacotoran:
Add the single page notification button to the version of the metadata
component that includes an update history (and therefore a "see all
updates" link).

The notification button requires a `base_path` therefore a `base_path` must
be passed to the metadata component wherever we want the button to
appear.

In the initial stages we do not want the button to appear everywhere
(i.e on every single page that uses a certain type of template or
publishing format) so there is an additional `include_notification_button`
flag to fine-tune where the button should appear.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add notification button to the metadata component ([PR #2437](https://github.com/alphagov/govuk_publishing_components/pull/2437))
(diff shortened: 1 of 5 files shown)


barrucadu (approved):
I'm not very well versed in rtl float issues, but the rest makes sense and looks good to me

andysellick (commented):
I'm not certain of the value of putting the notification button component inside another component in this situation. I can see how doing this will simplify rolling out this change across applications, but I think it adds unnecessary complexity to the component where the same outcome could be achieved by putting the notification button after the metadata component (some margin adjustments might be needed to fit with the design, admittedly).

PR #2443, opened 2021-11-11
Get single page notification button from personalisation API on load

danacotoran:
The single page notification button exists in two states: [default (unsubscribed)](https://components.publishing.service.gov.uk/component-guide/single_page_notification_button/default) and [already subscribed](https://components.publishing.service.gov.uk/component-guide/single_page_notification_button/already_subscribed).

If a user is authenticated and already subscribed to email notifications on the current page, the button should say "Stop getting emails about this page". To request the  user data for this layer of personalisation, there is a need to bypass the CDN, however it is not practical to serve all such requests from origin (especially on relatively high traffic pages such as the ones it is likely this feature will be added to).
As such the solution we've landed on is to use JS to perform this check on page load as an enhancement.

`account-api` serves the button HTML from the `/api/personalisation/check-email-subscription` endpoint – if the user is authenticated and already subscribed to the current page, the endpoint serves the "[Already subscribed](https://components.publishing.service.gov.uk/component-guide/single_page_notification_button/already_subscribed)" variant of the button. The button that already exists on the page is replaced with this updated version, so the user can see that they are already subscribed to emails about the current page.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
@@ -1,0 +1,48 @@
+/* global XMLHttpRequest */
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function SinglePageNotificationButton ($module) {
+    this.$module = $module
+    this.basePath = this.$module.querySelector('input[name="base_path"]').value
+    this.buttonLocation = this.$module.getAttribute('data-button-location')
+
+    this.personalisationEndpoint = '/api/personalisation/check-email-subscription?base_path=' + this.basePath
+    // This attribute is passed through to the personalisation API to ensure the updated button has the same button_location for analytics
+    if (this.buttonLocation) this.personalisationEndpoint += '&button_location=' + this.buttonLocation
+  }
+
+  SinglePageNotificationButton.prototype.init = function () {
+    var xhr = new XMLHttpRequest()
+    xhr.open('GET', this.personalisationEndpoint, true)
+
+    xhr.onreadystatechange = function () {
+      if (xhr.readyState === 4 && xhr.status === 200) {
+        var responseText = xhr.responseText
+        // if response text exists and is JSON parse-able, parse the response and get the button html
+        if (responseText && this.responseIsJSON(responseText)) {
+          var newButton = JSON.parse(responseText).button_html
+          var html = document.createElement('div')
+          html.innerHTML = newButton
+          // test that the html returned contains the button component; if yes, swap the button for the updated version
+          var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button .gem-c-single-page-notification-button__submit')
+          if (responseHasButton) {
+            this.$module.outerHTML = newButton
+          }
+        }
+      }
+    }.bind(this)
+    xhr.send()
+  }
+
+  SinglePageNotificationButton.prototype.responseIsJSON = function (string) {
+    try {
+      JSON.parse(string)
+    } catch (e) {
+      return false
+    }
+    return true
+  }
+  Modules.SinglePageNotificationButton = SinglePageNotificationButton
+})(window.GOVUK.Modules)
--- a/app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
+++ b/app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
@@ -1,23 +1,17 @@
 <%
-  page ||= ''
-  data_attributes ||= {}
-  base_path ||= nil
-  local_assigns[:margin_bottom] ||= 3
-  already_subscribed ||= false
-  text ||= already_subscribed ? t('components.single_page_notification_button.unsubscribe_text') : t('components.single_page_notification_button.subscribe_text')
-
+  component_helper = GovukPublishingComponents::Presenters::SinglePageNotificationButtonHelper.new(local_assigns)
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+
   wrapper_classes = %w(gem-c-single-page-notification-button govuk-!-display-none-print)
   wrapper_classes << shared_helper.get_margin_bottom
-  classes = "govuk-body-s gem-c-single-page-notification-button__submit"
 %>
 <% button_text = capture do %>
-  <svg class="gem-c-single-page-notification-button__icon" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="18" width="18" viewBox="0 0 459.334 459.334"><path fill="currentColor" d="M177.216 404.514c-.001.12-.009.239-.009.359 0 30.078 24.383 54.461 54.461 54.461s54.461-24.383 54.461-54.461c0-.12-.008-.239-.009-.359H175.216zM403.549 336.438l-49.015-72.002v-89.83c0-60.581-43.144-111.079-100.381-122.459V24.485C254.152 10.963 243.19 0 229.667 0s-24.485 10.963-24.485 24.485v27.663c-57.237 11.381-100.381 61.879-100.381 122.459v89.83l-49.015 72.002a24.76 24.76 0 0 0 20.468 38.693H383.08a24.761 24.761 0 0 0 20.469-38.694z"/></svg><%= text %>
+  <svg class="gem-c-single-page-notification-button__icon" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="18" width="18" viewBox="0 0 459.334 459.334"><path fill="currentColor" d="M177.216 404.514c-.001.12-.009.239-.009.359 0 30.078 24.383 54.461 54.461 54.461s54.461-24.383 54.461-54.461c0-.12-.008-.239-.009-.359H175.216zM403.549 336.438l-49.015-72.002v-89.83c0-60.581-43.144-111.079-100.381-122.459V24.485C254.152 10.963 243.19 0 229.667 0s-24.485 10.963-24.485 24.485v27.663c-57.237 11.381-100.381 61.879-100.381 122.459v89.83l-49.015 72.002a24.76 24.76 0 0 0 20.468 38.693H383.08a24.761 24.761 0 0 0 20.469-38.694z"/></svg><%= component_helper.button_text %>
 <% end %>
-<%= tag.form class: wrapper_classes, action: "/email/subscriptions/single-page/new", method: "POST", data: data_attributes do %>
-  <input type="hidden" name="base_path" value="<%= base_path %>">
+<%= tag.form class: wrapper_classes, action: "/email/subscriptions/single-page/new", method: "POST", data: component_helper.data do %>
+  <input type="hidden" name="base_path" value="<%= component_helper.base_path %>">
   <%= content_tag(:button, button_text, {
-    class: classes,
+    class: "govuk-body-s gem-c-single-page-notification-button__submit",
     type: "submit",
   }) %>
-<% end if base_path.presence %>
+<% end if component_helper.base_path %>
--- a/lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb
@@ -1,0 +1,40 @@
+module GovukPublishingComponents
+  module Presenters
+    class SinglePageNotificationButtonHelper
+      attr_reader :already_subscribed, :data_attributes, :base_path, :js_enhancement, :button_type, :button_location
+
+      def initialize(local_assigns)
+        @local_assigns = local_assigns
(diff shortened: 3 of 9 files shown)


app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -0,0 +1,43 @@
+/* global XMLHttpRequest */
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function SinglePageNotificationButton ($module) {
+    this.$module = $module
+    this.personalisationEndpoint = this.$module.getAttribute('data-personalisation-endpoint')
```
barrucadu, line 8: Why is it necessary to fetch the endpoint from a data attribute?   I see in the erb file you're generating this with the website root present, but doesn't it work when you use only the absolute path?

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -0,0 +1,43 @@
...
+          html.innerHTML = newButton
+          // test that the html returned contains the button component; if yes, swap the button for the updated version
+          var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button')
+          if (responseHasButton) {
+            this.$module.outerHTML = newButton
+          }
+        }
+      }
+    }.bind(this)
+    xhr.send()
```
andysellick, line 31: Trying to get my head around this... does this script need to take account of user cookie preferences? I can't remember how that works with regard to a user being signed in.

I think the thing I'm concerned about is that this script does an XMLHttpRequest by default, when most users (at least to start with) won't be signed into an account. Is there any way to make a check for this first before firing a network request?

spec/javascripts/components/single-page-notification-button-spec.js
```diff
@@ -0,0 +1,88 @@
+/* eslint-env jasmine, jquery */
```
andysellick, line 1: Nitpick: is `jquery` needed here? I can't see any in this file.

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
barrucadu: The account session cookie is an `HTTPOnly` cookie, so it's not accessible to javascript (but is still sent with `XMLHttpRequest`).  So we can't check for its presence here.

However, we *could* serve two versions of the page, with slightly different markup: one for users with a cookie and one for users without a cookie.  I think that'll work, so long as we set the right cache-control response header.

app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
```diff
@@ -10,6 +10,9 @@
   wrapper_classes = %w(gem-c-single-page-notification-button govuk-!-display-none-print)
   wrapper_classes << shared_helper.get_margin_bottom
   classes = "govuk-body-s gem-c-single-page-notification-button__submit"
+  
+  data_attributes[:module] = "single-page-notification-button"
```
barrucadu, line 14: Can we make this overridable by an app?  We can serve two versions of a page: one for users with a cookie, one for users without a cookie.  Only the version for users with a cookie needs to run the javascript.

spec/javascripts/components/single-page-notification-button-spec.js
```diff
@@ -0,0 +1,88 @@
...
+        '<input type="hidden" name="base_path" value="/current-page-path">' +
+        '<button class="govuk-body-s gem-c-single-page-notification-button__submit" type="submit">  <svg class="gem-c-single-page-notification-button__icon" aria-hidden="true" xmlns="http://www.w3.org/2000/svg" height="18" width="18" viewBox="0 0 459.334 459.334"><path fill="currentColor" d="M177.216 404.514c-.001.12-.009.239-.009.359 0 30.078 24.383 54.461 54.461 54.461s54.461-24.383 54.461-54.461c0-.12-.008-.239-.009-.359H175.216zM403.549 336.438l-49.015-72.002v-89.83c0-60.581-43.144-111.079-100.381-122.459V24.485C254.152 10.963 243.19 0 229.667 0s-24.485 10.963-24.485 24.485v27.663c-57.237 11.381-100.381 61.879-100.381 122.459v89.83l-49.015 72.002a24.76 24.76 0 0 0 20.468 38.693H383.08a24.761 24.761 0 0 0 20.469-38.694z"></path></svg>Get emails about this page</button>' +
+    '</form>'
```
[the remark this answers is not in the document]
danacotoran: Yeah, to be honest I literally copied the response from the real endpoint as I thought it would make most accurate/realistic 😄   I'll simplify it like you suggested though, you're right it doesn't have to be exactly the same.

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
danacotoran: @andysellick Michael suggested to make the JS stuff opt-in which I think is a great idea and shouldn't be a lot more additional work. Bonus: it will also appease this visual regression test which is complaining that `/api/personalisation/check-email-subscription` does not exist in the component guide https://github.com/alphagov/govuk_publishing_components/runs/4189636063?check_suite_focus=true#step:6:55

spec/javascripts/components/single-page-notification-button-spec.js
danacotoran: Nope, I let this in from earlier iterations of this file, good spot

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
andysellick: By 'opt-in' do you mean it'll only run if the user has accepted cookies?

spec/dummy/config/routes.rb
```diff
@@ -19,4 +19,5 @@
   get "error-summary", to: "welcome#errorsummary"
   get "tabsexample", to: "welcome#tabsexample"
   get "table", to: "welcome#table"
+  get "/api/personalisation/check-email-subscription", to: proc { [404, {}, [""]] }
```
andysellick, line 22: Probably worth putting a comment above this line to explain why it's here and which component it refers to, something like

```
we fake this URL to prevent the Single Page notification button from causing an error in the component guide
```

danacotoran: good point

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
danacotoran: > By 'opt-in' do you mean it'll only run if the user has accepted cookies?

@andysellick  What I meant about 'opt-in' here is get the component JS to run only when a user is logged in (as Michael's detailed above).
You brought up a good point about cookies however. I was not sure if the functionality should be present only when someone has accepted cookies or not. I brought up this question with team and it was concluded that the answer is no, because the only cookie this functionality is dependent on is the account session cookie, which is considered an essential one and thus does not require explicit consent.

alex9smith (approved):

spec/components/single_page_notification_button_spec.rb
```diff
@@ -40,4 +40,53 @@ def component_name
     render_component({ base_path: "/the-current-page", margin_bottom: 9 })
     assert_select '.gem-c-single-page-notification-button.govuk-\!-margin-bottom-9'
   end
+
+  it "has a data-module attribute for JavaScript, if the js-enhancement flag is present" do
+    render_component({ base_path: "/the-current-page", js_enhancement: true })
+    dom = Nokogiri::HTML(rendered)
+    form_data_module = dom.xpath("//form")[0].attr("data-module")
```
andysellick, line 47: Curious why Nokogiri is needed here - is it not possible to do something like `assert_select ".gem-c-single-page-notification-button[data-module="single-page-notification-button gem-track-click"]`?

spec/javascripts/components/single-page-notification-button-spec.js
```diff
@@ -0,0 +1,102 @@
...
+    initButton()
+
+    jasmine.Ajax.requests.mostRecent().respondWith({
+      status: 200,
+      contentType: 'application/json',
+      responseText: '{\n    "base_path": "/current-page-path",\n    "active": true,\n    "button_html": "<form class=\\"gem-c-single-page-notification-button new-button-for-test\\" action=\\"/email/subscriptions/single-page/new\\" method=\\"POST\\">\\n  <input type=\\"hidden\\" name=\\"base_path\\" value=\\"/current-page-path\\">\\n  <button class=\\"gem-c-single-page-notification-button__submit\\" type=\\"submit\\">Stop getting emails about this page\\n</button>\\n</form>"\n}'
+    })
+
+    var form = document.querySelector('form.gem-c-single-page-notification-button.new-button-for-test')
+    expect(form).toHaveText('Stop getting emails about this page')
```
andysellick, line 52: I wonder if this test needs to be more rigorous - should it check that the form contains a button and is therefore functional?

danacotoran: sure, that is a fair point. The form doesn't do anything without a button after all. Will update

spec/components/single_page_notification_button_spec.rb
danacotoran: Yep it is possible to have `assert_select ".gem-c-single-page-notification-button[data-module="gem-track-click single-page-notification-button"]` and the test will pass. But just because something is possible, doesn't mean it's the best/most accurate way.

Change the order of the `data-module`s, or add another `data-module` to the component and this test will fail. IMO the test should be strictly scoped to the relevant thing (which in this case is `data-module` value of `single-page-notification-button`). But I can be convinced to change it if there's some major disadvantage to using the Nokogiri approach? 🤷🏻‍♀️ 
Admittedly I am no expert in these things.

lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+  module Presenters
+    class SinglePageNotificationButtonHelper
+      attr_reader :already_subscribed, :data_attributes, :base_path, :js_enhancement, :button_type, :button_location
+
+      def initialize(local_assigns)
+        @local_assigns = local_assigns
+        @data_attributes = @local_assigns[:data_attributes] || {}
+        @js_enhancement = @local_assigns[:js_enhancement] || false
+        @already_subscribed = @local_assigns[:already_subscribed] || false
+        @base_path = @local_assigns[:base_path] || nil
```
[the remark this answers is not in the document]
danacotoran: Some of the options are expected to be booleans, others we expect will be some other type of object (e.g strings). 
I had the boolean ones default to `false`, expected string ones default to `nil` not sure if it's a convention or something I made up though. 😄
We could have them all use `nil` and it would not make a difference, it just wouldn't _feel right_.

lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb
```diff
@@ -0,0 +1,40 @@
...
+        @data_attributes[:action] = [button_type, "button", button_location].compact.join("-")
+        @data_attributes[:module] = module_names.join(" ")
+        @data_attributes[:category] = "Single-page-notification-button"
+        # This attribute is passed through to the personalisation API to ensure when a new button is returned from the API, it has the same button_location
+        @data_attributes[:button_location] = button_location
+        @data_attributes
+      end
+
+      def button_location_is_valid?
+        @local_assigns[:button_location] == "bottom" || @local_assigns[:button_location] == "top"
```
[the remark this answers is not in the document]
danacotoran: Good shout, have updated to `%w[bottom top].include? @local_assigns[:button_location]`

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -0,0 +1,48 @@
...
+    xhr.onreadystatechange = function () {
+      if (xhr.readyState === 4 && xhr.status === 200) {
+        var responseText = xhr.responseText
+        // if response text exists and is JSON parse-able, parse the response and get the button html
+        if (responseText && this.responseIsJSON(responseText)) {
+          var newButton = JSON.parse(responseText).button_html
+          var html = document.createElement('div')
+          html.innerHTML = newButton
+          // test that the html returned contains the button component; if yes, swap the button for the updated version
+          var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button')
```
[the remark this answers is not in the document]
danacotoran: I tried to account for most of the things that could go wrong when you fetch html from some endpoint and replace stuff on your page with that response. One of those possibilities to account for is that the server (for whatever reason) doesn't return the HTML structure you expect. Though this isn't completely foolproof of course.

spec/components/single_page_notification_button_spec.rb
andysellick: No, that's a good answer, thanks.

andysellick (approved):

PR #2444, opened 2021-11-11
Fix layout glitch seen when scrollbar is permanently visible

### LINKED ISSUE #2430: Bizarre behaviour on super navigation header when switching from tablet to desktop on safari
When on safari and swapping from tablet to desktop views, a bizarre behaviour is exhibited by the super navigation header:

https://user-images.githubusercontent.com/64783893/141271839-791dcc40-7f2e-4d6e-81cc-933544c9cd84.mov

Summary of screen recording:

- When moving from tablet to desktop very slowly, the mobile menu is briefly visible. This occurs approx between 750 and 770 px width
- When the client width increases further, the search button will overflow below the topic and government activity buttons
- The search icon returns to it's original position after another switch but done quickly, so that the mobile menu isn't visible as in the first bullet point

injms:
## What

Fixes #2430 - the layout would become disjointed as width that triggered CSS changes was different to the width that triggered JavaScript changes _only if_ the scrollbar was permanently visible. This was due to the use of `window.innerWidth` which reports the width of the viewport including the width of the scrollbar. Using `document.documentElement.clientWidth` reports the width of the viewport sans any bars.

## Why

Because layout bugs are not good to have.

### Before 

https://user-images.githubusercontent.com/1732331/141364868-c5008dbd-112f-4d65-baed-b7f9a1da83ce.mov

### After

https://user-images.githubusercontent.com/1732331/141364778-3fd75cee-e1b3-4d59-bd35-699a14554127.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add an explicit margin zero to the super navigation mobile menu button ([PR #2445](https://github.com/alphagov/govuk_publishing_components/pull/2445))
 * Remove brexit as a topic from the super navigation header ([PR #2446](https://github.com/alphagov/govuk_publishing_components/pull/2446))
 * Update the pseudo underline mixin and it's usage on the super navigation header ([PR #2439](https://github.com/alphagov/govuk_publishing_components/pull/2439))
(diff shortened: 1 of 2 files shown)


owenatgov (approved):
Works like a charm, cheers for fixing this!

Issue #598, opened 2017-05-16
Govspeak Summary is deprecated


alextea:
The [govspeak content guidance](http://govspeak-guide.herokuapp.com/#answer-summaries) says the summary formatting has been deprecated. 

The [related styles](https://github.com/alphagov/static/blob/dade167e53dbc3bdbb55aadd27b05c94ccdd74c5/app/assets/stylesheets/govuk-component/govspeak/_summary.scss) should be removed.

There's also an [issue](https://github.com/alphagov/govspeak/issues/109) to remove the code that creates the markup in the govspeak repo.

owenatgov:
It looks like this has been removed from govspeak and the work is to remove this from the gem. However it'd be useful to crawl the content store and see if the actually summary classes exist anywhere. Besides this, work should be quite simple.

owenatgov:
Following a query to the publishing API, this is stil being used on a number of pages:
- https://gov.uk/government/publications/approval-of-operating-plants-handling-or-processing-meat
- https://gov.uk/government/news/government-launches-information-sharing-partnership-on-cyber-security
- https://gov.uk/government/news/new-initiative-to-drive-down-costs-in-government-back-offices
- https://gov.uk/government/news/new-energy-deal-to-save-84m-from-government-bill
- https://gov.uk/government/news/prime-ministers-message-on-welfare-reform
- https://gov.uk/government/news/general-aviation-red-tape-challenge
- https://gov.uk/government/news/civil-servants-of-the-future-invited-to-apply-to-new-fast-track-apprenticeship-scheme
- https://gov.uk/government/news/more-government-contract-opportunities-published
- https://gov.uk/government/news/top-tips-to-help-small-organisations-win-government-contracts
- https://gov.uk/government/speeches/cyber-security-information-sharing-programme

Most of these are very old, circa 2013, but the most recent one is from this year which is troubling. Looking at these pages, removing the surrounding `div` and the styling so it's just a `p` tag with standard govspeak styling is going to have a minuscule effect on the layout of these pages, essentially a reduction in margin. My temptation is to move ahead with retiring this anyway. Interested in opinions.

owenatgov:
From @alex-ju on my above comment:
>Thanks for looking into these, Owen – now that we know where they’re being used it will make it easier to remove them – agree with you on the govspeak summary, I would move ahead as well

I will therefore go ahead with removing this styling from this repo.

[...]

PR #2447, opened 2021-11-12
Remove summary styling from govspeak

owenatgov:
## What
Removes the styling for the govspeak summary extension as well as the associated docs

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/598

To summarise (🥁 ):

- The summary extension has been retired so we shouldn't include this in the gem
- This styling has no impact on the user in practice compared to if the markup was just presented to the user without the wrapping div

No visual changes


jon-kirwan (approved):
💥 all good (just the CHANGELOG conflict to be resolved here too)

PR #2449, opened 2021-11-15
Don't show Brexit side nav on hub pages

sihugh:
## What

We removed the direct link from hub pages to the checker in https://github.com/alphagov/govuk_publishing_components/pull/2432 but we now want to remove the Brexit sidebar nav completely from these two pages ([businesses](https://www.gov.uk/guidance/brexit-guidance-for-businesses) and [individuals](https://www.gov.uk/guidance/brexit-guidance-for-individuals-and-families)) to stop circular journeys.

We can remove the logic around "brexit_start_page_content_id" because that page hasn't existed for a while.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


hannako (approved):
👍

PR #2454, opened 2021-11-17
Add organisation colour for Department for Levelling Up, Housing and Communities (DLUHC)

brucebolt:
DLUHC = Department for Levelling Up, Housing and Communities

## What
Adds the brand colour for DLUHC.

The colour has a contrast ratio of 14.75, so not adding a websafe colour for this organisation.

## Why
The same change has been made in [govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/2434), however we need to also make the change here until the next release of govuk-frontend.

[Zendesk ticket](https://govuk.zendesk.com/agent/tickets/4712566)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add brand colour for Department for Levelling Up, Housing and Communities (DLUHC) ([PR #2454](https://github.com/alphagov/govuk_publishing_components/pull/2454))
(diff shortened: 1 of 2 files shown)


injms (approved):
Looks to be working as expected:
Screenshot 2021-11-17 at 11 24 13

I'll add an issue so we don't forget to remove this when this change is made in GOV.UK Frontend.

Thanks for adding this Bruce!

brucebolt:
> I'll add an issue so we don't forget to remove this when this change is made in GOV.UK Frontend.
Brilliant, thank you.

PR #2459, opened 2021-11-22
Remove redundant API value from big number component

owenatgov:
## What
Removes a stray `href` from the big number API presets

## Why
This is a repeated value and is therefore redundant!

No visual changes.


injms (approved):
Nice catch!

PR #2462, opened 2021-11-22
Update search component

injms:
## What

 - Updates search component documentation to mention `margin_bottom` parameter - only the documentation was added here, not the functionality.
 - Adds the ability to change the spacing between the label and the input
 - Allows the search label to be wrapped in a heading - similar to [the `isPageHeading` setting in GOV.UK Frontend's Nunjucks macro.][1] This is best used in conjunction with `label_size`.

## Why

The documentation for the `margin_bottom` attribute was missing, so needed to be added.

The spacing between the label and the input needed to be more flexible to allow for the updated homepage layout.

Having the label wrapped as a heading will allow for a the structure of the page to make semantic sense without needing unnecessary repetition - for example, a heading saying "Search" followed by a label that says "Search GOV.UK" could be merged into one heading-label that says "Search GOV.UK".

## Visual Changes

Example with `label_margin_bottom` set to 9:

image

There are no visual changes when using `wrap_label_in_a_heading` - this will wrap the label in a heading element that is unstyled with the margin removed.

[1]: https://design-system.service.gov.uk/get-started/labels-legends-headings/#labels-as-page-headings

### DIFF
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -4,6 +4,7 @@
   aria_controls ||= nil
   button_text ||= t("components.search_box.search_button")
   id ||= "search-main-" + SecureRandom.hex(4)
+  label_margin_bottom ||= nil
   label_size ||= nil
   label_text ||= t("components.search_box.label")
   name ||= "q"
@@ -13,13 +14,6 @@
 
   data_attributes ||= {}
   data_attributes[:module] = 'gem-track-click'
-
-  label_classes = []
-  if (shared_helper.valid_heading_size?(label_size))
-    label_classes << "govuk-label govuk-label--#{label_size}"
-  else
-    label_classes << "gem-c-search__label"
-  end
 
   classes = %w[gem-c-search govuk-!-display-none-print]
   classes << (shared_helper.get_margin_top)
@@ -32,6 +26,14 @@
     classes << "gem-c-search--on-white"
   end
   classes << "gem-c-search--separate-label" if local_assigns.include?(:inline_label) or local_assigns.include?(:label_size)
+
+  label_classes = []
+  if (shared_helper.valid_heading_size?(label_size))
+    label_classes << "govuk-label govuk-label--#{label_size}"
+  else
+    label_classes << "gem-c-search__label"
+  end
+  label_classes << "govuk-!-margin-bottom-#{label_margin_bottom}" if [*1..9].include?(label_margin_bottom) and local_assigns.include?(:inline_label)
 %>
 
 <div class="<%= classes.join(" ") %>" data-module="gem-toggle-input-class-on-focus">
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -32,11 +28,31 @@
     classes << "gem-c-search--on-white"
   end
   classes << "gem-c-search--separate-label" if local_assigns.include?(:inline_label) or local_assigns.include?(:label_size)
```
owenatgov, line 30: A teeny tiny nitpick: you've mentioned why it's a bad idea to specify the `inline_label` using the `||=` convention at the top (because it's falsey, ruby will defer to the default value). Do you think its worth adding a comment to this effect? I'm thinking of a well-meaning dev seeing that it's not included at the top and adding it in, potentially breaking a few things.

I'm asking instead of suggesting partly because the tests would catch this.

spec/components/search_spec.rb
```diff
@@ -8,6 +8,7 @@ def component_name
   it "renders a search box with default options" do
     render_component({})
     assert_select ".gem-c-search.gem-c-search--on-white"
+    assert_select "label[class='^=govuk-\!-margin-bottom-']", count: 0
```
owenatgov, line 11: These "starts-with" attribute selectors aren't formatted quite right. They're passing because they're looking for nothing, but they're dud selectors so will always return nothing, thus making the tests redundant.

```suggestion
    assert_select "label[class^='govuk-\!-margin-bottom-']", count: 0
```

As a bit of extra evidence I did some local testing on this with a very rough test. This passed:

```
it "test" do
    render_component({})

    assert_select "div[class^='gem-c-']"
  end
```

This didn't:

```
it "test" do
    render_component({})

    assert_select "div[class='^gem-c-']"
  end
```

This applies also to lines 150 and 157.

owenatgov (commented):
A comment and a question. Very nice work here, some lovely tests.

owenatgov:
Approved the percy build 👍

spec/components/search_spec.rb
injms: Good catch 🕵️‍♂️ 

This has been fixed for the search component in https://github.com/alphagov/govuk_publishing_components/pull/2462/commits/ba38459d46a9e48af46a603db64623174689dacc.

And there were other instances of this error which I've fixed in https://github.com/alphagov/govuk_publishing_components/pull/2463.

owenatgov (commented):
I should've mentioned this in my original review but this just needs a changelog update and I'll approve it :)

injms:
@owenatgov change log updated - ready for review

owenatgov (approved):
🔍

app/views/govuk_publishing_components/components/_search.html.erb
injms: Ooh, this is a can of worms. I agree, it's a thing that might be overridden without someone meaning to - so we could suggest moving away from using `false ` values as the default.

For this component, I'd also suggest that renaming the `inline_label` to `label_position` of which the options are `inline` or `above` (or something similar, just throwing ideas out here).  Then we can set the default to be `inline` using the `||=`, run the two parameters in parallel while we update the component usage, and then deprecate it. I think there are a couple of breaking changes coming up, which is good timing.

I'll raise an issue to gently nudge a discussion around this and see what the community consensus is.

injms: Issue raised: https://github.com/alphagov/govuk_publishing_components/issues/2472

PR #2463, opened 2021-11-23
Fix selectors in spec

injms:
## What

All the uses of `[class='^=class-name']` have been replaced with the correct formatting: `[class^='class-name']`.

## Why

As [highlighted by @owenatgov in another review][1], there were selectors in the spec that weren't formatted correctly - so they were potentially able to giving false negatives and needed fixing.

## Visual Changes
No visual changes.

[1]: https://github.com/alphagov/govuk_publishing_components/pull/2462#discussion_r755090045


owenatgov (approved):
Legend 👍

PR #2464, opened 2021-11-23
Amend `explicit-cross-domain-links.js` code

danacotoran:
Amend `explicit-cross-domain-links` code to account for the possibility of someone accepting/rejecting cookies on the page that the cross-domain link is present. 

This was raised as a bug on the Brexit page, where the **Sign in** link is a cross-domain link (leading to DI Sign In territory). 
If someone has interacted with the cookie banner on this page, the link treats them as if they have not, i.e the `?cookie_consent=not_engaged` parameter remains unchanged, and the GA parameter is not added (if they have accepted cookies). The setting applies only after the page is reloaded.

## Before (the correct parameters are appended to the cross domain link only after the page is reloaded)
https://user-images.githubusercontent.com/7116819/143051955-22e6eb6a-f22f-48b1-b2e1-e6e59df52665.mov

## After (the correct parameters are appended to the cross domain link instantly when cookies are accepted or rejected)

https://user-images.githubusercontent.com/7116819/143052834-1f00cbc2-b2fe-41a6-8948-89e97a4e5b67.mov

Upon detecting the cookie consent or rejection event, the cross-domain link module should re-run and append the correct parameters to any cross domain links present on the page.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
@@ -66,7 +66,7 @@ describe('Explicit cross-domain linker', function () {
       expect(element.attr('href')).toEqual('/somewhere?cookie_consent=reject')
     })
 
-    describe('user accepts cookies', function () {
+    describe('user has accepted cookies', function () {
       it('modifies the link href to append cookie_consent parameter "accept" and adds _ga if trackers are present', function () {
         GOVUK.cookie('cookies_preferences_set', 'true')
         GOVUK.setConsentCookie({ usage: true })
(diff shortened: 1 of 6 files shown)


spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
```diff
@@ -93,6 +93,31 @@ describe('Explicit cross-domain linker', function () {
...
+        window.GOVUK.triggerEvent(window, 'cookie-consent')
+
+        expect(element.attr('href')).toEqual('/somewhere?cookie_consent=accept')
+      })
+
+      it('modifies the link href to append cookie_consent parameter "reject" if the cookie-reject event was fired', function () {
+        GOVUK.cookie('cookies_preferences_set', null)
+        explicitCrossDomainLinks.start(element)
+
+        GOVUK.cookie('cookies_preferences_set', 'true')
```
andysellick, line 114: These three lines appear to be duplicated in the previous test - could they be moved into a `beforeEach` function for this `describe` block?

andysellick (approved):
Overall looks good, one minor comment.

The code for `explicit-cross-domain-links` was already a bit of a spaghetti before these changes, I think once this change has gone in we should look at refactoring the code a little bit to make it clearer. Specifically there are a lot of guard clauses in the `start` function that maybe could be consolidated, and the whole thing could be restructured to make it more understandable. Something like:

```
window.addEventListener('cookie-consent', this.cookiesAreAccepted)
window.addEventListener('cookie-reject', this.cookiesAreRejected)

...

this.cookiesAreNotAnything = function () {
  this.decorate(element, 'cookie_consent=not-engaged')
}

this.cookiesAreAccepted = function () {
  // something here
}

this.cookiesAreRejected = function () {
  this.decorate(element, 'cookie_consent=reject')
}
```

spec/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.spec.js
danacotoran: Great point, have updated as suggested.

PR #2465, opened 2021-11-23
Add support to open Intervention links in new tab

chao-xian:
## What
Adds a parameter to the Intervention banner to allow links to open in a new tab.

## Why
In order to drive users to another online task, we don't want to completely hijack their original flow on GOV.UK. Adding the option to open links in a new tab helps to address this.

## Visual Changes

localhost_3212_component-guide_intervention_hide-intervention=true(iPhone 6_7_8)

Screenshot 2021-11-23 at 6 38 32 pm

### DIFF
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -35,7 +35,7 @@
     suggestion_link_text << accessible_text unless suggestion_link_text.include?(accessible_text)
   end
 %>
-<% if suggestion_text || (suggestion_link_text && suggestion_link_text) %>
+<% if suggestion_text || (suggestion_link_text && suggestion_link_url) %>
   <%= tag.section class: "gem-c-intervention", role: "region", aria: aria_attributes, data: data_attributes do %>
     <p class="govuk-body">
       <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -32,8 +32,7 @@
 
     suggestion_tag_options.merge!({ target: target, rel: rel })
 
-    accessible_text = I18n.t("components.intervention.accessible_link_text_suffix")
-    suggestion_link_text << accessible_text unless suggestion_link_text.include?(accessible_text)
+    suggestion_link_text = intervention_helper.accessible_text
   end
 %>
 <% if suggestion_text || (suggestion_link_text && suggestion_link_url) %>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -16,14 +16,28 @@
 
   intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(local_assigns)
   dismiss_href = intervention_helper.dismiss_link
+
+  suggestion_tag_options = {
+    class: "govuk-link gem-c-intervention__suggestion-link",
+    href: suggestion_link_url,
+    data: suggestion_data_attributes,
+  }
+
+  new_tab ||= false
+  if new_tab
+    target = "_blank"
+    rel = "noopener noreferrer"
+    rel << " external" unless suggestion_link_url && suggestion_link_url.start_with?("/", "https://gov.uk", "https://www.gov.uk")
+
+    suggestion_tag_options.merge!({ target: target, rel: rel })
+  end
 %>
 <% if suggestion_text or (suggestion_link_text and suggestion_link_text) %>
   <%= tag.section class: "gem-c-intervention", role: "region", aria: aria_attributes, data: data_attributes do %>
     <p class="govuk-body">
       <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
       <% if suggestion_link_text and suggestion_link_url %>
-        <%= tag.a suggestion_link_text, class: "govuk-link gem-c-intervention__suggestion-link", href: suggestion_link_url, data: suggestion_data_attributes %>
-
+        <%= tag.a suggestion_link_text, suggestion_tag_options %>
       <% end %>
     </p>
 
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -24,12 +24,15 @@
   }
 
   new_tab ||= false
-  if new_tab
+  if new_tab && (suggestion_link_text && suggestion_link_text)
     target = "_blank"
     rel = "noopener noreferrer"
     rel << " external" unless suggestion_link_url && suggestion_link_url.start_with?("/", "https://gov.uk", "https://www.gov.uk")
 
     suggestion_tag_options.merge!({ target: target, rel: rel })
+
+    accessible_text = I18n.t("components.intervention.accessible_link_text_suffix")
+    suggestion_link_text << accessible_text unless suggestion_link_text.include?(accessible_text)
   end
 %>
 <% if suggestion_text || (suggestion_link_text && suggestion_link_text) %>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -32,11 +32,11 @@
     suggestion_tag_options.merge!({ target: target, rel: rel })
   end
 %>
-<% if suggestion_text or (suggestion_link_text and suggestion_link_text) %>
+<% if suggestion_text || (suggestion_link_text && suggestion_link_text) %>
   <%= tag.section class: "gem-c-intervention", role: "region", aria: aria_attributes, data: data_attributes do %>
     <p class="govuk-body">
       <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
-      <% if suggestion_link_text and suggestion_link_url %>
+      <% if suggestion_link_text && suggestion_link_url %>
         <%= tag.a suggestion_link_text, suggestion_tag_options %>
       <% end %>
     </p>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -3,6 +3,7 @@
   suggestion_link_text ||= false
   suggestion_link_url ||= false
   suggestion_text ||= nil
+  new_tab ||= false
 
   data_attributes ||= {}
   data_attributes[:module] = 'intervention'
@@ -13,6 +14,8 @@
   aria_attributes[:label] = 'Intervention'
 
   local_assigns[:query_string] ||= request.query_string
+  local_assigns[:suggestion_link_text] = suggestion_link_text
+  local_assigns[:suggestion_link_url] = suggestion_link_url
 
   intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(local_assigns)
   dismiss_href = intervention_helper.dismiss_link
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -22,7 +26,7 @@
     <p class="govuk-body">
       <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
       <% if suggestion_link_text and suggestion_link_url %>
-        <%= tag.a suggestion_link_text, class: "govuk-link gem-c-intervention__suggestion-link", href: suggestion_link_url, data: suggestion_data_attributes %>
+        <%= tag.a suggestion_link_text, class: "govuk-link gem-c-intervention__suggestion-link", href: suggestion_link_url, data: suggestion_data_attributes, target: target %>
```
owenatgov, line 29: Could we add a `rel` attribute here to account for [security risks that come with opening links in new tabs](https://www.jitbit.com/alexblog/256-targetblank---the-most-underestimated-vulnerability-ever/)?

```suggestion
        <%= tag.a suggestion_link_text, class: "govuk-link gem-c-intervention__suggestion-link", href: suggestion_link_url, data: suggestion_data_attributes, target: target, rel: "noopener noreferrer" %>
```

owenatgov: We'd also want to add `external` to that rel attribute if the new tab was indeed an external site. I'm not sure how best to check for that though...

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -38,6 +38,16 @@ examples:
       suggestion_link_text: "You can now apply for a permit online."
       suggestion_link_url: "/permit"
 
+  open_suggestion_link_in_new_tab:
+    description: |
+      For example if you wanted to drive users to another online task, you don't want to completely hijack
+      their original flow on GOV.UK. Adding the option to open links in a new tab helps to address this.
+      Link text *must* tell the user that the link opens in a new tab.
```
owenatgov, line 45: I wonder if we can add that "(opens in a new tab)" text automatically if `new_tab` is specified.

owenatgov (commented):
A few comments here. Additionally I reckon it'd be worth adding a new test for this attribute in the spec.

owenatgov:
Percy build approved 👍

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: Love this suggestion.

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: 👍

chao-xian: [Done](https://github.com/alphagov/govuk_publishing_components/pull/2465/commits/5b436da4299f4c0026fdf7e19cafdf85a0885a2b)

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: [Done](https://github.com/alphagov/govuk_publishing_components/pull/2465/commits/9f13ad48ccf450c3b81df321f289c19d0af44128)

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -16,14 +16,31 @@
 
   intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(local_assigns)
   dismiss_href = intervention_helper.dismiss_link
+
+  suggestion_tag_options = {
+    class: "govuk-link gem-c-intervention__suggestion-link",
```
andysellick, line 21: Is the `gem-c-intervention__suggestion-link` class necessary? It doesn't seem to add any styles.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -16,14 +16,31 @@
...
+
+  new_tab ||= false
+  if new_tab && (suggestion_link_text && suggestion_link_text)
+    target = "_blank"
+    rel = "noopener noreferrer"
+    rel << " external" unless suggestion_link_url && suggestion_link_url.start_with?("/", "https://gov.uk", "https://www.gov.uk")
+
+    suggestion_tag_options.merge!({ target: target, rel: rel })
+
+    accessible_text = " (opens in a new tab)"
```
andysellick, line 34: This text should ideally come from a translation file, `en.yml`.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -16,14 +16,31 @@
...
   dismiss_href = intervention_helper.dismiss_link
+
+  suggestion_tag_options = {
+    class: "govuk-link gem-c-intervention__suggestion-link",
+    href: suggestion_link_url,
+    data: suggestion_data_attributes,
+  }
+
+  new_tab ||= false
+  if new_tab && (suggestion_link_text && suggestion_link_text)
```
andysellick, line 27: I'd suggest moving most of this code into the intervention component helper.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -16,14 +16,31 @@
...
+    rel << " external" unless suggestion_link_url && suggestion_link_url.start_with?("/", "https://gov.uk", "https://www.gov.uk")
+
+    suggestion_tag_options.merge!({ target: target, rel: rel })
+
+    accessible_text = " (opens in a new tab)"
+    suggestion_link_text << accessible_text unless suggestion_link_text.end_with?(accessible_text)
+  end
 %>
-<% if suggestion_text or (suggestion_link_text and suggestion_link_text) %>
+<% if suggestion_text || (suggestion_link_text && suggestion_link_text) %>
```
andysellick, line 38: This isn't your change, but the part of this statement after `&&` looks redundant - isn't it the same as the below?

`<% if suggestion_text || suggestion_link_text %>`

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -38,6 +38,17 @@ examples:
...
       suggestion_link_url: "/permit"
 
+  open_suggestion_link_in_new_tab:
+    description: |
+      For example if you wanted to drive users to another online task, you don't want to completely hijack
+      their original flow on GOV.UK. Adding the option to open links in a new tab helps to address this.
+      Link text should tell the user that the link opens in a new tab.
+      Note: "(opens in a new tab)" is added to link text if the phrase isn't included.
+    data:
+      suggestion_link_text: "You can now apply for a permit online (opens in a new tab)"
```
andysellick, line 48: Maybe change this line to just `"You can now apply for a permit online"` to demonstrate the auto text appending?

spec/components/intervention_spec.rb
```diff
@@ -63,4 +63,65 @@ def component_name
...
+      render_component(
+        suggestion_link_text: "Travel abroad",
+        suggestion_link_url: "/travel-abroad",
+        new_tab: true,
+      )
+
+      assert_select "a[target='_blank']"
+    end
+
+    it "renders with security attributes" do
```
andysellick, line 78: This test appears to do the same steps as the previous test but check for something different. Would it be worth combining the two? i.e. `it renders with target _blank and appropriate security attributes when new_tab is specified`

spec/components/intervention_spec.rb
```diff
@@ -63,4 +63,65 @@ def component_name
...
+    it "renders no target attribute without the new_tab option" do
+      render_component(
+        suggestion_link_text: "Travel abroad",
+        suggestion_link_url: "/travel-abroad",
+      )
+
+      assert_select "a[target='_blank']", false
+    end
+
+    it "appends accesible link text" do
```
andysellick, line 107: Nitpick: `accessible`

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: I don't think so because the banner can have just a link or just some text, whereas the check below is in case of both.

spec/components/intervention_spec.rb
chao-xian: Not a nitpick! It's a typo

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: Interesting. I don't know if we do or not really as it's what Max added in and I've only moved it to this hash. I'm open to removing it and seeing what happens as a case for making things tidy, but it might be a little out of scope for this PR otherwise.

app/views/govuk_publishing_components/components/_intervention.html.erb
andysellick: I don't understand... the logic in the brackets after the `||` is and-ing two variables that are the same variable - `suggestion_link_text`. Can't the second `suggestion_link_text` be removed? Or am I going mad? How does this work?

app/views/govuk_publishing_components/components/_intervention.html.erb
andysellick: The only place the class appears is in tests, so removing it would break those. But it isn't adding any styling. Agree, this is outside the scope of this PR, but can you raise an issue about it? Thanks.

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: omg sorry I think that's a typo from before as it should be link_text and link_url

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: Done

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: Done

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: Done

spec/components/intervention_spec.rb
chao-xian: I'd prefer to keep these separate. I feel that testing as specific as possible is a good thing - for clarity more than anything.

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -38,6 +38,17 @@ examples:
       suggestion_link_text: "You can now apply for a permit online."
       suggestion_link_url: "/permit"
 
+  open_suggestion_link_in_new_tab:
+    description: |
+      For example if you wanted to drive users to another online task, you don't want to completely hijack
```
andysellick, line 43: Nitpick: I don't think you need `For example` at the start of this sentence, it's perfectly clear without it.

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -38,6 +38,17 @@ examples:
...
+      their original flow on GOV.UK. Adding the option to open links in a new tab helps to address this.
+      Link text should tell the user that the link opens in a new tab.
+      Note: "(opens in a new tab)" is added to link text if the phrase isn't included.
```
andysellick, line 46: Are these newlines meant to render as newlines? You need to put an empty line between each one if that's the case, not sure if intended or not.

andysellick (approved):
Thanks for addressing my other comments, looks good. Just a couple of tiny nitpicks but nothing blocking.

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: Not a nitpick - thanks for the spot!

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: Intended - wanted them viewable within the a certain length in the source. Makes no difference as you say.

PR #2471, opened 2021-11-24
Fix single page notification button data attributes for tracking

danacotoran:
1. The data attributes for tracking were not named correctly. 
The `data-label` attribute  should be `data-track-label`, `data-action` should be `data-track-action`, `data-category` should be `data-track-category`.
2. Apply `data-module=gem-track-click` attribute to a wrapper div instead of the form itself. The reason for this is that
the form gets dynamically swapped out with the form returned from the personalisation endpoint. If `data-module=gem-track-click` was applied to the form, then in order for click tracking to fire correctly after the JS update has occurred, the track click module would have to be re-started on the updated form, which is not ideal. 
This is a workaround which can be credited to @andysellick

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
@@ -26,9 +26,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           var html = document.createElement('div')
           html.innerHTML = newButton
           // test that the html returned contains the button component; if yes, swap the button for the updated version
-          var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button .gem-c-single-page-notification-button__submit')
-          if (responseHasButton) {
-            this.$module.outerHTML = newButton
+          var responseButtonContainer = html.querySelector('form.gem-c-single-page-notification-button')
+          if (responseButtonContainer) {
+            this.$module.parentNode.replaceChild(responseButtonContainer, this.$module)
           }
         }
       }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Update search component ([PR #2462](https://github.com/alphagov/govuk_publishing_components/pull/2462))
 * Fix link to Crown Copyright in footer ([PR #2475](https://github.com/alphagov/govuk_publishing_components/pull/2475))
+* Fix single page notification button data attributes for tracking ([PR #2471](https://github.com/alphagov/govuk_publishing_components/pull/2471))
 
 ## 27.14.1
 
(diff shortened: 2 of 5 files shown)


danacotoran:
Re-drafted this while I investigate another potential tracking issue which could go in the same PR

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -28,7 +29,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           // test that the html returned contains the button component; if yes, swap the button for the updated version
           var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button .gem-c-single-page-notification-button__submit')
           if (responseHasButton) {
-            this.$module.outerHTML = newButton
+            var parent = this.$module.parentElement
+            var updatedModule = parent.insertBefore(html.firstChild, this.$module)
+            parent.removeChild(this.$module)
+            //  re-initialise click tracking on the updated element
+            var trackDetails = new window.GOVUK.Modules.GemTrackClick()
+            trackDetails.start($(updatedModule))
```
andysellick, line 37: I think I understand why the tracking is being reinitialised here but I wonder if there's an alternative approach. As long as some of the original HTML is present after the JS does the bit above, the tracking script should still work. If the entirety of the markup is rebuilt then I'd argue there might be a case for retaining the wrapping element if only to not have to reinitialise like this?

Related question - where is the tracking initialised in the first place? I can't see it - it seems odd that this code reinitialises it but doesn't start it in the first place (edit: just spotted, it's applied in the helper).

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -28,7 +29,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           // test that the html returned contains the button component; if yes, swap the button for the updated version
           var responseHasButton = html.querySelector('form.gem-c-single-page-notification-button .gem-c-single-page-notification-button__submit')
           if (responseHasButton) {
-            this.$module.outerHTML = newButton
+            var parent = this.$module.parentElement
+            var updatedModule = parent.insertBefore(html.firstChild, this.$module)
+            parent.removeChild(this.$module)
```
andysellick, line 34: I'm slightly nervous about removing `this.$module` entirely - although it looks like the module doesn't do anything else after this. Is that correct?

andysellick (commented):
Nothing wrong here but I'm wondering about the approach, although I could be wrong.

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
danacotoran: I have applied `data-module=gem-track-click` to a wrapper as per your suggestion, which eliminates the requirement to reinitialise tracking as I had done initially.

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
danacotoran: Yes the module does not do anything after this. The whole point of this script is solely to get the button to communicate accurately whether the current user has an active subscription to the current page.

andysellick (commented):
All looks good. I think originally there was a test for the JavaScript behaviour, but that seems to have disappeared? Would be worth putting it back in.

andysellick (approved):

PR #2476, opened 2021-11-25
Update Start button so info text is associated

### LINKED ISSUE #2083: Info text not associated with start button
This was raised as a question on the cross-government govuk-design-system Slack channel.

## Description of the issue
If you provide information text to a button, the text is rendered underneath the button but not linked to it in any way. This means it won't be read out to screenreader users when they focus on the button, and they could miss important information.

## Actual vs expected behaviour
Current behaviour:

When using the [start now button with info text](https://components.publishing.service.gov.uk/component-guide/button/start_now_button_with_info_text), the following HTML is output:

```html
<a class="gem-c-button govuk-button govuk-button--start" role="button" href="#">Start now...[svg here]</a>
<span class="gem-c-button__info-text">Sometimes you want to explain where a user is going to.</span>
```

This means the information text is completely separate from the button/link.

Expected behaviour: The information text is somehow linked to the button/link, e.g: using aria-labelledby

chris-gds:
## What

Update start button to link info text below

## Why

"If you provide information text to a button, the text is rendered underneath the button but not linked to it in any way. This means it won't be read out to screenreader users when they focus on the button, and they could miss important information."

Closes #2083 

## Visual Changes

No visual changes

### Screen reader example.

https://user-images.githubusercontent.com/71266765/143418992-921ca3f9-f737-4964-bf35-a64b877ebfd7.mov

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## 27.14.2
 
+* Update Start button so info text is associated ([PR #2476](https://github.com/alphagov/govuk_publishing_components/pull/2476))
 * Update search component ([PR #2462](https://github.com/alphagov/govuk_publishing_components/pull/2462))
 * Fix link to Crown Copyright in footer ([PR #2475](https://github.com/alphagov/govuk_publishing_components/pull/2475))
 * Fix single page notification button data attributes for tracking ([PR #2471](https://github.com/alphagov/govuk_publishing_components/pull/2471))
--- a/app/views/govuk_publishing_components/components/_button.html.erb
+++ b/app/views/govuk_publishing_components/components/_button.html.erb
@@ -27,5 +27,5 @@
 <% end %>
 
 <% if button.info_text %>
-  <%= tag.span button.info_text, class: button.info_text_classes %>
+  <%= tag.span button.info_text, button.info_text_options %>
 <% end %>
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -11,6 +11,7 @@ module GovukPublishingComponents
                   :rel,
                   :data_attributes,
                   :margin_bottom,
+                  :id,
                   :inline_layout,
                   :target,
                   :type,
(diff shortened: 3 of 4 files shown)


maxgds:
Hey Chris, this looks like a decent solution but I have a query. I can see you've definitely tested with voiceover as you've provided a video, but have you tested it with other screenreader/browser combinations? I'm a little bit concerned about the interaction between aria-hidden and aria-labelledby - is it reliably provided to AT if it's been aria-hidden in all cases? I tried googling to determine this without much success. An alternate approach might be instead of using labelledby, set an explicit label on the button so instead of:

```
<a class="gem-c-button govuk-button govuk-button--start" role="button" id="button-id-aa3dbebe" aria-labelledby="button-id-aa3dbebe info-text-id-60bfa2bc" draggable="false" href="#">
```
we wind up with
```
<a class="gem-c-button govuk-button govuk-button--start" role="button" aria-label="start-text info-text" draggable="false" href="#">
```

Let me know your thoughts!

chris-gds:
Good shout - let me do some deeper testing to see if this presents any issues.  The `hidden` attribute does create a feeling like something might not work as expected. 

I need to triple check this but I _think_ the main gotcha would that that `aria-label` wouldn't take the `ID` and convert into a string in this context.  However there might be a way around this should the hidden attribute cause an issue

chris-gds:
@maxgds I think this is all good. Works as expected 

JAWs

https://user-images.githubusercontent.com/71266765/143484626-161a88fa-47fd-4f8e-9f09-b40bf759f9ef.mov

NVDA Edge

https://user-images.githubusercontent.com/71266765/143484794-1343fc2c-3564-409e-bf5a-17dcedf4ac91.mov

NVDA FireFox

https://user-images.githubusercontent.com/71266765/143484623-19612eef-05a3-4d2e-889a-e7115677dcd4.mov

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update Start button with info text ([PR #2476](https://github.com/alphagov/govuk_publishing_components/pull/2476))
```
maxgds, line 12: Could you change either the PR title or this line so they're an exact match, please?

chris-gds: Course, done.

maxgds (approved):

PR #2478, opened 2021-11-25
Remove jQuery from ecommerce script

andysellick:
## What
Remove jQuery from the analytics ecommerce script.

This might look like a significant change because the script now assumes that its being passed a JS node rather than a jQuery object, but I checked and I'm fairly sure the only place this code is invoked is in `static-analytics.js` where it doesn't pass an element. The script is slightly strangely structured in that it's almost a module but if it isn't passed any elements it looks for them itself, in the `start` function. Theoretically we could take out this passed element and simplify the code, but I think that falls under the heading of making all of the analytics code more readable and usable.

## Why
We're removing jQuery from all of GOV.UK.

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/analytics/ecommerce.js
```diff
@@ -12,25 +12,25 @@
...
-        index = indexOverride ? parseInt(indexOverride, 10) - 1 : index
+      var searchQuery = GOVUK.analytics.stripPII(element.getAttribute('data-search-query')).substring(0, 100).toLowerCase()
+      var ecommerceRows = element.querySelectorAll('[data-ecommerce-row]')
+      var startPosition = parseInt(element.getAttribute('data-ecommerce-start-index'), 10)
+      var listTitle = element.getAttribute('data-list-title') || DEFAULT_LIST_TITLE
+      var variant = element.getAttribute('data-ecommerce-variant') || undefined
+      var trackClickLabel = element.getAttribute('data-track-click-label') || DEFAULT_TRACK_CLICK_LABEL
+
+      for (var i = 0; i < ecommerceRows.length; i++) {
+        var $ecommerceRow = ecommerceRows[i]
```
owenatgov, line 23: `ecommerceRows[i]` feels like it's short enough and communicates what it is without the need for this variable declaration. It also means we're scrubbing the convention of a `$` in front of variables to signify that they're jquery objects. What do you think about replacing the instances of `$ecommerceRow` below with `ecommerceRows[i]`?

owenatgov (commented):
One little code opinion. Cracking work as always!

I might not be the best person to give this a proper review and a test as I'm not super versed on the analytics code but I'm gonna hang around this PR and see if any discourse I could use to understand the code a bit better pops up if that's alright.

app/assets/javascripts/govuk_publishing_components/analytics/ecommerce.js
andysellick: My understanding is that if you're looping through an array and need to refer to the current array item more than once, it's more efficient to store that item in a variable than having to get it out of the array every time. It's a small optimisation in most cases (although in this case it saves 4 lookups).

Agree about the `$` though, have removed that, thanks.

owenatgov (approved):
Personally happy with the code 👍  Don't know if you want someone who's able to test this properly to take a look as well.

Issue #2238, opened 2021-07-30
Consolidate "white button with blue text" style


danacotoran:
The components gem currently has several components which use a similarly styled button, for instance: 
https://components.publishing.service.gov.uk/component-guide/govspeak/charts 
https://components.publishing.service.gov.uk/component-guide/subscription_links/with_copyable_feed_link 
https://components.publishing.service.gov.uk/component-guide/print_link

This style of button is not currently standardised, and each of these individual components has custom styling to achieve more or less the same result. 

Investigate if we can consolidate these styles into a new variation of the button component, in order to reduce repetition in our CSS.

chris-gds:
Agreed, in theory I would perhaps say these qualify as [secondary buttons](https://design-system.service.gov.uk/components/button/)

![image](https://user-images.githubusercontent.com/71266765/132886093-7054d3ba-40b7-443f-a887-97f10da3d54f.png)

andysellick:
Charts button now uses a standard secondary style button.

andysellick:
The remaining buttons referred to by this issue are:

[button in subscription links](https://components.publishing.service.gov.uk/component-guide/subscription_links/with_copyable_feed_link) | [print link button](https://components.publishing.service.gov.uk/component-guide/print_link)
---- | -----
<img width="415" height="81" alt="Image" src="https://github.com/user-attachments/assets/b1f491fb-5cad-4c37-a23d-145ede0703b7" /> | <img width="193" height="72" alt="Image" src="https://github.com/user-attachments/assets/8d67cab3-94f9-4a00-90c2-6eb0c8742d77" />

They're both similar (1px border, white background, contain SVG) but different (subs links button is bold, SVGs included in button text in first, background image in CSS in second).

Options are:

## Changing them to use a different button style, such as secondary button
Not really an option without causing visual inconsistencies. The print link button is used frequently alongside the 'get emails' component, which both use the same styles (see [example page at the bottom](https://www.gov.uk/guidance/creative-content-exchange-research-privacy-notice)). And the button in the subscription links component sits alongside bold links, which does broadly match the button style there.

<img width="475" height="80" alt="Image" src="https://github.com/user-attachments/assets/07122bad-97d2-4bfa-ae59-e25eee7a88a1" />

We could also change the 'get emails' button to match whatever new button style we use, and change the whole look of the subscription links component, but it feels like we'd be changing an increasing number of buttons rather than acknowledging that this style is in common use on GOV.UK.

[...]

PR #2480, opened 2021-11-26
Update charts to secondary button

chris-gds:
## What

Update Design of [button to match GOV.UK Design System](https://design-system.service.gov.uk/components/button/
) on GOVSpeak Charts.

## Why

[An issue was raised that highlighted several instances](https://github.com/alphagov/govuk_publishing_components/issues/2238) where we have unique buttons and button styles. We don't believe these are required and could be consolidated.  

These particular instances meet the criteria to match the Design System's secondary button properties.

### Before > After
image

## Anything else

Other buttons that meet criteria to be secondary buttons:
https://components.publishing.service.gov.uk/component-guide/subscription_links/with_copyable_feed_link
https://components.publishing.service.gov.uk/component-guide/print_link

## Design review link:

https://components-gem-pr-2480.herokuapp.com/component-guide/govspeak/charts/preview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,11 +10,12 @@
 ## Unreleased
 
 * Remove the last traces of jQuery ([PR #2702](https://github.com/alphagov/govuk_publishing_components/pull/2702))
-* Add tracking on the accordion ([#2693](https://github.com/alphagov/govuk_publishing_components/pull/2693))
+* Add tracking on the accordion ([PR #2693](https://github.com/alphagov/govuk_publishing_components/pull/2693))
+* Update charts to secondary button ([PR #2408](https://github.com/alphagov/govuk_publishing_components/pull/2480))
 
 ## 29.1.0
 
-* Remove "priority breadcrumbs" ([#2666](https://github.com/alphagov/govuk_publishing_components/pull/2666))
+* Remove "priority breadcrumbs" ([PR #2666](https://github.com/alphagov/govuk_publishing_components/pull/2666))
 
 ## 29.0.1
 
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
@@ -186,7 +186,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     toggleStatus.classList.add('govuk-visually-hidden', 'mc-toggle-status')
     toggleStatus.setAttribute('role', 'alert')
 
-    link.classList.add('govuk-body-s', 'mc-toggle-button')
+    link.classList.add('govuk-button', 'govuk-button--secondary')
     link.appendChild(toggleText)
     link.appendChild(toggleStatus)
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
@@ -271,26 +271,6 @@
     text-align: left;
   }
 
-  .mc-toggle-button {
-    border: 1px solid $govuk-border-colour;
-    color: $govuk-link-colour;
-    cursor: pointer;
-    margin: govuk-spacing(0);
-    padding: govuk-spacing(2);
-    background-color: govuk-colour("white");
-
-    &:hover {
-      background-color: govuk-colour("light-grey");
-      color: $govuk-link-hover-colour;
-    }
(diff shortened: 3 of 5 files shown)


chris-gds:
👋🏼  @mia-allers-gds / @nnagewad - Design feedback welcomed on these proposed changes.

mia-allers-gds:
I understand the reasoning for doing this but GOV.UK should be building on top of the design system, we don't have to follow them in every instance. If there are usability concerns with the current gov.uk buttons then I think we can work on them, but I would rather see us using nicer buttons everywhere

andysellick (approved):
This looks good @chris-gds. Definitely should remove these custom buttons - it's just adding more lines of CSS for users to download. We also have too many button variations for our button component and standardising to follow the DS is a good idea.

mia-allers-gds:
Hi @chris-gds I'm really keen for us to talk more about the button styles. I'm not convinced I agree with the use of the secondary button from the design system in this context. To me, these buttons have a different context to what you would find in, for example, a form or a cookie banner. I'd be keen for us to explore adapting the design system button into something more unique that fits the gov.uk content page layout more.

chris-gds:
Hey @mia-allers-gds - the objection to using the Design System secondary button on the chart in this context advocating for keeping the divergence. It would be good to understand more about the reasoning.

>To me, these buttons have a different 'vibe' to what you would find in, for example, a form or a cookie banner. I'd be keen for us to explore adapting the design system button into something more unique that fits the gov.uk content page layout more.

Are you thinking there's justification for a new _3rd_ button style to contribute to the Design System to address this?

> For what its worth, I think the current blue-outlined button could probably do with some love and a bit more affordance!

Can you expand on this?

nnagewad:
@chris-gds and @mia-allers-gds I've set up some time for us to chat in a Google Meet on Wednesday.

mia-allers-gds:
@chris-gds to answer your questions ahead of our chat:

> Are you thinking there's justification for a new 3rd button style to contribute to the Design System to address this?
I'm advocating for a 3rd button type for less important, in-page actions. It does not have to be submitted to the Design System. The Design System isn't a repository of every design element in the public sector, they only take things they think will be of value to a wide range of users

>Can you expand on this? (regarding improving the current button under discussion)
I think we have a lot of inconsistent buttons on gov.uk, and they should be consistent. If we replace them with a new style, I would want to improve what is there: the blue border button doesn't have much affordance, for example. So we could create something consistent and better. I'm not advocating for us to keep the blue-bordered buttons as-is

mia-allers-gds:
Update: @chris-gds @nnagewad and I had a chat about adding a third button style and decided we would go ahead with it.

**Why we need a third button style (distinct from the primary and secondary buttons from the Design System):**

- In-page interactions on content pages are secondary to consuming the content, and so buttons for these actions (toggles/ print/ rss subscription) should be less visually bold to avoid competing with the content
- A third button style will also allow us to create pages with primary and secondary interactions (such as start pages, forms, etc) that also have print/ rss/ toggles
- This new button style will be based off the Design System buttons
- The new buttons will need to be consistent but flexible, in case of the need for icons

chris-gds:
Thanks @mia-allers-gds, it was noted that the Design System's secondary button was not desirable in these 3 instances and a different style would be more suitable. 

Details of what this style is, when it's applicable, the benefits and the approach is still being decided. Also being factored into these conversations are [additional buttons on the gem](https://components.publishing.service.gov.uk/component-guide/button/preview).  Eg These have a "secondary quiet button" - currently there's is limited guidance around when or why this suitable to use and more research is needed into the context of their creation.

mia-allers-gds:
@chris-gds Could be that we expand on the secondary quiet button? Can chat about it further as we begin work I guess.

chris-gds:
> @chris-gds Could be that we expand on the secondary quiet button? Can chat about it further as we begin work I guess.

Potentially,  I think as a first step as GOVUK is a large consumer of the GOVUK Design System, and as this is essentially a paid for product, detailing a feature request of why this is believed to be needed would be appropriate. The response to this request (including timeline) can then inform next steps

chris-gds:
Alternative button style suggested here

https://github.com/alphagov/govuk-design-system-backlog/issues/154

chris-gds:
As this work is complete and it's unknown when this additional button style may be completed, I would recommend rolling out this work and then revising at later date as this update is an incremental improvement.

Acknowledging this might be closed without moving forward.

maxgds:
After discussing with our PM I'm closing this as it's a controversial change. It should be considered in future as part of reconsidering button styles as a whole.

PR #2481, opened 2021-11-26
Restore accessible autocomplete component

andysellick:
## What
Restores the accessible autocomplete component, with changes.

This uses the [accessible autocomplete](https://github.com/alphagov/accessible-autocomplete) code to progressively enhance a select element. This is based on the accessible autocomplete component that [existed briefly](https://github.com/alphagov/govuk_publishing_components/pull/1038) in the gem until we didn't need it anymore, but with some changes, specifically:

- we don't currently need tracking
- tracking might be needed in the future so this includes a custom `onConfirm` function (a custom `onConfirm` was originally included because we needed to [externally control the component](https://github.com/alphagov/finder-frontend/pull/970) in `finder-frontend`)
- the autocomplete `onConfirm` seems to have a bug where it fires when the autocomplete is blurred, but passes `undefined` even if something has been correctly entered in the autocomplete. An event listener has been added to the component to compensate for this
- all of the stuff to do with multiple selections has been removed, because we don't need that and it was really experimental
- tests have been updated

## Why
We're exploring using this component in a travel smart answer and it's currently the most suitable thing available.

## Visual Changes

Screenshot 2021-12-03 at 12 27 16

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accessible-autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accessible-autocomplete.js
@@ -1,61 +1,65 @@
 /* eslint-env jquery */
 /* global accessibleAutocomplete */
 // = require accessible-autocomplete/dist/accessible-autocomplete.min.js
-
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {};
 
 (function (Modules) {
-  'use strict'
+  function AccessibleAutocomplete ($module) {
+    this.$module = $module
+    this.selectElem = this.$module.querySelector('select')
+  }
 
-  Modules.AccessibleAutocomplete = function () {
-    var $selectElem
-
-    this.start = function ($element) {
-      $selectElem = $element.find('select')
-
-      var configOptions = {
-        selectElement: document.getElementById($selectElem.attr('id')),
-        showAllValues: true,
-        confirmOnBlur: true,
-        preserveNullOptions: true, // https://github.com/alphagov/accessible-autocomplete#null-options
-        defaultValue: ''
-      }
-
-      configOptions.onConfirm = this.onConfirm
-
-      new accessibleAutocomplete.enhanceSelectElement(configOptions) // eslint-disable-line no-new, new-cap
+  AccessibleAutocomplete.prototype.init = function () {
+    var configOptions = {
+      selectElement: document.getElementById(this.selectElem.getAttribute('id')),
+      autoselect: true,
+      confirmOnBlur: true,
+      preserveNullOptions: true, // https://github.com/alphagov/accessible-autocomplete#null-options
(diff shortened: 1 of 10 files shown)


andysellick:
I'm creating this with reservations - there's something of a disconnect between the autocomplete and the hidden `select`, in that it's possible to choose a valid option then edit it such that what the user thinks they've entered is different from what will actually be submitted in the hidden `select`. 

The use of the autocomplete with a hidden select was chosen to make the non-JS version slightly more helpful for users than an empty input but I'm strongly considering changing this component to use a [normal input and a datalist](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/datalist#examples).

alex-ju:
Before reviewing the code, if it's only needed in one application (be that `smart-answers`), shall we add it there instead (so we can iterate it quicker if needed)?

andysellick:
Thanks for the information @alex-ju. `content-data-admin` uses the autocomplete component from the gem (a really old gem when that was still a component) and I have some concerns about the autocomplete component in `content-publisher` - the code looks quite complicated (it has multiple configuration options including enhancing a standard input and a hidden select and having a multiple select option, which is more than I need) and there doesn't seem to be a test file.

I'm still trying to decide how to proceed. I think my options are:

1. port the autocomplete from content-publisher (although I might need to write a test file)
2. write an autocomplete in smart-answers that only does what I need i.e. enhances a regular input and datalist
3. write that 'simple' autocomplete here in the gem, then iterate it based on the content-publisher one with the aim of future consolidation

alex-ju:
I'm sharing some thoughts on this, but you are closer to the problem than I am.

> Thanks for the information @alex-ju. `content-data-admin` uses the autocomplete component from the gem (a really old gem when that was still a component) and I have some concerns about the autocomplete component in `content-publisher` - the code looks quite complicated (it has multiple configuration options including enhancing a standard input and a hidden select and having a multiple select option, which is more than I need) and there doesn't seem to be a test file.
> 
> I'm still trying to decide how to proceed. I think my options are:
> 
> 1. port the autocomplete from content-publisher (although I might need to write a test file)

The autocomplete code itself has unit tests in the main repo, hence in `content-publisher` we only made sure the different types of initialisations required in different areas of the project were captured in feature tests.

> 2. write an autocomplete in smart-answers that only does what I need i.e. enhances a regular input and datalist

This approach has the advantage of being able to test and iterate quickly without needing to release a new version of the component. It can also be retired with much ease.

> 3. write that 'simple' autocomplete here in the gem, then iterate it based on the content-publisher one with the aim of future consolidation

The work to converge/consolidate may fall between the cracks (as none the C19 or the Publishing team may be willing to take the time to do it).

andysellick:
@alex-ju I've created a new component in `smart-answers` based on this code and the code in `content-publisher`, see https://github.com/alphagov/smart-answers/pull/5694

PR #2482, opened 2021-11-29
Remove whitespace from pseudo-element content

injms:
## What

Removes the whitespace from the content declaration for pseduo-elements in the navigation header.

## Why

The spaces are probably being read aloud in screen readers, which isn't a helpful experience.

## Visual Changes

None.


owenatgov (approved):
Neat!

PR #2484, opened 2021-11-29
Update nav header search hover state underline

injms:
## What

Updates the hover state of the search link and search toggle button to have the same underline thickness as the other toggle buttons / links.

## Why

For visual consistency.

## Visual Changes

JavaScript unavailable before: 

JavaScript unavailable after: 

JavaScript available before:

JavaScript available after:


owenatgov (approved):
Great spot

PR #2492, opened 2021-11-30
Add margin_bottom option to success alert

danacotoran:
The GOV.UK success alert component has a fixed bottom margin which is fine in most use cases of this component.

Recently, however, the success alert component has been added to the top of publication pages whose content has a large margin on the top.
The result is that the visual space between the success alert and the content of the page is excessive, hence the need to introduce this option to the component.

## Currently: too much space
Screenshot 2021-11-30 at 17 24 14

---

https://trello.com/c/Z2eem0IC

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/success_alert.yml
+++ b/app/views/govuk_publishing_components/components/docs/success_alert.yml
@@ -37,3 +37,9 @@ examples:
         orci. Proin semper porttitor ipsum, vel maximus justo rutrum vel.
         Morbi volutpat facilisis libero. Donec posuere eget odio non egestas.
         Nullam sed neque quis turpis.
+  with_custom_margin_bottom:
+    description: |
+      The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). The default margins for the component are inherited from the [notification-banner](https://github.com/alphagov/govuk-frontend/blob/main/package/govuk/components/notification-banner/_index.scss) styles defined in the Design System.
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/success_alert.yml
```diff
@@ -37,3 +37,9 @@ examples:
         orci. Proin semper porttitor ipsum, vel maximus justo rutrum vel.
         Morbi volutpat facilisis libero. Donec posuere eget odio non egestas.
         Nullam sed neque quis turpis.
+  with_custom_margin_bottom:
+    description: |
+      The component accepts a number for margin bottom from 0 to 9 (0px to 60px) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). The default margins for the component are inherited from the [notification-banner](https://github.com/alphagov/govuk-frontend/blob/main/package/govuk/components/notification-banner/_index.scss) styles defined in the Design System.
```
andysellick, line 42: Maybe amend this to say what the margin is? e.g. `By default the component has a bottom margin of 50px, inherited from...`

andysellick (approved):
All good, obviously a test failure there that needs a fix but otherwise looks fine. I'd merge the two commits together since they relate to the same change.

app/views/govuk_publishing_components/components/docs/success_alert.yml
danacotoran: I wondered about doing that but I didn't think it would be a good idea because the documentation would immediately go out of date if GOVUK Frontend update the spacing on this component at all. I thought linking to the `scss` file in `govuk-frontend` was a good compromise.

PR #2493, opened 2021-12-01
Remove use of govuk-font from the big number component

owenatgov:
## What
Updates the [big number](https://components.publishing.service.gov.uk/component-guide/big_number) component to not use the design system's [`govuk-font`](https://components.publishing.service.gov.uk/component-guide/big_number) sass mixin.

## Why
Currently, the govuk-font mixin enforces responsive typography when in this case we want it to maintain the same font between screen sizes. Additionally, the current label font size is 16px which reverts to 14px on mobile, an inaccessible font size. We use a similar approach on the super navigation header, check that component for examples if you're curious about this.

### Trello cards
- https://trello.com/c/5VUAmVR7/682-3homepage-big-numbers-on-tablet-fix-sizing
- https://trello.com/c/RU79pucB/688-homepage-mobile-big-numbers-are-not-big

## Visual Changes
There is no visual change to the big number on tablet or desktop. Below is using [this page](https://components.publishing.service.gov.uk/component-guide/big_number/with_link) for reference.

### Mobile (tested at 321px)
Before:
Screenshot 2021-12-01 at 14 36 23

After:
Screenshot 2021-12-01 at 14 36 35

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
@@ -1,13 +1,25 @@
 .gem-c-big-number {
   margin-bottom: govuk-spacing(3);
+  font-family: $govuk-font-family;
 }
 
 .gem-c-big-number__value {
-  @include govuk-font($size: 80, $weight: bold);
+  font-size: 80px;
+  font-weight: 700;
+  line-height: 1;
+
+  @if $govuk-typography-use-rem {
+    font-size: govuk-px-to-rem(80px);
+  }
 }
 
 .gem-c-big-number__label {
-  @include govuk-font($size: 16, $weight: bold, $line-height: 2);
+  font-size: 16px;
+  font-weight: 700;
+
+  @if $govuk-typography-use-rem {
+    font-size: govuk-px-to-rem(16px);
+  }
 
   // This pseudo element is to bypass an issue with NVDA where block level elements are dictated separately.
   // What's happening here is that the label and the number technically have an inline relationship but appear to have a block relationship thanks to this element
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Remove use of govuk-font from the big number component ([PR #2493](https://github.com/alphagov/govuk_publishing_components/pull/2493))
+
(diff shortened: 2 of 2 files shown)


mia-allers-gds:
Looks good to me!

owenatgov:
@chao-xian Sorted out 👍

chao-xian (approved):

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,13 +1,25 @@
 .gem-c-big-number {
   margin-bottom: govuk-spacing(3);
+  font-family: $govuk-font-family;
```
36degrees, line 3: I know this has been merged already, but I'd suggest using the [`govuk-typography-common`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-typography-common) mixin here instead – which will take care of things like:

- anti-aliasing
- swapping out the font when printing
- any other changes to the 'core' typography introduced in the future

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,13 +1,25 @@
...
+
+  @if $govuk-typography-use-rem {
+    font-size: govuk-px-to-rem(80px);
+  }
 }
 
 .gem-c-big-number__label {
-  @include govuk-font($size: 16, $weight: bold, $line-height: 2);
+  font-size: 16px;
+  font-weight: 700;
```
36degrees, line 18: For similar reasons to above, I'd suggest using the [`govuk-typography-weight-bold`](https://frontend.design-system.service.gov.uk/sass-api-reference/#govuk-typography-weight-bold) mixin.

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,13 +1,25 @@
 .gem-c-big-number {
```
36degrees, line 1: I'd suggest including the `govuk-text-colour` mixin too – black text should be `#0b0c0c`, but at the minute this will be using the user agent default text colour. That'll probably be `#000000` in most cases, but may not always be, which could cause accessibility issues – see https://www.w3.org/TR/WCAG20-TECHS/F24.html.

36degrees (commented):
Sorry to catch these after the merge, but hopefully this is still useful…

PR #2494, opened 2021-12-02
Use the correct mixins for applying font in the big number component

owenatgov:
## What
A follow on from https://github.com/alphagov/govuk_publishing_components/pull/2493 to use design system mixins and variables in as many places as possible whilst not using `govuk-font`.

## Why
Following the merging of 2493, Ollie Byford from the design system team left some comments noting the existence of mixins like `govuk-typography-common` and how they would be useful here. See his review on 2493 for details.

### Trello cards
- https://trello.com/c/5VUAmVR7/682-3homepage-big-numbers-on-tablet-fix-sizing
- https://trello.com/c/RU79pucB/688-homepage-mobile-big-numbers-are-not-big

Will not be including visual changes as the only change is the difference in colour from system black (#000000) to the design system "black" which isn't perceivable to the naked eye.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
@@ -1,11 +1,12 @@
 .gem-c-big-number {
   margin-bottom: govuk-spacing(3);
-  font-family: $govuk-font-family;
+  @include govuk-typography-common;
+  @include govuk-text-colour;
 }
 
 .gem-c-big-number__value {
   font-size: 80px;
-  font-weight: 700;
+  @include govuk-typography-weight-bold;
   line-height: 1;
 
   @if $govuk-typography-use-rem {
@@ -15,7 +16,7 @@
 
 .gem-c-big-number__label {
   font-size: 16px;
-  font-weight: 700;
+  @include govuk-typography-weight-bold;
 
   @if $govuk-typography-use-rem {
     font-size: govuk-px-to-rem(16px);
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Use the correct mixins for applying font in the big number component ([PR #2494](https://github.com/alphagov/govuk_publishing_components/pull/2494))
 * Remove use of govuk-font from the big number component ([PR #2493](https://github.com/alphagov/govuk_publishing_components/pull/2493))
 
 ## 27.15.0


app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,11 +1,12 @@
 .gem-c-big-number {
   margin-bottom: govuk-spacing(3);
-  font-family: $govuk-font-family;
+  @include govuk-typography-common($font-family: $govuk-font-family);
+  color: $govuk-text-colour;
```
36degrees, line 4: Use the mixin – it'll take care of setting back to `#000000` when printing.

```suggestion
  @include govuk-text-colour;
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,11 +1,12 @@
...
 }
 
 .gem-c-big-number__value {
   font-size: 80px;
-  font-weight: 700;
+  @include govuk-typography-weight-bold($important: false);
```
36degrees, line 9: No need to pass `$important: false` – that's the default behaviour.

```suggestion
  @include govuk-typography-weight-bold;
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -15,7 +16,7 @@
 
 .gem-c-big-number__label {
   font-size: 16px;
-  font-weight: 700;
+  @include govuk-typography-weight-bold($important: false);
```
36degrees, line 19:
```suggestion
  @include govuk-typography-weight-bold;
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -1,11 +1,12 @@
 .gem-c-big-number {
   margin-bottom: govuk-spacing(3);
-  font-family: $govuk-font-family;
+  @include govuk-typography-common($font-family: $govuk-font-family);
```
36degrees, line 3: No need to pass the font family unless you want to override it for some reason.

```suggestion
  @include govuk-typography-common;
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
owenatgov: This is what happens when I do a page find in the DS sass reference and then don't skip beyond the first item 🤦

chao-xian (approved):
Thanks @36degrees! Speedy work @owenatgov

36degrees (approved):
Great stuff 💥

PR #2497, opened 2021-12-03
Import translations and sync locale files

1pretz1:
Imports translations from ASP.

More info in the commits ->

Trello:
https://trello.com/c/Xz78vNR4/2757-import-newly-sourced-translations-into-locale-files-5


config/i18n-tasks.yml
```diff
@@ -2,3 +2,11 @@ data:
   yaml:
     write:
       line_width: -1
+search:
+  paths:
+    - app/
+    - lib/
```
ChrisBAshton, line 8: I don't see any locales in `lib/` - why have we added this?

ChrisBAshton (approved):
This looks great, Peter 👍  one question below, but I'm sure there's a good answer to it.

config/i18n-tasks.yml
1pretz1: There are no locales stored in `lib/` but some are used there, e.g: https://github.com/alphagov/govuk_publishing_components/blob/5016189274096ef76176b34df7c3a473f95a8c16/lib/govuk_publishing_components/presenters/devolved_nations_helper.rb#L3. I18n-tasks doesn't look in `lib/` by default so will report keys as unused if there are any used in directories other than `app/`. I've updated the commit message which should make this clearer.

ChrisBAshton (approved):
Re-approved (for clarity)!

PR #2483, opened 2021-11-29
Update navigation header toggle button spacing

injms:
## What

Update navigation header link / toggle button spacing for the desktop layout.

Ticket: https://trello.com/c/aZhaaNyT

Preview page: https://components-gem-pr-2483.herokuapp.com/public

## Why

To make the spacing in the desktoplayout look more balanced.

## Visual Changes

Preview page: https://components-gem-pr-2483.herokuapp.com/public

### Before

components publishing service gov uk_public (7)

### After

govuk-publishing-components dev gov uk_public (18)

PR #2501, opened 2021-12-08
Alter use of pseudo-underline mixin to allow for different button sizes

owenatgov:
## What
Adds options to the `pseudo-underline` mixin within the super nav header to allow them to take different `left` and `right` attributes based on the button using the mixin.

## Why
Led by a bug report from Stephen McCarthy that the underline on the mobile menu button looks a bit weird.

The reason I've made the choice to alter how the mixin is initiated rather than just change the spacing is that this initial spacing is a conscious choice made in https://github.com/alphagov/govuk_publishing_components/pull/2483 This maintains the intent of 2483 whilst fixing this issue.

[Card](https://trello.com/c/SkO1xxdS/703-fix-pseudo-underline-on-mobile-menu-button)

### Before
Screenshot 2021-12-08 at 14 14 13

### After
Screenshot 2021-12-08 at 14 14 25

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Alter use of pseudo-underline mixin to allow for different button sizes ([#2501](https://github.com/alphagov/govuk_publishing_components/pull/2501))
+
 ## 27.16.0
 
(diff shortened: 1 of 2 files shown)


mia-allers-gds:
Looks much better, thank you!

chao-xian (commented):
Nice work Owen!

chao-xian (approved):

PR #2502, opened 2021-12-08
Re-work `explicit-cross-domain-links.js`

danacotoran:
It has transpired that the `explicit-cross-domain-links.js` script was not
set up correctly which is breaking our cross domain tracking currently.

The source of the issue is the `_ga` parameter which contains a timestamp
which actually expires and becomes invalid after two minutes.

The way the script worked previously was that the relevant cross-domain
links and forms on the page were decorated with the `_ga` and
`cookie_consent` parameters on page load. Doing it on page load is an
issue due to the time-sensitive nature of the `_ga` parameter, as
mentioned above.

This changes the way the script works so that instead of decorating
crossdomain links and forms with additional parameters on page load, we
decorate them when they are interacted with.
Besides ensuring that the `_ga` param is always valid and up-to-date, this
also makes it so that there is no longer a need to listen for cookie
banner events, so I also removed the `cookie-reject` event as part of this
commit (this event was explicitly introduced for the `explicit-cross-domain-links.js`
script and is not used anywhere else currently).

----

https://trello.com/c/VaRq2typ/27-fix-cross-domain-issue

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
@@ -6,33 +6,57 @@
 
   GOVUK.Modules.ExplicitCrossDomainLinks = function () {
     this.start = function ($module) {
-      var element = $module[0]
+      this.element = $module[0]
+      this.attribute = 'href'
+      this.attributeValue = this.element.getAttribute(this.attribute)
+      this.eventType = 'click'
+      if (!this.attributeValue) {
+        this.attribute = 'action'
+        this.attributeValue = this.element.getAttribute(this.attribute)
+        this.eventType = 'submit'
+      }
 
+      this.handleEvent = this.handleEvent.bind(this)
+      this.handleCookiesAccepted = this.handleCookiesAccepted.bind(this)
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -6,78 +6,79 @@
...
+        this.element.click()
       }
+    }
 
-      this.decorate(element, 'cookie_consent=accept')
+    this.handleCookiesAccepted = function () {
+      // If the cookie banner was engaged and usage cookie accepted, append ?_ga=clientid if available and cookie_consent=accept
+      var element = this.element
+      var attribute = this.attribute
+      this.decorate(element, 'cookie_consent=accept', this.attribute)
```
andysellick, line 55: You could use the `attribute` variable here, set on the previous line.

andysellick (approved):
Looks good, couple of small comments but happy to approve.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
danacotoran: Yup you are right, have corrected 🙏

PR #2509, opened 2021-12-09
Fix font for menu paragraphs

chao-xian:
## What / Why
Transport was not explicitly set on the menu items, so if viewing in Heroku for example the menu styling would appear broken.

### Before
Screenshot 2021-12-08 at 8 23 34 pm

### After
Screenshot 2021-12-09 at 4 07 54 pm


owenatgov (approved):
Luuuuuurvly

PR #2510, opened 2021-12-09
Add large mode on mobile only to search component

owenatgov:
## What
Adds `large-mobile` as an option for the `size` attribute on the [search component](https://components.publishing.service.gov.uk/component-guide/search). This means that the "large" styles, previously applied to all screen sizes via `size: "large"`, will only be applied to the search component on mobile.

## Why
Recommendation from our designer as part of the homepage work to make sure that the search box looks nice between screen sizes.

[Card](https://trello.com/c/Ss9cKnkL/702-homepage-stretch-goal-smaller-search-bar-on-desktop)

### Desktop
Screenshot 2021-12-09 at 16 17 16

### Mobile
Screenshot 2021-12-09 at 16 17 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix font for menu paragraphs [#2509](https://github.com/alphagov/govuk_publishing_components/pull/2509)
+* Add large mode on mobile only to search component ([PR #2510](https://github.com/alphagov/govuk_publishing_components/pull/2510))
 
 ## 27.17.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -1,6 +1,25 @@
 $input-size: 40px;
 $large-input-size: 50px;
 
+@mixin large-mode {
+  .gem-c-search__label {
+    @include govuk-font($size: 19, $line-height: $large-input-size);
+  }
+
+  .gem-c-search__input[type="search"] {
+    height: $large-input-size;
+  }
+
+  .gem-c-search__submit {
+    width: $large-input-size;
+    height: $large-input-size;
+
+    .gem-c-search__icon {
+      @include icon-positioning($large-input-size);
+    }
+  }
+}
+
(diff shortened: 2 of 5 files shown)


owenatgov:
Just realised this needs a test!

chao-xian:
I wonder if it's worth inverting this as mobile visits > desktop visits. Thoughts on that?

alex-ju:
> I wonder if it's worth inverting this as mobile visits > desktop visits. Thoughts on that?

+ 1 to this – it fits the mobile-first aproach

owenatgov:
@chao-xian Could you clarify what you mean when you say inverting?

chao-xian:
As in, we have desktop specific styles to make it small and the default is for mobile and already big. We can do that later if it's harder now considering we need to ship soon.

mia-allers-gds:
Looks great @owenatgov !

owenatgov:
@chao-xian Gotcha. I broadly agree with this approach, however I foresee 2 possible implementations and I have concerns about both:

1. Apply the big size as the default on mobile everywhere. This would be a design change for the component as the default on both desktop and mobile across govuk is the "small" version. I'm a little wary of this as it's a potentially significant design change that might alter the context it's being used in in some places. From a quick glance I've spotted this in use in [31 places](https://github.com/search?p=1&q=org%3Aalphagov+components%2Fcomponents%2Fsearch&type=Code). Not all of them will be live or relevant (some are just import declarations) but I'd like to do an audit, test and get a designers opinion first.
2. Re-organise the CSS so that we are applying small styles separately but maintaining. This feels a bit awkward because right now, small styles on mobile is the default so we'll have this default large styling but then we're overriding it almost everywhere.

I do think doing one of the above is the right thing to do, ideally 1. I'm happy to negotiate what we do now/alter this PR to encompass the wider work around this as it's not critical.

chao-xian:
I'll approve this for now and we can do 1 as a future fix. Sound good?

chao-xian (approved):

owenatgov:
@chao-xian Cool beans, thanks Kelv. I'll jot up an issue detailing what we've introduced and what we wanna do.

Could you please approve Percy?

PR #2516, opened 2021-12-13
Add flag to hide Intervention Banner component

chao-xian:
## What

Add a param to the component to hide the banner by default. This allows for logic in both the backend or Javascript to manipulate that component easily to display it based on whatever logic.

## Why
We want to selectively display a link to an online tree test for specific topics on GOV.UK. Adding this flag allows us to control whether to display the banner for our user research recruitment.

Related PR: https://github.com/alphagov/collections/pull/2630

[Trello card](https://trello.com/c/1o6QiDxC/646-add-banner-for-user-research-exercise-to-selected-pages)

## Visual Changes
Screenshot 2021-12-13 at 6 00 35 pm

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/intervention.yml
+++ b/app/views/govuk_publishing_components/components/docs/intervention.yml
@@ -38,6 +38,15 @@ examples:
       suggestion_link_text: "You can now apply for a permit online."
       suggestion_link_url: "/permit"
 
+  with_hide:
+    description: |
+      This example is for when we want to hide by default and display to the user based on some logic,
+      either in the backend or with Javascript.
+    data:
+      suggestion_link_text: "You may be invited to fill in a questionnaire"
+      suggestion_link_url: "/questionnaire"
+      hide: true
+
   open_suggestion_link_in_new_tab:
     description: |
       When sending users to another online task, you don't want to completely hijack
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleaased
+
+* Add flag to hide Intervention Banner component [#2516](https://github.com/alphagov/govuk_publishing_components/pull/2516)
+
 ## 27.18.0
 
 * Remove jQuery from page-content ([PR #2505](https://github.com/alphagov/govuk_publishing_components/pull/2505))
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -3,7 +3,9 @@
   suggestion_link_text ||= false
   suggestion_link_url ||= false
   suggestion_text ||= nil
+  hide ||= false
   new_tab ||= false
+
 
   data_attributes ||= {}
   data_attributes[:module] = 'intervention'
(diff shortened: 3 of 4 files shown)


owenatgov:
So, this code looks fine to me but I'm wondering about the need for a `hide` attribute. As this is only going to be used in collections, could you not just not render the component via the backend? Looking at https://github.com/alphagov/collections/pull/2630 you could so something like:

```ruby
if !@hide_recruitment_banner
    <%= render "govuk_publishing_components/components/intervention", {
        suggestion_text: "Help improve GOV.UK",
        suggestion_link_text: "Sign up to take part in user research",
        suggestion_link_url: "https://www.optimalworkshop.com/treejack/",
        new_tab: true,
      } %>
end
```

This feels simple enough that you could just do that, even if we wanted to roll it out elsewhere. That'd save the need for an additional API attribute to support in the gem. What do you think?

chao-xian:
@owenatgov Miller Columns is why

owenatgov:
@chao-xian Spell it out for me please.

chao-xian:
@owenatgov yes, soz. So even if it's hidden/shown according to the backend, as you're interacting with the miller columns there's fancy JS that interacts with the history api and the paths change but the page only fetches data from the backend, not the markup. So we have to manipulate the dom somehow and it was a lot easier to have the component present but hidden by default then change that attribute. I guess the alternative would be to hide it on load but this felt a more consistent way to control what's happening.

Here's the [JS in Collections](https://github.com/alphagov/collections/pull/2630/commits/fb6827c365948caa3521bcdc63f9302a2f3bc550#diff-789d7ba47d3014a1e3e6f434f0d91d3eb9e1ae2fb77f4bc8101bc46e523da3f9R331)

owenatgov:
@chao-xian Gotcha, thanks for the explanation. What a major bother. If this solution works then lets do it for now. I am quite keen to remove this once we retire the Miller columns from browse so I'll note this down somewhere.

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -38,6 +38,15 @@ examples:
       suggestion_link_text: "You can now apply for a permit online."
       suggestion_link_url: "/permit"
 
+  with_hide:
+    description: |
+      This example is for when we want to hide by default and display to the user based on some logic,
+      either in the backend or with Javascript.
```
owenatgov, line 44: nitpick: why is this on a different line?

owenatgov (approved):
A singular docs nitpick. Looking good 👍

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: It doesn't affect the output but just to keep it within a useful width for editing (simply a style guide thing which we don't have defined for MD specifically but is what we have for the rest of our code syntaxes)

owenatgov: I'm not too sure, personally. Breaking it presumes dev screen size and I _think_ most editors can wrap text anyway for easy of editing. Do you have any examples of where else we do this?

Caveat to this comment that I am absolutely not going to block this going out over a disagreement of MD styling. If you've gotta merge then merge away.

chao-xian: I know most editors will soft wrap but the general standard otherwise is ~80 chars and it just looks tidier... ¯\_(ツ)_/¯ 

Thank you again for your review!

PR #2517, opened 2021-12-14
Port the grid_helper sass mixin to the components gem

owenatgov:
## What
Replace the grid mixins from the super nav sass into a separate mixin file and update the grid helper in line with the most recent updates made during the new homepage work (https://github.com/alphagov/frontend/blob/main/app/assets/stylesheets/mixins/_grid-helper.scss).

## Why
The new homepage is using the same grid setup as the super nav. This will allow us to distribute the mixin between the super nav and the homepage to repeat code repetition.

No visual changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Fix font for menu paragraphs [#2509](https://github.com/alphagov/govuk_publishing_components/pull/2509)
 * Add large mode on mobile only to search component ([PR #2510](https://github.com/alphagov/govuk_publishing_components/pull/2510))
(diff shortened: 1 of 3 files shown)


alex-ju (approved):
I agree, sharing these mixins makes sense. We may want to have a conversation about where is the best place for them (we have `components/helpers` and `components/mixins` and I'm sure I'm not the only puzzled about the difference between them) and whether we'd like to prefix them, to avoid potential clashes with similar mixins defined at the application level.

owenatgov:
@alex-ju I'm also keen to have a chat about this, mainly as I'd forgotten that `components/helpers` existed 😅  A topic for the new year to try and bring some consistency to the codebase. I'll be cheeky and merge for now.

PR #2518, opened 2021-12-14
Remove Brexit call to action from contextual sidebar

sihugh:
## What
Remove the Brexit call to action from contextual sidebar

## Why
Use of the link is dwindling and there are other pressing things that might need greater visibility.

## Visual Changes
As expected, the element won't be rendered by the component anymore


lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -108,51 +108,6 @@ def tagged_to_brexit?
...
-        taxons = content_item.dig("links", "taxons").to_a
-        taxons.each do |taxon|
-          if brexit_cta_taxon_allow_list.include?(taxon["content_id"]) || parent_taxon_include?(taxon, brexit_cta_taxon_allow_list)
-            return true
-          end
-        end
-        false
-      end
-
-      def parent_taxon_include?(taxon, taxon_list)
```
hannako, line 140: I'm a bit sad to lose this lovely recursive method!

hannako (approved):
I spotted a few references to transition-check in the translation files. Which seems weird as that's been retired and the url is redirected to the brexit landing page. But I'll sort that in a separate PR, as it's unrelated to this.

This looks good to me 🥳

lib/govuk_publishing_components/presenters/contextual_navigation.rb
alex-ju: I was too, might actually be the last one in this repo 😢

PR #2521, opened 2021-12-14
Add missing class for slimmer header

chao-xian:
## What

This class is required for slimmer to inject service titles. For example the Service Manual.

## Why

The old template had this class, [and Slimmer expects it in order to inject a custom title header](https://github.com/alphagov/slimmer/blob/master/lib/slimmer/processors/inside_header_inserter.rb#L7-L8).

[Trello card](https://trello.com/c/7xbAX1Uk/711-bug-fix-service-manual-header)

## Visual Changes

None

## Related PRs

* https://github.com/alphagov/static/pull/2672
* https://github.com/alphagov/service-manual-frontend/pull/1021


owenatgov (commented):
Changelog pls then I'll approve 🙏

owenatgov (approved):
🎅

alex-ju (commented):
Alternatively, we could've updated `slimmer to use the existing `gem-c-header__logotype`. I am slightly worried that someone will find this class without any associated styles and will just get rid of it.

chao-xian:
Ah good point @alex-ju - which might be how it happened in the first place. I'll raise a couple PRs if I get a chance.

PR #2525, opened 2021-12-15
Add signup link component

andysellick:
## What
Add the signup link component from `collections`.

- unchanged apart from `app-c` CSS class prefix changing to `gem-c` and tweaks to the test structure to make it run here

## Why
We want to use it in a smart-answer that's currently under development. Once this component has been moved from `collections` to here in the gem, we'll remove the one in `collections` and point it at this one.

## Visual Changes
Screenshot 2021-12-15 at 14 55 14

Trello card: https://trello.com/c/i3brOfzz/431-add-multiple-country-select-question-type-to-smart-answers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add signup link component ([PR #2525](https://github.com/alphagov/govuk_publishing_components/pull/2525))
 * Remove jQuery from page-content ([PR #2505](https://github.com/alphagov/govuk_publishing_components/pull/2505))
 * Replace use of `includes()` in `explicit-cross-domain-links.js` with `indexOf()` alternative [#2515](https://github.com/alphagov/govuk_publishing_components/pull/2515)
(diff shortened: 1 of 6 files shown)


leenagupte (approved):
It matches the existing component in collections, so 👍  from me.

alex-ju:
It would've been great to prepare the PRs in `smart-answer` and `collection` to make sure we don't add a 3rd component to the existing 2 ones and leave tech debt behind. For reference `smart-answers` already has an [email-link](http://smart-answers-preview.herokuapp.com/component-guide/email-link) component and `collections` has a [signup-link](http://govuk-collections.herokuapp.com/component-guide/signup-link) component (the source of this PR, as I understand), both looking and functioning in a very similar way.

andysellick:
Thanks for the input @alex-ju, I wasn't aware `smart-answers` already had that component. Hopefully this doesn't change much though - I was planning to remove the component in `collections` but now I'll also remove the similar one in `smart-answers` and make sure all the functionality is covered.

PR #2534, opened 2021-12-17
Remove jQuery from download link tracker

andysellick:
## What
Removes jQuery from the download link tracker, which tracks clicks on links to downloadable assets.

It works as follows. Once users consent to analytics, the tracker attaches a click event listener to all pages on the body element. If the clicked element is a link (or inside a link) that matches a particular selector, a tracking event is fired to GA.

The download link tracker is called in `static-analytics.js` and the selector passed is `'a[href*="/government/uploads"], a[href*="assets.publishing.service.gov.uk"]'`. Broadly, these match links to downloadable assets.

I've had to change the code more than I expected to remove jQuery owing to how easy it is to attach an event listener to multiple elements in jQuery compared to the relative complexity of that in vanilla JS. This change has meant that the `getLinkFromEvent` function is no longer needed, as the check for the element type needs to be done much earlier in the code, before the tracking function is called.

## Why

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove jQuery from download link tracker ([PR #2534](https://github.com/alphagov/govuk_publishing_components/pull/2534))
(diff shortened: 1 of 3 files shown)


owenatgov (approved):
⬇️

PR #2536, opened 2021-12-20
Fix help-notice spacing in call-to-action

alex-ju:
## What
Fix `help-notice` spacing in `call-to-action`

## Why
When a `help-notice` element is used inside a `call-to-action` an excessive space is created as the padding of the outer element adds to the margin of the inner element. This change reduces this excessive space by removing the `margin-top` on the inner element if it is the first element within the `call-to-action` container.

## Visual Changes

<table>
<tr><td valign="top">

Screenshot 2021-12-20 at 10 57 23

</td><td valign="top">

Screenshot 2021-12-20 at 10 57 59

</td></tr>
</table>


jon-kirwan (approved):
Hi Alex, checked against https://www.gov.uk/foreign-travel-advice/australia/coronavirus and looks fine to me. One small thing to mention. Since the issue happens when the `help-notice` element is used inside a call-to-action - I wondered if the sass might need to be structured in the same way? e.g.
```
.call-to-action .help-notice:first-child {
    margin-top: 0;
}
```
Otherwise all looks good so have approved 👍

alex-ju:
> Hi Alex, checked against https://www.gov.uk/foreign-travel-advice/australia/coronavirus and looks fine to me. One small thing to mention. Since the issue happens when the `help-notice` element is used inside a call-to-action - I wondered if the sass might need to be structured in the same way? e.g.
> 
> ```
> .call-to-action .help-notice:first-child {
>     margin-top: 0;
> }
> ```
> 
> Otherwise all looks good so have approved 👍

Ah, yes, I thought I'm already within a `call-to-action` block. Will update – good spot!

PR #2538, opened 2021-12-21
Remove jQuery from external link tracker

andysellick:
## What
Removes jQuery from the external link tracker. This is a script that sends a tracking event when a consenting user clicks on a link to a page that isn't on www.gov.uk.

I've had to restructure this code quite a bit, partly because of the way event listeners work in JS vs jQuery and partly because the tests kept failing. 

The reason behind this was the `click` event listener was persisting between tests with the previous value for customDimension, so the test was technically passing but failing because the click event was occurring multiple times. This was originally solved with jQuery by using the `off` event listener remove, but in JS you have to pass the function that attached the listener, which was far more complex. Instead, the options are now attached to the external link tracker object, and get overwritten. This means the click events are still firing multiple times in the tests, but they're all firing correctly (in reality the tracker would only ever be initialised once). This tracker is only called here in the gem, by `static-analytics.js`.

Another change is how the code detects that the link is an external link. Before this was done using a clever selector, but the JS equivalent (`matches`) was resulting in some weird errors, so I've had to go a bit old school and parse the `href` attribute manually.

This code shares some similarities with the [download link tracker](https://github.com/alphagov/govuk_publishing_components/pull/2534), so maybe at some point we could try to consolidate them.

## Why

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove jQuery from external link tracker ([PR #2538](https://github.com/alphagov/govuk_publishing_components/pull/2538))
 * Remove jQuery from download link tracker ([PR #2534](https://github.com/alphagov/govuk_publishing_components/pull/2534))
 * **BREAKING:** Update feedback component to fix accessibility issues ([PR #2435](https://github.com/alphagov/govuk_publishing_components/pull/2435))
   You must make the following changes when you migrate to this release:
--- a/app/assets/javascripts/govuk_publishing_components/analytics/external-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/external-link-tracker.js
@@ -1,49 +1,63 @@
+// = require govuk/vendor/polyfills/Element/prototype/closest.js
 ;(function (global) {
   'use strict'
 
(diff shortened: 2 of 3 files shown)


owenatgov (commented):
One errant comment which I have a feeling has a very reasonable explanation that I'm missing. Also needs a changelog update.

I agree on your assessment that this could be combined with the download link tracker, not a problem for right now though. Could you make an issue for this please?

andysellick:
Issue logged: https://github.com/alphagov/govuk_publishing_components/issues/2550

@owenatgov are you referring to the multi line comment in the middle of the script? That was there originally so I just left it as it was.

owenatgov:
@andysellick So, I thought I'd made a comment in my review about the national archive link numbering but I suspect I forgot to hit enter to actually post said comment before submitting my review 🤦  Have slacked you about it.

andysellick:
@owenatgov to answer your question - I modified the spec file to add numbers to the links so that I could be sure that the right thing was working. There wasn't anything wrong with the test, but I found it hard during debugging to know at a glance which link had been clicked (and therefore which tracking event was working or not working) as they all had the same URL.

owenatgov (approved):
All good!

PR #2540, opened 2021-12-23
Remove jQuery from Google Analytics Universal Tracker

andysellick:
## What
Remove jQuery from Google Analytics Universal Tracker. This is the main script in the analytics code that contains all the things the rest of it depends on, like trackPageView and trackEvent, plus it also contains the snippet of JS from GA that makes everything work.

## Why

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


owenatgov (commented):
Needs a changelog update. Otherwise looking swell.

owenatgov (approved):
⭐

PR #2542, opened 2021-12-23
Remove jQuery from mailto-link-tracker

andysellick:
## What
Removes jQuery from the mailto link tracker, a script that tracks consenting clicks on links that point to an email address e.g. `<a href="mailto:someone@somewhere.something">Email me</a>`.

This script is similar to the [external link tracker](https://github.com/alphagov/govuk_publishing_components/pull/2538) and the [download link tracker](https://github.com/alphagov/govuk_publishing_components/pull/2534) once jQuery has been removed, so maybe once all these PRs are complete some kind of consolidation could be done.

## Why

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
@@ -1,36 +1,36 @@
+//= require ../vendor/polyfills/closest.js
 ;(function (global) {
   'use strict'
 
-  var $ = global.jQuery
   var GOVUK = global.GOVUK || {}
 
   GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {}
   GOVUK.analyticsPlugins.mailtoLinkTracker = function () {
-    var mailtoLinkSelector = 'a[href^="mailto:"]'
+    document.querySelector('body').addEventListener('click', function (event) {
+      var element = event.target
+      if (element.tagName !== 'A') {
+        element = element.closest('a')
+      }
 
-    $('body').on('click', mailtoLinkSelector, trackClickEvent)
+      if (!element) {
+        return
+      }
 
-    function trackClickEvent (evt) {
-      var $link = getLinkFromEvent(evt)
+      var href = element.getAttribute('href')
+      if (href.substring(0, 7) === 'mailto:') {
+        trackClickEvent(element, href)
+      }
+    })
+
+    function trackClickEvent (element, href) {
       var options = { transport: 'beacon' }
-      var href = $link.attr('href')
-      var linkText = $.trim($link.text())
+      var linkText = element.textContent
 
       if (linkText) {
-        options.label = linkText
+        options.label = linkText.trim()
       }
 
       GOVUK.analytics.trackEvent('Mailto Link Clicked', href, options)
-    }
-
-    function getLinkFromEvent (evt) {
-      var $target = $(evt.target)
-
-      if (!$target.is('a')) {
-        $target = $target.parents('a')
-      }
-
-      return $target
     }
   }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
@@ -17,6 +17,11 @@
       }
 
       var href = element.getAttribute('href')
+
+      if (!href) {
+        return
+      }
+
       if (href.substring(0, 7) === 'mailto:') {
         trackClickEvent(element, href)
       }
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
```diff
@@ -1,37 +1,36 @@
...
-  var $ = global.jQuery
   var GOVUK = global.GOVUK || {}
 
   GOVUK.analyticsPlugins = GOVUK.analyticsPlugins || {}
   GOVUK.analyticsPlugins.mailtoLinkTracker = function () {
-    var mailtoLinkSelector = 'a[href^="mailto:"]'
+    document.querySelector('body').addEventListener('click', function (event) {
+      var element = event.target
+      if (element.tagName !== 'A') {
+        element = element.closest('a')
```
alex-ju, line 11: Would be want to explicitly call a `closest` polyfill? It seems like we could do this via `govuk-frontend` or `govuk_publishing_components`
```
//= require govuk/vendor/polyfills/Element/prototype/closest.js
```
or
```
//= require ../vendor/polyfills/closest.js
```

I'm aware we may already call this polyfill with the other similar scripts, but until we review the polyfilling approach, we might want to explicitly import them where needed.

alex-ju (approved):
One comment about polyfilling `closest`, otherwise good to be merged.

app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
andysellick: Good question. Turns out that the reason this polyfill was added was because the govuk-frontend one [doesn't support IE8](https://github.com/alphagov/govuk_publishing_components/pull/1645#discussion_r470529237) (according to my past self). I know we don't support IE8 but I think at the time I made the decision on the basis that we still have some IE8 users and I don't want the JS to error for them if we can help it.

alex-ju: Ah, yes, I remember why we have the other polyfill, question is: do we want to add it to this script as well?

andysellick: Sorry - I'd completely misunderstood your question, I thought you were recommending we use the govuk-frontend polyfill over the one in the gem, but you were actually suggesting we just add a polyfill (I thought I'd added one already). Apologies, I've added a polyfill now (and updated the changelog).

andysellick: @alex-ju in updating the changelog and rebasing to master I found a new error caused indirectly by recent changes to the metadata component. I've added the change in a separate commit if you want to check before I merge. Including @danacotoran for reference.

spec/javascripts/components/metadata-spec.js
```diff
@@ -11,7 +11,15 @@ describe('The metadata component', function () {
 
   describe('when the "See all updates" link is clicked', function () {
     it('clicks the target if target exists and is not already expanded', function () {
-      var FIXTURE = '<div data-module="metadata"><a href="#toggle-me" class="js-see-all-updates-link"></a></div><div id="toggle-me" data-module="gem-toggle"><a data-controls="target" data-expanded aria-expanded="false"></a><div id="target" class="js-hidden">Target</div></div>'
+      console.log('BEGIN TEST')
```
alex-ju, line 14: I guess we can now get rid of this...

spec/javascripts/components/metadata-spec.js
```diff
@@ -21,6 +29,7 @@ describe('The metadata component', function () {
       var trigger = $(element).find('.js-see-all-updates-link')
       trigger[0].click()
       expect(GOVUK.Modules.GemToggle.prototype.toggleOnClick).toHaveBeenCalled()
+      console.log('END TEST')
```
alex-ju, line 32: ...and this

app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
```diff
@@ -17,6 +17,11 @@
       }
 
       var href = element.getAttribute('href')
+
+      if (!href) {
```
alex-ju, line 21: Do we also want to add a test in `mailto-link-tracker.spec.js` with an anchor without an href, to make sure we check this conditional statement is preserved over time.

spec/javascripts/components/metadata-spec.js
andysellick: 🤦  sorry, removed.

spec/javascripts/components/metadata-spec.js
andysellick: 🤦  sorry, removed.

app/assets/javascripts/govuk_publishing_components/analytics/mailto-link-tracker.js
andysellick: Good suggestion, have modified the initial test to check this.

PR #2552, opened 2022-01-06
Tweak metadata component "See all updates" interaction

danacotoran:
On some publications, the metadata component has a "See all updates" link. This is an in-page link which jumps to the list of updates at the bottom of the page like so: 

https://user-images.githubusercontent.com/7116819/148392637-026ed708-bd57-427c-9ebc-91497b2390e9.mov

This is on purpose in order to get around having the list of update notes at the top of the page. The reason for this is that the list of notes can be quite long and so it would not be ideal to have this list before the main content of the publication page as even when hidden behind a visibility toggle, once it's been expanded such a list could take a lot of scrolling to get past and inconvenience some users.

There was some feedback in UR (this was UR for a tangential bit functionality on a publication page) that suggested that people found it confusing that an additional click is required to show the list of updates (i.e after clicking "See all updates" at the top, one must click "see all updates" again at the bottom; the expected behaviour is that the list of update notes should be made visible straight away after the click at the top).

This is a workaround in order to achieve the intended outcome, which adds a simple script for the metadata component which fires a click event on the toggle trigger when the metadata "See all updates" link is clicked.

https://user-images.githubusercontent.com/7116819/148392972-80e91816-eff4-49fa-9e57-1e53068f79e2.mov

---- 

https://trello.com/c/XwSjaTET

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Tweak metadata component "See all updates" interaction ([PR #2552](https://github.com/alphagov/govuk_publishing_components/pull/2552))
+
 ## 28.0.0
 
 * **BREAKING:** Update feedback component to fix accessibility issues ([PR #2435](https://github.com/alphagov/govuk_publishing_components/pull/2435))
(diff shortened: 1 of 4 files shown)


owenatgov (commented):
Needs a changelog update, otherwise this looks great 💪

danacotoran:
Thanks @owenatgov I knew I was forgetting _something_

owenatgov (approved):
👀 ⬆️

PR #2554, opened 2022-01-07
Fix a bug in the scroll tracker

andysellick:
## What / why
Fixes the following problem with the scroll tracker.

- even though the tests pass, when run on an actual page the percentage tracking for 100% sometimes wasn't firing
- I think this was due to a rounding problem when the page height was a fraction rather than a whole number, the position of 100% was always slightly below what could be actually scrolled to
- the solution is to subtract one pixel from each calculated percentage position, which ensures that the page can be scrolled down far enough that the 100% percentage scroll track event can fire

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


danacotoran (approved):
📜 🛤️

PR #2556, opened 2022-01-07
Remove almost the last of the miscellaneous jQuery

andysellick:
## What
Remove the remaining bits of jQuery from the components gem. Actually, this doesn't quite remove all the jQuery as `modules.js` still requires it as it expects either a JS node or a jQuery object to be passed to it, but apart from that, I think this now removes all the jQuery.

## Why

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove almost the last of the miscellaneous jQuery ([PR #2556](https://github.com/alphagov/govuk_publishing_components/pull/2556))
 * Make metadata component "See all updates" link href less generic ([PR #2562](https://github.com/alphagov/govuk_publishing_components/pull/2562))
 * Update feedback component to add "display none" to "maybe" (spam prevention) button ([PR #2568](https://github.com/alphagov/govuk_publishing_components/pull/2568))
 * Apply margin bottom helper to big number component ([PR #2566](https://github.com/alphagov/govuk_publishing_components/pull/2566))
--- a/app/assets/javascripts/govuk_publishing_components/analytics/auto-track-event.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/auto-track-event.js
@@ -1,30 +1,30 @@
-;(function (global) {
-  'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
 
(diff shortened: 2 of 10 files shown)


alex-ju:
Due to issues created by a jasmine gem bump you will have to rebase this to master. Sorry for the inconvenience! 🤖

app/assets/javascripts/govuk_publishing_components/dependencies.js
```diff
@@ -5,8 +5,9 @@
 // = require jquery/dist/jquery
 // = require ./modules.js
 
-$(document).ready(function () {
+document.onreadystatechange = function () {
```
alex-ju, line 6: While this is technically correct, I wonder whether it's something worth doing since we import jQuery on L1 and the modules system on L2 requires jQuery. I'm thinking we could potentially leave `dependencies.js` as is, for backwards compatibility and offering applications who still require jQuery a mean of using it from the gem. Then for the applications that no longer require jQuery, we can gradually replace the `dependencies.js` with only importing a modules system that is not based on jQuery. This would, in theory, allow a transition rather than requiring a big bang approach. Aware we have static dealing with `dependencies.js` at the moment, which may require some additional parameter being passed via slimmer. Thoughts welcome.

alex-ju: The commit message on this (Remove jQuery from modules initialisation) may also be a bit misleading, as I immediately thought of `modules.js`

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove the last of the miscellaneous jQuery ([PR #2556](https://github.com/alphagov/govuk_publishing_components/pull/2556))
```
alex-ju, line 12: Not really _last_ is it? Because we still have jQuery in `modules.js`

alex-ju (approved):
This looks great, Andy! One comment in regard to `dependencies.js` which is perhaps more about the wider plan of removing jQuery gradually.

app/assets/javascripts/govuk_publishing_components/dependencies.js
andysellick: Yeah, that's sensible. I've removed this commit.

CHANGELOG.md
andysellick: True, have amended. I think I went into this PR more optimistic than I've come out of it 😞

Issue #2479, opened 2021-11-26
Add margin bottom option to big number component


jon-kirwan:
The **Big number** component currently doesn't have an option to change `margin-bottom` to override the default settings so it's necessary to either directly target the element e.g.

```
.gem-c-big-number {
  margin-bottom: govuk-spacing(6);
}
```

or add a wrapper class to the parent element of the component markup e.g.

```
.homepage-big-number__wrapper {
  &:nth-child(1) {
    margin-bottom: govuk-spacing(6);
  }
}
```
```
<div class="homepage-big-number__wrapper">
  <%= render "govuk_publishing_components/components/big_number", {
    number: t("homepage.index.ministerial_departments_count"),
  } %>
</div>
```
Or add a spacing override class on the parent element
```
<div class="govuk-!-margin-bottom-6">
  <%= render "govuk_publishing_components/components/big_number", {
    number: t("homepage.index.ministerial_departments_count"),
  } %>
</div>
```

PR #2566, opened 2022-01-10
Apply margin bottom helper to big number component

owenatgov:
## What
Applies the shared helper, which includes standardised `margin_bottom` behaviour, to [the big number component](https://components.publishing.service.gov.uk/component-guide/big_number).

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/2479

[Card](https://trello.com/c/u0DGQWQY/724-add-margin-bottom-to-big-number-component-and-apply-to-the-homepage)

### New docs section
Screenshot 2022-01-10 at 17 34 23

### DIFF
--- a/spec/components/big_number_spec.rb
+++ b/spec/components/big_number_spec.rb
@@ -81,6 +81,15 @@ describe "Big number", type: :view do
     assert_select ".gem-c-big-number__value[data-my-cool-attribute='cool']"
   end
 
+  it "sets a custom margin bottom if the margin_bottom attribute is present" do
+    render_component({
+      number: 500,
+      margin_bottom: 5,
+    })
+
+    assert_select '.gem-c-big-number.govuk-\!-margin-bottom-5'
+  end
+
   # The space mentioned in the below test is to handle screen readers printing dictations without a space between the number and the label
   # We don't want this to get removed accidentally, hence the following test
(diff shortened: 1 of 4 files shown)


owenatgov:
Just spotted that this needs a test for the new attribute. I initially thought that this wouldn't need one as this relies on an external helper which as its own tests but other components using the helper add tests for `margin_bottom`. Will follow convention here.

andysellick:
@owenatgov good shout. Yes, the tests don't need to be exhaustive because as you pointed out this functionality already has tests, but a basic test to check that the margin helper has been implemented correctly on this component would be good 👍

spec/components/big_number_spec.rb
```diff
@@ -81,6 +81,15 @@ def component_name
     assert_select ".gem-c-big-number__value[data-my-cool-attribute='cool']"
   end
 
+  it "sets a custom margin bottm if the margin_bottom attribute is present" do
```
andysellick, line 84: Nitpick: typo: `bottm`.

andysellick (approved):
One tiny typo, but otherwise looks good 👍

owenatgov:
@andysellick Would you mind approving Percy? Once you do (if you're happy with it) I'll merge this.

PR #2568, opened 2022-01-12
Update feedback component to add "display none" to "maybe" button

jon-kirwan:
TBC


maxgds (dismissed):
Because this means that users not using CSS will see it, although we don't anticipate them clicking it, can we shift it into the list below so it appears the same as the other buttons visually?

maxgds:
>  (and it also still includes the `hidden` attribute). I've double checked to be certain and looks like it's okay:

I missed that this is still there - I think we need to remove it as I think that's potentially what's hiding it from bots and letting spam through. Sadly this is only a guess.

maxgds:
> > (and it also still includes the `hidden` attribute). I've double checked to be certain and looks like it's okay:
> 
> I missed that this is still there - I think we need to remove it as I think that's potentially what's hiding it from bots and letting spam through. Sadly this is only a guess.

Also, I think on a lynx-like browser there is no CSS, inline or otherwise so users would see it.

maxgds (approved):
I'm not entirely sure this will help, or that the spam increase is definitely linked to our changes, but it's a good first step to see if it eliminates the problem.

KludgeKML:
It's anecdotal I know, but if it helps - I've put in a honeypot (an input field rather than a button, but same principle) into an open enquiries form before, and it definitely cut down spam volume on that. I think I used a gem to handle it, which added some boilerplate for anyone who could accidentally see it (but the gem may be too heavyweight to use here). I also couldn't comment on the accessibility implications.

An example of the code it made.

`<div class="xstvlqrdoacmi_1641984611">
   <style media="screen">.xstvlqrdoacmi_1641984611 {display:none;}</style>
   <label for="xstvlqrdoacmi">If you are a human, ignore this field</label>
   <input type="text" name="xstvlqrdoacmi" id="xstvlqrdoacmi" autocomplete="off" tabindex="-1">
</div>`

(Sorry if this has been covered before).

jon-kirwan:
Thanks @KludgeKML - good to know (likewise, whist looking in to this issue I also came across a few examples where the honeypot had been setup with an input field). Hoping this fixes the problem and / or it's only temporary but will keep this in mind if we do need to revisit 👍

maxgds:
I recognise that pattern having seen it elsewhere on the web, @KludgeKML  but it doesn't feel very GDSy to put a visible label saying "don't fill this in" - not that the "maybe" button is that far off it. In this case the solution we had seemed to be working previously so on the off chance that we have broken it and it's not coincidental it's worth reinstating it as-was.

PR #2570, opened 2022-01-14
Add check to big number to convert plus suffixes to subscript elements

owenatgov:
## What
Adds a small string-processing feature to the [big number component](https://components.publishing.service.gov.uk/component-guide/big_number). If a value passed to a big number is a string and ends with a plus symbol, strip it out and re-amend it to the end of the original value as a styled `span`.

## Why
This allows us to safely amend the positioning of the plus symbol so that it sits at the baseline of the big number value, adding to govuk's visual harmony. This comes as a recommendation from one of govuk's lead designers.

This solution has been tested extensively against the 3 major screen readers that govuk supports and our supported browsers.

[Card](https://trello.com/c/ts8lksM8/735-add-a-flag-to-bottom-align-the-in-big-numbers)

### Before
Screenshot 2022-01-14 at 15 50 01

### After
Screenshot 2022-01-14 at 15 50 19


mia-allers-gds:
@owenatgov looks amazing!

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -17,6 +17,18 @@
 
 %>
 <% if number %>
+  <% # We amend the capture here if it ends with a "+" to account for small symbols awkwardly sitting above the baseline in our typeface and amend them to sit at the baseline for the purpose of visual harmony %>
+  <% # Currently the only small symbol used with big numbers are pluses so they are is specifically targetted in the below code %>
+  <% if number.is_a? String
```
chao-xian, line 22: Wouldn't `<% ... %>` with the closing percent angle bracket not still work?

chao-xian (commented):
I have a code style question...

app/views/govuk_publishing_components/components/_big_number.html.erb
owenatgov: You are right. I was caught up in how my editor handles individual comment lines in `erb` files. I'll update this.

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -16,7 +16,19 @@
   end
 
 %>
-<% if number %>
+<% if number
+  # We amend the capture here if it ends with a "+" to account for small symbols awkwardly sitting above the baseline in our typeface and amend them to sit at the baseline for the purpose of visual harmony
```
chao-xian, line 20: Could we abstract this into a helper, so we don't have long chunk of logic here? See [example for Intervention Banner](https://github.com/alphagov/govuk_publishing_components/blob/master/lib/govuk_publishing_components/presenters/intervention_helper.rb)

chao-xian (changes requested):

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -1,25 +1,21 @@
...
-    value_classes << "gem-c-big-number__value--decorated"
-  end
-
+  value_classes = big_number_helper.value_classes
 %>
 <% if number %>
   <% big_number_value = capture do %>
     <%= tag.span class: value_classes, data: href ? nil : data_attributes do %>
-      <%= number %>
+      <%= raw(big_number_helper.formatted_number_string) %>
```
chao-xian, line 18: Rather than raw-ing this, can we set the HTML entity for plus in the helper instead?

lib/govuk_publishing_components/presenters/big_number_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+          class_list << "gem-c-big-number__value--decorated"
+        end
+
+        class_list
+      end
+
+      # We amend the number here if it ends with a "+" to account for small symbols awkwardly sitting above the baseline in our typeface and adjust them to sit at the baseline for the purpose of visual harmony
+      # Currently the only small symbol used with big numbers are pluses so they are specifically targetted in the below code
+      def formatted_number_string
+        if @number
```
chao-xian, line 23: Wondering whether if checking if this exists could be removed for setting a default instead of an empty string in the initializer instead?

app/views/govuk_publishing_components/components/_big_number.html.erb
owenatgov: I'm very uncomfortable with this `raw` but I wasn't sure how else to do it so I'm glad you've got an alternative idea. Could you please ELI5 your proposal?

lib/govuk_publishing_components/presenters/big_number_helper.rb
owenatgov: Also ELI5 this please

app/views/govuk_publishing_components/components/_big_number.html.erb
chao-xian: `&plus;` is the entity you want output. If you use the `.sanitize` method instead it'll do it for you: https://api.rubyonrails.org/classes/ActionView/Helpers/SanitizeHelper.html

lib/govuk_publishing_components/presenters/big_number_helper.rb
chao-xian: sorry for confusing you with my instead instead

app/views/govuk_publishing_components/components/_big_number.html.erb
owenatgov: After revisiting this a bit I've changed how it works so now it returns explicit data instead of html. It probably makes more sense not to handle rendering within the helper. The logic in the template is now a lot lighter than it was before I moved this to a helper. Let me know what you reckon.

app/views/govuk_publishing_components/components/_big_number.html.erb
```diff
@@ -1,25 +1,29 @@
...
 <% if number %>
   <% big_number_value = capture do %>
     <%= tag.span class: value_classes, data: href ? nil : data_attributes do %>
-      <%= number %>
+      <%
+        # The below check is to account for small symbols awkwardly sitting above the baseline in our typeface and adjust them to sit at the baseline for the purpose of visual harmony
+        # Currently the only small symbol used with big numbers are pluses so they are specifically applied in the below code
+      %>
+      <% if big_number_helper.number_has_plus_suffix? %>
+        <%= number[0...-1] %><%= tag.span raw("&plus;"), class: "gem-c-big-number__suffix" %>
```
chao-xian, line 23: At this point I think we might as well just go with `+` - I'm sorry I hadn't realised using entities with Rails was so tricky.

chao-xian (approved):
Just the plus thing to revert but looks good - thanks for doing the work on this!

app/views/govuk_publishing_components/components/_big_number.html.erb
owenatgov: No probs, it was worth a go for tidiness.

PR #2574, opened 2022-01-20
Update feedback component to resolve spam problem

jon-kirwan:
## What
https://trello.com/c/I0WfBF4P/1011-update-feedback-component-to-resolve-spam-problem

There has been an increase in spam coming from the feedback component from the 6th of Jan, which seems to coincide with the deployment of the new feedback component. https://govuk.zendesk.com/agent/tickets/4834840

**Solution - multiple changes**

- Changed the _Maybe_ button to be a link. Bots that click 'Maybe' will be redirected elsewhere (to the contact page)
- Replaced HTML comments with Ruby comments e.g. `<% # Maybe button exists only to try and capture clicks by bots %>` which may make it harder for more advanced bots to identify elements added for spam bots only
- Added fake (honeypot) input field
  - The input field includes a pattern attribute that will fail (and generate HTML validation) for anything other than an empty string 
  - Added inline JavaScript to check on form submission if the honeypot field has been filled in and to prevent submission 
- I’m not sure if we should also have some server side protection too? (to be investigated separately https://github.com/alphagov/feedback)
- For non CSS users, the input is labelled as **"This field is for robots only. Please leave blank"**

## Why

It seems likely that by resolving a number of accessibility issues (https://trello.com/c/MOshcFm1/771-update-feedback-component-to-hide-and-show-content-in-a-more-robust-way) we have inadvertently made it easier for spam bots to access the feedback component forms.

See:

- https://github.com/alphagov/govuk_publishing_components/pull/2435
- https://github.com/alphagov/govuk_publishing_components/pull/2568

### DIFF
--- a/app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
+++ b/app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
@@ -23,6 +23,12 @@
       <h3 class="gem-c-feedback__form-heading"><%= t("components.feedback.help_us_improve_govuk") %></h3>
       <p id="feedback_explanation" class="gem-c-feedback__form-paragraph"><%= t("components.feedback.dont_include_personal_info") %></p>
 
+      <% # Added for spam bots only %>
+      <div class="govuk-visually-hidden" aria-hidden="true">
+        <label for="feedback_maybe">This field is for robots only. Please leave blank</label>
+        <input id="feedback_maybe" type="text" pattern=".{0}" tabindex="-1" autocomplete="off" >
+      </div>
+
       <%= render "govuk_publishing_components/components/textarea", {
         label: {
           text: t("components.feedback.what_doing")
@@ -46,3 +52,23 @@
     </div>
   </div>
 </form>
+
+<%
+  # I've added the following script inline in case of a scenario where a bot is able to parse the page, 
+  # without downloading any of the external scripts. 
+  # This seems to be a more reliable way to make sure the script is executed.
+%>
+
+<script>
+  document.addEventListener("DOMContentLoaded", function () {
+    var input = document.querySelector("#feedback_maybe"),
+      form = document.querySelector("#something-is-wrong")
+
+    form.addEventListener("submit", spamCapture);
+
+    function spamCapture(e) {
+      if (input.value.length !== 0) return;
+      e.preventDefault();
+    }
+  });
+</script>
--- a/app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
@@ -1,17 +1,23 @@
-<%
-  contact_govuk_path = "/contact/govuk"
-%>
-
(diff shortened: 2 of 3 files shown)


maxgds:
> I’m not sure if we should also have some server side protection too? 

I think this will be necessary unfortunately, I expect bots are hitting a no-js page sometimes. I'd hope it would be as simple as wrapping the submission in an "if robot-input-is-blank" and just presenting the thanks message regardless.

app/views/govuk_publishing_components/components/feedback/_yes_no_banner.html.erb
```diff
@@ -1,17 +1,24 @@
...
 <div class="gem-c-feedback__prompt gem-c-feedback__js-show js-prompt" tabindex="-1">
   <div class="gem-c-feedback__prompt-questions js-prompt-questions" hidden>
     <h2 class="gem-c-feedback__prompt-question"><%= t("components.feedback.is_this_page_useful") %></h2>
     
     <ul class="gem-c-feedback__option-list">
-      <li class="gem-c-feedback__option-list-item" style="display: none" hidden="hidden">
-        <!-- Maybe button exists only to try and capture clicks by bots -->
-        <button data-track-category="yesNoFeedbackForm" data-track-action="ffMaybeClick" aria-expanded="false" style="display: none" hidden="hidden" aria-hidden="true">
+      <li class="gem-c-feedback__option-list-item govuk-visually-hidden" style="display: none" hidden>
+        <% # Maybe button exists only to try and capture clicks by bots %>
```
maxgds, line 7: This whole trap is a bit confusing to me as I've never been clear on why we expect bots to target this button rather than another... but that aside, why do you think turning it into a link will aid here? I'm also aware that it lands you on a page that contains a form - do we risk sending the nasty traffic away from our form in favour of spamming this one instead?

app/views/govuk_publishing_components/components/feedback/_problem_form.html.erb
```diff
@@ -46,3 +52,17 @@
     </div>
   </div>
 </form>
+
+<script>
```
maxgds, line 56: What's the reason for inlining the js rather than adding this to the existing js for the component?
https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/feedback.js

jon-kirwan: I've added it inline in case of a scenario where a bot is able to parse the page, without downloading any of the external scripts. This way seemed to be a more reliable way to make sure the script is executed (but otherwise i'd definitely prefer to include it in the external script).

maxgds: Seems like a good reason! Could you put a comment to that effect above it for future devs who wonder?

jon-kirwan: 👍 updated

maxgds (approved):
I'm hopeful that this will help resolve the spam issue even though the maybe button seems like a dark art. We may need to do more to handle it server side in the Feedback app, but for now this is good to go and we should be able to see the impact (or lack) quite quickly on release and integration into frontend apps.

PR #2575, opened 2022-01-20
Add pipes between buttons on super nav header

owenatgov:
## What
A couple of changes to the super nav header:

- Implement `spans` within the desktop buttons to create a visual pipe effect between buttons on desktop, like we have on mobile
- Change the colour of pipes to the design system `mid-grey`
- nudge the chevrons ever so slightly away from their respective buttons

## Why
This is to solve a problem where the desktop buttons appear to have a big gap between them when it's simply the width of the buttons which we need to maintain for the purpose of focus and open states. Solution produced alongside a designer.

[Card](https://trello.com/c/GHzPAVUT/725-desktop-padding-has-left-the-spacing-on-topics-and-gov-activity-inconsistent-timebox)

### Before
Screenshot 2022-01-20 at 17 22 13

### After
Screenshot 2022-01-21 at 16 54 56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@
 * Fix a bug in the scroll tracker ([PR #2554](https://github.com/alphagov/govuk_publishing_components/pull/2554))
 * Add check to big number to convert plus suffixes to subscript elements ([PR #2570](https://github.com/alphagov/govuk_publishing_components/pull/2570))
 * Update feedback component to resolve spam problem ([PR #2574](https://github.com/alphagov/govuk_publishing_components/pull/2574))
+=======
+* Add pipes between buttons on super nav header ([PR #2575](https://github.com/alphagov/govuk_publishing_components/pull/2575))
 
 ## 28.2.0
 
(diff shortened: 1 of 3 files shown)


mia-allers-gds:
@owenatgov these look super, super neat. Can I be really cheeky and ask for us to drop the pipes down to a darker grey? I know this isn't in the style guide but I like #767879

Screenshot 2022-01-20 at 17 33 45

Let me know your thoughts

chao-xian (approved):
Other than the colour change @mia-allers-gds has asked for, this LGTM

owenatgov:
@mia-allers-gds I feel uncomfortable adding "custom" colour codes into the gem as it could trip up any future devs trying to debug this if we ever want/need to change it and generally is another entity we have to maintain on top of existing stuff from the design system. I can see we have a couple of custom colours used in the gem so adding it wouldn't be a total showstopper but they're the exception rather than the rule so I'd like to avoid it if the alternatives don't look rubbish.

I imagine you've already looked at this but how do you feel about the design system dark grey? Here it is for reference:

Screenshot 2022-01-21 at 09 42 32

Open to continuing the negotiation if this solution is a bother.

owenatgov:
Myself and @mia-allers-gds chatted about this on slack and have found a suitable alternative approach.

PR #2581, opened 2022-01-24
Update accordion gem

chris-gds:


nnagewad:
Saw in the preview link. Looks good bud!

app/assets/javascripts/govuk_publishing_components/components/button.js
```diff
@@ -2,4 +2,4 @@
 // = require govuk/components/button/button.js
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {}
-window.GOVUK.Modules.GovukButton = window.GOVUKFrontend
+window.GOVUK.Modules.GovukButton = window.GOVUKFrontend.Button
```
andysellick, line 5: Thanks for the link that explains this change, but could you add it to the commit message as well to summarise? e.g. `GOVUK-frontend has changed the naming of components JavaScript modules so individual imports are now attached to window.GOVUKFrontend.[ComponentName] instead of window.GOVUKFrontend.`

app/assets/javascripts/govuk_publishing_components/components/skip-link.js
```diff
@@ -0,0 +1,5 @@
+// This component relies on JavaScript from GOV.UK Frontend
+// = require govuk/components/skip-link/skip-link.js
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.GovukSkipLink = window.GOVUKFrontend.SkipLink
```
andysellick, line 5: This looks like a new JS file for the skip link component. This needs to be highlighted for deployment to those applications that use this component as they'll need to include it. I don't think it technically counts as a breaking change, but maybe we should highlight it as that to ensure this gets done?

Applications using skip link are `collections-publisher, content-data-admin, content-publisher, datagovuk_find, finder-frontend, government-frontend, govuk-coronavirus-vulnerable-people-form, release, search-admin, signon, static`. Might be worth also checking this setup, I'm not surprised it's in `static` but a little unsure why it's in some of the others e.g. `finder-frontend`.

app/views/govuk_publishing_components/components/_checkboxes.html.erb
```diff
@@ -19,7 +19,7 @@
         <% end %>
 
         <% if cb_helper.error %>
-          <%= tag.span error, id: "#{id}-error", class: "govuk-error-message" %>
+          <%= tag.p error, id: "#{id}-error", class: "govuk-error-message" %>
```
andysellick, line 22: The link in the commit message is slightly wrong, should be `https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#update-the-html-for-error-messages`

spec/components/error_message_spec.rb
```diff
@@ -8,13 +8,13 @@ def component_name
   it "renders error message" do
     render_component(text: "Please enter your National Insurance number")
 
-    assert_select(".govuk-error-message", text: "Error: Please enter your National Insurance number")
+    assert_select("p.govuk-error-message", text: "Error: Please enter your National Insurance number")
```
andysellick, line 11: I see why you've modified this test but I don't think we need to specifically test the correct HTML element is used here - otherwise it would have previously looked for a `span`.

CHANGELOG.md
```diff
@@ -25,6 +25,7 @@
 
 ## 28.1.0
 
+* Bump `govuk-frontend` from 3.14.0 to 4.0.0 ([PR #2355](https://github.com/alphagov/govuk_publishing_components/pull/2533))
```
andysellick, line 28: This changelog entry doesn't match the PR. It should also appear beneath an `Unreleased` heading.

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -10,21 +10,235 @@
...
-        content: applications,
-      },
-      {
-        id: "components-gem",
-        label: "Components",
-        content: components,
-      }
-    ]
-  } %>
+        <div class="govuk-accordion" data-module="govuk-accordion" id="accordion-with-summary-sections">
```
andysellick: This looks wrong... it's replacing some logic to display the tabs component with a hard coded accordion. I think we discussed this and this was an error relating to a merge?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -22,44 +22,37 @@
...
-
-  unique_locales = locales.values.uniq
-
-  if unique_locales.length > 1
-    data_attributes[:locale] = locales
-  else
-    if unique_locales[0] != I18n.locale
-      data_attributes[:locale] = unique_locales[0]
-    end
-  end
```
andysellick, line 41: Can you provide some explanation why this locales code is being removed?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -22,44 +22,37 @@
...
+  data_attributes[:module] = 'govuk-accordion'
 %>
 <% if items.any? %>
   <%= tag.div(class: accordion_classes, id: id, data: data_attributes) do %>
     <% items.each_with_index do |item, i| %>
       <%
+        # Nunjucks starts a loop on 1 and the client side JavaScript also
+        # adopts this behaviour. To prevent things from breaking, the index
+        # here also need to be increase by one. (Nunjucks is used by GOV.UK
+        # Frontend, which this component is based on.)
```
andysellick, line 49: Thanks for explaining this oddity although I found it a little confusing - I don't think it needs to mention Nunjucks. If my understanding is correct, would the following suit?

`The GOVUK Frontend JavaScript for this component uses a loop starting at 1 for accessing heading ID values, hence the increment below.`

app/views/govuk_publishing_components/audit/show.html.erb
```diff
@@ -10,235 +10,21 @@
...
-              <% end %>
-            </div>
-          </div>
-        </div>
-      <% else %>
-        <p class="govuk-body">No components found.</p>
-      <% end %>
-    </div>
-  </div>
+<% if @in_application or @other_applications %>
```
andysellick, line 13: I've been reviewing this commit by commit, and I now see this commit fixes the problem with this page that I've already commented on. Could this commit be squashed into that one?

app/assets/javascripts/govuk_publishing_components/vendor/polyfills/common.js
```diff
@@ -0,0 +1,8 @@
+function nodeListForEach (nodes, callback) {
```
andysellick, line 1: If this is still required can this commit be squashed into the commit that removed it?

Where is it used? When I saw you'd removed it I had a look to confirm and couldn't see any other places where it is called.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,85 @@
+/* global nodeListForEach */
+//  = require ../vendor/polyfills/common.js
```
andysellick, line 2: In an earlier commit the polyfill for `indexOf` was removed, but code containing this is now restored in this commit. Can this polyfill be added back in as well?

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,5 +1,85 @@
+/* global nodeListForEach */
```
andysellick, line 1: It's a little confusing that this code was removed and then restored in the same PR. Could this commit be squashed into an earlier commit?

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,330 +1,2 @@
 @import "mixins/prefixed-transform";
-
```
jon-kirwan, line 2: I think this is probably fine... but just going to add that some `webkit-` prefixes are missing (that are present in the design system). e.g.
This is not specific to the Accordion - it's the same for all components (only observed here)

```
.js-enabled .govuk-accordion-nav__chevron {
    -webkit-box-sizing: border-box;
    box-sizing: border-box;
    ....
```
and
```
.js-enabled .govuk-accordion__section-header:hover {
    border-top-color: #b1b4b6;
    -webkit-box-shadow: inset 0 3px 0 0 #1d70b8;
    box-shadow: inset 0 3px 0 0 #1d70b8;
}
```

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -39,27 +37,45 @@
...
 
-      <%= tag.section(class: section_classes) do %>
-        <div class="gem-c-accordion__section-header">
-          <%= content_tag(shared_helper.get_heading_level, class: 'gem-c-accordion__section-heading', id: item[:heading][:id]) do %>
-            <%= tag.span(item[:heading][:text], id: "#{id}-heading-#{index}", data: item[:data_attributes], class: 'gem-c-accordion__section-button') %>
-          <% end %>
+        summary_classes = %w(govuk-accordion__section-summary govuk-body)
+      %>
+      <%= tag.div(class: section_classes) do %>
+        <div class="govuk-accordion__section-header">
```
jon-kirwan, line 60: Just wondering if it's worth swapping the `div` markup for `<%= tag.div(class: govuk-accordion__section-header) do %>` to keep consistent with the use of `tag` helpers elsewhere in the view.

jon-kirwan (commented):
Hi Chris,

This seems okay to me - the only thing I could find visually is that the chevrons are missing in IE8 (screenshots attached below) - but otherwise looks fine and works as expected.

There are some console errors logged for IE8 / 9 (similar to [here](https://github.com/alphagov/govuk_publishing_components/pull/2601#discussion_r797421805)). However, they seem to get logged for the home page of the gem and other components (e.g. action_link) so I don't it's related to the Accordion.

Otherwise, I've added two comments...

app/views/govuk_publishing_components/components/_accordion.html.erb
chris-gds: > This seems okay to me - the only thing I could find visually is that the chevrons are missing in IE8 (screenshots attached below) - but otherwise looks fine and works as expected.

Thanks for this, something most not be porting quite right as IE8 fallback should display an icon, let me review.

app/assets/javascripts/govuk_publishing_components/components/skip-link.js
chris-gds: So, [#2533](https://github.com/alphagov/govuk_publishing_components/pull/2533) to upgrade to `govuk-frontend 4.0` was dependant on completing before this could start - this `skiplink` is definitely part of it. 

Rather than block this piece of work I just pulled these commits in temporarily.   I plan to drop these once [#2533](https://github.com/alphagov/govuk_publishing_components/pull/2533) is merged so this shouldn't be part of it.

It's partly why I've kept it in Draft as well so this doesn't get mixed in or merged without this happening.

app/assets/javascripts/govuk_publishing_components/components/button.js
chris-gds: I think this could apply to [#2533](https://github.com/alphagov/govuk_publishing_components/pull/2533) during review

app/views/govuk_publishing_components/components/_accordion.html.erb
andysellick: Personally I wouldn't worry about the icons on IE8 - I think the meaning of the control is conveyed enough with the text.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
chris-gds: Hmm it's a good spot, I would expect the same prefixes to pull through. I'm wondering if this is automated around support and since GOVUK supports 2 back `-webkit-` is decided it's not needed.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: It's been removed as it's handled at the `govuk-frontend` side

app/views/govuk_publishing_components/components/_accordion.html.erb
chris-gds: +1 none issue

After some investigation it appears it's (unexpectedly) coming from the other end. 
https://github.com/alphagov/govuk-frontend/issues/2530

chris-gds:
@jon-kirwan did you notice any issues with manuals?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -22,7 +20,7 @@
   locales = {}
 
   data_attributes ||= {}
-  data_attributes[:module] = 'gem-accordion'
+  data_attributes[:module] = 'govuk-accordion gem-c-accordion'
   data_attributes[:anchor_navigation] = anchor_navigation
```
jon-kirwan, line 24: I think i've found an issue in Manuals Frontend in that the anchor links do not seem to expand the accordion sections that they are linked to. 

I think the issue is down to the naming convention used here `data_attributes[:module] = 'govuk-accordion gem-c-accordion'`. I think the module expects the format to be `gem-accordion`. As far as I can tell that's how it's initialising the component. 

```
      // eg selectable-table to SelectableTable
      function camelCaseAndCapitalise (string) { // gem-accordion
        return capitaliseFirstLetter(camelCase(string)) // GemAccordion
      }
```

jon-kirwan (changes requested):

app/views/govuk_publishing_components/components/_accordion.html.erb
chris-gds: This makes sense, I updated from `gem-accordion` to `gem-c-accordion` during this last wave of changes to match convention but of course, needs  `gem-accordion`

I'm not a fan of this mismatch, it's confusing, let me see if I can align.  Thank @jon-kirwan

jon-kirwan:
Manuals now seems fine for me. I've checked the anchor accordions examples and everything now seems to work correctly.

One further tiny issue. There is a typo in the 'What' section above in "Update the `govuk_publishing_components` _Accoridon_ to"...

chris-gds:
> Manuals now seems fine for me. I've checked the anchor accordions examples and everything now seems to work correctly.

Ace! Just waiting on 4.0 to be merged in (typo fixed)

PR #2582, opened 2022-01-24
[DO NOT MERGE] Remove vaccination-status from Google Analytics

gclssvglx:
## What

Remove the users response to the `vaccination-status` question from the `location` (URL) field in the [`check-travel-during-coronavirus` Smart Answer flow](https://github.com/alphagov/smart-answers/pull/5724) that we currently send to Google Analytics.

Linked - https://github.com/alphagov/smart-answers/pull/5736

## Why

A users vaccination status is considered health data, so we should remove it from being exposed outside of the application.

## Visual Changes
None

[Trello](https://trello.com/c/DInfO64C/540-mvp-innout-decide-what-to-do-from-a-data-privacy-angle)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/pii.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/pii.js
@@ -12,6 +12,9 @@
   var UNLOCK_TOKEN_PATTERN = /unlock_token=[a-zA-Z0-9-]+/g
   var STATE_PATTERN = /state=.[^&]+/g
 
+  // specific URL parameters to be redacted from smart answer URLs
+  var VACCINATION_STATUS_PATTERN = /vaccination_status=.[^&]+/g
+
   function shouldStripDates () {
     var metas = document.querySelectorAll('meta[name="govuk:static-analytics:strip-dates"]')
     return metas.length > 0
@@ -19,6 +22,11 @@
 
   function shouldStripPostcodes () {
     var metas = document.querySelectorAll('meta[name="govuk:static-analytics:strip-postcodes"]')
+    return metas.length > 0
+  }
+
+  function shouldStripVaccinationStatus () {
+    var metas = document.querySelectorAll('meta[name="govuk:static-analytics:strip-vaccination-status"]')
     return metas.length > 0
   }
 
@@ -43,6 +51,7 @@
   var pii = function () {
     this.stripDatePII = shouldStripDates()
     this.stripPostcodePII = shouldStripPostcodes()
+    this.stripVaccinationStatusPII = shouldStripVaccinationStatus()
     this.queryStringParametersToStrip = queryStringParametersToStrip()
   }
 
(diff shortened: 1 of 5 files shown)


spec/javascripts/govuk_publishing_components/analytics/google-analytics-universal-tracker.spec.js
```diff
@@ -350,4 +354,8 @@ describe('GOVUK.GoogleAnalyticsUniversalTracker', function () {
   function pageWantsPostcodesStripped () {
     $('head').append('<meta name="govuk:static-analytics:strip-postcodes" value="does not matter" />')
   }
+
+  function pageWantsVaccinationStatusStripped () {
+    $('head').append('<meta name="govuk:static-analytics:strip-vaccination-status" value="does not matter" />')
```
andysellick, line 359: I get from the test that one possible status for this is `fully-vaccinated` - just out of curiousity, what are the other possible statuses?

app/assets/javascripts/govuk_publishing_components/analytics/pii.js
```diff
@@ -12,6 +12,9 @@
   var UNLOCK_TOKEN_PATTERN = /unlock_token=[a-zA-Z0-9-]+/g
   var STATE_PATTERN = /state=.[^&]+/g
 
+  // specific URL parameters to be redacted from smart answer URLs
+  var VACCINATION_STATUS_PATTERN = /vaccination_status=.[^&]+/g
```
andysellick, line 16: I'm not very regex knowledgeable but I wanted to understand what this does so tried it out on regxr.com. Is it possible that this matches too much if the `vaccination_status` field is present but blank? I don't know if that would ever be possible though.

Screenshot 2022-01-26 at 16 03 18

andysellick (commented):
I've had a look through the JS, seems mostly sound, but I can't review the other stuff with any degree of confidence.

leenagupte:
Is it possible to create a generic `strip_pii` attribute, that accepts the parameter name, rather than hard-coding `vaccination_status`?

andysellick:
> Is it possible to create a generic `strip_pii` attribute, that accepts the parameter name, rather than hard-coding `vaccination_status`?

I've had a look through the code and I can't see anything that would do this currently (although I could be wrong). If there isn't something, it's definitely something that would be useful, I'll have a think.

leenagupte:
> > Is it possible to create a generic `strip_pii` attribute, that accepts the parameter name, rather than hard-coding `vaccination_status`?
> 
> I've had a look through the code and I can't see anything that would do this currently (although I could be wrong). If there isn't something, it's definitely something that would be useful, I'll have a think.

It looks like a function already exists to do this: https://github.com/alphagov/smart-answers/pull/5736#issuecomment-1022352154
TL;DR: https://github.com/alphagov/govuk_publishing_components/blob/dcaa0800ca66c920ccb195b796a67496c6d28e23/app/assets/javascripts/govuk_publishing_components/analytics/pii.js#L34

PR #2594, opened 2022-01-28
Add select component hint

andysellick:
## What
Adds a option for a hint to the select component.

- based on the Design system
- an optional hint attribute, with optional hint id
- the aria-describedby attribute of the select points to the hint, if an error is also present the describedby attribute includes both hint and error IDs

During this work I noticed that there's a lot of code in the template and some other refactoring that should be looked at (e.g. the helper file isn't named correctly - it's `select.rb` instead of `select_helper.rb`) but I think I'll do that separately to not over-complicate this PR.

## Why
Needed in `smart-answers`.

## Visual Changes
Screenshot 2022-01-28 at 15 06 59

Trello card: https://trello.com/c/P0YMGgDs/594-add-hint-to-select-component

### DIFF
--- a/app/views/govuk_publishing_components/components/_select.html.erb
+++ b/app/views/govuk_publishing_components/components/_select.html.erb
@@ -10,11 +10,22 @@
   heading_size = false unless shared_helper.valid_heading_size?(heading_size)
   error_message ||= nil
   error_id ||= nil
+  hint ||= nil
+  hint_id ||= nil
+
+  describedby = %w[]
 
   if error_message || error_id
     error_id = "error-#{SecureRandom.hex(4)}" unless error_id
-    aria_describedby = { describedby: error_id }
+    describedby << error_id
   end
+
+  if hint
+    hint_id = "hint-#{SecureRandom.hex(4)}" unless hint_id
+    describedby << hint_id
+  end
+
+  aria_describedby = { describedby: describedby } if describedby
 
   css_classes = %w(govuk-form-group gem-c-select)
   css_classes << "govuk-form-group--error" if error_message
@@ -35,6 +46,13 @@
       <%= tag.h1 label_tag(id, label, class: label_classes), class: "gem-c-title__text" %>
     <% else %>
       <%= label_tag(id, label, class: label_classes) %>
+    <% end %>
+
+    <% if hint %>
+      <%= render "govuk_publishing_components/components/hint", {
+        id: hint_id,
+        text: hint
+      } %>
     <% end %>
 
     <% if error_message %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_select.html.erb
```diff
@@ -37,6 +48,10 @@
       <%= label_tag(id, label, class: label_classes) %>
     <% end %>
 
+    <% if hint_text %>
+      <%= tag.span hint_text, id: hint_id, class: "govuk-hint" %>
```
kr8n3r, line 52: just to note that[ in govuk-frontend v3.8 `span` became `div`](https://github.com/alphagov/govuk-frontend/issues/1835) and in v4 styles which made the hint <span> display as block [were removed](https://github.com/alphagov/govuk-frontend/releases/tag/v4.0.0#:~:text=Update%20the%20HTML%20for%20hints)

migration can either be addressed here or in https://github.com/alphagov/govuk_publishing_components/pull/2533

alex-ju: The [hint component already uses a `div` element](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_hint.html.erb). However, we have two instances (character count and checkboxes) where the component is not being used and a `span` is still in place. I have updated #2533 to address these two special cases (until we update them to use the hint component). For this PR I suggest we use the hint component [as we do for most form components](https://github.com/alphagov/govuk_publishing_components/search?q=components%2Fhint).

andysellick: Thanks for the input both. Yes, agree that we should definitely use the hint component here, have changed 🤦

app/views/govuk_publishing_components/components/_select.html.erb
```diff
@@ -10,12 +10,23 @@
   heading_size = false unless shared_helper.valid_heading_size?(heading_size)
   error_message ||= nil
   error_id ||= nil
+  hint_text ||= nil
```
alex-ju, line 13: I seems like we use `hint` for the hint text, looking at other component's APIs. I would stick to `hint` for consistency, although I appreciate `hint_text` might be clearer.

alex-ju (approved):
Once comment related to the `hint` variable name we use in the component API, otherwise good to be merged.

app/views/govuk_publishing_components/components/_select.html.erb
kr8n3r: > The [hint component already uses a `div` element](https://github.com/alphagov/govuk_publishing_components/blob/master/app/views/govuk_publishing_components/components/_hint.html.erb). However, we have two instances (character count and checkboxes) where the component is not being used and a `span` is still in place. I have updated #2533 to address these two special cases (until we update them to use the hint component). For this PR I suggest we use the hint component [as we do for most form components](https://github.com/alphagov/govuk_publishing_components/search?q=components%2Fhint).

sorry din't see you folk have a hint component 👍

alex-ju: Thanks for the nudge, @kr8n3r – otherwise I wouldn't have found the two exceptions 😄

PR #2595, opened 2022-01-28
Document the triggerEvent code

andysellick:
## What
Add some documentation for `window.GOVUK.triggerEvent`.

## Why
Someone asked me how this worked recently and I couldn't remember, so I thought it might be worth writing a short description.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js
@@ -1,3 +1,4 @@
+// see https://github.com/alphagov/govuk_publishing_components/blob/main/docs/lib/trigger_event.md
 (function (root) {
   'use strict'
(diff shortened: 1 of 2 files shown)


alex-ju (commented):
A few initial thoughts on this:
 - Do we want to start documenting our JavaScript code (or at least what's in `/lib`) using a more standard approach (e.g. JSDoc)?
 - If we decide to move forward with having JavaScript code documentation as part of our main docs, do we want to find a way of pointing to these file from the code itself (in this case from `trigger-event.js`)?

chao-xian:
It's a good point seeing as we already have Rubydoc https://www.rubydoc.info/gems/govuk_publishing_components/34.10.1

MartinJJones (approved):
Could be worth adding a change log entry, other than that, the changes look good to me 👍

PR #2601, opened 2022-02-01
Update step-by-step

chris-gds:
## What

Align Design of current step-by-step to match GOVUK Design System Accordion

## Why

Align architecture, meet WCAG.

### Before > After 

step-by-step

## Anything else

Draft Design review


maxgds:
The initial state doesn't look right to me. It's closed, with a "show all steps" link, but the icon is pointing upwards as it usually does for the hide all wording, eg:
image

Could you take a look?

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -63,15 +63,116 @@ $top-border: solid 1px govuk-colour("mid-grey", $legacy: "grey-3");
   padding: 3px 3px 0 0;
 }
 
+.gem-c-step-nav____title-text-focus {
+  margin-bottom: 13px;
+  display: inline-block;
+}
+
+// Create Chevron icon aligned with text
+.gem-c-step-nav__chevron {
```
andysellick, line 72: It seems like this PR adds a lot of CSS to change the chevron from an SVG to CSS. Is this worth the effort when CSS might look different on different browsers but we can be much more sure the SVG won't?

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
```diff
@@ -63,15 +63,116 @@ $top-border: solid 1px govuk-colour("mid-grey", $legacy: "grey-3");
...
+
+  vertical-align: text-top;
+
+  // Reduce size, alter alignment
+  .gem-c-step-nav--large & {
+    vertical-align: top;
+  }
+
+  // IE8 fallback of icon
+  @include govuk-if-ie8 {
```
andysellick, line 90: I was checking this in browserstack and it looks like these changes have completely broken this component in IE8 - it doesn't appear.

Screenshot 2022-02-02 at 09 35 49

Also in IE9, the styling is a bit broken (current version looks very close to modern Chrome).

Screenshot 2022-02-02 at 09 36 25


Side point, is putting an IE8 fallback in worth the effort? With the SVG on IE8 the icon simply didn't appear, but the step nav was still usable.

app/views/govuk_publishing_components/components/docs/step_by_step_nav.yml
```diff
@@ -194,7 +194,7 @@ examples:
       - style, an attribute on the parent list that controls its appearance. Defaults to optional but can be set to 'choice' (adds indent and bullets to the list)
       - context, an optional text field to show some extra information after the link, usually a monetary value
       - active, whether to make the link highlighted (to indicate the current page)
-      - highlight_step, a general option to highlight the step. Usually the active link would be in the active step, but in some circumstances there might be an active step but no highlighted link.
+      - `highlight_step`, a general option to highlight the step. Usually the active link would be in the active step, but in some circumstances there might be an active step but no highlighted link.
```
andysellick, line 197: If you're highlighting `highlight_step`, shouldn't `context` and `active` also be highlighted?

app/views/govuk_publishing_components/components/docs/step_by_step_nav.yml
```diff
@@ -417,15 +417,15 @@ examples:
 
       The solution is to track the step by step journey the user is on.
 
-      This has been achieved by adding a query string to all of the internal links in the step by step that contains the content_id of the step by step.
+      This has been achieved by adding a query string to all of the internal links in the step by step that contains the `content_id` of the step by step.
 
       For example, if the step by step contains a link to [/check-uk-visa](/check-uk-visa), the component will renders the link as [/check-uk-visa?step-by-step-nav=d8f3c2e0-d544-4664-9616-ab71323e4d18](/check-uk-visa?step-by-step-nav=d8f3c2e0-d544-4664-9616-ab71323e4d18)
```
andysellick, line 422: There's a typo in this line (not your change) that could do with correcting - `the component will renders the link` should be `the component will render the link`.

andysellick (changes requested):
Changes seem to have significantly impacted IE8 and IE9. We're not necessarily required to fully support those browsers, but breaking existing functionality is unnecessary.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
chris-gds: >is putting an IE8 fallback in worth the effort? 

In this context you're right, I should remove it as we don't support 8 so why is it there? This was ported from the DS work as it had to support IE8.

🤔 IE8/9 have existing issues (below) - however I can resolve to resolve functional support.

Screenshot 2022-02-02 at 10 56 56

Screenshot 2022-02-02 at 10 54 50

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
chris-gds: I think would prefer to use SVG however, I think some of the reasons stemmed from IE8+ support, this more closely aligned with providing that icon and then enhancing.  (However as noted below this needs an update)

Also it was around reflow requirements namely `Firefox > Zoom Text Only` support (see image below).  It didn't scale or allow the same level of control.  

image

I'd actually prefer this icon to all be shared (Accordion, Step-by-step, + (upcoming update to) [Details](https://components.publishing.service.gov.uk/component-guide/details)) 🤔 - let me revisit

andysellick: Personally considering the small scale of IE8 usage and that the icon isn't essential (the text already conveys the meaning) I'd be happy if there wasn't an icon on IE8.

That's a good point about Firefox but again while it looks a little strange I don't think the user is missing anything since the text already conveys the meaning.

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
andysellick: I think the existing issues you're seeing with IE8/9 might be limited to the component guide pages for some reason. Checking a [live example](https://www.gov.uk/learn-to-drive-a-car) shows IE8 with full technical support and minor visual issues and IE9 with full support.

**IE8:**

Screenshot 2022-02-02 at 12 06 28

**IE9:**

Screenshot 2022-02-02 at 12 06 52

app/assets/stylesheets/govuk_publishing_components/components/_step-by-step-nav.scss
chris-gds: > I'd be happy if there wasn't an icon on IE8.

Agreed, happy to drop.

> while it looks a little strange I don't think the user is missing anything since the text already conveys the meaning.

I used to edge on this side of thinking too however as we've had [similar-ish issues](https://github.com/alphagov/govuk_publishing_components/pull/1799) (& [this](https://github.com/alphagov/govuk_publishing_components/pull/1793)) noted as WCAG concerns it I felt it made sense to factor this in for these updates

chris-gds:
> The initial state doesn't look right to me

@maxgds nice spot - resolved, the opposite initial state was set. 

>Changes seem to have significantly impacted IE8 and IE9. We're not necessarily required to fully support those browsers, but breaking existing functionality is unnecessary. 

@andysellick some prefixes were needed, these have been resolved. (screenshots in PR)
(I'll raise the JS issue on the gem examples that's causing these browsers to render strangely)

andysellick:
Thanks for fixing in IE8 @chris-gds. IE9 is an interesting one - it does seem to be visually and functionally broken in the component guide, but works okay in an application. That's... strange.

chris-gds:
Some testing failures relate to `tabs` which unrelated. Troubleshooting in parallel but hoping it won't block this

chris-gds:
@andysellick if you're happy to squeeze this in?

andysellick (approved):
I still think the amount of CSS added to replace the SVG is excessive but I understand the problems around this so happy to approve for now.

chris-gds:
> I still think the amount of CSS added to replace the SVG is excessive

I can understand this for step-by-step.  As the CSS was more deeply tested within the accordion 4.0 update I felt using the same technique made more sense. Otherwise it would have been a new code & QA to adjust hover, focus states and reflow requirements on the SVG that would have only been use here.  

If this was to be swapped ideally the SVG would be standalone file (rather than inline JS) that could be shared (details, accordion) that also fell back OK (<IE11) and met reflow requirements.

PR #2605, opened 2022-02-02
Change JavaScript modules from start to init

andysellick:
## What
Convert remaining JavaScript modules from GOV.UK legacy modules using jQuery (with a `start` function) to vanilla JavaScript GOV.UK modules (with an `init` function).

## Why
We're trying to remove jQuery and these modules need to be converted to match the GOV.UK Frontend module style and forthcoming changes to the modules initialisation code.

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
```diff
@@ -458,12 +456,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
   }
 
-  MagnaCharta.prototype.getChartId = function (module) {
+  MagnaCharta.prototype.getChartId = function () {
     var allCharts = document.querySelectorAll('table.js-barchart-table')
     var id = null
 
     for (var i = 0; i < allCharts.length; i++) {
-      if (allCharts[i] === module) {
+      if (allCharts[i] === this.$module) {
```
alex-ju, line 464: was this meant to be `if (allCharts[i] === this.$table) {`? Haven't expanded the two code, but judging by the above changes we don't seem to have a `this.$module` anymore

alex-ju (approved):
Thanks for aligning these scripts to the pattern, Andy! One question, otherwise it looks good to be merged to me. Don't forget the changelog entry.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js
andysellick: Yes, I think so. Good spot, thanks, will change before I merge 👍

andysellick: Slightly worrying that this didn't cause a test failure.

PR #2607, opened 2022-02-03
Migrate cross domain tracking script from static

andysellick:
**NOTE: a new version of the gem containing this PR cannot be deployed to `static` unless it is part of [this PR]() to remove the same code**

## What
Migrates the [cross domain tracking script](https://github.com/alphagov/static/blob/main/app/assets/javascripts/modules/cross-domain-tracking.js) and [its associated test](https://github.com/alphagov/static/blob/main/spec/javascripts/modules/cross-domain-tracking.spec.js) from `static`, and adds it to the analytics code.

Also changes some of the existing analytics test files to create a dummy `setDimension` object if one doesn't already exist. I was getting race condition errors running the tests until these lines were added - I'm not exactly sure why this migration was causing those errors, but there's a long history of the analytics tests destroying each other like this.

The migrated code is almost entirely unchanged. Very very minor tweaks:

- `GOVUK` references explicitly changed to `window.GOVUK` at the behest of the linter
- eslint stuff added to the spec file for jQuery and jasmine, again because of the linter
- no need to explicitly include the `trigger-event.js` script in the spec file

## Why
Cross domain tracking is part of analytics and we should have all this code in the same place.

## Visual Changes
None.

Partly but doesn't completely deal with: https://github.com/alphagov/static/issues/2443

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics.js
@@ -1,3 +1,5 @@
+// these scripts are specifically ordered
+
 //= require govuk_publishing_components/lib/cookie-functions
 //= require govuk_publishing_components/lib/extend-object
 //= require ./analytics/pii
@@ -18,3 +20,4 @@
 //= require ./analytics/explicit-cross-domain-links
 //= require ./analytics/track-click
 //= require ./analytics/track-select-change
+//= require ./analytics/cross-domain-tracking
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Migrate cross domain tracking script from static ([PR #2607](https://github.com/alphagov/govuk_publishing_components/pull/2607))
+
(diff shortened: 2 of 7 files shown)


app/assets/javascripts/govuk_publishing_components/analytics.js
```diff
@@ -18,3 +18,4 @@
 //= require ./analytics/explicit-cross-domain-links
 //= require ./analytics/track-click
 //= require ./analytics/track-select-change
+//= require ./analytics/cross-domain-tracking
```
alex-ju, line 21: If the scripts are not sensitive to order would be good to order them alphabetically; if they're sensitive we may need a comment to state this

alex-ju (approved):

app/assets/javascripts/govuk_publishing_components/analytics.js
andysellick: I think they are sensitive to order (I remember trying to do `require_tree` at some point and it failing) although I don't know what the order is - I think some of these scripts might need to be executed before the others, but some might not.

I put this one at the bottom because it was being executed in static after this analytics file was called. Have added a comment.

PR #2609, opened 2022-02-04
Update select.yml

KyleMacPherson:
## What
Prior to this change, the documentation for the select component suggests passing a parameter called `hint_text`,
however it looks that this should actually be `hint`. Comment on hint_text vs hint on original PR:
https://github.com/alphagov/govuk_publishing_components/pull/2594#discussion_r795723822


alex-ju (approved):

alex-ju:
Thanks for pointing this out @KyleMacPherson! It's going to be addressed in a separate PR

PR #2611, opened 2022-02-08
Refactor select component code

andysellick:
## What
Refactor the select component code. Specifically:

- rename the helper file from `select.rb` to `select_helper.rb` to fit with existing convention
- move most of the logic from the select component template file into the helper

Also fixes the [issue with the documentation](https://github.com/alphagov/govuk_publishing_components/pull/2609) raised in this PR (which seems to have got stuck in the build for some reason).

## Why
Code cleanup.

## Visual Changes
None.


alex-ju (approved):
Needs a rebase, but looks good!

PR #2613, opened 2022-02-09
[WIP] Remove all jQuery from the components gem

andysellick:
## What
Remove jQuery from the components gem, specifically from the modules initialisation code.

## Why
This is an experiment to test how applications with the gem behave when we remove jQuery. It is not currently intended for use in production.

## Visual Changes
None.


andysellick:
Making a note of a suggestion from @kevindew:

```
if (!GOVUK.Modules[moduleName].prototype.init) {
  if (!global.jQuery) throw "You need jQuery for this app"
  new GOVUK.Modules[moduleName]().start(global.jQuery(element))
  element.data(moduleNames[j] + '-module-started', true)
}
```

alex-ju (approved):
jQuery to plain JavaScript changes make sense to me; I admit I didn't got the chance to test it with applications, but it looks like you got this covered.

Nooshu:
Screenshot oof synthetic changes observed in SC.

Screen Shot 2022-03-25 at 12 42 49

Rough changes in data I collected (needs cleaning up to make more sense)
```
Collections: 73 to 41 = 43% / 32kb
Email alerts: 80 to 49 = 39% / 31KB
Feedback: 67 to 35 = 47% / 32kb
Finder Frontend: 105 to 73 = 31% / 32kb
Frontend: 71 to 39 = 45% / 32kb
government forntend: 70 to 38 = 46% / 32kb
info frontend: 64 to 33 = 49% / 31kb
licence finder: 98 to 66 = 33% / 32kb
Manuals frontend: 66 to 35 = 46% / 31kb
Service manual FE: 67 to 35 = 48% / 32kb
smart answers: 80 to 48 = 40% / 32kb
Whitehall 73 to 41 = 44% / 32kb
```

kevindew:
That's awesome, thanks for sharing @Nooshu. So many claps for @andysellick taking this through 👏 👏 👏 

Do you know which app is the 66KB/68KB - fifth one down on the left - It looks like it spikes in size and goes down?

Nooshu:
> That's awesome, thanks for sharing @Nooshu. So many claps for @andysellick taking this through 👏 👏 👏
> 
> Do you know which app is the 66KB/68KB - fifth one down on the left - It looks like it spikes in size and goes down?

I think that was licence-finder, I'm not sure why it spiked, Andy may know more. Maybe 2 versions were included for a day during the transition?

andysellick:
> I think that was licence-finder, I'm not sure why it spiked, Andy may know more. Maybe 2 jQuery versions were included for a day during the transition?

Yes, that must be it. Licence finder was too tricky to remove jQuery from (it has one massive JS file and no corresponding test file, running it locally is problematic and there's talk of retiring it at some point) so it's been given its own, which was deployed immediately prior to jQuery being removed from everywhere else, hence the spike, then the return to normal. Whitehall should have shown a similar pattern.

PR #2616, opened 2022-02-11
Add margin top and bottom option to inset text

andysellick:
## What
Adds options to control the margin top and bottom for the inset text component. Previously both were hard coded as 30px.

## Why
Need some more flexibility when using it in smart-answers.

## Visual Changes

Screenshot 2022-02-11 at 14 18 53

Trello card: https://trello.com/c/LUeE9FAO/641-product-team-template-card

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


maxf (approved):
Looks good :+1:

PR #2617, opened 2022-02-11
Pin rails to 6

alex-ju:
## What
Pin rails to ~6

## Why
Jenkins fails consistently on any current or old branch when testing with ruby 2.7 – the only workaround I could find after a day of investigations is to pin rails to 6 (as 7 seems to be causing the issue), so I'm proposing  this to unblock current work


sihugh (approved):
We should get this on a backlog to investigate

PR #2622, opened 2022-02-14
Update summary list

chris-gds:
## What

Update [`summary_list`](https://components.publishing.service.gov.uk/component-guide/summary_list) - adds a modifier class of `--no-actions` to empty rows only when there is a mix of rows with and without actions

## Why

[Prompted from 4.0 update](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#update-the-html-for-summary-lists), some rows within `summary_list` have an "action" that include links like "`change`, `edit`"  

Only applies 

> if you have a mix of rows with and without actions, add the govuk-summary-list__row--no-actions modifier class to the rows without actions

Inline with Design System / [GOVUK Frontend 4.0 update](https://github.com/alphagov/govuk_publishing_components/pull/2533)

> If you're not using Nunjucks macros, do not include an empty <span class="govuk-summary-list__actions"></span> in rows without actions. Instead, add the govuk-summary-list__row--no-actions modifier class to the row.

## Visual Changes

Previously when a column had "no actions" the column would still be present but would be empty.  This updates removes this `<dd class="govuk-summary-list__actions"></dd>` as the `:after` pseudo element that `no-actions` adds replicates this from a visual standpoint. This also allows Content to span the full container.

screen shot of the summary list before and after "govuk-summary-list__actions" being removed, the content spans the width of the container

When there is mixed actions, the "column" is restored visually: 

image

## Anything else

`govuk_publishing_components`  builds on the [GOVUK Design System](https://design-system.service.gov.uk/components/summary-list/) by adding `gem-c-summary-list__group-title` which adds a heading above the summary list in certain contexts that also can contain an action.  Should this be present the `summary_list` will still extend the full width.

image

Design System version of `summary_list`
https://design-system.service.gov.uk/components/summary-list/mixed-actions/index.html

[4.0.1](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md) (already merged) also has a fix that impacts this

### DIFF
--- a/app/views/govuk_publishing_components/components/_summary_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_summary_list.html.erb
@@ -32,12 +32,10 @@
 
       <% if delete.any? %>
         <% delete_main_link = capture do %>
-          <%
-            delete_section_link_text = delete[:link_text] || t("components.summary_list.delete")
-          %>
+          <% delete_section_link_text = delete[:link_text] || t("components.summary_list.delete") %>
           <%= link_to delete.fetch(:href),
-                    class: "govuk-link gem-link--destructive",
-                    data: delete.fetch(:data_attributes, {}) do %>
+                class: "govuk-link gem-link--destructive",
+                data: delete.fetch(:data_attributes, {}) do %>
             <%= delete_section_link_text %><%= tag.span " #{title}", class: "govuk-visually-hidden" unless delete[:link_text_no_enhance] -%>
           <% end %>
         <% end %>
@@ -62,39 +60,46 @@
           <%= delete_main_link %>
         <% end %>
       <% end %>
-
     <% end %>
 
+    <% any_body_actions = items.detect { |item| item.has_key?("edit") || item.has_key?("delete") } %>
     <% if items.any? %>
       <%= tag.dl class: "govuk-summary-list" do %>
         <% items.each do |item| %>
-          <%= tag.div class: "govuk-summary-list__row" do %>
+          <%
+            item_body_actions = (item.fetch(:edit, {}).any? || item.fetch(:delete, {}).any?)
+
+            if any_body_actions.present? && !item_body_actions.present?
+              classes = "govuk-summary-list__row govuk-summary-list__row--no-actions"
+            else
+              classes = "govuk-summary-list__row"
+            end
+          %>
+          <%= tag.div class: classes do %>
 
             <%= tag.dt item[:field], class: "govuk-summary-list__key" %>
             <%= tag.dd item[:value], class: "govuk-summary-list__value" %>
 
             <% if item.fetch(:edit, {}).any? %>
               <% edit_link = capture do %>
-                <%
-                  edit_link_text = item[:edit][:link_text] || t("components.summary_list.edit")
-                %>
+                <% edit_link_text = item[:edit][:link_text] || t("components.summary_list.edit") %>
                 <%= link_to item[:edit].fetch(:href),
-                            class: "govuk-link",
-                            data: item[:edit].fetch(:data_attributes, {}) do %>
-                  <%= edit_link_text %><%= tag.span " #{item[:field]}", class: "govuk-visually-hidden" unless item[:edit][:link_text_no_enhance] -%>
+                  class: "govuk-link",
+                  data: item[:edit].fetch(:data_attributes, {}) do %>
+                  <%= edit_link_text %>
+                  <%= tag.span " #{item[:field]}", class: "govuk-visually-hidden" unless item[:edit][:link_text_no_enhance] -%>
                 <% end %>
               <% end %>
             <% end %>
 
             <% if item.fetch(:delete, {}).any? %>
               <% delete_link = capture do %>
-                <%
-                  delete_link_text = item[:delete][:link_text] || t("components.summary_list.delete")
-                %>
+                <% delete_link_text = item[:delete][:link_text] || t("components.summary_list.delete") %>
                 <%= link_to item[:delete].fetch(:href),
-                            class: "govuk-link gem-link--destructive",
-                            data: item[:delete].fetch(:data_attributes, {}) do %>
-                  <%= delete_link_text %><%= tag.span " #{item[:field]}", class: "govuk-visually-hidden" unless item[:delete][:link_text_no_enhance] -%>
+                      class: "govuk-link gem-link--destructive",
+                      data: item[:delete].fetch(:data_attributes, {}) do %>
+                  <%= delete_link_text %>
+                  <%= tag.span " #{item[:field]}", class: "govuk-visually-hidden" unless item[:delete][:link_text_no_enhance] -%>
                 <% end %>
               <% end %>
             <% end %>
@@ -115,17 +120,17 @@
                 <% end %>
               <% end %>
             <% else %>
+            <% if edit_link || delete_link %>
               <%= tag.dd class: "govuk-summary-list__actions" do %>
                 <%= edit_link %>
                 <%= delete_link %>
               <% end %>
             <% end %>
-
           <% end %>
         <% end %>
       <% end %>
     <% end %>
-
-    <%= tag.div block, class: "gem-c-summary__block" if block %>
   <% end %>
+    <%= tag.div block, class: "gem-c-summary__block" if block %>
+    <% end %>
 <% end %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## 28.9.0
 
 * Add 'Invasion of Ukraine' CTA to pages tagged to the topical event ([PR #2657](https://github.com/alphagov/govuk_publishing_components/pull/2657))
+* Update summary list ([PR #2622](https://github.com/alphagov/govuk_publishing_components/pull/2622))
 
 ## 28.8.1
 
--- a/app/views/govuk_publishing_components/components/docs/summary_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/summary_list.yml
@@ -1,6 +1,6 @@
 name: Summary list
 description: Use the summary list to summarise information, for example, a user’s responses at the end of a form.
-body:  This component extends the [Summary list component in the Design System](https://design-system.service.gov.uk/components/summary-list/) allowing the rendering of multiple groups of lists, and actions at the group level.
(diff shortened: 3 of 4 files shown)


app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -115,7 +115,7 @@
                 <% end %>
               <% end %>
             <% else %>
-              <%= tag.dd class: "govuk-summary-list__actions" do %>
+              <%= tag.dd class: "govuk-summary-list__actions govuk-summary-list__row--no-actions" do %>
```
maxgds, line 118: Is this definitely right? The guidance says:
```
If you're not using Nunjucks macros, do not include an empty <span class="govuk-summary-list__actions"></span> in rows without actions. Instead, add the govuk-summary-list__row--no-actions modifier class to the row.```
But we weren't using an empty span in the first place - is this necessary? If it is, cool, but a bit more explanation would be good on the PR.

**Review by maxgds — CHANGES_REQUESTED** [review]

File: app/views/govuk_publishing_components/components/_summary_list.html.erb (hunk 1 above)
Comment (chris-gds) [reply app/views/govuk_publishing_components/components/_summary_list.html.erb:118]: I think so - does something standout?

> If you're not using Nunjucks macros, do not include an empty <span class="govuk-summary-list__actions"></span> in rows without actions.

We need the empty row

> Instead, add the `govuk-summary-list__row--no-actions` modifier class to the row.``` 

This has been done in this context

> But we weren't using an empty span in the first place - is this necessary? If it is, cool, but a bit more explanation would be good on the PR.

I can outline a bit more on the PR of why we're using `dd` instead of `span`

File: app/views/govuk_publishing_components/components/_summary_list.html.erb [hunk 2]
```diff
@@ -68,16 +68,14 @@
     <% if items.any? %>
       <%= tag.dl class: "govuk-summary-list" do %>
         <% items.each do |item| %>
-          <%= tag.div class: "govuk-summary-list__row" do %>
-
+          <%= tag.div class: "govuk-summary-list__row #{"govuk-summary-list__row--no-action" if !(item.fetch(:edit, {}).any? || item.fetch(:delete, {}).any?) }" do %>
+
```
jon-kirwan, line 72: I think there's a typo here. Looks like the class name should `govuk-summary-list__row--no-actions`

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -88,12 +86,10 @@
...
-                <%
-                  delete_link_text = item[:delete][:link_text] || t("components.summary_list.delete")
-                %>
+                <% delete_link_text = item[:delete][:link_text] || t("components.summary_list.delete") %>
                 <%= link_to item[:delete].fetch(:href),
-                            class: "govuk-link gem-link--destructive",
-                            data: item[:delete].fetch(:data_attributes, {}) do %>
+                      class: "govuk-link gem-link--destructive",
+                      data: item[:delete].fetch(:data_attributes, {}) do %>
                   <%= delete_link_text %><%= tag.span " #{item[:field]}", class: "govuk-visually-hidden" unless item[:delete][:link_text_no_enhance] -%>
```
jon-kirwan, line 93: Only a suggestion that perhaps this could be written a little differently so that the indentation, and the `<% end %>` of the block are a little easier to identify. Something like:

```
  <%= link_to delete.fetch(:href), class: "govuk-link gem-link--destructive", data: delete.fetch(:data_attributes, {}) do %>
    <%= delete_section_link_text %><%= tag.span " #{title}", class: "govuk-visually-hidden" unless delete[:link_text_no_enhance] -%>
  <% end %>
```

Same question / comment for line 78 too with regards to the edit link.

jon-kirwan (changes requested):
Couple of comments added. Also, even though the changes are not visual, is it necessary to rows with and without actions with one of the doc examples? Similar to the DS docs.

app/views/govuk_publishing_components/components/_summary_list.html.erb
chris-gds: Let me capture this one as it's tad out of scope.

jon-kirwan (commented):
From reviewing the visual regression tests, I think there could be a couple of additional issues:

- The `govuk-summary-list__actions` class is added for each item even when there are no actions to be displayed:

```
<% if edit_link && delete_link %>
  Render "edit" or "delete" link(s)
<% else %>
  <%= tag.dd class: "govuk-summary-list__actions" do %>
    <%= edit_link %>
    <%= delete_link %>
  <% end %>
<% end %>
```

When this happens, an additional cell is added (`display: table-cell;`) and less space available in the description.

components-gem-pr-2622 herokuapp com_component-guide_summary_list

Secondly, the `govuk-summary-list__row--no-actions` class is applied when we have all rows without actions. From a closer look at the DS docs, the class should only be added when there is a mix of rows with and without actions. And here another cell is added and the description is further compressed.

components-gem-pr-2622 herokuapp com_component-guide_summary_list (1)

chris-gds:
@jon-kirwan the plot thickens on this I've noticed an update that's due to come through in 4.0.1

https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md

app/views/govuk_publishing_components/components/_summary_list.html.erb
```diff
@@ -68,16 +66,25 @@
...
+            body_actions = (item.fetch(:edit, {}).any? || item.fetch(:delete, {}).any?)
+
+            if !title_actions || (title_actions && body_actions) 
+              classes = "govuk-summary-list__row"
+            else title_actions
+              classes = "govuk-summary-list__row govuk-summary-list__row--no-actions"
+            end
+              classes
+          %>
+          <%= tag.div class: classes do %>
```
jon-kirwan, line 81: I have spotted an issue with the way the `govuk-summary-list__row--no-actions` class is added when we have title actions (but no "edit" or "delete" action for an individual item). Which creates additional space.

govuk-publishing-components dev gov uk_component-guide_summary_list

So, I think we need to check for the existence of any occurences of "edit" or "delete" actions in the entire array and then use that in the loop when checking if we have a mix of rows with and without actions?

Perhaps something like this: 

```
<% any_body_actions = items.detect { |item| item.has_key?("edit") || item.has_key?("delete") } %>

<% if items.any? %>
  <%= tag.dl class: "govuk-summary-list" do %>
    <% items.each do |item| %>
      <%
        item_body_actions = (item.fetch(:edit, {}).any? || item.fetch(:delete, {}).any?)

        if any_body_actions.present? && !item_body_actions.present?
          classes = "govuk-summary-list__row govuk-summary-list__row--no-actions"
        else
          classes = "govuk-summary-list__row"
        end
      %>
      <%= tag.div class: classes do %>
```

jon-kirwan (changes requested):

app/views/govuk_publishing_components/components/_summary_list.html.erb
chris-gds: > Which creates additional space.

So, this is intentional - from my understanding of this mixed state of actions at the top and no actions, we want to create this faux column, the `no-actions` creates this space. It's sort of the reverse of the below:

image

jon-kirwan: I could be wrong (it's tricky because there is no equivalent in the design system for title actions).

But from the way I read the other examples, it seems we only want to add the extra column when you have a mixture of rows with and without actions, so that the columns align correctly.

Title actions, which appear aligned to the title, and not in the definition list, don't affect the alignment of the columns so we can stretch the description to the right margin and no need for the extra column.

I see what you mean though and it could be intentional.

chris-gds: Questioning my whole comprehension of this issue.  So my understanding is that was the original issue https://github.com/alphagov/govuk-frontend/issues/2495

image

So `--no-actions` was added otherwise if it's removed this 'extra space" is still present

https://user-images.githubusercontent.com/71266765/156416565-2fdb325d-c5f3-45df-bebd-c27dfd308b46.mov

> (it's tricky because there is no equivalent in the design system for title actions).

Agreed, whether this extra space should even exist on the gem version _if_ the title actions are present i'm not sure.

>when you have a mixture of rows with and without actions, so that the columns align correctly.

I've taken this to mean, visually at least, these are the same - the user doesn't differentiate between the two they see actions and scan vertically so this _should_ apply.

jon-kirwan: > So --no-actions was added otherwise if it's removed this 'extra space" is still present

One thing to add. I discovered that `--no-actions` needs to be removed from every row in order for the columns to stretch to fill the container (as it creates an empty `display: table-cell;` and pretty much a new column for the whole definition list).

https://user-images.githubusercontent.com/87758239/156551592-38264303-2a87-4936-b4f4-c8f1a325fb26.mov

chris-gds: This makes sense - as if it's not applied then it would be a "no action" summary list so would extend.  This is a new change that's part of this PR.

image

This modifier persists the faux column when there are mixed actions - in this context where `gem-c-summary-list__group-title` exists.  I think right now my question is more is this understanding correct?  

As it's only on the gem we have `gem-c-summary-list__group-title`.  From a UX POV I believe this is correct but I've asked the DS to confirm either way.

image

Additional action point, there's not an example of a summary-list with no group title _and_ mixed actions. I'll add this.

chris-gds: @jon-kirwan after a chat with the DS and reviewing the original issue - I've totally changed my mind.

- Updated with your suggestions
- Extending the full width when there is no action 
- This also applies regardless of the title actions (letting the content flow the full-width)
- Updated the docs to include include this mix

jon-kirwan (approved):
Looks great, thanks Chris! 👍

maxgds (commented):
I haven't had a full look at this since updates, but since it's approved by Jon I'm removing my request for changes review and am happy for you to ferry this out.

jon-kirwan (approved):

chris-gds:
@maxgds I think your change request is still "active" on this

maxgds (approved):

maxgds:
> @maxgds I think your change request is still "active" on this

Sorry, I thought review->add comments would have cleared it without needing me to explicitly approve, approved now.

PR #2623, opened 2022-02-14
Clear DOM remnants after FilterComponentsSpec

alex-ju:
## What
Clear DOM remnants after `FilterComponentsSpec`

## Why
We clear the search input after this test suite, but not the generate list of components, leaving a few DOM elements behind which can potentially interfere with other tests.

### Before
Screenshot 2022-02-14 at 17 56 22

### After
Screenshot 2022-02-14 at 17 56 47


injms (approved):
Good catch - thank you!

Issue #2555, opened 2022-01-07
Jasmine Ruby gem is EOL


alex-ju:
3.99 is the final version of Jasmine for Ruby. We'll have to consider whether to migrate to `jasmine-browser-runner` npm package (which is the direct replacement for the `jasmine` gem) or use the `jasmine` npm package and run specs in NodeJS

_Originally posted by @alex-ju in https://github.com/alphagov/govuk_publishing_components/pull/2545#pullrequestreview-846667200_

kevindew:
Good shout opening this. I keep seeing signs of it and thinking 😱 

I'd be keen to get involved in working out what we do about this and our plan going forward. I feel perhaps that once we get rid of jQuery we might want to see about how we modernise JS testing or even modernise our JS in general.

alex-ju:
Thanks @kevindew and @ollietreend for getting involved with this!

Issue #2621, opened 2022-02-14
Jasmine tests fail with Rails 7


alex-ju:
Jenkins deployment fails in Ruby 2.7. As a workaround we temporarily [pinned Rails to 6.x](https://github.com/alphagov/govuk_publishing_components/pull/2617) until further investigations on the issue are carried.

```
Puma starting in single mode...
* Puma version: 5.6.1 (ruby 2.7.5-p203) ("Birdie's Version")
*  Min threads: 0
*  Max threads: 5
*  Environment: test
*          PID: 2468
* Listening on http://0.0.0.0:46281/
Use Ctrl-C to stop
Waiting for server on localhost:46281...
jasmine server started
Waiting for suite to finish in browser ...
rake aborted!
Selenium::WebDriver::Error::JavascriptError: javascript error: jsApiReporter is not defined
  (Session info: headless chrome=85.0.4183.83)
#0 0x5614ed894a39 <unknown>
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/response.rb:72:in `assert_ok'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/response.rb:34:in `initialize'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/http/common.rb:88:in `new'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/http/common.rb:88:in `create_response'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/http/default.rb:114:in `request'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/http/common.rb:64:in `call'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/bridge.rb:167:in `execute'
/var/lib/jenkins/bundles/ruby/2.7.0/gems/selenium-webdriver-3.142.7/lib/selenium/webdriver/remote/w3c/bridge.rb:567:in `execute'
[...]

PR #2625, opened 2022-02-15
Switch to `jasmine-browser-runner`

alex-ju:
## What
First attempt at swapping `jasmine` gems with npm equivalents

## Why
[Jasmine gem is EOL](https://github.com/alphagov/govuk_publishing_components/issues/2555)

## Notes
~~Works locally, but fails in Jenkins (help on this much appreciated, have spent some time trying to debug, but not much success as I can't figure out which chromedriver version we're using in CI)~~ Solved by moving to GitHub Actions


Rakefile
```diff
@@ -36,4 +36,11 @@ task lint: %i[rubocop environment] do
   sh "yarn run lint"
 end
 
-task default: [:lint, :spec, "app:jasmine:ci"]
+desc "Jasmine"
+task :jasmine do
+  sh "which chromedriver"
+  sh "chromedriver --version"
+  sh "yarn run jasmine"
```
kevindew, line 43: I assume you'd have to run an assets precompile step before this task?

Rakefile
```diff
@@ -36,4 +36,11 @@ task lint: %i[rubocop environment] do
   sh "yarn run lint"
 end
 
-task default: [:lint, :spec, "app:jasmine:ci"]
+desc "Jasmine"
+task :jasmine do
+  sh "which chromedriver"
+  sh "chromedriver --version"
```
kevindew, line 42: Do we have a pain point that Ruby installs a chromedriver which isn't easy for node to find?

spec/javascripts/support/jasmine_helper.rb
```diff
@@ -1,7 +0,0 @@
-require "jasmine_selenium_runner/configure_jasmine"
-
-class ChromeHeadlessJasmineConfigurer < JasmineSeleniumRunner::ConfigureJasmine
-  def selenium_options
-    { options: GovukTest.headless_chrome_selenium_options }
```
kevindew, line 5: Just to bear in mind for when this progresses more, without this we lose a useful env var flag so that the tests can run within a Docker environment: https://github.com/alphagov/govuk_test/blob/53a8c728caecf412b48ff807437d0c75b5992f13/lib/govuk_test.rb#L30 we may have to think about whether there's a nice way we can share this across apps.

kevindew (commented):
Just snooping by because I saw this pop up and it looked interesting. Awesome stuff getting started on this - does look quite the painful migration.

ollietreend:
Oh, this looks great @alex-ju! 🚀 

I've had [so much pain](https://github.com/alphagov/govuk_publishing_components/issues/2621#issuecomment-1048007995) trying to get Jasmine to run today. But it looks like you've already solved it with this PR.

kevindew:
Hey @alex-ju, I hope it's ok but I've added a bunch of commits to this to resolve some problems and make things a bit cleaner.

Happy to move my commits off if they're divisive at all.

A rough summary is:

- Switched chromedriver to be made available from the webdrivers gem (saves needing to install it twice) - this means the tests run in govuk-docker
- Set up `yarn run jasmine:ci` and `yarn run jasmine:browser` as ways to run the tests in both environments - [govuk-docker will need modifying](https://github.com/alphagov/govuk-docker/blob/cb0cf26cd4311dfbc3c1657e887a1e7506baf3e2/projects/govuk_publishing_components/docker-compose.yml#L23-L27)
- Only test with the specific files we need rather than all JS
- Move more bits to yarn so rake is just wrapper around a yarn task - to be consistent with JS/CSS linting.

alex-ju:
> Hey @alex-ju, I hope it's ok but I've added a bunch of commits to this to resolve some problems and make things a bit cleaner.
> 
> Happy to move my commits off if they're divisive at all.
> 
> A rough summary is:
> 
> - Switched chromedriver to be made available from the webdrivers gem (saves needing to install it twice) - this means the tests run in govuk-docker
> - Set up `yarn run jasmine:ci` and `yarn run jasmine:browser` as ways to run the tests in both environments - [govuk-docker will need modifying](https://github.com/alphagov/govuk-docker/blob/cb0cf26cd4311dfbc3c1657e887a1e7506baf3e2/projects/govuk_publishing_components/docker-compose.yml#L23-L27)
> - Only test with the specific files we need rather than all JS
> - Move more bits to yarn so rake is just wrapper around a yarn task - to be consistent with JS/CSS linting.

Lovely additiona @kevindew! Sorry I wasn't able to progress with this today. It looks like it's close to get a review.

kevindew:
> > Hey @alex-ju, I hope it's ok but I've added a bunch of commits to this to resolve some problems and make things a bit cleaner.
> > Happy to move my commits off if they're divisive at all.
> > A rough summary is:
> > 
> > * Switched chromedriver to be made available from the webdrivers gem (saves needing to install it twice) - this means the tests run in govuk-docker
> > * Set up `yarn run jasmine:ci` and `yarn run jasmine:browser` as ways to run the tests in both environments - [govuk-docker will need modifying](https://github.com/alphagov/govuk-docker/blob/cb0cf26cd4311dfbc3c1657e887a1e7506baf3e2/projects/govuk_publishing_components/docker-compose.yml#L23-L27)
> > * Only test with the specific files we need rather than all JS
> > * Move more bits to yarn so rake is just wrapper around a yarn task - to be consistent with JS/CSS linting.
> 
> Lovely additiona @kevindew! Sorry I wasn't able to progress with this today. It looks like it's close to get a review.

That's no problem, we thought you were on leave today.

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,25 @@
+{
+  "srcDir": "spec/dummy/public/assets",
+  "srcFiles": [
+    "govuk_publishing_components/dependencies-*.js",
+    "govuk_publishing_components/all_components-*.js",
+    "govuk_publishing_components/modules-*.js",
+    "govuk_publishing_components/analytics-*.js",
+    "component_guide/accessibility-test-*.js",
+    "component_guide/filter-components-*.js"
+  ],
```
ollietreend, line 10: I've noticed that the CSS file isn't included with this config, but it was previously included when using the gem.

This may not be a problem, so long as the Jasmine tests only cover functionality and don't cover styling such as `element.style.backgroundColor` etc.

However, for the sake of being comprehensive and making things behave as expected, it'd probably be worth adding it in here:

```suggestion
  ],
  "cssFiles": [
    "application-*.css",
    "component_guide/application-*.css"
  ],
```

Rakefile
```diff
@@ -36,4 +36,9 @@ task lint: %i[rubocop environment] do
   sh "yarn run lint"
 end
 
-task default: [:lint, :spec, "app:jasmine:ci"]
+desc "Jasmine"
+task :jasmine do
+  sh "yarn run jasmine:ci"
+end
```
ollietreend, line 42: In our Rails applications, I'm going to suggest we define this task in its own file at: `lib/tasks/jasmine.rake`. This would make it consistent with similar tasks, such as [`lint.rake` in content-publisher](https://github.com/alphagov/content-publisher/blob/96c51d89ae3eea8a43b2276a9443b08599d0e8eb/lib/tasks/lint.rake).

This repo is a bit weird since it's both a gem and a Rails app, so I'll leave it up to you to decide which location you think is best for the `:jasmine` task to be defined.

spec/support/jasmine-browser.json
```diff
@@ -0,0 +1,25 @@
...
+  "helpers": [
+    "helpers/*.js",
+    "vendor/*.js"
+  ],
+  "env": {
+    "stopSpecOnExpectationFailure": false,
+    "stopOnSpecFailure": false,
+    "random": false
+  },
```
ollietreend, line 23: I noticed that the defaults from the `jasmine-browser-runner` [Getting started guide](https://github.com/jasmine/jasmine-browser-runner#getting-started) are:

```json
  "env": {
    "random": true
  }
```

Is there a reason why we've diverged from the defaults? i.e. is it a deliberate choice, or has it happened by accident?

package.json
```diff
@@ -5,7 +5,10 @@
   "scripts": {
     "lint:js": "standardx 'app/assets/javascripts/**/*.js' 'spec/javascripts/**/*.js'",
     "lint:scss": "stylelint app/assets/stylesheets/",
-    "lint": "yarn run lint:js && yarn run lint:scss"
+    "lint": "yarn run lint:js && yarn run lint:scss",
+    "jasmine:prepare": "bundle exec rake app:assets:clobber && bundle exec rake app:assets:precompile",
```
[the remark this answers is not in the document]
ollietreend: FAO @kevindew

Rakefile
ollietreend: FAO @kevindew

package.json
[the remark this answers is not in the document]
kevindew: That's a good shout - I will update.

spec/support/jasmine-browser.json
kevindew: Sure, I've added it in for equivalence. Somewhat mixed feelings on whether it should be in but could see it needed for a component considering DOM visual state.

Rakefile
kevindew: Yeah looks like there's already a precedence here for Rake tasks defined in that file so I think it's ok here, does some good to have separate files for apps.

spec/support/jasmine-browser.json
kevindew: I hope we don't need that. I turned it off with a rather cocky commit message after a few attempts locally worked fine. Hopefully build passes.

spec/support/jasmine-browser.json
ollietreend: If nothing else, I think it just reduces the "WTF factor" if anything does come to depend upon CSS. I don't think the `cssFiles` option was as well documented, so this avoids somebody trying to scramble around to find it in the future.

I expect the [visual regression tests](https://github.com/alphagov/govuk_publishing_components/blob/master/.github/workflows/visual-regression-tests.yml) are probably the right way to test most of our CSS though.

Rakefile
ollietreend: Good point. I hadn't even spotted that the `Rakefile` already contained other tasks 🙂

spec/support/jasmine-browser.json
ollietreend: And presumably `"stopSpecOnExpectationFailure": false` and `"stopOnSpecFailure": false` just allow the entire test suite to run despite errors, rather than failing fast at the first failure.

I guess that seems like a useful feature. Maybe that's why we've specifically enabled it?

Rakefile
alex-ju: yes, clobber and precompile required, have added them afterwards

Rakefile
alex-ju: yes, I think that was the case

spec/support/jasmine-browser.json
alex-ju: That's a leftover from me fixing a test that failed intermittently, but the fix got out in a separate PR and I forgot to switch back to true 🙈

kevindew: Ah yes these would be good to think about more, thanks for raising @ollietreend.

We'd want `"stopOnSpecFailure": false` in CI as we naturally want to know all the tests that fail. But if I understand `"stopSpecOnExpectationFailure": false` right - that'd be a test carrying on once an expectation fails which seems unusual to me. 

Looking at the repo though - the defaults are: https://github.com/jasmine/jasmine-browser-runner/blob/763ffc8a9b11e9aa621eb9bfe9709a1527f8bc89/lib/examples/default_config.json#L13-L17. These actually match what we have so we can probably just drop this block all together with an expectation that the upstream project remains sensible.

ollietreend: The use of `"examples"` in that file path – `lib/examples/default_config.json` – leaves me suspicious that it's not the real source of truth for config defaults. It seems to be the template used when generating new config via `npm jasmine-browser-runner init` according to https://github.com/jasmine/jasmine-browser-runner/commit/22e32a6a2444ba6d82fd549955714e1409d90402.

Instead, I think Jasmine's defaults are actually defined in the main `jasmine-core` package here: https://github.com/jasmine/jasmine/blob/9e10743a1c7c0a0fc5696261af8b791d9dfe40ce/src/core/Env.js#L36-L134

Regardless, it looks like we were using the default config values, so they should be safe to remove! ✨

kevindew: Oh cool thanks for looking into that.

kevindew:
@ollietreend @alex-ju do we think this is all ready to merge now?

alex-ju:
> @ollietreend @alex-ju do we think this is all ready to merge now?

I've updated the commands in our docs to reference the new way of running jasmine tests via yarn. Since I'm a contributor I guess Ollie would be the best person to provide a formal review.

ollietreend (approved):
This looks good to me ✨ 🚀

PR #2629, opened 2022-02-21
Add print support for rtl languages

chris-gds:
## What 

This adds supports for most / all rtl languages while being printed[1]

## Why

Some languages read right-to-left, when these languages are active the print stylesheets do not adjust to cater for this.

### Before > After

Screenshot of GOVUK page with arabic content while print has been selected and modal is open, the text is aligned left-to-right, this update changes this text to right-to-left

##  Anything else

A solution that should future proof for further localisation, however mindful at present some of these codes / languages are not supported at present on GOVUK
 
Being added to the `govuk_ publishing_components` however only being called per app
Related `government-frontend`  PR: https://github.com/alphagov/government-frontend/pull/2367

Testing link
https://www.gov.uk/government/news/statement-by-saudi-arabia-the-uae-the-uk-and-the-us-about-the-situation-in-yemen-and-the-region.ar

[1]https://meta.wikimedia.org/wiki/Template:List_of_language_names_ordered_by_code

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
@@ -13,3 +13,4 @@
 @import "components/print/step-by-step-nav";
 @import "components/print/textarea";
 @import "components/print/title";
+// @import "components/print/language-support";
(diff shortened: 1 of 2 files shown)


chris-gds:
Closing as after discussion incremental approach was taken

Other considerations

- rtl languages at present (on screen) typically apply a blanket change from ltr to rtl however this impacts layout and and non-translated text.

- this approach adds support for languages that we currently don't translate for, adding bloat.

chris-gds:
Other considerations: https://github.com/alphagov/government-frontend/pull/2367#issuecomment-1054343295

PR #2630, opened 2022-02-21
Use GitHub Actions for Continuous Integration

ollietreend:
## What

This PR configures the repository to use GitHub Actions for continuous integration instead of Jenkins.

It's mostly a copy & paste of the [CI configuration for `govspeak`](https://github.com/alphagov/govspeak/blob/main/.github/workflows/ci.yml) which also uses GitHub Actions to test and publish to RubyGems.

It uses a matrix to run tests against both versions of Ruby that this gem supports: 2.6 and 2.7. It also handles publishing new versions of the gem to RubyGems.org.

### Dependencies

The following dependencies are installed for the test suite to run:

- **Ruby** – the gem is tested against both Ruby 2.6 and Ruby 2.7 using a [matrix build](https://docs.github.com/en/actions/using-workflows/advanced-workflow-features#using-a-build-matrix)
- **Node.js** – the latest LTS release will be used, which is currently Node.js 16
- **govuk_content_schemas** – the `govuk_schema` gem expects [alphagov/govuk-content-schemas](https://github.com/alphagov/govuk-content-schemas) to be available at `../govuk-content-schemas/`. The branch `deployed-to-production` will be checked out, which replicates the [default behaviour of Jenkins](https://github.com/alphagov/govuk-jenkinslib/blob/499503e3893ad184be923cddbd41fcc8b45d4897/vars/govuk.groovy#L618-L622).

## Why

To unblock work to make this gem support Ruby 2.7 and Rails 7. Previous attempts have been hampered by problems with Jenkins (#2621), resulting in this gem being pinned to Rails 6 (#2617).

Also, to adopt [GOV.UK RFC-123](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-123-github-actions-ci.md).

## Visual Changes

None. This is a dev workflow change, so there are no changes to the gem itself.


alex-ju (approved):
This all looks good to me, pending the rubygems key

.github/workflows/ci.yml
```diff
@@ -0,0 +1,68 @@
...
+        ruby: [2.6, 2.7]
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+
+    # Clone the govuk-content-schemas repo, a dependency of the govuk_schemas gem
+    # Make it available at ../govuk-content-schemas, which is where govuk_schemas expects it to be
+    - run: git clone -b deployed-to-production --single-branch https://github.com/alphagov/govuk-content-schemas ../govuk-content-schemas
+
+    # Gemfile.lock isn't needed when testing and releasing this gem – in fact, its existence causes problems.
```
kevindew, line 19: It'd be good to explain these problems in the commit. 

I could anticipate that this could cause problems as well since the contents of Gemfile.lock and gems installed could differ and cause test failures.

PR #2631, opened 2022-02-21
Try running in Heroku without a Gemfile.lock

kevindew:
Temporary PR to try out Heroku running without Gemfile.lock - as we're expecting challenges


kevindew:
This didn't work.

PR #2637, opened 2022-02-24
Update real user metrics LUX.js to v300

injms:
## What

Upgraded Speedcurve RUM script to version 300.

## Why

In order to not break the real user metrics reporting this script needs to be kept up to date.

## Visual Changes
None.


alex-ju (approved):

injms:
Force pushed to add changelog entry.

PR #2646, opened 2022-03-01
Improve custom dimensions code

andysellick:
## What
Improve the analytics custom dimensions code.

- more specifically, the PageContent code called from custom dimensions to supply custom dimensions 26 and 27
- on mainstream browse pages, they should report not only the number of sections and total number of links in the two right hand columns (if visible) but also the left hand column, if that is the only column
- also the code to identify the application and page as collections and browse assumed the meta tag content was 'collections', which is true on live but in Dev for some reason it was 'Collections', which made testing this difficult, so have lowercased that identifier to make it more certain
- also added tests for these two new columns

## Why
We're fixing the analytics on the browse page so we can understand it better before making any changes.

## Visual Changes
None.

Trello card: https://trello.com/c/uyaekaKj/394-fix-google-analytics-page-path-on-ajax-events-on-mainstream-browse-pages

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
@@ -14,7 +14,7 @@
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-title').length
       case isMainstreamBrowsePage():
-        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length
+        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length || document.querySelectorAll('#root ul').length
       case isTopicPage():
         return document.querySelectorAll('.topics-page nav.index-list').length
       case isPolicyAreaPage():
@@ -45,7 +45,7 @@
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-document-list li a').length
       case isMainstreamBrowsePage():
-        return countVisible(document.querySelectorAll('#subsection ul a')) || document.querySelectorAll('#section ul a').length
+        return countVisible(document.querySelectorAll('#subsection ul a')) || document.querySelectorAll('#section ul a').length || document.querySelectorAll('#root ul a').length
       case isTopicPage():
         return document.querySelectorAll('.topics-page .index-list ul a').length ||
           document.querySelectorAll('.topics-page .topics ul a').length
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
```diff
@@ -14,7 +14,7 @@
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-title').length
       case isMainstreamBrowsePage():
-        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length
+        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length || document.querySelectorAll('#root ul').length
```
Rosa-Fox, line 17: It only counts the `#section ul` and `#root ul` as having 1 section each... I guess that makes sense (looking at https://www.gov.uk/browse/births-deaths-marriages/marriage-divorce) as they don't have sections in the way that the `#subsection ul` part has. Just wanted to check that was expected behaviour? 

Also if it is expanded enough to see the `'#subsection ul'` then it only counts the subsections within that menu... again, is this expected behaviour rather than it adding the total of all 3 menus (say if it were for https://www.gov.uk/browse/births-deaths-marriages/marriage-divorce)?

Rosa-Fox: If so then it looks good to me!

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
```diff
@@ -71,7 +71,7 @@
   function getMetaAttribute (selector) {
     var element = document.querySelector(selector)
     if (element) {
-      return element.getAttribute('content')
+      return element.getAttribute('content').toLowerCase()
```
Rosa-Fox, line 74: 👍

Rosa-Fox (commented):
Looks good to me... just a couple of questions to make sure I understand the functionality!

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
andysellick: Yes, this is expected behaviour. Good to ask though, as it is a bit confusing. `root` and `section` only ever have one list, which is why counting is a bit simpler, whereas `subsection` can have many lists, hence the extra effort of the `countVisible` function to calculate it.

And yes, we only want to count the column that is visible on the right. So if it can't count the `subsection`, it tries to count the `section` and if it can't find that it counts the `root`.

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
andysellick: Couldn't figure out why this was working on live but not local. Still weird to me that live is `collections` but local is `Collections`...

alex-ju (approved):

PR #2647, opened 2022-03-01
Update accordion tests

chris-gds:
## What

Update testing on the accordion

## Why

Tests [introduced in this PR](https://github.com/alphagov/govuk_publishing_components/pull/2581) were not actually checking against a rendered page.

## Visuals

Testing run against this URL
http://govuk-publishing-components.dev.gov.uk/accordionexample

image

### DIFF
--- a/spec/features/accordion_spec.rb
+++ b/spec/features/accordion_spec.rb
@@ -1,0 +1,46 @@
+require "rails_helper"
+
+describe "Accordion", js: true, type: :view do
+  scenario "There is a page with the accordion rendered" do
+    given_i_visit_a_page_with_the_accordion_component
+    then_the_accordion_loads
+    then_the_accordion_javascript_loads
+    when_i_click_the_first_accordion_section
+    then_the_accordion_opens
+    when_i_click_the_first_accordion_section
+    then_the_accordion_has_data_attributes
+  end
+
+  def given_i_visit_a_page_with_the_accordion_component
+    visit "/accordionexample"
+  end
+
+  def then_the_accordion_loads
+    expect(page).to have_css(".gem-c-accordion", visible: :visible)
+    expect(page).not_to have_content("This is the content for Writing well for the web.")
+  end
+
+  def then_the_accordion_javascript_loads
+    expect(page).to have_css(".gem-c-accordion--active", visible: :visible)
+    expect(page).to have_css(".govuk-accordion__show-all", visible: :visible)
+    expect(page).not_to have_content("This is the content for Writing well for the web.")
+  end
+
+  def when_i_click_the_first_accordion_section
+    page.first(".govuk-accordion__section-button").click
+  end
+
+  def then_the_accordion_opens
+    expect(page).to have_selector(".govuk-accordion__section.govuk-accordion__section--expanded", visible: :visible)
+    expect(page).to have_content("This is the content for Writing well for the web.")
+  end
+
+  def then_the_accordion_has_data_attributes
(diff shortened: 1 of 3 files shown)


spec/features/accordion_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    expect(page).to have_css(".gem-c-accordion--active", visible: true)
+  end
+
+  def when_i_click_the_first_accordion_section
+    page.first(".govuk-accordion__section-button").click
+  end
+
+  def then_the_accordion_opens
+    puts page.body
+    expect(page).to have_selector(".govuk-accordion__section.govuk-accordion__section--expanded", visible: true)
```
injms, line 32: Could we also check for the content of the open section? That way we know that the CSS is doing the correct thing.

For example:
```ruby
expect(page).to have_content("This is the content for Writing well for the web.")
```

spec/features/accordion_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+    when_i_click_the_first_accordion_section
+    then_the_accordion_has_data_attributes
+  end
+
+  def given_i_visit_a_page_with_the_accordion_component
+    visit "/accordionexample"
+  end
+
+  def then_the_accordion_loads
+    expect(page).to have_css(".gem-c-accordion", visible: true)
```
injms, line 19: Do you think it's worth checking the accordion hasn't got any sections open when it's loaded?

```ruby
expect(page).not_to have_content("This is the content for Writing well for the web.")
```

spec/features/accordion_spec.rb
```diff
@@ -0,0 +1,43 @@
...
+
+describe "Accordion", js: true, type: :view do
+  scenario "There is a page with the accordion rendered" do
+    given_i_visit_a_page_with_the_accordion_component
+    then_the_accordion_loads
+    then_the_accordion__javascript_loads
+    when_i_click_the_first_accordion_section
+    then_the_accordion_opens
+    when_i_click_the_first_accordion_section
+    then_the_accordion_has_data_attributes
```
injms, line 11: Is it worth having a check for the "Show all sections" button as well?

spec/features/accordion_spec.rb
injms, line 19: Using `visible: :visible` is preferred [^1], as using `false` can lead to confusion since it finds both visible _and_ invisible elements - so `false` is the equivalent to `visible: :all`, rather than being the equivalent to `visible: :hidden`.

 ```suggestion
     expect(page).to have_css(".gem-c-accordion", visible: :visible)
```


[^1]: https://docs.rubocop.org/rubocop-rspec/cops_rspec/capybara.html#rspeccapybaravisibilitymatcher

spec/features/accordion_spec.rb
chris-gds: I'm on the fence about what's worth replicating testing wise that's done on the [Design System side](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/accordion/accordion.test.js) - however I've added a sense check on `25`

injms: Good point, we don't need to duplicate tests. The sense check makes sense - thank you!

injms (approved):
Super - thanks for fixing this!

spec/features/accordion_spec.rb
chris-gds: huh, thank you - really good to know.

PR #2654, opened 2022-03-02
Add spellcheck to input 

chris-gds:
## What

Update `inputs` to include `spellcheck=""false"` by default - allows the option to be changed to `true` when required

## Why

Inline with [GOVUK Design System Guidance](https://design-system.service.gov.uk/patterns/email-addresses/), noted as part of accessible format request epic

Many inputs like `email` and `name` have no value being spellchecked due to their nature.  However other fields do, allows this option to be manually passed if required.

## Visual Changes

NA

## Anything else

Docs have been updated, code formatting added + `autocomplete` has been moved to be grouped next to email as this is highly recommended to be used in conjunction

### DIFF
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -7,6 +7,7 @@
   describedby ||= nil
   enterkeyhint ||= nil
   id ||= "input-#{SecureRandom.hex(4)}"
+  spellcheck ||= "false"
   type ||= "text"
   value ||= nil
   inputmode ||= nil
@@ -63,6 +64,7 @@
     inputmode = "numeric"
     pattern = "[0-9]*"
   end
+
 %>
 
 <%= content_tag :div, class: form_group_css_classes do %>
@@ -114,6 +116,7 @@
     name: name,
     pattern: pattern,
     readonly: readonly,
+    spellcheck: spellcheck,
     tabindex: tabindex,
     type: type,
     value: value
--- a/app/views/govuk_publishing_components/components/docs/input.yml
+++ b/app/views/govuk_publishing_components/components/docs/input.yml
@@ -24,12 +24,19 @@ examples:
         text: "What is your name?"
       name: "name"
   specific_input_type:
-    description: By default the input will be type="text". This parameter accepts an alternative, e.g. "search" or "email". No validation is done on this input.
+    description: By default the input will be `type="text"`. This parameter accepts an alternative, e.g. `search` or `email`. `spellcheck="false"` is the default this can be changed by passing a `true` value. Consider adding `autocomplete`. No validation is done on this input.
     data:
       label:
         text: "What is your email address?"
       name: "address"
       type: "email"
+  with_autocomplete:
+    data:
+      label:
+        text: "Automatically complete the field with a user's email address (in supporting browsers)"
+      name: "name"
+      type: "email"
+      autocomplete: "email"
   numeric_input:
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -63,6 +64,10 @@
     inputmode = "numeric"
     pattern = "[0-9]*"
   end
+
+  if type == "email"
+    spellcheck = "false"
+  end
```
jon-kirwan, line 70: I'm just wondering if we should also include `type == "number"`? (I see that this type does get modified to `text` like so):
```
  if type == "number"
    type = "text"
    inputmode = "numeric"
    pattern = "[0-9]*"
  end
```
... but looking at the example it feels like we should probably set `spellcheck = "false"` too. Like for the 'account number example https://components.publishing.service.gov.uk/component-guide/input/numeric_input/preview

jon-kirwan (changes requested):
Just one comment on whether we should apply `spellcheck = "false"` to other input types.

Also, is there any benefit to adding test for this too?

app/views/govuk_publishing_components/components/_input.html.erb
chris-gds: I like it, let me update.

chris-gds:
> Just one comment on whether we should apply `spellcheck = "false"` to other input types


Looking at the way other input types are used there's nothing added or recommended, is there another you have in mind?
https://design-system.service.gov.uk/components/text-input/



>Also, is there any benefit to adding test for this too? 

I did think as it's not user controllable maybe not, however I've added a check.

jon-kirwan (commented):
All looks fine. I have a question around `name="name"` though and whether this might be a tricky one because of the different variations on name in use.

For example, I checked against `feedback` for [Request an accessible format...](https://github.com/alphagov/feedback/blob/f50bd0b3e70b8cf5471dd3359a66b90c458815a4/app/views/contact/govuk/accessible_format_requests/contact_details.html.erb#L42) and we use name="full_name" and on the [contact form](https://www.gov.uk/contact/govuk) it's `name="contact[name]"`

chris-gds:
> For example, I checked against feedback for [Request an accessible format...](https://github.com/alphagov/feedback/blob/f50bd0b3e70b8cf5471dd3359a66b90c458815a4/app/views/contact/govuk/accessible_format_requests/contact_details.html.erb#L42) and we use `name="full_name"` and on the [contact form](https://www.gov.uk/contact/govuk) it's `name="contact[name]"`

This is tricky as `name` is used for various labelling eg ` name: "lead-times",` so it's not like blanket `name` can be used - maybe a different approach is needed perhaps it should always be `false` unless it's explicitly set as `true`.

app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -119,17 +119,20 @@ examples:
       name: "name"
       width: 10
   with_search_icon:
+    description: |
+        Adds a search icon, `spellcheck="true"` is also present when the value is passed.
```
jon-kirwan, line 123: A very finicky comment (apologies!) but just wonder if this could be misinterpreted that spellcheck is added when `search_icon: true` is present.

Perhaps should word slightly differently. Something like...

```
Adds a search icon.

[spellcheck](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/spellcheck) can also be added to indicate that the element should be, if possible, checked for spelling errors
```

chris-gds: Works for me. ✅

jon-kirwan (approved):
Thanks! 👍

PR #2662, opened 2022-03-07
[DO NOT MERGE] Add GTM tracking module initial version

andysellick:
## What
Adds JavaScript to handle Google Tag Manager (GTM) tracking.

- listens for clicks on elements with data-module="gtm-tracking"
- expects a data-gtm-attributes attribute on the clicked element, containing an object that can be extracted using JSON parse to get the required attributes to push to the GTM dataLayer

All this script currently does is push to the dataLayer. It requires that GTM has been initialised at the application level using the [Google Tag Manager script component](https://components.publishing.service.gov.uk/component-guide/google_tag_manager_script) for anything to actually be registered in GA.

This PR is currently incomplete as there are unanswered questions about the attributes and how they will work. Mainly:

- for attributes that are component specific, how do we add these to components - are they hard coded into the component or always passed
- what the values for a lot of these attributes should be

## Why
We're migrating towards GA4 with Google Tag Manager and this is part of the initial work to prepare for that.

## Visual Changes
None.

Trello card: https://trello.com/c/PBL5ZR5c/841-ga4-push-gtm-data-from-clicks-to-the-action-link-component

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics.js
@@ -18,3 +18,5 @@
 //= require ./analytics/explicit-cross-domain-links
 //= require ./analytics/track-click
 //= require ./analytics/track-select-change
+
+//= require ./analytics/gtm-tracking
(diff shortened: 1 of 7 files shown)


andysellick:
Closing this PR as the technical complexity and effort involved to add tracking to all components is demonstrably too much. Will try a different approach.

Issue #2674, opened 2022-03-09
Slight visual discrepency for navigation header search button when JavaScript is not available


injms:
The navgiation header has a visually identical search toggle button and search link - the search toggle button is visible and the link hidden when JavaScript is available; and visa versa when JavaScript is unavailable.

The blue search toggle button has a hairline border at the bottom:
![www gov uk_ (4)](https://user-images.githubusercontent.com/1732331/157485656-9cdabce3-6231-49a9-a131-c320e930131c.png)

https://github.com/alphagov/govuk_publishing_components/blob/86b361c94d13d1566c3a483674faf04df796b1ce/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss#L708

When JavaScript is turned off, the search link should be visually identical - but it lacks the dark bottom border:

![www gov uk_ (5)](https://user-images.githubusercontent.com/1732331/157489767-be50c174-7618-48fc-8aff-af955e5e307b.png)

I think that this is okay to be added in here:

https://github.com/alphagov/govuk_publishing_components/blob/86b361c94d13d1566c3a483674faf04df796b1ce/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss#L472-L477

PR #2686, opened 2022-03-16
Add border to search icon when JS disabled

chris-gds:
## What

Add border to bottom of search icon on the [`layout_super_navigation_header`](https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header) when JS is disabled

## Why

> The blue search toggle button has a hairline border at the bottom/ When JavaScript is turned off, the search link should be visually identical - but it lacks the dark bottom border

Closes https://github.com/alphagov/govuk_publishing_components/issues/2674

## Visual Changes

image


injms:
I think that this looks good and would be happy to approve, but I don't believe I should have the power to approve it as I've moved teams.

alex-ju (approved):
Approving based on Ian's comment

PR #2687, opened 2022-03-16
Remove emails from accessible format request pilot

edwardkerry:
The pilot scheme is being paused so we are temporarily removing the participating emails from the attachment component.

This will prevent links to the accessible format request form being displayed on inaccessible attachments, and instead display the mail_to link.

### DIFF
--- a/lib/govuk_publishing_components/presenters/attachment.rb
+++ b/lib/govuk_publishing_components/presenters/attachment.rb
@@ -1,15 +1,11 @@
 module GovukPublishingComponents
   module Presenters
     class Attachment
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/attachment.rb
```diff
@@ -4,12 +4,7 @@ class Attachment
       # DfE, DWP, DHSC, HMRC, DVSA and PHE are taking part in a pilot to use a form
```
KludgeKML, line 4: Should this comment be changed to reflect the current reality?

edwardkerry: thank you! updated it

1pretz1 (approved):

PR #2689, opened 2022-03-17
Add variant tracking to card component

chris-gds:
## What

Add tracking `dimension26` and `dimension32` for tracking sections and number of links

## Why

New [variant browse pages](https://github.com/alphagov/collections/pull/2673) have been designed and built that are based in `collections` need analytics added for Browse Level 0 & 1.

## Visual Changes

NA

Passes the number of sections and links, for ref - 

dimension26 being highlighted on omnibug google analytics tracking plugin that reveals the number 8 which the number of links on the page

## Anything else?
 
This is being added as part of an AB test.  Depending on the outcome and changes this may need to be either a) removed b) altered

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
 * **BREAKING:** Remove all jQuery from the components gem ([PR #2613](https://github.com/alphagov/govuk_publishing_components/pull/2613))
 * Add compatibility with Sprockets 4, see [Sprockets documentation](https://github.com/rails/sprockets/blob/master/UPGRADING.md) on how to upgrade your Rails app ([PR #2691](https://github.com/alphagov/govuk_publishing_components/pull/2691))
 * Use a data attribute to specify RUM (real user monitoring) script location ([PR #2682](https://github.com/alphagov/govuk_publishing_components/pull/2682))
+* Add variant tracking to card component ([PR #2689](https://github.com/alphagov/govuk_publishing_components/pull/2689))
 
 ## 28.9.2
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
@@ -13,6 +13,8 @@
         return document.querySelectorAll('[data-track-count="accordionSection"]').length
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-title').length
+      case isNewBrowsePage():
+        return document.querySelectorAll('[data-track-count="cardList"]').length
       case isMainstreamBrowsePage():
         return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length || document.querySelectorAll('#root ul').length
       case isTopicPage():
@@ -44,6 +46,8 @@
         return document.querySelectorAll('a[data-track-category="navAccordionLinkClicked"]').length
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-document-list li a').length
+      case isNewBrowsePage():
+        return document.querySelectorAll('[data-track-count="cardLink"]').length
       case isMainstreamBrowsePage():
         return countVisible(document.querySelectorAll('#subsection ul a')) || document.querySelectorAll('#section ul a').length || document.querySelectorAll('#root ul a').length
       case isTopicPage():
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
```diff
@@ -14,7 +14,7 @@
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-title').length
       case isMainstreamBrowsePage():
-        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length || document.querySelectorAll('#root ul').length
+        return countVisible(document.querySelectorAll('#subsection ul')) || document.querySelectorAll('#section ul').length || document.querySelectorAll('#root ul').length || document.querySelectorAll('.gem-c-cards__list').length
```
andysellick, line 17: I'm wondering about this approach. Rather than extending the existing `isMainstreamBrowsePage` functions, would it be cleaner to create a whole new set that we can adapt as needed? Something like this.

```
PageContent.getNumberOfSections = function () {
...
  case isNewBrowsePage():
    return document.querySelectorAll('.gem-c-cards__list').length

...

PageContent.getNumberOfLinks = function () {
...
  case isNewBrowsePage():
      document.querySelectorAll('.gem-c-cards__link').length

...

  function isNewBrowsePage () {
    return getMetaAttribute(metaApplicationSelector) === 'collections' &&
      getMetaAttribute(metaFormatSelector) === 'new_browse_page'
  }
```

Obviously it will involve more code changes but it might be more flexible if we need to adapt it, and easier to clean up if we retire the existing browse page code.

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
```diff
@@ -45,7 +45,7 @@
       case isDocumentCollectionPage():
         return document.querySelectorAll('.document-collection .group-document-list li a').length
       case isMainstreamBrowsePage():
-        return countVisible(document.querySelectorAll('#subsection ul a')) || document.querySelectorAll('#section ul a').length || document.querySelectorAll('#root ul a').length
+        return countVisible(document.querySelectorAll('#subsection ul a')) || document.querySelectorAll('#section ul a').length || document.querySelectorAll('#root ul a').length || document.querySelectorAll('.gem-c-cards__link').length
```
andysellick, line 48: Generally we try to avoid referencing a class name internal to a component outside of a component, because components are subject to change and doing so can be brittle - but that might not always be possible. Is there an alternative way to count items on the new browse page without referencing a component class?

chris-gds: >so can be brittle

100% agreed, how does a `data-` approach sound instead for this? Adding an attribute to the component - for tracking.  I feel like doing DOM selections element is equally quite brittle.  I mean it's less likely the `ul > a` format will change however it might (eg an `ol` for curated content)

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
chris-gds: I'm open to do this - it isolates it more and gives more granular control for these new templates. 

Thinking out loud, it would be nice for this to exist in the component - or the page - and pass into tracking.

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
andysellick: Yeah, a data attribute approach could work, because those would be defined on the browse page and passed to the component, so much more specific and controllable.

app/assets/javascripts/govuk_publishing_components/analytics/page-content.js
andysellick: Yeah, there's definitely some things about how this tracking works that aren't ideal - having this script hidden away separate from the browse pages that explicitly relies upon the markup structure of those pages (again) feels brittle. But hopefully we can address issues like this in the upcoming GA work.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Add compatibility with Sprockets 4, see [Sprockets documentation](https://github.com/rails/sprockets/blob/master/UPGRADING.md) on how to upgrade your Rails app ([PR #2691](https://github.com/alphagov/govuk_publishing_components/pull/2691))
 * Use a data attribute to specify RUM (real user monitoring) script location ([PR #2682](https://github.com/alphagov/govuk_publishing_components/pull/2682))
+* Add variant tracking to card component ([PR #2689](https://github.com/alphagov/govuk_publishing_components/pull/2689))
```
andysellick, line 14: Can this match the PR name please?

andysellick (commented):
Couple of minor things, but otherwise good.

andysellick (approved):

PR #2693, opened 2022-03-22
Add tracking on the accordion

chris-gds:
## What

Add click event tracking for [Accordion's](https://components.publishing.service.gov.uk/component-guide/accordion) "Show all sections" button behind a feature flag.

Also, add click event tracking for each section, also behind a feature flag.

## Why

For PA's to track when button is interacted with for new GOVUK Explore, Browse Pages AB variant. Replicates bespoke functionality found on `/coronavirus`.

## Visual Changes

Example capture on `collections`

https://user-images.githubusercontent.com/71266765/160830285-f25dcd8a-47a8-4c74-8b88-8129cdd8c8a9.mov

https://user-images.githubusercontent.com/71266765/160697806-59bd56b0-5b4b-4ffd-a014-82e3877a6cea.mov

## Anything else

Issue raised to capture tracking testing that is missing from this PR.
https://github.com/alphagov/govuk_publishing_components/issues/2694

Related PRs
https://github.com/alphagov/collections/pull/2673
https://github.com/alphagov/collections/pull/2709

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -28,6 +28,7 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
   GemAccordion.prototype.init = function () {
     // Indicate that JavaScript has worked
     this.$module.classList.add('gem-c-accordion--active')
+    this.$module.querySelector('.' + this.showAllControls).classList.add('gem-c-accordion__show-all')
 
     // Feature flag for anchor tag navigation used on manuals
     if (this.$module.getAttribute('data-anchor-navigation') === 'true') {
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -14,6 +14,8 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     this.sectionHeaderClass = 'govuk-accordion__section-header'
     this.sectionInnerContent = 'govuk-accordion__section-content'
     this.showAllControls = 'govuk-accordion__show-all'
+    this.sectionButton = 'govuk-accordion__section-button'
+    this.headingText = 'govuk-accordion__section-heading-text'
 
     // Translated component content and language attribute pulled from data attributes
     this.$module.actions = {}
@@ -38,6 +40,10 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     // Feature flag for "Show all sections" GA click event tracking
     if (this.$module.getAttribute('data-track-show-all-clicks') === 'true') {
       this.addAccordionOpenAllTracking()
+    }
+    // Feature flag for each section GA click event tracking
+    if (this.$module.getAttribute('data-track-sections') === 'true') {
+      this.addEventListenerSections()
     }
   }
 
@@ -107,6 +113,26 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
         window.GOVUK.analytics.trackEvent('pageElementInteraction', action, options)
       }
     })
+  }
+
+  GemAccordion.prototype.addEventListenerSections = function () {
+    var sections = this.$module.querySelectorAll('.' + this.sectionButton)
+    nodeListForEach(sections, function (section) {
+      section.addEventListener('click', this.addAccordionSectionTracking.bind(this, section))
+    }.bind(this))
+  }
+
+  // If the Accordion's sections are opened on click, then pass them to the GA event tracking
+  GemAccordion.prototype.addAccordionSectionTracking = function (section) {
+    var expanded = section.getAttribute('aria-expanded') === 'false'
+    var label = section.querySelector('.' + this.headingText).textContent
+    var action = expanded ? 'accordionOpened' : 'accordionClosed'
+    var options = { transport: 'beacon', label: label }
+
+    if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
+      window.GOVUK.analytics.trackEvent('pageElementInteraction', action, options)
+      console.log('pageElementInteraction', action, options)
+    }
   }
 
   Modules.GemAccordion = GemAccordion
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -13,6 +13,7 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     this.sectionClassExpanded = 'govuk-accordion__section--expanded'
     this.sectionHeaderClass = 'govuk-accordion__section-header'
     this.sectionInnerContent = 'govuk-accordion__section-content'
+    this.showAllControls = 'govuk-accordion__show-all'
 
     // Translated component content and language attribute pulled from data attributes
     this.$module.actions = {}
@@ -32,6 +33,10 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     if (this.$module.getAttribute('data-anchor-navigation') === 'true') {
       this.openByAnchorOnLoad()
       this.addEventListenersForAnchors()
+    }
+    // Feature flag for "Show all sections" GA click event tracking
+    if (this.$module.getAttribute('data-track-show-all-clicks') === 'true') {
+      this.addAccordionOpenAllTracking()
     }
   }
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -13,6 +13,7 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     this.sectionClassExpanded = 'govuk-accordion__section--expanded'
     this.sectionHeaderClass = 'govuk-accordion__section-header'
     this.sectionInnerContent = 'govuk-accordion__section-content'
+    this.showAllControls = 'govuk-accordion__show-all'
```
andysellick, line 16: Not sure it's worth declaring this as a variable when it only gets used once and the variable name is almost as long as the thing it refers to.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -90,5 +95,19 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     }
   }
 
+  // To track the Accordion's "Show all sections" / "Hide all sections" button click events and pass them to the GA event tracking
+  // This is added via JS after page load by the Accordion component,
```
andysellick, line 99: I think this line of this comment isn't needed?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -5,6 +5,7 @@
   id ||= "default-id-#{SecureRandom.hex(4)}"
   items ||= []
   anchor_navigation ||= false
+  event_tracking ||= false
```
andysellick, line 8: Would a better name for this variable be something like `track_show_all_clicks`?

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -309,3 +309,27 @@ examples:
...
+          content:
+            html: <p class="govuk-body" id="anchor-nav-test">This is content for accordion 2 of 2</p>
+        - heading:
+            text: Know your audience
+          content:
+            html: <p class="govuk-body">This is the content for Know your audience.</p>
+        - heading:
+            text: How people read
+          content:
+            html: <p class="govuk-body">This is the content for How people read.</p>
```
andysellick, line 335: This example seems overly long - it only needs to demonstrate the show all button click, two accordion sections. will be enough.

docs/analytics/analytics.md
```diff
@@ -297,3 +297,6 @@ analytics tracker without attempting to strip PII from it.  For example:
   GOVUK.analytics.setDimension(1, new GOVUK.Analytics.PIISafe('this-is-not-an@email-address-but-it-looks-like-one'));
   GOVUK.analytics.trackEvent('report title clicked', new GOVUK.Analytics.PIISafe('this report title looks like it contains a P0 5TC ode but it does not really'));
 ````
+#### Tooling
+
+[Ominbug](https://chrome.google.com/webstore/detail/omnibug/bknpehncffejahipecakbfkomebjmokl?hl=en) is a useful extension to easily capture and review the events being sent to Google Analytics.
```
andysellick, line 302: Omnibug.

spec/dummy/app/views/welcome/accordion_example.html.erb
```diff
@@ -1,6 +1,7 @@
 <h1>A test page in the dummy app</h1>
 <p>This is for testing the accordion component</p>
 <%= render "govuk_publishing_components/components/accordion", {
+  event_tracking: true,
```
andysellick, line 4: Why does the dummy app need to be updated?

app/views/govuk_publishing_components/components/_accordion.html.erb
chris-gds: Works for me, will update

spec/dummy/app/views/welcome/accordion_example.html.erb
chris-gds: So the plan, at the moment, is to [extend the testing](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/features/accordion_spec.rb) so the dummy app loads GA and can check this event fires as expected.  I'm aware other testing does this a different way by testing a [copy of the element instead of the element](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/components/checkboxes-spec.js).

However I feel like these aren't real tests as we're almost trying to keep copies of the HTML fixture in sync rather than testing what the user actually sees.  As the Accordion testing happens on an actual page, I thought it make more sense to continue with this approach.  (If I create a new `accordion-spec.js` file to test tracking i'm un-doing this work)

andysellick: Okay, that makes sense, but I'd save this change for that future work - it doesn't add anything to this PR.

chris-gds: Sure, happy to capture to tackle outside of this PR

app/assets/javascripts/govuk_publishing_components/components/accordion.js
chris-gds: Omph - I agree in isolation, however this is more sticking to convention, I felt mixing the way we select elements in the same file was potentially more confusing. I'm tempted to leave this as is for this reason.

spec/dummy/app/views/welcome/accordion_example.html.erb
chris-gds: https://github.com/alphagov/govuk_publishing_components/issues/2694

andysellick: Sorry, I think I misunderstood this slightly and I've looked at the code again. This component, despite having a JS file, does not appear to have a corresponding JS test file. That is not good, but it is not your fault. However, since you're adding some JS to this file, I think this is the right moment to create a JS test file and put some tests in for this code, although I don't expect you to complete the file by adding any further tests for the rest of the already existing JS.

I think also that the [issue you raised](https://github.com/alphagov/govuk_publishing_components/issues/2694) should be updated to more simply state that the whole of the JS for the accordion should be tested. In terms of the point raised about testing against the actual markup of the component vs a copy of the markup, that's valid, but we don't currently have a mechanism for that, and there is still value in testing the JS in this way. For example, the first thing the accordion JS does is add a class to the accordion - which can be tested fully regardless of whether it is run against the real markup of the component or a copy.

andysellick:
@chris-gds percy seems to be failing for a random and seemingly unrelated reason. I've had this in the past and rebasing fixed it.

spec/dummy/app/views/welcome/accordion_example.html.erb
chris-gds: > This component, despite having a JS file, does not appear to have a corresponding JS test file. That is not good, but it is not your fault. However, since you're adding some JS to this file, I think this is the right moment to create a JS test file and put some tests in for this code, although I don't expect you to complete the file by adding any further tests for the rest of the already existing JS.

> I think also that the https://github.com/alphagov/govuk_publishing_components/issues/2694 should be updated to more simply state that the whole of the JS for the accordion should be tested. 

[So, there's a lot of testing happening on the Design System side](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/accordion/accordion.test.js), I didn't wish to duplicate testing here as it's redundant. "gem-only" features that exist here have been layered on top and these are [tested](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/features/accordion_spec.rb).  I'm wondering for this GA tracking to restore this file.

Is this a comment to add testing to re-test these via the JS?

> In terms of the point raised about testing against the actual markup of the component vs a copy of the markup, that's valid, but we don't currently have a mechanism for that, 

As this is now rendering on the `dummy_app` (http://govuk-publishing-components.dev.gov.uk/accordionexample) is this not a valid mechanism? (We use this for `/tabsexample` as well?)

> and there is still value in testing the JS in this way.  For example, the first thing the accordion JS does is add a class to the accordion - which can be tested fully regardless 

So, this has been tested here, against rendered markup,  Should I duplicate this in a JS file?
https://github.com/alphagov/govuk_publishing_components/blob/main/spec/features/accordion_spec.rb#L24

> of whether it is run against the real markup of the component or a copy.

In this context it's against the rendered accordion, I'm wondering what coverage is perhaps missing from the testing here:  https://github.com/alphagov/govuk_publishing_components/blob/main/spec/features/accordion_spec.rb

andysellick: I think I haven't made myself clear. You've added some JS functionality and that should have a corresponding test. Something like:

```
// accordion-spec.js

var FIXTURE = document.createElement('div')
new GOVUK.Modules.accordion(FIXTURE).init()
spyOn(window.GOVUK.analytics, 'trackEvent')

var showAllLink = FIXTURE.querySelector('.show-all-link')
showAllLink.click()
expect(window.GOVUK.analytics.trackEvent).toHaveBeenCalledWith('whatever') // show all

showAllLink.click()
expect(window.GOVUK.analytics.trackEvent).toHaveBeenCalledWith('whatever') // hide all

```

chris-gds: Ace, I'm on this page - I'm restoring the file and doing something similar just for tracking

andysellick (approved):
As long as the test issue is addressed relatively soon I'm happy to approve.

chris-gds:
Outstanding: https://github.com/alphagov/govuk_publishing_components/issues/2694

PR #2698, opened 2022-03-24
Adjust how modules.start is run

andysellick:
## What
Take away the document event listener for page ready that wraps the call to `modules.start`.

## Why
The document event listener was added to replace jQuery's document.ready, but now that jQuery isn't included anymore I don't think either are needed, and I have concerns about how well `DOMContentLoaded` is supported in older browsers.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


kevindew (approved):
Looks good to me as we expect the JS to be in the footer and DOM already loaded. Don't think browser support should be an issue though, if you're using a browser without this support you'd have a lot of other problems.

andysellick:
This doesn't work at all - I think the reason the `document.ready` was originally in place was due to the ordering of the scripts. Going to close for now.

PR #2702, opened 2022-03-29
Remove the last traces of jQuery

andysellick:
## What
Turns out there were some tiny bits of jQuery left in the gem, or references to it.

## Why
We took jQuery out of GOV.UK last week.

## Visual Changes
None.


sihugh (approved):
LGTM

PR #2708, opened 2022-03-31
Set up a GOV.UK conventional dependabot file

kevindew:
The previous one was auto-generated by dependabot in https://github.com/alphagov/govuk_publishing_components/commit/6a9aa1b334851d905ed757d7d5d3d5045e83a707 and wasn't representative of the conventional approach for GOV.UK dependencies.

It also contained a bunch of dependencies this project didn't have and was missing a few (such as stylelint-config-gds).

As part of this I've also cleaned up the gemspec development dependency pins, which seemed superfluous.


ollietreend (approved):
It's lovely to lose the `ignore:` list! 🎉 
Thanks @kevindew ✨

PR #2714, opened 2022-04-01
[WIP] Refactor Feedback for AB test for Insights

andyhd:
⚠️ Work in progress - do not merge

* Currently the variation displayed is selected by the `variant` URL parameter - this is only a temporary implementation for use during development
* No effort has yet been put into making the refactored component accessible, performant or completely functional
* No tests yet

## What

* Refactor feedback to use a data-driven state machine
* Add two alternate feedback multi-step workflows

## Why

The Insights team wants to conduct a multivariate test of the Feedback component. The variations they want to test are branching multi-step workflows. The existing component does not support branching, so this change implements a data-driven state machine to model the existing workflow and the two new variations.

## Visual Changes

The existing component remains unchanged visually


andyhd:
Closing as this work was abandoned

PR #2719, opened 2022-04-06
use locale files to generate header links in public_layout instead of constant in public_layout_helper

patrickpatrickpatrick:
## What
Instead of using a constant set in `public_layout_helper`, `public_layout_helper` now uses new values in the locale files to generate the links and structure of the footer in `public_layout.html.erb`.

## Why
The task was to make the footer links and headings the same as those in the header but upon seeing how it was implemented I have done this by adding to the footer links and headings to the locale files.

[Trello Card Link](https://trello.com/c/0O5wJh7e/878-small-amends-to-the-sitewide-footer)

### Before
cropped

### After
Screenshot 2022-04-19 at 09 45 28


patrickpatrickpatrick:
also aware that the FOOTER_META needs to use locale files also but thought that would be better as a seperate branch because it's a different concern

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -1,294 +1,7 @@
...
-                  track_category: "footerClicked",
-                  track_action: "footerLinks",
-                  track_label: "Transparency and freedom of information releases",
-                },
-              },
-            },
-          ],
-        },
-      ].freeze
+      FOOTER_NAVIGATION_RATIO = [2, 1].freeze
```
hannalaakso, line 4: Is there a particular reason we need to freeze the variable here? Is it just to guard against someone changing this but not updating the data to reflect? Might be worth leaving a comment about it?

I'm also wondering if `FOOTER_NAVIGATION_COLUMNS` might be a more descriptive name? I had to look at the code to understand what it was for. But not a blocker.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -333,7 +46,23 @@ class PublicLayoutHelper
       attr_reader :footer_navigation, :footer_meta, :cookie_banner_data
 
       def initialize(local_assigns)
-        @footer_navigation = local_assigns[:footer_navigation] || FOOTER_NAV
+        @footer_navigation = local_assigns[:footer_navigation] || I18n.t("components.layout_footer.navigation_links").each_with_index.map do |menu, i|
```
hannalaakso, line 49: I'm wondering what will happen if `navigation_links` in the particular translation file is empty? Will this throw up an error?

hannalaakso (commented):
Looks like a good tidy up and a solution to the problem. Left a couple of comments.

Also, would it be worth updating the content for https://components.publishing.service.gov.uk/component-guide/layout_footer/with_navigation/preview to keep it in sync?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
chao-xian: `.freeze` is for making the constant a constant. Name change suggestion seems sensible.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
chao-xian: It'll default to `en.yml`

chao-xian (approved):
LGTM

lib/govuk_publishing_components/presenters/public_layout_helper.rb
hannalaakso: Thanks @chao-xian, I got that but was querying if there was a particular scenario where it might be at risk of being overwritten.

PR #2722, opened 2022-04-08
add new custom dimension for new browse page template analytics

patrickpatrickpatrick:
## What
This PR adds a new a dimension to the analytics that is derived from the metatag `navigation-list-type`. This metatag will appear on the new browse pages and will state whether they are a 'curated' or a 'default' page. 

## Why
The new browse pages are a hierarchy of different levels each containing a default set or a curated set of related pages. Before this PR, the level of hierarchy of a page and whether or not the page was curated was stored in a single dimension. This would make checking analytics for a specific combination of level and/or hierarchy more difficult, so it was requested that this dimension was split into two. `navigation-list-type` will indicate if a page is curated and `navigation-page-type` will indicate what level of the hierarchy a page is.


hannalaakso (commented):
I've tested this together with https://github.com/alphagov/collections/pull/2740 and the correct dimensions are being set 🎉  

Do we need to add some tests for this new meta tag, similar to what we do for `navigation-page-type`?

patrickpatrickpatrick:
> I've tested this together with [alphagov/collections#2740](https://github.com/alphagov/collections/pull/2740) and the correct dimensions are being set 🎉
> 
> Do we need to add some tests for this new meta tag, similar to what we do for `navigation-page-type`?

oh sure can add new tests if they're required

hannalaakso (approved):

Issue #1968, opened 2021-03-10
JS bug on pages that use the govspeak magna charta stacked chart


owenatgov:
On pages using the magna charta stacked variant, there is a js error triggered by the stacked chart specifically. This is causing any js that has yet to be run after the offending chart to break, including the feedback component.

[Examples page](https://www.gov.uk/government/publications/government-digital-inclusion-strategy/government-digital-inclusion-strategy)

andysellick:
Error seems to occur here: https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/govspeak/magna-charta.js#L264

![Screenshot 2022-03-14 at 16 18 48](https://user-images.githubusercontent.com/861310/158215399-48f0494c-07c4-405c-afae-91ca57a7d9ca.png)

PR #2731, opened 2022-04-14
Fix JS bug in MagnaCharta for stacked charts

MartinJJones:
## What
Improve the error handling for stacked charts when the table source HTML does not include the required table header elements.

## Why
Without the error handling, not only will the Chart will not be rendered, but any remaining JavaScript will not be run after the error, leading to further issues with other components in the bundled JS file.

## Visual Changes

| **Before**  | **After** |
| ----------- | ----------- |
| If the table header elements are not present, the chart is not rendered, however the "toggleLink" button is still displayed with misleading text | If the table header elements are not present, the chart and toggleLink button are not rendered |
| Screenshot 2022-04-14 at 09 55 27  | Screenshot 2022-04-14 at 09 56 11 |

Fixes https://github.com/alphagov/govuk_publishing_components/issues/1968

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix JS bug in MagnaCharta for stacked charts ([PR #2731](https://github.com/alphagov/govuk_publishing_components/pull/2731))
+
 ## 29.6.0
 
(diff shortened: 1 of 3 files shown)


maxgds (approved):
Nice work.

MartinJJones:
Resolved merge conflict in CHANGELOG.MD and force-pushed the changes.

PR #2745, opened 2022-04-29
remove coronavirus topic from menu bar

patrickpatrickpatrick:
## What
Remove Coronavirus link from the menu bar.

## Why
This change was requested by the Covid team.

### Before
Screenshot 2022-04-29 at 10 03 27

### After
Screenshot 2022-04-29 at 10 03 03


AgaDufrat:
It looks like the first commit is removing the topic from the menu bar but your second commit (updating changelog) is adding it back: 
Screenshot 2022-04-29 at 10 24 33

patrickpatrickpatrick:
> It looks like the first commit is removing the topic from the menu bar but your second commit (updating changelog) is adding it back: Screenshot 2022-04-29 at 10 24 33

whoops had added it back for the screenshots which i forgot about, fixing that now

AgaDufrat (approved):

PR #2747, opened 2022-04-29
Update cross domain linking script to use init

danacotoran:
It was noted that cross-domain linking between the account and the DI domain stopped working around 25th March.

Around that time, a lot of good work was done around removing jQuery from GOVUK. As part of that work there were some [changes](https://github.com/alphagov/govuk_publishing_components/commit/613cfcc8f71c96d8099bf28dd51231f48bbec3b9) to the script which starts the modules in `govuk_publishing_components` where support for legacy modules was removed.

The explicit-cross-domain-links script was still using `start()` to initialise so it stopped working around that time. This changes it to `init()`.

To test: on this page https://www.gov.uk/email/subscriptions/single-page/new?topic_id=keeping-a-pet-pig-or-micropig, when you click **Create a GOV.UK account or sign in**, the URL of the next page should have a cookie consent query parameter which reflects whether you've accepted cookies, rejected cookies, or not engaged with the cookie banner.

-----

https://trello.com/c/EM69ov98

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
@@ -1,89 +1,88 @@
-;(function (global) {
-  'use strict'
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
 
-  var GOVUK = global.GOVUK || {}
-  GOVUK.Modules = GOVUK.Modules || {}
+(function (Modules) {
+  function ExplicitCrossDomainLinks ($module) {
+    this.$module = $module
+  }
 
-  GOVUK.Modules.ExplicitCrossDomainLinks = function () {
-    this.start = function ($module) {
-      this.element = $module[0]
-      this.attribute = 'href'
-      this.attributeValue = this.element.getAttribute(this.attribute)
-      this.eventType = 'click'
-      if (!this.attributeValue) {
-        this.attribute = 'action'
-        this.attributeValue = this.element.getAttribute(this.attribute)
-        this.eventType = 'submit'
-      }
-
-      this.handleEvent = this.handleEvent.bind(this)
-      this.handleCookiesAccepted = this.handleCookiesAccepted.bind(this)
-      // Listens for the 'submit' event if the element is a form, and the 'click' event if it is a link
-      this.element.addEventListener(this.eventType, this.handleEvent)
+  ExplicitCrossDomainLinks.prototype.init = function () {
+    this.attribute = 'href'
+    this.attributeValue = this.$module.getAttribute(this.attribute)
+    this.eventType = 'click'
+    if (!this.attributeValue) {
+      this.attribute = 'action'
+      this.attributeValue = this.$module.getAttribute(this.attribute)
+      this.eventType = 'submit'
     }
 
-    this.handleEvent = function (e) {
-      //  prevent default: we want the link href and/or form action to be decorated before we navigate away
-      e.preventDefault()
-      var cookieBannerEngaged = GOVUK.cookie('cookies_preferences_set')
-      var cookieConsent = GOVUK.getConsentCookie()
+    this.handleEvent = this.handleEvent.bind(this)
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -1,32 +1,15 @@
...
-      this.handleEvent = this.handleEvent.bind(this)
-      this.handleCookiesAccepted = this.handleCookiesAccepted.bind(this)
-      // Listens for the 'submit' event if the element is a form, and the 'click' event if it is a link
-      this.element.addEventListener(this.eventType, this.handleEvent)
-    }
+(function (Modules) {
+  function ExplicitCrossDomainLinks ($module) {
+    this.$module = $module
 
     this.handleEvent = function (e) {
```
andysellick, line 8: I see that `handleEvent` was already inside the `start` function but I think normally we try to include the bulk of the JS outside of the constructor function, unless it's necessary. Can this be moved to its own function within the module at the same level as `this.init()`? i.e.

```
(function (Modules) {
  function ExplicitCrossDomainLinks ($module) {
    this.$module = $module
  }

  ExplicitCrossDomainLinks.prototype.init = function () {
    // ...
  }

  ExplicitCrossDomainLinks.prototype.handleEvent = function () {
    // ...
  }
```

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -85,5 +68,22 @@
     }
   }
 
-  global.GOVUK = GOVUK
-})(window)
+  ExplicitCrossDomainLinks.prototype.init = function () {
+    this.element = this.$module
```
andysellick, line 72: Seems like an unnecessary reassignment of a variable - could the script be updated to use `this.$module` instead of `this.element`?

andysellick (commented):
Argh, sorry, thought I'd got all these scripts converted from start to init. Many apologies.

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
danacotoran: Question: what is the benefit of adding functions to the prototype which do not need to be public (i.e. there is no need to access them from outside this module as the purpose they serve is highly specific to this module)?

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
danacotoran: Good spot, no idea why I did this, will update!

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
andysellick: I think there's a technical reason to not put too much into the constructor but I can't remember what it is unfortunately (I remember being called out on the same thing on one of my PRs in the past). Other than that I'd say the reasons were convention (generally our constructors are short, containing only the minimum of stuff i.e. `this.$module = $module`) and readability (`init` is the place where the code starts executing, so having that as near the top as possible is good).

app/assets/javascripts/govuk_publishing_components/analytics/explicit-cross-domain-links.js
```diff
@@ -1,89 +1,88 @@
...
-      }
+      var linker = new window.gaplugins.Linker(trackers[0])
+      var attrValue = element.getAttribute(attribute)
 
-      element.setAttribute(attribute, attributeValue)
+      element.setAttribute(attribute, linker.decorate(attrValue))
+    })
+  }
+
+  ExplicitCrossDomainLinks.prototype.init = function () {
```
andysellick, line 71: Tiny nitpick - could you put the `init` function as the first function? Just because we do that everywhere else and it's the first bit of code that executes.

andysellick (approved):
🚀

PR #2750, opened 2022-05-03
Realign list of topics in navigation header

injms:
## What

Updates the `columns` mixin with the updated number of items in the list of topics in the navigation header.

## Why

An [item was removed from the topics listed in the navigation header][removed], which requires the CSS grid layout to be rejigged to avoid lopsided columns.

### Before
image

### After
image

[removed]: https://github.com/alphagov/govuk_publishing_components/pull/2745


maxgds (approved):

injms:
Force pushed after sorting out a changelog conflict.

PR #2751, opened 2022-05-03
Fix print styles for organisation logos

### LINKED ISSUE #2374: Printing pages with organisation logos can have a overly large logo
## Problem 

If an organisation has a non-crest, custom image logo it will be printed out at it's unresized dimensions - this can lead to a logo being surprisingly large. For example, on a [page with the Environment Agency page][eg]:

## Expected behaviour

The logo should be small, which is more in keeping with the screen version of the page, and doesn't overwhelm the layout.

[eg]:https://www.gov.uk/government/publications/social-distancing-when-signing-and-handing-over-waste-transfer-and-consignment-notes-in-person-rps-c8/social-distancing-when-signing-and-handing-over-waste-transfer-and-consignment-notes-in-person-rps-c8

patrickpatrickpatrick:
## What
Sets a width for the organisation logo component when being printed so that it is not excessively large if image used for organisation logo is a large image. This resolves issue https://github.com/alphagov/govuk_publishing_components/issues/2374.

## Why
When printing a page with the organisation logo component, there was no width set on the image so it would be printed at the size of the image itself. This meant if the image was large then the image would appear much larger than it had on the page itself. A potential fix would be to hide the organisation logo when printing but as the organisation logo is the only indication on the page when printed that the page is originating from an organisation I thought it was important to retain it.

## Visual Changes
N.B. this is print view only

### Before
Screenshot 2022-05-03 at 14 16 52

### After
Screenshot 2022-05-03 at 14 17 05

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Extend GTM analytics click tracking ([PR #2786](https://github.com/alphagov/govuk_publishing_components/pull/2786))
 * GOVUK Frontend 4.1.0 updates ([PR #2794](https://github.com/alphagov/govuk_publishing_components/pull/2794))
+* Fix print styles for organisation logos ([PR #2751](https://github.com/alphagov/govuk_publishing_components/pull/2751))
 
 ## 29.10.0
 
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Approved, but can you add a CHANGELOG entry please? We're going to need to update the component asset includes for all applications that use this component, so it would be helpful to have this in the CHANGELOG, thanks.

PR #2753, opened 2022-05-03
add wrapper to the component card to ensure link spacing

patrickpatrickpatrick:
## What
Add a wrapper to the content of the card component.

## Why
When testing for how usable the collections pages were with a magnifier tool, I noticed that there was a difference to how the cards were implemented in the govuk_publishing_components gem to how they were implemented on GOV.UK itself. The cards on GOV.UK have a wrapper containing the sub-heading and description so that the area of the card that acts as a link is only the portion of the card containing text and not the entire area of the card. This means there is a space between each card above and below on the grid. In the govuk_publishing_components implementation of the card, there is no wrapper which means the entire card acts as a link and so there is no spacing above and blow each card on the grid. This could potentially make the site less usable if using a magnifying tool as the user could accidentally click on the wrong link. 

## Before
Screenshot 2022-05-04 at 09 36 29

## After
Screenshot 2022-05-04 at 09 35 01

[Related Trello Card](https://trello.com/c/Fpj1ARx0/933-do-frontend-resilience-cross-browser-and-assistive-tech-testing-on-level-0-1-templates)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
@@ -50,12 +50,19 @@
   // We use grid to split the container into column widths, so manage the horizontal spacing with
   // internal margins.
   margin: 0 govuk-spacing(3);
-  padding: govuk-spacing(3) 0 govuk-spacing(6) 0;
+  padding: govuk-spacing(1) 0 govuk-spacing(4) 0;
+}
+
+.gem-c-cards__list-item-wrapper {
+  // this wrapper ensures that the clickable area of the card only
+  // covers the area of the card containing text so in a grid of cards
+  // there is space above and below each link
+  padding: govuk-spacing(2) govuk-spacing(6) govuk-spacing(2) 0;
   position: relative;
 }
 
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
```diff
@@ -51,11 +51,15 @@
   // internal margins.
   margin: 0 govuk-spacing(3);
   padding: govuk-spacing(3) 0 govuk-spacing(6) 0;
+}
+
```
hannalaakso, line 55: Could we leave a comment here to explain that the wrapper is there to limit the size of the touch area to differentiate from the one below?

patrickpatrickpatrick:
tested on iOS on device, when pressing between two cards neither link is followed which indicates there is a good enough gap to differentiate between cards

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
patrickpatrickpatrick: i have added a comment to the stylesheet explaining the purpose of the wrapper

PR #2755, opened 2022-05-09
Make auditing check static for missing assets

andysellick:
## What
If present, use data from `static` to make warnings about missing assets in applications more accurate. Specifically:

- if `static` includes an asset and the application doesn't, no longer raise a warning (asset not actually missing)
- if `static` includes an asset and the application does as well, raise a new warning (asset is duplicated)

We only need to do this for certain applications, specifically the ones that use `static` (`static` is included in the auditing as well, so obviously we need to not compare `static` with itself).

## Why
The warnings in the auditing sometimes contained false positives, which made the warnings generally less reliable. Once we've made the auditing more reliable we can start to, well, rely on it more.

The result of this work is that most of the applications in the auditing now have more warnings. The next step is to manually check and test that these warnings are accurate.

## Visual Changes

Added a bit of visual information about this in the interface (this only appears for applications that use `static`).

Screenshot 2022-05-09 at 10 42 14

Warnings for duplication of assets:

Screenshot 2022-05-13 at 12 22 30

### DIFF
--- a/app/models/govuk_publishing_components/audit_comparer.rb
+++ b/app/models/govuk_publishing_components/audit_comparer.rb
@@ -4,6 +4,22 @@ module GovukPublishingComponents
 
     def initialize(gem_data, results, simple)
       if gem_data[:gem_found]
+        @applications_using_static = %w[
+          collections
+          email-alert-frontend
+          feedback
+          finder-frontend
+          frontend
+          government-frontend
+          info-frontend
+          licence-finder
+          manuals-frontend
+          service-manual-frontend
+          smart-answers
+          whitehall
+        ]
+
+        @static_data = find_static(results)
         @gem_data = gem_data
         @applications_data = sort_results(results, simple)
         @gem_data[:components_by_application] = get_components_by_application || []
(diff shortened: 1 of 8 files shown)


app/models/govuk_publishing_components/audit_comparer.rb
```diff
@@ -4,6 +4,22 @@ class AuditComparer
...
+          finder-frontend
+          frontend
+          government-frontend
+          info-frontend
+          licence-finder
+          manuals-frontend
+          service-manual-frontend
+          smart-answers
+          whitehall
+        ].sort
```
injms, line 20: These are already alphabetical, so is `.sort` needed?

andysellick:
@injms I've realised that this PR is incomplete, apologies for asking you to review it. I'll add some more commits (so you won't have to go back through the code) and ping you when it's ready.

andysellick:
@injms I've reworked this, and it works better (details in the desc) but I've realised a problem. Probably best to discuss direct.

andysellick:
Had a problem where `static` had the title component in a stylesheet for the error page, which is only used on the error page, but that meant that the auditing thought static included title everywhere, so produced some inaccurate warnings.

Have solved this by removing the error stylesheet from static: https://github.com/alphagov/static/pull/2777 as it wasn't really needed anyway.

patrickpatrickpatrick (approved):
looks good to me

PR #2760, opened 2022-05-09
Add GTM analytics click tracking

andysellick:
## What
Add a generic script for pushing data to the dataLayer for Google Tag Manager. This is an initial approach that may change over time.

- adds generic Google Tag Manager 'click to track' script, that depends upon there being a dataLayer to work
- dataLayer is currently only initialised using the GTM component and environment variables, so should only work on integration
- the analytics-ga4 script is currently not included anywhere, this will be included conditionally in static based on the same environment variables, to avoid code being deployed to live

Right now we're concerned with not doing anything that adds any code to the public site, so these scripts will only be included in static in development environments and on integration.

The script also includes one approach to dealing with tracking on expandable items (specifically accordions right now) where we want to record the open/closed state of the accordion when clicked. This is done by looking for an element with `aria-expanded` in the clicked element. If found, this is used to override the `ui.state` value sent to the dataLayer.

## Why
We want to start progressing with the GA upgrade.

## Visual Changes
None.

Trello card: https://trello.com/c/V0RFrSMK/43-test-technical-approach-to-tracking-on-tabs-and-accordions

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
@@ -1,0 +1,49 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  function GtmClickTracking (module) {
+    this.module = module
+    this.trackingTrigger = 'data-gtm-event-name' // elements with this attribute get tracked
+  }
+
+  GtmClickTracking.prototype.init = function () {
+    var trackClicksOn = [this.module]
+    if (!this.module.getAttribute(this.trackingTrigger)) {
+      trackClicksOn = this.module.querySelectorAll('[' + this.trackingTrigger + ']')
+    }
+
+    for (var i = 0; i < trackClicksOn.length; i++) {
+      trackClicksOn[i].addEventListener('click', this.trackClick.bind(this))
+    }
+  }
+
+  GtmClickTracking.prototype.trackClick = function (event) {
+    if (window.dataLayer) {
+      var target = event.currentTarget
+      var data = {
+        event: 'analytics',
+        event_name: target.getAttribute('data-gtm-event-name'),
+        // get entire URL apart from domain
+        link_url: window.location.href.substring(window.location.origin.length),
+        ui: JSON.parse(target.getAttribute('data-gtm-attributes'))
+      }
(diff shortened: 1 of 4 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
```diff
@@ -0,0 +1,94 @@
...
+        'test-1-1': 'test 1-1 value',
+        'test-1-2': 'test 1-2 value'
+      }
+      var attributes2 = {
+        'test-2-1': 'test 2-1 value',
+        'test-2-2': 'test 2-2 value'
+      }
+      element.innerHTML =
+        '<div data-gtm-event-name="event1-name"' +
+          'data-gtm-attributes=\'' + JSON.stringify(attributes1) + '\'' +
```
injms, line 55: To avoid the need for escaping and to be consistent with the previous and next line, double quotes could be used:
```suggestion
          'data-gtm-attributes="' + JSON.stringify(attributes1) + '"' +
```

andysellick: I thought so too - but it doesn't work.

`stringify` returns a string containing `"` so the rendered HTML becomes something invalid like `<div data-gtm-attributes="{ "attribute": "value" }">` and the test fails.

(I was going to switch the quotes/ticks around but the linter enforces `'` on the outer boundary)

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
```diff
@@ -0,0 +1,48 @@
...
+    }
+  }
+
+  GtmClickTracking.prototype.trackClick = function (event) {
+    if (window.dataLayer) {
+      var target = event.currentTarget
+      var data = {
+        event: 'analytics',
+        event_name: target.getAttribute('data-gtm-event-name'),
+        link_url: window.location.href.substring(window.location.origin.length),
```
injms, line 29: I _think_ this gets the path of the URL - is it worth adding a small comment to add context and explain what this does?

injms (approved):
This looks good from my understanding of what's required for the Google Analytics upgrade. 

I think there's a separate part that's needed later - when the analytics code is less fluid - to make sure that this is well documented.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
andysellick: Thanks, added.

PR #2761, opened 2022-05-09
Add Architecture Design Records 

gclssvglx:
## What

We should document the design decisions related to this project.

An [Architecture Design Record](https://cognitect.com/blog/2011/11/15/documenting-architecture-decisions) is an ideal - and structured - way to do this and is also the GOV.UK preferred approach.

## Why

So that others know why a particular decision was taken, what options were also considered and why they were rejected.

## Visual Changes
None

[Trello](https://trello.com/c/Z59PS5na/44-create-an-adr-to-document-the-decision-around-analytics-code-location)

### DIFF
--- a/docs/adr/0002-google-analytics-4-tracking-and-code-location.md
+++ b/docs/adr/0002-google-analytics-4-tracking-and-code-location.md
@@ -1,0 +1,48 @@
+# 2. Google Analytics 4 tracking and code location
+
+Date: 2022-05-09
+
+## Status
+
+Accepted
+
+## Context
+Glossary of terms/acronyms
+
+- GA - Google Analytics
+- UA - Universal analytics, sometimes referred to as GA3, the version of GA we are upgrading from
+- GA4 - the new version of UA that we are upgrading to
+- GTM - Google Tag Manager
+
+GOV.UK currently uses UA for analytics. However, UA is due to be retired next year. We will have to upgrade to GA4. We are taking this opportunity to do a full overhaul of our existing analytics. This will involve rethinking how our analytics works, bringing consistency into the way we do it, and introducing GTM.
+
+## Decision
+We will put the new code for GA4/GTM analytics into govuk_publishing_components:
+
+- other options were static or a new gem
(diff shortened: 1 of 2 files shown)


docs/adr/0002-google-analytics-4-tracking-and-code-location.md
```diff
@@ -0,0 +1,19 @@
+# 2. Google Analytics 4 tracking and code location
+
+Date: 2022-05-09
+
+## Status
+
+Accepted
```
Rosa-Fox, line 7: I wonder if this should start off being 'pending' or 'TBC' to start off with? I guess it will depend if we use the Github PR reviews for ADR comments or if we use a google doc and then the PR is more of a final version? It looks like the ADRs are also documented in the dev docs (for example [Github](https://github.com/alphagov/smart-answers/blob/main/docs/arch/003-testing-flows.md) and [dev docs](https://docs.publishing.service.gov.uk/repos/smart-answers/arch/003-testing-flows.html)) so I wonder if there should be a mention of also adding to the docs in the other general ADR file ([0001-record-architecture-decisions.md](https://github.com/alphagov/govuk_publishing_components/pull/2761/files#diff-e2c27172663e2ac0af76f95037853f6a52e5f77889e90ad1c1f8644279f74402)).

.adr-dir
```diff
@@ -0,0 +1 @@
+docs/adr
```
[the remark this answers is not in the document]
gclssvglx: Oh, this is auto-created when you use the `adr-tools`, but you use a non-default location. The default is `doc/adr` and we're using `docs/adr`. I've removed it as I don't think we either need or will use the `adr-tools` here as all it does is create a template file - which we can simply copy from the initial ADR anyway.

docs/adr/0002-google-analytics-4-tracking-and-code-location.md
gclssvglx: Yeah, good point - I think you got to this PR before I'd had chance to create the Google doc that we've agreed to work in 🚤  But it's worth changing it anyway - now `In progress`. Thanks @Rosa-Fox

GDSNewt (approved):

PR #2767, opened 2022-05-12
Remove console.log

andysellick:
## What
Spotted a `console.log` in the accordion.js, removing.

## Why
Not needed in production - looks like it was added for debug and accidentally left in.

## Visual Changes
None.


injms (approved):
🕵️

PR #2769, opened 2022-05-12
Add HTTP protocol measurement to RUM

injms:
## What

This adds in a new measurement to the payload that will be sent to Speedcurve's LUX service - like all metrics, this will only be sent if and when a user opts in.

## Why

Knowing which HTTP protocol is being used is important for understanding how the different protocols affect load times and our user's experiences.

Letting LUX know which protocol is being used will allow us to compare the protocols and what effect they have on the performance metrics monitored.

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-measurer.js
```diff
@@ -157,3 +157,31 @@ if (
...
+      var getEntriesByType = performance.getEntriesByType('navigation')
+
+      if (getEntriesByType.length > 0) {
+        var httpProtocol = performance.getEntriesByType('navigation')[0].nextHopProtocol
+        LUX.addData("http-protocol", httpProtocol)
+      }
+    })
+  }
+} catch (e) {
+  console.error('Error in LUX reporting the HTTP protocol:', e)
```
andysellick, line 186: Would it be worth including `window.location` in this error? (or is that included in the `e` object?)

spec/javascripts/govuk_publishing_components/rum/http-protocol-measurement-spec.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-env jasmine */
+/* global LUX */
+
+// This test will only work if the browser running it supports the HTML5
+// Performance API.
+if (typeof window.performance !== 'undefined') {
+  describe('The LUX RUM function should report the the HTTP protocol', function () {
+    // Because the script being tests is loaded and run as soon as possible, a
```
andysellick, line 8: Typo `being tests` should be `being tested`?

spec/javascripts/govuk_publishing_components/rum/http-protocol-measurement-spec.js
```diff
@@ -0,0 +1,26 @@
+/* eslint-env jasmine */
+/* global LUX */
+
+// This test will only work if the browser running it supports the HTML5
+// Performance API.
+if (typeof window.performance !== 'undefined') {
+  describe('The LUX RUM function should report the the HTTP protocol', function () {
```
andysellick, line 7: I'm not sure I understand how this test works, although my knowledge of this whole area is limited. I guess I'm used to tests that create a condition then compare the outcome with an expected set of data. But this test doesn't create a condition - it just expects things. Is there any comment you could add to explain this a little?

andysellick (approved):
I have a few comments, but non-blocking as the core of this seems solid, as far as I understand it.

spec/javascripts/govuk_publishing_components/rum/http-protocol-measurement-spec.js
injms: Fixed - thank you!

app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-measurer.js
injms: Good shout - I've added it in.

spec/javascripts/govuk_publishing_components/rum/http-protocol-measurement-spec.js
injms: I've added some more in depth detail about this that I hope makes sense. 

It's a difficult and different thing to test, as the script runs automatically and shouldn't require initialisation because it should run as soon as possible to get the most accurate measurement. Unfortunately I can't think of a better way of testing this.

PR #2773, opened 2022-05-12
Update LUX reporter script to v301

injms:
## What
Update the LUX reporter script from v300 to v301.

## Why
If the version of the script gets too old, no data will be able to be sent to Speedcurve. Since we self host this script, we need to make sure that it's the latest version.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


Nooshu:
I think we will also have to check for the existence of `performance.getEntriesByType` looking at the errors coming through in the RUM error logs
Screenshot 2022-05-16 at 17 22 50

Nooshu (approved):
Looks good Ian, much easier to review the unminified version.

PR #2774, opened 2022-05-13
Tracking changes for the footer

patrickpatrickpatrick:
## What

Add new data-attributes to links in the footer. For most links in the footer, this involves a new helper method which works out what data attributes a link should have and then adds it to the hash that is used to generate each link. The only exception to this is the copyright and license links. As they have html that is hardcoded in the locale file, they have hardcoded data-tracking variables in their parent element.

## Why
The new data-attributes added to the links in the footer allow the region of the footer to be more easily determined from analytics themselves rather than having to check on the template. The attributes are in the following format and are on all links in the footer (including the copyright links). 

```
  data-track-category: footerClicked
  // this depends on where in the footer the url is
  data-track-action: topicsLink || governmentActivityLink || supportLink || copyrightLink
  data-track-label: <link destination URL>
  data-track-options: {
   dimension29: <link text>
  }
```

[Relevant Trello Card](https://trello.com/c/sn4of3m9/336-implement-consistent-analytics-tracking-on-the-sitewide-footer)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Update LUX to v301 ([PR #2773](https://github.com/alphagov/govuk_publishing_components/pull/2773))
 * Fix http protocol reporting in Safari ([PR #2781](https://github.com/alphagov/govuk_publishing_components/pull/2781))
+* Tracking changes for the footer ([PR #2774](https://github.com/alphagov/govuk_publishing_components/pull/2774))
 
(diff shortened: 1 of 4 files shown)


andysellick (commented):
There doesn't seem to be a test file for `public_layout_helper.rb` yet but I think it might be good to create one now?

MartinJJones (approved):
Changes and new test file look good to me, nice work

PR #2778, opened 2022-05-16
Update analytics logic for new browse page metatags

patrickpatrickpatrick:
## What
Update the conditional statement in `page-content.js `which checks to see if a page is a new browse page. Change logic of counting sections in Browse Level 2 templates to count the number of sections instead of returning the value '1'.

## Why
The values of metatags for new browse pages were changed which made previously written conditional statements in `page-content.js `always return false. The conditional statements have been updated to use the new metatag values so that now counting of sections/links should work on all browse page levels.

Initially the logic for counting sections on Browse Level 2 templates assumed that default pages consisted of 1 section. This is sometimes not the case (there can be related topics) so the logic has been changed to count the number of sections on the page.

[Relevant Trello Card](https://trello.com/c/REnMbbdH/993-dimensions-26-and-27-for-new-topics-level-0-1-pages-not-showing)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update analytics logic for new browse page metatags ([PR #2778](https://github.com/alphagov/govuk_publishing_components/pull/2778))
 * Tracking changes for the footer ([PR #2774](https://github.com/alphagov/govuk_publishing_components/pull/2774))
 
 ## 29.9.0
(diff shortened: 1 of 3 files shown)


andysellick (commented):
I think this looks good - does the page-content test need updating?

patrickpatrickpatrick:
> I think this looks good - does the page-content test need updating?

don't think there ever was a test written for it, should i add it

andysellick:
@patrickpatrickpatrick there's already a test for `page-content.js`, which I think includes this, so it'll probably just need extending/tweaking to reflect your changes.

MartinJJones (approved):
Nice work!, the new test added looks good as well 👍

PR #2781, opened 2022-05-17
Fix HTTP protocol reporting in Safari

injms:
## What

Fix the HTTP protocol reporting into Speedcurve RUM as Safari was not reporting any data.

This used `DOMContentLoaded` to make sure that `performance.getEntriesByType()` was not undefined. However, this was not reporting anything in Safari. It appears that the event listener was being added after the event was fired.

Checking the `readyState` of the document lets us see whether the document is loading or ready - and then, depending on the state, either the `load` event has a listener attached or the measurement is done immediately.

Adds extra check to make sure that `performance.getEntriesByType` is not undefined - so can be used without throwing an error.

## Why

Monitoring the type of HTTP connection that a user's using will allow us to see how the different HTTP protocols affect performance - with Safari missing a large chunk of users were being missed out from this investigation.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Update LUX to v301 ([PR #2773](https://github.com/alphagov/govuk_publishing_components/pull/2773))
+* Fix http protocol reporting in Safari ([PR #2781](https://github.com/alphagov/govuk_publishing_components/pull/2781))
 
 ## 29.8.0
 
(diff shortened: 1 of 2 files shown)


Nooshu (approved):
This looks great Ian, thanks for adding in the fix! It should hopefully cut down on the errors we were seeing in RUM. Also, the Safari Fix too. I was wondering why we weren't seeing any Custom Data.

PR #2786, opened 2022-05-20
Extend GTM analytics click tracking

andysellick:
## What
Extend the GTM click tracking, specifically:

- if an element contained an expanding element (indicated by the aria-expanded attribute) information about that was included in the dataLayer push
- this now checks if the clicked element itself has that attribute
- will be used on the accordion show/hide all element

Also updates the accordion component:

- allows data attributes to be added to the show/hide all link
- this means we can trigger the GTM click tracking on this element and include the expanded/collapsed state, all without any custom analytics code in the component

## Why
We want to track the state of expanding/collapsing elements on user interaction. At the moment there's some custom code that does this, but we should be able to do it by passing data attributes and having the overall module be smart enough to handle it without additional custom code.

## Visual Changes
None.

Trello card: https://trello.com/c/V0RFrSMK/43-test-technical-approach-to-tracking-on-tabs-and-accordions

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Extend GTM analytics click tracking ([PR #2786](https://github.com/alphagov/govuk_publishing_components/pull/2786))
 * GOVUK Frontend 4.1.0 updates ([PR #2794](https://github.com/alphagov/govuk_publishing_components/pull/2794))
 
 ## 29.10.0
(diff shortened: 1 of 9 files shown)


andysellick:
@gclssvglx I've had a think about the point you raised about the `checkExpandedState` function and I can't currently think of a better way of structuring it. I think the problem is:

- it needs to check either for the existence of an attribute or a child element that has that attribute (which returns two different types of values)
- it needs to return either the value of that attribute (which is a string of 'true' or 'false') or return nothing, i.e. `false`

I don't want to pressure you to approve, but I think this code is going to be rewritten a lot long before it gets to production, so we're going to have lots of opportunity to improve it.

gclssvglx (approved):
Looking good - nice job @andysellick real noodle twister 👍

PR #2788, opened 2022-05-20
Add SVG thumbnail styles for Whitehall

benjamineskola:
## What

Adding CSS for SVG attachment thumbnail icons now being used by Whitehall.

## Why

We've added some SVG icons to Whitehall in alphagov/whitehall#6550 as a transitional step towards using the attachment components. However, because they're not rendered by whitehall itself, the CSS to style these icons needs to be available in other frontend apps, so this would appear to be the correct place for it (although adding to the legacy Whitehall-supporting css is obviously not ideal).

This is actually trying to remain close to the style of the attachment component, without depending on the component 

### Before

The original icon that we're replacing:

Screenshot 2022-05-20 at 17 09 59

### After

The icon properly styled, when the CSS in this PR is applied:

Screenshot 2022-05-19 at 10 27 09

The icon improperly styled:

Screenshot 2022-05-19 at 10 22 25

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -37,7 +37,8 @@
       margin-left: -($thumbnail-width + govuk-spacing(6) - $govuk-border-width);
       padding-bottom: govuk-spacing(3);
 
-      img {
+      img,
+      svg {
         display: block;
         width: $thumbnail-width;
         height: 140px;
@@ -55,6 +56,11 @@
         }
 
         box-shadow: 0 2px 2px rgba(govuk-colour("black"), .4);
+      }
+
+      svg {
+        fill: govuk-colour("mid-grey", $legacy: "grey-3");
+        stroke: govuk-colour("mid-grey", $legacy: "grey-3");
       }
     }
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -56,6 +56,23 @@
...
+        background: govuk-colour("white");
+        box-shadow: 0 2px 2px rgba(govuk-colour("black"), .4);
+        fill: govuk-colour("mid-grey", $legacy: "grey-3");
+        stroke: govuk-colour("mid-grey", $legacy: "grey-3");
+
+        @include govuk-if-ie8 {
+          max-width: none;
+          border: $govuk-border-width solid govuk-colour("mid-grey", $legacy: "grey-3");
+        }
+      }
```
DilwoarH, line 75: Given that the change is basically a duplicate of the above, I would suggest doing the following:

Change [line 40](https://github.com/alphagov/govuk_publishing_components/pull/2788/files#diff-5315bc6594eb2695f5ff5d7a3585e58b342f1bf830fc16bb995f6e1dbb6eeb11R40) from `img` to `img, svg`.

and have the following changes for SVG specific:

```suggestion

      svg {
        fill: govuk-colour("mid-grey", $legacy: "grey-3");
        stroke: govuk-colour("mid-grey", $legacy: "grey-3");
      }
```

DilwoarH (changes requested):
Looks good - one small change and also change notes need to be added to the following section: 
https://github.com/alphagov/govuk_publishing_components/blob/main/CHANGELOG.md#unreleased

DilwoarH (changes requested):
All good, just the change notes and happy to approve

PR #2794, opened 2022-05-24
GOVUK Frontend 4.1.0 updates

andysellick:
## What
Make some minor changes to our component markup following the release of `govuk-frontend` 4.1.0, specifically:

- remove 'aria-live' attribute from character count component (not needed anymore)
- remove 'pattern' attribute from date input component (not needed anymore)

## Why
We don't import markup for components from `govuk-frontend`, we've hard coded it into `.erb` templates and heavily modified it. This means that when we get a new release of `govuk-frontend` we need to check the release notes to see if anything has changed. For this release only two minor things have changed.

- https://github.com/alphagov/govuk-frontend/releases/tag/v4.1.0
- https://github.com/alphagov/govuk_publishing_components/blob/main/docs/upgrade-govuk-frontend.md

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* GOVUK Frontend 4.1.0 updates ([PR #2794](https://github.com/alphagov/govuk_publishing_components/pull/2794))
+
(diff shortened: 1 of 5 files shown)


patrickpatrickpatrick (approved):
Looks good to me. Will make sure to check next time before merging in future.

PR #2795, opened 2022-05-25
Move the emergency_banner from static

patrickpatrickpatrick:
## What

Move the [`emergency_banner`](https://docs.publishing.service.gov.uk/repos/static/emergency-banner.html) component from [`static`](https://github.com/alphagov/static) to the [`govuk_publishing_components`](https://github.com/alphagov/govuk_publishing_components/) gem. 

### What Has Been Done

Move the templates, documentation, stylesheets to this gem from `static`. Write new tests for the "new" component.

## Why

Part of an ongoing effort to refactor static and retire slimmer as outlined [here](https://trello.com/c/zLiC5VwG/7-objective-retire-static-and-slimmer).


app/assets/stylesheets/govuk_publishing_components/components/_emergency-banner.scss
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  .homepage & {
+    border-bottom: govuk-spacing(1) solid govuk-colour("white");
+    border-top: govuk-spacing(1) solid govuk-colour("white");
+    margin-bottom: (0 - govuk-spacing(2));
+    margin-top: 0;
+    position: relative;
+    z-index: 10;
+  }
```
injms, line 19: Having the external `homepage` class affect the styles breaks the component principle of isolation[^1], specifically:

> it does not rely on external selectors to style its children

The only way around this I can think of doing this is by having this variation set by a parameter into the component, for example:

```erb
<%= render "govuk_publishing_components/components/emergency_banner", {
  campaign_class: "notable-death",
  heading: "His Royal Highness Henry VIII",
  short_description: "1491 to 1547",
  link_text: "Override Link Text",
  link: "https://www.gov.uk/",
  homepage: true,
} %>
```
 and then setting up a [homepage specific template in static](https://github.com/alphagov/static/tree/main/app/views/root) (`gem_layout_homepage_only.html.erb` for example) that only the homepage template in the `frontend` app uses. 

It's a bit more work now - but it keeps the isolation principle, and adding this functionality will be one less thing to need to update as Static is reduced in size.

[^1]: https://github.com/alphagov/govuk_publishing_components/blob/cec4f2f50dc0187e19d4e4a49c550ff0f7f6c4be/docs/component_principles.md#a-component-is-isolated-when

injms (changes requested):
A couple of things need changing before this can be merged in - details in the comments.

If possible, can the commits be more descriptive about the why of the changes? The what is covered by the code changes themselves, so needs less explanation; but the why is important to help future us know the context of the changes. For example, why does this particular change help reduce the reliance on Static and Slimmer?

For this component, the code copied across (and no other changes where possible) could be a single commit so it's clear what's been copied from Static. Then subsequent commits can be the improvements you're making - this gives a clearer indication of what's gone on, what you've been able to improve, what other things have been tried, what is trickier to iterate on, and what's out of scope.

patrickpatrickpatrick:
in line with @injms [suggested changes](https://github.com/alphagov/govuk_publishing_components/pull/2795#discussion_r896566760), a new PR in https://github.com/alphagov/frontend/pull/3260 has been opened

app/assets/stylesheets/govuk_publishing_components/components/_emergency-banner.scss
```diff
@@ -0,0 +1,74 @@
...
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(6) 0;
+  }
+}
+
+.gem-c-emergency-banner--homepage {
+  border-bottom: govuk-spacing(1) solid govuk-colour("white");
+  border-top: govuk-spacing(1) solid govuk-colour("white");
+  margin-bottom: (0 - govuk-spacing(2));
```
injms, line 16: This could be clearer if a [negative number is used][1]:

```suggestion
  margin-bottom: govuk-spacing(-2);
```

[1]: https://frontend.design-system.service.gov.uk/sass-api-reference/#example-using-negative-spacing

app/views/govuk_publishing_components/components/docs/emergency_banner.yml
```diff
@@ -0,0 +1,51 @@
...
+      When presenting any type of emergency banner on homepage the styling is slightly different.
+    data:
+      campaign_class: "notable-death"
+      heading: "His Royal Highness Henry VIII"
+      short_description: "1491 to 1547"
+      link: "https://www.gov.uk/"
+    embed: |
+      <div class="homepage">
+        <%= component %>
+      </div>
```
injms, line 51: I think the homepage example needs updating with the `homepage` parameter.

app/assets/stylesheets/govuk_publishing_components/components/_emergency-banner.scss
```diff
@@ -0,0 +1,74 @@
...
+
+  @include govuk-media-query($from: tablet) {
+    padding: govuk-spacing(6) 0;
+  }
+}
+
+.gem-c-emergency-banner--homepage {
+  border-bottom: govuk-spacing(1) solid govuk-colour("white");
+  border-top: govuk-spacing(1) solid govuk-colour("white");
+  margin-bottom: (govuk-spacing(-2));
```
injms, line 16: Does the `govuk-spacing(-2)` need to be wrapped in brackets?

app/views/govuk_publishing_components/components/docs/emergency_banner.yml
```diff
@@ -0,0 +1,47 @@
...
+body: |
+  There are three types of events that would lead GOV.UK to add an emergency banner to the top of each page on the web site; a notable death, a national emergency or a local emergency.
+
+  See the [opsmanual](https://docs.publishing.service.gov.uk/manual/emergency-publishing.html#adding-emergency-publishing-banners) for information about what the Emergency Banner is and when it should be deployed.
+
+accessibility_criteria: |
+  The links must:
+
+  * be keyboard focusable
+
```
injms, line 12: Sorry, I've just spotted this - I think this is the last thing!

There's a `shared_accessibility_criteria` that can be used when a component uses links and/or buttons:

```suggestion
shared_accessibility_criteria:
- link
```

This can then show the full `link` accessibility criteria from:

https://github.com/alphagov/govuk_publishing_components/blob/22d85c2c96e98ac6534260c84f02f07d7cf38269/app/models/govuk_publishing_components/shared_accessibility_criteria.rb#L3-L18

patrickpatrickpatrick: oh that's useful!

patrickpatrickpatrick:
have tested these changes in integration

patrickpatrickpatrick:
> One additional thing that needs to be changed (which won't appear in tests and which I am writing here also so I don't forget) is that the local emergency background colour needs to be updated to ensure a high enough contrast between the text and the background. I have asked for a designer's input on this and they are going to get back to me on a new colour.
> 
> Screenshot 2022-06-15 at 14 41 48

this has now been changed to a new colour
Screenshot 2022-07-05 at 16 26 56

PR #2797, opened 2022-05-25
Refactor the accordion component code

andysellick:
## What
Refactor some things in the accordion code. Specifically:

- remove some variables that weren't used
- stop having class name variables that don't start with `.` except it's needed every time the variable is used
- rename the class name and element selector variables to clearly distinguish them
- remove the class that was added on JS initialisation as it doesn't do anything except allow a test to pass - we should have a better way of testing our code than that

## Why
I made some unrelated changes in the accordion recently and this was bothering me. 

I'm 99.9% certain the variables I've removed aren't used but worth double checking. I think we used to have the ability to pass translated text to the component for the 'show/hide' and 'show/hide all' links (since they're created by JS, which can't access translations from Rails) but don't have that since now most of the accordion component JS is managed by code from `govuk-frontend`. At some point it might be good to think about how to bring that functionality back.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -25,7 +25,6 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
   GemAccordion.prototype.init = function () {
     // Indicate that JavaScript has worked
-    this.$module.classList.add('gem-c-accordion--active')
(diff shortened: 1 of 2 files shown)


querkmachine:
> I think we used to have the ability to pass translated text to the component […] At some point it might be good to think about how to bring that functionality back.

We're working on that at the moment! https://github.com/alphagov/govuk-frontend/issues/1714

Signed,
Your friendly neighbourhood Design System team.

andysellick:
Brilliant, thanks @querkmachine!

AshGDS (approved):
Looks good to me @andysellick :+1:

PR #2801, opened 2022-05-27
Improve documentation

andysellick:
## What
Minor tweaks to some of our documentation.

## Why
Some areas of our documentation were not as clear as they could be.

## Visual Changes
Changes to docs only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,19 +35,30 @@ yarn run jasmine:browser
 yarn run jasmine:ci
 ```
 
-### Further documentation
+### Using the gem
 
 - [Install and use this gem](docs/install-and-use.md)
 - [Use a component in your application](docs/use-components.md)
-- [Generate a new component](docs/generate-a-new-component.md)
-- [Testing a component](docs/testing-components.md)
-- [Move a component from an application to the gem](docs/moving-components-upstream-into-this-gem.md)
+
+### Managing the gem
+
 - [Publish/release a new version of the gem](docs/publishing-to-rubygems.md)
 - [Keep this gem in sync with the Design System](docs/upgrade-govuk-frontend.md)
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -37,17 +37,30 @@ yarn run jasmine:ci
 
 ### Further documentation
 
+Using the gem
```
injms, line 40: Do you think that this should be a heading?

```suggestion
#### Using the gem
```

docs/auditing.md
```diff
@@ -2,6 +2,8 @@
 
 The component guide includes some auditing of components and applications that can provide useful information. It can be found at `/component-guide/audit`.
 
+It is designed to be run locally as it gets information about applications and how they use components from local files. Note that if your local copy of an application is not up to date the information reported by the auditing may be incorrect.
```
injms, line 5: I think that the local apps need to be up to date is worth highlighting in a way that can't be missed - not necessarily as I've suggested, just in any way that highlights it strongly:

```suggestion
It is designed to be run locally as it gets information about applications and how they use components from local files. 

> **If your local copy of an application is not up to date the information reported by the auditing may be incorrect**
```

README.md
injms: Or add a colon at the end:
```suggestion
Using the gem:
```

andysellick: Have updated the headings here a bit based on this suggestion, hopefully a bit clearer now.

docs/auditing.md
andysellick: Thanks, have highlighted this (I couldn't think of a better way than you'd suggested, it's clear at least).

injms (approved):
Smashing - thanks!

PR #2802, opened 2022-05-27
Add date of birth autocomplete option for date input

jon-kirwan:
## What

Add a new `dob_autocomplete_enabled` API attribute, when set to true, this will set the `autocomplete` attribute on the 3 date input fields to `bday-day`, `bday-month` and `bday-year`.


app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -2,11 +2,11 @@
   id ||= "input-#{SecureRandom.hex(4)}"
   name ||= nil
   items ||= [
-    { :name => "day", :width => 2 },
-    { :name => "month", :width => 2 },
-    { :name => "year", :width => 4 }
+    { :name => "day", :width => 2, autocomplete: "bday-day" },
```
maxgds, line 5: I think I'd rather see these autocomplete strings only added in the case where the date is a DOB and autocomplete is on, rather than added here and removed in many cases. It feels more expandable if we were to ever have alternate autocomplete patterns, and generally cleaner.

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -26,6 +26,15 @@ examples:
       component as follows: `custom-name[day]`, `custom-name[month]`, `custom-name[year]`
     data:
       name: "dob"
+  with_autocomplete_for_date_of_birth_fields:
+    description: |
+      Use the `autocomplete_date_of_birth` option when you're asking for a date of birth. This lets browsers autofill the information on a user’s behalf if they've entered it previously.
```
maxgds, line 31: Please replace "a date of birth" with "the user's date of birth".

maxgds (changes requested):
Nothing major, but some small tweaks requested.

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -26,6 +26,15 @@ examples:
...
+
+      Note that the option should only be passed to the component if the date of birth asked for is the date of birth of the person filling the form in and not asked for on behalf of someone else.
+
+      Current browser support for the date of birth values is limited although it seems to have better support in third party applications.
```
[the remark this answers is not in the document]
jon-kirwan: I've now removed the line from the documentation. This is added in the pull request above under 'Anything else'

app/views/govuk_publishing_components/components/_date_input.html.erb
jon-kirwan: I think I've addressed the changes and have moved the autocomplete values to where the component options are passed in -

`autocomplete: autocomplete_date_of_birth ? "bday-#{item[:name]}" : nil`

Added changes under a second temporary commit.

maxgds (approved):

PR #2803, opened 2022-05-31
Add js search to table component

davidtrussler:
## What
Updates the table component to add a `With filter` option.

## Why
This was previously existing functionality that requires restoration. See [Add JS Search to Table Component](https://trello.com/c/cNYAJjvM/443-add-js-search-to-country-list-in-travel-advice)

## Visual Changes

The new component is added to the table section of the [Component Guide](https://components.publishing.service.gov.uk/component-guide/table) and appears as below. 

Screenshot 2022-05-31 at 16 39 43

The component displays results in response to user input as shown below

Screenshot 2022-05-31 at 16 43 47

Screenshot 2022-05-31 at 16 44 23

Screenshot 2022-05-31 at 16 45 09


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Updates the table component to add a `With filter` option ([PR #2803](https://github.com/alphagov/govuk_publishing_components/pull/2803))
```
andysellick, line 12: CHANGELOG entries should match the title of the PR.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,44 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function TableFilter ($module) {
+    this.$module = $module
+    this.searchInput = $module.querySelector('input[name="filter"]')
+    this.tableRows = $module.querySelector('table').querySelectorAll('.govuk-table__body .govuk-table__row')
```
andysellick, line 8: Does this code need to consider whether any of these elements might not be found? I don't know if there's a scenario where they wouldn't, but for example if there was no `table` element then this line would error.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,63 @@
...
 %>
 
-<%= GovukPublishingComponents::AppHelpers::TableHelper.helper(self, caption, {
-    sortable: sortable,
-    caption_classes: caption_classes
-  }) do |t| %>
+<div <% if filterable %> data-module="table-filter" <% end %> >
+  <% if filterable %>
+    <div class="govuk-table-filter">
+      <%= render "govuk_publishing_components/components/input", {
```
andysellick, line 14: Something to flag for deployment - because this component now relies on another component, we need to make sure that every application that uses this component also includes the assets for the input component. Although in this case it's fine, because `static` includes the assets for the input component already, so no need to worry. I'm just raising this for awareness.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,63 @@
...
+        label: {
+          text: "Search"
```
andysellick, line 16: All component text should be added to and then pulled in from the translation files (`en.yml`) e.g.

```erb
<%= t("components.table.filter_label") %>
```

Same for the 'no results' text.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,63 @@
...
+        },
```
andysellick, line 17: Does `Search` adequately convey the meaning of this control?

Also `Search` and `Filter` seem to be used interchangeably in a few places across this PR, but I think users might consider them to be different things and therefore this could be confusing.

app/views/govuk_publishing_components/components/_table.html.erb
andysellick, line 14: Have you thought about how this filter will work with a screen reader? I've just tested it in Voiceover and there doesn't seem to be any feedback to indicate what's happening when you type something into the box.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -16,6 +16,19 @@ $table-row-even-background-colour: govuk-colour("light-grey", $legacy: "grey-4")
...
+  display: none;
+}
+
+.govuk-table__message {
+  @include govuk-font($size: 19);
+}
+
+.govuk-table__row--hidden,
+.govuk-table__message--hidden {
+  display: none;
```
andysellick, line 29: I suspect that there's already a class in `govuk-frontend` available to us that will hide things. Might be worth asking in the Design System slack channel, might save you a few lines of CSS here.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,63 @@
...
+  filterable ||= false
 %>
 
-<%= GovukPublishingComponents::AppHelpers::TableHelper.helper(self, caption, {
-    sortable: sortable,
-    caption_classes: caption_classes
-  }) do |t| %>
+<div <% if filterable %> data-module="table-filter" <% end %> >
+  <% if filterable %>
+    <div class="govuk-table-filter">
```
andysellick, line 13: These changes will need to be tested in the [table spec file](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/components/table_spec.rb) as well. You'd probably want to check that if the filter option is turned on that the input is rendered, and maybe test that the 'no results' text is rendered as well, maybe that the div has the right data-module on it, etc.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -16,6 +16,19 @@ $table-row-even-background-colour: govuk-colour("light-grey", $legacy: "grey-4")
   color: $govuk-secondary-text-colour;
 }
 
+.govuk-table-filter {
+  display: none;
+}
+
+.govuk-table__message {
+  @include govuk-font($size: 19);
```
andysellick, line 24: I think you could remove this class entirely and use `govuk-body` in the template instead.

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,100 @@
+/* eslint-env jasmine, jquery */
```
andysellick, line 1: I think you can remove `jquery` from this line as I didn't see any jQuery in this file.

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,100 @@
...
+    expect(rows[1].classList.contains(hiddenRowClass)).toBe(false)
+    expect(message.classList.contains(hiddenMessageClass)).toBe(true)
+
+    // User enters '£3'
+    searchInput.value = '£3'
+    tableFilter.updateRows()
+    expect(rows[0].classList.contains(hiddenRowClass)).toBe(true)
+    expect(rows[1].classList.contains(hiddenRowClass)).toBe(false)
+    expect(message.classList.contains(hiddenMessageClass)).toBe(true)
+  })
```
andysellick, line 99: Should there also be a test that checks that everything reappears if the user clears the content of the filter/search input?

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,100 @@
...
+
+  it('Sets the correct classes when the `updateRows` method is called', function () {
+    var rows = document.querySelectorAll('.govuk-table__row')
+    var message = document.querySelector('.govuk-table__message')
+    var hiddenRowClass = tableFilter.hiddenRowClass
+    var hiddenMessageClass = tableFilter.hiddenMessageClass
+
+    // User enters 'b'
+    searchInput.value = 'b'
+    tableFilter.updateRows()
```
andysellick, line 60: Why do these tests need to explicitly call the `updateRows` function? If the code has been initialised properly then isn't this unnecessary? I think you should be able to call `tableFilter.init()` once instead of explicitly calling the `updateRows` function multiple times.

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,100 @@
...
+      '</div>'
+
+    window.setFixtures(FIXTURE)
+
+    element = document.querySelector('[data-module="table-filter"]')
+    searchInput = document.querySelector('input')
+    tableFilter = new GOVUK.Modules.TableFilter(element)
+  })
+
+  it('Calls the `updateRows` method when the user types in the input field', function () {
```
andysellick, line 42: Why is it important to test that a specific function within the code has been called? Is it not enough to test that the expected outcome happens when certain things are put into the search box? (which is what the next test does anyway)

andysellick (changes requested):
Looks like a solid start but I've raised a few questions and I think some more consideration for accessibility might be needed.

CHANGELOG.md
davidtrussler: Updated 👍

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,44 @@
...
+    this.filter = $module.querySelector('.govuk-table-filter')
```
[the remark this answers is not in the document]
davidtrussler: Done 👍

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
davidtrussler: I think that if this component is being rendered the elements will by definition all be present. Additionally I have not been able to find any equivalents in any existing components.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -16,6 +16,19 @@ $table-row-even-background-colour: govuk-colour("light-grey", $legacy: "grey-4")
   color: $govuk-secondary-text-colour;
 }
 
+.govuk-table-filter {
```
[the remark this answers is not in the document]
davidtrussler: I have renamed classes for elements added in this work as suggested by the naming convention.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,63 @@
...
-  sortable ||=false
+  sortable ||= false
+  filterable ||= false
 %>
 
-<%= GovukPublishingComponents::AppHelpers::TableHelper.helper(self, caption, {
-    sortable: sortable,
-    caption_classes: caption_classes
-  }) do |t| %>
+<div <% if filterable %> data-module="table-filter" <% end %> >
```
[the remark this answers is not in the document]
davidtrussler: Good point. I’ve used the suggested capture block here which tidies this up quite a bit. I’m still not wholly convinced the structure of the whole component is ideal though since we are adding a couple of new elements outside of the component itself.

app/views/govuk_publishing_components/components/_table.html.erb
davidtrussler: Makes sense: the text has been added to the translation files (keys only in locales other than `en`)

app/views/govuk_publishing_components/components/_table.html.erb
davidtrussler: This has been updated for consistency. I think though that the actual text that should be displayed and the style in which it is rendered is more a design specialism.

app/views/govuk_publishing_components/components/_table.html.erb
davidtrussler: I have added a new test to the RSpec file for this component. My knowledge of RSpec is minimal though so this will need a more experienced look.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
davidtrussler: Yes, good call. Ta.

spec/javascripts/components/table-filter-spec.js
davidtrussler: I’m not sure about this. I would say that scenario is already covered by the existing instances and I have also been removing unnecessary and over-zealous testing. Very happy to discuss the test coverage and how we should approach that more broadly in general though (see below).

spec/javascripts/components/table-filter-spec.js
davidtrussler: I think this and the following comment are two sides of the same coin: it is about separating the initialisation of the component from the methods that are called when it has been initialised so that there is not a dependency between them. 

I’ve reduced the number of test cases for this and clarified what they are testing but for now I have retained a call to the method. 

I am aware that this is in effect testing a private method (i.e. one that would not be accessed by a user) but remain to be convinced that this is not viable in this case though I’d be interested in other thoughts on this.

spec/javascripts/components/table-filter-spec.js
davidtrussler: We are testing here that the event listener has been added when the component is initialised. I have updated the description to better reflect that. The test that follows deals with the outcome of calling the method. By using separate tests for these two things we are breaking the dependency of one on the other. I’m happy to be convinced otherwise but for me that is a more valuable way to test the overall functionality: if the component as a whole breaks it is clearer where the problem has occurred.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,44 @@
...
+  }
+
+  // Reads value of input and filters content
+  TableFilter.prototype.updateRows = function () {
+    var value = this.searchInput.value
+    var hiddenRowClass = this.hiddenRowClass
+    var hiddenRows = 0
+
+    this.tableRows.forEach(function (row) {
+      row.classList.remove(hiddenRowClass)
```
jon-kirwan, line 28: Unfortunately, IE11 doesn't support `forEach` on a NodeList https://developer.mozilla.org/en-US/docs/Web/API/NodeList/forEach and it doesn't look like we have a polyfill here included... so I think you'll have to use a for loop here (or convert the NodeList to an Array).

ie11-screen-capture

app/views/govuk_publishing_components/components/_table.html.erb
jon-kirwan: Yes, I think an aria-live attribute could be added to the parent `div`.

```
<% if filterable %>
  <div data-module="table-filter" aria-live="polite">
  ...
```

andysellick: And maybe something to associate the search input with the table?

spec/javascripts/components/table-filter-spec.js
andysellick: That makes sense.

spec/javascripts/components/table-filter-spec.js
andysellick: I suppose my question would be what is the value of such an approach? There isn't a scenario where separation of the two things you're talking about would happen. You're testing a private method and in doing so you're not testing that the JS as an end-to-end unit works fully i.e. once initialised, user input causes the right effect. I'd prefer to test modules as realistically as possible - this is a small example, but applied to a larger block of code could introduce all kinds of brittleness.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
davidtrussler: Great, thanks, I've used that in place of `display: none` here.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,44 @@
...
+    this.filter = $module.querySelector('.js-gem-c-table__filter')
+    this.message = $module.querySelector('.js-gem-c-table__message')
+    this.hiddenRowClass = 'govuk-table__row--hidden'
+    this.hiddenMessageClass = 'gem-c-table__message--hidden'
+  }
+
+  TableFilter.prototype.init = function () {
+    this.$module.updateRows = this.updateRows.bind(this)
+    this.filter.style.display = 'block'
+    this.searchInput.addEventListener('input', this.$module.updateRows)
```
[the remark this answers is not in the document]
davidtrussler: In fact this has now been replaced with the hidden class in a previous change.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
davidtrussler: Oops, my bad - I kinda thought about that then forgot. Also didn’t test cross browser! I’ve rewritten that now and tested it more fully.

spec/javascripts/components/table-filter-spec.js
davidtrussler: I have updated the test to more explicitly check the effects of user input.

app/views/govuk_publishing_components/components/_table.html.erb
davidtrussler: I’ve been looking into this and we discussed it at the FE meeting yesterday as well. There’s a couple of things to be taken into consideration: 
- As far as I can see the only changes to the table that a screen reader (or VoiceOver anyway) responds to is content changes. So in this case not changing the display value of a table row or even adding/removing it from the DOM (surprisingly) will trigger that. But we are not changing the content of a table row here. 
- It may be considered anyway that wrapping an entire table in a live region would cause a potentially massive amount of content to be related to the user which is not really acceptable.

A similar implementation was done [here](https://www.gov.uk/foreign-travel-advice) but in this case we only update a count of the number of rows displayed as a result of user action which would satisfy both of the points above. There is no requirement here to report the number of rows so unless we add that or have that as a kind of visually hidden element I am struggling to know how to deal with this.

andysellick: I think the key thing is that when a user types into the input something should tell them that they've made an impact on the UI somehow. Otherwise a screen reader user could be typing and they have no idea if what they're doing is actually working.

I like the idea of some visually hidden text that updates with a count of matching rows, that seems better than reading aloud the whole table (which could take some time, and also not clearly immediately show what has changed).

spec/javascripts/components/table-filter-spec.js
andysellick: Thanks for adding that. Unfortunately I think I can foresee a problem. Correct me if I'm wrong, but it looks like the general state of the test fixture isn't reset after each test - you're using a `beforeAll` to setup things once. Since the tests run in a random order, it's entirely possible that the reset test you've added could be run first, in which case it doesn't do anything - the input value is already empty.

For this specific test I'd recommend doing a search first, checking the results are as expected, then clearing the input and checking that everything has been reset to normal.

As a general strategy I'd also suggest that your tests be as self contained as possible to avoid race conditions that could incorrectly pass or fail following tests. Usually we recreate the fixture and any associated variables for each test using a `beforeEach` to ensure that the state of the code is as known as possible for each test.

spec/javascripts/components/table-filter-spec.js
andysellick: Great, thanks.

andysellick:
I've approved the visual changes to this PR. One was the thing that's being added (which is fine, obviously) and the other is a slightly weird one to do with the image card component, which is unrelated to this change and is occurring on a few other PRs at the moment.

app/views/govuk_publishing_components/components/_table.html.erb
davidtrussler: I’ve added an element here to display (to assistive technology) the number of rows that are shown. It’s pretty similar to the way this is done on the Travel Advice page.

jon-kirwan: I think this is fine.

There are a few differences between VoiceOver, NVDA and JAWS here:
- JAWS is fine.
- NVDA and Chrome is fine
- NVDA and Firefox. Changes are not announced (which is the same behaviour seen on https://www.gov.uk/foreign-travel-advice). I'm not sure if that's an issue with NVDA and Firefox and `aria-live`
- VoiceOver also seems fine although sometimes the announcement is interrupted.

davidtrussler: Thanks for looking at that @jon-kirwan, especially on other Screen Readers. I only tested it with VoiceOver and I noticed it was a little bit inconsistent but it’s the same as the Travel Advice page so I guess that’s acceptable.

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,121 @@
...
+    })
+
+    it('Sets up the event listener', function () {
+      searchInput.dispatchEvent(event)
+      expect(tableFilter.updateRows).toHaveBeenCalledTimes(1)
+    })
+  })
+
+  describe('When the user inputs a value', function () {
+    beforeEach(function () {
```
andysellick, line 70: Is there any significant difference in the two `beforeEach` functions you've defined here? They look broadly similar although the second one has some extra bits (but nothing that contradicts the other?). If so, you could write one and put it at the top of the file for all tests?

spec/javascripts/components/table-filter-spec.js
```diff
@@ -0,0 +1,121 @@
...
+
+    it('Shows the correct rows when the case of the input term differs from the table row content', function () {
+      searchInput.value = 'OP'
+      searchInput.dispatchEvent(event)
+      expect(rows[0].classList.contains(hiddenClass)).toBe(true)
+      expect(rows[1].classList.contains(hiddenClass)).toBe(false)
+      expect(filterCount.textContent).toEqual('Number of filtered rows: 1')
+    })
+
+    it('Shows all the rows when the input is cleared', function () {
```
andysellick, line 109: This test is much better, but can you also check that the hidden number of rows text changes when the query is entered and then is reset when the query is cleared? I know that other tests check that it gets set correctly with matching and non-matching queries, but this is the only place where the test clears the input and we should check that everything returns to normal as expected.

andysellick (commented):
Tests look much better, good stuff. Still seem to be some problems with getting screen readers to announce the hidden text properly though, and I think I've found a few ways to optimise the JS a little.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,55 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  var SETTINGS = {
```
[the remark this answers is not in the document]
davidtrussler: Yes, makes sense. I've updated that. I just copied the way the SETTINGS variable was set from elsewhere but it's not really very necessary in this instance.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,55 @@
...
+  TableFilter.prototype.init = function () {
+    this.$module.updateRows = this.updateRows.bind(this)
+    this.filter.classList.remove(this.hiddenClass)
+    this.searchInput.addEventListener('input', this.$module.updateRows)
+  }
+
+  // Reads value of input and filters content
+  TableFilter.prototype.updateRows = function () {
+    var value = this.searchInput.value
+    var hiddenClass = this.hiddenClass
```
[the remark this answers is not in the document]
davidtrussler: Yes, good point. That was left over from an earlier iteration. 👍 
I've removed the variable declaration.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,55 @@
...
+    var hiddenRows = 0
+    var length = this.tableRows.length
+
+    for (var i = 0; i < length; i++) {
+      var row = this.tableRows[i]
+
+      row.classList.remove(hiddenClass)
+
+      if (!row.textContent.toUpperCase().includes(value.toUpperCase())) {
```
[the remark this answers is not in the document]
davidtrussler: Makes sense. Refactored accordingly.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,55 @@
...
+        row.classList.add(hiddenClass)
+        hiddenRows++
+      }
+    }
+
+    this.filterCount.textContent = this.filterCount.getAttribute(SETTINGS.text.count) + ': ' + (length - hiddenRows)
```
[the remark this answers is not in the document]
davidtrussler: Good point, I have updated the message announced along these lines.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,71 @@
...
+    <div class="js-gem-c-table__filter govuk-!-display-none">
+      <%= render "govuk_publishing_components/components/input", {
+        label: {
+          text: t("components.table.filter_label")
+        },
+        name: "filter",
+        controls: "filterCount"
+      } %>
+
+      <p id="filterCount" aria-live="polite" class="js-filter-count govuk-visually-hidden" data-count-text="<%= t("components.table.filter_count") %>"></p>
```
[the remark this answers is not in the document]
davidtrussler: I’ve updated this a little bit along the lines of the implementation on the [GOVUK Search page](https://www.gov.uk/search). This seems to be something of an improvement on the current version. Mostly this involves wrapping the `aria-live` region around the whole filter section (including both the input and the message relating the number of rows updated by the user’s input) and changing the level of this to `assertive`.

app/assets/javascripts/govuk_publishing_components/components/table-filter.js
```diff
@@ -0,0 +1,52 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
```
andysellick, line 3: Sorry, but I've just realised that the name of any JS file for a component needs to match the component name (`table.js`) in order for the component auditing to pick up that it exists. Can you rename it?

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,72 @@
...
           <% end %>
         <% end %>
       <% end %>
     <% end %>
   <% end %>
+<% end %>
+
+<% if filterable %>
+  <div data-module="table-filter">
+    <div class="js-gem-c-table__filter govuk-!-display-none" aria-live="assertive">
```
andysellick, line 55: I think I'm getting closer to the problem with the screenreader announcement - this `aria-live` attribute needs to be on the`<p id="filterCount">`.

andysellick: Doing so seems to improve screen reader announcement, but not to 100%. I think part of the problem is that most screenreaders will only announce the hidden text if it changes, which means you could continue to type something that matches nothing and only receive one announcement. But that might be okay.

config/locales/en.yml
```diff
@@ -330,5 +330,9 @@ en:
     summary_list:
       delete: Delete
       edit: Change
+    table:
+      filter_count: results for the term entered
```
andysellick, line 334: It's possible to find only one result, so the announced text is `1 results for the term entered`, which isn't perfect. I don't think this should be a blocking issue though, I'm happy for it to be addressed in a follow up PR.

app/views/govuk_publishing_components/components/_table.html.erb
```diff
@@ -2,46 +2,71 @@
...
         <% end %>
       <% end %>
     <% end %>
   <% end %>
+<% end %>
+
+<% if filterable %>
+  <div data-module="table-filter">
+    <div class="js-gem-c-table__filter govuk-!-display-none">
+      <%= render "govuk_publishing_components/components/input", {
```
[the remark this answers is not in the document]
andysellick: Great. Can this be documented in the YML file?

andysellick (commented):
Before you merge can some of the commits be squashed together?

app/views/govuk_publishing_components/components/_table.html.erb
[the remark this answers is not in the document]
davidtrussler: Added documentation on this change.

config/locales/en.yml
davidtrussler: Yes that's true. Annoying but if we can return to it I guess that's the way to go.

andysellick (approved):
Looks good. Well done. Maybe squash that last commit before merging 👍

CHANGELOG.md
```diff
@@ -29,6 +29,7 @@
 * Remove accessible format request pilot support ([PR #2826](https://github.com/alphagov/govuk_publishing_components/pull/2826))
 * GTM analytics add page views ([PR #2814](https://github.com/alphagov/govuk_publishing_components/pull/2814))
 * Add id attribute to error alert component ([PR #2825](https://github.com/alphagov/govuk_publishing_components/pull/2825))
+* Add js search to table component ([PR #2803](https://github.com/alphagov/govuk_publishing_components/pull/2803))
```
DilwoarH, line 32: You probably want to move this to under the `Unreleased` header

DilwoarH (approved):
Small thing i noticed

CHANGELOG.md
davidtrussler: OK, have moved this.

PR #2811, opened 2022-06-13
Add details click tracking

AshGDS:
Hi @andysellick 

Would you be able to look at this PR? Thanks :+1: (I'm guessing i'll need to update the changelog before merge too)

## What

This adds click tracking to the GOV.UK details component. The code:

- Refactors the `aria-expanded` getter function into a generic attribute getter so it can be used for `aria-expanded` and `open`
- Checks if an element can have an 'open' attribute before trying to grab that attribute. This is needed as the closed state for a details attribute is when the `open` attribute doesn't exist. Therefore, you need to check wether the element is allowed to have an `open` attribute in the first place, otherwise every single non-details element would be reported back as `closed` since they are all missing the `open` attribute. This is currently just checking for `govuk-details` in the `data-module` attribute - there might be a safer way to check for this. But this should allow us to extend it to include any component that uses `open` in the future.
- I used slighlty modified tests from the `aria-expanded` tests to make the tests with `open` . Since this code is quite similar to the `aria-expanded` code, do you think we could reduce code duplication between the tests somehow?

## Why

The `open` attribute is similar to aria-expanded but has a slightly different 'closed' state, so this needs to be accounted for with extra code.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
@@ -25,11 +25,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           link_url: window.location.href.substring(window.location.origin.length),
           ui: JSON.parse(target.getAttribute('data-gtm-attributes')) || {}
         }
-        var ariaExpanded = this.checkExpandedState(target)
-        if (ariaExpanded) {
-          data.ui.state = ariaExpanded === 'false' ? 'opened' : 'closed'
+        var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+        /*
+          the details component uses an 'open' attribute instead of aria-expanded, so we need to check if we're on a details component.
+          since details deletes the 'open' attribute when closed, we need this boolean, otherwise every element which
+          doesn't contain an 'open' attr would be pushed to gtm as a closed element.
+        */
+        var canHaveOpenAttribute = target.getAttribute('data-module') ? target.getAttribute('data-module').includes('govuk-details') : false
+
+        if (ariaExpanded || canHaveOpenAttribute) {
           data.ui.text = data.ui.text || target.innerText
+
+          if (ariaExpanded) {
+            data.ui.state = (ariaExpanded === 'false') ? 'opened' : 'closed'
+          } else if (canHaveOpenAttribute) {
+            var openAttribute = this.getClosestAttribute(target, 'open')
+            data.ui.state = (openAttribute == null) ? 'opened' : 'closed'
+          }
         }
+
         window.dataLayer.push(data)
       }
     }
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
```diff
@@ -25,11 +25,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-        var ariaExpanded = this.checkExpandedState(target)
-        if (ariaExpanded) {
-          data.ui.state = ariaExpanded === 'false' ? 'opened' : 'closed'
+        var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+        /*
+          the details component uses an 'open' attribute instead of aria-expanded, so we need to check if we're on a details component.
+          since details deletes the 'open' attribute when closed, we need this boolean, otherwise every element which
+          doesn't contain an 'open' attr would be pushed to gtm as a closed element.
+        */
+        var canHaveOpenAttribute = target.getAttribute('data-module') ? target.getAttribute('data-module').includes('govuk-details') : false
```
andysellick, line 34: I'm interested by your approach - why did you choose to look for the specific `data-module` rather than the `details` element itself?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
```diff
@@ -25,11 +25,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+        if (ariaExpanded || canHaveOpenAttribute) {
           data.ui.text = data.ui.text || target.innerText
+
+          if (ariaExpanded) {
+            data.ui.state = (ariaExpanded === 'false') ? 'opened' : 'closed'
+          } else if (canHaveOpenAttribute) {
+            var openAttribute = this.getClosestAttribute(target, 'open')
```
andysellick, line 42: I see why, but this code now calls `getClosestAttribute` twice, which seems inefficient. It might be possible as part of altering the details component detection on line 34 to keep a hold of the `details` element itself as a variable, which could be referenced here to find the `open` attribute, instead of calling this function a second time?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
AshGDS: This was me being overly cautious - I was not sure if that would detect `<details>` elements that aren't part of the GOV.UK components. But thinking about it now, we're already doing a blind check for just `aria-expanded` anyway so that isn't an issue. This is just me forgetting how the GOV.UK ecosystem works! I'll change it to check the element itself. Thanks @andysellick :+1:

AshGDS:
Thanks @andysellick - I've made these changes, but might need some help with implementing `.closest` - I could only get `.querySelector` to work instead which wouldn't grab the parent, if that's important to us.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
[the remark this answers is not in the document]
andysellick: I think it should work if you do the following (tested and it seems to work)

```
var detailsElement = target.closest('details')
```

It's a little inefficient, because `closest` will traverse the whole way up the DOM until it finds what its looking for or runs out of elements. But I think if we tried to be more specific (like assuming the details element is one level up) we'd make the code too brittle, so I think this works.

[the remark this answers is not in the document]
AshGDS: Thanks @andysellick - i've updated this, and I think I've rewritten the test so it makes sense.

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
```diff
@@ -158,6 +158,57 @@ describe('Google Tag Manager click tracking', function () {
     })
   })
 
+  describe('doing tracking on an details element', function () {
```
andysellick, line 161: Nitpick: should be `doing tracking on a details element`

andysellick (approved):
Looks good. Can you add an entry to the CHANGELOG before merging please?

PR #2813, opened 2022-06-14
Add Options to Accordion Section Tracking

patrickpatrickpatrick:
## What

The function that fires an analytics event when an accordion is opened or closed can now be passed optional tracking options. This means custom dimensions can now be specified when a user clicks on an accordion section.

## Why

Similar to how link clicks can also append custom dimensions to the analytics event, a request was made that accordion open and close should also include the position (index) of the accordion on the page when clicked. This change means this dimension and any others can be easily added to the accordion open/close event.

[Relevant Trello Card](https://trello.com/c/1Y6WwRZh/1075-add-tracking-for-position-of-accordion-section-on-new-browse-level-2-curated-topics)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -144,6 +144,18 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     var action = expanded ? 'accordionOpened' : 'accordionClosed'
     var options = { transport: 'beacon', label: label }
 
+    // optional parameters are added to the parent
+    // heading not the button that is clicked
+    var extraOptions = section.parentElement && section.parentElement.getAttribute('data-track-options')
+
+    // this uses the same logic as track-click.js handleClick
+    // means we can add a custom dimensions on click
+    // (such as the index of the accordion on the page)
+    if (extraOptions) {
+      extraOptions = JSON.parse(extraOptions)
+      for (var k in extraOptions) options[k] = extraOptions[k]
+    }
+
     if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -144,6 +144,18 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
     var action = expanded ? 'accordionOpened' : 'accordionClosed'
     var options = { transport: 'beacon', label: label }
 
+    // optional parameters are added to the parent
+    // heading not the button that is clicked
+    var extraOptions = section.parentElement && section.parentElement.getAttribute('data-track-options')
```
hannalaakso, line 149: Just for the reference of anyone else reviewing this and wondering why we're creating an object for just one attribute: I think we have a convention to set page scoped dimensions (eg. category) as their own data attributes on the element and then create an options object for any event scoped ones. We usually do this for links I think and this is a heading but it makes sense to follow the same convention here.

hannalaakso (commented):
I've tested this together with https://github.com/alphagov/collections/pull/2825 and I can see the dimension 114 with the correct position being fired off on click from the accordion 👍  Although this doesn't happen on the Detailed guidance accordion, not sure if it should? Example on this page: `/browse/business/setting-up`

patrickpatrickpatrick:
> I've tested this together with [alphagov/collections#2825](https://github.com/alphagov/collections/pull/2825) and I can see the dimension 114 with the correct position being fired off on click from the accordion 👍 Although this doesn't happen on the Detailed guidance accordion, not sure if it should? Example on this page: `/browse/business/setting-up`

ah yeah, I've added that now to the other PR

hannalaakso (commented):
Tested together with https://github.com/alphagov/collections/pull/2825 and the dimension is firing off correctly on both types of accordions now 👍

PR #2814, opened 2022-06-15
[DO NOT MERGE] GTM analytics add page views

andysellick:
## What
Adds page views for Google Tag Manager code.

Will need manual testing to check the status code part of the event. It's not possible in JS to get a normal page's status code, so we rely on a variable that is set by the error page (in static) to determine the page status code.

## Why
Page views are a fundamental part of measuring user analytics.

## Visual Changes
None.

Trello card: https://trello.com/c/mizG5QKy/4-add-tracking-page-views


app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -0,0 +1,46 @@
...
+      if (window.dataLayer) {
+        var data = {
+          event: 'config_ready',
+          page: {
+            location: this.getLocation(),
+            referrer: this.getReferrer(),
+            title: this.getTitle(),
+            status_code: this.getStatusCode()
+          }
+        }
```
gclssvglx, line 41: Sorry to keep banging the drum for a generic GA4 data structure, I feel it would be best to create such a beast though - that way it's consistent throughout - less cognitive load on the devs and it'll really help the data teams with the YoY/UA mapping issue too.

Currently, we have this...
```
gtm_attributes: {
  event: ‘analytics’,
  event_name: 
  link_url:
  ui: {
    type: TYPE OF ELEMENT OR COMPONENT,
    text: NAME OR TITLE OF THE SPECIFIC ELEMENT,
    index: INDEX OF THE SPECIFIC ELEMENT, e.g. 1 (of 4) (optional),
    'index-total': TOTAL NUMBER OF ELEMENTS IN THE PARENT, e.g. 4 (optional),
    section: PART OF THE PAGE INTERACTED WITH (optional),
    state: STATE OF THE ELEMENT WHEN INTERACTED e.g. ‘opened’ (optional),
  }
}
```

If we change this structure slightly (and update the existing code of course) then map in these attributes  something like this...

```
data: {
  event: ‘config_ready’,
  event_name:  this.getTitle(),
  url: this.getLocation(),
  referrer: this.getReferrer(),
  status: this.getStatusCode(),
  state: {
    type: 'pageload',
    text: 'n/a',
    index: 'n/a',
    'index-total': 'n/a',
    section: 'n/a',
  }
}
```
So, we it doesn't matter what we call it, but we keep a consistent structure inside it and `n/a` anything that is not relevant for that specific interaction type.

app/assets/javascripts/govuk_publishing_components/analytics-ga4.js
```diff
@@ -1 +1,4 @@
+//= require ./analytics-ga4/gtm-page-views
 //= require ./analytics-ga4/gtm-click-tracking
+
+window.GOVUK.Gtm.sendPageView() // this will need integrating with cookie consent before production
```
gclssvglx, line 4: Nice!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -13,6 +13,13 @@
...
             title: this.getTitle(),
             status_code: this.getStatusCode()
+          },
+          publishing: {
+            document_type: this.getMetaContent('format'),
+            publishing_app: this.getMetaContent('publishing-application'),
+            rendering_app: this.getMetaContent('rendering-application'),
+            schema_name: this.getMetaContent('schema-name'),
+            content_id: this.getMetaContent('content-id')
           }
```
gclssvglx, line 40: Just curious, do we need this data in the GA4 event or is this mimicking what UA currently does? I can see a case for `content-id` and maybe `format`, but the rest feel unnecessary unless they are actively being use for filtering in reports etc.  If we go with the generic structure this feels like it belongs in `state`.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -0,0 +1,96 @@
...
+    getMetaContent: function (name) {
+      var tag = document.querySelector('meta[name="govuk:' + name + '"]')
+      if (tag) {
+        return tag.getAttribute('content').toLowerCase()
+      } else {
+        return 'n/a'
+      }
+    },
+
+    getLanguage: function () {
```
AshGDS, line 77: Is there a reason this uses `title` instead of the `<html>` tag which would have `<html lang="en">` ? 

(My guess is, if something isn't in English then the `<title>` tag has a `lang` attribute, and if it doesn't have that attribute we assume it's English?)

Some pages don't have a lang tag on their `title ` but do on `html` - for example https://gov.uk or https://www.gov.uk/coronavirus

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -20,6 +20,14 @@
...
             content_id: this.getMetaContent('content-id')
+          },
+          taxonomy: {
+            browse_topic: this.getMetaContent('section'),
+            main_topic: this.getMetaContent('taxon-slug'),
+            main_topic_id: this.getMetaContent('taxon-id'),
+            level_1: this.getMetaContent('themes'),
+            topics: this.getMetaContent('taxon-slugs'),
+            topics_ids: this.getMetaContent('taxon-ids')
           }
```
gclssvglx, line 40: Wow more data - I'm curious to know how this stuff is being used in GA.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -28,6 +28,15 @@
...
+          },
+          content: {
+            language: this.getLanguage(),
+            history: this.getHistory(),
+            withdrawn: this.getWithDrawn(),
+            government: this.getMetaContent('government'),
+            political_status: this.getMetaContent('political-status'),
+            org_ids: this.getMetaContent('analytics:organisations'),
+            world_location: this.getMetaContent('analytics:world-locations')
           }
```
gclssvglx, line 40: Quite an intense amount of data - assuming it's all used I can see why the data teams are keen to get this in - it seems it pretty much covers everything!

gclssvglx (commented):
This is looking fantastic @andysellick - I have one bigish suggestion around data structure. 

I love that you've managed to get hold of all that data - amazing job. But, I am curious to know if it's all actually used/needed though - this seems like a good opportunity to skinny down some of the data we capture if we can.

AshGDS (commented):
Hi @andysellick 

Looks good to me code wise - I like how the tests are written too (helps me to learn good TDD!)

I've added one question.

Thanks :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: That's a really good point, thanks for picking that up. The example page I was building this on had it on the `title` tag and it hadn't occurred to me that it would be somewhere else. I'll check a few more examples and rework this.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: Yes, there's a lot - and I haven't finished implementing it yet either - see https://docs.google.com/document/d/1LGLRb6zMvJBjzMYPgLAcBWbS-GsKGUzLcumCsZgiUhM/edit#

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: I've updated the code to use the lang attribute on the `html` element instead - fairly sure from looking through our code that we use that element more consistently for that purpose.

AshGDS: Thanks @andysellick - looks good! I'm guessing the `if (html)` condition doesn't need an `else` condition since the likelihood of `html` not existing is near impossible?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: Having a consistent data structure is good, but apparently the analysts are happy with the structure as it is because a page view is a distinct thing from other kinds of events (even though they're all events).

And yes, there could be empty parts to the data, but they should always have `n/a` values.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: Apparently there's a whole load of data we don't currently capture that is desirable, which is reflected in this new structure. Yeah, I agree it's a lot.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
andysellick: That's a good point! Thanks, have amended.

andysellick:
Thanks for the reviews both, however following progress in understanding how metatags work I'm expanding this PR to include more of the schema data that I previously thought wasn't easily available. Will shout again when it's ready for re-review.

andysellick:
I _think_ this is ready for review now. Changes implemented are the new consistent names for the data items, plus some additional items that I hadn't included before but are now displayed as meta tags that are read by the JavaScript.

There are a few outstanding issues regarding naming conventions but they relate to existing naming conventions so I think rather than complicate this PR further I'll address them in a separate one. Also I haven't implemented banners or PII yet, but again I'll do that in another PR.

gclssvglx (approved):
Great work @andysellick 👍

AshGDS (approved):
Looks good @andysellick 👍

PR #2817, opened 2022-06-16
Change menubar P elements to use H elements

matthillco:
## What

Improve the accessibility of the main menu by changing `<p>` elements to `<hX>` elements. [Trello](https://trello.com/c/W317oTLq/1067-fix-menubar-items-that-use-p-tags-in-titles-and-not-heading-elements)

## Why

Paragraph elements aren't announced by screen-readers in the same way as heading elements. This can hinder screen-reader users from properly navigating the page. This small fix helps those users by triggering a heading announcement.

## Visual Changes

Not applicable

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add gem-track-click to specific sections of layout_footer ([PR #2800](https://github.com/alphagov/govuk_publishing_components/pull/2800))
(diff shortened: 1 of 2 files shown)


injms (approved):
Super work - thanks for ploughing through the caching problems to get this done 👍 

It's worth noting that there's no visual difference in the pull request description so people know that's what to expect when reviewing - I put "None" under the "Visual changes" heading when this is the case.

PR #2822, opened 2022-06-20
Fix issue with detecting aria-expanded

AshGDS:
Hi @andysellick 

Would you be able to approve this PR if all is OK?

Thanks :+1:

## What

Adds an if statement to check wether `aria-expanded` is inside an accordion component

## Why

Without this, imagine you had the following HTML:

```HTML
    <div data-gtm-event-name="event3-name" data-gtm-attributes='${JSON.stringify(attributes)}' class="clickme">
        <div class='content'>
            abc
        </div>
        <div class='accordion'>
            <button aria-expanded="false">Show</button>
        </div>
    </div>
```

Clicking on the parent div would trigger the accordion click tracking to run, since it has a nested `aria-expanded` element inside it. We only want that tracking to run when the accordion is clicked.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
@@ -27,7 +27,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
 
         // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
-        if (target.closest('[data-module="govuk-accordion gem-accordion"]') || target.closest('[data-module="gem-accordion govuk-accordion"]')) {
+        if (target.closest('.gem-c-accordion')) {
           var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
         }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
@@ -25,7 +25,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           link_url: window.location.href.substring(window.location.origin.length),
           ui: JSON.parse(target.getAttribute('data-gtm-attributes')) || {}
         }
-        var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+
+        // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
+        if (target.closest('[data-module="govuk-accordion gem-accordion"]') || target.closest('[data-module="gem-accordion govuk-accordion"]')) {
+          var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+        }
+
         /*
           the details component uses an 'open' attribute instead of aria-expanded, so we need to check if we're on a details component.
           since details deletes the 'open' attribute when closed, we need this boolean, otherwise every element which
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
@@ -119,6 +119,7 @@ describe('Google Tag Manager click tracking', function () {
         text: 'some text'
       }
       element.setAttribute('data-gtm-event-name', 'event3-name')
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
```diff
@@ -25,7 +25,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
           link_url: window.location.href.substring(window.location.origin.length),
           ui: JSON.parse(target.getAttribute('data-gtm-attributes')) || {}
         }
-        var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+
+        // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
+        if (target.closest('[data-module="govuk-accordion gem-accordion"]') || target.closest('[data-module="gem-accordion govuk-accordion"]')) {
```
andysellick, line 30: I haven't come across this before - presumably it was necessary to include both these conditions because the two items in `data-module` could occur in either order? Would a simpler alternative be to look for just one of them in the returned value?

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
```diff
@@ -152,6 +153,7 @@ describe('Google Tag Manager click tracking', function () {
           text: 'some text'
         }
       }
+      element.setAttribute('data-module', 'gem-accordion govuk-accordion')
```
andysellick, line 156: Is it necessary to set this again? It's already done in the `beforeEach` function.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
AshGDS: @andysellick I tested this on the console on the component guide Accordion, and it looks like querySelectors require an exact match, so just doing `data-module="govuk-accoridon"` doesn't work as you need the full `data-module` value. So I added a check for both orders of the `data-module`. Is this OK?

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
AshGDS: This was to test both orders of `data-module` - it looks very similar to the beforeEach! Is this OK?

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
```diff
@@ -254,6 +257,7 @@ describe('Google Tag Manager click tracking', function () {
           text: 'Hide'
         }
       }
+      clickOn.setAttribute('data-module', 'gem-accordion govuk-accordion')
```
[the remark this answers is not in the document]
AshGDS: I've answered in the review comment above

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.js
andysellick: That's fine, although it seems over-complicated - could you search for `target.closest('.gem-c-accordion')` instead?

AshGDS: Thanks @andysellick I'm happy to do that - though I was wondering what your thoughts are on coupling the analytics up to classes? I'm guessing that class is very unlikely to change?

andysellick: I think it's very unlikely to change, yes. But you raise a good point - in the event that the classname (or even the data-module attribute) were to change, how would we know? I suppose the analytics would stop working for this expanding element on the accordion - hopefully that would be ultimately picked up by some of the testing that @gclssvglx is looking into?

AshGDS: Thanks @andysellick - yep it should be picked up his testing, I can change it to use the class instead :+1:

AshGDS: @andysellick I've updated this now :+1:

andysellick (approved):
Looks good. Can you add a CHANGELOG entry?

Issue #2819, opened 2022-06-16
Layout super navigation header initialises gem-track-click inefficiently


andysellick:
The layout header uses the `gem-track-click` module to track links in the dropdowns. The way this is being done is to apply the `data-module="gem-track-click"` [attribute to each individual link](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb#L103), which is at least a few dozen links.

![Screenshot 2022-06-16 at 17 08 11](https://user-images.githubusercontent.com/861310/174116245-74e8e4e7-9d50-41de-8ad5-501c7d9bf42b.png)

This is working entirely correctly however `gem-track-click` doesn't need to be activated like this. In the [documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics/track-click.md) it shows that the module can be applied to a parent element and tracking attributes added to children (see 'Specific tracking can also be applied to elements within a container'), in order that the module is only instantiated once.

I'm concerned that the way it's being used at the moment in the header means that unnecessary load is being added to browsers to create each instance of the module for every link - it's probably using more memory than is necessary.

(Interestingly it appears to have been configured correctly in the footer, with the `data-module` applied only to the parent)

patrickpatrickpatrick:
I recently refactored the footer gem-track click, can pick this up too to make them in line w/ each other

andysellick:
When you take a look, I'd be interested to know what impact the change has on performance. I was thinking of starting by putting a console log into the `gem-track-click` init code (to count how many times it gets called before and after the change) but also looking at memory usage in the browser.

PR #2823, opened 2022-06-20
Remove gem-track-click for individual links, use in header instead

patrickpatrickpatrick:
## What

Remove `data-module="gem-track-click"` from each of the links in the footer, add `data-module="gem-track-click"` to the parent header component instead. 

## Why

Click tracking can be initialised on a parent component instead of each individual component. The previous implementation meant that for each clickable link in the header the module for tracking links was being initialised unnecessarily. Using this alternative approach has benefits such as reducing the number of event listeners by 31 and decreasing memory usage. 

Screenshot 2022-06-20 at 11 27 52

Fixes https://github.com/alphagov/govuk_publishing_components/issues/2819


andysellick (approved):
Looks good to me. Just wondering if it should have a changelog entry... probably not? Think you're fine.

If you add fixes https://github.com/alphagov/govuk_publishing_components/issues/2819 into the description, it'll automatically close that issue when this PR is merged.

PR #2826, opened 2022-06-20
Remove code for accessible format request pilot.

KludgeKML:
## What
Support for the Request Accessible Format form is being removed.

## Why
All participating departments have pulled out of the pilot, so removing code makes sense while we rethink. It may be reinstated from the git history, or if the pilot does not recur, we keep the code tidy by removing it.

## Visual Changes
No visual changes, but With Accessible Format Request form option has been removed from items rendered for visual regression check.

https://trello.com/c/iZEzpeFy/1340-remove-dfe-from-the-accessible-format-pilot

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove accessible format request pilot support ([PR #2826](https://github.com/alphagov/govuk_publishing_components/pull/2826))
 * Add gem-track-click to specific sections of layout_footer ([PR #2800](https://github.com/alphagov/govuk_publishing_components/pull/2800))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove accessible format request pilot support ([PR](https://github.com/alphagov/govuk_publishing_components/pull/2826))
```
maxgds, line 12: Could you add the PR number to the linking text.

maxgds (approved):
One tiny tweak requested, otherwise good - doesn't need a re-review.

maxgds:
> @maxgds could you approve the visual change?

Done

PR #2829, opened 2022-06-22
Check if primary_publisher nil OR empty in meta_tags

patrickpatrickpatrick:
## What

Add check to meta_tags to make sure primary_publisher is not an empty array before attempting to access a hash in the array.

## Why

Tests were intermittently failing on collections and after some digging, it was discovered that this was because that the random schema generated for tests can occasionally result in `primary_publishing_organisation` being set to an empty array. This would cause accessing the key of the first hash of the array in `meta_tags` to fail which would mean tests would fail. This commit adds `.blank?` to the primary_publisher check which means that if the array is empty then an error will not be caused by this line. This fixes the collections tests but in theory should also fix other intermittently failing tests in other applications.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Nice. Might be worth getting a BE dev to double check this.

chao-xian (approved):
👍

PR #2830, opened 2022-06-22
Add new fields for spam detection

AshGDS:
Hi @andysellick 

Would you be able to review this PR? 

Not sure if it needs new tests, and if so what the tests would be.

Thanks :+1:

## What

Adds:
- A timer hidden field to see how quickly the form is submitted. Uses `setInterval` to update the field.
- A paste count hidden field and listener which detects how many pastes are inserted into the form `textarea`.
- A keypress count hidden field and listener which detects how many keypresses are inserted into the form `textarea`.

It then works in combination with the backend to filter out the spam: https://github.com/alphagov/feedback/pull/1452

## Why

We are experimenting using JS to provide metadata which might help reduce spam on the feedback component.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -142,6 +142,19 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       '<h2>Sorry, we’re unable to send your message as you haven’t given us any information.</h2>',
       ' <p>Please tell us what you were doing or what went wrong</p>'
     ].join('')
+
+    var honeyPot = document.createElement('input')
+    honeyPot.setAttribute('type', 'hidden')
+    honeyPot.setAttribute('name', 'token')
+    var timer = 0;
+    honeyPot.setAttribute('value', timer)
+
+    var timerUpdater = setInterval( function() {
+        timer = timer + 0.5;
+        honeyPot.setAttribute('value', timer)
+    }, 500);
+    this.somethingIsWrongForm.appendChild(honeyPot)
+
   }
 
   Feedback.prototype.setHiddenValuesNotUsefulForm = function (gaClientId) {
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -75,11 +75,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     }.bind(this))
 
     this.somethingIsWrongButton.addEventListener('click', function (e) {
-      this.timerFunction = function () {
+      this.timerInterval = setInterval(function () {
         this.timer = this.timer + 1
         this.timerHoneyPot.setAttribute('value', this.timer)
-      }
-      this.timerInterval = setInterval(this.timerFunction.bind(this), 1000)
+      }.bind(this), 1000)
     }.bind(this))
 
     this.somethingIsWrongForm.addEventListener('paste', function (e) {
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -83,21 +83,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       }.bind(this), 1000)
     }.bind(this))
 
-    this.pasteListener = function () {
-      this.pastes = this.pastes + 1
-      this.pastesHoneyPot.setAttribute('value', this.pastes)
-    }.bind(this)
-
-    this.keypressListener = function () {
-      this.keypresses = this.keypresses + 1
-      this.keypressHoneyPot.setAttribute('value', this.keypresses)
-    }.bind(this)
-
-    this.whatDoingInput.addEventListener('paste', this.pasteListener)
-    this.whatWrongInput.addEventListener('paste', this.pasteListener)
-    this.whatDoingInput.addEventListener('keypress', this.keypressListener)
-    this.whatWrongInput.addEventListener('keypress', this.keypressListener)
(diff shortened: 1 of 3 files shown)


andysellick (commented):
Looks good and seems to work as intended. Yes, definitely need some tests for this. It should be possible to recreate the behaviour we're intending to capture in a test. You should be able to use [jasmine clock](https://jasmine.github.io/api/2.6/Clock.html) for the timer part.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -74,6 +74,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.setHiddenValuesNotUsefulForm(gaClientId)
     }.bind(this))
 
+    this.somethingIsWrongButton.addEventListener('click', function (e) {
+      this.timerFunction = function () {
+        this.timer = this.timer + 1
+        this.timerHoneyPot.setAttribute('value', this.timer)
+      }
+      this.timerInterval = setInterval(this.timerFunction.bind(this), 1000)
```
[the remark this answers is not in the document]
AshGDS: Thanks @andysellick - I was anticipating having to reinstate the `setInterval` somewhere else, but that never happened so I have changed it to what you've written 👍

AshGDS:
Thanks @andysellick - I have added tests now. I had to move the `paste` and `keypress` listener directly onto the `textarea` elements  (It was on the whole `form` originally.) It seems like in the tests, when I would dispatch an event programatically on the `textarea`, the event would not bubble upwards to the listener on the `form`. Therefore I've moved the listener onto both of the `textarea` inputs so that the tests work.

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -142,6 +167,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    this.pastesHoneyPot.setAttribute('type', 'hidden')
+    this.pastesHoneyPot.setAttribute('name', 'pastes')
+    this.pastesHoneyPot.setAttribute('value', this.pastes)
+    this.somethingIsWrongForm.appendChild(this.pastesHoneyPot)
+
+    this.keypressHoneyPot = document.createElement('input')
+    this.keypressHoneyPot.setAttribute('type', 'hidden')
+    this.keypressHoneyPot.setAttribute('name', 'keypresses')
+    this.keypressHoneyPot.setAttribute('value', this.keypresses)
+    this.somethingIsWrongForm.appendChild(this.keypressHoneyPot)
```
andysellick, line 191: Lines 175 to 191 seem to be doing the same thing three times with slight variations. Could you also do something like

```
function createHoneyPotElement(reference, name, value) {
  reference = document.createElement('input')
  reference.setAttribute('type', 'hidden')
  reference.setAttribute('name', name)
  reference.setAttribute('value', value)
  this.somethingIsWrongForm.appendChild(reference)
}

createHoneyPotElement(this.timerHoneyPot, 'timer', this.timer)
createHoneyPotElement(this.timerHoneyPot, 'pastes', this.pastes)
createHoneyPotElement(this.timerHoneyPot, 'keypresses', this.keypresses)
```

(might need `this.timer` (etc.) to be defined before calling, but might not)

This is a non-blocking issue though - the code works fine as it is.

spec/javascripts/components/feedback-spec.js
```diff
@@ -735,6 +738,56 @@ describe('Feedback component', function () {
...
+    it('has a paste detection field', function () {
+      var $form = document.querySelector('.gem-c-feedback #something-is-wrong')
+      var $pastes = $form.querySelector('input[name=pastes]')
+      var $whatDoing = $form.querySelector('[name=what_doing]')
+      var $whatWrong = $form.querySelector('[name=what_wrong]')
+
+      var pasteEvent = new window.ClipboardEvent('paste')
+
+      $whatDoing.dispatchEvent(pasteEvent)
+      $whatWrong.dispatchEvent(pasteEvent)
```
andysellick, line 771: Another non-blocking comment. We have a [script for creating events](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/trigger-event.js) that you might find useful here, might be able to write as:

```
window.GOVUK.triggerEvent($whatDoing, 'paste')
window.GOVUK.triggerEvent($whatWrong, 'paste')
```

app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -74,6 +76,28 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.setHiddenValuesNotUsefulForm(gaClientId)
     }.bind(this))
 
+    this.somethingIsWrongButton.addEventListener('click', function (e) {
```
andysellick, line 79: I like that the timer isn't initialised until the user clicks the button to expand the form, it fits with only starting once the user is actually using the form but also means that there's no interval happening on every page on GOV.UK unless it's actually needed. 

Would it be possible to similarly limit the other listeners (paste and keypress), perhaps by initialising them inside this click listener as well?

AshGDS:
Thanks @andysellick - good point, I'll make those changes. I'll also need to check that accessibility tools like Dragon voice to text are still able to submit text into the textareas. If inserting text with these types of software are detected as 0 keypresses then the paste/keypress method may need to be removed.

AshGDS:
@andysellick As I've removed the paste/keypress listener, your previous feedback may not be relevant anymore. If everything is OK with the timer code, would you be able to approve the PR, and then I'll add it to the CHANGELOG? Thanks!

andysellick (approved):
Looks good! The visual diff error should resolve if you rebase from `main`.

PR #2831, opened 2022-06-23
Standardise some analytics attributes

andysellick:
## What
Rename some of the attributes used by GA4 page views.

## Why
Trying to clean up some of the code and standardise the names. Ideally the name of something in the content item would persist through to the metadata, the JS and on into GA4, rather than being slightly different at various stages, which could cause confusion in future.

## Visual Changes
None.

Trello cards: 

- https://trello.com/c/mizG5QKy/4-add-tracking-page-views
- https://trello.com/c/GDLkrUws/335-page-views-data-naming-inconsistencies

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
@@ -11,6 +11,7 @@ describe('Google Tag Manager click tracking', function () {
 
   afterEach(function () {
     document.body.removeChild(element)
+    window.location.hash = ''
   })
 
   describe('configuring tracking incompletely', function () {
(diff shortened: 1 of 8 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-click-tracking.spec.js
```diff
@@ -11,6 +11,7 @@ describe('Google Tag Manager click tracking', function () {
 
   afterEach(function () {
     document.body.removeChild(element)
+    window.location.hash = ''
```
GDSNewt, line 14: I can't envision a scenario in testing where we would want the hash to persist across tests so I think this is fine.

GDSNewt (approved):
LGTM :shipit:

PR #2840, opened 2022-06-29
Add extra options for show/hide all sections accordion click

patrickpatrickpatrick:
## What

Adds ability to add extra options to the click anayltics event for showing/hiding all the sections of an accordion.

## Why

In a [previous PR ](https://github.com/alphagov/govuk_publishing_components/pull/2813)I added the ability to add extra options to click events for opening sections in an accordion but I did not include the Show/Hide All Sections button. For tracking purposes it means that if an attribute is not passed in an event, then an event cannot be filtered on that attribute. Therefore adding the ability to pass attributes to the Show/Hide All Sections button means it can be more easily inlcuded in searches/filters on click events.

[Relevant Trello Card](https://trello.com/c/hDJRk2hS/1096-add-tracking-for-position-of-show-hide-all-sections-in-accordion-on-new-browse-level-2-curated-topics)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -124,6 +124,16 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
       var label = expanded ? 'Show all sections' : 'Hide all sections'
       var action = expanded ? 'accordionOpened' : 'accordionClosed'
       var options = { transport: 'beacon', label: label }
+
+      var extraOptions = event.target && event.target.getAttribute('data-track-options')
+
+      // this uses the same logic as track-click.js handleClick
+      // means we can add a custom dimensions on click
+      if (extraOptions) {
+        extraOptions = JSON.parse(extraOptions)
+        for (var k in extraOptions) options[k] = extraOptions[k]
+      }
+
       if (window.GOVUK.analytics && window.GOVUK.analytics.trackEvent) {
         window.GOVUK.analytics.trackEvent('pageElementInteraction', action, options)
       }
--- a/app/views/govuk_publishing_components/components/docs/accordion.yml
+++ b/app/views/govuk_publishing_components/components/docs/accordion.yml
@@ -189,7 +189,7 @@ examples:
 
       Each item can also have a `data_attributes` hash. These `data_attributes` are placed on the `button` that triggers the opening and closing - useful for differentiating between each section of the accordion.
 
-      Data attributes can also be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created).
+      Data attributes can also be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created). If `track_options` within `data_attributes_show_all` is set, then it is possible to pass a custom dimension when 'Show/Hide all' is clicked.
     data:
       data_attributes:
           gtm: gtm-accordion
@@ -197,6 +197,7 @@ examples:
       data_attributes_show_all:
         gtm-event-name: example
         gtm-attributes: "{ 'ui': { 'type': 'type value', 'section': 'section value' } }"
+        tracking-options: "{ 'dimension114': 1 }"
       items:
         - heading:
             text: Writing well for the web
--- a/spec/javascripts/components/accordion-spec.js
+++ b/spec/javascripts/components/accordion-spec.js
@@ -16,6 +16,16 @@ describe('Accordion component', function () {
       '</div>' +
       '<div class="govuk-accordion__section">' +
         '<div class="govuk-accordion__section-header">' +
+          '<h2 class="govuk-accordion__section-heading" data-track-count="accordionSection" data-track-options="{&quot;dimension114&quot;:1}">' +
+            '<button type="button" class="govuk-accordion__section-button" aria-expanded="false">' +
+              '<span class="govuk-accordion__section-heading-text">' +
+                'Accordion Section Heading Text' +
+              '</span>' +
+            '</button>' +
+          '</h2>' +
+        '</div>' +
+      '<div class="govuk-accordion__section">' +
+        '<div class="govuk-accordion__section-header">' +
           '<h2 class="govuk-accordion__section-heading">' +
             '<span class="govuk-accordion__section-button" id="default-id-078ef791-heading-1">Writing well for the web</span>' +
           '</h2>' +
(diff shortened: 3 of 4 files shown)


app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -189,7 +189,7 @@ examples:
 
       Each item can also have a `data_attributes` hash. These `data_attributes` are placed on the `button` that triggers the opening and closing - useful for differentiating between each section of the accordion.
 
-      Data attributes can also be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created).
+      Data attributes can also be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created). If `track_options` within `data_attributes_show_all` is set, then it is possible to pass a custom dimension when 'Show/Hide all' is clicked.
```
andysellick, line 192: Can you add an example of this to the documentation? (you might need to create a separate entry)

spec/javascripts/components/accordion-spec.js
```diff
@@ -49,22 +63,50 @@ describe('Accordion component', function () {
...
+  })
+
+  it('applies custom attributes to click event if specified in show/hide section', function () {
+    var object = {
+      'track-options': {
+        dimensionExample: 'Example value'
+      }
+    }
+
+    accordion.setAttribute('data-show-all-attributes', JSON.stringify(object))
```
andysellick, line 84: Could this technique (of setting the required attributes on the fixture in the test itself, rather than the markup at the top of the file) be used in the previous test as well? It makes the test more self-contained and readable.

spec/javascripts/components/accordion-spec.js
```diff
@@ -49,22 +63,50 @@ describe('Accordion component', function () {
...
+      },
+      'track-options': {
+        dimensionExample: 'Example value'
       }
     }
+
     var wrappingObject = {
       'gtm-event-name': 'example',
-      'show-all-attributes': JSON.stringify(object)
+      ...object
```
andysellick, line 104: Is there a particular reason for using spread syntax here, over the previous method? I ask because it's not supported in IE and it would be good to be able to run the tests in IE if we need to, and this would block that.

spec/javascripts/components/accordion-spec.js
```diff
@@ -49,22 +63,50 @@ describe('Accordion component', function () {
...
     }
     accordion.setAttribute('data-show-all-attributes', JSON.stringify(wrappingObject))
     startAccordion()
+    expect(document.querySelector('.govuk-accordion__show-all').getAttribute('data-track-options')).toEqual(JSON.stringify({ dimensionExample: 'Example value' }))
     expect(document.querySelector('.govuk-accordion__show-all').getAttribute('data-gtm-event-name')).toEqual('example')
-    expect(document.querySelector('.govuk-accordion__show-all').getAttribute('data-show-all-attributes')).toEqual(JSON.stringify(object))
+    expect(document.querySelector('.govuk-accordion__show-all').getAttribute('data-ui')).toEqual(JSON.stringify(object.ui))
```
andysellick, line 110: I think this is modifying the intent of the original test, which is to test that the show/hide all link has been correctly given a `data-show-all-attributes` attribute (needed for the GA4 work).

Suggestion - could this test be reverted and a new test added to check this new functionality? It feels related but distinct.

spec/javascripts/components/accordion-spec.js
```diff
@@ -7,40 +7,54 @@ describe('Accordion component', function () {
...
+    </div>
+  `
 
   function startAccordion () {
     new GOVUK.Modules.GemAccordion(accordion).init()
   }
 
   beforeEach(function () {
+    spyOn(GOVUK.analytics, 'trackEvent')
+    spyOn(GOVUK.Modules, 'GemTrackClick').and.callFake(function () { this.init = function () {} })
```
andysellick, line 56: I can't see where `GemTrackClick` is referenced - can you explain more about why this line is needed?

spec/javascripts/components/accordion-spec.js
```diff
@@ -7,40 +7,54 @@ describe('Accordion component', function () {
...
+        <div class="govuk-accordion__section-header">
+          <h2 class="govuk-accordion__section-heading" data-track-count="accordionSection" data-track-options="{&quot;dimension114&quot;:1}">
+            <button type="button" class="govuk-accordion__section-button" aria-expanded="false">
+              <span class="govuk-accordion__section-heading-text"><span class="govuk-accordion__section-heading-text-focus">Accordion Section Heading Text</span></span>
+              <span class="govuk-visually-hidden govuk-accordion__section-heading-divider">, </span>
+                <span class="govuk-accordion__section-toggle" data-nosnippet="">
+                  <span class="govuk-accordion__section-toggle-focus">
+                    <span class="govuk-accordion-nav__chevron govuk-accordion-nav__chevron--down"></span>
+                    <span class="govuk-accordion__section-toggle-text">Show<span class="govuk-visually-hidden"> this section</span>
+                  </span>
```
andysellick, line 27: It looks like these additional lines are to make the markup closer to the actual rendered content, but is there a particular reason for this? I'd generally recommend that fixture markup is as simple as possible - if these extra spans are needed for a test then fair enough, but I can't see them being referenced anywhere.

andysellick (commented):
Thanks for adding some tests, although I've raised a few questions. Also would it be possible to add some tests for https://github.com/alphagov/govuk_publishing_components/pull/2813? (either in this PR as a separate commit or in a separate PR, whichever makes sense).

patrickpatrickpatrick:
Moved all new tests to separate tests and deleted previous changes to existing ones. Reverted all ES6 changes and removed extraneous mark-up and statements.

> Thanks for adding some tests, although I've raised a few questions. Also would it be possible to add some tests for #2813? (either in this PR as a separate commit or in a separate PR, whichever makes sense).

`it('applies custom attributes to click event if specified in section')` should test the change I made in that PR, as the additional data attribute for the sections are added directly to the element and aren't appended using javascript like for the show all sections data attributes

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -48,13 +48,18 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
...
         var showAll = this.$module.querySelector('.' + this.showAllControls)
         var values = JSON.parse(showAllAttributes)
         var keys = Object.keys(values)
         for (var i = 0; i < keys.length; i++) {
-          showAll.setAttribute('data-' + keys[i], values[keys[i]])
+          if (typeof values[keys[i]] === 'object') {
+            showAll.setAttribute('data-' + keys[i], JSON.stringify(values[keys[i]]))
+          } else {
+            showAll.setAttribute('data-' + keys[i], values[keys[i]])
+          }
```
andysellick, line 62: As discussed, turns out this change isn't necessary - the existing code will already append any given attribute onto the show all element.

andysellick (commented):
Almost ready to approve! Just one thing, which we just discussed.

andysellick (approved):
🚀

PR #2842, opened 2022-06-30
Add GTM Personally Identifiable Information remover

AshGDS:
Hi @andysellick 

Would you be able to approve this if all is OK?

Thanks :+1:

## What

This takes the PII removal code from Universal Analytics and moves it to our gtm code. Some modifications to the PII code have been made, which are:
- It now prevents email addresses prepended with a `#` from having the `#` removed in redaction. This was needed as a valid test was failing due to the # unexpectedly being caught in the redaction.
- A function  - `stripPIIWithOverride()` - was added to manually override wether dates and postcodes are stripped. This was added because we are only applying the PII to certain values now. In Universal Analytics, PII was being applied to every value, unless it was wrapped in a `PIISafe` class. Enabling Dates and Postcodes on UA was also done in a way that applied it to every stripPII calls - you weren't easily able to have some values only strip emails, and then other values strip email, postcodes and dates.
- The `PIISafe` class now exists within the `PIIRemover` module - in UA it existed in `GOVUK.Analytics` which created an unnecessary dependancy. Theoretically the `PIIRemover` can be used outside of analytics code now.
- The `referrer` value in Pageview tracking has maximum redaction on it - we'll need to measure the impact of this (if it creates any false positives.)

## Why

We need to be able to remove PII from values pushed to GTM. We are removing all the UA code eventually so separate file were created.

## Testing

You could test it by running `static` with a local publishing components, and the GTM environment variables in static's `docker-compose.yml`. You can then view `gem_layout.html.erb` in your browser, and edit the page title in `_gem_base.html.erb` to contain an email address.  You would then see the redaction on the title working if you inspect `window.dataLayer`.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
@@ -5,7 +5,7 @@
   var GOVUK = global.GOVUK || {}
 
   GOVUK.Gtm = {
-    PiiRemover: new GOVUK.PiiRemover(),
+    PIIRemover: new GOVUK.PIIRemover(),
     sendPageView: function () {
       if (window.dataLayer) {
         var data = {
@@ -50,15 +50,15 @@
     },
 
     getLocation: function () {
-      return this.PiiRemover.stripPII(document.location.href)
+      return this.PIIRemover.stripPII(document.location.href)
     },
 
     getReferrer: function () {
-      return this.PiiRemover.stripPIIWithOverride(document.referrer, true, true)
+      return this.PIIRemover.stripPIIWithOverride(document.referrer, true, true)
     },
 
     getTitle: function () {
-      return this.PiiRemover.stripPII(document.title)
+      return this.PIIRemover.stripPII(document.title)
     },
 
     // window.httpStatusCode is set in the source of the error page in static
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
@@ -1,9 +1,11 @@
+//= require ./pii-remover
 ;(function (global) {
   'use strict'
 
   var GOVUK = global.GOVUK || {}
 
   GOVUK.Gtm = {
+    PiiRemover: new GOVUK.PiiRemover(),
     sendPageView: function () {
       if (window.dataLayer) {
         var data = {
@@ -48,15 +50,15 @@
     },
 
     getLocation: function () {
-      return document.location.href
+      return this.PiiRemover.stripPII(document.location.href)
     },
 
     getReferrer: function () {
-      return document.referrer
+      return this.PiiRemover.stripPII(document.referrer, true)
     },
 
     getTitle: function () {
-      return document.title
+      return this.PiiRemover.stripPII(document.title)
     },
 
     // window.httpStatusCode is set in the source of the error page in static
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
@@ -54,7 +54,7 @@
     },
 
     getReferrer: function () {
-      return this.PiiRemover.stripPII(document.referrer, true)
+      return this.PiiRemover.stripPIIWithOverride(document.referrer, true, true)
     },
 
     getTitle: function () {
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
@@ -1,0 +1,119 @@
+;(function (global) {
+  'use strict'
+
+  var GOVUK = global.GOVUK || {}
+  var EMAIL_PATTERN = /[^\s=/?&#]+(?:@|%40)[^\s=/?&]+/g
+  var POSTCODE_PATTERN = /\b[A-PR-UWYZ][A-HJ-Z]?[0-9][0-9A-HJKMNPR-Y]?(?:[\s+]|%20)*[0-9](?!refund)[ABD-HJLNPQ-Z]{2,3}\b/gi
+  var DATE_PATTERN = /\d{4}(-?)\d{2}(-?)\d{2}/g
+  var ESCAPE_REGEX_PATTERN = /[|\\{}()[\]^$+*?.]/g
+
+  // specific URL parameters to be redacted from accounts URLs
+  var RESET_PASSWORD_TOKEN_PATTERN = /reset_password_token=[a-zA-Z0-9-]+/g
+  var UNLOCK_TOKEN_PATTERN = /unlock_token=[a-zA-Z0-9-]+/g
+  var STATE_PATTERN = /state=.[^&]+/g
+
+  function shouldStripDates () {
+    var metas = document.querySelectorAll('meta[name="govuk:static-analytics:strip-dates"]')
+    return metas.length > 0
+  }
+
+  function shouldStripPostcodes () {
+    var metas = document.querySelectorAll('meta[name="govuk:static-analytics:strip-postcodes"]')
+    return metas.length > 0
+  }
+
+  function queryStringParametersToStrip () {
+    var meta = document.querySelector('meta[name="govuk:static-analytics:strip-query-string-parameters"]')
+    var value = false
+    if (meta) {
+      value = meta.getAttribute('content')
+    }
+    var parameters = []
+
+    if (value) {
+      var split = value.split(',')
+      for (var i = 0; i < split.length; i++) {
+        parameters.push(split[i].trim())
+      }
+    }
+
+    return parameters
+  }
+
+  var PiiRemover = function () {
+    this.stripDatePII = shouldStripDates()
+    this.stripPostcodePII = shouldStripPostcodes()
+    this.queryStringParametersToStrip = queryStringParametersToStrip()
+  }
+
+  PiiRemover.prototype.PIISafe = function (value) {
+    this.value = value
+  }
+
+  PiiRemover.prototype.stripPIIFromString = function (string, stripAllOverride) {
+    var stripped = string.replace(EMAIL_PATTERN, '[email]')
+    stripped = stripped.replace(RESET_PASSWORD_TOKEN_PATTERN, 'reset_password_token=[reset_password_token]')
+    stripped = stripped.replace(UNLOCK_TOKEN_PATTERN, 'unlock_token=[unlock_token]')
+    stripped = stripped.replace(STATE_PATTERN, 'state=[state]')
+    stripped = this.stripQueryStringParameters(stripped)
+
+    if (this.stripDatePII === true || stripAllOverride === true) {
+      stripped = stripped.replace(DATE_PATTERN, '[date]')
+    }
+    if (this.stripPostcodePII === true || stripAllOverride === true) {
+      stripped = stripped.replace(POSTCODE_PATTERN, '[postcode]')
+    }
+    return stripped
+  }
+
+  PiiRemover.prototype.stripPIIFromObject = function (object) {
+    if (object) {
+      if (object instanceof this.PIISafe) {
+        return object.value
+      } else {
+        for (var property in object) {
+          var value = object[property]
+
+          object[property] = this.stripPII(value)
+        }
+        return object
+      }
+    }
+  }
+
+  PiiRemover.prototype.stripPIIFromArray = function (array) {
+    for (var i = 0, l = array.length; i < l; i++) {
+      var elem = array[i]
+
+      array[i] = this.stripPII(elem)
+    }
+    return array
+  }
+
+  PiiRemover.prototype.stripPII = function (value) {
+    if (typeof value === 'string') {
+      return this.stripPIIFromString(value)
+    } else if (Object.prototype.toString.call(value) === '[object Array]' || Object.prototype.toString.call(value) === '[object Arguments]') {
+      return this.stripPIIFromArray(value)
+    } else if (typeof value === 'object') {
+      return this.stripPIIFromObject(value)
+    } else {
+      return value
+    }
+  }
+
+  PiiRemover.prototype.stripQueryStringParameters = function (string) {
+    for (var i = 0; i < this.queryStringParametersToStrip.length; i++) {
+      var parameter = this.queryStringParametersToStrip[i]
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -0,0 +1,134 @@
...
+      }
+    }
+
+    return parameters
+  }
+
+  var PIIRemover = function () {
+    this.stripDatePII = shouldStripDates()
+    this.stripPostcodePII = shouldStripPostcodes()
+    this.queryStringParametersToStrip = queryStringParametersToStrip()
```
andysellick, line 46: I know this is from the original code, but it suddenly seems odd. Why spend three lines here attaching these three functions to `this` when they could be initially declared as e.g. `PIIRemover.prototype.shouldStripDates = function` (etc.) ? Any idea?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -0,0 +1,134 @@
...
+  PIIRemover.prototype.stripPIIFromArray = function (array) {
+    for (var i = 0, l = array.length; i < l; i++) {
+      var elem = array[i]
+
+      array[i] = this.stripPII(elem)
+    }
+    return array
+  }
+
+  PIIRemover.prototype.stripPII = function (value) {
```
andysellick, line 108: Would it make this code more readable if this function was one of the first in the file? This seems like the main function that's called, that then calls other functions.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -0,0 +1,134 @@
...
+      var parameter = this.queryStringParametersToStrip[i]
+      var escaped = parameter.replace(ESCAPE_REGEX_PATTERN, '\\$&')
+      var regexp = new RegExp('((?:\\?|&)' + escaped + '=)(?:[^&#\\s]*)', 'g')
+      string = string.replace(regexp, '$1[' + parameter + ']')
+    }
+
+    return string
+  }
+
+  GOVUK.PIIRemover = PIIRemover
```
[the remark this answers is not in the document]
andysellick: Then we can do the same with other GA4 scripts.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
+# Stripping Personally Identifiable Information (PII)
+
+A module exists for stripping out PII from values. It can accept strings, query parameters, arrays or objects. If the value is an array or object, the child values are looped through and then treated as strings.
+
+99% of the code has been taken from the legacy universal analytics PII redaction code.
```
andysellick, line 5: I think you can remove this line - it's belongs more in the PR description than the documentation.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+- State values become `state=[state]`
+
+Additional values that can be redacted are:
+- Dates - `2022-02-22` becomes `[date]`
+- Postcodes - `sw1a 1aa` becomes `[postcode]`
+
+
+These aren't enabled by default due to the higher likelihood of producing false positives.
+
+Query string parameters are redacted if there is presence of a `meta[name="govuk:static-analytics:strip-query-string-parameters"]` meta tag, with comma separated query string parameters keys in its `content` attribute. The usage of this query string PII removal in production is unknown at this time.
```
andysellick, line 22: Suggest moving this paragraph into the section on query string parameters.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+
+
+### Enabling additional redaction
+
+In order to enable additional redaction you can:
+
+#### Add meta tags to your application via the `Meta Tags` publishing component:
```
andysellick, line 29: Not sure this is a heading, it feels like it belongs with the previous sentence, i.e. 

```
In order to enable additional redaction you can add meta tags to your application via the `Meta Tags` publishing component.
```

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+- Dates - `2022-02-22` becomes `[date]`
+- Postcodes - `sw1a 1aa` becomes `[postcode]`
+
+
+These aren't enabled by default due to the higher likelihood of producing false positives.
+
+Query string parameters are redacted if there is presence of a `meta[name="govuk:static-analytics:strip-query-string-parameters"]` meta tag, with comma separated query string parameters keys in its `content` attribute. The usage of this query string PII removal in production is unknown at this time.
+
+
+### Enabling additional redaction
```
andysellick, line 25: Should be a `##` heading rather than a `###` heading? It feels like a section in its own right, rather than a subheading of `What is redacted`.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+
+In order to enable additional redaction you can:
+
+#### Add meta tags to your application via the `Meta Tags` publishing component:
+
+##### For query string parameters:
```
andysellick, line 31: Nitpick: I don't think headings should end with `:`.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+  local_assigns: {
+    strip_dates_pii: true,
+    strip_postcodes_pii: true
+  }
+} %>
+```
+
+The `shouldStripDates()` and `shouldStripPostcodes()` functions in the JavaScript will then check if these values exist in order to strip dates and postcodes.
+
+However, this will redact dates and postcodes from every call of the PIIRemover. In order to apply extra redaction to specific values, you can use the `stripPIIWithOverride()` function
```
andysellick, line 50: Nitpick: full stop at end of sentence.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+
+### Preventing redaction from running on a value
+
+There is a `PIISafe` class that you can wrap your objects or strings in.
+
+`var myValue = new this.PIIRemover.PIISafe(myValue)`
+
+If `myValue` is then passed through to a `stripPII` function, instead of having its PII redacted, its original value will be returned, bypassing any redaction.
+
+## Basic use
```
andysellick, line 71: I'd suggest moving this section so that it follows the introductory `What is redacted` section.

docs/analytics-gtm/pii-remover.md
```diff
@@ -0,0 +1,103 @@
...
+var example5 = PIIRemover.stripPII(myArray)
+// ['[email]', 'hello world']
+
+// Meta tag of <meta name="govuk:static-analytics:strip-query-string-parameters" content="strip-parameter-1,strip-parameter-2" />
+var myQueryString = '/test?strip-parameter-1=secret&strip-parameter-2=more-secret'
+var example6 = PIIRemover.stripPII(myQueryString)
+// /test?strip-parameter-1=[strip-parameter-1]&strip-parameter-2=[strip-parameter-2]
+```
+
+The modifications from the universal analytics PII code are that:
```
andysellick, line 101: It feels like this section isn't necessary - it's something to explain in the PR or commit messages, but not immediately relevant to anyone in future wanting to find out about this code. We'll be removing the old analytics code eventually, so ideally the documentation should be as standalone as possible.

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.spec.js
```diff
@@ -247,4 +247,34 @@ describe('Google Tag Manager page view tracking', function () {
...
+    document.title = 'example@gov.uk'
+    expected.page.title = '[email]'
+
+    spyOnProperty(document, 'referrer', 'get').and.returnValue('https://gov.uk/example@gov.uk')
+    expected.page.referrer = 'https://gov.uk/[email]'
+
+    // We can't spy on location, so instead we use an anchor link to change the URL temporarily
+
+    // Reset the URL so we can build the expected link string
+    const linkForURLMock = document.createElement('a')
```
andysellick, line 261: Historically we've avoided using `const` as it isn't supported in IE10 or below. Obviously the tests don't run in users browsers, so it's safer here, but ideally the tests should pass in older browsers as well so I think maybe sticking with `var` is safer.

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -0,0 +1,230 @@
+/* eslint-env jasmine, jquery */
```
andysellick, line 1: I think there's no jQuery in here so this could be removed?

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -0,0 +1,230 @@
...
+/* global GOVUK */
+
+describe('GOVUK.PIIRemover', function () {
+  var pii
+
+  beforeEach(function () {
+    resetHead()
```
andysellick, line 8: Does `resetHead` need to be called before every test if it's already called after every test? That seems like it might be unnecessary.

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -0,0 +1,230 @@
...
+        'this is a [postcode] postcode',
+        'this is a [date] date'
+      ]
+
+      arr = pii.stripPII(arr)
+      expect(arr).toEqual(strippedArr)
+    })
+  })
+
+  describe('when there is a a govuk:static-analytics:strip-postcodes meta tag present', function () {
```
andysellick, line 119: Nitpick: `when there is a a`

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -0,0 +1,230 @@
...
+    it('doesn\'t strip out UUIDs in URLs', function () {
+      expect(pii.stripPostcodePII).toEqual(true)
+      var string = pii.stripPII('gov.uk/thing?postcode=sw1a1aa&uuid=d6c2de5d-ef90-45d1-82d4-5f2438369eea')
+      expect(string).toEqual('gov.uk/thing?postcode=[postcode]&uuid=d6c2de5d-ef90-45d1-82d4-5f2438369eea')
+    })
+  })
+
+  describe('when there is a govuk:static-analytics:strip-query-string-parameters meta tag present', function () {
+    afterEach(function () {
+      resetHead()
```
andysellick, line 140: You don't need to call `resetHead` again here - any `afterEach` called at a higher level applies to any tests beneath it, so it's already been called by the main `afterEach`.

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -0,0 +1,230 @@
...
+
+    it('doesn\'t match params without a query string prefix', function () {
+      pageWantsQueryStringParametersStripped(['parameter'])
+      pii = new GOVUK.PIIRemover()
+      var string = pii.stripPII('parameter=secret')
+      expect(string).toEqual('parameter=secret')
+    })
+  })
+
+  describe('when there is a a govuk:static-analytics:strip-dates meta tag present', function () {
```
andysellick, line 172: Nitpick: `when there is a a`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
```diff
@@ -48,15 +50,15 @@
     },
 
     getLocation: function () {
-      return document.location.href
+      return this.PIIRemover.stripPII(document.location.href)
     },
 
     getReferrer: function () {
-      return document.referrer
+      return this.PIIRemover.stripPIIWithOverride(document.referrer, true, true)
```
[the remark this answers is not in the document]
AshGDS: @andysellick This is a consequence of what the implementation guidance said for this value - since referrer talked about worst case scenario redaction, Tara and I agreed to use date and postcode stripping on the referrer.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
AshGDS: @andysellick I'm guessing it's because they're meant to be 'private' methods, so putting it on the `prototype` would allow whoever is using a PIIRemover instance to call them. Also it's attaching the returned value of the functions to those `this` variables, it's not attaching the function itself! Happy to refactor it if you think it should be changed.

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
AshGDS: @andysellick the meta tag functions at the bottom of the file are using jQuery - should I replace them with vanilla JS?

AshGDS:
Thanks for the thorough review @andysellick - I've made most of the changes and responded to a suggestion where I haven't. In hindsight, I shouldn't have copied the PII tests over from the UA section and just added mine to the end, without reviewing the past tests first!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.js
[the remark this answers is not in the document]
andysellick: Okay, that's fine. I still think this is inconsistent but I don't think it should hold up this PR. I'll make a note and we can come back to it later.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
andysellick: Ah of course, thanks.

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Check if primary_publisher nil OR empty in meta_tags ([PR #2829](https://github.com/alphagov/govuk_publishing_components/pull/2829))
 * Add date of birth autocomplete option for date input ([PR #2802](https://github.com/alphagov/govuk_publishing_components/pull/2802))
+* Add personally identifiable information (PII) remover to GTM ([PR #2842](https://github.com/alphagov/govuk_publishing_components/pull/2842))
```
andysellick, line 14: Nitpick: can this be consistent with the PR name? (easiest thing is to just rename the PR).

andysellick (approved):
Great work! 👏

PR #2844, opened 2022-07-01
Rename GA4 analytics attribute state to action

andysellick:
## What / why
Changes the name of the state attribute in the new GA4 code to action.

- rename is to avoid ambiguity
- will also need updating in applications where data attributes are passed

## Visual Changes
None.

Trello card: https://trello.com/c/BRyCFCVf/334-uistate-should-be-uiaction


MartinJJones (approved):
Looks good to me 👍

andysellick:
Thanks @MartinJJones !

PR #2845, opened 2022-07-01
Update popular links in the super navigation header

beccapearce:
## What
Changing the popular links in the search bar to match the popular links on the homepage

## Why
Adding the check benefits link to popular links is a request from content.

### Before
Screenshot 2022-07-01 at 10 58 30

### After
Screenshot 2022-07-01 at 12 45 50

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


richardTowers (approved):
Looks good to me!

PR #2849, opened 2022-07-05
Add file locations to component audits

andysellick:
## What
Adds links to the file locations of components included in applications to the auditing pages of the component guide, in the 'Components' tab.

- previously the 'components by application' section of the auditing listed the components, how many applications used them, and the names of those applications
- now lists the above, plus links to the specific files where the components are used (in github)
- also updates this part of the auditing to only match applications where the component is included in ruby or templates, previously matched for any instance (which meant false positives e.g. would say an application was using a component even if it only included the component asset)

## Why
I think this is going to be useful.

### Before
Screenshot 2022-07-05 at 10 10 14

### After
Screenshot 2022-07-05 at 10 09 41

### DIFF
--- a/app/views/govuk_publishing_components/audit/_components.html.erb
+++ b/app/views/govuk_publishing_components/audit/_components.html.erb
@@ -63,10 +63,28 @@
           <% @components[:components_by_application].each do |component| %>
             <div class="govuk-summary-list__row">
               <dt class="govuk-summary-list__key">
-                <%= component[:component] %> (<%= component[:count] %>)
+                <%= component[:component] %> (<%= pluralize(component[:count], 'use') %>)
               </dt>
               <dd class="govuk-summary-list__value">
-                <%= component[:list] %>
+                <% component[:locations].each do |application| %>
+                  <% github_link = 'https://github.com/alphagov/' + application[:name] + '/blob/main/' %>
+                  <details class="govuk-details govuk-!-margin-bottom-2" data-module="govuk-details">
+                    <summary class="govuk-details__summary">
+                      <span class="govuk-details__summary-text">
+                        <%= application[:name] %> (<%= application[:locations].length %>)
+                      </span>
+                    </summary>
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/audit/_components.html.erb
```diff
@@ -66,7 +66,26 @@
                 <%= component[:component] %> (<%= component[:count] %>)
               </dt>
               <dd class="govuk-summary-list__value">
-                <%= component[:list] %>
+                <p><%= component[:list] %></p>
```
injms, line 69: Could the list of applications that the component is used be shown only once in the `details` element?

For example: 

image

app/views/govuk_publishing_components/audit/_components.html.erb
```diff
@@ -66,7 +66,26 @@
...
+                <% component[:locations].each do |application| %>
+                  <% github_link = 'https://github.com/alphagov/' + application[:name] + '/blob/main/' %>
+                  <details class="govuk-details govuk-!-margin-bottom-2" data-module="govuk-details">
+                    <summary class="govuk-details__summary">
+                      <span class="govuk-details__summary-text">
+                        <%= application[:name] %> (<%= application[:locations].length %>)
```
injms, line 75: Just a suggestion to make it clearer what the numbers mean:

```suggestion
                        <%= application[:name] %> (<%= application[:locations].length %> uses)
```

It should hopefully make it clearer the difference between a number next to the component and the number next to the application.

injms (approved):
No blocking changes - just a couple of suggestions that shouldn't prevent this from being merged in. Super!

app/views/govuk_publishing_components/audit/_components.html.erb
andysellick: Oh yes, that's much more sensible, thanks. Have changed.

PR #2851, opened 2022-07-07
Add white arrow to action link options

patrickpatrickpatrick:
For the new cost of living banner, the design used the already existing
action link component but with a white arrow instead of the default
colours that were available (so that it is the same colour as the link
text). The white arrow .svg/.png have been added, the documentation has
been updated with the new option and a test for the new arrow colour
has been added.


app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -92,6 +92,20 @@
   }
 }
 
+.gem-c-action-link--white-arrow {
+  &:before {
+    content: "";
+    display: table-cell;
```
andysellick, line 98: You don't need to define `content` or `display` here - these are already set on the default `:before` element on this component - https://github.com/alphagov/govuk_publishing_components/blob/d65aeef6119a6a36938b437da7741d079defdae6/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss#L6

andysellick (commented):
Looks good, one small adjustment and please can you add a changelog entry?

chao-xian:
This is how it looks when used on the Homepage:

localhost_3005_

chao-xian (approved):
Looks like Patrick has addressed Andy's comments. I've checked this runs ok.

andysellick (approved):

PR #2852, opened 2022-07-07
Reword component docs implying tracking

andysellick:
## What
Clean up some of the components documentation that talks about tracking and data attributes.

- a lot of components accept data attributes
- some components have built in tracking of some form that can use these data attributes to automatically apply tracking
- docs don't always clarify the difference, updating to clarify
- also cleaning up a few docs where quote marks were being used unnecessarily

## Why
There's been some past confusion over the difference and how tracking happens.

## Visual Changes
To text and examples in documentation pages.


app/views/govuk_publishing_components/components/docs/cookie_banner.yml
```diff
@@ -17,34 +17,34 @@ examples:
...
     description: Use this type of banner in your service if it only uses cookies for analytics.
     data:
       title: Can we store analytics cookies on your device?
       text: Analytics cookies help us understand how our website is being used.
       confirmation_message: You’ve accepted all cookies. You can `<a class='govuk-link' href='/cookies'>change your cookie settings</a>` at any time.
       services_cookies:
-        "yes":
-          text: "Yes"
+        yes:
+          text: Yes
```
injms, line 41: I think that `yes` and `no` are being converted to booleans, thanks to YAMLs "Norway Problem"; this is what comes through in the guide:

image

The fix is to quote any instance of "yes" or "no":

```suggestion
        "yes":
          text: "Yes"
```

PR #2853, opened 2022-07-07
Fix broken preview link for "previous and next navigation" component

jon-kirwan:
## What

Fix broken preview link for "previous and next navigation" component

## Why

To see the preview of the default ("previous" only preview)


andysellick (approved):
Nice spot, well done.

PR #2862, opened 2022-07-15
Removes additional `id` argument from `table_helper`

davidtrussler:
## What
The changes in this PR removes an argument in the `table_helper` that had been added in a previous piece of work, and was intended to allow a value for an `id` parameter to be passed to the table component. This value is now added to the `opt` argument instead.

## Why
While the initial change was not problematic within this repo, the Table Helper is also used outside the repo. When calls are made to it from elsewhere only two arguments are assumed leading to unexpected results in the table component there.

## Visual Changes
There should be no visual changes to the table component. The important check is that the table retains an `id` parameter in the mark-up.

Screenshot 2022-07-15 at 11 38 31

### DIFF
--- a/lib/govuk_publishing_components/app_helpers/table_helper.rb
+++ b/lib/govuk_publishing_components/app_helpers/table_helper.rb
@@ -1,7 +1,7 @@
 module GovukPublishingComponents
   module AppHelpers
     class TableHelper
-      def self.helper(context, caption = nil, id = nil, opt = {})
+      def self.helper(context, caption = nil, opt = {})
         builder = TableBuilder.new(context.tag)
 
         classes = %w[gem-c-table govuk-table]
@@ -10,7 +10,7 @@ module GovukPublishingComponents
         caption_classes = %w[govuk-table__caption]
         caption_classes << opt[:caption_classes] if opt[:caption_classes]
 
-        context.tag.table class: classes, id: id do
+        context.tag.table class: classes, id: opt[:table_id] do
           context.concat context.tag.caption caption, class: caption_classes
           yield(builder)
         end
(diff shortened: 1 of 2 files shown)


andysellick (approved):
Good job.

DilwoarH (commented):
Looks great - a test would make it even better :)

lib/govuk_publishing_components/app_helpers/table_helper.rb
```diff
@@ -1,3 +1,6 @@
+# This is part of the Public API of this gem
+# Any changes to the signature will require a major version change
```
kevindew, line 2: Thanks for this. I'd move this to be above the self.helper method - since that method is part of the public API and that is the place with the signature.

I'd also suggest that it's only non-backwards compatible changes to the signature.

davidtrussler: Ah yes, makes sense. I've updated the comment.

DilwoarH (approved):
LGTM 🥇

DilwoarH (changes requested):
We need a change note

DilwoarH:
> We need a change note

It will be a breaking change as it'll break the helpers currently using the ID argument.

kevindew:
> > We need a change note
> 
> It will be a breaking change as it'll break the helpers currently using the ID argument.

It'll be ok to record it as a patch release given the breaking change was a bug.

davidtrussler:
> > > We need a change note
> > 
> > 
> > It will be a breaking change as it'll break the helpers currently using the ID argument.
> 
> It'll be ok to record it as a patch release given the breaking change was a bug.

OK, well right now I've added it to the CHANGELOG under `Unreleased`. When it is released it will be version `29.15.1`, bumped from `29.15.0` if that sounds right.

DilwoarH (approved):

PR #2864, opened 2022-07-18
GA4 analytics schema rework

andysellick:
## What
Rework the code based on changes to the GA4 schema. This includes:

- the creation of a new file that contains the schema for events, which is then used by the event click tracking script (hopefully we can use it for other schemas as well)
- restructuring of the page view data
- some new documentation
- miscellaneous other changes (see commit messages for details)

## Why
We're redesigning the data schema for events.

## Visual Changes
None.

Trello card: https://trello.com/c/obytEbqo/153-update-existing-code-with-schema-changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* GA4 analytics schema rework ([PR #2864](https://github.com/alphagov/govuk_publishing_components/pull/2864))
+
 ## 29.15.1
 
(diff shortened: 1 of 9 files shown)


AshGDS (approved):
Looks good, great work @andysellick 

I think it's probably worth removing your example from the `accordion.yml` since this implementation can't be reflected properly in the `.yml` due to the use of JSON's. Maybe link to the click tracking documentation somewhere in the `.yml` instead?

andysellick:
> I think it's probably worth removing your example from the `accordion.yml` since this implementation can't be reflected properly in the `.yml` due to the use of JSON's. Maybe link to the click tracking documentation somewhere in the `.yml` instead?

Good shout. Think I'll address that separately so we can get this merged and unblock some stuff.

PR #2870, opened 2022-07-22
Add link click tracking

AshGDS:
Hi @andysellick,

Would you be able to approve this PR? Thanks!

## What

Adds GTM link click tracking to the repo.

When initialised, the JS adds an event listener to the `<body>` of the page with the following listeners:

- A `click` listener (for left clicks, control clicks, and "Meta key" clicks - "Meta key" clicks are command + clicks or windows key + clicks.)
- A `mousedown` listener (for middle clicks.)
- A `contextmenu` listener (for right clicks / context menu keypresses.)

When one of these listeners are fired, they check if the `event.target` is an `<a>` tag with a `href`. If so, it will then check if the link is:

- A `download` link - This is either a href to `assets.publishing.service.gov.uk` or a href on the `www.gov.uk` domain pointing to the `/government/uploads/` path.
- An `external` link - This is a href that is not on the `www.gov.uk` domain. 
    - It also checks if a `href` starts with `/` and does not start with `//` to determine if it's a relative link like `/bank-holidays` , but not a protocol relative link like `//google.com`

If the href matches one of the above, it is pushed to the dataLayer using the event schema.

## Why

So that we can track left clicks, right clicks, middle clicks, control + clicks and command + clicks on the following types of links: external links, download links, and mailto: links.

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      if (href.length < 7) {
+        return false
+      }
+      return href.substring(0, 7) === 'mailto:'
+    },
+
+    isDownloadLink: function (href) {
+      var assetsDomain = 'assets.publishing.service.gov.uk/'
+      var uploadsPath = '/government/uploads/'
+      if (!href) {
```
andysellick, line 100: Nitpick: tiny optimisation, this guard clause could be the first thing in this function.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+        return false
+      }
+
+      if (this.hrefPointsToDomain(href, assetsDomain)) {
+        return true
+      }
+
+      var isInternalLink = this.hrefPointsToDomain(href, this.internalLinksDomain) || this.hrefPointsToDomain(href, this.internalLinksDomain.replace('www.', ''))
```
andysellick, line 108: I'd suggest having a second variable for `this.internalLinksDomain.replace('www.', '')` rather than recalculating it each click e.g. `this.internalLinksDomainWithoutWww = 'gov.uk/'`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      if (this.stringStartsWith(href, uploadsPath)) {
+        return true
+      }
+    },
+
+    isExternalLink: function (href) {
+      if (!href) {
+        return false
+      }
+      var isInternalLink = this.hrefPointsToDomain(href, this.internalLinksDomain) || this.hrefPointsToDomain(href, this.internalLinksDomain.replace('www.', ''))
```
andysellick, line 123: This looks like a repetition of line 108 - would it be possible to calculate this once and store the value somehow rather than calling it twice?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      if (!isInternalLink && !this.hrefIsRelative(href)) {
+        return true
+      }
+    },
+
+    trackClickEvent: function (schema) {
+      window.dataLayer = window.dataLayer || []
```
andysellick, line 130: The dataLayer should be defined once by the GTM code snippet, so the approach I've been taking is to do something like

```
if (window.dataLayer) {
  window.dataLayer.push(thing)
}
```

...but happy to discuss the approach.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      this.stringStartsWith(href, schemaRelativeDomain)
+    },
+
+    stringStartsWith: function (string, stringToFind) {
+      return string.substring(0, stringToFind.length) === stringToFind
+    },
+
+    hrefIsRelative: function (href) {
+      // Checks that a link is relative, but is not a protocol relative url
+      return this.stringStartsWith(href, '/') && !this.stringStartsWith(href, '//')
```
andysellick, line 150: If I understand what this is doing correctly, I think a slightly more efficient way of doing it might be:

```
return href[0] === '/' && href[1] !== '/'
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+    }
+
```
andysellick, line 152: Nitpick: empty line.

docs/analytics-gtm/gtm-link-click-tracker.md
```diff
@@ -0,0 +1,62 @@
+# Google Tag Manager link click tracker
+
+A module exists for tracking link clicks on GOV.UK.
```
andysellick, line 3: Maybe `A module for tracking link clicks on GOV.UK.`?

docs/analytics-gtm/gtm-link-click-tracker.md
```diff
@@ -0,0 +1,62 @@
...
+
+When initialised, the JS adds an event listener to the `<body>` of the page with the following listeners:
+
+- A `click` listener (for left clicks, control clicks, and "Meta key" clicks - "Meta key" clicks are command + clicks or windows key + clicks.)
```
andysellick, line 7: Genuine interest - what's a windows key + click? I thought if you press the windows key it opens the Start menu?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      }
+
+      if (!element) {
+        return
+      }
+
+      var clickData = {}
+      var href = element.getAttribute('href')
+
+      if (this.isMailToLink(href)) {
```
andysellick, line 30: Each of the three functions in this `if` (`isMailtoLink`, `isDownloadLink` and `isExternalLink`) first check for the existence of `href` and return false if it's not there. I think it would be more efficient to have that check once, here, prior to the `if` - if the script can't find a `href` attribute then it might as well finish here.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+        this.handleClick(event)
+      }
+    },
+
+    isMailToLink: function (href) {
+      if (!href) {
+        return false
+      }
+
+      if (href.length < 7) {
```
andysellick, line 91: What's the value of this guard clause? `length` is probably a quicker thing to execute than `substring` so it's fractionally more efficient if the href definitely isn't a mailto link, but it means that for every link that is a mailto, it always has to do both.

Had a quick look for which is quicker but can't find anything - I'd be tempted to just have the `substring` call, but happy to discuss.

docs/analytics-gtm/gtm-link-click-tracker.md
```diff
@@ -0,0 +1,62 @@
...
+    - If the `href` does not meet both of these criteria, then it is considered an external link.
+
+## Basic use
+
+```
+//= require ./analytics-ga4/gtm-schemas
+//= require ./analytics-ga4/gtm-link-click-tracker
+
+window.GOVUK.analyticsGA4.linkClickTracker.trackLinkClicks()
+```
```
andysellick, line 27: I'm not sure we need to document how to include the script, as we'll be including the script in GOV.UK anyway. Also we won't be including it in this way (once we integrate it into the cookies code) so I wouldn't want someone unfamiliar with the analytics code copying and pasting this into an application.

docs/analytics-gtm/gtm-link-click-tracker.md
```diff
@@ -0,0 +1,62 @@
...
+
+`<a href="https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/742746/A_quick_guide_to_govt_healthy_eating_update.pdf">A Quick Guide to the Government’s Healthy Eating Recommendations</a>`
+
+In the example above, on a left click of the link, the following would be pushed to the dataLayer, using the `eventSchema` found in `gtm-schemas.js`:
+
+
```
andysellick, line 33: Nitpick: double empty line.

spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.spec.js
```diff
@@ -0,0 +1,416 @@
+/* eslint-env jasmine, jquery */
+
+describe('GOVUK.analyticsGA4.linkClickTracker', function () {
+  'use strict'
+  var GOVUK = window.GOVUK
+  var $ = window.jQuery
```
andysellick, line 6: This might be a pain, but can we avoid using jQuery? We're going to have to eventually remove all jQuery from the site, even tests, so the frontend team agreed a while ago not to add any more.

andysellick: I suspect one thing that will be tricky is removing the click event listeners, but there's ways around this if it's a problem, let me know.

andysellick (commented):
Looks really good. I've made some initial comments about the code but haven't tested it locally yet.

app/assets/javascripts/govuk_publishing_components/analytics-ga4.js
```diff
@@ -2,6 +2,8 @@
 //= require ./analytics-ga4/gtm-schemas
 //= require ./analytics-ga4/pii-remover
 //= require ./analytics-ga4/gtm-page-views
+//= require ./analytics-ga4/gtm-link-click-tracker
 //= require ./analytics-ga4/gtm-click-tracking
 
 window.GOVUK.Gtm.sendPageView() // this will need integrating with cookie consent before production
+window.GOVUK.analyticsGA4.linkClickTracker.trackLinkClicks()
```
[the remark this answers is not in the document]
AshGDS: Yep - I think you suggested analyticsGA4 in a previous PR and I'm happy to stick with that!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+      var href = element.getAttribute('href')
+
+      if (this.isMailToLink(href)) {
+        clickData.type = 'email'
+        clickData.external = 'true'
+      } else if (this.isDownloadLink(href)) {
+        clickData.type = 'download'
+        clickData.external = 'false'
+      } else if (this.isExternalLink(href)) {
+        clickData.type = 'generic link'
```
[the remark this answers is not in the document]
AshGDS: @andysellick Yep this is from the implementation guide. I'm guessing this is because compared to a mailto and download links, these are "generic links" - but I agree that external link would make more sense.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,156 @@
...
+        clickData.external = 'true'
+      }
+
+      if (Object.keys(clickData).length > 0) {
+        clickData.event_name = 'navigation'
+        clickData.text = element.textContent.trim()
+        clickData.url = href
+        clickData.link_method = this.getClickType(event)
```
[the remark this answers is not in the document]
AshGDS: @andysellick This was from a conversation between Tara and I when talking about adding left, right, middle, command click and control click to the code - she ran this by Jeff and he said:

"Assuming the developers can pass the specific click type, adding an additional parameter should be fine although I’d probably go with link_method or link_type to fit in with the other default link parameters."

docs/analytics-gtm/gtm-link-click-tracker.md
AshGDS: @andysellick `command` and `windows key` are considered meta keys. If you pressed the windows key alone it would open the start menu, but not if you  held it and clicked at the same time. It looks like Windows key + Click doesn't do anything different on links, but on a Mac the meta key + click opens the link in a new tab. 

I doubt anyone on a Windows machine will do a Windows Key + Click so I don't think we will need to check in the JS wether they are on a Mac when tracking this. But if you think we should ensure they are on a Mac first, then I can add an if statement to check which platform the user is on. 

I believe you can filter GA reports by Operating System as well, so if we were getting Windows Key + Click events they could be filtered out to only show Command + Click events.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
AshGDS: @andysellick This was me incorrectly thinking that the JS would crash if you ran `substring(0, 7)` on a string shorter than 7 characters. Happy to remove it!

AshGDS:
Hi @andysellick 

I've removed jQuery from the tests, resolved the code issues, and have left conversations unresolved that you may want to respond to.

There's a weird bug with the new click tracking tests. It looks like depending on the order that tests are run, my new link click tracking tests receive this error:

`GOVUK.analytics.setDimension is not a function thrown`

Which I think might be because the old link click listeners are trying to run while my new link click tests are running. I've tried to solve it by importing `GOVUK.analytics.setDimension` into my tests `beforeEach()` but the error still happens depending on the order the tests are run. What should we do about this? Should I try and investigate further? What is weird is that the old tests have code in them to remove their listeners after each test as well.

Thanks :+1: 

image

docs/analytics-gtm/gtm-link-click-tracker.md
andysellick: No that's fine I was just interested, thanks for the info.

AshGDS:
@andysellick I'm happy for this to be merged by someone else if it gets approved while I'm away :+1:

JamesCGDS:
Hi @andysellick,

It looks like there are some unresolved conversations in the PR but I think @AshGDS may have addressed them - I'm happy to approve this if you are?

andysellick:
@JamesCGDS I think this needs fewer commits, but I'll take another look over the code as well.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,157 @@
...
+    },
+
+    isExternalLink: function (href) {
+      var isInternalLink = this.hrefPointsToDomain(href, this.internalLinksDomain) || this.hrefPointsToDomain(href, this.internalLinksDomainWithoutWww)
+      if (!isInternalLink && !this.hrefIsRelative(href)) {
+        return true
+      }
+    },
+
+    trackClickEvent: function (schema) {
```
andysellick, line 130: Not sure this needs to be a separate function - it's only a few lines and it's only called from one place.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/gtm-link-click-tracker.js
```diff
@@ -0,0 +1,157 @@
...
+      if (window.dataLayer) {
```
andysellick, line 131: We should definitely check earlier in the code for the presence of the dataLayer, probably right at the start.

docs/analytics-gtm/gtm-link-click-tracker.md
```diff
@@ -0,0 +1,54 @@
...
+```
+{
+    "event": "analytics",
+    "event_data": {
+        "event_name": "navigation",
+        "type": "download",
+        "url": "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/742746/A_quick_guide_to_govt_healthy_eating_update.pdf",
+        "text": "A Quick Guide to the Government’s Healthy Eating Recommendations",
+        "index": "n/a",
```
andysellick, line 34: We should update these `n/a` examples as they're now `null`.

andysellick (commented):
Couple of small things that I think we should address before merging. Also this PR is currently 24 commits - we can definitely merge some of them together.

JamesCGDS:
Due to issues with merging commits, this PR has been replaced by PR 2904 - https://github.com/alphagov/govuk_publishing_components/pull/2904.

**Credit goes to @AshGDS for writing the code used in PR 2904.**

PR #2877, opened 2022-07-26
Add helper use to auditing

andysellick:
## What
Extend the auditing to include details of where applications are referencing code from the gem.

## Why
We've had more than one instance where changing something in the gem has caused an error in an application because the application relied upon that code. This can give developers insight into where gem code is being used outside of the gem, and hopefully avoid similar future errors.

## Visual Changes
Extra bit added to the bottom of the 'components' tab of the auditing page.

Screenshot 2022-07-26 at 15 22 53


injms (approved):
Good idea! Thanks for adding this functionality in 👍

PR #2878, opened 2022-07-27
Update GA4 analytics page view data structure

andysellick:
## What
Change the data structure/schema of a GTM page view.

- sub object should be `page_view` rather than `page`

## Why
Because I got it wrong the first time.

## Visual Changes
None.


gclssvglx (approved):
Looks great - thanks @andysellick 👍

PR #2884, opened 2022-07-29
Ensure tab clicks grab the tabs href for gtm

AshGDS:
Hi @andysellick 

Would you be able to approve this? Though, I'm wondering if we should let our new dev James review it first as a task, and if there any requested changes we can see if he would like to make them. Thanks :+1:

## What

This makes it so our GTM event click tracker grabs the `href` of a tab when it is clicked. For example, if a 'More information' tab is clicked, the value of the url key should be: `url: #more-information` 

This uses `event.target` to ensure that an `<a>` tag was clicked on. You can't use the existing `target` variable as this has already traversed up the DOM due to the use of `.closest`, so it wouldn't let us know if what was clicked on was the tab itself - it just gives us the element with `data-ga4` on it.

## Why

When clicking tabs, the `url` value in our GTM schema was passing through the current browser URL. This is an issue because our click tracking runs first, before the tab has changed. Therefore it was grabbing the browser URL before the tab had changed, so it was outdated. For example if you were on `/coronavirus` and clicked on the `More information` tab, our tracking should track either `#more-information` or `/coronavirus#more-information` as this is useful to the analytics. However because we were grabbing the page URL, it was always being displayed as `/coronavirus` as our GTM code runs before the tab code which changes the URL.

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Expand print link component ([PR #2900](https://github.com/alphagov/govuk_publishing_components/pull/2900))
 * Add link click tracking ([PR #2904](https://github.com/alphagov/govuk_publishing_components/pull/2904))
+* Ensure tab clicks grab the tabs href for gtm ([PR #2884](https://github.com/alphagov/govuk_publishing_components/pull/2884))
 
 ## 30.1.0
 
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Looks good. I'm slightly concerned that this code is getting a bit specific and has a few edge cases in it, which feels brittle, but I think let's go for it for now and review/discuss the code in general soon.

Also please add a CHANGELOG entry, thanks 👍

PR #2889, opened 2022-08-03
Update GA4 schema to use null instead of 'n/a' for undefined values

andysellick:
## What
Change the GA4 tracking code so that empty values passed to GA4 are now `null` and not `n/a`.

Had a brief thought about using `undefined` instead of `null` but decided on `null` on the basis that null is an assignment value, meant for assigning to a variable as a representation of no value (apparently).

Note that this code is only enabled on integration.

## Why
Currently we're sending `n/a` values for any event_data parameters that don't have a value in order to nullify previously passed data. We believe that GA4 will automatically return `(not set)` for any dimensions that don't have a value. So we think it's safe to not track a value at all, rather than tracking `n/a`.

## Visual Changes
None.

Trello card: https://trello.com/c/HoDiZaWd/348-push-null-values-rather-than-n-a-in-eventdata-parameters

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.spec.js
@@ -8,6 +8,7 @@ describe('Google Tag Manager page view tracking', function () {
   beforeEach(function () {
     saved.title = document.title
     document.title = 'This here page'
+    var nullValue = null
     expected = {
       event: 'page_view',
       page_view: {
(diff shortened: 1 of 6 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/gtm-page-views.spec.js
```diff
@@ -16,30 +16,30 @@ describe('Google Tag Manager page view tracking', function () {
...
-
-        section: 'n/a',
-        taxon_slug: 'n/a',
-        taxon_id: 'n/a',
-        themes: 'n/a',
-        taxon_slugs: 'n/a',
-        taxon_ids: 'n/a',
-
-        language: 'n/a',
+        document_type: null,
```
JamesCGDS, line 19: Might there be any mileage in storing `null` in a variable and then using the variable as the value? Just thinking if we needed to change `null` to something else in future, we'd only need to change the variable

andysellick: Good shout, have updated.

JamesCGDS (approved):

Issue #2863, opened 2022-07-18
🏞 Image card component - image lacks `width` and `height` attribute


injms:
The image in the image card component doesn't currently have the ability to use the `width` and `height` attribute; at the moment the image markup looks like this:

```html
<img 
  alt="" 
  class="gem-c-image-card__image"
  loading="lazy" 
  sizes="(max-width: 640px) 100vw, (max-width: 1020px) 33vw, 300px" 
  src="image-promo.jpg"
  srcset="image-promo.jpg 610w, image-promo-480.jpg 480w, image-promo-320.jpg 320w, image-promo-240.jpg 240w, image-promo-170.jpg 170w" 
/>
```

Setting a width and a height will allow the browser to know how much space the image should take up. Browsers use these dimensions to calculate the image's aspect ratio and reserve the space needed in the page layout - this reduces the amount of layout shift on the page.[^1]

The width and height are overridden by the page by the presence of any CSS targeting the image - as the image card component has styles for the image, this won't affect the visual presentation once loaded. But this should make the experience whilst the page is loading better as it will help avoid page jank aka cumulative layout shift.

As GOVUK (should) only be using images with a 3:2 aspect ratio[^2], this could be hardcoded into the component rather than added as an extra option. Because it's only the aspect ratio that the browser requires - not the actual pixel values of the image - the values could be set to just 300 and 200.

[^1]: For more details on why setting the `width` and `height` attribute is a good thing, see [Optimize Cumulative Layout Shift: Images without dimensions](https://web.dev/optimize-cls/#images-without-dimensions).
[^2]: https://design-system.service.gov.uk/styles/images/#aspect-ratio

jon-kirwan:
I've re-opened the issue as unfortunately the fixed width and height attributes are leading to incorrectly sized images in some places:

![www gov uk_government_organisations_attorney-generals-office (1)](https://user-images.githubusercontent.com/87758239/180800188-cb5060fb-53bd-46ca-b1bf-5328a37e7d48.png)
![www gov uk_government_organisations_attorney-generals-office](https://user-images.githubusercontent.com/87758239/180800236-5009377e-a532-4cf5-b594-1190e7bb8557.png)

PR #2895, opened 2022-08-05
Include size attributes on image card component

jon-kirwan:
## What
https://trello.com/c/iz7WJaeG/1404-include-width-and-height-size-attributes-on-image-card-component-2869

Include width and height size attributes on image card component

- https://components-gem-pr-2895.herokuapp.com/component-guide/image_card
- https://components-gem-pr-2895.herokuapp.com/component-guide/image_card/large_version

## Why
See https://github.com/alphagov/govuk_publishing_components/issues/2863

### CLS (before)

https://user-images.githubusercontent.com/87758239/183419165-41224472-c3b2-4bb2-bf3a-3e786f6b190f.mov

### CLS (after)

https://user-images.githubusercontent.com/87758239/183419124-d9d54a1b-3a6c-4e23-b06d-772eefce2f3b.mov

## Testing

https://www.integration.publishing.service.gov.uk/government/organisations/attorney-generals-office (which includes the image component with `large` and standard options) in -

- [x] Internet Explorer 11 on Windows 10 
- [x] Edge 103 on Windows 10 
- [x] Chrome 104 on Windows 10 
- [x] Chrome 104 on macOS Monterey
- [x] Safari 15.3 on macOS Monterey
- [x] Firefox 103 on macOS Big Sur
- [x] Chrome on Galaxy S22
- [x] Safari on iPhone 13

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/image_card.yml
+++ b/app/views/govuk_publishing_components/components/docs/image_card.yml
@@ -4,6 +4,8 @@ body: |
   An image and links, meant for use for news articles and people. Includes optional paragraph and additional links.
 
   The component is generally to be used within a grid column but has no grid of its own, so the text and the images in the examples below may not always line up. This will normally look tidier, for example [on pages like this](https://www.gov.uk/government/organisations/attorney-generals-office).
+
+  Images should have an aspect ratio of 3:2.
 accessibility_criteria: |
   The component must:
 
@@ -12,6 +14,10 @@ accessibility_criteria: |
   - if the contents of the component are in a different language than the rest of the document, include an appropriate `lang` attribute to correctly identify the language used in the component
 shared_accessibility_criteria:
   - link
+embed: |
+  <div class="govuk-!-width-one-third">
+    <%= component %>
+  </div>
 examples:
   default:
     data:
@@ -195,6 +201,10 @@ examples:
         text: "Announcement"
       heading_text: "Something has happened nearby possibly"
       description: "Following a news report that something has happened, further details are emerging of the thing that has happened and what that means for you."
+    embed: |
+      <div class="govuk-!-width-full">
+        <%= component %>
+      </div>
   with_data_attributes:
     description: Data attributes can be applied as required. Note that the component does not include built in tracking. If this is required consider using the [track click script](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics/track-click.md).
     data:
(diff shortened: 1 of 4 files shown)


andysellick (approved):
Nice work @jon-kirwan 👏

andysellick:
Only one thing I've just noticed when reviewing the percy diff... this change has restricted the final example (with srcset) to these new dimensions, where previously it was a much bigger image. Might need to check that won't cause a problem first.

jon-kirwan:
This is a little tricky. I think it could be an issue because width and height attributes override image dimensions in the `srcset`.

From web.dev, [modern best practice](https://web.dev/optimize-cls/#modern-best-practice) is as follows:

> What about responsive images?
When working with [responsive images](https://web.dev/serve-responsive-images), `srcset` defines the images you allow the browser to select between and what size each image is. To ensure `` width and height attributes can be set, each image should use the same aspect ratio.
```
<img
  width="1000"
  height="1000"
  src="puppy-1000.jpg"
  srcset="puppy-1000.jpg 1000w, puppy-2000.jpg 2000w, puppy-3000.jpg 3000w"
  alt="Puppy with balloons"
/>
```

So, I think the correct approach is to have one set of width and height attributes for all variants (300 x 200) but to include height and image attributes in the CSS like so:

```
height: auto;
width: 100%;
```

Then the image can be resized to the width of this container.

This works fine on GOV.UK pages where images are constrained by the Design System layout.

But on the component page - the images scale up to around 960px (the width of `govuk-width-container`).

Can we include layout elements on the component guide (for this component only)?

injms:
> Can we include layout elements on the component guide (for this component only)?

Yes, I think that `embed` will do that, for example:

```yaml
varation_name:
  description: This is a description.
  data:
    key: pair
  embed: |
    <div class="something-something-darkside">
      <%= component %>
    </div>
```

jon-kirwan:
Thanks @injms for the 'embed' suggestion.

There are two things to highlight with reference to the component guide:

- All example code snippets now contain additional layout markup. I'm not sure if I should add some information to explain that?
- Additional text, will now wrap with the column. For example, `with_more_information` option - [before](https://components.publishing.service.gov.uk/component-guide/image_card/with_more_information/preview) and [after](https://components-gem-pr-2895.herokuapp.com/component-guide/image_card/with_more_information/preview)

app/views/govuk_publishing_components/components/docs/image_card.yml
```diff
@@ -283,3 +293,7 @@ examples:
         /assets/govuk_publishing_components/image-card-srcset/cop26-320.jpg: 320w
         /assets/govuk_publishing_components/image-card-srcset/cop26-240.jpg: 240w
         /assets/govuk_publishing_components/image-card-srcset/cop26-170.jpg: 170w
+    embed: |
+      <div class="govuk-!-width-full">
```
injms, line 297: I think we should wrap this in a one-third container, otherwise it looks odd when it goes full width:

```suggestion
      <div class="govuk-!-width-one-third">
```

injms: Actually, ignore me on this! The image diff shows the main branch version at full width - the image is broken on the live site which confused me.

Issue #2167, opened 2021-06-28
Investigate how the checkboxes enhancement in `govuk-frontend` v3.13.0 integrates with the local implementation


alex-ju:
Investigate how the [checkboxes.js](https://github.com/alphagov/govuk-frontend/blob/main/src/govuk/components/checkboxes/checkboxes.js) enhancement released in [`govuk-frontend` v3.13.0](https://github.com/alphagov/govuk_publishing_components/pull/2164) integrates with [the local enhancement](https://github.com/alphagov/govuk_publishing_components/blob/master/app/assets/javascripts/govuk_publishing_components/components/checkboxes.js) in `govuk_publishing_components`. They both try to cover the same feature of providing 'exclusive' checkbox items that result in unselecting the other options.

PR #2896, opened 2022-08-08
Remove GOV.UK specific code for handling exclusive checkboxes

andysellick:
## What
Remove GOV.UK specific code for handling the [exclusive checkboxes option](https://components.publishing.service.gov.uk/component-guide/checkboxes/checkbox_with_exclusive_item).

## Why
This is already included in `govuk-frontend` as of v.3.13.0.

Fixes https://github.com/alphagov/govuk_publishing_components/issues/2167

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,9 +6,10 @@
 - Don't include changes that are purely internal. The CHANGELOG should be a
   useful summary for people upgrading their application, not a replication
   of the commit log.
-  
+
 ## Unreleased
 
+* Remove GOV.UK specific code for handling exclusive checkboxes ([PR #2896](https://github.com/alphagov/govuk_publishing_components/pull/2896))
 * Add links and sections count tracking for Cost of living hub ([PR #2921](https://github.com/alphagov/govuk_publishing_components/pull/2921))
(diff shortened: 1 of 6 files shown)


jon-kirwan (commented):
Nice work. I've just one query. I think the following lines can be removed from the helper since they don't appear to be referenced in the view now (previously added here https://github.com/alphagov/govuk_publishing_components/commit/c95180272c2471cca8ee9e0d4f34f3f6a39aacc0)

https://github.com/alphagov/govuk_publishing_components/blob/c8552cc05b81abb6df8ce331e627b4edc1ff2ea9/lib/govuk_publishing_components/presenters/checkboxes_helper.rb#L18

and

https://github.com/alphagov/govuk_publishing_components/blob/c8552cc05b81abb6df8ce331e627b4edc1ff2ea9/lib/govuk_publishing_components/presenters/checkboxes_helper.rb#L32

andysellick:
> I think the following lines can be removed from the helper since they don't appear to be referenced in the view now (previously added here [c951802](https://github.com/alphagov/govuk_publishing_components/commit/c95180272c2471cca8ee9e0d4f34f3f6a39aacc0))

Great spot, thanks @jon-kirwan! Have removed.

jon-kirwan (approved):
All good thanks! 🚀

PR #2898, opened 2022-08-09
Fix auditing path cleaning

andysellick:
## What
Fixes a bug in the component auditing, in the section that shows where components are used in applications.

- fixes a bug where file paths that contained a + character were being cut short, resulting in incorrect links to component references in applications
- example was `guide.html+print.erb` in government-frontend, which was incorrectly coming out as `guide.html`

## Why
Was generating links to files that didn't exist.

## Visual Changes
None.

### DIFF
--- a/app/models/govuk_publishing_components/audit_applications.rb
+++ b/app/models/govuk_publishing_components/audit_applications.rb
@@ -150,7 +150,7 @@ module GovukPublishingComponents
     end
 
     def clean_file_path(file)
-      file[/(?<=#{Regexp.escape(@path.to_s)}\/)[\/a-zA-Z_-]+.[a-zA-Z.]+/]
+      file[/(?<=#{Regexp.escape(@path.to_s)}\/)[\/a-zA-Z_-]+.[a-zA-Z+.]+/]
     end
 
     def clean_file_name(name)


app/models/govuk_publishing_components/audit_applications.rb
```diff
@@ -150,7 +150,7 @@ def find_code_references(file, src, regex)
     end
 
     def clean_file_path(file)
-      file[/(?<=#{Regexp.escape(@path.to_s)}\/)[\/a-zA-Z_-]+.[a-zA-Z.]+/]
+      file[/(?<=#{Regexp.escape(@path.to_s)}\/)[\/a-zA-Z_-]+.[a-zA-Z+.]+/]
```
injms, line 153: Just double checking that we don't want to escape the `+` as it's a special character?

andysellick: Added, can't hurt.

andysellick: Ah, no, it can hurt. Linter throws up this:

```
Offenses:

app/models/govuk_publishing_components/audit_applications.rb:153:69: C: [Correctable] Style/RedundantRegexpEscape: Redundant escape inside regexp literal
      file[/(?<=#{Regexp.escape(@path.to_s)}\/)[\/a-zA-Z_-]+.[a-zA-Z\+.]+/]
                                                                    ^^

209 files inspected, 1 offense detected, 1 offense autocorrectable
```

There might be a way around this linting issue but I think the original code is fine - the linter seems to think escaping isn't needed, and the code seems to work as intended.

injms: Super! Thanks for checking 👍

injms (approved):

PR #2904, opened 2022-08-15
Gtm add external link tracking

JamesCGDS:
Hi @andysellick,

Would you be able to approve this PR? Thanks!

## What

Adds GTM link click tracking to the repo.

When initialised, the JS adds an event listener to the `<body>` of the page with the following listeners:

- A `click` listener (for left clicks, control clicks, and "Meta key" clicks - "Meta key" clicks are command + clicks or windows key + clicks.)
- A `mousedown` listener (for middle clicks.)
- A `contextmenu` listener (for right clicks / context menu keypresses.)

When one of these listeners are fired, they check if the `event.target` is an `<a>` tag with a `href`. If so, it will then check if the link is:

- A `download` link - This is either a href to `assets.publishing.service.gov.uk` or a href on the `www.gov.uk` domain pointing to the `/government/uploads/` path.
- An `external` link - This is a href that is not on the `www.gov.uk` domain. 
    - It also checks if a `href` starts with `/` and does not start with `//` to determine if it's a relative link like `/bank-holidays` , but not a protocol relative link like `//google.com`

If the href matches one of the above, it is pushed to the dataLayer using the event schema.

## Why

So that we can track left clicks, right clicks, middle clicks, control + clicks and command + clicks on the following types of links: external links, download links, and mailto: links.

## Visual Changes

None.


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Expand print link component ([PR #2900](https://github.com/alphagov/govuk_publishing_components/pull/2900))
+* Add link click tracking ([PR #2904](https://github.com/alphagov/govuk_publishing_components/pull/2904))
```
andysellick, line 13: Nitpick - can this match the PR name? (either change it here or rename the PR is fine)

andysellick (approved):
Looks good. I'd be tempted to merge the linting fixes into an earlier commit if possible, and merge the two docs commits together, but otherwise fine. You should close the corresponding PR and refer to this one, and give @AshGDS credit 😁

PR #2906, opened 2022-08-17
Update gtm naming conventions

JamesCGDS:
## What
This PR renames several analytics modules and files in order to make naming conventions more consistent and generic (e.g. any instances of `gtm-` have been replaced with `ga4-`). For example:

- `gtm-click-tracking`/`GtmClickTracking` has been renamed to `ga4-event-tracker`/`GA4EventTracker`
- `gtm-link-click-tracker`/`linkClickTracker` has been renamed to `ga4-link-tracker`/`linkTracker`

This will have a knock-on effect in applications where the above modules are still used as a data attribute. Therefore, these changes are breaking and will need to be fixed in applications where these modules (e.g. `gtm-click-tracking`) appear (such as `collections` and `frontend`).

In addition, the data attributes in the example component documentation have been updated to `custom_data_attr` to make the example more generic.

## Why
Prior to these changes, some of our analytics modules had similar names e.g. `gtm-click-tracking.js` and `gtm-link-click-tracker.js`. In order to improve the differentiation between these modules (and also to provide consistency in naming conventions e.g. using `tracker` instead of `tracking`), these changes have been made.

## Visual Changes
None.


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Expand print link component ([PR #2900](https://github.com/alphagov/govuk_publishing_components/pull/2900))
 * Add link click tracking ([PR #2904](https://github.com/alphagov/govuk_publishing_components/pull/2904))
 * Ensure tab clicks grab the tabs href for gtm ([PR #2884](https://github.com/alphagov/govuk_publishing_components/pull/2884))
+* **BREAKING:** Update gtm naming conventions ([PR #2906](https://github.com/alphagov/govuk_publishing_components/pull/2906))
```
andysellick, line 15: You're right, this is technically a breaking change, but as this code isn't live in production I'm not sure it warrants a breaking release - I don't want to unnecessarily alarm anyone or make them look through code to check that merging this new release into an app isn't going to break anything. What do you think?

andysellick (approved):
Good stuff, just one comment, but happy to approve ahead of resolving. Good documentation updates too.

CHANGELOG.md
JamesCGDS: Cool makes sense, I'll update the log and merge. Thanks Andy 👍

PR #2908, opened 2022-08-18
Allow accordion to accept custom data module

andysellick:
## What
Allow the accordion to be passed a custom value for `data-module`. The accordion already fills this in with its own module, so this must be preserved alongside whatever is passed.

## Why
Because we want to add tracking to accordions without having to wrap them in extra elements just to add the `data-module` attribute. Also because I already told @gclssvglx accordions can do this and it turns out they couldn't.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
All good :+1:

PR #2915, opened 2022-08-19
Integrate GA4 analytics code with cookie consent mechanism

andysellick:
## What
Integrate the GA4 code with the cookie consent mechanism. Currently GA4 just runs on page load, but limited only to integration.

Adds an init script to the `window.GOVUK.analyticsGA4` object, which when called (from `static`) will make the GA4 analytics code work. Currently, this does the following.

- checks for the consent cookie
- if found, sets `window.GOVUK.analyticsGA4.consent` to `true` for use later
- any JS in the `window.GOVUK.analyticsGA4.analyticsModules` object will have its `init` function called, if one exists
- this includes all of the GA4 code including pageviews (all restructured as part of this PR)

This means a pageview (and all the rest of the analytics code) should happen on the page where consent is given, without needing a page reload. This is important otherwise we'd miss out on a pageview for that first page of the users journey.

For any `GOVUK.Modules`, we add the [wait for consent](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/javascript-modules.md#modules-and-cookie-consent) mechanism. This is currently limited to the GA4 event tracker.

Still to do:

- stop relying on the GTM component for including the GTM JS from Google (will do in a separate PR as this PR has enough in it already)

## Why
This change will mean that only users who consent to cookies will have the analytics code included in their browser and executed.

## Visual Changes
None.

Trello card: https://trello.com/c/vrMj18et/353-integrate-gtm-with-cookie-consent-mechanism

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
@@ -11,6 +11,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   Ga4EventTracker.prototype.init = function () {
+    var consentCookie = window.GOVUK.getConsentCookie()
+
+    if (consentCookie && consentCookie.settings) {
+      this.startModule()
+    } else {
+      this.startModule = this.startModule.bind(this)
+      window.addEventListener('cookie-consent', this.startModule)
+    }
+  }
+
+  // triggered by cookie-consent event, which happens when users consent to cookies
+  Ga4EventTracker.prototype.startModule = function () {
     if (window.dataLayer) {
       this.module.addEventListener('click', this.trackClick.bind(this), true) // useCapture must be true
     }
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
@@ -11,68 +11,68 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   Ga4EventTracker.prototype.init = function () {
-    this.module.addEventListener('click', this.trackClick.bind(this), true) // useCapture must be true
+    if (window.dataLayer) {
+      this.module.addEventListener('click', this.trackClick.bind(this), true) // useCapture must be true
+    }
   }
 
   Ga4EventTracker.prototype.trackClick = function (event) {
-    if (window.dataLayer) {
-      var target = this.findTrackingAttributes(event.target)
-      if (target) {
-        var schema = new window.GOVUK.analyticsGA4.Schemas().eventSchema()
+    var target = this.findTrackingAttributes(event.target)
+    if (target) {
+      var schema = new window.GOVUK.analyticsGA4.Schemas().eventSchema()
 
-        try {
-          var data = target.getAttribute(this.trackingTrigger)
-          data = JSON.parse(data)
-        } catch (e) {
-          // if there's a problem with the config, don't start the tracker
-          console.error('GA4 configuration error: ' + e.message, window.location)
-          return
-        }
-
-        schema.event = 'event_data'
-        // get attributes from the data attribute to send to GA
-        // only allow it if it already exists in the schema
-        for (var property in data) {
-          if (property in schema.event_data) {
-            schema.event_data[property] = data[property]
-          }
-        }
-
-        // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
-        if (target.closest('.gem-c-accordion')) {
-          var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
-        }
-
-        /*
-          the details component uses an 'open' attribute instead of aria-expanded, so we need to check if we're on a details component.
-          since details deletes the 'open' attribute when closed, we need this boolean, otherwise every element which
-          doesn't contain an 'open' attr would be pushed to gtm as a closed element.
-        */
-        var detailsElement = target.closest('details')
-
-        if (ariaExpanded) {
-          schema.event_data.text = data.text || target.innerText
-          schema.event_data.action = (ariaExpanded === 'false') ? 'opened' : 'closed'
-        } else if (detailsElement) {
-          schema.event_data.text = data.text || detailsElement.textContent
-          var openAttribute = detailsElement.getAttribute('open')
-          schema.event_data.action = (openAttribute == null) ? 'opened' : 'closed'
-        }
-
-        /* If a tab was clicked, grab the href of the clicked tab (usually an anchor # link) */
-        var tabElement = event.target.closest('.gem-c-tabs')
-        if (tabElement) {
-          var aTag = event.target.closest('a')
-          if (aTag) {
-            var href = aTag.getAttribute('href')
-            if (href) {
-              schema.event_data.url = href
-            }
-          }
-        }
-
-        window.dataLayer.push(schema)
+      try {
+        var data = target.getAttribute(this.trackingTrigger)
+        data = JSON.parse(data)
+      } catch (e) {
+        // if there's a problem with the config, don't start the tracker
+        console.error('GA4 configuration error: ' + e.message, window.location)
+        return
       }
+
+      schema.event = 'event_data'
+      // get attributes from the data attribute to send to GA
+      // only allow it if it already exists in the schema
+      for (var property in data) {
+        if (property in schema.event_data) {
+          schema.event_data[property] = data[property]
+        }
+      }
+
+      // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
+      if (target.closest('.gem-c-accordion')) {
+        var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
+      }
+
+      /*
+        the details component uses an 'open' attribute instead of aria-expanded, so we need to check if we're on a details component.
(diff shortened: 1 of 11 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -1,113 +1,114 @@
...
+  PageViewTracker.getReferrer = function () {
+    return this.PIIRemover.stripPIIWithOverride(document.referrer, true, true)
+  }
+
+  PageViewTracker.getTitle = function () {
+    return this.PIIRemover.stripPII(document.title)
+  }
+
+  // window.httpStatusCode is set in the source of the error page in static
+  // https://github.com/alphagov/static/blob/main/app/views/root/_error_page.html.erb#L32
```
injms, line 66: Currently line 32 linked to here is `<div class="govuk-grid-row">`, which I don't think is what's wanted here. Could we link to a specific commit to avoid this kind of drift?

```suggestion
  // https://github.com/alphagov/static/blob/1c734451f2dd6fc0c7e80beccbdcbfa5aaffd0e4/app/views/root/_error_page.html.erb#L41-L43
```

andysellick: Good shout, have updated, thanks.

gclssvglx (approved):
I don't pretend to understand all of this, but from the logic - the overall approach looks sound and very sensible to me. Very happy to approve, but looking to @AshGDS and @JamesCGDS to give the code a through eyeball.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -0,0 +1,27 @@
+var initFunction = function () {
```
AshGDS, line 1: Should this have it's own:

 ```
window.GOVUK.analyticsGA4 = window.GOVUK.analyticsGA4 || {}
```
So that it doesn't depend on being imported after another file has declared `window.GOVUK.analyticsGA4`?

This would never happen in our analytics implementation, since we have the page views etc. being imported first which would declare what `window.GOVUK.analyticsGA4` is. But if someone wanted to use this cookie consent pattern with our modules only, I think it would crash as `window.GOVUK.analyticsGA4` wouldn't exist.

File: spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js [hunk 3]
```diff
@@ -177,99 +177,99 @@ describe('Google Tag Manager page view tracking', function () {
       content.setAttribute('lang', 'wakandan')
       expected.page_view.language = 'wakandan'
 
-      GOVUK.analyticsGA4.pageViewTracker.sendPageView()
+      GOVUK.analyticsGA4.analyticsModules.PageViewTracker.init()
       expect(window.dataLayer[0]).toEqual(expected)
     })
 
     it('set incorrectly', function () {
```
AshGDS, line 184: Should this be `sets` or `sets it`?

AshGDS (approved):
@andysellick looks good so I will approve - I've added some minor non-blocking questions

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      }
+      TestModule.prototype.startModule = function () {
+        testObject.testFunction()
+      }
+
+      testModule = new TestModule()
+      spyOn(testObject, 'testFunction')
+    })
+
+    it('do not run if consent is not given', function () {
```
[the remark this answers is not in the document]
andysellick: The full sentence (including the bit from the parent describe function) is 

`Initialising GA4 Modules depending on cookie consent to run do not run if consent is not given`

...which is a bit of a mouthful but I think grammatically correct, as it's plural.

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
```diff
@@ -0,0 +1,109 @@
...
+      GOVUK.setCookie('cookies_policy', '{"essential":false,"settings":false,"usage":false,"campaigns":false}')
+      GOVUK.analyticsGA4.init()
+
+      testModule.init()
+      expect(testObject.testFunction).not.toHaveBeenCalled()
+    })
+
+    it('run if consent is given', function () {
```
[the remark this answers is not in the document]
andysellick: See above comment, `runs` would refer to a singular thing, `run` for plural. I think.

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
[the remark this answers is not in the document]
AshGDS: Thanks @andysellick - I forgot that it combines it with the describe!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
andysellick: Yes, that makes sense, have added, thanks.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
andysellick: The full sentence would be `Google Tag Manager page view tracking returns a pageview with a language set incorrectly`, so I think that's correct. Although it doesn't read very well within that function, granted.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -1,171 +1,171 @@
...
-        this.internalLinksDomain = 'www.gov.uk/'
-        this.internalLinksDomainWithoutWww = 'gov.uk/'
-        this.handleClick = this.handleClick.bind(this)
-        this.handleMousedown = this.handleMousedown.bind(this)
-        document.querySelector('body').addEventListener('click', this.handleClick)
-        document.querySelector('body').addEventListener('contextmenu', this.handleClick)
-        document.querySelector('body').addEventListener('mousedown', this.handleMousedown)
-      }
-    },
+  var Ga4LinkTracker = function () {}
```
JamesCGDS, line 9: Would there be any mileage in turning this into an object rather than a function?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -1,113 +1,114 @@
...
-            first_published_at: this.stripTimeFrom(this.getMetaContent('first-published-at')),
-            updated_at: this.stripTimeFrom(this.getMetaContent('updated-at')),
-            public_updated_at: this.stripTimeFrom(this.getMetaContent('public-updated-at')),
-            publishing_government: this.getMetaContent('publishing-government'),
-            political_status: this.getMetaContent('political-status'),
-            primary_publishing_organisation: this.getMetaContent('primary-publishing-organisation'),
-            organisations: this.getMetaContent('analytics:organisations'),
-            world_locations: this.getMetaContent('analytics:world-locations')
-          }
+  var PageViewTracker = function () {}
```
JamesCGDS, line 8: Would there be any mileage in turning this into an object rather than a function?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
```diff
@@ -11,68 +11,78 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   Ga4EventTracker.prototype.init = function () {
-    this.module.addEventListener('click', this.trackClick.bind(this), true) // useCapture must be true
+    if (window.GOVUK.analyticsGA4.consent) {
```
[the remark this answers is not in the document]
andysellick: That's an excellent point. I'm not certain in which order our code executes, but this would fail in the following scenario.

- user consents to cookies on a previous page
- Modules code runs first, looks for cookie consent flag, finds false, does not execute analytics Module
- analytics code runs, sets consent flag to true based on previously set cookie

It does work if the analytics code runs first, but we can't guarantee that:

- user consents to cookies on a previous page
- analytics code runs, sets consent flag to true based on previously set cookie
- Modules code runs first, looks for cookie consent flag, finds true, executes

Fortunately this doesn't mean that tracking would happen without consent, but it does mean that when users consent we might miss out on tracking data.

This might be a bit of a thorny one to sort out right now, so I think the best option for the moment is to go back to the original model [already in use by some Modules](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/javascript-modules.md#modules-and-cookie-consent), which is to use `window.GOVUK.getConsentCookie()` to check, rather than relying on a flag set by the analytics code.

This solution is less efficient (more than one bit of code would read the cookie) but for now I think the best way forward. I'll record this problem in our technical notes for later investigation.

Thanks for raising.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: Yes, I definitely think so. I'd been worried that these look too similar to our existing `Modules`, which could cause confusion in future. Also if I restructure them as objects, it involves fewer changes to the existing code. Good shout, have updated.

JamesCGDS (approved):
Looks good!

PR #2916, opened 2022-08-22
Fix bugs with gtm external link tracking

AshGDS:
Hi @JamesCGDS - would you be able to look at this PR? Thanks :+1:

## What
- Changes `event: analytics` to `event: event_data` in the external link tracking GTM object (as requested by analysts.)
- Makes sure that anchor links (jump links), such as `#other-ways-to-apply` are not treated as external links.

## Why
- The `event` key was changed so that the data is properly tracking the the analytics dashboard.
- The anchor links should not be treated as external links, as they just jump to a location on the same page you're already on.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -61,7 +61,7 @@
         clickData.link_method = this.getClickType(event)
 
         var schema = new window.GOVUK.analyticsGA4.Schemas().eventSchema()
-        schema.event = 'analytics'
+        schema.event = 'event_data'
 
         // get attributes from the clickData object to send to GA
         // only allow it if it already exists in the schema
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -124,7 +124,7 @@
 
     isExternalLink: function (href) {
       var isInternalLink = this.hrefPointsToDomain(href, this.internalLinksDomain) || this.hrefPointsToDomain(href, this.internalLinksDomainWithoutWww)
-      if (!isInternalLink && !this.hrefIsRelative(href)) {
+      if (!isInternalLink && !this.hrefIsRelative(href) && !this.hrefIsAnchor(href)) {
         return true
       }
     },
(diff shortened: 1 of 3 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -51,6 +51,11 @@ describe('GOVUK.analyticsGA4.linkTracker', function () {
               '<a href="https://gov.uk/some-path">Another local link</a>' +
               '<a href="gov.uk/some-path">Another local link</a>' +
               '<a href="//gov.uk/some-path">Another local link</a>' +
+            '</div>' +
+            '<div class="internal-links">' +
```
JamesCGDS, line 55: Just to check: would this need to be `<div class="anchor-links">`?

JamesCGDS (approved):
All good 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -146,6 +146,10 @@
     hrefIsRelative: function (href) {
       // Checks that a link is relative, but is not a protocol relative url
       return href[0] === '/' && href[1] !== '/'
+    },
+
+    hrefIsAnchor: function (href) {
```
andysellick, line 151: Happy with the code as it is (and it's already been merged) but thought I'd raise this as a discussion point. Another approach here would be to invert this function so that it already returns the desired value, instead of returning the value then inverting it, i.e.

```JavaScript
hrefIsNotAnchor: function (href) {
  return href[0] !== '#'
}

if (!isInternalLink && !this.hrefIsRelative(href) && this.hrefIsNotAnchor(href)) {
  //
}
```

I think this would make the code slightly easier to read (and microscopically faster), but interested to know what everyone else thinks.

JamesCGDS: Personally, I find the original easier to read. If we do want to change this though, would we also want to change `!isInternalLink` and `hrefIsRelative` to `isNotInternalLink` and `hrefIsNotRelative` respectively? Otherwise the difference in the checks could potentially get confusing.

andysellick: Agreed (I hadn't checked whether those functions could be as easily inverted though).

PR #2919, opened 2022-08-24
Add example for passing data attributes to textarea

DilwoarH:
## What
This adds an example of how to pass data attributes to text areas.

## Why
The option is available but isn't documented.


injms (approved):
Good spot! Thank you 👍

PR #2920, opened 2022-08-24
Remove unused classes in intervention component

jon-kirwan:
## What
Replace unused classes in intervention component.

## Why
The component has selectors for `govuk-link gem-c-intervention__suggestion-link` and `govuk-link gem-c-intervention__dismiss` which are only used for targeting in tests but has no specific styling.


injms (approved):
🕵️ Nice catch and tidy up 👍

PR #2921, opened 2022-08-25
Add links and sections count tracking for Cost of living hub

AgaDufrat:
## What
Add links and sections count tracking for Cost of living hub

## Why
We need to add tracing to populate custom dimensions 26 (number of sections) and 27 (number of links) on the Cost of living hub pages as requested by a performance analyst.

`data-track-count="accordionSection"` is already added to the accordion sections https://github.com/alphagov/collections/blob/241b33a/app/views/cost_of_living_landing_page/show.html.erb#L22

`data-track-count="contentLink"` will be added to links within accordions and links within inset text on the page.

Meta tag `navigation-page-type` will be added to the Cost of living page.

[Trello](https://trello.com/c/0pjp3DgK/1288-add-meta-tags-that-power-custom-dimensions-m)

## Visual Changes
n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   of the commit log.
   
 ## Unreleased
+
+* Add links and sections count tracking for Cost of living hub ([PR #2921](https://github.com/alphagov/govuk_publishing_components/pull/2921))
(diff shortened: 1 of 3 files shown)


hannalaakso (approved):
I've tested this locally together with https://github.com/alphagov/collections/pull/2927 and the four dimensions are being set correctly. Also seems like a low risk change so happy to approve it 👍

PR #2922, opened 2022-08-30
Add `gtm.blocklist` values to GTM snippet

gclssvglx:
## What

Include the following with the GTM snippet, to add a block list to the dataLayer:

```
dataLayer = [
  { 'gtm.blocklist' : ['customPixels', 'customScripts', 'html', 'nonGoogleScripts'] },
]
```

## Why

To prevent anyone from adding custom code through GTM.

## Visual Changes

None

[Trello](https://trello.com/c/G2dyvbZR/97-add-blacklist-to-gtm-snippet)

### DIFF
--- a/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
+++ b/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
@@ -1,13 +1,19 @@
 <%
   gtm_auth ||= nil
   gtm_preview ||= nil
+  blocklist ||= nil # an array of the form ['customPixels', 'customScripts', 'html', 'nonGoogleScripts']
 
   gtm_attributes = %w(gtm_cookies_win=x)
     gtm_attributes << "gtm_auth=" + gtm_auth if gtm_auth
     gtm_attributes << "gtm_preview=" + gtm_preview if gtm_preview
   gtm_attributes = gtm_attributes.join('&')
 %>
+
 <script>
+<% if blocklist %>
+window.dataLayer = window.dataLayer || []
+window.dataLayer.push({ 'gtm.blocklist' : <%= blocklist %> })
+<% end %>
 var doNotTrack = ( navigator.doNotTrack === '1' || navigator.doNotTrack === 'yes' || navigator.msDoNotTrack === '1' || window.doNotTrack === '1' )
 if (!doNotTrack) {
   (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
--- a/app/views/govuk_publishing_components/components/docs/google_tag_manager_script.yml
+++ b/app/views/govuk_publishing_components/components/docs/google_tag_manager_script.yml
@@ -11,3 +11,7 @@ examples:
   default:
     data:
       gtm_id: GTM-XXXXXXX
+  with_blocklist:
+    data:
+      gtm_id: GTM-XXXXXXX
+      blocklist: "['customPixels', 'customScripts', 'html', 'nonGoogleScripts']"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## UNRELEASED
+
+* Add gtm.blocklist values to GTM snippet ([PR #2922](https://github.com/alphagov/govuk_publishing_components/pull/2922))
+
+
 ## 30.4.1
 
 * Revert addition of `awesome_print` gem ([PR #2943](https://github.com/alphagov/govuk_publishing_components/pull/2943))
--- a/spec/components/google_tag_manager_script_spec.rb
+++ b/spec/components/google_tag_manager_script_spec.rb
@@ -5,10 +5,22 @@ describe "Google tag manager script", type: :view do
     "google_tag_manager_script"
   end
 
-  it "renders a script tag" do
+  it "renders a script tag without a blocklist" do
     render_component(gtm_id: "GTM_ID")
 
     assert_select "script"
+    expect(rendered).not_to include "gtm.blocklist"
+    expect(rendered).to include "window,document,'script','dataLayer','GTM_ID'"
+  end
+
+  it "renders a script tag with a blocklist" do
+    render_component(
+      gtm_id: "GTM_ID",
+      blocklist: [],
+    )
+
+    assert_select "script"
+    expect(rendered).to include "gtm.blocklist"
     expect(rendered).to include "window,document,'script','dataLayer','GTM_ID'"
   end
 end


app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
```diff
@@ -8,6 +8,10 @@
   gtm_attributes = gtm_attributes.join('&')
 %>
 <script>
+dataLayer = [
+  { 'gtm.blacklist' : ['customPixels', 'customScripts', 'html', 'nonGoogleScripts'] },
```
andysellick, line 12: Just want to mention that this component is also used by `content-data-admin` and `content-publisher`, so we should check that this change won't cause those applications a problem.

gclssvglx: Thanks for alerting me - didn't realise. Yes, I'll include devs from those teams to review.

ChrisBAshton: Does this work with `blocklist` instead of `blacklist`? My quick search [suggests it should](https://developers.google.com/tag-platform/tag-manager/web/restrict), and it's a more inclusive term to use if possible.

kevindew: Yup, a dataLayer variable is already set in Content Publisher: https://github.com/alphagov/content-publisher/blob/4be9a6b0b2249ce7b513349777d96b684f5e8a73/app/views/layouts/application.html.erb#L7-L14 I presume there is something similar in Content Data.

This would cause a problem as I understand as it will override the previously set dataLayer variable and replace anything already set.

I imagine you've got a choice of a) making this a breaking change and set a new global dataLayer variable or b) checking whether a dataLayer variable exists, and creating it if not before appending to it. (e.g. `window.dataLayer = window.dataLayer || []`)

I don't know a huge amount about how things work but I'm wondering whether b) is the better option as I'm assuming if an app like Content Publisher adds metadata to datalayer after this script has run, it is too late and we've already sent information to tag manager that misses it?

---

minor nit, we're missing a var declaration here

gclssvglx: > Does this work with `blocklist` instead of `blacklist`? My quick search [suggests it should](https://developers.google.com/tag-platform/tag-manager/web/restrict), and it's a more inclusive term to use if possible.

Yes, quite right @ChrisBAshton - looks like it should. Updated throughout - thanks for spoting.

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
```diff
@@ -8,6 +8,10 @@
   gtm_attributes = gtm_attributes.join('&')
 %>
 <script>
+dataLayer = [
```
andysellick, line 11: Just want to check - should this be `window.dataLayer` or is this something different?

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
ChrisBAshton: Thanks Graham.
+1 Kevin - [Content Data Admin defines its own `dataLayer` variable too](https://github.com/alphagov/content-data-admin/blob/c1444b5d268a11079e80455c75bef0236308321f/app/views/layouts/application.html.erb#L7-L11). I assume it would be overwritten by the changes in this PR.

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
gclssvglx: To be honest I don't fully understand the distinction - they appear to be synonymous 🤷 

Here's some console output...

Screenshot 2022-09-01 at 08 56 23

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
andysellick: I think `window.dataLayer` and `var dataLayer` will produce two different things, so you might want to check on that. I know that if `window.dataLayer` exists and `var dataLayer` doesn't, calling `dataLayer` will produce `window.dataLayer`, but not the other way around.

I suspect the best way to do this PR is to add an option to the component to pass an optional blocklist value. The code might look something like this:

```erb
<%
  gtm_auth ||= nil
  gtm_preview ||= nil
  blocklist ||= nil # an array of the form ['customPixels', 'customScripts', 'html', 'nonGoogleScripts']

  gtm_attributes = %w(gtm_cookies_win=x)
    gtm_attributes << "gtm_auth=" + gtm_auth if gtm_auth
    gtm_attributes << "gtm_preview=" + gtm_preview if gtm_preview
  gtm_attributes = gtm_attributes.join('&')
%>
<script>
<% if blocklist %>
  window.dataLayer = window.dataLayer || []
  window.dataLayer.push({ 'gtm.blocklist' : <%= blocklist %> })
<% end %>
var doNotTrack = ( navigator.doNotTrack === '1' || navigator.doNotTrack === 'yes' || navigator.msDoNotTrack === '1' || window.doNotTrack === '1' )
if (!doNotTrack) {
  (function(w,d,s,l,i){w[l]=w[l]||[];w[l].push({'gtm.start':
  new Date().getTime(),event:'gtm.js'});var f=d.getElementsByTagName(s)[0],
  j=d.createElement(s),dl=l!='dataLayer'?'&l='+l:'';j.async=true;j.src=
  'https://www.googletagmanager.com/gtm.js?id='+i+dl+'&<%= raw gtm_attributes %>';f.parentNode.insertBefore(j,f);
})(window,document,'script','dataLayer','<%= gtm_id %>');
}
</script>
```

That way the apps currently using this will stick with the default configuration and not have your new blocklist included, and when you call it in `static` you can pass a blocklist to it.

You'll need to update the component `YML` file to demonstrate the example, and include a test of the erb template. I can help you with those if you want.

gclssvglx: I like this approach 👍 Thanks @andysellick - will update the code as suggested.

gclssvglx: PS Still unsure of the inner workings of `dataLayer`, but my [brief experiment](https://github.com/alphagov/govuk_publishing_components/pull/2922#discussion_r960337231) seems to indicate they are either the same or at least synonymous.

gclssvglx: Thanks @kevindew - great explanation

andysellick (approved):
Looks good. Can you add a CHANGELOG entry before merging please?

app/views/govuk_publishing_components/components/docs/google_tag_manager_script.yml
```diff
@@ -11,3 +11,7 @@ examples:
   default:
     data:
       gtm_id: GTM-XXXXXXX
+  with_blocklist:
+    data:
+      gtm_id: GTM-XXXXXXX
+      blocklist: "['customPixels', 'customScripts', 'html', 'nonGoogleScripts']"
```
kevindew, line 17: Should this be an array?

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
```diff
@@ -1,13 +1,19 @@
...
   gtm_attributes = %w(gtm_cookies_win=x)
     gtm_attributes << "gtm_auth=" + gtm_auth if gtm_auth
     gtm_attributes << "gtm_preview=" + gtm_preview if gtm_preview
   gtm_attributes = gtm_attributes.join('&')
 %>
+
 <script>
+<% if blocklist %>
+window.dataLayer = window.dataLayer || []
+window.dataLayer.push({ 'gtm.blocklist' : <%= blocklist %> })
```
kevindew, line 15: This seems to treat the variable as a string? whereas I'd expect it to be dealing with an array?

kevindew (commented):
Just spotted a couple of things that look a bit fishy.

The other thing I'm not clear of is why we're doing this change and not setting blocklist in the app that's using it? Are we expecting this code to be used in lots of apps and thus be repetitive? I had assumed it'd just be Static.

Also do we expect people to need to configure this? I wonder a little why we're setting this as an option - as I'd guess the reason to add this was for consistency but it being configurable means only the default is consistent.

andysellick:
@gclssvglx I've just realised that this change won't be needed soon anyway - as we're going to stop using this component for the GA4 work, and replace it with the Google snippet as part of our JavaScript (hoping to finish this week). Sorry!

app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
gclssvglx: Thanks @kevindew - I would expect that it depends on what `blocklist` is set to. The expectation is `# an array of the form ['customPixels', 'customScripts', 'html', 'nonGoogleScripts']` as per the comment. But without TypeScript...!

app/views/govuk_publishing_components/components/docs/google_tag_manager_script.yml
gclssvglx: Thanks @kevindew - this is the doc-block, which caused me formatting issues while trying to keep the linter-passing `['customPixels', 'customScripts', 'html', 'nonGoogleScripts']` in the example on-screen. So, yes is should be an array, the issue was trying to make it look like one in the example.

PR #2923, opened 2022-08-30
Modify GTM values for download links in response to analyst review

AshGDS:
Hi @andysellick,

Would you be able to merge this PR if everything looks OK? Thanks :+1:

## What

This branch updates the values sent to GTM when clicking our download links. This branch:

- Changes the `event_name` value to `file_download`.
- Changes the `type` value to `generic download`.
- Makes download links `external: true` instead of `false`.
- Adds the ability to detect download `/preview` links, and pushes them with the `type: preview` value. This uses regex, instead of a generic check for the presence of `/preview` in the string, or `/preview` being the last characters in the string. This should prevent a false positive if the file itself has the word `preview` in it - a valid download URL could contain `/preview-of-something.pdf` - and loose checking for `/preview` in this case wouldn't incorrectly class this as a preview link.

## Why

These changes were requested in the analyst review of this trello card: https://trello.com/c/vyvSCGWx/315-add-tracking-download-links

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -44,18 +44,20 @@
       }
 
       if (this.isMailToLink(href)) {
+        clickData.event_name = 'navigation'
         clickData.type = 'email'
         clickData.external = 'true'
       } else if (this.isDownloadLink(href)) {
-        clickData.type = 'download'
-        clickData.external = 'false'
+        clickData.event_name = 'file_download'
+        clickData.type = this.isPreviewLink(href) ? 'preview' : 'generic download'
+        clickData.external = 'true'
       } else if (this.isExternalLink(href)) {
+        clickData.event_name = 'navigation'
         clickData.type = 'generic link'
         clickData.external = 'true'
       }
 
       if (Object.keys(clickData).length > 0) {
-        clickData.event_name = 'navigation'
         clickData.text = element.textContent.trim()
         clickData.url = href
         clickData.link_method = this.getClickType(event)
@@ -127,6 +129,18 @@
       if (!isInternalLink && !this.hrefIsRelative(href) && !this.hrefIsAnchor(href)) {
         return true
       }
+    },
+
+    isPreviewLink: function (href) {
+      /*  Regex looks for:
+      1. The file extension period (the character '.')
+      2. any alphanumeric characters (so we can match any file type such as jpg, pdf, mp4.)
+      3. the presence of '/preview'.
+      For example, .csv/preview or .mp4/preview will be matched.
+      Regex is used over JS string methods as this should work with anchor links, query string parameters and files that may have 'preview' in their name.
+      */
+      var previewRegex = /\.\w+\/preview/i
+      return previewRegex.test(href)
     },
 
     hrefPointsToDomain: function (href, domain) {
--- a/docs/analytics-gtm/ga4-link-tracker.md
+++ b/docs/analytics-gtm/ga4-link-tracker.md
@@ -10,12 +10,15 @@ When initialised, the JS adds an event listener to the `<body>` of the page with
 
 When one of these listeners are fired, they check if the `event.target` is an `<a>` tag with a `href`. If so, it will then check if the link is:
 
-- A `download` link - This is either a href to `assets.publishing.service.gov.uk` or a href on the `www.gov.uk` domain pointing to the `/government/uploads/` path.
+- A `generic download` - This is either a href to `assets.publishing.service.gov.uk` or a href on the `www.gov.uk` domain pointing to the `/government/uploads/` path.
+- A `preview` link - This is either a href to `assets.publishing.service.gov.uk` or a href on the `www.gov.uk` domain pointing to the `/government/uploads/` path which has `/preview` in the URL after the file path.
 - A `mailto` link - This is an email href that starts with `mailto:`.
-- An `external` link - This is a href that is not on the `www.gov.uk` domain. 
+- An `generic link`  - This is a href that is not on the `www.gov.uk` domain.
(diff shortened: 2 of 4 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -299,8 +304,7 @@ describe('GOVUK.analyticsGA4.linkTracker', function () {
         var link = linksToTest[i]
         GOVUK.triggerEvent(link, 'click')
         expected.event_data.url = link.getAttribute('href')
-        expected.event_data.type = 'download'
-        expected.event_data.external = 'false'
```
andysellick, line 303: Why remove the check for the `external` parameter? Is it enough that `type` is set to the correct value? I notice this has been changed for the other tests as well.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -419,6 +424,20 @@ describe('GOVUK.analyticsGA4.linkTracker', function () {
         expect(window.dataLayer[0]).toEqual(expected)
       }
     })
+
+    it('detects files with preview in their filename as download links instead of preview links', function () {
```
andysellick, line 428: Should this commit `Add extra test to ensure preview links are being detected properly` be merged with the first commit? Unclear why it's separate, can you explain?

docs/analytics-gtm/ga4-link-tracker.md
```diff
@@ -10,13 +10,16 @@ When initialised, the JS adds an event listener to the `<body>` of the page with
...
 - A `mailto` link - This is an email href that starts with `mailto:`.
-- An `external` link - This is a href that is not on the `www.gov.uk` domain. 
+- An `generic link`  - This is a href that is not on the `www.gov.uk` domain.
     - To calculate this, it checks if the href starts with `http://www.gov.uk/`, `https://www.gov.uk/`, `//www.gov.uk/`, `www.gov.uk/`, `http://gov.uk/`, `https://gov.uk/`, `//gov.uk`, or `gov.uk/` (NB: `gov.uk/` and `www./` are technically relative links as a href must contain a `/` or http method at the start to be treated as non-relative).
-    - It also checks if a `href` starts with `/` and does not start with `//` to determine if it's a relative link like `/bank-holidays` , but not a protocol relative link `//google.com`
+    - It also checks if a `href` starts with `/` and does not start with `//` to determine if it's a relative link like `/bank-holidays` , but not a protocol relative link such as `//google.com`.
     - If the `href` does not meet both of these criteria, then it is considered an external link.
 
+Event's can either be a `navigation` or a `file_download`
```
andysellick: Nitpick: `Events`.

docs/analytics-gtm/ga4-link-tracker.md
```diff
@@ -18,7 +18,7 @@ When one of these listeners are fired, they check if the `event.target` is an `<
     - It also checks if a `href` starts with `/` and does not start with `//` to determine if it's a relative link like `/bank-holidays` , but not a protocol relative link such as `//google.com`.
     - If the `href` does not meet both of these criteria, then it is considered an external link.
 
-Event's can either be a `navigation` or a `file_download`
+Event's can either have an `event_name` of `navigation` or a `file_download`. Download and preview links will use the `file_download` value, while generic external links and mailto links will use `navigation`.
```
andysellick, line 21: Nitpick: remove unnecessary `a` i.e. `Events can either have an... or file_download`.

docs/analytics-gtm/ga4-link-tracker.md
```diff
@@ -25,10 +28,10 @@ In the example above, on a left click of the link, the following would be pushed
 
 ```
 {
-    "event": "analytics",
+    "event": "event_data",
     "event_data": {
```
andysellick, line 32: Probably out of scope of the code change, but it seems odd that the value of `event` and another key are both `event_data`. Is this correct?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -47,8 +47,9 @@
         clickData.type = 'email'
         clickData.external = 'true'
       } else if (this.isDownloadLink(href)) {
-        clickData.type = 'download'
-        clickData.external = 'false'
+        clickData.event_name = 'file_download'
```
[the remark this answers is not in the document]
AshGDS: Good spot @andysellick - I had accidentally made the `type` in the tests `file_download` in the `beforeEach` function. Instead it should be in the `event_name`, so the tests were passing as the `beforeEach` `event_name` was still `navigation`, and the incorrect `type: 'file_download' I set was being overridden in each individual test.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: These tests will always have the same external value (the performance analyst has said our assets.publishing.service download links are external), so it has been defined at the beforeEach level once, instead of individually in each test. I think everything in this code is now external: 'true', so external: 'false' will only be useful once we implement internal link tracking. I'll make sure external is only defined in the beforeEach now.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: Yep, they should be one commit - I've never merged a commit before which is why they're separate (but James is going to help me with this)

docs/analytics-gtm/ga4-link-tracker.md
AshGDS: I think it's correct as it was requested by our Performance Analyst - I'm guessing there's a good reason for this (maybe they need the event key in order to determine which object they need to grab and process, since they never know what type of object is being pushed to the dataLayer?)

AshGDS:
Thanks @andysellick - I've made the requested changes, and merged the commits down to a single commit. :+1:

andysellick (approved):
Looks good, good job.

PR #2934, opened 2022-08-31
Change the way we generate pretty print data for our documentations

DilwoarH:
## What
This change adds the awesome_print package to pretty print the rails code. This is a more stable solution and allows us to use things like `:or` keys.

## Why
We found that the pretty_data function was changing `:or` keys to `"or"` strings. This caused copy and pasting of the code to fail as "or" would cause the component to throw an error.

## Visual Changes

https://components-gem-pr-2934.herokuapp.com/public

### Before

Screenshot 2022-08-31 at 3 00 04 pm

### After

Screenshot 2022-08-31 at 3 00 27 pm

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Modify GTM values for download links in response to analyst review ([PR #2923](https://github.com/alphagov/govuk_publishing_components/pull/2923/))
 * Add Cost of Living hub links to side wide navbar and footer ([PR #2939](https://github.com/alphagov/govuk_publishing_components/pull/2939))
+* Change the way we generate pretty print data for our documentations ([PR #2934](https://github.com/alphagov/govuk_publishing_components/pull/2934))
 
 ## 30.3.0
 
--- a/govuk_publishing_components.gemspec
+++ b/govuk_publishing_components.gemspec
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rouge"
   s.add_dependency "sprockets", ">= 3"
 
+  s.add_development_dependency "awesome_print"
   s.add_development_dependency "capybara"
   s.add_development_dependency "faker"
   s.add_development_dependency "gds-api-adapters"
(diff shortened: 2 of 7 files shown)


DilwoarH:
Example of this working: https://components-gem-pr-2934.herokuapp.com/component-guide/checkboxes/checkbox_with_exclusive_item

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change the way we generate pretty print data for our documentations ([PR #2934](https://github.com/alphagov/govuk_publishing_components/pull/2934))
```
andysellick, line 12: Nitpick: can this match the PR title? (or vice versa)

andysellick (approved):
One tiny thing to address, but otherwise good stuff.

CHANGELOG.md
DilwoarH: done

govuk_publishing_components.gemspec
```diff
@@ -24,6 +24,7 @@ Gem::Specification.new do |s|
   s.add_dependency "rouge"
   s.add_dependency "sprockets", ">= 3"
 
+  s.add_development_dependency "awesome_print"
```
kevindew, line 27: @DilwoarH This dependency is used outside development, as component guide is also used in apps that run this gem (whereas development dependencies are only installed when you're debugging the gem).

This is why Static had problems with this - and any other apps would (unless they already had the `awesome_print` gem installed. 

Rather than moving this to be an app dependency (and adding the overhead that all the apps then need to manage it). I'd consider

a) Do we need components to be fussy about whether things are a symbol or a string? it's quite precise - but maybe the horse has bolted
b) Lets see if we can use `pp` instead which is part of Ruby Standard Library, so wouldn't need managing by applications (which seems a lot for one line of code)

DilwoarH: Thanks for this @kevindew, this has now been reverted.

Regarding your points.

a) I think this is probably the best way to move forward. The change to the documentation was to display the documentation as expected by the author.
b) This is the one I initially went with but it is very limited and styling standards didn't meet the expected coding styles we had in place already.

I will speak with the rest of the community and look at the solutions suggested.

PR #2936, opened 2022-09-01
Adds content list on components guide

DilwoarH:
## What
Adds content list on components guide

## Why
This will allow users to jump to the example that they need without having to scroll through the whole page.


andysellick (approved):
Nice feature. Worth adding to the changelog? Maybe not, I'm not sure.

DilwoarH:
> Nice feature. Worth adding to the changelog? Maybe not, I'm not sure.

I was thinking about it but decided against it because it's not a change to the component's gem.

injms (approved):
Super!

PR #2938, opened 2022-09-01
Add all headings to content list

DilwoarH:
## What
This change adds all headings in the page as part of the content list

## Why
This allows users to have a whole page structural overview rather than just a list of components.

### Before

Screenshot 2022-09-01 at 4 16 07 pm

### After

Screenshot 2022-09-01 at 4 15 43 pm


injms (approved):
Super - thank you!

PR #2948, opened 2022-09-08
Refactor ga4-link-tracker and add new features

AshGDS:
## What

1. Allows shift clicks to be tracked, using `event.shiftKey`.
2. Allows share and follow link clicks to be tracked, using a `data-ga4-link` attribute on `_share_links.html.erb`.
3. Tracks internal download paths (e.g. `/government/uploads`) as internal links instead of external at the request of the analysts.
4. Adds the current hostname as an internal domain when tracking links.
5. Removes hard coding of www.gov.uk as an internal domain.
6. Removes hard coding of assets.publishing.gov.uk as a download domain.
7. Removes hard coding of /government/uploads/ as a download path.
8. Allows the developer to specify arrays for the internal domains to detect, download domains to detect, and internal domain download paths to check. This makes the code more suitable for domains that aren't www.gov.uk.

## Why

1. Shift clicks were added at the request of the analysts.
2. Share and follow link click tracking was added at the request of the analysts.
3. Tracking internal download paths (e.g. `/government/uploads`) as internal links instead of external was made at the request of the analysts.
4. Adds the current hostname as an internal domain when tracking links. This allows this code to be used on different domains - as previously the internal domain was hard coded as `www.gov.uk`
5. Removes hard coding of www.gov.uk as an internal domain to make the code more generic.
6. Removes hard coding of assets.publishing.gov.uk as a download domain to make the code more generic.
7. Removes hard coding of /government/uploads/ as a download path to make the code more generic.
8. When running `trackLinkClicks()`, it allows the developer to specify arrays which internal domains to detect, download domains to detect, and download paths to check. This makes the code more suitable for running on domains that aren't www.gov.uk. The code can be used on different domains, without any hardcoded `www.gov.uk` specific values in the code. 
It can now be decided what domains in hrefs are classed as internal domains - for example `helpforhouseholds.campaign.gov.uk` could be classed as an internal domain now. It can now be decided what domains are classed as a 'dedicated download' domains (a domain name which will always be a file download, such as `assets.publishing.service.gov.uk` or `dropbox.com`). It can now be decided what URL paths are classed as internal download paths, such as `/government/uploads` or `/government/another-new-download-path`. 

## Visual changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4.js
@@ -6,4 +6,4 @@
 //= require ./analytics-ga4/ga4-event-tracker
 
 window.GOVUK.analyticsGA4.pageViewTracker.sendPageView() // this will need integrating with cookie consent before production
-window.GOVUK.analyticsGA4.linkTracker.trackLinkClicks()
+window.GOVUK.analyticsGA4.linkTracker.trackLinkClicks({ internalDownloadPaths: ['/government/uploads/'], dedicatedDownloadDomains: ['assets.publishing.service.gov.uk'] })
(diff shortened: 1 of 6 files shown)


AshGDS:
Closed as it was quite far behind a large refactor that was merged to main, and was simpler to add changes to the newer version of main.

PR #2949, opened 2022-09-09
Fix underline on organisation logo component

andysellick:
## What
Fix the underline on the organisation logo component.

## Why
This appears to be being caused by a [bug in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1359501&q=underline&can=2) related to the use of `position: relative` and `top: 3px` - the underline on the element is not adjusted, meaning that the text is pushed down 3px, slightly overlapping the underline (if you increase the value for `top` you can make the underline eventually appear above the text).

The bug has only been recently reported but we don't really need to have this CSS on this element - the same effect can be achieved by increasing the padding above the element by 3px.

## Visual Changes
Tested in various Chromes and Chrome on mobile.

Before | After
------ | ------
Screenshot 2022-09-09 at 09 36 32 | Screenshot 2022-09-09 at 09 36 37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix underline on organisation logo component ([PR #2949](https://github.com/alphagov/govuk_publishing_components/pull/2949))
 * Integrate GA4 analytics code with cookie consent mechanism ([PR #2915](https://github.com/alphagov/govuk_publishing_components/pull/2915))
 
 ## 30.4.1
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Looks good to me, nice work!,

NickColley (commented):
Non-blocking: Is it possible to avoid the use of relative positioning to solve this, thereby simplifying the CSS itself?

andysellick:
> Non-blocking: Is it possible to avoid the use of relative positioning to solve this, thereby simplifying the CSS itself?

Sorry @NickColley, can you clarify? I've removed the relative positioning.

chao-xian (approved):

NickColley:
Ah yeah, I missed that in the diff, nice fix. sorry for the confusion

PR #2955, opened 2022-09-12
Add ecommerce tracking

JamesCGDS:
## What

This PR adds enhanced ecommerce tracking scripts that will gather data on pages such as search and push this data to the `dataLayer` where it can be accessed by Performance Analysts.

Caveat: on lines `90` and `91`, `internalLinksDomain` and `internalLinksDomainWithoutWww` are set directly because otherwise `isExternalLink` will return an error. This implementation is temporary and will be fixed following a refactor of the `ga4-link-tracker` code by @AshGDS (https://github.com/alphagov/govuk_publishing_components/pull/2961).

## Why

These changes are being made as part of our migration to GA4.

## Visual Changes

N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * GA4 analytics add single dataLayer push function ([PR #2960](https://github.com/alphagov/govuk_publishing_components/pull/2960))
 * Remove unused classes in intervention component ([PR #2920](https://github.com/alphagov/govuk_publishing_components/pull/2920))
 * Include size attributes on image card component ([PR #2895](https://github.com/alphagov/govuk_publishing_components/pull/2895))
+* Add ecommerce tracking ([PR #2955](https://github.com/alphagov/govuk_publishing_components/pull/2955))
 
 ## 30.5.2
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
@@ -1,0 +1,141 @@
+// = require govuk/vendor/polyfills/Element/prototype/closest.js
+;(function (global) {
+  'use strict'
+
+  var GOVUK = global.GOVUK || {}
+  GOVUK.analyticsGA4 = GOVUK.analyticsGA4 || {}
+
+  GOVUK.analyticsGA4.Ga4EnhancedEcommerceTracker = {
+    PIIRemover: new GOVUK.analyticsGA4.PIIRemover(),
+    DEFAULT_LIST_TITLE: 'Site search results',
+
+    init: function (isNewPageLoad = true) {
+      if (window.dataLayer) {
+        this.searchResultsBlocks = document.querySelectorAll('[data-ga4-ecommerce]')
+        this.isNewPageLoad = isNewPageLoad
+
+        if (!this.searchResultsBlocks.length === 0) {
+          return
+        }
+
+        /* If the results are updated by JS, the URL of the page will change and this needs to be visible to PA's,
+        hence the pageView object push to the dataLayer. We do not need to send a pageView object on page load as
+        this is handled elsewhere. */
+        if (!this.isNewPageLoad) {
+          var pageViewTracker = window.GOVUK.analyticsGA4.analyticsModules.PageViewTracker
+
+          if (pageViewTracker) {
+            pageViewTracker.init()
+          }
+        }
+
+        for (var i = 0; i < this.searchResultsBlocks.length; i++) {
+          this.trackSearchResults(this.searchResultsBlocks[i])
+
+          if (this.isNewPageLoad) {
+            this.searchResultsBlocks[i].addEventListener('click', this.handleClick.bind(this))
+          }
+        }
+      }
+    },
+
+    trackSearchResults: function (searchResultsBlock) {
+      var schema = this.populateEcommerceSchema(searchResultsBlock)
+
+      this.clearPreviousEcommerceObject()
+      window.dataLayer.push(schema)
+    },
+
+    handleClick: function (event) {
+      var searchResultsBlock = event.target.closest('[data-ga4-ecommerce]')
+      var isSearchResult = event.target.getAttribute('data-ecommerce-path')
+
+      if (isSearchResult) {
+        var searchResult = event.target
+        var schema = this.populateEcommerceSchema(searchResultsBlock, true, searchResult)
+
+        this.clearPreviousEcommerceObject()
+        window.dataLayer.push(schema)
+      }
+    },
+
+    populateEcommerceSchema: function (searchResultsBlock, searchResultClicked = false, searchResult = null) {
+      // Limiting to 100 characters to avoid noise from extra long search queries and to stop the size of the payload going over 8k limit.
+      var searchQuery = this.PIIRemover.stripPII(searchResultsBlock.getAttribute('data-search-query')).substring(0, 100).toLowerCase()
+      var variant = searchResultsBlock.getAttribute('data-ecommerce-variant') || undefined
+      var ecommerceRows = searchResultsBlock.querySelectorAll('[data-ecommerce-row]')
+      var listTitle = searchResultsBlock.getAttribute('data-list-title') || this.DEFAULT_LIST_TITLE
+      var startPosition = parseInt(searchResultsBlock.getAttribute('data-ecommerce-start-index'), 10)
+
+      var ecommerceObject = {
+        event: 'search_results',
+        search_results: {
+          event_name: searchResultClicked && searchResult ? 'select_item' : 'view_item_list',
+          term: searchQuery,
+          sort: variant,
+          results: this.getResultsCount(searchResultsBlock),
+          ecommerce: {
+            items: []
+          }
+        }
+      }
+
+      // Populate items array
+      if (searchResultClicked && searchResult) {
+        ecommerceObject.search_results.ecommerce.items.push({
+          item_id: searchResult.getAttribute('data-ecommerce-path'),
+          item_name: searchResult.textContent,
+          item_list_name: listTitle,
+          index: this.getIndex(searchResult, startPosition)
+        })
+
+        ecommerceObject.event_data = {
+          external: GOVUK.analyticsGA4.analyticsModules.Ga4LinkTracker.isExternalLink(searchResult.getAttribute('data-ecommerce-path')) ? 'true' : 'false'
+        }
+      } else {
+        for (var i = 0; i < ecommerceRows.length; i++) {
+          var ecommerceRow = ecommerceRows[i]
+          var path = ecommerceRow.getAttribute('data-ecommerce-path')
+
+          /* If the element does not have a data-ecommerce-index attribute, we set one so that we can use it later when setting the index property
+          on the ecommerce object. This for loop will always run on page load and so data-ecommerce-index will be available when we use it in the
+          initial if block above that checks if a search result has been clicked. */
+          if (!ecommerceRow.getAttribute('data-ecommerce-index')) {
+            ecommerceRow.setAttribute('data-ecommerce-index', i + 1)
+          }
+
+          ecommerceObject.search_results.ecommerce.items.push({
+            item_id: path,
+            item_list_name: listTitle,
+            index: this.getIndex(ecommerceRow, startPosition)
+          })
+        }
+      }
+
+      return ecommerceObject
+    },
+
+    getIndex: function (element, startPosition) {
+      return parseInt(element.getAttribute('data-ecommerce-index')) + startPosition - 1
+    },
+
+    clearPreviousEcommerceObject: function () {
+      window.dataLayer.push({ search_results: { ecommerce: null } })
+    },
+
+    getResultsCount: function (searchResultsBlock) {
+      // This returns a string e.g. '12,345 results'. Therefore to extract the number, we need to remove the comma and split the string at the space
+      var resultsCount = searchResultsBlock.querySelector('#js-result-count')
+
+      if (!resultsCount) {
+        return null
+      }
+
+      resultsCount = resultsCount.textContent.replace(',', '')
+      resultsCount = resultsCount.split(' ')[0]
+      return parseInt(resultsCount)
+    }
+  }
+
+  global.GOVUK = GOVUK
+})(window)
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
@@ -1,0 +1,296 @@
+/* eslint-env jasmine */
+
+describe('Google Analytics ecommerce tracking', function () {
+  var GOVUK = window.GOVUK
+  var searchResultsParentEl
+  var searchResults
+  var searchResultsHeading
+  var resultToBeClicked
+  var onPageLoadExpected
+  var onSearchResultClickExpected
+  var pageViewExpected
+
+  function agreeToCookies () {
+    GOVUK.setCookie('cookies_policy', '{"essential":true,"settings":true,"usage":true,"campaigns":true}')
+  }
+
+  beforeEach(function () {
+    searchResultsHeading = document.createElement('h2')
+    searchResultsHeading.id = 'js-result-count'
+    searchResultsHeading.innerHTML = '12,345 results'
+
+    searchResultsParentEl = document.createElement('div')
+    searchResultsParentEl.setAttribute('data-ga4-ecommerce', '')
+    searchResultsParentEl.setAttribute('data-search-query', 'test-search-query')
+    searchResultsParentEl.setAttribute('data-module', 'ga4-enhanced-ecommerce-tracker')
+    searchResultsParentEl.setAttribute('data-ecommerce-variant', 'test-ecommerce-variant')
+    searchResultsParentEl.setAttribute('data-list-title', 'test-list-title')
+    searchResultsParentEl.setAttribute('data-ecommerce-start-index', '1')
+
+    searchResults = document.createElement('div')
+    searchResults.innerHTML =
+      '<a data-ecommerce-path="/coronavirus" data-ecommerce-row="1" data-ecommerce-index="1" href="/coronavirus">Coronavirus (COVID-19): guidance and support</a>' +
+
+      '<a data-ecommerce-path="/guidance/travel-abroad-from-england-during-coronavirus-covid-19" data-ecommerce-row="1" data-ecommerce-index="2" href="/guidance/travel-abroad">Travel abroad</a>' +
+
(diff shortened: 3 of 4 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 30.5.3
```
AshGDS, line 10: This needs to be moved back to `## Unreleased` (I think a rebase causes this to happen as your change stays on the same line, but the heading changes)

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+      var schema = this.populateEcommerceSchema(true, searchResult)
+
+      this.clearPreviousEcommerceObject()
+      window.dataLayer.push(schema)
+    }
+  }
+
+  Ga4EnhancedEcommerceTracker.prototype.populateEcommerceSchema = function (searchResultClicked = false, searchResult = null) {
+    // Limiting to 100 characters to avoid noise from extra longs search queries
+    // and to stop the size of the payload going over 8k limit.
```
AshGDS, line 68: Is this is a comment from UA, if so I think it should be removed (the 8k limit part)

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+    var listTitle = this.module.getAttribute('data-list-title') || this.DEFAULT_LIST_TITLE
+    var startPosition = parseInt(this.module.getAttribute('data-ecommerce-start-index'), 10)
+    var ecommerceObject = {}
+
+    ecommerceObject.event = 'search_results'
+    ecommerceObject.search_results = {}
+    ecommerceObject.search_results.event_name = searchResultClicked && searchResult ? 'select_item' : 'view_item_list'
+    ecommerceObject.search_results.term = searchQuery
+    ecommerceObject.search_results.sort = variant
+    ecommerceObject.search_results.results = parseInt(this.getMetaTagContent('govuk:search-result-count'))
```
AshGDS, line 81: Is it OK that if the meta tag doesn't exist, `parseInt(null)` would return `NaN` and push that to the dataLayer? Should it fallback to `ecommerceRows.length` here if the meta tag doesn't exist?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
```diff
@@ -0,0 +1,295 @@
...
+    searchResultsParentEl = document.createElement('div')
+    searchResultsParentEl.setAttribute('data-search-query', 'test-search-query')
+    searchResultsParentEl.setAttribute('data-module', 'ga4-enhanced-ecommerce-tracker')
+    searchResultsParentEl.setAttribute('data-ecommerce-variant', 'test-ecommerce-variant')
+    searchResultsParentEl.setAttribute('data-list-title', 'test-list-title')
+    searchResultsParentEl.setAttribute('data-ecommerce-start-index', '1')
+
+    searchResults = document.createElement('div')
+    searchResults.innerHTML =
+      '<a data-ecommerce-path="/coronavirus" data-ecommerce-row="1" data-ecommerce-index="1" data-track-category="navFinderLinkClicked" data-track-action="Search.1" data-track-label="/coronavirus" data-track-options="{&quot;dimension28&quot;:20,&quot;dimension29&quot;:&quot;Coronavirus (COVID-19): guidance and support&quot;}" class="gem-c-document-list__item-title   govuk-link" href="/coronavirus">Coronavirus (COVID-19): guidance and support</a>' +
```
AshGDS, line 31: Might be worth removing the data values and the css class which aren't needed for these elements, so it's easier to read

AshGDS (commented):
Thanks @JamesCGDS - I've added some comments :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  'use strict'
+
+  function Ga4EnhancedEcommerceTracker (module) {
+    this.module = module
+  }
+
+  Ga4EnhancedEcommerceTracker.prototype.init = function (isNewPage = true) {
```
andysellick, line 11: We don't normally pass parameters to the `init` function of a module like this, normally because Modules are called and initialised by the start Modules script. Instead, normal practice is to have another data attribute on the element in question and read it when the constructor is called, or in the init function. For example:

```HTML
<div data-module="Ga4EnhancedEcommerceTracker" data-is-new-page="false"></div>
```

```JavaScript
  function Ga4EnhancedEcommerceTracker (module) {
    this.module = module
    this.isNewPage = this.module.getAttribute('data-is-new-page') || true
  }
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+      this.startModule = this.startModule.bind(this)
+      window.addEventListener('cookie-consent', this.startModule)
+    }
+  }
+
+  Ga4EnhancedEcommerceTracker.prototype.startModule = function () {
+    this.PIIRemover = new window.GOVUK.analyticsGA4.PIIRemover()
+    this.DEFAULT_LIST_TITLE = 'Site search results'
+
+    window.dataLayer = window.dataLayer || []
```
andysellick, line 27: I think we should try to avoid initialising the dataLayer inside our code. Instead, I think we should be checking to see whether it exists as an early step, and backing out if it doesn't (on the assumption that GTM has failed to initialise for some reason, perhaps because cookie consent hasn't been given).

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+    if (this.isNewPage) {
+      this.module.addEventListener('click', this.handleClick.bind(this))
+    } else {
+      if (window.GOVUK.analyticsGA4.analyticsModules.PageViewTracker) {
+        var pageViewTracker = window.GOVUK.analyticsGA4.analyticsModules.PageViewTracker
+        window.dataLayer.push({
+          event: 'page_view',
+          page_view: {
+            location: pageViewTracker.getLocation(),
+            referrer: pageViewTracker.getReferrer()
```
andysellick, line 38: I'm a bit confused by this section of code - it looks like you're getting the page view code and using part of it to create a new page view in here? I feel like this could increase maintenance if we decide to change the data structure of page views. Instead, could you call the page view code directly?

```JavaScript
var pageViewTracker = window.GOVUK.analyticsGA4.analyticsModules.PageViewTracker
if (pageViewTracker) {
  pageViewTracker.init()
}
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+    var ecommerceObject = {}
+
+    ecommerceObject.event = 'search_results'
+    ecommerceObject.search_results = {}
+    ecommerceObject.search_results.event_name = searchResultClicked && searchResult ? 'select_item' : 'view_item_list'
+    ecommerceObject.search_results.term = searchQuery
+    ecommerceObject.search_results.sort = variant
+    ecommerceObject.search_results.results = parseInt(this.getMetaTagContent('govuk:search-result-count'))
+    ecommerceObject.search_results.ecommerce = {}
+    ecommerceObject.search_results.ecommerce.items = []
```
andysellick, line 83: Could this block be written out as an object instead of individual declarations? If it's possible I think it would help to make the structure of the object more plain.

```JavaScript
var ecommerceObject = {
  event: 'search_results',
  search_results: {
    event_name: searchResultClicked && searchResult ? 'select_item' : 'view_item_list',
    // etc.
  }
}
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+
+    if (searchResultClicked && searchResult) {
+      ecommerceObject.search_results.ecommerce.items.push({
+        item_id: searchResult.getAttribute('data-ecommerce-path'),
+        item_name: searchResult.textContent,
+        item_list_name: listTitle,
+        index: this.getIndex(searchResult, Array.prototype.indexOf.call(ecommerceRows, searchResult)) + startPosition
```
andysellick, line 90: Can you explain what this line does?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+
+  Ga4EnhancedEcommerceTracker.prototype.getMetaTagContent = function (metaTagName) {
+    var element = document.querySelector(`meta[name='${metaTagName}']`)
+    if (!element) {
+      return null;    
+    }
+    return element.getAttribute('content');
+  }
+
+  Modules.Ga4EnhancedEcommerceTracker = Ga4EnhancedEcommerceTracker
```
andysellick, line 132: Since this gets referred to in a few places, would it be cleaner to call this something shorter? Maybe just `Ga4EcommerceTracker`?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
```diff
@@ -0,0 +1,295 @@
+/* eslint-env jasmine */
+
+describe('Google Analytics ecommerce tracking', function () {
```
andysellick, line 3: I'd consider merging the commit that adds this test with the commit that creates the code it refers to, since they're linked.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -40,7 +40,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         })
       }
     }
-    
+
```
andysellick, line 43: Could these linting fixes be merged with the commit that added these files?

andysellick (commented):
Looks generally good, although it's a little hard to follow in places and I'm not sure it should be structured as one of our `GOVUK.Modules`. Let's have a chat and see what solution would be best.

CHANGELOG.md
JamesCGDS: Done.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: I think the comment should stay because I think it does relate to GA. Good spot re the typo, I've fixed that now.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: We can't use `ecommerceRows.length` because that doesn't return the total result count. As per our discussion, I've changed this function to get the total results from the `h2`.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
JamesCGDS: Agreed, I've removed those.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,133 @@
...
+          }
+        })
+      }
+    }
+
+    this.trackSearchResults()
+  }
+
+  Ga4EnhancedEcommerceTracker.prototype.trackSearchResults = function () {
+    var schema = this.populateEcommerceSchema(this.module)
```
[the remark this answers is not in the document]
JamesCGDS: Good spot!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: Agreed, I've changed this to just be `if(window.dataLayer){ ... }`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: As per our discussion earlier, I've implemented your above suggestion with a view to refactor the `PageViewTracker` code at a later date to accommodate scenarios where we only need to push a few select values of the `pageView` object to the `dataLayer` rather than the whole thing.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: Agreed, this is much easier to read.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: As per our discussion earlier, this code has been refactored to be more readable following a suggestion from @AshGDS.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: Agreed. As per our discussion earlier, this will be addressed as part of a wider task to look at naming conventions.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
JamesCGDS: Done.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: Done.

JamesCGDS:
Thank you @andysellick and @AshGDS for your comments and input, it's much appreciated 👍  I've addressed those comments now, if you have any further feedback just give me a shout!

AshGDS:
Looks good to me but I will leave it for Andy to approve :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
```diff
@@ -0,0 +1,144 @@
...
+      if (searchResultClicked && searchResult) {
+        ecommerceObject.search_results.ecommerce.items.push({
+          item_id: searchResult.getAttribute('data-ecommerce-path'),
+          item_name: searchResult.textContent,
+          item_list_name: listTitle,
+          index: this.getIndex(searchResult, startPosition)
+        })
+
+        GOVUK.analyticsGA4.analyticsModules.Ga4LinkTracker.internalLinksDomain = 'www.gov.uk/'
+        GOVUK.analyticsGA4.analyticsModules.Ga4LinkTracker.internalLinksDomainWithoutWww = 'gov.uk/'
```
andysellick, line 93: Hopefully a quick one. Can you recap why these need to be set? And will it have any detrimental effect on the rest of the code once these two lines have run?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
```diff
@@ -0,0 +1,296 @@
...
+    searchResultsParentEl.setAttribute('data-module', 'ga4-enhanced-ecommerce-tracker')
+    searchResultsParentEl.setAttribute('data-ecommerce-variant', 'test-ecommerce-variant')
+    searchResultsParentEl.setAttribute('data-list-title', 'test-list-title')
+    searchResultsParentEl.setAttribute('data-ecommerce-start-index', '1')
+
+    searchResults = document.createElement('div')
+    searchResults.innerHTML =
+      '<a data-ecommerce-path="/coronavirus" data-ecommerce-row="1" data-ecommerce-index="1" href="/coronavirus">Coronavirus (COVID-19): guidance and support</a>' +
+
+      '<a data-ecommerce-path="/guidance/travel-abroad-from-england-during-coronavirus-covid-19" data-ecommerce-row="1" data-ecommerce-index="2" href="/guidance/travel-abroad-from-england-during-coronavirus-covid-19">Travel abroad from England during coronavirus (COVID-19)</a>' +
```
andysellick, line 34: Non blocking comment - it's good to base test data on real data but we can make our test data more readable by shortening things to the minimum i.e. the text and href of these links could be much shorter - as long as they're unique.

andysellick (approved):
Looks great, well done!

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.spec.js
JamesCGDS: Cool, I've just pushed up a change that will shorten those test links.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
JamesCGDS: I’ve just pushed up a change that removes these. Initially, we were having problems without setting it but I’ve just tested this locally and everything works fine without them. The only place where we’d need to keep them is in the tests as the link tracker isn’t initialised there.

PR #2959, opened 2022-09-13
Fix cookie-consent events breaking GA4 tests

AshGDS:
Hi @andysellick - would you be able to approve this when you get a chance? Thanks :+1:

## What

Ensures the `cookie-consent` event listener is removed after the `init-ga4.spec.js` tests.
## Why

We were adding a `cookie-consent` listener in our tests, which initialises our GA4 modules when the event is fired. However `cookie-functions-spec.js` fires `cookie-consent` events, which caused multiple instances of GTM to be initialised, breaking our tests as multiple click listeners/page view listeners were active.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -18,9 +18,7 @@ var initFunction = function () {
     }
     // to be added: cross domain tracking code
   } else {
-    window.addEventListener('cookie-consent', function () {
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
```
andysellick, line 10: Probably don't need a CHANGELOG for this... it's an internal change, relating only to testing, so I don't think would even qualify as a patch release.

andysellick (approved):

PR #2961, opened 2022-09-14
Refactor ga4-link-tracker and add new features

AshGDS:
Hi @andysellick,

Would you be able to approve this if all is OK?

Thanks :+1:

## What

- Allows shift clicks to be tracked, using the `event.shiftKey` boolean on click, as requested by analysts.
- Allows share and follow link clicks to be tracked. This uses a boolean on the ruby template import - either `track_as_follow` or `track_as_share`. If one these exist, a `JSON` is added into a `data-attribute` called `data-ga4-link`, which contains all the GTM data that needs to be pushed. Some values need JS though and can't be determined in the template. Therefore they are marked with a value of `'[populated_via_js]'` in the template, and then this is detected in the JavaScript as something that needs to be replaced with a new value determined in the JS.
- Tracks internal download paths as internal links instead of external, as requested by analysts.
- Adds the current hostname as an internal domain when tracking links. This allows this code to be used on any domain, instead of being hard coded to `www.gov.uk`, which was causing a bug on integration.
- Allows the internal domains, download domains, and download paths to be set using a `config` object. This allows other domains to use our code with better control over what's detected as internal/external/download links.
- Allows click listeners to be disabled on initialisation by defining a `disableListeners` boolean in the `config` object. This allows other modules to access the link tracker functions without any extra click listeners running. 
- Fixes an edge case where a href without a trailing slash such as `https://www.gov.uk` would be classed as an external link, because we are only checking for domains with a trailing slash at the end, in this case`https://www.gov.uk/`, which is safer than checking without a trailing slash.

## Why

- This fixes a few bugs (the ones in the implementation board + a couple that are not in there.)
- This allows the code to be used outside of www.gov.uk.
- The `config` object is needed for our tests, as the current hostname is `localhost` when testing, but we are testing against `www.gov.uk`, so we can add `www.gov.uk` as an internal domain when testing. 
- This allows the "link detection" functions to be used without adding additional click listeners to the DOM. Our eCommerce tracking needs to determine whether search results are external but we don't want it to initialise a new link tracker with event listeners, as this would duplicate events. The functions can't be used alone without initialising a new instance of the link tracker, because the variables needed for determining internal/external/download links are defined at initialisation, and are then referenced in the other functions.

I have also added `external="true"` or `external="false"` to download tests, as now download links can be internal or external so this was needed to classify them for the testing instead of `external` being essentially hardcoded in `beforeEach`.

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -43,25 +51,51 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
         return
       }
 
-      if (this.isMailToLink(href)) {
+      var linkAttributes = element.getAttribute('data-ga4-link')
+      if (linkAttributes) {
+        linkAttributes = JSON.parse(linkAttributes)
+        clickData = Object.assign(clickData, linkAttributes)
```
andysellick, line 57: I don't believe `Object.assign` is supported by Internet Explorer. Fortunately (as long as my understanding of assign is correct), we already have a backwards-compatible script for merging two objects, see: https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/govuk_publishing_components/lib/extend-object-spec.js

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -43,25 +51,51 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
...
+
+        /* Since external links can't be determined in the template, we use _ as a signal
+        for our JavaScript to determine this value. */
+        if (clickData.external === '_' && clickData.url) {
```
andysellick, line 61: I'm not sure I understand why this is using `_` as a flag - why can't it use `true/false`?

andysellick (commented):
I've given this a quick look and made a few comments. It's a bit hard to follow the changes - I know I've asked for you to squash related commits together, but I think this PR would really benefit from being broken into individual commits for each change, particularly as there's so many. Would it be possible to break this up a bit?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Thanks @andysellick - will use that instead :+1:

AshGDS:
> I've given this a quick look and made a few comments. It's a bit hard to follow the changes - I know I've asked for you to squash related commits together, but I think this PR would really benefit from being broken into individual commits for each change, particularly as there's so many. Would it be possible to break this up a bit?

Thanks @andysellick - will do :+1:

AshGDS:
Hi @andysellick - I've split out the changes into separate commits.

I've also changed the `_` values in the template for [Add ability to track share links and follow links](https://github.com/alphagov/govuk_publishing_components/pull/2961/commits/f831ac0dd252ac3a9d1a51c9557a51e1b23e6440) to be `[populated_via_js]` instead.

I'll fix `Object.assign` alongside any other changes you request.

Thanks :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -43,25 +43,51 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
...
+        clickData.url = href
+        clickData.text = element.textContent.trim()
+        clickData.link_method = this.getClickType(event)
       } else if (this.isDownloadLink(href)) {
         clickData.event_name = 'file_download'
         clickData.type = this.isPreviewLink(href) ? 'preview' : 'generic download'
         clickData.external = this.isExternalLink(href) ? 'true' : 'false'
+        clickData.url = href
+        clickData.text = element.textContent.trim()
+        clickData.link_method = this.getClickType(event)
```
andysellick, line 88: Non blocking comment - there's some repetition in this file with these lines. While they will never be executed more than once, we might want to think about a refactor at some point to make it a bit cleaner.

app/views/govuk_publishing_components/components/_share_links.html.erb
```diff
@@ -42,6 +43,26 @@
                 'socialNetwork': link[:icon],
                 'socialTarget': link[:href]
               }
+              ga4_link_data = {
```
andysellick, line 46: Non-blocking comment - it might be good to abstract this ruby code out into a helper method to make the template a bit cleaner, as we do with some other components.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -7,12 +7,18 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
   'use strict'
 
   var Ga4LinkTracker = {
-    init: function () {
+    init: function (config = {}) {
```
andysellick, line 10: I think having named function parameters like this is an ES6 thing? If so it's not supported by some older IEs that we want to include. Can this be rewritten somehow? (assuming I'm correct, if not just ignore me).

AshGDS:
Just rebased, and noticed where I've done

`config = {}`

This will need to be changed as it's ES6 - I will do this alongside the other changes once it has been reviewed by someone :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -7,12 +7,18 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
...
       if (window.dataLayer) {
-        this.internalLinksDomain = 'www.gov.uk/'
-        this.internalLinksDomainWithoutWww = 'gov.uk/'
+        this.config = config
```
andysellick, line 12: I'm not clear on where this config will be passed from once the code is in use. Can you provide more detail?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -19,9 +19,11 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
         this.handleClick = this.handleClick.bind(this)
         this.handleMousedown = this.handleMousedown.bind(this)
 
-        document.querySelector('body').addEventListener('click', this.handleClick)
-        document.querySelector('body').addEventListener('contextmenu', this.handleClick)
-        document.querySelector('body').addEventListener('mousedown', this.handleMousedown)
+        if (!this.config.disableListeners) {
```
andysellick, line 22: Which other modules are using this code?

andysellick (commented):
Thanks for breaking into separate commits, made it much easier to review. I still have some questions and confusions, hopefully most of them can be answered without making significant changes.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Yep, you're correct - will be simple to fix :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick Good point - I think initially I thought I'd have to surround `stopTracking()` with a check for `!this.config.disableListeners`, but it seems to still work without this check, so I had accidentally left it as `this.config`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick `ga4-enhanced-ecommerce-tracker.js` is using this code to check if a link is external.

Theoretically, `ga4-enhanced-ecommerce-tracker.js` might work without having to initialise it's own link tracker instance, since `ga4-link-tracker.js` would initialise on each pageload. However then it would rely on `ga4-link-tracker.js` loading before `ga4-enhanced-commerce-tracker.js` which seems a bit risky. 

Therefore, I was thinking we would let `ga4-enhanced-ecommerce-tracker.js` initialise it's own link tracker, but without any click listeners. What do you think about this?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick Good point - I think more link tracker types are going to be added soon (There's still attachment, internal , and navigation) so it might be safer to keep it like this for now, and optimise it once we know what those link types involve.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: I've changed it to just be `config` now.

andysellick: I seem to have stumbled upon something else accidentally! Good spot.

What I meant was that this whole block of code receives `config` as an optional parameter, but what calls this code and passes that?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick you might be right that all links in GOV.UK search are internal, but I'm not sure about the other finders. Since all the finders share the same template, I believe the eCommerce code is going to run on all the finders as soon as it's deployed even though we only scoped in GOV.UK search.

Basically, when you click on a search result, the eCommerce GTM object contains:

```JSON
"event_data": {
    "external": "true/false"
}
```

Which uses `Ga4LinkTracker.isExternalLink()` to determine wether the link is external or not. And in order to determine if something is external, the link tracker module needs to know what is classed as internal, which is grabbed when it is initialised. It relies on multiple other functions inside the link tracker too, which is why one function couldn't be simply duplicated into `ga4-enhanced-ecommerce-tracker`.

I think you might be correct that we won't need to initialise `ga4-link-tracker.js` again, but then the eCommerce code will only work if `ga4-link-tracker.js` has initialised first so it creates a dependency on the run order. This is because when `init()` is called on `ga4-link-tracker.js`, it sets what the internal domains are, the download domains, the assets paths, etc. which are needed to run `.isExternalLink()`. So if this has run before the eCommerce tracking, we wouldn't need to initialise another one. 

However a benefit of having two initialisations is that the search code may have different domains/paths which it wants to determine as internal/download links - and this would work as we can pass through a different config object to this initialisation. I'm not sure if that would ever be needed though.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick The tests call this so that it can test against `www.gov.uk` while the current hostname is `localhost`. It was also helpful in case we wanted to initialise the link tracking without the click listeners. It also allows for extra internal domains / download domains / download paths to be added in case there are multiple domains to class as internal, or multiple domains to be classed as download links, or multiple paths to be classed as download paths. It essentially removes the hard coding of the functionality only working for `www.gov.uk`'s MVP purposes and allows it to be a bit more future proof.

AshGDS:
@andysellick It looks like I need to update the percy screenshots - would you know how I go about doing this? Thanks :+1:

andysellick:
> @andysellick It looks like I need to update the percy screenshots - would you know how I go about doing this? Thanks 👍

@AshGDS in this case the change is deliberate, so we just tell Percy that this is okay. I can approve the Percy build for you.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: That sounds good. For future reference for testing, you can use a `spyOn` to override a function like `getHostname` so that it returns whatever you want, something like:

```
spyOn(Ga4LinkTracker, 'getHostname').and.returnValue('www.whateveryouwant.com')
```

AshGDS: Thanks @andysellick - I think at one point when I initially built this I was trying to spy on `window.location.hostname` which it wouldn't let me do, I didn't think of just wrapping it in a function!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: I think I'm happy to go ahead with this as it is but the close linking of these two bits of code feels brittle. Let's include this when we look at refactoring this.

(having thought about it I'm fairly sure all of our search results are internal links, but @gclssvglx or @Rosa-Fox might know for certain)

andysellick (approved):

PR #2964, opened 2022-09-15
Add click tracking to Yes/No buttons on feedback component

AshGDS:
Hi @andysellick @JamesCGDS 

Would you be able to approve this PR if all is OK?

Thanks :+1:

## What

- Adds `data-module="ga4-event-tracker"` to the feedback component parent `<div>`
- Adds a `data-ga4` JSON to the feedback component Yes/No buttons so that they are tracked on click by `ga4-event-tracker.js`

## Why

- They are needed to fulfil the MVP requirement of tracking the clicks on the Yes/No buttons
- I've hard coded the English versions of "Yes" "No" and "Is this page useful?" in the `JSON`. In the live feedback component, these strings are potentially translated into 60+ other languages, depending on what language the user's browser is set to. I'm assuming we should always send English to GA4, otherwise the analysts would have to collate all these different languages in the dashboard just to see wether Yes/No was clicked.

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
@@ -412,4 +412,65 @@ describe('Google Analytics event tracking', function () {
       expect(window.dataLayer[0].event_data.url).toEqual(undefined)
     })
   })
+
+  describe('doing tracking on the feedback component', function () {
+    beforeEach(function () {
+      var attributes = {
+        event_name: 'form_submit',
+        type: 'feedback',
+        text: 'Yes',
+        section: 'Is this page useful?'
+      }
+      element.innerHTML =
+        '<div data-module="ga4-event-tracker">' +
+      '</div>'
+
+      var yesButton = document.createElement('button')
+      yesButton.setAttribute('data-ga4', JSON.stringify(attributes))
+      yesButton.id = 'yesButton'
+
+      attributes.text = 'No'
+      var noButton = document.createElement('button')
+      noButton.setAttribute('data-ga4', JSON.stringify(attributes))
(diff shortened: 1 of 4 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
```diff
@@ -412,4 +412,65 @@ describe('Google Analytics event tracking', function () {
...
+      document.body.appendChild(element)
+      new GOVUK.Modules.Ga4EventTracker(element).init()
+
+      window.dataLayer = []
+    })
+
+    it('should track "Yes" button clicks', function () {
+      expected = new GOVUK.analyticsGA4.Schemas().eventSchema()
+
+      expected.event = 'event_data'
```
JamesCGDS, line 449: Could `expected` be set in the `beforeEach` and then just set `expected.event_data.text` in the individual tests? Only because it looks like all the properties are the same except for `expected.event_data.text`. Although if you feel that makes it more confusing to read, happy to leave it as it is 👍

andysellick (approved):
Looks good, although I'll let you investigate James' comment.

There's also a wider conversation we might need to have about enabling analytics by default on components. This one is a bit of a special case, because it's only ever used in one place with one configuration, so I think it's okay, but generally if we're adding tracking to components (either by adding a module or hard coding it into the JS) we should make it an option, rather than the default. Or should we? Like I said, wider conversation.

Also please add a CHANGELOG entry.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
AshGDS: Thanks @JamesCGDS - I've changed this to reflect what you've said :+1:

PR #2967, opened 2022-09-16
Remove usage of default parameters

JamesCGDS:
## What
When initialising ecommerce tracking, the `Ga4EnhancedEcommerceTracker.init()` function relies on a default parameter being set to a boolean. As this is an ES6 feature, it has been removed.

## Why
Using ES6 features can cause compatibility issues with older browsers and so it has been removed.

## Visual Changes
N/A


AshGDS (approved):
All good :+1:

PR #2970, opened 2022-09-20
Remove default parameters from ecommerce tracker

injms:
## What

Remove the default parameters from a function in the GA4 enhanced ecommerce tracking JavaScript.

Removing the defaults doesn't seem to have an effect as the defaults were
falsely anyway.

## Why

We shouldn't be using ES6 in production. This was preventing applications from being deployed as the uglifier/compresser was throwing an error when encountering the default parameters:

```
Uglifier::Error: Unexpected token operator «=», expected punc «,». To use ES6 syntax, harmony mode must be enabled with Uglifier.new(:harmony => true).
```

[Log output from frontend](https://ci.integration.publishing.service.gov.uk/job/frontend/job/set-up-individual-component-css/17/console)

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-enhanced-ecommerce-tracker.js
@@ -59,7 +59,7 @@
       }
     },
 
-    populateEcommerceSchema: function (searchResultsBlock, searchResultClicked = false, searchResult = null) {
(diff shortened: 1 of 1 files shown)


andysellick:
@JamesCGDS one for you to check.

Related to https://github.com/alphagov/govuk_publishing_components/pull/2967

JamesCGDS:
Thank you @injms, I missed that one. I’ve just added a commit onto this to call the parameters when they’re needed. Everything seems to work well and all the tests pass without this, but I thought it might make the code a little easier to read.

JamesCGDS (approved):

PR #2971, opened 2022-09-20
Change GTM push of unset values from 'null' to 'undefined'

AshGDS:
Hi @andysellick 

Would you be able to approve this? Thanks :+1:

## What

- This will make our GA4 code push any unset values as `undefined` instead of `null`.

## Why

- This should now display unset values as `(not set)` instead of `""` in the GA4 dashboard.
- I had to remove `var nullValue` from the page view tests, as `standardx` didn't like us setting a `var` to `undefined`
- Ecommerce has not been touched, because `we should set any missing values to undefined rather than null - except for the ecommerce parameters (and any other edge cases where [Jeff] may recommend against it.)`

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -7,7 +7,7 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
 
   var PageViewTracker = {
     PIIRemover: new window.GOVUK.analyticsGA4.PIIRemover(), // imported in analytics-ga4.js
-    nullValue: null,
+    nullValue: undefined,
 
     init: function () {
       if (window.dataLayer) {
@@ -102,7 +102,7 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
     // return only the date from given timestamps of the form
     // 2022-03-28T19:11:00.000+00:00
     stripTimeFrom: function (value) {
-      if (value !== null) {
+      if (value !== undefined) {
         return value.split('T')[0]
       } else {
         return this.nullValue
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
@@ -4,7 +4,6 @@ describe('Google Tag Manager page view tracking', function () {
   var GOVUK = window.GOVUK
   var saved = {}
   var expected
-  var nullValue = null
 
   beforeAll(function () {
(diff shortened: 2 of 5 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
```diff
@@ -4,7 +4,6 @@ describe('Google Tag Manager page view tracking', function () {
   var GOVUK = window.GOVUK
   var saved = {}
   var expected
-  var nullValue = null
```
andysellick, line 7: Could this not be changed to `undefined`? I assume there was some kind of weirdness involved?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -102,7 +102,7 @@ window.GOVUK.analyticsGA4.analyticsModules = window.GOVUK.analyticsGA4.analytics
     // return only the date from given timestamps of the form
     // 2022-03-28T19:11:00.000+00:00
     stripTimeFrom: function (value) {
-      if (value !== null) {
+      if (value !== undefined) {
```
andysellick, line 105: Is this a related change? I'm just interested in how this works... did something fail if this was left as `null`?

andysellick (commented):
Looks good, just a couple of questions, also this one probably needs a CHANGELOG entry.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
AshGDS: @andysellick yep, without changing this, the if statement will return true as `undefined !== null`, but then it will crash as it tries to do `undefined.split()`

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
AshGDS: @andysellick Yep, `standardx` said `ga4-page-views.spec.js:7:7: It's not necessary to initialize 'nullValue' to undefined.`

andysellick: Interesting... wonder why it also didn't complain about setting `this.nullValue` in the page view tracker? Ah well, fair enough.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
andysellick: Oh of course if there isn't a value it will default to being the undefined default... so it is a related change. Thanks.

AshGDS: Maybe it should be `value !== null && value != undefined` ?

andysellick (approved):

PR #2973, opened 2022-09-20
Use section for emergency banner

patrickpatrickpatrick:
## What

Small adjustments to the template of emergency banner for accessibility reasons.

## Why

When the emergency banner was deployed on GOV.UK, some changes were made based on the feedback from an accessibility specialist. These changes involved removing the role attribute from the parent div and changing the element of the parent div to a section. This is to avoid any accidental duplication of roles on any page of GOV.UK. In accordance with this change, the aria-label has been changed to aria-labelledby instead.

Co-author with @maxf who made the changes to static (which will be removed when[ the emergency banner static PR](https://github.com/alphagov/static/pull/2788) has been merged).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add ecommerce tracking documentation ([PR #2997](https://github.com/alphagov/govuk_publishing_components/pull/2997))
+* Use section for emergency banner ([PR #2973](https://github.com/alphagov/govuk_publishing_components/pull/2973))
 
 ## 31.0.0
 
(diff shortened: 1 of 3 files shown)


hannalaakso (approved):
I've compared the outputted markup against the banner that was deployed on production and the `section` tag and `aria-labelledby` linked with `id` match what was published there 👍

hannalaakso (changes requested):
Apologies for revoking the approval but I've just realised that it'd be good to add a test for the `aria-labelledby` to check it matches the `id` on the heading.

patrickpatrickpatrick:
have added that test

hannalaakso (approved):
LGTM 👍

PR #2975, opened 2022-09-20
Add new cookies set by gtag

andysellick:
## What / why
- the cookie code keeps a list of cookies that are allowed to be set, which prevents any of the code using the cookie functions from setting arbitrary cookies that we wouldn't know about
- this list is also used to clean up the cookies set when consent is withdrawn
- gtag is adding three new cookies, one for production, staging and integration, each using a unique name, which should not change over time (as it is based on the ID for each of these environments)
- adding these cookies to the list of allowed cookies, not because we set them (gtag does) but because we want to make sure that we clear them

## Visual Changes
None.

Trello card: https://trello.com/c/nBnBJE4d/375-inconsistent-setting-of-ga-cookies


andysellick:
> All looks sensible, but should there be accompanying changes to the page at https://www.gov.uk/help/cookie-details? Or is it too early for that?

This change only makes sure that if these new cookies are set, they'll be deleted - nothing here creates a new cookie. So this change is safe to deploy to live. But we will update that page before GA4 goes onto production.

maxgds:
> This change only makes sure that if these new cookies are set, they'll be deleted

Nitpick: that's not entirely true, as you noted in the PR this adds them to the permitted list and previously they couldn't have been set in order to need deletion - "the cookie code keeps a list of cookies that are allowed to be set, which prevents any of the code using the cookie functions from setting arbitrary cookies that we wouldn't know about" So long as when we start setting them for real we don't miss it that's cool, though.

andysellick:
@maxgds yes, you're correct. I meant that the code that will add these cookies won't be live on production until we've updated the cookie details page, but yes, technically if something else tried to set these cookies they'd be permitted to.

gclssvglx (approved):

PR #2979, opened 2022-09-22
Add optional gtag snippet

andysellick:
## What
Add the ability to switch GA4 from GTM to Gtag.

- if an environment variable for gtag has been set in static and passed as a variable attached to the window.GOVUK.analyticsGA4.vars object, use gtag instead of GTM
- this is for testing on integration, we will only add the gtag env var on there for now
- note that this code snippet for gtag is adapted from the GTM one, since the gtag one from Google involves <script> tags, rather than the convenience of a self executing function that creates the <script> tags and appends them to the DOM. However, both original snippets do essentially the same thing, so the GTM one has been adapted to suit and should perform exactly the same task as the original gtag snippet recommended by Google

Corresponding PR to use this in `static` is here: https://github.com/alphagov/static/pull/2886

## Why
We want to see if we can make Gtag work as intended. Note that this code will not work on any environment unless the correct env vars are set - initially we will only test this on integration.

## Visual Changes
None.

Trello card: https://trello.com/c/Oo0T8ZCa/178-spike-into-back-out-plan-from-gtm-to-gtag

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add optional gtag snippet ([PR #2979](https://github.com/alphagov/govuk_publishing_components/pull/2979))
+
 ## 30.7.1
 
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
All good :+1:

PR #2980, opened 2022-09-22
Refactor analytics code snippets

andysellick:
## What
Rewrite the code from Google that pulls in Tag Manager, so that if the `auth` and `preview` variables are not present, we don't pass them as part of the script URL.

## Why
Apparently the snippet with the extra bits in should never be used in a live site, because environments (the snippet we were using) are expensive in terms of performance, so we should avoid sending too much traffic to them - environment endpoints can become overloaded and shut down.

## Visual Changes
None.

Trello card: https://trello.com/c/zpERINGm/223-rewrite-google-snippets

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js.erb
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js.erb
@@ -6,28 +6,33 @@ window.GOVUK.analyticsGA4 = window.GOVUK.analyticsGA4 || {};
 
   var core = {
     load: function () {
-      if (window.GOVUK.analyticsGA4.vars.gtag_id) { // initialise gtag
-        (function (id) {
-          window.dataLayer = window.dataLayer || [];
-          function gtag(){dataLayer.push(arguments);}
-          gtag('js', new Date());
-          gtag('config', window.GOVUK.analyticsGA4.vars.gtag_id);
-          var firstScript = document.getElementsByTagName('script')[0];
-          var newScript = document.createElement('script');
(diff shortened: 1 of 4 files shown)


JamesCGDS (approved):
All good so will approve this 👍  Non-blocking comment - could the `fix` commit be squashed?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js.erb
```diff
@@ -6,28 +6,33 @@ window.GOVUK.analyticsGA4 = window.GOVUK.analyticsGA4 || {};
...
+        // initialise GTM
+        window.dataLayer = window.dataLayer || []
+        window.dataLayer.push({ 'gtm.start': new Date().getTime(), event: 'gtm.js' })
+
+        var firstScript = document.getElementsByTagName('script')[0]
+        var newScript = document.createElement('script')
+        var dl = 'dataLayer' != 'dataLayer' ? '&l=' + 'dataLayer' : ''
+
+        newScript.async = true
+        newScript.src = 'https://www.googletagmanager.com/gtm.js?id=' + window.GOVUK.analyticsGA4.vars.id + dl + '&gtm_auth=' + window.GOVUK.analyticsGA4.vars.auth + '&gtm_preview=' + window.GOVUK.analyticsGA4.vars.preview + '&gtm_cookies_win=x'
```
gclssvglx: Should we not be environment specific? On integration and staging this is fine, but remove the `gtm_auth` and `gtm_preview` on production?

gclssvglx (commented):
Looking good @andysellick - but I do have one question...

PR #2982, opened 2022-09-23
Lint ga4-core

andysellick:
## What
Fix up some linting issues in GA4-core.

- because this file is a .js.erb it was ignored by our linter
- but it turns out it had some linting problems, which were only picked up later when the gem with this code in was installed in an application and included, built and then linted, causing errors
- linting issues addressed

## Why
Just tried to build `static` with the gem containing this unlinted code and it broke 😞 

## Visual Changes
None.

Trello card: https://trello.com/c/zpERINGm/223-rewrite-google-snippets

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


gclssvglx (approved):
Yes, this is strange that on CI it precompile's the assets then lints 🤔 Nice one - looks good me 👍

PR #2985, opened 2022-09-23
Fix bug when getting target element from URL hash

MartinJJones:
## What
Update the accordion component to only look for a target element within the module itself, not the entire document.

## Why
This fixes a bug when `anchor_navigation` is set to `true` and an ID is passed in for an element that exists on the page, but is not contained within the accordion component.

For example: https://www.gov.uk/guidance/how-to-publish-on-gov-uk/creating-and-updating-pages#section-title

## Further Info

Fixes: https://github.com/alphagov/govuk_publishing_components/issues/2086
Trello: https://trello.com/c/q5P6u7id/1507-ps-17-accordion-anchor-link-navigation-breaks-with-invalid-anchor

### Using querySelector when the ID contains a colon
There was also a need to escape the colon character when using `querySelector`, for example `$this.module.querySelector('#fnref:1')` would throw an error [^1]

[^1]: [Bugzilla - querySelector doesn't work if the ID contains a colon (:)](https://bugzilla.mozilla.org/show_bug.cgi?id=883044)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -59,11 +59,9 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
   // Navigate to and open accordions with anchored content on page load if a hash is present
   GemAccordion.prototype.openByAnchorOnLoad = function () {
+    if (!window.location.hash) return
     var splitHash = window.location.hash.split('#')[1]
-
-    if (window.location.hash && document.getElementById(splitHash)) {
-      this.openForAnchor(splitHash)
-    }
+    this.openForAnchor(splitHash)
   }
 
   // Add event listeners for links to open accordion sections when navigated to using said anchor links on the page
@@ -80,7 +78,9 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
   // Find the parent accordion section for the given id and open it
   GemAccordion.prototype.openForAnchor = function (hash) {
-    var target = document.getElementById(hash)
+    hash = hash.replace(':', '\\:')
+    var target = this.$module.querySelector('#' + hash)
+    if (!target) return
     var $section = this.getContainingSection(target)
     var $header = $section.querySelector(this.sectionHeader)
     var $expanded = this.getContainingSection($section)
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -59,9 +59,15 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
   // Navigate to and open accordions with anchored content on page load if a hash is present
   GemAccordion.prototype.openByAnchorOnLoad = function () {
+    if (!window.location.hash) return
     var splitHash = window.location.hash.split('#')[1]
+    var targetID = window.location.hash
```
jon-kirwan, line 64: Tiny change suggestion here but it looks like 'ID' tends to be written as 'Id' so perhaps we should update the variable name to `targetId`

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -80,7 +86,11 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
   // Find the parent accordion section for the given id and open it
   GemAccordion.prototype.openForAnchor = function (hash) {
-    var target = document.getElementById(hash)
+    if (hash.indexOf(':') !== -1) {
+      hash = hash.replace(':', '\\:')
+    }
+
```
jon-kirwan, line 92: Another tiny suggestion here - but I think the code above looks quite similar to the code here:

```
GemAccordion.prototype.openForAnchor = function (hash) {
  if (hash.indexOf(':') !== -1) {
    hash = hash.replace(':', '\\:')
  }
  ...
```

```
GemAccordion.prototype.openByAnchorOnLoad = function () {
  if (!window.location.hash) return
  var splitHash = window.location.hash.split('#')[1]
  var targetID = window.location.hash

  if (targetID.indexOf(':') !== 1) {
    targetID = targetID.replace(':', '\\:')
  }
  ...
```
Maybe we could have a separate function that could work for both?

jon-kirwan (approved):
Looks great thanks! I've added two minor suggestions but not critical changes by any means so have approved

jon-kirwan (approved):
All good! 👍

MartinJJones:
Amended note in the Changelog

PR #3004, opened 2022-10-06
Fix issues with GA4 link tracking identified by performance analysts

AshGDS:
Hi @andysellick / @JamesCGDS - would either (or both) of you be able to approve this PR? 

Also, do you think this is too many changes in one branch? I like to tackle multiple tickets at a time, but I appreciate that some people would treat each commit as a separate PR.

Thanks :+1:

## What

- Strips out `_gl` and `_ga` query parameters from being pushed with link clicks
- Adds a `link_domain` field which tracks what protocol and domain the clicked link is pointing to (e.g. `https://www.gov.uk` or `https://example.com`)
- Adds a `link_path_parts` object which stores strings. These are the clicked link's path, split at every 100 characters.
- Changes the `link_method` key to just be `method`.
- Strips line breaks and multiple spaces in link text before being pushed to the dataLayer.
- Updates the docs to reflect the above changes.

## Why

- The `_gl` and `_ga` query parameters are used for cross domain tracking to identify a user across domains. It isn't useful for the link click tracking as it won't tell us anything, and they are long query parameters which is wasting precious space.
- Adds ` link_domain` value. This was requested by the PAs as it will be useful for them to see which domains are being clicked. It will also be needed to reconstruct long URLs in Data Studio that have been split up - related to the next point: 
- Adds a `link_path_parts` object. GA4 only allows event values to be a maximum of 100 characters. Therefore, a lot of the link URLs were being cut off after 100 characters, which means we are pushing incomplete data to GA4. To solve this, we have created an object which dynamically adds the path of the URL, split every 100 characters. We have set a limit of splitting it into a maximum of 5 parts (500 characters) - this can be changed if the analysts want. The analysts also requested we add the link path to this object, even when the whole link is under 100 characters.
- Changes the `link_method` key to just be `method` because GA4 already has a `method` key configured, so there was no point in having our own key called `link_method`.
- We strip line breaks and multiple spaces in link text as this was causing visual issues in the GA4 dashboard.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -61,8 +61,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
           clickData.external = this.isExternalLink(clickData.url) ? 'true' : 'false'
         }
 
-        if (clickData.link_method === 'populated-via-js') {
-          clickData.link_method = this.getClickType(event)
+        if (clickData.method === 'populated-via-js') {
+          clickData.method = this.getClickType(event)
         }
 
         if (clickData.index) {
@@ -78,21 +78,21 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
         clickData.external = 'true'
         clickData.url = href
         clickData.text = element.textContent.trim()
-        clickData.link_method = this.getClickType(event)
+        clickData.method = this.getClickType(event)
       } else if (this.isDownloadLink(href)) {
         clickData.event_name = 'file_download'
         clickData.type = this.isPreviewLink(href) ? 'preview' : 'generic download'
         clickData.external = this.isExternalLink(href) ? 'true' : 'false'
         clickData.url = href
         clickData.text = element.textContent.trim()
-        clickData.link_method = this.getClickType(event)
+        clickData.method = this.getClickType(event)
       } else if (this.isExternalLink(href)) {
         clickData.event_name = 'navigation'
         clickData.type = 'generic link'
         clickData.external = 'true'
         clickData.url = href
         clickData.text = element.textContent.trim()
-        clickData.link_method = this.getClickType(event)
+        clickData.method = this.getClickType(event)
       }
 
       if (Object.keys(clickData).length > 0) {
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -97,6 +97,10 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
 
       if (Object.keys(clickData).length > 0) {
+        if (clickData.url) {
+          clickData.url = this.removeCrossDomainParams(clickData.url)
+        }
+
         var schema = new window.GOVUK.analyticsGa4.Schemas().eventSchema()
         schema.event = 'event_data'
 
@@ -236,6 +240,13 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       return string.substring(0, stringToFind.length) === stringToFind
     },
 
+    stringEndsWith: function (string, stringToFind) {
+      if (stringToFind.length > string.length) {
+        return false
+      }
+      return string.substring(string.length - stringToFind.length, string.length) === stringToFind
+    },
+
     hrefIsRelative: function (href) {
       // Checks that a link is relative, but is not a protocol relative url
       return href[0] === '/' && href[1] !== '/'
@@ -247,6 +258,22 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
 
     getHostname: function () {
       return window.location.hostname
+    },
+
+    removeCrossDomainParams: function (url) {
+      if (url.indexOf('_ga') !== -1 || url.indexOf('_gl') !== -1) {
+        // _ga & _gl are values needed for cross domain tracking, but we don't want them included in our click tracking.
+        url = url.replace(/_ga=([^&]*)/, '')
+        url = url.replace(/_gl=([^&]*)/, '')
+
+        // The following code cleans up inconsistencies such as gov.uk/&&, gov.uk/?&hello=world, gov.uk/?, and gov.uk/&.
+        url = url.replaceAll(/(&&)+/g, '&')
+        url = url.replace('?&', '?')
+        if (this.stringEndsWith(url, '?') || this.stringEndsWith(url, '&')) {
+          url = url.substring(0, url.length - 1)
+        }
+      }
+      return url
     }
   }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -77,21 +77,21 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
         clickData.type = 'email'
         clickData.external = 'true'
         clickData.url = href
-        clickData.text = element.textContent.trim()
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
         clickData.method = this.getClickType(event)
       } else if (this.isDownloadLink(href)) {
         clickData.event_name = 'file_download'
         clickData.type = this.isPreviewLink(href) ? 'preview' : 'generic download'
         clickData.external = this.isExternalLink(href) ? 'true' : 'false'
         clickData.url = href
-        clickData.text = element.textContent.trim()
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
         clickData.method = this.getClickType(event)
       } else if (this.isExternalLink(href)) {
         clickData.event_name = 'navigation'
         clickData.type = 'generic link'
         clickData.external = 'true'
         clickData.url = href
-        clickData.text = element.textContent.trim()
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
         clickData.method = this.getClickType(event)
       }
 
@@ -187,6 +187,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
           domainsArrays.push(domainWithoutWww)
         }
       }
+    },
+
+    removeLinesAndExtraSpaces: function (text) {
+      text = text.replace(/(\r\n|\n|\r)/gm, ' ') // Replace line breaks with 1 space
+      text = text.replace(/\s+/g, ' ') // Replace instances of 2+ spaces with 1 space
+      return text
     },
 
     getClickType: function (event) {
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -52,8 +52,7 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       var clickData = {}
       var linkAttributes = element.getAttribute('data-ga4-link')
       if (linkAttributes) {
-        linkAttributes = JSON.parse(linkAttributes)
-        clickData = window.GOVUK.extendObject(clickData, linkAttributes)
+        clickData = JSON.parse(linkAttributes)
 
         /* Since external links can't be determined in the template, we use populated-via-js as a signal
         for our JavaScript to determine this value. */
@@ -66,40 +65,40 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
         }
 
         if (clickData.index) {
-          clickData.index = parseInt(linkAttributes.index)
+          clickData.index = parseInt(clickData.index)
         }
 
         if (clickData.index_total) {
-          clickData.index_total = parseInt(linkAttributes.index_total)
+          clickData.index_total = parseInt(clickData.index_total)
         }
       } else if (this.isMailToLink(href)) {
         clickData.event_name = 'navigation'
         clickData.type = 'email'
         clickData.external = 'true'
         clickData.url = href
-        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent)
         clickData.method = this.getClickType(event)
       } else if (this.isDownloadLink(href)) {
         clickData.event_name = 'file_download'
         clickData.type = this.isPreviewLink(href) ? 'preview' : 'generic download'
         clickData.external = this.isExternalLink(href) ? 'true' : 'false'
         clickData.url = href
-        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent)
         clickData.method = this.getClickType(event)
       } else if (this.isExternalLink(href)) {
         clickData.event_name = 'navigation'
         clickData.type = 'generic link'
         clickData.external = 'true'
         clickData.url = href
-        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent)
         clickData.method = this.getClickType(event)
       }
 
       if (Object.keys(clickData).length > 0) {
         if (clickData.url) {
-          this.removeCrossDomainParams(clickData)
-          this.populateLinkDomain(clickData)
-          this.populateLinkPathParts(clickData)
+          clickData.url = this.removeCrossDomainParams(clickData.url)
+          clickData.link_domain = this.populateLinkDomain(clickData.url)
+          clickData.link_path_parts = this.populateLinkPathParts(clickData.url)
         }
 
         var schema = new window.GOVUK.analyticsGa4.Schemas().eventSchema()
@@ -117,14 +116,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
-    populateLinkPathParts: function (clickData) {
-      var href = clickData.url
+    populateLinkPathParts: function (href) {
       var path = ''
-      if (this.hrefIsRelative(href)) {
-        path = href
-      } else if (this.isMailToLink(href)) {
+      if (this.hrefIsRelative(href) || this.isMailToLink(href)) {
         path = href
       } else {
+        // This regex matches a protocol and domain name at the start of a string such as https://www.gov.uk, http://gov.uk, //gov.uk
         path = href.replace(/^(http:||https:)?(\/\/)([^\/]*)/, '') // eslint-disable-line no-useless-escape
       }
 
@@ -132,49 +129,33 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
         return
       }
 
-      /* Divide the length into hundreds, to determine how many string "parts" to make.
-      For example a 150 char string length divided by 100 would return 1.5,
-      so we round up, which returns 2, which makes sense as we would need one segment for chars 0-100, and another for chars 100-150 . */
-      var hundreds = Math.ceil(path.length / 100)
-
-      var LIMIT = 5
-
-      if (hundreds > LIMIT) {
-        hundreds = LIMIT // Limit the amount of data we send to 500 chars
+      /*
+      This will create an object with 5 keys that are indexes ("1", "2", etc.)
+      The values will be each part of the link path split every 100 characters, or undefined.
+      For example: {"1": "/hello/world/etc...", "2": "/more/path/text...", "3": undefined, "4": undefined, "5": undefined}
+      Undefined values are needed to override the persistent object in GTM so that any values from old pushes are overwritten.
+      */
+      var parts = path.match(/.{1,100}/g)
+      var obj = {}
+      for (var i = 0; i < 5; i++) {
+        obj[(i + 1).toString()] = parts[i]
       }
-
-      clickData.link_path_parts = {}
-
-      for (var i = 0; i < LIMIT; i++) {
-        var stringIndex = (i + 1).toString() // Start the GTM index from 1 for analysts benefit
-        clickData.link_path_parts[stringIndex] = undefined // Ensure the object keys are cleared, so link click paths aren't mixed on separate pushes
-      }
-
-      for (i = 0; i < hundreds; i++) {
-        var startIndex = i * 100
-        stringIndex = (i + 1).toString() // Start the GTM index from 1 for analysts benefit
-        // If it's the last index, substring from the start index to the end of the path string. Else, grab the next 100 characaters.
-        if (i === LIMIT) {
-          clickData.link_path_parts[stringIndex] = path.substring(startIndex, href.length)
-        } else {
-          clickData.link_path_parts[stringIndex] = path.substring(startIndex, startIndex + 100)
-        }
-      }
+      return obj
     },
 
-    populateLinkDomain: function (clickData) {
-      var href = clickData.url
+    populateLinkDomain: function (href) {
       // We always want mailto links to have an undefined link_domain
       if (this.isMailToLink(href)) {
-        return
+        return undefined
       }
 
       if (this.hrefIsRelative(href)) {
-        clickData.link_domain = this.getProtocol() + '//' + this.getHostname()
+        return this.getProtocol() + '//' + this.getHostname()
       } else {
+        // This regex matches a protocol and domain name at the start of a string such as https://www.gov.uk, http://gov.uk, //gov.uk
         var domainRegex = /^(http:||https:)?(\/\/)([^\/]*)/ // eslint-disable-line no-useless-escape
         var domain = domainRegex.exec(href)[0]
-        clickData.link_domain = domain
+        return domain
       }
     },
 
@@ -190,6 +171,7 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
     },
 
     removeLinesAndExtraSpaces: function (text) {
+      text = text.trim()
       text = text.replace(/(\r\n|\n|\r)/gm, ' ') // Replace line breaks with 1 space
       text = text.replace(/\s+/g, ' ') // Replace instances of 2+ spaces with 1 space
       return text
@@ -309,9 +291,6 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
     },
 
     stringEndsWith: function (string, stringToFind) {
-      if (stringToFind.length > string.length) {
-        return false
-      }
       return string.substring(string.length - stringToFind.length, string.length) === stringToFind
     },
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +116,22 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
+    populateLinkDomain: function (clickData) {
```
andysellick, line 165: I'm a bit uncertain here - this function doesn't seem to return anything? It might be working in some clever way that I've missed but generally for clarity I'd either have a function return something or have it act on a `this` variable.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +116,22 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+      var href = clickData.url
+      // We always want mailto links to have an undefined link_domain
+      if (this.isMailToLink(href)) {
```
andysellick, line 168: I think this function might be called elsewhere in the code (although I haven't checked in detail) - would it be possible to store the value for this once, rather than calling this function multiple times?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +116,22 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+        return
+      }
+
+      if (this.hrefIsRelative(href)) {
+        clickData.link_domain = this.getProtocol() + '//' + this.getHostname()
+      } else {
+        var domainRegex = /^(http:||https:)?(\/\/)([^\/]*)/ // eslint-disable-line no-useless-escape
```
andysellick, line 175: Maybe worth adding a quick comment to explain what this regex does?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -260,7 +276,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       return window.location.hostname
     },
 
-    removeCrossDomainParams: function (url) {
+    getProtocol: function () {
+      return window.location.protocol
+    },
+
+    removeCrossDomainParams: function (clickData) {
```
andysellick, line 335: Same comment here about returning values as for the `populateLinkDomain` function.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -32,34 +34,28 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
...
-              '<a href="nationalarchives3.gov.uk/one.pdf">National Archives PDF</a>' +
+              '<a href="http://nationalarchives.gov.uk" link_domain="http://nationalarchives.gov.uk"> National Archives </a>' +
+              '<a href="https://nationalarchives.gov.uk" link_domain="https://nationalarchives.gov.uk"></a>' +
+              '<a href="https://nationalarchives.gov.uk/one.pdf" link_domain="https://nationalarchives.gov.uk">National Archives PDF</a>' +
             '</div>' +
             '<div class="protocol-relative-external-links">' +
-              '<a href="//nationalarchives1.gov.uk"> National Archives </a>' +
-              '<a href="//nationalarchives2.gov.uk"></a>' +
-              '<a href="//nationalarchives3.gov.uk/one.pdf">National Archives PDF</a>' +
+              '<a href="//nationalarchives.gov.uk"> National Archives </a>' +
```
andysellick, line 55: The reason this test data was originally numbered was because during some debugging it was much easier to differentiate between the links if they had different domains - might be a feature worth keeping (or you could change the URL to have a unique identifier, as you did with the links on line 37 onwards).

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -32,34 +34,28 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
       links = document.createElement('div')
       links.innerHTML =
           '<div class="fully-structured-external-links">' +
-              '<a href="http://www.nationalarchives1.gov.uk"> National Archives </a>' +
-              '<a href="https://www.nationalarchives2.gov.uk"></a>' +
-              '<a href="https://www.nationalarchives3.gov.uk/one.pdf">National Archives PDF</a>' +
+              '<a href="http://www.nationalarchives.gov.uk/1" link_domain="http://www.nationalarchives.gov.uk"> National Archives </a>' +
```
andysellick: Might be worth adding a comment here to explain why the `link_domain` attribute is here and that it's not expected in normal usage.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -103,20 +99,7 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
         var link = linksToTest[i]
         GOVUK.triggerEvent(link, 'click')
 
-        expected.event_data.url = link.getAttribute('href')
-        expected.event_data.text = link.innerText.trim()
-        expect(window.dataLayer[0]).toEqual(expected)
-      }
-    })
-
-    it('detects external click events on missing http/https external links', function () {
```
andysellick, line 112: Just wanted to check that this bit was deleted on purpose - it's not immediately clear to me what's changed here.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -143,6 +127,7 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
 
       GOVUK.triggerEvent(nestedImage, 'click')
 
+      expected.event_data.link_domain = 'http://www.nationalarchives.gov.uk'
```
andysellick, line 148: Any reason why sometimes the expected `link_domain` is hardcoded and sometimes it reads it from the attribute? Might be good to be consistent if possible?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -263,56 +252,44 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
 
       links = document.createElement('div')
       links.innerHTML = '<div class="fully-structured-download-links">' +
-            '<a href="https://assets.publishing.service.gov.uk/one.pdf" external="true">PDF</a>' +
-            '<a href="https://assets.publishing.service.gov.uk/two.xslt" external="true">Spreadsheet</a>' +
-            '<a href="https://www.gov.uk/government/uploads/system/three.doc" external="false">Document</a>' +
+            '<a href="https://assets.publishing.service.gov.uk/one.pdf" external="true" link-domain="https://assets.publishing.service.gov.uk">PDF</a>' +
```
andysellick: Nitpick: you've used `link-domain` here but `link_domain` previously, might be good to be consistent.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +117,51 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
+    populateLinkPathParts: function (clickData) {
+      var href = clickData.url
+      var path = ''
+      if (this.hrefIsRelative(href)) {
+        path = href
+      } else if (this.isMailToLink(href)) {
```
andysellick, line 125: Might be worth storing the of `isMailToLink` if possible to prevent having to call it more than once. Not sure it's possible though.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +117,51 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
+    populateLinkPathParts: function (clickData) {
+      var href = clickData.url
+      var path = ''
+      if (this.hrefIsRelative(href)) {
```
andysellick, line 123: This and the following `else if` result in the same outcome. Would it be possible either combine them or maybe even invert the whole statement e.g.

```
if (some condition I don't know exactly) {
  path = href.replace(/^(http:||https:)?(\/\/)([^\/]*)/, '') // eslint-disable-line no-useless-escape
} else {
  path = href
}
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +117,51 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+      }
+
+      /* Divide the length into hundreds, to determine how many string "parts" to make.
+      For example a 150 char string length divided by 100 would return 1.5,
+      so we round up, which returns 2, which makes sense as we would need one segment for chars 0-100, and another for chars 100-150 . */
+      var hundreds = Math.ceil(path.length / 100)
+
+      var LIMIT = 5
+
+      if (hundreds > LIMIT) {
```
andysellick, line 142: Not sure which would be more efficient, but you can also do `hundreds = Math.min(hundreds, LIMIT)`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +117,51 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+        hundreds = LIMIT // Limit the amount of data we send to 500 chars
+      }
+
+      clickData.link_path_parts = {}
+
+      for (var i = 0; i < LIMIT; i++) {
+        var stringIndex = (i + 1).toString() // Start the GTM index from 1 for analysts benefit
```
andysellick, line 149: If you want the loop to be 1 indexed you can do `for (var i = 1; i <= LIMIT; i++) {`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -116,6 +117,51 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+        clickData.link_path_parts[stringIndex] = undefined // Ensure the object keys are cleared, so link click paths aren't mixed on separate pushes
+      }
+
+      for (i = 0; i < hundreds; i++) {
```
andysellick, line 153: There's probably a reason for this, but would it be possible to combine this loop with the loop above? They seem related. It's also a little confusing why one uses `LIMIT` and the other uses `hundreds`, when you've already compared the two. Why is that?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -34,19 +42,19 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
       links = document.createElement('div')
       links.innerHTML =
           '<div class="fully-structured-external-links">' +
-              '<a href="http://www.nationalarchives.gov.uk/1" link_domain="http://www.nationalarchives.gov.uk"> National Archives </a>' +
-              '<a href="https://www.nationalarchives.gov.uk/2" link_domain="https://www.nationalarchives.gov.uk"></a>' +
-              '<a href="https://www.nationalarchives.gov.uk/3.pdf" link_domain="https://www.nationalarchives.gov.uk">National Archives PDF</a>' +
+              '<a href="http://www.nationalarchives.gov.uk/1" link_domain="http://www.nationalarchives.gov.uk" path="/1"> National Archives </a>' +
```
andysellick, line 45: Might help to have a comment explaining what the non-valid `path` attribute is for.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -644,5 +718,70 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
...
+          1: '/supercalifragilisticexpialidocious-even-though-the-sound-of-it-is-something-qui'
+        },
+        {
+          1: '/supercalifragilisticexpialidocious-even-though-the-sound-of-it-is-something-qui'
+        }
+      ]
+
+      var linksToTest = document.querySelectorAll('.long-links a')
+      for (var i = 0; i < linksToTest.length; i++) {
+        defaultLinkPathParts = {
```
andysellick, line 797: This object is being defined multiple times throughout this file - could this bit be moved to the top level `beforeEach`?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -77,21 +77,21 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
         clickData.type = 'email'
         clickData.external = 'true'
         clickData.url = href
-        clickData.text = element.textContent.trim()
+        clickData.text = this.removeLinesAndExtraSpaces(element.textContent.trim())
```
andysellick, line 80: Maybe put the `trim()` call inside `this.removeLinesAndExtraSpaces`? Seems odd to consistently call the function with the parameter wrapped in another function.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -189,6 +189,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
+    removeLinesAndExtraSpaces: function (text) {
+      text = text.replace(/(\r\n|\n|\r)/gm, ' ') // Replace line breaks with 1 space
+      text = text.replace(/\s+/g, ' ') // Replace instances of 2+ spaces with 1 space
```
andysellick, line 194: Good comments 👍

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -719,6 +719,31 @@ describe('GOVUK.analyticsGa4.linkTracker', function () {
...
     })
 
+    it('cleans up link text with multiple lines and spaces', function () {
+      linkTracker = GOVUK.analyticsGa4.analyticsModules.Ga4LinkTracker
+      linkTracker.init({ internalDomains: ['www.gov.uk'] })
+
+      links.innerHTML = '<div class="messy-text">' +
+      '<a href="https://example.com">\nI \nam \non \nmultiple \n\n\n lines</a>' +
+      '<a href="https://www.gov.uk/government/uploads/example.pdf">I     have     lots   of    spaces</a>' +
+      '<a href="mailto:hello@example.com">I     have     lots   of    spaces \n and \n\n many \n\n\n lines</a>' +
```
andysellick, line 729: Not sure how this actually works, but is it worth checking real line breaks? e.g.

```
'<a href="mailto:hello@example.com">I     have     lots   of    spaces \n and \n\n many ' + 
'many ' +
'lines</a>' +
```

andysellick (commented):
Some comments on code structure, but mostly makes sense. Not for this PR but maybe in future worth splitting into separate PRs, as was quite a lot to go through.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -236,6 +240,13 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       return string.substring(0, stringToFind.length) === stringToFind
     },
 
+    stringEndsWith: function (string, stringToFind) {
+      if (stringToFind.length > string.length) {
```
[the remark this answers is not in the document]
AshGDS: I think you're right that I should change it - I was worried that if `stringToFind`  was longer than the `string` it would cause a problem as the start index for the substring would be negative, but it seems that JS allows a negative index.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick objects are passed by reference, so the object passed through to the function gets modified directly. So `clickData` gets modified without the need to return anything. Should I change it so that it returns the updated values instead?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick Yep it does get called previously. I think the code might get a bit more confusing if we try and optimise here though but I'll see how it looks

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: @andysellick It can't because it needs to be reset before each link click, instead of before each test

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: @andysellick I don't think JS counts line breaks outside of quotes. I just tried your string in jsFiddle and it doesn't add any line breaks for `many many lines</a>"`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: See here https://github.com/alphagov/govuk_publishing_components/pull/3004#discussion_r990078024

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: Can do, I simplified the domain so that I didn't have to manually add a `link_domain` attribute to each link, but I can add that in.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: @andysellick I deleted it because they were testing broken links - if the link doesn't have `http://` or `https://` or `//` at the start it's treated as an internal link. For example `href="google.com` on our homepage, JS resolves it to be  `https://www.gov.uk/google.com` - so this test was wrong as links looking like that wouldn't exist

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: @andysellick Some tests always have the same domain name, so I could hard code it in the test whereas some tests had a bunch of different domains, so I had to mark them up so the `expected` value would change depending on the domain. Should I use the `link_domain` attribute everywhere instead?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: See https://github.com/alphagov/govuk_publishing_components/pull/3004#discussion_r990089935

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Yes it should be `if(this.hrefIsRelative(href) || this.isMailToLink(href))`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Good idea :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Good idea :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick I basically need to create the 5 object keys first and set them as `undefined`. It uses LIMIT as that's how many keys we want the object to have.  We need all 5 values to be set to undefined, so that any previous values from a previous push will be overwritten.

Then, I can overwrite the `undefined` with the string values. This uses `hundreds` as it allows us to only loop the amount of times needed to add the string to the object. If we used `LIMIT`, we are always assuming that the string will fill up 5 segments. With `hundreds`, we're getting the amount of segments based on the strings length, and then if it hasn't taken up 5 keys, we are sending `undefined` for the remaining keys which we know exist because of our previous loop.

If I were to combine them, it would get complicated, because I would be trying to add the string to the object, and then if the string has less than 5 segments, I'm then going to need to add the keys with the value `undefined` after it to ensure the object always has 5 keys. 

Essentially it's like how you created the `eventSchema`, we:

- Create the schema object with the keys that will exist in GTM, all of them set as undefined
- Add values to any keys which are relevant

I guess the reason the object is created in a loop is so that the `LIMIT` value can simply be modified, and then the schema object, and the max amount of segments that can be added to the object are both updated accordingly.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: I didn't implement this in the end as I think it'll make the code a bit too confusing - let me know if you want me to explain in more detail.

AshGDS:
Thanks @andysellick - I've made the requested changes where I could, and have left conversations unresolved where I need your input (make sure to expand the hidden conversations!)

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: I guess if it works okay and there's no risk of a problem then it's okay - I'm just not sure what the accepted practice is for this kind of situation.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: Okay, as long as you've had a look, that's fine.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
andysellick: Ah okay, thanks.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
andysellick: No, that makes sense, thanks.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: I think that this whole section could be written more simply. I've just tried replacing 134-159 with this, and it seems to work. Can you check and see if this would be a suitable alternative?

```JavaScript
var parts = path.match(/.{1,100}/g)
var obj = {}
for (var i = 0; i < 5; i++) {
  obj[(i + 1).toString()] = parts[i]
}
clickData.link_path_parts = obj
```

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
andysellick: Ah yes, sorry, missed that. However if you leave this one in place (I see now that it resets each time through the loop) you can remove all the other instances in the rest of the file (e.g. line 511, 579, etc.) and put those all in a new `beforeEach` at the start of the file.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Nice solution, thanks @andysellick

AshGDS:
Thanks @andysellick - I've made those changes.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -247,6 +308,27 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+    getProtocol: function () {
+      return window.location.protocol
+    },
+
+    removeCrossDomainParams: function (clickData) {
+      var url = clickData.url
+      if (url.indexOf('_ga') !== -1 || url.indexOf('_gl') !== -1) {
+        // _ga & _gl are values needed for cross domain tracking, but we don't want them included in our click tracking.
+        url = url.replace(/_ga=([^&]*)/, '')
+        url = url.replace(/_gl=([^&]*)/, '')
```
andysellick, line 322: I think it might be possible to combine these two lines into a single regex: `_g[al]=([^&]*)`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -247,6 +308,27 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+
+        // The following code cleans up inconsistencies such as gov.uk/&&, gov.uk/?&hello=world, gov.uk/?, and gov.uk/&.
+        url = url.replaceAll(/(&&)+/g, '&')
+        url = url.replace('?&', '?')
```
andysellick, line 326: Just want to check... this line uses `replace` but the previous line uses `replaceAll` - is that okay? Will there only ever be on instance of `?&`?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -44,13 +44,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
       var href = element.getAttribute('href')
 
       if (!href) {
         return
       }
-
+      var clickData = {}
       var linkAttributes = element.getAttribute('data-ga4-link')
       if (linkAttributes) {
         linkAttributes = JSON.parse(linkAttributes)
```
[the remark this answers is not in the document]
AshGDS: Good spot :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick I think the idea is there'd only be one instance of `?` in the URL, so `?&` would only happen once

andysellick (approved):
Well done. Would it be possible to merge some of the fixup commits?

PR #3011, opened 2022-10-12
Move the GTM blocklist code

andysellick:
## What / why
Discovered this problem during testing. The blocklist has to come first or it doesn't work.

- make it so that the blocklist is the very first thing that is added to the dataLayer when GTM initialises
- otherwise it doesn't work

## Visual Changes
None.


gclssvglx (approved):
Nice one Andy 👍

PR #3014, opened 2022-10-13
Set up gem so stylesheets can be loaded individually

injms:
## What

Set the component gem up so that rendering applications can (optionally) load each component's stylesheet individually, as described in [RFC #149][rfc].

### TL;DR

An array for stylesheet names is set up in the rendering application using a helper. Templates can then add the names of stylesheets to the array, which is then used by the application to output all the stylesheets needed on the page.

Example of rendering application upgrade: https://github.com/alphagov/frontend/pull/3342/

### Initial set up

There are 3 types of stylesheet that are used across GOVUK:

- CSS from components in the gem
- CSS from components in the application
- CSS from views in the application

The helper adds each type of stylesheet to the array - gem components, app components, and views.

Any other styles should be added in the application's `application.scss` file.

Four shorthand methods for adding each type of stylesheet to the array:

- `add_gem_component_stylesheet`
- `add_app_component_stylesheet`
- `add_view_stylesheet`
- `add_stylesheet_path`

The first three take the name of the Sass file relevant to the component or view:

```ruby
add_gem_component_stylesheet("accordion")
```
```ruby
add_app_component_stylesheet("calendar")
```
```ruby
add_view_stylesheet("homepage")
```

There is also 1 generic method that takes a file path:

```ruby
add_stylesheet_path("views/dir-1/_partial.css")
```

### Precompiling

The stylesheets need to be added to the rendering applications list of assets to precompile. The `get_component_css_paths` method is available from the AssetHelper module in GovukPublishingComponents.

```ruby
#config/application.rb
module Frontend
  class Application < Rails::Application
    include GovukPublishingComponents::AppHelpers::AssetHelper
    ...
    config.assets.precompile << get_component_css_paths
```

Stylesheets for components in the gem are automatically generated, any app component and view stylesheets should be added to the application's `manifest.js`.

#### Nested components in GOV.UK Frontend

GOV.UK Frontend has components that use other components - for example, both the radio button and character count component include the label component. These are deduped by GOV.UK Frontend when compiling the CSS to a single file.

As this pull request includes the stylesheets for the components individually, this will mean that both the radio stylesheet and the character count stylesheet will include the label component.

This shouldn't make a difference in visual appearance of the site, since the stylesheets were included in alphabetical order and continue to do so. However, it means that there still is some CSS that is unnecessarily included twice.

The page sizes were calculated from individual compiled CSS files, so there still is a performance improvement - but it could be further improved.

Ideally there'd be a way of including only the components without including the nested components; for example, the partial for radio inputs would include both the label and radio:

```ruby
add_gem_component_stylesheet("label")
add_gem_component_stylesheet("radio")
```

which would then be deduped before being added to the array of stylesheets.

#### Increased request header size

Rails automatically adds the stylesheets on the page to the request `preload` header. This means that it gets bigger with every additional stylesheet that's added to the page. Coupled with a large Content Security Policy (CSP), this means that some pages have a header size that nginx doesn't like at all - so causes nginx to return a 502 error. This error is seen on Integration, but not locally or on Heroku.

A pull request has been raised to fix this by [increasing the cache size in NGINX](https://github.com/alphagov/govuk-puppet/pull/11846).

**Note:**

We will use this feature in the `frontend` app first and review the performance impact before rolling out to other frontend apps.

The increased cache size in NGINX is large enough for the loading of individual CSS assets in the `frontend` app. However, we will need to increase the cache size further before being used in other apps such as `government-frontend` that use a larger number of components, and before it is used to load individual JS assets.

### Step by step changes to the rendering application

**Set up the stylesheets to be precompiled**. Add any app components and views in `app/assets/config/manifest.js` (and then remove them from `app/assets/stylesheets/application.scss`).

```js
//= link components/_calendar.css
//= link views/_homepage.css
```

Add list of gem components to be precompiled in `config/application.rb` (and then all gem component `@import`s can also be removed from `app/assets/stylesheets/application.scss`):

```ruby
#config/application.rb
module Frontend
  class Application < Rails::Application
    include GovukPublishingComponents::AppHelpers::AssetHelper

   ...

    config.assets.precompile << get_component_css_paths
```

#### Using the publishing components gem without the `static` rendering app

The `static` rendering application already includes several components from the publishing components gem. To avoid requesting the same stylesheet twice, we exclude any stylesheets included in `static` by default.

If you are using the individual assets loading feature in an app that does not rely on `static`, we recommend adding the config file below in your application to ensure that the components render correctly.

```rb
# config/initializers/govuk_publishing_components.rb
GovukPublishingComponents.configure do |c|
  c.exclude_css_from_static = false
end
```

**Add the app components and views** to the array. This bit will likely take the most time - and really benefits from a comprehensive list of what pages use which templates (and template variations).

Add the `add_view_stylesheet` and `add_app_component_stylesheet` to the top of each partial template that needs it:

```rb
add_view_stylesheet("homepage")
```

```rb
add_app_component_stylesheet("calendar")
```

Each stylesheet that is used individually will need an extra `@import` added to allow them to use GOV.UK Frontend variables and mixins:

```scss
@import "govuk_publishing_components/individual_component_support";
```

**Hoist the content of the body to the top of the base template in `app/views/layouts/application.html.erb`**. This ensures that all the components are added to the array before the array is used in the `<head>` of the template.

On line 1:

```rb
<% content_for :body do %>
  <!-- Everything that was between <body> and </body> --->
<% end %>
```

And then between `<body>` and `</body>`:

```rb
<%= yield :body %>
```

**Add the stylesheets just before the closing `</head>` tag of the base template**:

```rb
<%= stylesheet_link_tag "application.css", integrity: false %>
<%=
  render_component_stylesheets
%>
```

## Why

See [RFC #149][rfc] for more details about this approach.

## Visual Changes

None.

[rfc]: https://github.com/alphagov/govuk-rfcs/pull/152


lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -0,0 +1,155 @@
...
+      end
+
+      # Some applications have view-specific stylesheets - use this method
+      # to add them to the list
+      def add_view(view)
+        unless is_already_used?(view)
+          @view_stylesheets_being_used << "views/_#{view}.css"
+          @stylesheet_list << view
+        end
+      end
```
kevindew, line 140: These 3 options seem a bit limiting, for example you couldn't have any stylesheets in a nested folder (e.g. view/dir-1/_partial.scss).

I'd imagine you'd do better with a generic method that takes a file path e.g: add_path("views/dir-1/_partial.css") which was called by the other convenience methods e.g.

```
def add_gem_component(component)
  add_path("govuk_publishing_components/components/_#{component}.css")
end
```

lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -0,0 +1,155 @@
...
+          layout-footer
+          layout-for-public
+          layout-header
+          layout-super-navigation-header
+        ]
+      end
+
+      # Used to add a component that exists in the gem to the list
+      def add_gem_component(component)
+        unless is_already_used?(component)
```
kevindew, line 119: This check and subsequent ones looks like it has some problems:

- If there is any cases of a stylesheet having the same name it wouldn't include it. E.g if you had a view named and component sharing the same name the last added would be rejected
- if you had a view that had the same name as any of the reserved ones for static it'd not be included
- if you had an app that didn't use static you'd never be able to use the static components (I imagine it'd be better to configure this with an initialiser to state whether an app uses static)

lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -0,0 +1,155 @@
+module GovukPublishingComponents
+  module AppHelpers
+    class AssetHelper
+      # This is a list of all the components that have stylesheets - so a new
+      # component needs to be added to this list rather than being laboriously
+      # added and removed from every single rendering application.
+      COMPONENT_CSS_PATHS = %w[
```
kevindew, line 7: This constant doesn't seem to be used anywhere?

lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -0,0 +1,155 @@
+module GovukPublishingComponents
+  module AppHelpers
+    class AssetHelper
```
kevindew, line 3: We should have some unit tests for this class

lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -0,0 +1,155 @@
...
+        # Set up the lists of gem components, app components, and views.
+        @gem_component_stylesheets_being_used = []
+        @app_component_stylesheets_being_used = []
+        @view_stylesheets_being_used = []
+
+        # This list includes components already included in Static; taken from
+        # https://github.com/alphagov/static/blob/198a598682df40ce4a2c3c286c06244297c18cf0/app/assets/stylesheets/application.scss
+        #
+        # This is used to dedupe stylesheets without needing the full path.
+        @stylesheet_list = %w[
```
kevindew, line 95: It'd be more consistent for this to be flagged as a class constant.

app/assets/stylesheets/govuk_publishing_components/individual_component_support.scss
```diff
@@ -0,0 +1,10 @@
+$govuk-new-link-styles: true;
```
kevindew, line 1: This filename should have an underscore since it's used as an include rather than an entry point.

kevindew (commented):
Amazing work getting so far with this, I've not done a thorough review but thought I'd drop some quick comments in. I'm also due to be off for a while from Wednesday so please do seek more reviewers and don't wait for me.

I think there's a few things that are missing from this before it's mergable:

- There's no tests for this, I've suggested one place for unit tests but I expect there should be more than that
- There should be a changelog entry.
- The PR description contains really good documentation but I imagine that is something that should be in this gem rather than just in the PR.
- It would have problems with apps that don't use static.
- It'd be good to have this used by the component guide that runs in the dummy app so that it is implicitly used as part of testing of this gem (and can be easily demo'd)

I'd also suggest breaking this up into smaller PR's so it is easier to review, as this is already daunting at 145 files, I imagine you can break up the set-up work from the implementation.

Finally, the use of ApplicationController to initialise this seems unusual. I'd expect you can remove this. I'd imagine you shouldn't need any request time initialisation (this can happen in helpers) and for application level initialisation you should use a Rails initialiser (there is prior art for this in [GovukPublishingComponents::Config](https://github.com/alphagov/govuk_publishing_components/blob/ee732a885b44d7883825fc5cc36fc1179a30d60e/lib/govuk_publishing_components/config.rb) with an example of it as part of an initialiser in [Content Publisher](https://github.com/alphagov/content-publisher/blob/c098d5f38e6194defbaf54b8da9d475f249bc84f/config/initializers/govuk_publishing_components.rb))

Hope this helps

lib/govuk_publishing_components/app_helpers/asset_helper.rb
kevindew: Ah I just saw it in the PR description. It seems surprising that an app would reach in and use this constant. I'd wonder if you can instead make it's usage better through a method, e.g. `GovukPublishingComponents.precompile_individual_component_stylesheet_assets`

You could save having to list all the components by doing something like: `Dir["#{GovukPublishingComponents::Config.gem_directory}/app/assets/stylesheets/govuk_publishing_components/components/_*.scss"]` to generate the list of files.

I also expect it'd be good to have:

```
  if defined? @individual_stylesheets
    list_of_stylesheets = @individual_stylesheets.get_stylesheets.map do | component |
      stylesheet_link_tag(component, integrity: false)
    end
    raw(list_of_stylesheets.join(""))
  end
```

as something that is handled by a method too, that'd allow cleaner app usage and allow the method to be tested.

MartinJJones:
@kevindew thanks for the taking a look at this and all the helpful feedback, will work on this further today 👍

leenagupte (dismissed):
It's not easy to pick up someone else's work and ideas and I think you've done a great job of trying to progress this.

Whilst I agree with this change in principle I have some concerns about how it's being implemented. I've added some inline comments but there are two reasons that I've requested changes on this PR (the scary red X)

1. There are no tests for the helper.
2. The use of the `@individual_stylesheets` global variable that is defined in the calling application.

With regards to point 2, this is big concern. The gem shouldn't have to know or rely on implementation of the caller. Highly coupled code like this will be harder to maintain in the long run. [1]
@kevindew has made some good suggestions for helper methods. I'm also happy to talk in person about potential implementations that are more robust and don't rely on implementation in the calling application. 

[1]: https://www.bridge-global.com/blog/why-are-solid-principles-vital-when-designing-software-in-the-object-oriented-world

MartinJJones:
Thanks for the feedback as well @leenagupte, will be sure to take a further look and make the changes required.

> I'm also happy to talk in person about potential implementations that are more robust and don't rely on implementation in the calling application.

Definitely interested in talking through this further, will reach out to you to arrange a catch-up, thanks

lib/govuk_publishing_components/app_helpers/asset_helper.rb
kevindew: Ah not quite, I meant introduce one method that accepts the relatively low level construct of the full path to an asset and that method has the business logic for dealing with a stylesheet being added.

Then any other methods are just wrappers around the low level method for convenience. 

So you'd end up with something like:

```
def add_gem_component(component)
  add_path("govuk_publishing_components/components/_#{component}.css")
end

def add_path(full_path)
   # all the business logic for adding a stylesheet
end
```

leenagupte: Ah, and the `add_view_component` method can accept a whole path instead and pass this directly to `add_path`. Right. Thanks! Makes sense.

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: > Am I correct in understanding that the add_gem_component, add_component and add_view methods will be used by the rendering applications?

Yes, although I don't think they need to be called in a specific order, will test this further to confirm if this is correct.

> Does that mean that it is expected that the stylesheet for the gem will be the one used even if a view in a rendering application has one with the same name?

I think this would depend on the order in which the component stylesheet paths are added to `@stylesheet_list`, with the current implementation, the below scenario is possible:

- `views/_COMPONENT_NAME.css` is used and `_COMPONENT_NAME` added to`@stylesheet_list` first
- The view template renders a component with the same name, either from the gem or the app itself, `components/_COMPONENT_NAME.css`
- The `is_already_used?` method will stop this component name from being added to `@stylesheet_list`, as it is already in included
```
def is_already_used?(component)
        @stylesheet_list.include?(component)
end
```
- The stylesheet for `components/_COMPONENT_NAME.css` would not be included on the frontend app and the component would not render correctly

Will try and add a test for this scenario.

> Are there any cases where there are stylesheets with the same name?

I haven't come across any examples of this, but will also check this further.

MartinJJones: One possible fix for this is to update the methods so that the full path is added to `@stylesheet_list`.

**From**

      ```Ruby
def add_view(view)
        unless is_already_used?(view)
          @view_stylesheets_being_used << "views/_#{view}.css"
          @stylesheet_list << view
        end
      end
```

**To**

```Ruby
def add_view(view)
        unless is_already_used?(view)
          @view_stylesheets_being_used << "views/_#{view}.css"
          @stylesheet_list << "views/_#{view}.css"
        end
      end
```

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: `@stylesheet_list` should contain a unique list of component stylesheets that are required for the page being rendered. It's used to prevent the same stylesheet being requested multiple times, several buttons on a page, for example.

In the example above, we are pre-populating the `@stylesheet_list` with the stylesheets we know are already included in static. If a stylesheet is missed, it would lead to duplicate code being sent to the frontend, for example `button.css` is included in `static`, we would then add `button.css` again for the `frontend` rendering app.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -1,4 +1,8 @@
 <%
+  if defined? @individual_stylesheets
+    @individual_stylesheets.add_gem_component("accordion")
+  end
```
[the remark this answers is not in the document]
MartinJJones: We have removed the `@individual_stylesheets` instance variable and replaced with a helper as recommended. We were able to do this using [Railtie](https://api.rubyonrails.org/classes/Rails/Railtie.html) to [initialize the AssetHelper](https://github.com/alphagov/govuk_publishing_components/pull/3014/commits/818a1f19718d82b955f655a80060ce0c7fd38fff)

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: Thanks, the methods have been updated so we can now support any file path, for example:

Generic method that takes any file path

```Ruby
      def add_stylesheet_path(component_path)
        unless is_already_used?(component_path)
          STYLESHEET_LIST << component_path
        end
      end
```

Convenience method that calls `add_stylesheet_path`

```Ruby
      def add_gem_component(gem_component)
        add_stylesheet_path("govuk_publishing_components/components/_#{gem_component}.css")
      end
```

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: > - If there is any cases of a stylesheet having the same name it wouldn't include it. E.g if you had a view named and component sharing the same name the last added would be rejected.
> - if you had a view that had the same name as any of the reserved ones for static it'd not be included

This potential issue was fixed by including the path to the stylesheet as well as the stylesheet component name.

app/assets/stylesheets/govuk_publishing_components/individual_component_support.scss
MartinJJones: Thanks, if it's OK, I've captured this point to work on as part of this Trello card - Trello card: https://trello.com/c/OX7Vgl8B/1698-individual-loading-of-assets-documentation-and-sharing

My thinking here is that the name of `individual_component_support` is currently consistent with the naming of `component_support.scss` and `govuk_frontend_support.scss` in the [same directory](https://github.com/alphagov/govuk_publishing_components/tree/main/app/assets/stylesheets/govuk_publishing_components), but we may want to change these file names as well.

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: Thanks, we have refactored this.

We can now use the `get_component_css_paths` method from GovukPublishingComponents to get the list of stylesheets to precompile.

We've also added `render_component_stylesheets` to render the stylesheets in application.html

MartinJJones:
Most of the feedback from the initial code review have been addressed, however we have decided not to use the individual loading of assets feature in the component guide as part of this change.

> It'd be good to have this used by the component guide that runs in the dummy app so that it is implicitly used as part of testing of this gem (and can be easily demo'd)

We can add this change in the future and captured this on a Trello card - https://trello.com/c/OX7Vgl8B/1698-individual-loading-of-assets-documentation-and-sharing

We have feature tests in place to check the individual loading of assets feature is working correctly. Lastly, since this feature is experimental and will not be used in all applications, it will be good to keep the component guide as it is (not using the individual loading of assets feature). The visual regression tests run on the component guide app will help ensure that this change does not cause any side effects to an application not using the feature.

app/assets/stylesheets/govuk_publishing_components/individual_component_support.scss
KludgeKML: I think @kevindew's point here is that this is different from` component_support.scss` and `govuk_frontend_support.scss`, which are both intended to be included in other apps, whereas `individual_component_support.scss` is only included in other stylesheets in this gem and isn't intended to be used outside the gem.

lib/govuk_publishing_components/app_helpers/asset_helper.rb
MartinJJones: Done 👍

app/assets/stylesheets/govuk_publishing_components/individual_component_support.scss
jon-kirwan: Done 👍

KludgeKML:
@MartinJJones I've made a slight edit to the description - the tags were invisible in the rendered markdown, so I've wrapped them in code block backticks to make them show up.

MartinJJones:
> @MartinJJones I've made a slight edit to the description - the tags were invisible in the rendered markdown, so I've wrapped them in code block backticks to make them show up.

@KludgeKML Good spot, thanks for updating the description. I've just [made the same change in the gem documentation](https://github.com/alphagov/govuk_publishing_components/pull/3014/commits/a44321e1370f99be158336e1bfebc5b2419b6194) as well.

KludgeKML (approved):
Looks good to me, I think all the comments from Leena and Kev are now addressed.

lib/govuk_publishing_components/app_helpers/asset_helper.rb
KludgeKML: Now updated to make it clear that these are all stylesheets.

KludgeKML:
@leenagupte I'm going to dismiss your review, it appears to be the only way to mark your requested changes as fixed.

PR #3015, opened 2022-10-13
Replace YAML anchor with list to fix VRT action

injms:
## What

Replace the ignore file list YAML anchor with an actual list.

## Why

GitHub Actions doesn't currently support YAML anchors, so this list of files and paths to ignore needs to be written twice - once for the pushes and once  or the pull requests.

This caused the YAML file to halt the running of the workflow - so visual regression tests weren't being run.

image

## Visual Changes
None.

### DIFF
--- a/.github/workflows/visual-regression-tests.yml
+++ b/.github/workflows/visual-regression-tests.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-    paths-ignore: &ignore-list
+    paths-ignore: # Until anchors are supported by GitHub Actions, we have to write this list twice.
       - ".github/**"
       - "docs/**"
       - ".gitignore"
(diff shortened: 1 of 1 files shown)


.github/workflows/visual-regression-tests.yml
```diff
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - main
-    paths-ignore: &ignore-list
+    paths-ignore: # Until anchors are support by GitHub Actions, we have to write this list twice.
```
AshGDS, line 7: Nitpick, should it be `supported` instead of `support`?

AshGDS (approved):
Looks good, thanks for fixing the issue as I kept seeing it on my branches. I added a small non-blocking comment.

MartinJJones (approved):
Looks good 👍

.github/workflows/visual-regression-tests.yml
injms: Yes, yes it should 🤦

AshGDS: Thanks @injms :+1:

PR #3024, opened 2022-10-17
Add tracking to super navigation header dropdowns

AshGDS:
Hi @andysellick / @JamesCGDS 

Would one of you (or both) be able to review this? Thanks :+1:

## What

- Adds tracking to the dropdown menus of the GOV.UK super navigation header. These are the dropdowns for `Topics`, `Government Activity`, the `Search` button, and the `Menu` button on mobile (the `Menu` button  expands and collapses the whole menu on mobile.)
- This uses `ga4-event-tracker` which listens for a click on all `ga4-event-tracker` modules, then checks for a `data-ga4` data attribute JSON on the `event.target` and pushes it to GTM if it exists.

This is the element in question:

Desktop:
image

Mobile:
image

The GTM object which is output when clicking one of these dropdowns looks like this:
```JSON
{
    "event": "event_data",
    "event_data": {
        "event_name": "select_content",
        "type": "super navigation header",
        "text": "Topics",
        "index": 2,
        "index_total": 4,
        "action": "opened"
    },
    "govuk_gem_version": "31.1.1"
}
```

## Why

- This is a high priority interaction to add. 
- I've had to do a bit of maths (just some addition) to make the `index` and `index_total` conform to what the Performance Analyst's want.  This is why I'm doing things like `"index_total": navigation_links.length + 2` because this makes the indexes correct across the elements. The order is:
    - Mobile 'Menu' dropdown (index 1)
    - Topics dropdown (index 2)
    - Government activity dropdown (index 3)
    - Search button (index 4)
    - The index_total is 4 across all of them

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
@@ -51,8 +51,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       }
 
-      // Ensure it only tracks aria-expanded in an accordion/super navigation header element, instead of in any child of the clicked element
-      if (target.closest('.gem-c-accordion') || target.closest('.gem-c-layout-super-navigation-header')) {
+      /* Ensure it only tracks aria-expanded in an accordion or element with data-ga4-expandable on it. */
+      if (target.closest('.gem-c-accordion') || target.closest('[data-ga4-expandable]')) {
         var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
       }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
@@ -51,8 +51,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       }
 
-      // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
-      if (target.closest('.gem-c-accordion')) {
+      // Ensure it only tracks aria-expanded in an accordion/super navigation header element, instead of in any child of the clicked element
+      if (target.closest('.gem-c-accordion') || target.closest('.gem-c-layout-super-navigation-header')) {
         var ariaExpanded = this.getClosestAttribute(target, 'aria-expanded')
       }
 
--- a/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
@@ -15,7 +15,7 @@
   hide_navigation_menu_text = t("components.layout_super_navigation_header.menu_toggle_label.hide", :label => "navigation")
   show_navigation_menu_text = t("components.layout_super_navigation_header.menu_toggle_label.show", :label => "navigation")
 %>
-<header role="banner" class="gem-c-layout-super-navigation-header" data-module="gem-track-click ga4-event-tracker" data-track-links-only>
+<header role="banner" class="gem-c-layout-super-navigation-header" data-module="gem-track-click ga4-event-tracker" data-track-links-only data-ga4-expandable>
   <div class="gem-c-layout-super-navigation-header__container govuk-width-container govuk-clearfix">
     <div class="gem-c-layout-super-navigation-header__header-logo">
       <a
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
```diff
@@ -51,8 +51,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       }
 
-      // Ensure it only tracks aria-expanded in an accordion element, instead of in any child of the clicked element
-      if (target.closest('.gem-c-accordion')) {
+      // Ensure it only tracks aria-expanded in an accordion/super navigation header element, instead of in any child of the clicked element
+      if (target.closest('.gem-c-accordion') || target.closest('.gem-c-layout-super-navigation-header')) {
```
andysellick, line 55: I did start this, but I'm now concerned that this line is going to be slowly expanded to add further searches for specific classes, which feels brittle and unsustainable. I wonder instead if we could include an attribute on the element itself to trigger this behaviour?

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -217,6 +231,13 @@
         hidden
         id="super-search-menu-toggle"
         type="button"
+       data-ga4="<%= {
```
andysellick, line 234: Is the indentation here slightly wrong?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
```diff
@@ -462,4 +462,49 @@ describe('Google Analytics event tracking', function () {
       expect(window.dataLayer[0]).toEqual(expected)
     })
   })
+
+  describe('doing tracking on the super navigation header', function () {
```
andysellick, line 466: Following on my comment about the `target.closest` line, one benefit would be that we could make this section of testing more generic, and not have to add specific tests for each kind of new thing.

andysellick (commented):
Generally looks good. I'd like us to think some more about the approach and how it can be more future proof.

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -78,6 +78,13 @@
         hidden
         id="super-navigation-menu-toggle"
         type="button"
+        data-ga4="<%= {
+          "event_name": "select_content",
+          "type": "super navigation header",
+          "text": "Menu",
+          "index": 1,
+          "index_total": navigation_links.length + 2
+        }.to_json %>"
```
JamesCGDS, line 87: Would it be possible to add a `section` property here and set it to the `text` value? There's a comment in the implementation guide saying this would be useful.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
AshGDS: Good idea - maybe it checks for the presence of `data-ga4-expandable` ?

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
AshGDS: I think so - this has exposed how reliant I am on `standardx` to make my JS look good - without it I'm sloppy!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
AshGDS: Thanks @andysellick I have used a `data-ga4-expandable` attribute to make tracking `aria-expanded` more generic and not dependent on component classes. I didn't remove the `.gem-c-accordion` check as I didn't want to break the accordion tracking with this branch.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
AshGDS: Thanks @andysellick - I renamed the test to make it more generic, I wasn't sure how to change the test itself to make it more generic but let me know if you have an idea.

AshGDS:
Thanks @JamesCGDS & @andysellick - I've made the requested changes now. I pushed it as a separate commit to make it easier to review, then if everything is OK I'll squash the commit.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.js
andysellick: Great stuff. I've added this to our tech notes to address separately for the accordion.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-event-tracker.spec.js
andysellick: I guess we can do that when we remove the explicit accordion class, so fine for now.

andysellick:
Changes look good, looks like there's one comment from @JamesCGDS still outstanding, if he's not back soon I can take a second look.

PR #3028, opened 2022-10-19
Include the words 'opens in new tab' as part of the share link

jon-kirwan:
## What

Include the words ‘opens in new tab’ as part of the share link

https://trello.com/c/PhMa7my6/1578-include-the-words-opens-in-new-tab-as-part-of-the-share-link

**Review URL**
- See https://components-gem-pr-3028.herokuapp.com/component-guide/share_links

## Why

- The 'Share on...' links open in a new tab without informing the user. This can be disorientating for screen reader users who may not know that the window has opened 
- Also, to follow the Design System guidance here: https://design-system.service.gov.uk/styles/typography/#opening-links-in-a-new-tab

## Before (component only with "Sharing will open the page in a new tab" text)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199773690-ad01526f-8971-4441-a6e4-3d4914f092a6.png"></a>

## After (component only "Sharing will open the page in a new tab" text)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199773681-c8a6ef15-6c45-4512-848c-f4789e9c9b17.png"></a>

## Before (component with "Sharing..." text and title)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199773693-92d2dc23-35f4-4745-9e42-62edd91e2f76.png"></a>

## After (component with "Sharing..." text and title)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199773686-ce48c111-a22c-427e-ab94-68cf8a118e91.png"></a>

## Before (news article page with component)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199776186-d7130707-41f6-43da-b6aa-5839838baaa9.png"></a>

## After (news article page with component)

<a target="_blank" rel="noopener noreferrer" href="https://user-images.githubusercontent.com/87758239/199776174-04eddb36-1798-49ae-8be7-59546ca411f5.png"></a>

### DIFF
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -232,16 +232,16 @@ en:
       navigation_search_heading: Search and popular pages
       navigation_search_subheading: Search
       popular_links:
-      - label: 'Check benefits and financial support you can get'
-        href: '/check-benefits-financial-support'
+      - label: Check benefits and financial support you can get
+        href: "/check-benefits-financial-support"
       - label: 'Limits on energy prices: Energy Price Guarantee'
-        href: '/government/publications/energy-bills-support/energy-bills-support-factsheet-8-september-2022'
-      - label: 'Find a job'
-        href: '/find-a-job'
-      - label: 'Coronavirus (COVID-19)'
-        href: '/coronavirus'
+        href: "/government/publications/energy-bills-support/energy-bills-support-factsheet-8-september-2022"
+      - label: Find a job
+        href: "/find-a-job"
+      - label: Coronavirus (COVID-19)
+        href: "/coronavirus"
       - label: 'Universal Credit account: sign in'
-        href: '/sign-in-universal-credit'
+        href: "/sign-in-universal-credit"
       popular_links_heading: Popular on GOV.UK
       search_text: Search GOV.UK
     metadata:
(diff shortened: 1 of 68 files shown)


maxgds (changes requested):
I think this wants further enhancement. DAC raised it as a screen reader issue but it's an issue for all users that things are opening in a separate place to the default. The design system guidance suggests a warning at the top of a list of links that is visible to all - but you've not added this. The hidden text you've added is available only to screen readers which is good and recommended by the Design System because those users might access the link out of context, but we also need to support sighted users who might otherwise be disoriented. Looking at the component, I think you're going to need design involvement to decide how to present that extra text.

config/locales/en.yml
```diff
@@ -292,21 +292,25 @@ en:
...
+          href: https://ukstandswithukraine.campaign.gov.uk/
+        title: Invasion of Ukraine
       world_locations: World locations
     search_box:
       input_title: Search
       label: Search on GOV.UK
       search_button: Search
+    share_links:
+      all_opens_in_new_tab: Sharing will open the page in a new tab
+      open:
```
maxgds, line 312: All looks good except for this "open" key in the translations which seems to be redundant. If we can remove that we're good to go.

maxgds (changes requested):
One small change needed across the translation files, but otherwise looking good.

maxgds (approved):

PR #3031, opened 2022-10-21
Version 31.2.0

patrickpatrickpatrick:
## 31.2.0

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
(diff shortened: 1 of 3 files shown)


chao-xian:
I think this warrants a minor version rather than a patch bump, as in I think this is a 31.2.0

patrickpatrickpatrick:
> I think this warrants a minor version rather than a patch bump, as in I think this is a 31.2.0

do i need to update branch name or is that fine

chao-xian:
It doesn't matter so up to you.

chao-xian (approved):

PR #3032, opened 2022-10-21
Fix search referrer bug

JamesCGDS:
## What
This PR contains two key changes:

- Fixes a bug where the `referrer` property on the `page_view` object was not updating correctly after dynamic page updates
- Adds a new `dynamic` property to the `page_view` object

## Why
When a user is on a search results page and applies a filter, the URL of the page changes and PA's need to be able to track the URLs a user has visited using the `page_view.referrer` property. In the current implementation, however, the `page_view.referrer` property was not updating and remained the same whenever a dynamic update was triggered. This PR ensures that the `page_view.referrer` property is updated correctly by using a parameter that it receives from [`live_search.js`](https://github.com/alphagov/finder-frontend/blob/e0e8f32000f708b2b2a67983d49f45d9f04c27f8/app/assets/javascripts/live_search.js#L121) in the `finder-frontend` repository.

In addition, PA's need to be able to distinguish between a fresh page load and a dynamic page update and the `dynamic` property has therefore been added to the `page_view` object. As requested by the PA's, it is a stringified boolean.

## Visual Changes
None.

## N.B.
Related PR - https://github.com/alphagov/finder-frontend/pull/2906

As this PR changes how `GOVUK.analyticsGa4.Ga4EcommerceTracker.init()` receives and handles parameters, these changes are breaking.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ## Unreleased
 
 * Fix issue with blue action link arrow svg ([PR #3039](https://github.com/alphagov/govuk_publishing_components/pull/3039))
-
+* **BREAKING:** Fix referrer bug ([PR #2906](https://github.com/alphagov/finder-frontend/pull/2906))
 
 ## 31.2.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@
 ## Unreleased
 
 * Fix issue with blue action link arrow svg ([PR #3039](https://github.com/alphagov/govuk_publishing_components/pull/3039))
-* **BREAKING:** Fix referrer bug ([PR #2906](https://github.com/alphagov/finder-frontend/pull/2906))
+* **BREAKING:** Fix referrer bug ([PR #3032](https://github.com/alphagov/govuk_publishing_components/pull/3032))
 
 ## 31.2.0
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
@@ -9,12 +9,17 @@
     PIIRemover: new GOVUK.analyticsGa4.PIIRemover(),
     DEFAULT_LIST_TITLE: 'Site search results',
 
-    init: function (isNewPageLoad) {
+    init: function (dynamicPageViewData) {
       if (window.dataLayer) {
+        /* The dynamicPageViewData object is only passed to the init() function as a result of an AJAX request
+        (in live_search.js in the finder-frontend repository). Otherwise it will not be available and this indicates a fresh page load.
+        This is needed to trigger a fresh pageViewTracker push to the dataLayer on dynamic page updates (line 30) and to prevent multiple
+        click listeners from being applied (line 42). */
+        var isNewPageLoad = !dynamicPageViewData
+
         /* The data-ga4-ecommerce attribute may be present on several DOM elements e.g. search results and spelling
         suggestions, hence why document.querySelectorAll is required */
         this.searchResultsBlocks = document.querySelectorAll('[data-ga4-ecommerce]')
-        this.isNewPageLoad = isNewPageLoad
 
         if (!this.searchResultsBlocks.length === 0) {
           return
@@ -23,18 +28,18 @@
         /* If the results are updated by JS, the URL of the page will change and this needs to be visible to PA's,
         hence the pageView object push to the dataLayer. We do not need to send a pageView object on page load as
         this is handled elsewhere. */
-        if (!this.isNewPageLoad) {
+        if (dynamicPageViewData) {
           var pageViewTracker = window.GOVUK.analyticsGa4.analyticsModules.PageViewTracker
 
           if (pageViewTracker) {
-            pageViewTracker.init()
+            pageViewTracker.init(dynamicPageViewData)
           }
         }
 
         for (var i = 0; i < this.searchResultsBlocks.length; i++) {
           this.trackSearchResults(this.searchResultsBlocks[i])
 
-          if (this.isNewPageLoad) {
+          if (isNewPageLoad) {
             this.searchResultsBlocks[i].addEventListener('click', this.handleClick.bind(this))
           }
         }
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
@@ -9,13 +9,13 @@
     PIIRemover: new GOVUK.analyticsGa4.PIIRemover(),
     DEFAULT_LIST_TITLE: 'Site search results',
 
-    init: function (dynamicPageViewData) {
+    init: function (referrer) {
       if (window.dataLayer) {
-        /* The dynamicPageViewData object is only passed to the init() function as a result of an AJAX request
+        /* The referrer parameter is only passed to the init() function as a result of an AJAX request
         (in live_search.js in the finder-frontend repository). Otherwise it will not be available and this indicates a fresh page load.
-        This is needed to trigger a fresh pageViewTracker push to the dataLayer on dynamic page updates (line 30) and to prevent multiple
-        click listeners from being applied (line 42). */
-        var isNewPageLoad = !dynamicPageViewData
+        This is needed to trigger a fresh pageViewTracker push to the dataLayer on dynamic page updates and to prevent multiple
+        click listeners from being applied on this.searchResultsBlocks elements. */
+        var isNewPageLoad = !referrer
 
         /* The data-ga4-ecommerce attribute may be present on several DOM elements e.g. search results and spelling
         suggestions, hence why document.querySelectorAll is required */
@@ -28,11 +28,11 @@
         /* If the results are updated by JS, the URL of the page will change and this needs to be visible to PA's,
         hence the pageView object push to the dataLayer. We do not need to send a pageView object on page load as
         this is handled elsewhere. */
-        if (dynamicPageViewData) {
+        if (referrer) {
           var pageViewTracker = window.GOVUK.analyticsGa4.analyticsModules.PageViewTracker
 
           if (pageViewTracker) {
-            pageViewTracker.init(dynamicPageViewData)
+            pageViewTracker.init(referrer)
           }
         }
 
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## 31.2.0
 
 * Feedback component visual updates ([PR #2894](https://github.com/alphagov/govuk_publishing_components/pull/2894))
+* **BREAKING:** Fix referrer bug ([PR #2906](https://github.com/alphagov/finder-frontend/pull/2906))
```
AshGDS, line 13: Your change is in the wrong place

AshGDS (approved):
Looks good to me, there's a small change needed for the CHANGELOG.

CHANGELOG.md
JamesCGDS: My mistake, fixed now

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -9,13 +9,15 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
+    init: function (dynamicPageViewData) {
+      dynamicPageViewData = dynamicPageViewData || {}
+
       if (window.dataLayer) {
         var data = {
           event: 'page_view',
           page_view: {
             location: this.getLocation(),
-            referrer: this.getReferrer(),
+            referrer: dynamicPageViewData.referrer || this.getReferrer(),
```
andysellick, line 20: It looks like we're only interested in the `dynamicPageViewData` value for `referrer`? If that's the case do we need to pass a whole object or just this value?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
```diff
@@ -9,12 +9,17 @@
     PIIRemover: new GOVUK.analyticsGa4.PIIRemover(),
     DEFAULT_LIST_TITLE: 'Site search results',
 
-    init: function (isNewPageLoad) {
+    init: function (dynamicPageViewData) {
       if (window.dataLayer) {
+        /* The dynamicPageViewData object is only passed to the init() function as a result of an AJAX request
+        (in live_search.js in the finder-frontend repository). Otherwise it will not be available and this indicates a fresh page load.
+        This is needed to trigger a fresh pageViewTracker push to the dataLayer on dynamic page updates (line 30) and to prevent multiple
```
andysellick, line 16: This is a helpful comment. My only thought is that having specific line number references may get out of date very quickly, could these instead refer to functions? Might last longer.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
```diff
@@ -310,4 +312,43 @@ describe('Google Tag Manager page view tracking', function () {
...
+
+      expect(window.dataLayer[0]).toEqual(expected)
+    })
+
+    it('when passed a parameter for the virtual_page_view and referrer', function () {
+      expected.page_view.referrer = 'https://gov.uk/referrer'
+      expected.page_view.dynamic = 'true'
+      GOVUK.analyticsGa4.analyticsModules.PageViewTracker.init({
+        referrer: 'https://gov.uk/referrer',
+        dynamic: 'true'
```
andysellick, line 348: You've covered all the possibilities of these two keys existing/not existing very thoroughly here. I'm just curious though - I thought that they're intrinsically linked i.e. if there's a value for `referrer` then `dynamic` is always true, otherwise both are missing. Can you explain the three scenarios being tested here?

docs/analytics-ga4/ga4-ecommerce-tracker.md
```diff
@@ -17,22 +17,24 @@ In `analytics-ga4.js`:
 In `live_search.js` in the `finder-frontend` repository (https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L129):
 
 ```JavaScript
-GOVUK.analyticsGa4.Ga4EcommerceTracker.init(isNewPage) // isNewPage is a boolean and is used to detect whether a new page has been loaded or not
+GOVUK.analyticsGa4.Ga4EcommerceTracker.init(dynamicPageViewData) 
+
+/* The init() function takes an object (dynamicPageViewData) as an optional parameter to detect whether dynamic updates have been made to the page via AJAX requests. The values in the object are then used to determine what is pushed to the dataLayer. If a parameter is not passed, this indicates a fresh page load (e.g. by refreshing the page or using the pagination) and that the page has not been dynamically updated via JS. */
```
andysellick, line 22: This seems like part of the documentation - why is it in a JS code snippet comment?

docs/analytics-ga4/ga4-ecommerce-tracker.md
```diff
@@ -17,22 +17,24 @@ In `analytics-ga4.js`:
...
 ```
 
 Unlike the other GA4 tracking scripts in this repository, ecommerce tracking is not initialised by `init-ga4.js`. Instead, it is initialised by the JS that runs on finder pages (specifically, the `LiveSearch.prototype.Ga4EcommerceTracking()` function). It is called:
 
-On page load - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L55
+On page load (where no parameters are passed) - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L55
 
-And whenever search results are updated - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L121
+And whenever search results are updated (where an object parameter is passed) - https://github.com/alphagov/finder-frontend/blob/main/app/assets/javascripts/live_search.js#L121
```
andysellick, line 29: There's a few links to specific line numbers in github files in this doc, which seems like it'll get out of date quickly (there's a way of making a more permanent link although I can't remember off the top of my head what it is).

As a general rule it would be good if this documentation was standalone even if e.g. github was down. Are these links simply references to 'this is where this happens in the code'? Is there any more explanation you could add here to mean that readers don't always have to follow the links?

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* **BREAKING:** Fix referrer bug ([PR #2906](https://github.com/alphagov/finder-frontend/pull/2906))
```
andysellick, line 12: This is the wrong detail - it should refer to this PR (`https://github.com/alphagov/govuk_publishing_components/pull/3032`) not the one in `finder-frontend`.

andysellick (commented):
Looking good, a few comments/mainly questions to help my understanding. I'd suggest merging the test commit into the code changes commit, as they're directly related (presumably the tests would fail without the code changes, and vice versa).

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
JamesCGDS: A little later in the code I also need to access another property (`dynamicPageViewData.dynamic`) so that the PA’s have something to differentiate dynamic page updates

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
JamesCGDS: That’s very true, I’ve updated the comment

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
JamesCGDS: I think this is sort of linked to the first comment - if we're happy with keeping the code as it is, this means that the `referrer` and `dynamic` properties are being conditionally set by different values on the `dynamicPageViewData` object and I'd just want to be absolutely sure that everything is working as it should in all possible scenarios

docs/analytics-ga4/ga4-ecommerce-tracker.md
JamesCGDS: No reason, just an oversight on my part. I've moved it now

docs/analytics-ga4/ga4-ecommerce-tracker.md
JamesCGDS: Very true. I’ve removed the links now and just put code snippets in as they were just there to show what happens in the code, as you say

CHANGELOG.md
JamesCGDS: My mistake, updated now

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
andysellick: Been wrestling with this over lunch. I think I'm coming around to the thought that this doesn't feel quite right - we're passing an object that contains two values, but only really using one of them (referrer). I'm quite happy to pass an object (it's more future proof, although I'm not aware of anything else we'd need to pass) but the code and tests currently imply that different values might be passed in that object, when in reality there are only two scenarios: nothing is passed, or referrer and dynamic (set to true) are passed.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
andysellick: Yes, that's what I'm getting at. I'll resolve this comment though as the other thread is talking about the same thing.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
JamesCGDS: Cool, thanks for this @andysellick. What I've done is I've changed the implementation slightly so that it's only using a single string parameter now rather than an object. Like you say, I'm not aware of anything further that would need to be passed and, thinking about it, I think passing an object when we only need a single value might not be the most intuitive. If this needs extending in the future we can make it an object but, in the interests of keeping the code more readable in its current form, perhaps just passing the string is best.

AshGDS (approved):
Looks good, good spot on the things I missed @andysellick 

I'd squash the commits once Andy has seen the changes

JamesCGDS:
@AshGDS - cool, will do. Just kept the commits separate for the time being so it was easier to see the changes.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
andysellick: Thanks @JamesCGDS, this feels much tidier 👍

andysellick (approved):
Looks good, just need to squash some commits 👍

PR #2866, opened 2022-07-20
Remove PrimaryLinks JS Module

MartinJJones:
## What
Remove the "PrimaryLinks" JS Module and related tests from govuk-publishing-components, no documentation to remove.

## Why
The "PrimaryLinks" module is not accessible and not widely used across GOVUK.

After searching the content store for the `primary-item` CSS class which is used by the PrimaryLinks module, 28 items were found. All of the 28 items used the "government-frontend" rendering app, however the "government-frontend" app does not initialise the PrimaryLinks module, so has no impact on how these pages are rendered.

### Breakdown of content items found

**CMA content - 24 pages**
The PrimaryLinks module has not worked on these pages for over 5 years, following the migration of "Specialist Frontend" to "Government Frontend".

**DVSA content - 4 pages**
The Primary Links module for these pages stopped working a couple of months ago following the migration of "Manuals Frontend" to "Government Frontend".

## Visual Changes
**Note**: The example below shows the visual difference when the component is initialised to help demonstrate what the PrimaryLinks module does, it is not actually initialised on goverment-frontend.

| Before | After |
| - | - |
| example-showing-primary-module-in-use | live-example-no-primary-link-module |

Trello: https://trello.com/c/nezjGrzF/1388-investigate-and-potentially-fix-primary-links-module

Issue #2780, opened 2022-05-17
Several Instances of HTML in locale files


patrickpatrickpatrick:
## What

Remove all instances of HTML in locale files.

## Why

When I was working on a card to add data-tracking attributes to the copyright and license links in the footer, I noticed that the actual elements I was to add data to were actually in the locale files themselves. I was able to add a listener to the container of the link but to me it seems quite unmaintainable to have essentially 60 copies of the same elements with different text. If a class needed to be changed or an attribute then something could easily be missed. In fact there is already an issue caused by this - the `a` element in copyright_html links to a 404 in every other language apart from in `en.yml`. 

The instances of html in locale files are as follows: 

```
opendocument_html: This file is in an <a href='https://www.gov.uk/guidance/using-open-document-formats-odf-in-your-organisation' target=%{target} class='govuk-link'>OpenDocument</a> format

request_format_details_html: If you use assistive technology (such as a screen reader) and need a version of this document in a more accessible format, please email <a href='mailto:%{alternative_format_contact_email}' target='_blank' class='govuk-link'>%{alternative_format_contact_email}</a>. Please tell us what format you need. It will help us if you say what assistive technology you use.

phase_banner_html: This is a new service – your <a class="govuk-link" href="https://signin.account.gov.uk/contact-us?supportType=PUBLIC">feedback</a> will help us to improve it.

copyright_html: <a class="govuk-footer__link govuk-footer__copyright-logo" href="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/">© Crown copyright</a>

licence_html: All content is available under the <a class="govuk-footer__link" href="https://www.nationalarchives.gov.uk/doc/open-government-licence/version/3/" rel="license">Open Government Licence v3.0</a>, except where otherwise stated
```

It seems to me that most of these have the same format of `prefix` `link` `suffix`, so perhaps an alternative structure for this pattern could be

```
licence_link
    prefix: All content is available under the 
    link_text: Open Government Licence v3.0
    suffix: except where otherwise stated
```

maxgds:
I disagree with the suggested solution of prefix/link text/suffix because grammar would potentially break if you were doing clean exact string translations of sentence fragments. There are some options discussed on stackoverflow here: https://stackoverflow.com/questions/2543936/rails-i18n-translating-text-with-links-inside

If this only affects links we could probably get away with something like:

[...]

PR #3038, opened 2022-10-25
Change "+" to "Show" in related navigation and metadata block components

jon-kirwan:
## What
The "**+ 3 more**" / "**show fewer**" button text from the [related navigation component](https://components.publishing.service.gov.uk/component-guide/related_navigation) and [metadata block component](https://components.publishing.service.gov.uk/component-guide/metadata/extensive_country_list/) is not descriptive enough to identify its function or purpose when navigating out of context.

Change "**+**" to "**Show**".

**Example URLs:**

- https://www.gov.uk/check-uk-visa
- https://www.gov.uk/cymraeg

**Review URL:**

- [related navigation component example - with extensive world locations](https://components-gem-pr-3038.herokuapp.com/component-guide/related_navigation/with_extensive_world_locations/preview)
- [metadata block component example - with extensive country list](https://components-gem-pr-3038.herokuapp.com/component-guide/related_navigation/with_extensive_world_locations/preview)

## Why
WCAG fail and the button text is confusing for users (especially for screen reader users).

## Anything else
- ~~I decided against including any HTML in the locale file based on this [issue: several Instances of HTML in locale files](https://github.com/alphagov/govuk_publishing_components/issues/2780)~~
- A [similar issue - affecting the primaryLinks" JS Module](https://github.com/alphagov/govuk_publishing_components/pull/2866) was previously fixed by removing the disclosure functionality. I've not applied the same fix since the related navigation component is used more widely and similar changes would have a bigger impact. Also, the component does not have the same problems e.g. lack of communication to screen reader users on changes or that the "**+ more**" button is removed when clicked

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Support Rails Content Security Policy nonce on inline JavaScript ([PR #3173](https://github.com/alphagov/govuk_publishing_components/pull/3173))
 * Restyle subscription link ([PR #3177](https://github.com/alphagov/govuk_publishing_components/pull/3177))
+* Change "+" to "Show" in related navigation and metadata block components ([PR #3038](https://github.com/alphagov/govuk_publishing_components/pull/3038))
 
 ## 34.2.0
 
--- a/app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
+++ b/app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
@@ -14,7 +14,9 @@
        data-controls="toggle-<%= toggle_id %>"
        data-expanded="false"
        data-toggled-text="<%= t("components.metadata.toggle_less") %>">
-        <%= t("components.metadata.toggle_more", number: remaining.length) %>
+        <%= t("components.metadata.toggle_more",
+          show: t('components.accordion.show'),
+          number: remaining.length) %>
     </a>
   </div>
   <span id="toggle-<%= toggle_id %>" class="gem-c-metadata__toggle-items js-hidden"><%= remaining.to_sentence.html_safe %></span>
--- a/app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
+++ b/app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
@@ -13,9 +13,9 @@
        class="gem-c-metadata__definition-link govuk-!-display-none-print"
        data-controls="toggle-<%= toggle_id %>"
        data-expanded="false"
-       data-toggled-text="<%= t("components.metadata.toggle_less") %>">
-        <%= t("components.metadata.toggle_more",
-          show: t('components.accordion.show'),
+       data-toggled-text="<%= t("common.toggle_less") %>">
+        <%= t("common.toggle_more",
+          show: t('common.show'),
           number: remaining.length) %>
     </a>
   </div>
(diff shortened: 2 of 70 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 
 * Feedback component visual updates ([PR #2894](https://github.com/alphagov/govuk_publishing_components/pull/2894))
 * Add tracking to super navigation header dropdowns ([PR #3024](https://github.com/alphagov/govuk_publishing_components/pull/3024))
+* Add context to related navigation component "more" and "show fewer" buttons ([PR #3038](https://github.com/alphagov/govuk_publishing_components/pull/3038))
```
maxgds, line 14: This needs to move into an unreleased heading.

spec/components/related_navigation_spec.rb
```diff
@@ -158,6 +158,28 @@ def construct_links(type, base_path, title, document_type = nil, locale = nil)
...
+      { title: "What qualification levels mean", base_path: "/what-different-qualification-levels-mean" },
+      { title: "Find a traineeship", base_path: "/find-traineeship" },
+    ].each do |item|
+      content_item["links"]["ordered_related_items"] << { "title" => item[:title], "base_path" => item[:base_path] }
+    end
+
+    render_component(content_item: content_item)
+
+    assert_select ".gem-c-related-navigation__section-link[href=\"/apply-apprenticeship\"]", text: "Find an apprenticeship"
+    assert_select ".gem-c-related-navigation__toggle", text: /\+\s2 more[.\n+\s]+related content/
```
maxgds, line 178: I have some reservations about the grammar here. This works fine for "two more world locations" but gets a bit weird when it comes to "two more related content". I'd suggest you run it by Anika to see if this is good enough or whether we want to expand these to "two more related content list items" or something similar.

app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
```diff
@@ -14,7 +14,9 @@
        data-controls="toggle-<%= toggle_id %>"
        data-expanded="false"
        data-toggled-text="<%= t("components.metadata.toggle_less") %>">
-        <%= t("components.metadata.toggle_more", number: remaining.length) %>
+        <%= t("components.metadata.toggle_more",
+          show: t('components.accordion.show'),
```
maxgds, line 18: I can see why you've used the existing translation for show, but I think it would be cleaner to give each component its own translation otherwise you're creating a dependency between the two - a dependency makes sense if you're including one in the other but that's not happening here. I know that's going to be a lot of files to update and it will be repetitive but it's the cleaner approach and keeps each component self contained.

maxgds: Actually, perhaps it makes sense to move this translation into "common" and have these two components, plus the accordion pull it from there to avoid repetition but keep it cleaner.

maxgds (changes requested):

app/views/govuk_publishing_components/components/metadata/_sentence.html.erb
jon-kirwan: Thanks, yep that'll be much tidier. I've now moved a few things under `common`:

- `accordion.show` (used by `accordion`, `related_navigation`, `metadata`)
- `accordion.hide` (only used in `accordion` now but I probably best to keep it together with `show`)
- `metadata.toggle_less` (used by `related_navigation`, `metadata`)
- `metadata.toggle_more` (used by `related_navigation`, `metadata`)

maxgds (approved):
Looks good.

jon-kirwan:
> Looks good.

Thank you. I think it might need approval in Percy too if you could also hit approve there?

maxgds:
> > Looks good.
> 
> Thank you. I think it might need approval in Percy too if you could also hit approve there?


Done

PR #3039, opened 2022-10-25
Fix issue with blue action link arrow svg

patrickpatrickpatrick:
## What

Replace the svg that was being used for the blue arrow action link.

## Why

The tail of the blue arrow svg used for action links was misaligned and not in the centre of the caret. It was previously using a different SVG to the other arrows for action link. This was because unlike the other SVGs the arrow for the blue svg is transparent. To fix the styling issue the blue arrow now uses the svg that the other arrows use but has a mask so it remains transparent as it was before.,

### Before

Screenshot 2022-10-25 at 16 28 13

### After

Screenshot 2022-10-25 at 16 28 41


matthillco:
Looks good to me. I don't have Percy permissions though, please request a visual approval from someone with permissions.

hannalaakso (approved):
LGTM 👍

PR #3041, opened 2022-10-27
Prevent unnecessary repetition in code when adding link click tracking

AshGDS:
Hi @andysellick @JamesCGDS,

I've tried to make `ga4-link-tracker` more efficient, would you be able to approve if you agree with my changes? Let me know if you disagree with my proposed change. Also let me know if what I've written doesn't make sense. Thanks :+1:

## What
On `ga4-link-tracker`,  when a link is clicked there are two ways we can build the object needed to be pushed to GTM. These are:

- The JS categorising a link based on the link's destination, and building the GTM object from that, then pushing it up.
- Grabbing a hardcoded GTM JSON from a data-attribute on the `<a>` tag, and then parsing it and pushing it to GTM. This is needed when the link can't be categorised based on its destination. For example, header links need to be categorised based on where they are rendered in the template - it doesn't matter what their link destination is. Therefore the solution was to just hardcode the GTM object they needed onto each link itself. (This is the `data-ga4-link` implementation which is used on the Share/Follow links currently.) This is basically the same as `data-ga4` we use for the event tracking.

The issue with having a hardcoded GTM JSON on each `<a>` tag, is that it is quite inefficient due to how many links can exist on a page. If you have a JSON on every `<a>` tag, that adds up quickly in terms of bytes added to the HTML. There are roughly 116 homepage links (including the header and footer) so if we were to add the header, footer, and homepage link click tracking, it would add 116 JSONs to the homepage HTML as data attributes. Initially I thought this would add a lot of file size to the HTML, but because we have HTML compression on the production server this isn't too massive of an issue. I tested adding 26 individual JSONs to the header and without compression it added 5KB to the HTML. But with server compression it only added 500 bytes to the HTML. So doing some maths, if we have 116 JSONs it might add 2.2KB to the homepage size with compression. Across all of GOV.UK, since there are 26 header links and 25 footer links that would be an extra 1KB added to every page. I'm not sure how big of an issue an extra 1KB is - though I think the newer implementation is more efficient and would have less of an increase in HTML size.

Therefore, I have come up with a solution which I think is cleaner for both the template and JS. Instead of adding the a JSON in the template to each link we do this:

- In the template, tag the links we want to track with a category attribute called `data-ga4-link-category`. For example, `<a href="http://test.com" data-ga4-link-category='follow'>` for tracking a "Follow us" link.
- In the JS, we grab this category string, and then we build the GTM object we need based on the category. (e.g. `if (linkCategory === "follow") [build the correct gtm object] else if (linkCategory === "something else")....`) 
- Most values we need for the object can just be grabbed via JS (such as the url, link text, etc.) but there are a couple values we can only get from the Ruby template - these are `index` and `index_total`. Therefore I have added the index of each link as a data attribute on the link called `data-ga4-index`. This allows us to easily grab what index the link is in a list. I've then added a `data-ga4-index-total` attribute, which is added on the parent of the list of links. This allows us to grab the `index_total` on a list of links without any duplication in the HTML.

## Why

- The current `data-ga4-link` implementation is inefficient in my opinion.
- This new implementation keeps the GTM schema out of the template, so if the schema changes we wouldn't need to change the templates
- There is less JS/template code overall with this implementation.

## Visual Changes
None.


AshGDS:
Superseded by https://github.com/alphagov/govuk_publishing_components/pull/3057

PR #3042, opened 2022-10-27
Track clicks on links with child elements

patrickpatrickpatrick:
## What

Update the


app/assets/javascripts/govuk_publishing_components/analytics/track-click.js
```diff
@@ -30,6 +30,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   GemTrackClick.prototype.handleClick = function (target) {
+    console.log('HELLO')
+
```
jon-kirwan, line 34: Looks like there's a stray console.log message here to be removed.

jon-kirwan (changes requested):
Hi, all looks fine. Just looks like there's a leftover `console.log` to be removed and then i'll approve 👍

app/assets/javascripts/govuk_publishing_components/analytics/track-click.js
patrickpatrickpatrick: ah thanks, will remove that

jon-kirwan (approved):
🚀

PR #3043, opened 2022-10-31
Add track link test

andysellick:
## What / why
Adds a test and small documentation update for https://github.com/alphagov/govuk_publishing_components/pull/3042

## Visual Changes
None.


jon-kirwan (approved):
All seems good to me and have approved 👍 Only thing to mention is if the commits should be combined in to one?

andysellick:
> All seems good to me and have approved 👍 Only thing to mention is if the commits should be combined in to one?

They're unrelated, so I initially thought they should actually go in separate PRs, but as it's just a small thing I thought I'd tack it into this one. If we needed to revert these changes we'd want them to be separate.

PR #3058, opened 2022-11-08
Use one button in navbar

patrickpatrickpatrick:
## What

Change the implementation of the sitewide super navigation header layout component to use one 'Menu' button at all screen sizes with one toggle menu instead of two buttons with two toggle menus. 

Previous implementation was that each of the two toggle buttons would be contained within another toggle dropdown menu on mobile. This has been removed, as there is now only one menu which is not closable.

Increase size of clickable area for links with descriptions to be the height of the link and description.

## Why

Tree testing has shown that users are confused by our current implementation of two buttons on desktop. We have a design proposal that simplifies this for our users and lessen confusion.

[Relevant Trello Card](https://trello.com/c/tsBPaWVM/1403-build-streamlined-menu-iteration-xl)

### Desktop

Screenshot 2022-11-08 at 16 29 41

Screenshot 2022-11-08 at 16 32 56

image

### Mobile

Screenshot 2022-11-08 at 16 33 04

Screenshot 2022-11-08 at 16 33 13

Screenshot 2022-11-08 at 16 33 18

### Desktop

Screenshot 2022-11-08 at 16 30 03

Screenshot 2022-11-08 at 16 30 12

### Mobile

Screenshot 2022-11-08 at 16 38 07

Screenshot 2022-11-08 at 16 30 28

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
@@ -92,32 +92,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     return closestPrevious(previousElement, elementType)
   }
 
-  // When moving from one screen size to another (eg mobile to desktop) we need
-  // to find out whether a submenu is open so the parent menu can be kept open.
-  var hasSubMenusOpen = function ($menu) {
-    return $menu.querySelectorAll('button[aria-expanded="true"]').length > 0
-  }
-
-  // Detect the current viewport width. Return the string `desktop` or `mobile` so it can be interpolated to access the
-  // `data-toggle-{desktop|mobile}-group` attribute.
-  //
-  // Use `matchedia` which mitigates issues in modern browsers when scrollbars are forced in MacOS (using `document.documentElement.clientWidth`
-  // for viewport width detection has issues in Chrome and Firefox, and `window.innerWidth` has issues in Safari).
-  //
-  //  Fall back to `document.documentElement.clientWidth` for legacy browsers.
-  var windowSize = function () {
-    var viewport = false
-    if (typeof window.matchMedia === 'function') {
-      viewport = window.matchMedia('(min-width:' + SETTINGS.breakpoint.desktop + 'px)').matches
-    } else {
-      viewport = document.documentElement.clientWidth >= SETTINGS.breakpoint.desktop
-    }
-    return viewport ? 'desktop' : 'mobile'
-  }
   function SuperNavigationMegaMenu ($module) {
     this.$module = $module
-    this.$navigationToggle = this.$module.querySelector('#super-navigation-menu-toggle')
-    this.$navigationMenu = this.$module.querySelector('#super-navigation-menu')
     this.$searchToggle = this.$module.querySelector('#super-search-menu-toggle')
     this.$searchMenu = this.$module.querySelector('#super-search-menu')
 
@@ -132,55 +108,25 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     )
 
     this.hiddenButtons = this.$module.querySelectorAll('button[hidden]')
-
-    this.lastWindowSize = null
   }
 
-  SuperNavigationMegaMenu.prototype.windowSize = windowSize
-
   SuperNavigationMegaMenu.prototype.updateStates = function () {
-    if (this.windowSize() === 'mobile' && this.lastWindowSize !== 'mobile') {
-      this.$navigationToggle.removeAttribute('hidden')
+    // Check whether any of the dropdown menus are open.
+    var $openButton = this.$module.querySelector('[aria-expanded="true"][data-toggle-desktop-group="top"]')
+    // Gets the open dropdown menu
+    var $openMenu = $openButton ? this.$module.querySelector('#' + $openButton.getAttribute('aria-controls')) : null
+    // If there is an open dropdown menu, get the height of the dropdown menu.
+    var margin = $openMenu ? $openMenu.offsetHeight : 0
 
-      // Hides navigation menu unless a submenu is open - this could be common
-      // as the desktop view has the navigation submenu as top level in the
-      // menu.
-      if (!hasSubMenusOpen(this.$navigationMenu)) {
-        hide(this.$navigationToggle, this.$navigationMenu)
-      }
-
-      this.$module.style.marginBottom = '0'
-    }
-
-    // Hide the navigation toggle button and show the navigation submenu:
-    if (this.windowSize() === 'desktop' && this.lastWindowSize !== 'desktop') {
-      this.$navigationToggle.setAttribute('hidden', 'hidden')
-      this.$navigationMenu.removeAttribute('hidden')
-    }
-
-    // The dropdown menu height is only needed when in desktop mode as this is
-    // when the dropdown menu is using `absolute` positioning. This requires
-    // JavaScript to work out the height and make space for the dropdown menu.
-    if (windowSize() === 'desktop') {
-      // Check whether any of the dropdown menus are open.
-      var $openButton = this.$module.querySelector('[aria-expanded="true"][data-toggle-desktop-group="top"]')
-      // Gets the open dropdown menu
-      var $openMenu = $openButton ? this.$module.querySelector('#' + $openButton.getAttribute('aria-controls')) : null
-      // If there is an open dropdown menu, get the height of the dropdown menu.
-      var margin = $openMenu ? $openMenu.offsetHeight : 0
-
-      // Make space for the dropdown menu.
-      this.$module.style.marginBottom = margin + 'px'
-    }
-
-    this.lastWindowSize = this.windowSize()
+    // Make space for the dropdown menu.
+    this.$module.style.marginBottom = margin + 'px'
   }
 
   SuperNavigationMegaMenu.prototype.buttonHandler = function (event) {
     var $target = closestParentIncluding(event.target, 'button')
     var $targetMenu = this.$module.querySelector('#' + $target.getAttribute('aria-controls'))
 
-    var toggleGroupAttribute = 'data-toggle-' + this.windowSize() + '-group'
+    var toggleGroupAttribute = 'data-toggle-desktop-group'
     var toggleGroupName = $target.getAttribute(toggleGroupAttribute)
     var toggleGroupList = this.$module.querySelectorAll('[' + toggleGroupAttribute + '="' + toggleGroupName + '"]')
 
@@ -194,9 +140,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     toggle($target, $targetMenu)
 
-    if (this.windowSize() === 'desktop') {
-      this.$module.style.marginBottom = $targetMenu.offsetHeight + 'px'
-    }
+    this.$module.style.marginBottom = $targetMenu.offsetHeight + 'px'
   }
 
   SuperNavigationMegaMenu.prototype.init = function () {
@@ -217,6 +161,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       $element.removeAttribute('hidden')
 
       var closestSiblingLink = closestPrevious($element, 'a')
+
       if (closestSiblingLink) {
         closestSiblingLink.setAttribute('hidden', 'hidden')
       }
@@ -236,8 +181,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     //     second level navigation menu
     hide(this.$searchToggle, this.$searchMenu)
     this.updateStates()
-
-    this.lastWindowSize = this.windowSize()
 
     // The menu needs to be updated when the window is resized to make sure that
     // the space needed for the dropdown menu is correct.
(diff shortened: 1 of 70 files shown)


app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -65,109 +65,87 @@
...
+          text_for_hide: hide_menu_text,
+          text_for_show: show_menu_text,
+          toggle_desktop_group: "top",
+          toggle_mobile_group: "top",
+          tracking_key: tracking_label,
+          ga4: {
+            event_name: "select_content",
+            type: "header menu bar",
+            text: link[:label],
+            index: 2,
```
AshGDS, line 109:
```suggestion
            index: 1,
```

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -65,109 +65,87 @@
...
+            index_total: 3,
```
AshGDS, line 110:
```suggestion
            index_total: 2,
```

AshGDS (changes requested):
Hi @patrickpatrickpatrick 

These values will need to be modified so the GA4 tracking is accurate - as there's one less button on the menu now we just need to reduce the indexes by 1.

Thanks :+1:

patrickpatrickpatrick:
> Hi @patrickpatrickpatrick
> 
> These values will need to be modified so the GA4 tracking is accurate - as there's one less button on the menu now we just need to reduce the indexes by 1.
> 
> Thanks 👍

Sorry if this is me not understanding the purpose of the indexes; I thought the reason they were offset originally (started at 2, which is what the buttons are on live) was because they were taking into account the presence of the logo link (which would be link number 1). Is that assumption not correct?

AshGDS:
> > Hi @patrickpatrickpatrick
> > These values will need to be modified so the GA4 tracking is accurate - as there's one less button on the menu now we just need to reduce the indexes by 1.
> > Thanks 👍
> 
> Sorry if this is me not understanding the purpose of the indexes; I thought the reason they were offset originally (started at 2, which is what the buttons are on live) was because they were taking into account the presence of the logo link (which would be link number 1). Is that assumption not correct?

These are for the buttons which expand and collapse. Previously we had:

- The 'Show/Hide menu button' (only visible on mobile) as index 1
- The 'Topics' dropdown as index 2
- The 'Government activity' dropdown as index 3
- The 'Search' button as index 4

But now we have:
- The 'Show/Hide menu button' (both visible on mobile/desktop) as index 1
- The 'Search' button as index 2

Hope that makes sense :+1:

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -119,22 +120,12 @@ $after-button-padding-left: govuk-spacing(5);
...
-    margin: 0 (0 - govuk-spacing(6));
-  }
-
-  @include govuk-media-query($from: "desktop") {
-    background: none;
-    float: left;
-    padding: 0;
-    margin: 0;
-  }
+  background: none;
```
hannalaakso, line 126: Could I check what this style is doing? I'm not seeing a cascading background style from anywhere that we need to override but I might be missing something.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -119,22 +120,12 @@ $after-button-padding-left: govuk-spacing(5);
...
+  float: left;
```
hannalaakso, line 127: Could I check whether it's necessary to float the container here, just because I think it's good to avoid floating things if possible. It looks like the float is used for stopping the items overflowing the nav when the menu is closed but this could be avoided if we didn't set `inline-block` on .`gem-c-layout-super-navigation-header__navigation-items` but kept it as `block` - this could be okay since we only have now one item inside `gem-c-layout-super-navigation-header__navigation-items` so we don't need to use `inline-block` to align the items anymore, at least on mobile. Here's what I mean but I might be missing a reason why this won't work:

Screenshot 2022-11-24 at 18 11 30

hannalaakso (commented):
Looks great when testing the component locally, not spotted any issues 🙌 I've started to review the code and left some comments, will come back to this later.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -295,102 +304,90 @@ $after-button-padding-left: govuk-spacing(5);
     padding: govuk-spacing(3) 0;
   }
 
+  @include govuk-media-query($until: "desktop") {
+    position: absolute;
+    right: 34px;
```
hannalaakso, line 309: I do wonder whether instead of using absolutely positioning elements and having to set their exact locations in pixels we could align the logo, the menu button and the search button using `display: inline-block`. Currently they're positioned with a combination of floats and absolutely positioning and magicky numbers that could potentially be a bit brittle. However the existing navbar already does this so I guess not something we need to fix this in this PR.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -637,45 +696,54 @@ $after-button-padding-left: govuk-spacing(5);
...
 }
 
-.gem-c-layout-super-navigation-header__navigation-second-items--topics {
-  @include govuk-media-query($from: "desktop") {
-    @include columns($items: 17, $columns: 2, $selector: "li", $flow: column);
+.gem-c-layout-super-navigation-header__column--government-activity {
+  position: relative;
+
+  @include govuk-media-query($until: "desktop") {
+    &:after {
```
hannalaakso, line 731: Only a small thing but maybe we could leave a comment to explain that the pseudo element is there to only (?) add a border, it took me a little while to work out. Alternatively it would be more consistent with how we do dividers on the [homepage](https://www.gov.uk/) above "Featured" etc. if the divider was added as a top border on the heading, I guess this would require some toggling of a special class on `.gem-c-layout-super-navigation-header__column-header`. However a class might be more explicit in its function than a pseudo element (it would be neat if the [heading with border component](https://components.publishing.service.gov.uk/component-guide/heading/with_border/preview) admitted specific border styles, I've just raised this in a separate channel).

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -765,5 +840,15 @@ $after-button-padding-left: govuk-spacing(5);
...
   }
+
+  @include govuk-media-query($from: "desktop") {
+    margin: govuk-spacing(5) auto 0;
+    max-width: 960px;
+  }
+}
+
+.gem-c-layout-super-navigation-header__column-header {
+  font-size: 24px;
```
hannalaakso, line 853: If the class is only adding the font size we could consider setting it with `.govuk-heading-m`

patrickpatrickpatrick:
have removed all the absolutely positioned styling, everything uses display inline-block now

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
patrickpatrickpatrick: ah yep, i can actually totally remove that and use a border instead (not sure why i didn't do this before, i think there was a redesign where this line was going to change to a vertical one on desktop)

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
patrickpatrickpatrick: i think the reason i did it this way is that govuk-heading classes have responsive margins

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -65,140 +65,92 @@
...
-          "index_total": navigation_links.length + 2,
-          "section": "Menu"
-        }.to_json %>"
-      >
-        <span class="gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner">
-          Menu
+
+      <%
+        link = t("components.layout_super_navigation_header.navigation_link")
+        unique_id = SecureRandom.hex(4)
```
matthillco, line 71: What's the purpose of this unique number appended to the component ID? It would seem to be unnecessary, as there's only one menu and one button, so the original ID without this number would be sufficient?

patrickpatrickpatrick: This is here just because in the [preview of the component there are multiple variations](https://components.publishing.service.gov.uk/component-guide//layout_super_navigation_header/preview).

hannalaakso (commented):
I've been doing some testing and overall looks really good but I noticed a couple of issues:

- The underline should span the width of the word menu and the arrow
Screenshot 2022-12-08 at 11 44 58
- The underline should stay visible when the menu is open
Screenshot 2022-12-08 at 11 47 43
- The layout becomes misaligned at about 300% zooming on Mac Safari, this doesn't happen in the current component. Could be one to look at later rather than blocking the PR now.
Screenshot 2022-12-08 at 16 16 02
- If JavaScript is enabled and the stylesheet fails to load, there is a lot of margin inserted by JS which pushes the page content down. The margin by JS wasn't introduced by this PR, it already existed on the component but I'm not seeing the issue in the screenshot in the current component. A real edge case but would be good to at least record for later to look at. 
Screenshot 2022-12-08 at 14 28 17

patrickpatrickpatrick:
> I've been doing some testing and overall looks really good but I noticed a couple of issues:
> 
> * The underline should span the width of the word menu and the arrow
>   Screenshot 2022-12-08 at 11 44 58
> * The underline should stay visible when the menu is open
>   Screenshot 2022-12-08 at 11 47 43
> * The layout becomes misaligned at about 300% zooming on Mac Safari, this doesn't happen in the current component. Could be one to look at later rather than blocking the PR now.
>   Screenshot 2022-12-08 at 16 16 02
> * If JavaScript is enabled and the stylesheet fails to load, there is a lot of margin inserted by JS which pushes the page content down. The margin by JS wasn't introduced by this PR, it already existed on the component but I'm not seeing the issue in the screenshot in the current component. A real edge case but would be good to at least record for later to look at.
>   Screenshot 2022-12-08 at 14 28 17

these should all be fixed now

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -65,140 +65,93 @@
...
+          label: show_menu_text,
+        },
+        class: "gem-c-layout-super-navigation-header__navigation-top-toggle-button",
+        data: {
+          text_for_hide: hide_menu_text,
+          text_for_show: show_menu_text,
+          toggle_desktop_group: "top",
+          toggle_mobile_group: "top",
+          tracking_key: tracking_label,
+          ga4: {
```
AshGDS, line 105: Please change this from `ga4` to `ga4_event` as we have changed the attribute name

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -233,12 +184,12 @@
         hidden
         id="super-search-menu-toggle"
         type="button"
-        data-ga4-event="<%= {
+        data-ga4="<%= {
```
AshGDS, line 187: Please change this from `ga4` to `ga4_event` as we have changed the attribute name

AshGDS (changes requested):

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -92,32 +92,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
-    } else {
-      viewport = document.documentElement.clientWidth >= SETTINGS.breakpoint.desktop
-    }
-    return viewport ? 'desktop' : 'mobile'
-  }
   function SuperNavigationMegaMenu ($module) {
     this.$module = $module
-    this.$navigationToggle = this.$module.querySelector('#super-navigation-menu-toggle')
-    this.$navigationMenu = this.$module.querySelector('#super-navigation-menu')
+    this.$navigationToggle = this.$module.querySelector('#super-navigation-menu')
```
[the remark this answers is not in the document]
patrickpatrickpatrick: removed the variable

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
patrickpatrickpatrick: have changed this

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
patrickpatrickpatrick: have changed this

AshGDS (approved):
Thanks @patrickpatrickpatrick

chao-xian (approved):

PR #3067, opened 2022-11-15
Add instructions for upgrading Rails

MuriloDalRi:
https://trello.com/c/LZctyIYq/3007-automatically-add-extra-instructions-to-dependabot-prs-for-rails-1

### DIFF
--- a/docs/pull_request_template.md
+++ b/docs/pull_request_template.md
@@ -16,3 +16,5 @@
 
 
 ### After
+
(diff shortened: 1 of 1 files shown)


MuriloDalRi:
> I don't think you want that in this repo as it's a gem that supports multiple Rails versions. I don't think the steps listed would be very helpful and could be rather misleading.

Thanks Kevin, will remove this one!

PR #3068, opened 2022-11-15
Add aria attributes parameter to the button component

dnkrj:
## What
The button component currently only allows an aria label, no other aria tags.
This commit merges in additional aria attributes passed to the helper.

## Why
There is currently no way to indicate the a button controls another area of the page or is described by another section of the page. This commit merges in additional aria attributes passed to the helper.


andysellick:
@dnkrj are you aware of https://github.com/alphagov/govuk_publishing_components/pull/3088 ? Looks like you're both trying to do the same thing.

dnkrj:
Thanks Andy,
Jonny picked up that ticket from me when I was on second line, have closed
my draft now.

PR #3071, opened 2022-11-16
Set attributes for single page notification button based on Account API response

hannalaakso:
Co-authored with @hannako 

## What
There is a tight coupling between the single page notification button component and the Account API which this PR unpicks. The account API is [returning the full button HTML ](https://github.com/alphagov/account-api/blob/8706f606fa41a5ac44fa602f3da8482d3b0ee9f6/app/controllers/personalisation/check_email_subscription_controller.rb#L31-L46) when there is a govuk_account session present. This makes it hard to extend the functionality of the component as any changes would need to be duplicated in account api.

## How
This PR modifies the JavaScript that receives the API response to now only update the necessary button attributes instead of expecting to receive the entire button HTML and replacing the whole button. Specifically, the PR:
- Updates the tracking attribute on the button HTML according to the value of `active` in the API response (`active` indicates whether the user has subscribed to the relevant notifications)
- Appends the value of  `data-button-location` , if it is provided in the markup, to the above tracking attribute value 
- Updates the button text if both `data-button-text-unsubscribe` and `data-button-text-subscribe` are provided in the markup attributes
- Adds a new wrapper class `.gem-c-single-page-notication-button__text` to more reliably target the button text
- Adds tests to check whether the tracking attribute, including optional location, and the custom button text are updated correctly upon receiving the API response. Removes some no longer relevant test classes (`.old-button-for-test` and `.new-button-for-test`) since the whole button is no longer being replaced. 
- Adds an additional test for checking that the button is rendered visible when the API call completes (the button is initially hidden when JS runs)

## Why 
We currently receive the entire notification button HTML from the Account API to replace the notification button markup rendered on page load - [this is done to make the button reflect whether the user has subscribed to notifications ](https://github.com/alphagov/govuk_publishing_components/pull/2443) for the particular topic or document collection. 

This PR is in preparation for some later work to refactor the API. Once this PR has been merged, and we are no longer using the `button_location` or `button_html` from the [API response](https://github.com/alphagov/account-api/blob/444c22665d171954b61ebeabbfa22dd64bd01af1/app/controllers/personalisation/check_email_subscription_controller.rb#L31-L39), we will remove them from the API and only return the `active` flag. 

## A note about testing
Both @hannako and I have tested the button to work correctly by running it locally via [goverment-frontend](https://github.com/alphagov/government-frontend/pull/2535) with a GOV.UK Account.

## Left to do later
Once the API response has been refactored on the backend to not return the entire button HTML, we should also remove [passing the button location to the Account API in the JS](https://github.com/alphagov/govuk_publishing_components/blob/8ef92af0084403dd9c31142321f359c15ec96f5b/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js#L14) and [the corresponding test for it](https://github.com/alphagov/govuk_publishing_components/blob/8ef92af0084403dd9c31142321f359c15ec96f5b/spec/javascripts/components/single-page-notification-button-spec.js#L28).

## Visual changes
None

### HTML for subscribe button
 Screenshot 2022-11-18 at 14 18 16

### HTML for unsubscribe button

Screenshot 2022-11-18 at 14 17 37

Trello https://trello.com/c/qjGJ7b5P/1507-set-tracking-attribute-and-location-for-the-single-page-notification-button-component-based-on-simplified-api-response

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
@@ -28,15 +28,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       if (xhr.readyState === 4) {
         if (xhr.status === 200) {
           var responseText = xhr.responseText
-          // if response text exists and is JSON parse-able, parse the response and get the button html
+          // if response text exists and is JSON parse-able, parse the response and update the button html
           if (responseText && this.responseIsJSON(responseText)) {
-            var newButton = JSON.parse(responseText).button_html
-            var html = document.createElement('div')
-            html.innerHTML = newButton
-            // test that the html returned contains the button component; if yes, swap the button for the updated version
-            var responseButtonContainer = html.querySelector('form.gem-c-single-page-notification-button')
-            if (responseButtonContainer) {
-              this.$module.parentNode.replaceChild(responseButtonContainer, this.$module)
+            var active = JSON.parse(responseText).active
+
+            var customSubscribeText = this.$module.getAttribute('data-button-text-subscribe')
+            var customUnsubscribeText = this.$module.getAttribute('data-button-text-unsubscribe')
+            // Only set custom button text if both text items are provided
+            var customText = customSubscribeText && customUnsubscribeText
+
+            // Append '-[button-location]' to the tracking data attribute value if data-button-location is set
+            var optionalButtonLocation = this.$module.getAttribute('data-button-location') ? '-' + this.$module.getAttribute('data-button-location') : ''
+
+            // If response returns active, user has subscribed to notifications
+            if (active === true) {
+              this.$module.setAttribute('data-track-action', 'Unsubscribe-button' + optionalButtonLocation)
+
+              if (customText) {
+                this.$module.querySelector('.gem-c-single-page-notication-button__text').textContent = customUnsubscribeText
+              }
+            } else {
+              this.$module.setAttribute('data-track-action', 'Subscribe-button' + optionalButtonLocation)
+
+              if (customText) {
+                this.$module.querySelector('.gem-c-single-page-notication-button__text').textContent = customSubscribeText
+              }
             }
           }
         }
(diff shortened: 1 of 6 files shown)


hannako:
Superseded by https://github.com/alphagov/govuk_publishing_components/pull/3073

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -28,15 +28,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+
+            // Append '-[button-location]' to the tracking data attribute value if data-button-location is set
+            var optionalButtonLocation = this.$module.getAttribute('data-button-location') ? '-' + this.$module.getAttribute('data-button-location') : ''
+
+            // If response returns active, user has subscribed to notifications
+            if (active === true) {
+              this.$module.setAttribute('data-track-action', 'Unsubscribe-button' + optionalButtonLocation)
+
+              if (customText) {
+                this.$module.querySelector('.gem-c-single-page-notication-button__text').innerHTML = customUnsubscribeText
```
matthillco, line 48: Consider using `textContent` rather than `innerHTML` as there's no longer any HTML involved in this assignment. (Could be left as is, although `textContent` does render more quickly.)

app/assets/javascripts/govuk_publishing_components/components/single-page-notification-button.js
```diff
@@ -28,15 +28,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+              }
+            } else {
+              this.$module.setAttribute('data-track-action', 'Subscribe-button' + optionalButtonLocation)
+
+              if (customText) {
+                this.$module.querySelector('.gem-c-single-page-notication-button__text').innerHTML = customSubscribeText
```
matthillco, line 54: Consider using `textContent` rather than `innerHTML` as there's no longer any HTML involved in this assignment. (Could be left as is, although `textContent` does render more quickly.)

PR #3074, opened 2022-11-18
Update link for alpha phase to avoid redirect

36degrees:
## What

The URL `/service-manual/phases/ideal-alphas` refers to [a page from the original Service Manual][1] before the content was migrated in 2016.

Since the content was migrated, that URL now redirects to `/service-manual/agile-delivery/how-the-alpha-phase-works`.

Update the link in the alpha banner to point to the same place.

[1]: https://github.com/alphagov/government-service-design-manual/blob/master/service-manual/phases/ideal-alphas.md

## Why

Avoids an unnecessary redirect.


andysellick (approved):
Thanks, good spot 👍

PR #3075, opened 2022-11-21
Remove the redundant region role from the cookie banner

matthillco:
## What
The cookie banner is currently nested in two very similar regions:
```
<div id="global-cookie-message" role="region" aria-label="cookie banner" [...]>
  <div class="govuk-cookie-banner js-banner-wrapper" role="region" aria-label="Cookies on GOV.UK">
```
That means there is a region announced as “cookie banner” which includes the another region called “Cookies on GOV.UK”.

## Why
This is confusing to screen reader users. Landmarks should be used sparingly and reflect the structure of the page. This structure should only be announced as one region.

## How
By removing the `region` and `aria-label` from the outer element, we retain the full functionality of the component without needing to change the structure itself. Now only one region will be announced.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Update component auditing following print style changes ([PR #3128](https://github.com/alphagov/govuk_publishing_components/pull/3128))
+* Remove the redundant region role from the cookie banner ([PR #3075](https://github.com/alphagov/govuk_publishing_components/pull/3075))
 
 ## 34.1.0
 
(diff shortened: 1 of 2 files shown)


hannalaakso (commented):
This change makes sense to me as it aligns ours with the Design System component 👍 However there's an issue in that now the confirmation message `.gem-c-cookie-banner__confirmation` isn't wrapped inside a container that has the region and aria-label so the confirmation doesn't get announced to assistive technologies. This was already case on the existing code but since we removed the duplicate properties in this PR, we'll need to move `.gem-c-cookie-banner__confirmation` inside `.gem-c-cookie-banner`, that might require some small style or JavaScript changes as well. There's an example of multiple messages inside the component markup [here](https://design-system.service.gov.uk/components/cookie-banner/server-side-multiple-messages-confirmation-visible/index.html).

hannalaakso (approved):
We're going with a simpler fix to remove the duplicated attributes for now and I've captured the above mentioned tech debt that has resulted from the component diverging from the Design System component [here](https://trello.com/c/j82uodb1/189-align-cookie-banner-with-govuk-design-system-cookie-banner).

PR #3081, opened 2022-11-24
Extend the `input` and `textarea` components to use the `dir` attribute

davidtrussler:
## What
Extends input and textarea components to accept a value for the `dir` attribute

## Why
In a previous piece of work (see[ Add RTL support for attachments](https://trello.com/c/RT5RbtmT/896-add-rtl-support-for-attachments)) we added support for right-to-left languages to the attachments pages on the Whitehall publishing app. This was done by adding a class to the parent of the `input` and `textarea` elements that would be used by CSS to render the text correctly. It became clear that it would be a more elegant and semantic solution to achieve this by use of the `dir` attribute of the elements. In addition this is recommended practice for instances where CSS is not loaded. 

Currently these components do not accept a value for that attribute so the changes here update these components accordingly. 

As part of this work it was also necessary to update the hint, label and error message components so that the correct value for the `dir` element can also be displayed on those elements. Another concern is to enable the element (`input` or `textarea`) and its corresponding `hint`, `label` and `error messages` to be able to render with different values for the `dir` attribute. This is for the specific use of the component in the Whitehall publishing app where a language is specified for a document. 

To this end the input and textarea components are considered with the following in mind: 
- the `dir` attribute is not set by default
- the `dir` attribute is set to the same value on both the `input`/`textarea` elements and their associated hint, label and error messages by default via the `right_to_left` boolean attribute of the component
- the `dir` attribute of the hint, label and error messages can optionally be set separately via the `right_to_left_field` boolean (for `input` or `textarea`) and `right_to_left_help` boolean (for hint, label and error messages) attributes of the component.

### Updated components

|element|example|
|---|---|
|hint|Screenshot 2023-01-05 at 12 31 18|
|label|Screenshot 2023-01-05 at 12 44 42|
|error message|Screenshot 2023-01-05 at 12 50 42|
|error items|Screenshot 2023-01-05 at 12 51 43|
|input|Screenshot 2023-01-05 at 14 16 26|
|input with separate `dir` attributes|Screenshot 2023-01-05 at 14 22 16|
|textarea|Screenshot 2023-01-05 at 14 18 19|
|textarea with separate `dir` attributes|Screenshot 2023-01-05 at 14 19 09|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Support Rails Content Security Policy nonce on inline JavaScript ([PR #3173](https://github.com/alphagov/govuk_publishing_components/pull/3173))
 * Restyle subscription link ([PR #3177](https://github.com/alphagov/govuk_publishing_components/pull/3177))
 * Change "+" to "Show" in related navigation and metadata block components ([PR #3038](https://github.com/alphagov/govuk_publishing_components/pull/3038))
+* Extend the `input` and `textarea` components to use the `dir` attribute ([PR #3081](https://github.com/alphagov/govuk_publishing_components/pull/3081))
 
 ## 34.2.0
 
--- a/app/views/govuk_publishing_components/components/_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_input.html.erb
@@ -31,6 +31,8 @@
   heading_level ||= nil
   prefix ||= nil
   suffix ||= nil
+  right_to_left ||= false
+  right_to_left_help = right_to_left_help.nil? ? right_to_left : right_to_left_help
 
   css_classes = %w(gem-c-input govuk-input)
   css_classes << "govuk-input--error" if has_error
@@ -70,7 +72,8 @@
     <% label_markup = capture do %>
       <%= render "govuk_publishing_components/components/label", {
         html_for: id,
-        heading_size: heading_size
+        heading_size: heading_size,
+        right_to_left: right_to_left_help
       }.merge(label.symbolize_keys) %>
     <% end %>
 
@@ -86,7 +89,8 @@
   <% if hint %>
     <%= render "govuk_publishing_components/components/hint", {
       id: hint_id,
-      text: hint
+      text: hint,
+      right_to_left: right_to_left_help
     } %>
   <% end %>
 
@@ -95,6 +99,7 @@
       id: error_id,
       text: error_message,
       items: error_items,
+      right_to_left: right_to_left_help
     } %>
   <% end %>
 
@@ -107,6 +112,7 @@
     autofocus: autofocus,
     class: css_classes,
     data: data,
+    dir: right_to_left ? "rtl" : nil,
     enterkeyhint: checked_enterkeyhint,
     id: id,
     inputmode: inputmode,
--- a/app/views/govuk_publishing_components/components/_label.html.erb
+++ b/app/views/govuk_publishing_components/components/_label.html.erb
@@ -7,6 +7,7 @@
   bold ||= false
   heading_size = false unless shared_helper.valid_heading_size?(heading_size)
   is_page_heading ||= false
+  right_to_left ||= false
 
   css_classes = %w[gem-c-label govuk-label]
   css_classes << "govuk-label--s" if bold
@@ -19,14 +20,14 @@
 
 <% if is_page_heading %>
   <%= tag.h1 text, class: "govuk-label-wrapper" do %>
-    <%= tag.label text, id: id, for: html_for, class: css_classes %>
+    <%= tag.label text, id: id, for: html_for, class: css_classes, dir: right_to_left ? "rtl" : nil %>
   <% end %>
 <% else %>
-  <%= tag.label text, id: id, for: html_for, class: css_classes %>
+  <%= tag.label text, id: id, for: html_for, class: css_classes, dir: right_to_left ? "rtl" : nil %>
 <% end %>
 
 <% if hint_text.present? %>
-  <%= tag.span id: hint_id, class: hint_text_css_classes do %>
+  <%= tag.div id: hint_id, class: hint_text_css_classes, dir: right_to_left ? "rtl" : nil do %>
     <%= hint_text %>
   <% end %>
 <% end %>
--- a/app/views/govuk_publishing_components/components/_textarea.html.erb
+++ b/app/views/govuk_publishing_components/components/_textarea.html.erb
@@ -17,6 +17,8 @@
   hint_id = "hint-#{SecureRandom.hex(4)}"
   has_error ||= error_message || error_items.any?
   error_id = "error-#{SecureRandom.hex(4)}"
+  right_to_left ||= false
+  right_to_left_help = right_to_left_help.nil? ? right_to_left : right_to_left_help
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
@@ -39,13 +41,17 @@
 
 <%= content_tag :div, class: form_group_css_classes do %>
   <% if label %>
-    <%= render "govuk_publishing_components/components/label", { html_for: id }.merge(label.symbolize_keys) %>
+    <%= render "govuk_publishing_components/components/label", {
+      html_for: id,
+      right_to_left: right_to_left_help
+    }.merge(label.symbolize_keys) %>
   <% end %>
 
   <% if hint %>
     <%= render "govuk_publishing_components/components/hint", {
       id: hint_id,
-      text: hint
+      text: hint,
+      right_to_left: right_to_left_help
     } %>
   <% end %>
 
@@ -54,11 +60,13 @@
       id: error_id,
       text: error_message,
       items: error_items,
+      right_to_left: right_to_left_help
     } %>
   <% end %>
 
   <%= tag.textarea name: name,
     class: css_classes,
+    dir: right_to_left ? "rtl" : nil,
     id: id,
     rows: rows,
     maxlength: maxlength,
--- a/app/views/govuk_publishing_components/components/docs/error_message.yml
+++ b/app/views/govuk_publishing_components/components/docs/error_message.yml
@@ -21,3 +21,16 @@ examples:
       items:
         - text: Error 1
         - text: Error 2
+  with_dir_attribute:
+    description: Allows the correct display of right to left languages
+    data:
+      text: "An error message displayed right to left"
+      id: "error_id_2"
+      right_to_left: true
+  with_items_and_dir_attribute:
+    description: To allow the correct display of right to left languages on error items
+    data:
+      right_to_left: true
+      items:
+        - text: Error 1 displayed right to left
+        - text: Error 2 displayed right to left
--- a/app/views/govuk_publishing_components/components/docs/input.yml
+++ b/app/views/govuk_publishing_components/components/docs/input.yml
@@ -183,3 +183,35 @@ examples:
         text: "Given name"
       name: given-name
       enterkeyhint: "next"
+  with_dir_attribute:
+    description: |
+      Allows the correct display of right to left languages.
+
+      By default the input element and the label both display text in a left to right direction.
+
+      When the `right_to_left` parameter of the input component is set to `true` both the input and the label, hint and error message display their content in a right to left direction. 
+    data:
+      label:
+        text: "Some input text to be displayed right to left with a label that displays in the same direction"
+      hint: "Some hint text that displays in the same text direction as the label"
+      error_message: "An error message that displays in the same text direction as the label"
+      name: rtl-input-text
+      value: "العربيَّة"
+      right_to_left: true
+
+  with_separate_dir_attributes_for_field_and_help_text:
+    description: |
+      Allows the correct display of right to left languages.
+
+      By default the input element and the label both display text in a left to right direction.
+
+      If the input field and the help text (label, hint and error messages) are required to display in different directions the `right_to_left_help` attribute can be set as false to override the `right_to_left` attribute.
+    data:
+      label:
+        text: "Some input text that displays right to left with a label that displays left to right"
+      hint: "Some hint text that displays in the same text direction as the label"
+      name: rtl-input-text
+      value: "العربيَّة"
+      right_to_left: true
+      right_to_left_help: false
+      error_message: "An error message that displays in the same text direction as the label"
--- a/app/views/govuk_publishing_components/components/docs/label.yml
+++ b/app/views/govuk_publishing_components/components/docs/label.yml
@@ -65,3 +65,14 @@ examples:
       text: "Label with an ID"
       id: "id-for-the-label"
       html_for: "id-that-matches-input-6"
+  with_dir_attribute:
+    description: |
+      To allow the correct display of right to left languages.
+
+      When the `right_to_left` parameter is set to `true` any hint text displays in the same text direction as the label.
+    data:
+      text: "العربيَّة"
+      html_for: "id-that-matches-input-7"
+      hint_id: "should-match-aria-describedby-input"
+      hint_text: "Hint text displayed right to left"
+      right_to_left: true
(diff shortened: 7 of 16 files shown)


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Extends input and textarea components to accept a value for the dir attribute ([PR #3081](https://github.com/alphagov/govuk_publishing_components/pull/3081))
```
andysellick, line 12: Nitpick: can this match the PR name.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -31,6 +31,8 @@
   heading_level ||= nil
   prefix ||= nil
   suffix ||= nil
+  text_direction_rtl ||= false
+  text_direction_label_ltr ||= false
```
andysellick, line 35: It might just be me, but reading this with one option referring to `rtl` and the other being `ltr` is a bit confusing. Is there a reason they're not both `_rtl` or both `_ltr`?

app/views/govuk_publishing_components/components/docs/input.yml
```diff
@@ -183,3 +183,17 @@ examples:
         text: "Given name"
       name: given-name
       enterkeyhint: "next"
+  with_dir_attribute:
+    description: |
+      To allow the correct display of right to left languages.
+
+      When `text_direction_rtl` is set to true both the input and the label are displayed right to left.
+
+      When `text_direction_label_ltr` is set to true the textarea is displayed right to left and the label is displayed left to right
```
andysellick, line 192: Again this might just be me but I'm finding this a bit confusing. These two options both control the same things, but one overrides the other. I see that you want one flag to set everything to right to left and one flag to override that for the label - can you explain the use case for this? Would it be simpler to have one option for the label and one for the input?

There should also be a test that checks this behaviour as it doesn't look like it behaves as described. If I only set `text_direction_label_ltr` everything is displayed left to right - the textarea is not displayed right to left as described here.

app/views/govuk_publishing_components/components/_textarea.html.erb
```diff
@@ -17,6 +17,8 @@
   hint_id = "hint-#{SecureRandom.hex(4)}"
   has_error ||= error_message || error_items.any?
   error_id = "error-#{SecureRandom.hex(4)}"
+  text_direction_rtl ||= false
+  text_direction_label_ltr ||= false
```
andysellick, line 21: General suggestion - could the variable names used be clearer? Maybe use a general format like `text_direction_[element]_[direction]` e.g. `text_direction_input_right ||= false`. Later variables like `dir_textarea` could also use this convention for consistency. Distinguishing between `ltr` and `rtl` compared to `left` and `right` at a glance is harder.

andysellick (commented):
Looks like a great addition but the code is a little confusing and I think could do with some more tests in place.

davidtrussler:
@andysellick Thanks for the comments. In retrospect I think you may be right that the approach of defaulting both the `label` and the corresponding element to the same thing may cause more confusion than it solves. I'll think about that aspect further but my original thoughts were that for them to be the same would be the expected scenario and to set them differently would be an exception. But maybe that's not really the case.

kevindew (commented):
Nice job putting this together. 

I've got some suggestions on the syntax of calling this that I think could improve it (I hope I'm not contradicting any of your thoughts @andysellick, these are just loosely held opinions). I imagine there's some value in thrashing the syntax out a bit outside of implentation, since it's hard to change later.

I imagine we don't really need the full name `text_direction_rtl` for a boolean. I imagine we could go for `rtl` or `right_to_left` if we want a boolean, or dir if we accept a string (could be more flexible).

So I'd imagine if we did used `right_to_left` we could have:

```
  name: "input",
  right_to_left: true
} %>
```

Result: We'd get a right to left input, no label.

If we did: 

```
  label: {
    text: "My label"
  },
  name: "input",
  right_to_left: true
} %>
```

We'd get a right-to-left input with a right-to-left label.

And if we did:

```
  label: {
    text: "My label",
    right_to_left: false
  },
  name: "input",
  right_to_left: true
} %>
```

We'd get a right-to-left input with a left-to-right label.

Equivalents with dir:

```
  name: "input",
  dir: "rtl"
} %>
```

We'd get a right to left input, no label.
 
```
  label: {
    text: "My label"
  },
  name: "input",
  dir: "rtl"
} %>
```

We'd get a right-to-left input with a right-to-left label.

```
  label: {
    text: "My label",
    dir: "ltr"
  },
  name: "input",
  dir: "rtl"
} %>
```

We'd get a right-to-left input with a left-to-right label.

andysellick:
@kevindew niiiice. Yeah, go with that, that's much clearer.

I am still really interested to know what scenario would require a `rtl` input but a `ltr` label - can anyone enlighten me?

kevindew:
thanks, what do you think is better, a right to left Boolean or a text dir input?

And yes, I can understand that being surprising. It’s for publishing applications as the labels are expected to be in English.

So if you’re producing an Arabic translation of a document in Whitehall we want the user to input right to left, but the label of the field is right to left.

A thought did just occur to me about what to do about hints/error messages. As I guess they’ll remain left to right whatever is passed in hmm.

andysellick:
Ah that makes sense now, thanks.

I think the `right_to_left` boolean applied at the right point in the components options is clearest. That does depend on every component that's likely to need this to accept options in the same structured way (which might not be the case) but that's fixable.

I think this work is quite exciting - it could set a pattern for rolling out this functionality to any component that could need it in the future. Haven't thought much about the details, but I wonder if there's a way of making it easy to apply to any other components in the future? (might be a code thing, could be as simple as writing a short guide)

davidtrussler:
@andysellick I've updated this in line with the comments form yourself and @kevindew. 

I haven't done anything WRT the comment about other stuff like hints and error messages but the more I think about it the more I think that ought to be in here. It's not necessary for the purpose intended originally for these changes but it feels only half done without them now. Do you have any thoughts?

app/views/govuk_publishing_components/components/_label.html.erb
```diff
@@ -15,14 +16,16 @@
...
   <%= tag.h1 text, class: "govuk-label-wrapper" do %>
-    <%= tag.label text, id: id, for: html_for, class: css_classes %>
+    <%= tag.label text, id: id, for: html_for, class: css_classes, dir: dir %>
   <% end %>
 <% else %>
-  <%= tag.label text, id: id, for: html_for, class: css_classes %>
+  <%= tag.label text, id: id, for: html_for, class: css_classes, dir: dir %>
 <% end %>
 
 <% if hint_text.present? %>
```
kevindew, line 31: I guess this hint text should also follow the right_to_left option.

kevindew (commented):
@davidtrussler this looks great, thanks for digging into this further. I think you made the right choice adding it to form hint and error. 

What I'm now not totally sure on though is the best syntax to use to manage it. I think having label imply it is a bit surprising, since you wouldn't expect the label nested arguments to affect things not nested. I had a little play around and found it was probably easiest to allow fields to have two different right_to_left attributes, one for the field and one for the help. Both default to the standard right_to_left attribute if they're not set. 

This idea seemed hard to explain so I thought it was easiest communicated through a diff: https://github.com/alphagov/govuk_publishing_components/compare/7b5a7255b2c76b9592bce735a6dbeff40b89ae5a...rtl-kevin

Happy to have a chat over it

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -63,14 +64,19 @@
     type = "text"
     inputmode = "numeric"
   end
+
+  dir = right_to_left ? "rtl" : nil
 %>
 
 <%= content_tag :div, class: form_group_css_classes do %>
   <% if label %>
+    <% label_rtl = label[:right_to_left] == false ? label[:right_to_left] : right_to_left %>
```
andysellick, line 73: I think I understand this logic - if the label rtl is set to false, then use that, otherwise use the 'general' right_to_left value, is that right? It might be fractionally easier to read as shown below, but I notice that this logic repeats for the hint and error - we might have reached the tipping point where the amount of logic in the view justifies putting some of this code into a helper file, where code comments and a dedicated helper test would help clarify the functionality.

```
label_rtl = label[:right_to_left] == false ? false : right_to_left
```

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -84,17 +90,23 @@
...
     <%= render "govuk_publishing_components/components/hint", {
       id: hint_id,
-      text: hint
+      text: hint,
+      right_to_left: hint_rtl
     } %>
   <% end %>
 
   <% if has_error %>
+    <% error_rtl = label_rtl.nil? ? right_to_left : label_rtl %>
```
andysellick, line 103: Asking mainly for my own understanding (it's early, my brain's still waking up) - the logic for hint and error differs from the logic from label? What's the difference there?

app/views/govuk_publishing_components/components/_input.html.erb
andysellick: Since this functionality is being applied to more than one component and could apply to others in future, it might be something to add to the shared helper.

app/views/govuk_publishing_components/components/docs/error_message.yml
```diff
@@ -15,9 +15,22 @@ examples:
     data:
       text: "Please enter your National Insurance Number"
       id: "error_id"
+  with_dir_attribute:
+    description: Error items are a common pattern where a collection of error is passed with each item having a text attribute of the error
```
andysellick, line 19: This description and the one on the `with_items_and_dir_attribute` example are the same, and don't describe what is being demonstrated.

andysellick (commented):
A few more thoughts and questions. Also can some of these commits be combined? Ideally having the overall change to each component in a single commit would be good (i.e. change to template, test and documentation).

davidtrussler:
> @davidtrussler this looks great, thanks for digging into this further. I think you made the right choice adding it to form hint and error.
> 
> What I'm now not totally sure on though is the best syntax to use to manage it. I think having label imply it is a bit surprising, since you wouldn't expect the label nested arguments to affect things not nested. I had a little play around and found it was probably easiest to allow fields to have two different right_to_left attributes, one for the field and one for the help. Both default to the standard right_to_left attribute if they're not set.
> 
> This idea seemed hard to explain so I thought it was easiest communicated through a diff: [7b5a725...rtl-kevin](https://github.com/alphagov/govuk_publishing_components/compare/7b5a7255b2c76b9592bce735a6dbeff40b89ae5a...rtl-kevin)
> 
> Happy to have a chat over it

@kevindew I've updated the input component in line with what you have suggested (in [this commit](https://github.com/alphagov/govuk_publishing_components/pull/3081/commits/01ddd227120041da1ec007f5ff9e95059548d735)), though I've changed that a bit to hopefully be pretty clear to the user. This gives two new variations on the component documentation as below. If that looks OK I'll extend it to the textarea component. 

|Single right-to-left attribute|Multiple right-to-left attributes|
|-|-|
|Screenshot 2022-12-22 at 12 58 20|Screenshot 2022-12-22 at 12 59 02|

kevindew:
Thanks @davidtrussler - that looks good to me 👍

app/views/govuk_publishing_components/components/docs/error_message.yml
davidtrussler: Oops, that was a bit of lazy copy and paste. Fixed now 👍

app/views/govuk_publishing_components/components/_input.html.erb
davidtrussler: This is now outdated and improved by the later changes along the lines suggested by Kevin.

app/views/govuk_publishing_components/components/_input.html.erb
davidtrussler: This is now outdated and improved by the later changes along the lines suggested by Kevin.

app/views/govuk_publishing_components/components/_label.html.erb
davidtrussler: Yes, good point, have added that. 👍

app/assets/stylesheets/govuk_publishing_components/components/_hint.scss
```diff
@@ -1 +1,5 @@
 @import "govuk/components/hint/hint";
+
+.govuk-hint {
+  display: block;
```
andysellick, line 4: Is this needed? `.govuk-hint` is a div element, which is block by default.

app/views/govuk_publishing_components/components/docs/label.yml
```diff
@@ -65,3 +65,12 @@ examples:
       text: "Label with an ID"
       id: "id-for-the-label"
       html_for: "id-that-matches-input-6"
+  with_dir_attribute:
+    description: |
+      To allow the correct display of right to left languages.
```
andysellick, line 70: Maybe add something like `Dir also applies to label hint, as shown.`?

app/views/govuk_publishing_components/components/docs/error_message.yml
```diff
@@ -15,9 +15,22 @@ examples:
     data:
       text: "Please enter your National Insurance Number"
       id: "error_id"
+  with_dir_attribute:
```
andysellick, line 18: Nitpick: I'd put this example beneath the `with_items` example, as `with_items` feels like more of a prominent use case.

app/views/govuk_publishing_components/components/_input.html.erb
```diff
@@ -107,6 +113,7 @@
     autofocus: autofocus,
     class: css_classes,
     data: data,
+    dir: right_to_left_field ? "rtl" : nil,
```
andysellick, line 116: I might not be fully understanding this, so please feel free to correct me - there are three rtl options: the normal one, `field` and `help`. `help` is applied to the label, hint and error message, and `field` is applied to the input itself. But the normal one isn't applied directly to any elements, it's just used to determine the others, is that right?

If that's correct, could `right_to_left` not be directly applied to the input, and `field` removed? 

```Ruby
right_to_left ||= false
right_to_left_help = right_to_left_help.nil? ? right_to_left : right_to_left_help

# label: right_to_left_help
# hint: right_to_left_help
# error: right_to_left_help
# input: right_to_left ? "rtl" : nil
```

andysellick (commented):
Looking much better, good job. Just a few minor things and one thing that is probably me not understanding correctly.

app/assets/stylesheets/govuk_publishing_components/components/_hint.scss
davidtrussler: That's interesting, actually when the `.govuk-hint` class is used in the label component (though not in the other classes) it is a span. If it is an inline element the text floats to the left and looks wrong (which was the reason for adding it). I've changed it to only apply where it is used in the label component.

app/views/govuk_publishing_components/components/docs/label.yml
davidtrussler: Yes, good point. Have added that. 👍

app/views/govuk_publishing_components/components/docs/error_message.yml
davidtrussler: Sounds like a fair point. Have changed that. 👍

app/views/govuk_publishing_components/components/_input.html.erb
davidtrussler: Hmm, yes, that makes sense. And will apply equally to the textarea component. I've updated the input and textarea components to reflect this with changes to the example documentation, the html and the test.

app/assets/stylesheets/govuk_publishing_components/components/_hint.scss
davidtrussler: Yes, I agree, that would be the better solution. I've updated that to a div and removed the addition to the CSS.

davidtrussler:
> Looking much better, good job. Just a few minor things and one thing that is probably me not understanding correctly.

Thanks @andysellick - I think I've covered the points you've raised if you could give it a look over 😄

andysellick (approved):
Nice work. I also like the clean commit structure.

I was wondering if this is needed again in the future whether it would be possible to implement some kind of shared code for adding this to components, but I think that's out of scope for now.

PR #3082, opened 2022-11-24
Simplify the way ga4 tracking is added to accordions

AshGDS:
Hi @andysellick / @JamesCGDS 

Would you be able to review this? Thanks 👍 

## What
Previously we were passing a GTM object to each accordion's render statement inside of a frontend app's template. See an example here:
 https://github.com/alphagov/collections/blob/2304bf53592f39e35270413fdb27c00c7cb3190a/app/views/world_wide_taxons/accordion.html.erb#L31 

Instead, it's simpler to just pass a boolean - `ga4_tracking: true` - in the accordion render of these frontend apps, and then we handle the GTM object in one place inside of `govuk_publishing_components` instead.

We were also using the accordion's class to check wether we tracked `aria-expanded`. We are now using a data attribute for this, called `data-ga4-expandable`

This isn't a breaking change as the old code is still in the repo, as that code was used to add some new UA tracking so we have to leave it in.

## Why

- This gives us the benefit of only having to maintain our GTM object in one repo, and we won't have to modify templates in multiple frontend apps if our GTM schema changes.
- Reduces technical debt
- Now only uses `data-ga4-` attributes which is better.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/accordion.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/accordion.js
@@ -43,9 +43,10 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
 
     // look for data attributes to put onto the 'show/hide all' link
     var showAllAttributes = this.$module.getAttribute('data-show-all-attributes')
+    var showAll
     if (showAllAttributes) {
       try {
-        var showAll = this.$module.querySelector(this.showAllControls)
+        showAll = this.$module.querySelector(this.showAllControls)
         var values = JSON.parse(showAllAttributes)
         var keys = Object.keys(values)
         for (var i = 0; i < keys.length; i++) {
@@ -54,6 +55,16 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
       } catch (e) {
         console.error('Could not read accordion data attributes error: ' + e.message, window.location)
       }
+    }
+
+    // if GA4 is enabled, set attributes on 'show all sections' for tracking using ga4-event-tracker
+    var dataModule = this.$module.getAttribute('data-module')
+    var isGa4Enabled = dataModule ? dataModule.indexOf('ga4-event-tracker') !== -1 : false
+    if (isGa4Enabled) {
+      var indexTotal = this.$module.querySelectorAll('.govuk-accordion__section').length
+      var showAllAttributesGa4 = { event_name: 'select_content', type: 'accordion', index: 0, index_total: indexTotal }
+      showAll = this.$module.querySelector(this.showAllControls)
+      showAll.setAttribute('data-ga4-event', JSON.stringify(showAllAttributesGa4))
     }
   }
 
--- a/app/views/govuk_publishing_components/components/_accordion.html.erb
+++ b/app/views/govuk_publishing_components/components/_accordion.html.erb
@@ -23,10 +23,12 @@
 
   data_attributes ||= {}
   ((data_attributes[:module] ||= "") << " " << "govuk-accordion gem-accordion").strip!
+  ga4_tracking ||= false
+  data_attributes[:module] << " ga4-event-tracker" if ga4_tracking
+  data_attributes[:ga4_expandable] = '' if ga4_tracking
   data_attributes[:anchor_navigation] = anchor_navigation
   data_attributes[:track_show_all_clicks] = track_show_all_clicks
   data_attributes[:track_sections] = track_sections
-
   data_attributes_show_all ||= nil
   data_attributes[:show_all_attributes] = data_attributes_show_all if data_attributes_show_all
 
@@ -56,6 +58,17 @@
         index = i + 1
 
         item[:data_attributes] ||= nil
+
+        if ga4_tracking
+          item[:data_attributes] ||= {}
+          item[:data_attributes][:ga4_event] = {
+            event_name: "select_content",
+            type: "accordion",
+            text: item[:heading][:text],
+            index: index,
+            index_total: items.length
+          }.to_json
+        end
 
         section_classes = %w(govuk-accordion__section)
         section_classes << 'govuk-accordion__section--expanded' if item[:expanded]
--- a/app/views/govuk_publishing_components/components/docs/accordion.yml
+++ b/app/views/govuk_publishing_components/components/docs/accordion.yml
@@ -181,21 +181,33 @@ examples:
                     <a class="govuk-link" href="#">Retiring your service</a>
                   </li>
               </ul>'
+  with_ga4_tracking:
+    description: |
+      Allows you to add GA4 tracking to an accordion. This will add a data module and data-attributes with JSONs to the accordion. See the [ga4-event-tracker documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/ga4-event-tracker.md) for more information.
+    data:
+      ga4_tracking: true
+      items:
+        - heading:
+            text: Writing well for the web
+          content:
+            html: <p class="govuk-body">This is the content for Writing well for the web.</p>
+        - heading:
+            text: Writing well for specialists
+          content:
+            html: <p class="govuk-body">This is the content for Writing well for specialists.</p>
   with_data_attributes:
     description: |
       Adds custom data attributes to each section of the accordion. Accepts a hash, so multiple attributes can be added.
 
       The `data_attributes` option applies attributes to the outermost element in the accordion. Each item can also have a `data_attributes` hash, which are placed on the `button` that triggers the opening and closing - useful for differentiating between each section of the accordion.
 
-      Data attributes can be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created). More details on how this can be used with the GA4 event tracking can be found in the 'Advanced' section of the [event tracking documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/ga4-event-tracker.md).
+      Data attributes can be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created). Currently, this is only used for tracking with Universal Analytics.
 
       If `track_options` within `data_attributes_show_all` is set, then it is possible to pass a custom dimension when 'Show/Hide all' is clicked.
     data:
       data_attributes:
         custom_data_attr: custom-data-attr-accordion
       data_attributes_show_all:
-        custom_data_attr-event-name: example
-        custom_data_attr-attributes: "{ 'ui': { 'type': 'type value', 'section': 'section value' } }"
         tracking-options: "{ 'dimension114': 1 }"
       items:
         - heading:
--- a/docs/analytics-ga4/ga4-event-tracker.md
+++ b/docs/analytics-ga4/ga4-event-tracker.md
@@ -46,21 +46,12 @@ Many components already accept data attributes in this way (see the [component g
 
 ### Overview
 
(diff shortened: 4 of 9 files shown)


docs/analytics-ga4/ga4-event-tracker.md
```diff
@@ -72,16 +63,14 @@ In some situations we want to track dynamic elements like the 'Show/hide all sec
...
+- Enable tracking for an accordion using the `ga4_tracking` option
+  - In the ruby template, `data-show-all-attributes` is appended to the main component element with the GA4 JSON
 - the 'Show/hide all' link is created by govuk-frontend JavaScript
 - the accordion JavaScript reads `data-show-all-attributes`
-  - it creates a `data-` attribute on the 'Show/hide all' link for each key in the JSON
-  - in the example above, the result will be `data-ga4="{ "event_name": "select_content", "type": "accordion" }"`
-- wrap the accordion in the event tracking script
-  - tracking will be activated by clicks on elements with a `data-ga4` attribute
-  - it checks for an `aria-expanded` attribute if the clicked element or parent element contains a `data-ga4-expandable` value or the `gem-c-accordion` class. If it detects one of those, it sets the `action` of the GA data accordingly.
+  - it moves the `data-show-all-attributes` JSON to the 'Show/hide all' link, renaming the attribute to `data-ga4-event`
```
andysellick, line 70: Struggling a bit this afternoon post-lunch so I could be completely wrong about this, but checking in the component guide I can't see any attribute on the show/hide all link anymore. Has something broken or am I missing something obvious?

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -181,21 +181,49 @@ examples:
...
+            html: <p class="govuk-body">This is the content for Writing well for specialists.</p>
+          data_attributes:
+            custom_data_attr: custom-data-attr-accordion-item-2
+        - heading:
+            text: Know your audience
+          content:
+            html: <p class="govuk-body">This is the content for Know your audience.</p>
+          data_attributes:
+            custom_data_attr: custom-data-attr-accordion-item-3
+        - heading:
```
andysellick, line 208: Nitpick: not sure you need this much data to demonstrate this example, a couple of steps would be fine.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -23,10 +23,12 @@
 
   data_attributes ||= {}
   ((data_attributes[:module] ||= "") << " " << "govuk-accordion gem-accordion").strip!
+  ga4_tracking ||= false
+  data_attributes[:module] << " ga4-event-tracker" if ga4_tracking
+  data_attributes[:ga4_expandable] = '' if ga4_tracking
```
andysellick, line 28: I think there should be a test in the `.spec` file to check that these attributes are added correctly, along with the individual attributes added for each step further down.

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -45,6 +47,14 @@
     end
   end
 
+  if ga4_tracking
+    data_attributes[:show_all_attributes] = {
```
andysellick, line 51: Does this need a bit more in the JS test to check that these attributes get copied onto the show/hide all button correctly? (I haven't checked, there might already be something from the previous iteration, don't know if it would need updating or not).

andysellick (commented):
My brain's gone a bit foggy this afternoon so this might not be particularly thorough or coherent, but looks good, like the approach, this will definitely save us some effort - just a few comments.

docs/analytics-ga4/ga4-event-tracker.md
AshGDS: Good spot - I am getting confused as well. It looks like the initial implementation of `show_all_attributes` takes a JSON containing JSONs, and this would apply each JSON as a `data-` attribute to the Show All link. However we just need to apply the one JSON directly on the data attribute now.

I've also just done a git blame and realised the `data-show-all-attributes` isn't needed for UA - you introduced it specifically for GA4. I got confused because the example wasn't in our GTM schema, and I think that's just because you wrote the code for this before we had a schema, so I thought the examples were UA stuff. (See [here](https://github.com/alphagov/govuk_publishing_components/commit/f193ef04feec77e79ca666d474e0221c6fd03ba9#diff-0353e1304e7acef271c38d2ac799bf23412749d788d92d9e3e8fab8b6b02a2bbR197))

Therefore I should be able to rename the attribute to `data-ga4-show-all-attributes`

docs/analytics-ga4/ga4-event-tracker.md
```diff
@@ -72,16 +63,14 @@ In some situations we want to track dynamic elements like the 'Show/hide all sec
...
+  - in the example above, the result will be `data-ga4-event="{ "event_name": "select_content", "type": "accordion", index: "0", index_total: "0" }"`
+  - tracking will be activated by clicks on elements with a `data-ga4-event` attribute
+  - it checks for an `aria-expanded` attribute, as the accordion will contains a `data-ga4-expandable` value. This sets the `action` of the GA data accordingly.
```
JamesCGDS, line 73: Minor typo - "as the accordion will contain a `data-ga4-expandable` value"

Nitpick - would it be possible to say "This sets the `action` property of the `event_data` object accordingly."? I think this comes down to personal preference though so happy to leave it as it is if you prefer.

docs/analytics-ga4/ga4-event-tracker.md
AshGDS: I'm getting confused about wether I can rename `data_show_all_attributes` in the accordions, but I think I figured it out.

Initially I thought `data_show_all_attributes`  was being used for UA, so I kept it in place. Then I saw that you introduced `data_show_all_attributes` specifically for GA4, so then I thought I could remove it. But it looks like another team have used this GA4 `data_show_all_attributes` attribute to add tracking to UA, so now I can't remove it.

@andysellick the code to add attributes to the Show All attributes link takes a JSON containing JSONs, so I'll have to update the template so I can either:
- Update the template so that it passes a JSON containing a JSON to `data-show-all-attributes` (this feels like adding technical debt)
- Duplicate the JavaScript used for`data-show-all-attributes` code, and make a specific one for GA4 with the attribute name `data-ga4-show-all-attributes` (I think this is preferable)

docs/analytics-ga4/ga4-event-tracker.md
AshGDS: Thanks @JamesCGDS, I've updated the docs

docs/analytics-ga4/ga4-event-tracker.md
AshGDS: @andysellick I think i've resolved this, see the latest commits...

app/views/govuk_publishing_components/components/docs/accordion.yml
AshGDS: Thanks @andysellick - good point. I've made it 2 steps.

app/views/govuk_publishing_components/components/_accordion.html.erb
AshGDS: Thanks @andysellick - there's now a spec test for the module, `data-ga4-expandable` and `data-ga4-show-all-attributes` and the `data-ga4-event` on the 1 accordion item in the test.

AshGDS:
Thanks @andysellick @JamesCGDS - I think I've resolved the changes you requested, feel free to re-review when you have a chance. Will squash the commits once everything is approved :+1:

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -55,6 +56,12 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
         console.error('Could not read accordion data attributes error: ' + e.message, window.location)
       }
     }
+
+    var showAllAttributesGa4 = this.$module.getAttribute('data-ga4-show-all-attributes')
+    if (showAllAttributesGa4) {
+      showAll = this.$module.querySelector(this.showAllControls)
+      showAll.setAttribute('data-ga4-event', showAllAttributesGa4)
```
andysellick, line 63: Instead of passing these attributes to the component and then moving them onto the showall button, couldn't we just hardcode them? Something like this:

```Javascript
// look for data attributes to put onto the 'show/hide all' link
var showAll = this.$module.querySelector(this.showAllControls)
var showAllAttributes = this.$module.getAttribute('data-show-all-attributes')
if (showAllAttributes) {
  try {
    var values = JSON.parse(showAllAttributes)
    var keys = Object.keys(values)
    for (var i = 0; i < keys.length; i++) {
      showAll.setAttribute('data-' + keys[i], values[keys[i]])
    }
  } catch (e) {
    console.error('Could not read accordion data attributes error: ' + e.message, window.location)
  }
}

// if GA4 is enabled, set attributes on 'show all sections' for tracking using ga4-event-tracker
var isGa4Enabled = this.$module.hasAttribute('data-ga4-expandable')
if (isGa4Enabled) {
  var indexTotal = this.$module.querySelectorAll('.govuk-accordion__section').length
  var showAllAttributesGa4 = {"event_name":"select_content","type":"accordion","index":0,"index_total": indexTotal}
  showAll.setAttribute('data-ga4-event', JSON.stringify(showAllAttributesGa4))
}
```

AshGDS: Thanks @andysellick - I've added this change and then will move it to `ga4-core.js` once the link tracker PR is merged.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -55,6 +56,16 @@ window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
         console.error('Could not read accordion data attributes error: ' + e.message, window.location)
```
JamesCGDS, line 56: This code isn't part of the PR but I just noticed it and wanted to check - is this meant to be here?

andysellick: Yes, it is. In the event that incorrect data is passed to the component we want an error to happen in the console.

Ideally it would be noticed by the developer implementing it, but we also have stuff to monitor this that should help us identify configuration problems like this.

JamesCGDS (approved):
Looks good to me 👍

app/views/govuk_publishing_components/components/docs/accordion.yml
```diff
@@ -181,21 +181,33 @@ examples:
...
 
       The `data_attributes` option applies attributes to the outermost element in the accordion. Each item can also have a `data_attributes` hash, which are placed on the `button` that triggers the opening and closing - useful for differentiating between each section of the accordion.
 
-      Data attributes can be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created). More details on how this can be used with the GA4 event tracking can be found in the 'Advanced' section of the [event tracking documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/ga4-event-tracker.md).
+      Data attributes can be added to the 'Show/hide all' link using the `data_attributes_show_all` option, primarily where custom tracking is required. These attributes are read from the accordion markup and then added to the link by JavaScript (which is how the link is created).
```
andysellick, line 204: I think this now only applies (or is needed) for UA instead of GA4? Is it worth clarifying this here?

andysellick (commented):
Nice work. Just two minor documentation comments then I think we're good to go 👍

app/assets/javascripts/govuk_publishing_components/components/accordion.js
andysellick: Does this still need moving to `ga4-core.js`? I can't think what else will need this right now. Maybe move it later when we need it, shouldn't have to block this PR.

AshGDS: This is due to `ga4-core` not being initialised in `govuk_publishing_components`, so the accordion would crash on the developer frontend which showcases the component. Therefore we decided to share the function at a later date.

Issue #3097, opened 2022-12-01
Intervention banner does not support "dismiss"


chao-xian:
The intervention banner can render an SVG `(x)` which implies it can be dismissed, but there is no code to support that functionality.

See https://components.publishing.service.gov.uk/component-guide/intervention

PR #3101, opened 2022-12-02
Fix intervention dismiss

chao-xian:
## What

The dismiss feature was never working on the intervention button. This PR adds that missing functionality.

See https://components-gem-pr-3101.herokuapp.com/component-guide/intervention/with_dismiss_link/preview

Fixes https://github.com/alphagov/govuk_publishing_components/issues/3097

[Trello](https://trello.com/c/MIwoWClW/1514-bug-fix-dismiss-button-on-intervention-banner-component-for-research-panel-recruitment-banner-m)

Fixes https://github.com/alphagov/govuk_publishing_components/issues/3097

### Before

https://user-images.githubusercontent.com/424772/211383759-d25f1045-3ea3-4966-9afb-07ec278a6366.mov

### After

https://user-images.githubusercontent.com/424772/211383925-994944de-7097-4503-8d8c-0abc75fea583.mov

### Cookies

https://user-images.githubusercontent.com/424772/219225646-e03a8db1-c64c-4cbc-a0ef-56d12d46a043.mov

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/intervention.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/intervention.js
@@ -1,0 +1,28 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Intervention ($module) {
+    this.$module = $module
+    this.$banner = this.$module.querySelector('.gem-c-intervention')
+    this.$closeLink = this.$module.querySelector('.gem-c-intervention__dismiss-link')
+  }
+
+  Intervention.prototype.init = function () {
+    this.$module.close = this.handleClose.bind(this)
+
+    if (this.$closeLink) {
+      this.$closeLink.addEventListener('click', this.$module.close)
+    }
+  }
+
+  Intervention.prototype.handleClose = function (event) {
+    if (event) {
+      event.preventDefault()
+    }
+
+    this.$module.style.display = 'none'
+  }
+
+  Modules.Intervention = Intervention
+})(window.GOVUK.Modules)
--- a/app/assets/javascripts/govuk_publishing_components/components/intervention.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/intervention.js
@@ -4,23 +4,64 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 (function (Modules) {
   function Intervention ($module) {
     this.$module = $module
-    this.$banner = this.$module.querySelector('.gem-c-intervention')
-    this.$closeLink = this.$module.querySelector('.gem-c-intervention__dismiss-link')
+    this.$closeLink = this.$module.querySelector('.js-dismiss-link')
+    this.$campaignName = this.$module.getAttribute('data-intervention-name')
+    this.$campaignCookie = window.GOVUK.cookie('intervention_campaign') || ''
   }
 
   Intervention.prototype.init = function () {
-    this.$module.close = this.handleClose.bind(this)
+    this.$cookieHasCampaign = this.cookieHasCampaign()
 
     if (this.$closeLink) {
+      this.$module.close = this.handleClose.bind(this)
       this.$closeLink.addEventListener('click', this.$module.close)
+    }
+
+    if (this.$cookieHasCampaign) {
+      this.hideBanner()
     }
   }
 
   Intervention.prototype.handleClose = function (event) {
-    if (event) {
-      event.preventDefault()
+    event.preventDefault()
+
+    if (this.$cookieHasCampaign) {
+      this.setCookies()
+    }
+    this.hideBanner()
+  }
+
+  Intervention.prototype.cookieHasCampaign = function () {
+    if (this.$campaignCookie) {
+      return this.cookieValues().includes(this.$campaignName)
+    }
+    return false
+  }
+
+  Intervention.prototype.cookieValues = function () {
+    if (this.$campaignCookie !== null && this.$campaignCookie.length > 0) {
+      return this.$campaignCookie.split(',')
+    }
+  }
+
+  Intervention.prototype.appendCookieValues = function () {
+    return this.$campaignCookie + ',' + this.$campaignName
+  }
+
+  Intervention.prototype.setCookies = function () {
+    if (this.$campaignCookie) {
+      window.GOVUK.setCookie('intervention_campaign', this.appendCookieValues(), { days: 30 })
+    } else {
+      window.GOVUK.setCookie('intervention_campaign', this.$campaignName, { days: 30 })
+    }
+  }
+
+  Intervention.prototype.hideBanner = function () {
+    if (!this.$cookieHasCampaign) {
+      this.setCookies()
     }
 
+    this.$module.hidden = true
     this.$module.style.display = 'none'
   }
 
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -15,12 +15,15 @@
   aria_attributes ||= {}
   aria_attributes[:label] = 'Intervention'
 
-  local_assigns[:query_string] ||= request.query_string
-  local_assigns[:suggestion_text] = suggestion_text
-  local_assigns[:suggestion_link_text] = suggestion_link_text
-  local_assigns[:suggestion_link_url] = suggestion_link_url
+  options = {
+    params: request.params,
+    query_string: request.query_string,
+    suggestion_text: suggestion_text,
+    suggestion_link_text: suggestion_link_text,
+    suggestion_link_url: suggestion_link_url,
+  }
 
-  intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(local_assigns)
+  intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(options)
   dismiss_href = intervention_helper.dismiss_link
 
   suggestion_tag_options = {
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -1,6 +1,7 @@
 <%
   add_gem_component_stylesheet("intervention")
 
+  name ||= ""
   dismiss_text ||= false
   suggestion_link_text ||= false
   suggestion_link_url ||= false
@@ -10,6 +11,8 @@
 
   data_attributes ||= {}
   suggestion_data_attributes ||= {}
+  data_attributes[:module] = "intervention"
+  data_attributes["intervention-name"] = name
 
   aria_attributes ||= {}
   aria_attributes[:label] = 'Intervention'
@@ -40,8 +43,6 @@
     suggestion_link_text = intervention_helper.accessible_text
   end
 
-  data_attributes[:module] = "intervention"
-
   section_options = {
     class: "gem-c-intervention",
     role: "region", aria: aria_attributes,
@@ -60,7 +61,7 @@
 
     <% if dismiss_text %>
       <p class="govuk-body">
-        <%= tag.a class: "govuk-link gem-c-intervention__dismiss-link", href: dismiss_href do %>
+        <%= tag.a class: "govuk-link js-dismiss-link", href: dismiss_href do %>
           <svg class="gem-c-intervention__dismiss-icon"
             width="19" height="19" viewBox="0 0 19 19"
             aria-hidden="true"
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -10,7 +10,6 @@
 
   data_attributes ||= {}
   suggestion_data_attributes ||= {}
-  dismiss_data_attributes ||= {}
 
   aria_attributes ||= {}
   aria_attributes[:label] = 'Intervention'
@@ -41,6 +40,8 @@
     suggestion_link_text = intervention_helper.accessible_text
   end
 
+  data_attributes[:module] = "intervention"
+
   section_options = {
     class: "gem-c-intervention",
     role: "region", aria: aria_attributes,
@@ -59,7 +60,7 @@
 
     <% if dismiss_text %>
       <p class="govuk-body">
-        <%= tag.a class: "govuk-link", href: dismiss_href, data: dismiss_data_attributes do %>
+        <%= tag.a class: "govuk-link gem-c-intervention__dismiss-link", href: dismiss_href do %>
           <svg class="gem-c-intervention__dismiss-icon"
             width="19" height="19" viewBox="0 0 19 19"
             aria-hidden="true"
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -16,6 +16,7 @@
   aria_attributes[:label] = 'Intervention'
 
   local_assigns[:query_string] ||= request.query_string
+  local_assigns[:suggestion_text] = suggestion_text
   local_assigns[:suggestion_link_text] = suggestion_link_text
   local_assigns[:suggestion_link_url] = suggestion_link_url
 
@@ -44,7 +45,7 @@
   }
   section_options.merge!({ hidden: true }) if hide
 %>
-<% if suggestion_text || (suggestion_link_text && suggestion_link_url) %>
+<% if intervention_helper.show? %>
   <%= tag.section **section_options do %>
     <p class="govuk-body">
       <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
--- a/app/views/govuk_publishing_components/components/_intervention.html.erb
+++ b/app/views/govuk_publishing_components/components/_intervention.html.erb
@@ -18,6 +18,8 @@
   aria_attributes[:label] = 'Intervention'
 
   options = {
+    name: name,
+    dismiss_text: dismiss_text,
     params: request.params,
     query_string: request.query_string,
     suggestion_text: suggestion_text,
--- a/app/views/govuk_publishing_components/components/docs/intervention.yml
+++ b/app/views/govuk_publishing_components/components/docs/intervention.yml
@@ -6,7 +6,10 @@ body: |
   that would be useful to them. This component would be used to add this personalised content and would
   indicate to the user that this is not normally part of the page, but has been added for them specifically.
 
-  The dismiss link points to the current URL with the "hide-intervention" query string parameter set to "true". The backend should check for this string in order to render the same page without the intervention.
+  The dismiss link works without Javascript by pointing to the current URL with the "hide-intervention" 
+  query string parameter set to "true". It's progressively enhanced and sets a cookie to remember that 
+  the user has dismissed the banner before. The cookie requires a "name" parameter, the value of which 
+  is stored in the cookie to distinguish which campaign banner the user has dismissed.
 
 accessibility_criteria: |
   The intervention component must:
@@ -33,9 +36,11 @@ examples:
     data:
       suggestion_text: You should renew your permit every 6 months.
       dismiss_text: Hide this suggestion
+      name: campaign-name
 
   with_suggestion_link_only:
     data:
+      suggestion_text: You should renew your permit every 6 months.
       suggestion_link_text: You can now apply for a permit online.
       suggestion_link_url: /permit
 
@@ -44,6 +49,7 @@ examples:
       This example is for when we want to hide by default and display to the user based on some logic,
       either in the backend or with Javascript.
     data:
+      suggestion_text: You should renew your permit every 6 months.
       suggestion_link_text: You may be invited to fill in a questionnaire
       suggestion_link_url: /questionnaire
       hide: true
@@ -55,6 +61,7 @@ examples:
       Link text should tell the user that the link opens in a new tab.
       Note: "(opens in a new tab)" is added to link text if the phrase isn't included.
     data:
+      suggestion_text: You should renew your permit every 6 months.
       suggestion_link_text: You can now apply for a permit online
       suggestion_link_url: www.google.com/permit
       new_tab: true
@@ -76,6 +83,7 @@ examples:
         track-dimension-index: 29
         track-label: clicked suggestion
       dismiss_text: Hide this suggestion
+      name: another-campaign-name
       dismiss_data_attributes:
         track-category: interventionBanner
         track-action: interventionDismissed
--- a/app/views/govuk_publishing_components/components/docs/intervention.yml
+++ b/app/views/govuk_publishing_components/components/docs/intervention.yml
@@ -33,6 +33,9 @@ examples:
       suggestion_link_url: /travel-abroad
 
   with_dismiss_link:
+    description: | 
+      Name is required in order to set a cookie. The name value should be distinct to the campaign for the banner, 
+      so that other banners using this component are not hidden by accident.
     data:
       suggestion_text: You should renew your permit every 6 months.
       dismiss_text: Hide this suggestion
--- a/app/views/govuk_publishing_components/components/docs/intervention.yml
+++ b/app/views/govuk_publishing_components/components/docs/intervention.yml
@@ -38,12 +38,6 @@ examples:
       dismiss_text: Hide this suggestion
       name: campaign-name
 
-  with_suggestion_link_only:
-    data:
-      suggestion_text: You should renew your permit every 6 months.
-      suggestion_link_text: You can now apply for a permit online.
-      suggestion_link_url: /permit
-
   with_hide:
     description: |
       This example is for when we want to hide by default and display to the user based on some logic,
--- a/lib/govuk_publishing_components/presenters/intervention_helper.rb
+++ b/lib/govuk_publishing_components/presenters/intervention_helper.rb
@@ -4,6 +4,7 @@ module GovukPublishingComponents
       def initialize(local_assigns)
         @accessible_text_suffix = I18n.t("components.intervention.accessible_link_text_suffix")
         @query_string = local_assigns[:query_string]
+        @suggestion_text = local_assigns[:suggestion_text]
         @suggestion_link_text = local_assigns[:suggestion_link_text]
         @suggestion_link_url = local_assigns[:suggestion_link_url]
       end
@@ -29,9 +30,13 @@ module GovukPublishingComponents
         rel
       end
 
+      def show?
+        @suggestion_text || (@suggestion_link_text && @suggestion_link_url)
+      end
+
     private
 
-      attr_reader :accessible_text_suffix, :query_string, :suggestion_link_text, :suggestion_link_url
+      attr_reader :accessible_text_suffix, :query_string, :suggestion_text, :suggestion_link_text, :suggestion_link_url
     end
   end
 end
--- a/lib/govuk_publishing_components/presenters/intervention_helper.rb
+++ b/lib/govuk_publishing_components/presenters/intervention_helper.rb
@@ -1,12 +1,13 @@
 module GovukPublishingComponents
   module Presenters
     class InterventionHelper
-      def initialize(local_assigns)
+      def initialize(options = {})
         @accessible_text_suffix = I18n.t("components.intervention.accessible_link_text_suffix")
-        @query_string = local_assigns[:query_string]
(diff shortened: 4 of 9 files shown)


app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -9,6 +9,7 @@
   data_attributes ||= {}
   suggestion_data_attributes ||= {}
   dismiss_data_attributes ||= {}
+  dismiss_data_attributes
```
leenagupte, line 12: Is this needed?

leenagupte (commented):
The backend parts look good. I've left a few inline comments / suggestions. 

The Javascript module looks very much like the module for the modal dialogue component, so I'm sure it's ok, but it should definitely be reviewed by someone with more JS knowledge.

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,28 @@
...
+      this.$closeLink.addEventListener('click', this.$module.close)
+    }
+  }
+
+  Intervention.prototype.handleClose = function (event) {
+    if (event) {
+      event.preventDefault()
+    }
+
+    this.$module.style.display = 'none'
```
andysellick, line 24: Might be worth checking from an accessibility perspective what's best here - whether to hide the element or remove it from the page.

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -9,15 +9,20 @@
...
+  options = {
+    params: request.params,
+    query_string: request.query_string,
+    suggestion_text: suggestion_text,
+    suggestion_link_text: suggestion_link_text,
+    suggestion_link_url: suggestion_link_url,
+  }
 
-  intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(local_assigns)
+intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(options)
```
andysellick, line 25: Nitpick: indent this line?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
andysellick, line 24: How does this not reappear if the page is reloaded? Or what's the plan for how that should work?

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -35,14 +40,16 @@
...
+
   section_options = {
     class: "gem-c-intervention",
     role: "region", aria: aria_attributes,
     data: data_attributes,
   }
   section_options.merge!({ hidden: true }) if hide
 %>
-<% if suggestion_text || (suggestion_link_text && suggestion_link_url) %>
+<% if intervention_helper.show? %>
```
andysellick, line 52: There's a lot of logic before this thing that decides whether the component gets shown or not - might be better if this line was earlier. Maybe outside the scope of this PR.

andysellick (commented):
Mainly looked at the JS. I thought originally that the dismiss functionality of this component would be handled at an application level? Fine if that's changed, just wondering.

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: Argh...

andysellick: (sorry)

chao-xian: I can only think to set a cookie (if accepted)... I thought about also setting a query param but that wouldn't help on page refresh nor going to another page that also has the banner. Cookies it is then...

andysellick:
> @andysellick I wasn't aware that the intention was to make it per app, but given that the intervention banner is used for the same sort of thing (research recruitment) across several apps (2 at a time) it makes more sense to do this here? It also just made it seem like it's broken

Yeah I think it probably does make more sense to put it here. I guess it just never got implemented in applications.

app/assets/javascripts/govuk_publishing_components/components/intervention.js
andysellick: https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js#L169 some potentially useful code here to help with that.

lib/govuk_publishing_components/presenters/intervention_helper.rb
```diff
@@ -16,10 +17,10 @@ def accessible_text
       end
 
       def dismiss_link
-        if @query_string.present?
+        return "?hide-intervention=true" unless @query_string.present?
+
+        unless @query_string.include?("hide-intervention=true")
```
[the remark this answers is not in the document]
chao-xian: I think I should still be able to use `Addressable::URI.path()` if the path is used...

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: I can only get it to this point, but it's less readable:

```
<%
  dismiss_text ||= false
  suggestion_link_text ||= false
  suggestion_link_url ||= false
  suggestion_text ||= nil
  hide ||= false
  new_tab ||= false

  data_attributes ||= {}
  suggestion_data_attributes ||= {}
  dismiss_data_attributes ||= {}
  dismiss_data_attributes

  aria_attributes ||= {}
  aria_attributes[:label] = 'Intervention'

  options = {
    params: request.params,
    query_string: request.query_string,
    suggestion_text: suggestion_text,
    suggestion_link_text: suggestion_link_text,
    suggestion_link_url: suggestion_link_url,
  }

  intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new(options)
  return unless intervention_helper.show?

  dismiss_href = intervention_helper.dismiss_link

  suggestion_tag_options = {
    class: "govuk-link",
    href: suggestion_link_url,
    data: suggestion_data_attributes,
  }

  data_attributes[:module] = "intervention"

  section_options = {
    class: "gem-c-intervention",
    role: "region", aria: aria_attributes,
    data: data_attributes,
  }
  section_options.merge!({ hidden: true }) if hide

  if new_tab && (suggestion_link_text && suggestion_link_url)
    target = "_blank"
    rel = intervention_helper.security_attr

    suggestion_tag_options.merge!({ target: target, rel: rel })

    suggestion_link_text = intervention_helper.accessible_text
  end
%>

<%= tag.section **section_options do %>
  <p class="govuk-body">
    <%= tag.span suggestion_text, class: "gem-c-intervention__textwrapper" if suggestion_text %>
    <% if suggestion_link_text && suggestion_link_url %>
      <%= tag.a suggestion_link_text, **suggestion_tag_options %>
    <% end %>
  </p>

  <% if dismiss_text %>
    <p class="govuk-body">
      <%= tag.a class: "govuk-link gem-c-intervention__dismiss-link", href: dismiss_href, data: dismiss_data_attributes do %>
        <svg class="gem-c-intervention__dismiss-icon"
          width="19" height="19" viewBox="0 0 19 19"
          aria-hidden="true"
          fill="none" stroke="currentColor">
          <path d="M0.499997 9.5C0.499997 4.52944 4.52944 0.499999 9.5 0.499999C14.4706 0.5 18.5 4.52944 18.5 9.5C18.5 14.4706 14.4706 18.5 9.5 18.5C4.52943 18.5 0.499997 14.4706 0.499997 9.5Z"/>
          <path d="M13.3477 5.4209L9.67383 9.09473L6 5.4209"/>
          <path d="M6 13.4209L9.67383 9.74707L13.3477 13.4209"/>
        </svg>
        <%= dismiss_text %>
      <% end %>
    </p>
  <% end %>
<% end %>
```

lib/govuk_publishing_components/presenters/intervention_helper.rb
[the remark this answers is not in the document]
chao-xian: Have addressed your first comment. Will see if I can update to use Addressable in a later commit

spec/lib/govuk_publishing_components/components/intervention_helper_spec.rb
```diff
@@ -9,40 +9,65 @@
...
+
+      it "returns falsey value when only some of required params are passed" do
+        intervention_helper = GovukPublishingComponents::Presenters::InterventionHelper.new({ suggestion_link_text: "Text", params: {} })
 
         expect(intervention_helper.show?).to be_falsey
       end
+
+      it "returns falsey value when required params and hide query param is passed" do
+        params = {
+          "a" => "b",
```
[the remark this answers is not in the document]
chao-xian: Nope! Good spot, thank you

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: What's the difference with doing this and the way cookie banners does `.hidden = true`? https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js#L105

andysellick: I think the hidden attribute works even if stylesheets are disabled/not loaded, but I'm not certain.

chao-xian: Seeing as cookies is the main way we do it on all GOV.UK pages I'm going to follow that convention...

lib/govuk_publishing_components/presenters/intervention_helper.rb
```diff
@@ -16,10 +17,10 @@ def accessible_text
       end
 
       def dismiss_link
-        if @query_string.present?
-          "#{@query_string}&hide-intervention=true"
-        else
+        if !@query_string.present?
           "?hide-intervention=true"
+        elsif !@query_string.include?("hide-intervention=true")
+          "#{@query_string}&hide-intervention=true"
```
leenagupte, line 23: I think this change can be removed completely. It's doing the same as the original if/else, except it's doing a negative check.

i.e. it's the same as leaving it as:

```ruby
if @query_string.present?
   "#{@query_string}&hide-intervention=true"
else
   "?hide-intervention=true"
end
```

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: Was looking at the wrong place. Here https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js#L67

display = 'none' as well

lib/govuk_publishing_components/presenters/intervention_helper.rb
chao-xian: 🤦

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: oh _now_ I realise why doing this per app _might_ be better because you can be running multiple campaigns using the intervention component across GOV.UK so setting a cookie to remember if you dismissed one, if I go ahead and implement it the way I can only think here, would hide _all_ of them including ones the user _won't have seen yet_... Thoughts @andysellick ?

chao-xian: One thing I could do I suppose is set the cookie name from a data attribute based on one of the options passed to the component?

andysellick: Yes, that could work. The JS would need to read that in and check the cookie to determine whether to allow the banner to be shown or not on page load. Not too complex, hopefully.

chao-xian:
@andysellick progressive enhancement all working... please could you re-review. Thanks!

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+  function Intervention ($module) {
+    this.$module = $module
+    this.$banner = this.$module.querySelector('.gem-c-intervention')
+    this.$closeLink = this.$module.querySelector('.gem-c-intervention__dismiss-link')
+    this.$campaignName = this.$module.getAttribute('data-intervention-name')
+    this.$interventionCookie = window.GOVUK.cookie('intervention_campaign')
+  }
+
+  Intervention.prototype.init = function () {
+    this.$module.close = this.handleClose.bind(this)
```
andysellick, line 14: Should this line be moved into the `if` statement below?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+      event.preventDefault()
+    }
+
+    if (this.cookieHasCampaign()) {
+      this.setCookies()
+    }
+    this.hideBanner()
+  }
+
+  Intervention.prototype.cookieHasCampaign = function () {
```
andysellick, line 36: This function is called four times in this file - would it be possible to call it once at the start, store the value, then refer to that for the rest?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+    if (!this.$interventionCookie) {
```
andysellick, line 37: Code reduction suggestion:

```JavaScript
if (this.$interventionCookie) {
  return this.cookieValues().includes(this.$campaignName)
}
return false

```

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+      this.$closeLink.addEventListener('click', this.$module.close)
+    }
+
+    if (this.cookieHasCampaign()) {
+      this.hideBanner()
+    }
+  }
+
+  Intervention.prototype.handleClose = function (event) {
+    if (event) {
```
andysellick, line 26: Do you need to check for `event` here? Is there a situation where it wouldn't be present?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+  Intervention.prototype.cookieHasCampaign = function () {
+    if (!this.$interventionCookie) {
+      return false
+    } else {
+      var included = this.cookieValues().includes(this.$campaignName)
+      return included
+    }
+  }
+
+  Intervention.prototype.readCampaignCookie = function () {
```
andysellick, line 45: Could this function's return value be stored on module initialisation, rather than calling this function twice in the code?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+    var cookieString = this.readCampaignCookie()
+
+    if (cookieString !== null && cookieString.length > 0) {
+      return cookieString.split(',')
+    }
+  }
+
+  Intervention.prototype.appendCookieValues = function () {
+    var commaPrefixed = ','.concat(this.$campaignName)
+    return this.readCampaignCookie().concat(commaPrefixed)
```
andysellick, line 59: Interested by the use of `concat` here as apparently it's intended for merging arrays, which I don't think any of these things are?

I think this function could be simplified to

```JavaScript
return this.readCampaignCookie() + ',' + this.$campaignName
```

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -35,14 +39,17 @@
     suggestion_link_text = intervention_helper.accessible_text
   end
 
+  data_attributes[:module] = "intervention"
+  data_attributes["intervention-name"] = name
```
andysellick, line 43: Nitpick: could these two lines setting `data_attributes` be moved up to around line 11, to put them together with the initial declaration for `data_attributes`?

app/views/govuk_publishing_components/components/_intervention.html.erb
```diff
@@ -53,7 +60,7 @@
 
     <% if dismiss_text %>
       <p class="govuk-body">
-        <%= tag.a class: "govuk-link", href: dismiss_href, data: dismiss_data_attributes do %>
+        <%= tag.a class: "govuk-link gem-c-intervention__dismiss-link", href: dismiss_href do %>
```
andysellick, line 63: Since the `gem-c-intervention__dismiss-link` class is used entirely as a JS hook and not for styling, could it be renamed to indicate this? Something like `js-dismiss-link` would be fine.

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -33,9 +36,11 @@ examples:
     data:
       suggestion_text: You should renew your permit every 6 months.
       dismiss_text: Hide this suggestion
+      name: campaign-name
 
   with_suggestion_link_only:
     data:
+      suggestion_text: You should renew your permit every 6 months.
```
andysellick, line 43: I'm confused why `suggestion_text` has been added to all of these examples, can you explain why it wasn't here before? I think for this specific example it's been added in error - the example name suggests that this is to show the component without suggestion text.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,91 @@
...
+      var close = document.querySelector('.gem-c-intervention__dismiss-link')
+      close.click()
+
+      expect(GOVUK.cookie).toHaveBeenCalled()
+      var bannerCookie = GOVUK.cookie('intervention_campaign')
+      expect(bannerCookie).toEqual('test-intervention-name')
+    })
+  })
+
+  describe('cookies are already set', function () {
```
andysellick, line 48: Nitpick: describe name should be `when cookies are already set`?

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,91 @@
...
+      new GOVUK.Modules.Intervention(element).init()
+
+      var banner = document.querySelector('.gem-c-intervention')
+
+      expect(banner).toBeHidden()
+    })
+  })
+})
+
+describe('another concurrent banner campaign', function () {
```
andysellick, line 61: Why is this an entirely separate describe block? It should be part of the main `Intervention banner component` describe block (much of the `before` and `after` code looks duplicated).

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,91 @@
...
+    var element = document.querySelector('[data-module="intervention"]')
+    new GOVUK.Modules.Intervention(element).init()
+  })
+
+  it('appends campaign value to the cookie', function () {
+    var close = document.querySelector('.gem-c-intervention__dismiss-link')
+    close.click()
+
+    var updatedCookie = GOVUK.cookie('intervention_campaign')
+    expect(updatedCookie).toEqual('existing-value,test-intervention-name')
```
andysellick, line 84: Is there a way to check that the expiry date of the cookie has been extended? My understanding of the code is that since we're using a single cookie, when there is an existing campaign the new one is appended and the expiry date extended to 30 days from now. 

Currently I think the behaviour of the code is either:

- 1st Jan: user hides intervention banner on campaign A, expires in 30 days
- 29th Jan: user hides intervention banner on campaign B, uses existing cookie, extends cookie expiry
- campaign A and B now have the banner reappear 30 days after the 29th Jan

One campaign having an effect on all campaigns doesn't feel ideal, but better than the alternative:

- 1st Jan: user hides intervention banner on campaign A, expires in 30 days
- 29th Jan: user hides intervention banner on campaign B, uses existing cookie, doesn't change expiry date
- campaign A and B have the banner reappear 30 days after the original cookie, i.e. one day from the 29th Jan

If neither of those is ideal there might be a way of storing the individual campaign expiry date inside the cookie, then setting the cookie expiry to the latest value. But that way further complexity lies.

spec/javascripts/components/intervention-spec.js
```diff
@@ -0,0 +1,91 @@
...
+  })
+
+  afterEach(function () {
```
andysellick, line 87: Nitpick: normally we include the `afterEach` function immediately after the `beforeEach` for readability.

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,80 @@
...
+(function (Modules) {
+  function Intervention ($module) {
+    this.$module = $module
+    this.$banner = this.$module.querySelector('.gem-c-intervention')
+    this.$closeLink = this.$module.querySelector('.gem-c-intervention__dismiss-link')
+    this.$campaignName = this.$module.getAttribute('data-intervention-name')
+    this.$interventionCookie = window.GOVUK.cookie('intervention_campaign')
+  }
+
+  Intervention.prototype.init = function () {
```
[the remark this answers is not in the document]
chao-xian: I thought this falls under Essential?

spec/javascripts/components/intervention-spec.js
chao-xian: I think scenario 1 is what I should aim for...

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: It was missing previously as the examples only showed banners with a dismiss link with no text... which was useless.

spec/javascripts/components/intervention-spec.js
chao-xian: I had it under the same block but for clarity wanted it separated. The duplication is ok in this in my view - it's only one duplication. The other thing was about making sure it was for _two_ different banners and to initialise a different use of the component.

andysellick: The main reason for keeping all tests in a single describe block is so that they're kept together when run in the browser, currently they're separate and there's no obvious indication what the other describe block is testing without grepping through the code:

Screenshot 2023-02-20 at 14 37 40

The other things you mention should be resolvable - I'd be happy to pair if that would help?

spec/javascripts/components/intervention-spec.js
andysellick: Agreed, it's far simpler. Would be good to have a test to check that it's doing that, if possible.

app/assets/javascripts/govuk_publishing_components/components/intervention.js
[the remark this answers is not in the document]
andysellick: That's a good point. Okay, never mind.

app/views/govuk_publishing_components/components/docs/intervention.yml
andysellick: If that's the case this example should be removed entirely. Also should `suggestion_text` be a required option for the component to render?

app/assets/javascripts/govuk_publishing_components/components/intervention.js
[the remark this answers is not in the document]
chao-xian: _phew_

spec/javascripts/components/intervention-spec.js
chao-xian: As agreed from our call, I won't add a test for this as it's impossible to affect and read the cookie expiration values in Jasmine. Plus it feels like testing for browser behaviour at this point. The expected behaviour of the cookie expiry extending when clicking on another campaign banner was checked during the call.

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: Addressed in https://github.com/alphagov/govuk_publishing_components/pull/3101/commits/5bdf8c768d8b7b7ad5a39b45f38dd80f11561483

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: It is a String function https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/concat 

and supported as far back as IE 

https://caniuse.com/mdn-javascript_builtins_string_concat

and much easier to read than the + operator

chao-xian: but maybe it can be simplified...

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: thanks - done

app/assets/javascripts/govuk_publishing_components/components/intervention.js
chao-xian: good point - removerd

app/views/govuk_publishing_components/components/_intervention.html.erb
chao-xian: thanks - done

app/assets/javascripts/govuk_publishing_components/components/intervention.js
```diff
@@ -0,0 +1,70 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function Intervention ($module) {
+    this.$module = $module
+    this.$closeLink = this.$module.querySelector('.js-dismiss-link')
+    this.$campaignName = this.$module.getAttribute('data-intervention-name')
+    this.$interventionCookie = window.GOVUK.cookie('intervention_campaign')
+    this.$campaignCookie = window.GOVUK.cookie('intervention_campaign') || ''
```
andysellick, line 10: campaignCookie and interventionCookie appear to be almost the same variable, is there a reason they couldn't be one?

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -6,7 +6,10 @@ body: |
   that would be useful to them. This component would be used to add this personalised content and would
   indicate to the user that this is not normally part of the page, but has been added for them specifically.
 
-  The dismiss link points to the current URL with the "hide-intervention" query string parameter set to "true". The backend should check for this string in order to render the same page without the intervention.
+  The dismiss link works without Javascript by pointing to the current URL with the "hide-intervention" 
+  query string parameter set to "true". It's progressively enhanced and sets a cookie to remember that 
+  the user has dismissed the banner before. The cookie requires a "name" parameter, the value of which 
+  is stored in the cookie to distinguish which campaign banner the user has dismissed.
```
andysellick, line 12: I just want to check this... `name` isn't a required parameter for the component, what happens if one isn't passed?

app/views/govuk_publishing_components/components/docs/intervention.yml
```diff
@@ -21,7 +24,7 @@ accessibility_excluded_rules:
 examples:
   default:
     data:
-      suggestion_text: You should renew your permit every 6 months.
+      suggestion_text: You should renew your permit every 6 months. Required.
```
andysellick, line 27: No worries, but why was 'Required' added here?

lib/govuk_publishing_components/presenters/intervention_helper.rb
```diff
@@ -1,11 +1,13 @@
...
   module Presenters
     class InterventionHelper
-      def initialize(local_assigns)
+      def initialize(options = {})
         @accessible_text_suffix = I18n.t("components.intervention.accessible_link_text_suffix")
-        @query_string = local_assigns[:query_string]
-        @suggestion_link_text = local_assigns[:suggestion_link_text]
-        @suggestion_link_url = local_assigns[:suggestion_link_url]
+        @params = options[:params] || nil
+        @query_string = options[:query_string] || nil
```
andysellick, line 7: Is defaulting these values to `nil` necessary? I thought that would be the default already if not supplied.

lib/govuk_publishing_components/presenters/intervention_helper.rb
```diff
@@ -29,9 +31,16 @@ def security_attr
         rel
       end
 
+      def show?
+        # return false if suggestion_text.blank? || params["hide-intervention"] == "true"
```
andysellick, line 35: Line of debug needs to be removed?

andysellick:
Some of the visual difference checker fails can be fixed by rebasing this PR against `main`, happy to approve visual changes once that's done.

app/views/govuk_publishing_components/components/docs/intervention.yml
chao-xian: It's now required as discussed

andysellick (approved):
Well done! 👏 Thanks for bearing with all my requests for change.

PR #3102, opened 2022-12-05
Use the base path for World Location related links

brucebolt:
## What
We are currently creating the URL of World Location related links by using the World Location's name.

However there are multiple examples where the name does not match the slug (e.g. an acronym in the slug, but spelt out in full in the name).

This was attempted to be solved by hardcoding a list of special cases, but that list has now become out of date.

Since the content item already includes the base path (which will always be up-to-date), we should use that and remove the list of exceptions.

## Why
There are currently broken related links on documents that link to World Locations.

## Visual Changes
No visual changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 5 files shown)


JonathanHallam (approved):
Totally forgot about this one but it's looking good to me!

PR #3108, opened 2022-12-06
[DO NOT MERGE] Refactor button component print styles

jon-kirwan:
## What / why

## Visual Changes
None.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components_print.scss
@@ -4,7 +4,6 @@
 @import "govuk_frontend_support";
 
 @import "components/print/accordion";
-@import "components/print/button";
 @import "components/print/cards";
 @import "components/print/contents-list";
 @import "components/print/emergency-banner";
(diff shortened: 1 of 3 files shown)


app/views/layouts/govuk_publishing_components/application.html.erb
```diff
@@ -16,7 +16,7 @@
     <%= csrf_meta_tags %>
     <%= favicon_link_tag "govuk_publishing_components/favicon-production.png" %>
 
-    <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
+    <%= stylesheet_link_tag "component_guide/application", media: "all" %>
```
andysellick, line 19: Is it worth putting this change in a separate commit? It's related, but it feels like a significant thing all of its own.

andysellick (approved):

app/views/layouts/govuk_publishing_components/application.html.erb
jon-kirwan: Yep good idea. I'll remove it from here.

jon-kirwan:

Issue #3005, opened 2022-10-06
Print stylesheets are separate from screen stylesheets


injms:
Stylesheets imported from GOV.UK Frontend contain both screen and print styles in the same file:

[Example](https://github.com/alphagov/govuk-frontend/blob/cdd027bef7fbbea83756e514808cbde00f46aa2e/src/govuk/helpers/_typography.scss#L30-L32):

```scss
@mixin govuk-typography-common($font-family: $govuk-font-family) {

   ...

  @include govuk-media-query($media-type: print) {
    font-family: $govuk-font-family-print;
  }
}
```

Stylesheets for components in this gem have separate files for print and screen, for example:
+ [print/_accordion.scss](https://github.com/alphagov/govuk_publishing_components/blob/282383794b148a13cba3fb9236f4a081ff07b814/app/assets/stylesheets/govuk_publishing_components/components/print/_accordion.scss)
+ [_accordion.scss](https://github.com/alphagov/govuk_publishing_components/blob/282383794b148a13cba3fb9236f4a081ff07b814/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss)

This means that the component gem contains stylesheets which have both rules for print and screen, and stylesheets that are only print.

This isn't breaking things because the non-print stylesheets are included with `media="all"` - but this is confusing.

I'd suggest that the component gem stylesheets should mirror the behaviour of GOV.UK Frontend, and have both screen and print stylesheets in the same file. This is a better option than trying to separate the print and screen styles out of GOV.UK Frontend, as that's very difficult / almost impossible.

Positives:
+ Easier to add print styles to a component, since the print stylesheet doesn't need to be manually added to a rendering application
+ Good to follow GOV.UK Frontend's lead on this - separating their styles into screen and print would be very difficult
+ One less concurrent download
+ No way for devs to forget to include the print CSS when they bring a new component into a rendering application

Negatives:
- much more difficult to audit which components have print stylesheets
- this isn't broken, so does it need fixing?

maxgds:
We discussed this previoiusly and people seemed generally positive about it - I can't recall whether we foresaw any potential issues and my bad for not actually writing it up. I'd add to the positives:
1 less concurrent download
No way for devs to forget to include the print css when they bring a new component into a rendering app

PR #3110, opened 2022-12-06
Remove print stylesheets from components

andysellick:
## What / why
Remove print stylesheets from components, in favour of using print mixins in the screen stylesheet. This will reduce requests, simplifies component management, and may reduce asset size in some components.

This PR will only move styles from the print scss to screen scss, to avoid a breaking change that would result from removing them altogether. We'll delete the empty files in a follow up PR.

Remaining tasks:

- [x] include change to component guide layout template screen stylesheet to 'all'
- [x] update documentation referring to print styles
- [x] remove print styles for organisation logo, as seems to be better without
- [x] update card print styles, reduce spacing, remove arrows
- [x] cookie banner seems to have no print styles, should be hidden when printed
- [x] update auditing (a [separate PR](https://github.com/alphagov/govuk_publishing_components/pull/3128) in progress)

## Things to note

We've had to update the way the components gem uses its stylesheets, previously it had one for `screen` and one for `print` but we've had to change it so that application.css is now `all` (because it now contains the print styles). This had led to some good changes e.g. accordion and step by step nav were had significant copies of their screen styles in their print styles, which have now been removed.

We've checked and our public applications don't need this change - they already use `all`. In a few cases they don't specify (in which case it defaults to `all`) but we'll update those to make them more explicit.

## Testing

Chrome dev tools has an option to render the page in the browser in print mode, but it lies! You can try various things but the easiest way to check how it looks in print mode is to just open print preview (CMD P on a Mac).

We've checked how different browsers behave when given an empty print stylesheet and Chrome and IE seem to be fine.

Public applications tested with this PR:

- [x] collections
- [x] email-alert-frontend
- [x] feedback
- [x] finder-frontend
- [x] frontend
- [x] government-frontend (checked in Chrome and IE11 - no issues, although (unrelated) IE11 print retains page grid column layout when it shouldn't)
- [x] info-frontend
- [x] licence-finder
- [x] service-manual-frontend
- [x] smart-answers
- [x] whitehall
- [x] datagovuk

Publishing apps to test (?)

(everything below this point is unlikely to ever be printed, so we could save ourselves some effort here...)

- [ ] content-publisher
- [ ] content-data-admin
- [ ] collections-publisher
- [ ] travel-advice-publisher

Other apps to test (?)

- [ ] signon
- [ ] release

## Visual Changes
Some improvements to print styles have been achieved, including removing background colour from step by step nav header, reducing the squashed appearance of organisation logos, cleaning up the appearance of the card component and hiding the cookie banner.

Fixes: https://github.com/alphagov/govuk_publishing_components/issues/3005


app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
```diff
@@ -1,116 +1 @@
...
-}
-
-.gem-c-step-nav__link {
-  margin-bottom: .3em;
-}
-
-.gem-c-step-nav__circle-step-label,
-.gem-c-step-nav__circle-step-colon {
-  display: none;
-}
```
JamesCGDS, line 116: Just out of curiosity - what's the effect of removing these styles and not copying them over into the `print` mixin? The PR does mention that some styles were being duplicated and can be removed entirely; is that what's happening here?

app/assets/stylesheets/govuk_publishing_components/components/print/_accordion.scss
```diff
@@ -1,28 +1 @@
-// This is in place until a proper fix comes through from GOV.UK Frontend.
-//
-// TODO: Replace with the print styles that will come from GOV.UK Frontend.
-
-@import "govuk_publishing_components/component_support";
```
JamesCGDS, line 5: Just checking - is it okay that this `import` isn't being copied over as well?

MartinJJones: Thanks James, it was initially copied over as well, but after further investigation it is not required.

Further details are included on this comment: https://github.com/alphagov/govuk_publishing_components/pull/3114#discussion_r1040929056

JamesCGDS: Cool, thanks Martin! 👍

app/assets/stylesheets/govuk_publishing_components/components/print/_step-by-step-nav.scss
MartinJJones: Overall the effect is positive, we ended up with fewer lines off CSS, while improving the print styles.

Further details are included on this closed PR's description, including screenshots showing before and after:

https://github.com/alphagov/govuk_publishing_components/pull/3118#issue-1479197815

JamesCGDS: Great, thanks Martin 👍

JamesCGDS (approved):

app/views/layouts/govuk_publishing_components/application.html.erb
```diff
@@ -16,7 +16,7 @@
     <%= csrf_meta_tags %>
     <%= favicon_link_tag "govuk_publishing_components/favicon-production.png" %>
 
-    <%= stylesheet_link_tag "component_guide/application", media: "screen" %>
+    <%= stylesheet_link_tag "component_guide/application", media: "all" %>
     <%= stylesheet_link_tag "component_guide/print", media: "print" %>
```
kevindew, line 20: Should we get rid of this too? I think it doesn't do anything.

kevindew (commented):
This is awesome 👍 

I've only given it a quick scan as it looks quite well reviewed already.

Only thing I wonder is if you actually cause confusion trying to avoid the breaking change here. I guess we're at risk of ending up with a print stylesheet with 0 bytes in it. I wonder if it's actually better to be explicit as it'll mean apps have to act and don't risk leaving an odd situation linger. I'm expecting we actually need apps to act as we want them to drop using media: "screen" so a breaking change (with a breaking build) might make that easier to be explained.

app/views/layouts/govuk_publishing_components/application.html.erb
andysellick: Good shout - I've just pushed another commit to remove this and a range of other references to print stuff. Not quite all of it yet though - still pondering whether to remove all the print style files as you suggest and just have the breaking change. Thanks for having a look 👍

andysellick: Think we're going to go with the breaking change as suggested @kevindew. We'll hold off merging this for today so we can get a good run at it on Monday, when we can update individual apps as dependabot warrants. Fairly sure tests will fail without the print styles so it shouldn't be deployable.

PR #3114, opened 2022-12-06
[DO NOT MERGE] Refactor accordion component print styles

MartinJJones:
## What / why

## Visual Changes
After updating the CSS class names, the accordion toggle buttons no longer show when printing as intended.

| Before | After |
| --- | --- |
| accordion-print-before | accordion-print-after |


app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
```diff
@@ -1,7 +1,36 @@
 @import "mixins/prefixed-transform";
 @import "govuk/components/accordion/accordion";
+@import "govuk_publishing_components/component_support";
```
andysellick, line 3: This is interesting - any idea why `component_support` was previously needed by the print stylesheet? Is it still needed here?

andysellick (commented):
Accordion styles look good, just one hopefully small question.

app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
MartinJJones: Good catch, it was originally added as part of this commit - https://github.com/alphagov/govuk_publishing_components/commit/1ec93f29deab43bcbdb761670a79e8964483f9af to:

> "Fix all_components_print import, `components/print/accordion` requires `govuk-frontend` mixins as it imports the main `components/accordion`"

None of the other screen stylesheets import `component_support`, as it is expected to be imported already by the application, for example in application.scss - https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L8.

I will remove the import for `component_support`, thanks

andysellick: Great, thanks for investigating. Happy to approve now.

MartinJJones:

PR #3115, opened 2022-12-06
[DO NOT MERGE] Refactor Govspeak content (HTML publications) component print styles

jon-kirwan:
## What / why

## Visual Changes
None.


andysellick (approved):
Govspeak content print looks good 👍 

We should audit all these styles at some point as well (not now) - I suspect `.govuk-sticky-element` might have been removed (or at least partly removed).

jon-kirwan:
> Govspeak content print looks good 👍
> 
> We should audit all these styles at some point as well (not now) - I suspect `.govuk-sticky-element` might have been removed (or at least partly removed).

Yep, doesn't seem to be in use. Perhaps we should create an issue for any suspect styles which can be removed?

jon-kirwan:

PR #3117, opened 2022-12-06
[DO NOT MERGE] Step nav header print

andysellick:
## What / why

## Visual Changes
Had to add some extra lines to remove the background colour when printed, as this was now coming from the regular screen styles.


andysellick:

PR #3118, opened 2022-12-06
[DO NOT MERGE] Refactor step by step navigation print styles

andysellick:
## What / why

## Visual Changes
Several! The print styles for this component were quite complex, and after fiddling for a bit I wondered what it would look like if they were removed entirely - and it turned out the only problem was that the steps weren't visible. So I've added the one rule to make the steps visible (which has to be `!important` unfortunately because of precedence higher up) but dramatically simplifies the code.

Before | After
------ | ------
Screenshot 2022-12-06 at 14 08 27 | Screenshot 2022-12-06 at 14 08 36


MartinJJones (approved):

andysellick:
Closed for https://github.com/alphagov/govuk_publishing_components/pull/3110

PR #3121, opened 2022-12-06
WIP Add option to turn headings into anchor link headings

ChrisBAshton:
## What

Defines a new `is_anchor` parameter which, when set to `true`, enhances the heading into an anchor link heading. This is a POC while frontend devs evaluate a few different UX approaches.

We still need to consider a couple of points:

1. CSS and JS
2. Do all frontend apps have govuk_publishing_components installed, and specifically the 'heading' component?

## Why
For GIFT week, we're exploring making it easier for users to share links to specific sections on GOV.UK pages, in a similar style to GitHub ([example](https://github.com/alphagov/govuk-developer-docs#before-running-the-app)).

Trello: https://trello.com/c/Gz6dirLk/36-turn-headings-into-anchor-links

## Visual Changes
TBC

### DIFF
--- a/app/views/govuk_publishing_components/components/_heading.html.erb
+++ b/app/views/govuk_publishing_components/components/_heading.html.erb
@@ -1,6 +1,7 @@
 <%
   brand ||= false
   lang = local_assigns[:lang].presence
+  is_anchor = local_assigns[:is_anchor] || false
 
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
   heading_helper = GovukPublishingComponents::Presenters::HeadingHelper.new(local_assigns)
@@ -17,3 +18,9 @@
   id: heading_helper.id,
   lang: lang
 ) %>
+<% if is_anchor %>
+<a href="#<%= heading_helper.id %>">
+  <span aria-hidden="true">#</span>
+  <span class="govuk-visually-hidden">Section titled #{text}</span>
+</a>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/heading.yml
+++ b/app/views/govuk_publishing_components/components/docs/heading.yml
@@ -85,3 +85,10 @@ examples:
     data:
       text: "Ein gweinidogion"
       lang: "cy"
+  with_anchor:
+    description: |
+      Turns the heading into an anchor link heading, so that users can easily retrieve the link to this specific section.
+    data:
+      text: "Heading"
+      id: "heading"
+      is_anchor: true
--- a/spec/components/heading_spec.rb
+++ b/spec/components/heading_spec.rb
@@ -118,4 +118,12 @@ describe "Heading", type: :view do
     render_component(text: "Inverted", inverse: true)
     assert_select ".gem-c-heading.gem-c-heading--inverse"
   end
+
+  it "augments heading with anchor link" do
+    render_component(text: "Heading", heading_level: 2, id: "heading", is_anchor: true)
+    assert_select "h2#heading"
+    assert_select "h2#heading + a[href='#heading']"
+    assert_select "h2#heading + a[href='#heading'] span[aria-hidden='true']"
+    assert_select "h2#heading + a[href='#heading'] span.govuk-visually-hidden"
+  end
 end


andysellick:
Really interested to see this bit of work, it's definitely something worth investigating.

Quick answer to one question - all public facing frontend apps (and many of our non-public facing) have `govuk_publishing_components` installed. The list below shows which ones are currently using the heading component, but it is available to all.

- collections
- collections-publisher
- datagovuk_find
- email-alert-frontend
- feedback
- finder-frontend
- frontend
- government-frontend
- licence-finder
- release
- service-manual-frontend
- smart-answers
- whitehall

One problem is that not all headings use the heading component - as well as (probably) a few instances of hard coded headings, we've got all the headings that are part of govspeak content, and therefore can't use the heading component (or rather, currently don't - it might be possible to re-engineer govspeak to use the heading component). It's also worth pointing out the [page title component](https://components.publishing.service.gov.uk/component-guide/title), which is generally used at the top of pages (so might not require an anchor link) but worth checking to make sure it's not mid-page anywhere.

Another approach could be to have a bit of JS here that runs on page load in all applications, finds all the headings in page content and progressively enhances them with anchor links. That would save having to modify all instances of this component to turn on this option in every application. If we did that though we'd need to rely on the JS to make the anchor IDs consistent, presumably based on the text of the heading, don't know if we'd need to consider what happens when content is tweaked (e.g. if the heading text is modified very slightly, it would alter the anchor and make all past bookmarks 
invalid) but I'm not saying that's a showstopper, more something to check and discuss. Something like (sorry I'm typing this quickly before my machine restarts to update)

```JavaScript
var mainContent = document.getElementById('content')
var h2s = mainContent.querySelectorAll('h2')
var h3s = mainContent.querySelectorAll('h3')
// (etc. although there's probably a better way of getting all heading tags)
for (var i = 0; i < h2s.length; i++) {
  var heading = h2s[i]
  var anchor = heading.getAttribute('id')
  if (!anchor) {
    anchor = heading.textContent
  }
  // process anchor i.e. add dashes
  // wrap heading in link 
}
```

ChrisBAshton:
Thanks for your thoughts, @andysellick, and thanks for confirming that the heading component is available universally.

You're right that GovSpeak currently doesn't use the heading component. I have a draft PR that would make it use the heading component (currently only for headings that 'opt in' to the new behaviour - via some added markdown - but in theory we could modify this to _always_ use the heading component and then only pass the `is_anchor` parameter if they've opted in). PR here: https://github.com/alphagov/govspeak/pull/263.
I think we probably ought to do that, rather than use the heading component in some cases and not use it in others. 🤔 

The JS approach is one approach we're considering: https://docs.google.com/document/d/1gDyiU097Fk0L-XvNECzu0Kbx3romLrsp57WPvCv9z_w/edit#heading=h.e872eidf4ab4. It was my preference to begin with, though Jon, Hanna and Anika have their concerns - see [Slack thread](https://gds.slack.com/archives/C04EETAK5MW/p1670255763676309?thread_ts=1670241627.890529&cid=C04EETAK5MW).

andysellick:
Thanks @ChrisBAshton. The availability of JS is a good point - I'd also wonder about whether slow loading pages would work properly for an entirely JS implementation. Lot of legwork to convert all the heading components though.

Maybe there's a middle ground using the best of each approach - we could convert govspeak to use the heading component, then modify the heading component to always generate an ID for itself based on the heading text (or [use the one supplied](https://components.publishing.service.gov.uk/component-guide/heading/with_id_attribute)).

We could then write a JS module to only turn on the functionality for certain page types. We have a similar approach with scroll tracking, which is controlled by a `data-module` attribute added to meta tags within specific templates (sometimes with additional logic to limit it to only certain pages within that page type). That would mean that while all headings have an ID, we only apply the new link functionality to e.g. detailed guides, where we could measure usage for a time to see how it performs (some tracking could be added to the module).

ChrisBAshton:
Closed as only a spike, not production code.
For suggested next steps, visit https://docs.google.com/document/d/1gDyiU097Fk0L-XvNECzu0Kbx3romLrsp57WPvCv9z_w/edit.

Issue #2902, opened 2022-08-11
We might have to remove jQuery from all the tests after all


andysellick:
Dependabot just raised [bump jasmine-core to 4.3.0](https://github.com/alphagov/govuk_publishing_components/pull/2873), which is producing several errors in the jasmine/JavaScript tests like this:

```
Uncaught TypeError: jasmine.util.argsToArray is not a function thrown
```

The gem relies on a hard coded local copy of [jasmine-jquery](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/helpers/jasmine-jquery-2.0.5.js) to provide helper functions when running tests. This in turn relies upon a function inside `jasmine-core` called `jasmine.util.argsToArray()`, which was [unfortunately removed](https://github.com/jasmine/jasmine/commit/e2e2275d41f7a380a36b0855ada19ac5e7a9fa02) in the `4.3.0` release.

Upgrading `jasmine-jquery` seemed like the logical first step, but the latest version still has this problem. Furthermore, the owners of the repo are now unable to maintain it and are seeking someone to take over responsibility.

I've [fixed this for now](https://github.com/alphagov/govuk_publishing_components/pull/2873) by hard coding the `argsToArray()` function inside our local copy of `jasmine-jquery`, but this is obviously not an ideal solution for the future. We're probably going to need to remove our dependency on this package, either by removing jQuery from all of our tests or by finding another way.

AshGDS:
We've discovered that `finder-frontend` doesn't use `jasmine-jquery` for its tests, which prompted us to investigate what `jasmine-jquery` is doing for us. It looks like it adds extra matchers (e.g. `.toBeVisible()`, `.toBeHidden()`) on tests to make it easier to write them. However we seem to be only using these extra matchers on a few tests, so we should be able to change those tests and remove `jasmine-jquery`. Here are the matchers it provides, that we will have to replace: https://www.npmjs.com/package/jasmine-jquery#jquery-matchers

hannalaakso:
I've used Cheerio before with Jest to create a jQuery like API to query in tests, I wonder if it could be used here.

MartinJJones:
This appears to be blocking the bump jasmine-core and jasmine-browser-runner #5231 

- Updates jasmine-core from 5.13.0 to 6.0.1
- Updates jasmine-browser-runner from 3.0.0 to 4.0.0

MartinJJones:
[...]

PR #3123, opened 2022-12-07
Remove jasmine jquery from tests 1

JamesCGDS:
## What
This PR removes the following `jasmine-jquery` helper functions and replaces them with a vanilla JS alternative:

- `toBeVisible()`
- `toContainElement()`
- `toContainText()`
- `toExist()`
- `toHaveAttr()`

## Why
The `jasmine-jquery` helper functions were causing [errors when running tests](https://github.com/alphagov/govuk_publishing_components/issues/2902) and so they have been replaced with a vanilla JS alternative. This also removes our tests' reliance on the `jasmine-jquery` library (which is no longer actively maintained).

## Visual Changes
N/A

### DIFF
--- a/spec/javascripts/components/cookie-banner-spec.js
+++ b/spec/javascripts/components/cookie-banner-spec.js
@@ -64,7 +64,7 @@ describe('Cookie banner', function () {
     var cookieBannerConfirmation = document.querySelector('.gem-c-cookie-banner__confirmation')
 
     expect(element.offsetWidth > 0 || element.offsetHeight > 0).toBeTruthy()
-    expect(cookieBannerMain.offsetWidth > 0 || cookieBannerMain.offsetHeight > 0 ).toBeTruthy()
+    expect(cookieBannerMain.offsetWidth > 0 || cookieBannerMain.offsetHeight > 0).toBeTruthy()
     expect(cookieBannerConfirmation).toBeHidden()
   })
 
--- a/spec/javascripts/components/cookie-banner-spec.js
+++ b/spec/javascripts/components/cookie-banner-spec.js
@@ -63,8 +63,8 @@ describe('Cookie banner', function () {
     var cookieBannerMain = document.querySelector('.js-banner-wrapper')
     var cookieBannerConfirmation = document.querySelector('.gem-c-cookie-banner__confirmation')
 
-    expect(element).toBeVisible()
-    expect(cookieBannerMain).toBeVisible()
+    expect(element.offsetWidth > 0 || element.offsetHeight > 0).toBeTruthy()
+    expect(cookieBannerMain.offsetWidth > 0 || cookieBannerMain.offsetHeight > 0 ).toBeTruthy()
     expect(cookieBannerConfirmation).toBeHidden()
   })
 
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
Looks good :+1:

What I did for repetitive things is make a function inside `SpecHelper.js` - then you can access that function in your tests on the `window.` object. I think that might be helpful to make the `.toBeVisible()` test refactor make more sense.

For example 

```JavaScript
expect(confirmationMessage.offsetWidth > 0 || confirmationMessage.offsetHeight > 0).toBeTruthy()
```

Would become

```JavaScript
expect(window.isElementHidden(confirmationMessage)).toBeTruthy() 
```

spec/javascripts/components/cookie-banner-spec.js
```diff
@@ -63,8 +63,8 @@ describe('Cookie banner', function () {
     var cookieBannerMain = document.querySelector('.js-banner-wrapper')
     var cookieBannerConfirmation = document.querySelector('.gem-c-cookie-banner__confirmation')
 
-    expect(element).toBeVisible()
-    expect(cookieBannerMain).toBeVisible()
+    expect(element.offsetWidth > 0 || element.offsetHeight > 0).toBeTruthy()
```
andysellick, line 66: Jasmine jQuery uses jQuery's `visible` selector, which is basically this but has an [additional check](https://github.com/jquery/jquery/blob/a684e6ba836f7c553968d7d026ed7941e1a612d8/src/css/hiddenVisibleSelectors.js#L8) using `getClientRects`, is it worth adding this here?

Also since this repeats in a few places would it be a good idea/possible to create a function for this and call it?

spec/javascripts/components/step-by-step-nav-spec.js
```diff
@@ -171,12 +171,14 @@ describe('A stepnav module', function () {
...
   it('inserts a button into each step to show/hide content', function () {
-    var $titleButton = $element.find('.js-step-title-button')
+    var $titleButtons = $element[0].querySelectorAll('.js-step-title-button')
 
-    expect($titleButton).toHaveClass('gem-c-step-nav__button--title')
-    expect($titleButton).toHaveAttr('aria-expanded', 'false')
-    expect($titleButton[0]).toHaveAttr('aria-controls', 'step-panel-topic-step-one-1')
-    expect($titleButton[1]).toHaveAttr('aria-controls', 'step-panel-topic-step-two-1')
+    for(var i = 0; i < $titleButtons.length; i++) {
+      expect($titleButtons[i]).toHaveClass('gem-c-step-nav__button--title')
```
andysellick, line 177: Where does `toHaveClass` come from? I can't find any documentation on it as a vanilla JS function (but I could be looking in the wrong places).

andysellick (commented):
Really great stuff, fantastic to see this happening. I just had a couple of questions, hope you don't mind me jumping in like this...

spec/javascripts/components/step-by-step-nav-spec.js
andysellick: Ah brilliant, thanks, hadn't seen that.

JamesCGDS:
Cool, thanks @AshGDS for that suggestion, I'll make that change 👍

spec/javascripts/components/cookie-banner-spec.js
JamesCGDS: Good spot, I'll make those changes. Thanks @andysellick

spec/javascripts/components/step-by-step-nav-spec.js
JamesCGDS: Basically @AshGDS and I divided the work between us and `toHaveClass` is a `jasmine-jquery` function that Ash will be replacing so it will be replaced when he raises his PR.

PR #3129, opened 2022-12-08
Remove jasmine jquery from tests 2

AshGDS:
Hi @JamesCGDS / @andysellick would you be able to approve this? Thanks :+1:

## What

This PR removes the following `jasmine-jquery` helper functions and replaces them with a vanilla JS alternative:

- toBeDisabled()
- toBeHidden()
- toHaveClass()
- toHaveText()
- setFixtures() - this has just been copied over from `jasmine-jquery` instead of rewritten

This is being merged into `remove_jasmine_jquery_from_tests` first so there's no `CHANGELOG.md` commit.

## Why

The `jasmine-jquery` helper functions were causing https://github.com/alphagov/govuk_publishing_components/issues/2902 and so they have been replaced with a vanilla JS alternative. This also removes our tests' reliance on the `jasmine-jquery` library (which is no longer actively maintained).

### DIFF
--- a/spec/javascripts/components/cookie-banner-spec.js
+++ b/spec/javascripts/components/cookie-banner-spec.js
@@ -63,9 +63,9 @@ describe('Cookie banner', function () {
     var cookieBannerMain = document.querySelector('.js-banner-wrapper')
     var cookieBannerConfirmation = document.querySelector('.gem-c-cookie-banner__confirmation')
 
-    expect(window.isElementVisible(element)).toBeTruthy()
-    expect(window.isElementVisible(cookieBannerMain)).toBeTruthy()
-    expect(window.isElementHidden(cookieBannerConfirmation)).toEqual(true)
+    expect(window.specHelpers.isElementVisible(element)).toBeTruthy()
+    expect(window.specHelpers.isElementVisible(cookieBannerMain)).toBeTruthy()
+    expect(window.specHelpers.isElementHidden(cookieBannerConfirmation)).toEqual(true)
   })
 
   it('should show the cookie banner when preferences have not been actively set', function () {
@@ -77,9 +77,9 @@ describe('Cookie banner', function () {
     var cookieBannerMain = document.querySelector('.js-banner-wrapper')
     var cookieBannerConfirmation = document.querySelector('.gem-c-cookie-banner__confirmation')
 
-    expect(window.isElementVisible(element)).toBeTruthy()
-    expect(window.isElementVisible(cookieBannerMain)).toBeTruthy()
-    expect(window.isElementHidden(cookieBannerConfirmation)).toEqual(true)
+    expect(window.specHelpers.isElementVisible(element)).toBeTruthy()
+    expect(window.specHelpers.isElementVisible(cookieBannerMain)).toBeTruthy()
+    expect(window.specHelpers.isElementHidden(cookieBannerConfirmation)).toEqual(true)
   })
 
   it('should hide the cookie banner when preferences have been actively set', function () {
@@ -88,7 +88,7 @@ describe('Cookie banner', function () {
     var element = document.querySelector('[data-module="cookie-banner"]')
     new GOVUK.Modules.CookieBanner(element).init()
 
-    expect(window.isElementHidden(element)).toEqual(true)
+    expect(window.specHelpers.isElementHidden(element)).toEqual(true)
   })
 
   it('sets a default consent cookie', function () {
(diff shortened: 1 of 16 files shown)


spec/javascripts/helpers/SpecHelper.js
```diff
@@ -36,3 +36,19 @@ afterEach(function () {
...
+  return !this.isElementVisible(element)
+}
+
+function isClassOnElement(element, className) {
+    return element.classList.contains(className)
+}
+
+function setFixtures(html) {
+  return jasmine.getFixtures().proxyCallTo_('set', arguments)
+}
```
JamesCGDS, line 54: Just had a thought: could we put these functions into a `specHelpers` object? I think it might make the file look a little cleaner/help with readability.

AshGDS: Thanks @JamesCGDS , i've done this now

JamesCGDS (approved):
Looks good to me, nice work 👍

kevindew:
Amazing about removing more jQuery 🚀 

Just a quick FYI about the CHANGELOG.md, since this will just be an internal change and not one for consumers of the gem there shouldn't need to be any reference to this in the changelog.

spec/javascripts/components/cookie-banner-spec.js
```diff
@@ -161,15 +161,15 @@ describe('Cookie banner', function () {
     new GOVUK.Modules.CookieBanner(element).init()
 
     var acceptCookiesButton = document.querySelector('[data-accept-cookies]')
-    var mainCookieBanner = document.querySelector('.gem-c-cookie-banner__message')
+    var mainCookieBanner = document.querySelector('.js-banner-wrapper')
```
andysellick, line 164: Just curious why this class changed?

spec/javascripts/helpers/SpecHelper.js
```diff
@@ -40,3 +40,7 @@ function isElementVisible(element) {
 function isElementDisabled(selector) {
   return document.querySelector(selector+':disabled') !== null
 }
+
+var isElementHidden = function(element) {
```
andysellick: Why is `isElementHidden` declared differently from `isElementDisabled`? Fine if there's a reason, but otherwise would be good to be consistent (I think the two ways of declaring are basically the same anyway).

spec/javascripts/components/step-by-step-nav-spec.js
```diff
@@ -446,17 +446,19 @@ describe('A stepnav module', function () {
 
     it('opens the steps it has remembered', function () {
       var $step1 = $element.find('#topic-step-one')
-      expect($step1).toHaveClass('step-is-shown')
-      expect($step1.find('.js-panel')).not.toHaveClass('js-hidden')
+      expect(window.isClassOnElement($step1[0], 'step-is-shown')).toEqual(true)
+      var $jsPanel = $step1.find('.js-panel')
+      expect(window.isClassOnElement($jsPanel[0], 'js-hidden')).toEqual(false)
```
andysellick: Totally non-blocking suggestion here, but I think this could be tweaked to reduce the reliance on jQuery's `find` function.

```Javascript
var $jsPanel = $step1[0].querySelector('.js-panel')
expect(window.isClassOnElement($jsPanel, 'js-hidden')).toEqual(false)
```

spec/javascripts/components/toggle-input-class-on-focus-spec.js
```diff
@@ -23,11 +23,14 @@ describe('A toggle class module', function () {
 
     it('applies the focus style on focus and removes it on blur', function () {
       var searchInput = document.querySelector('.js-class-toggle')
-      expect(searchInput).not.toHaveClass('focus')
+      expect(window.isClassOnElement(searchInput, 'js-class-toggle')).toEqual(true)
```
andysellick: I'm not sure what this line is for? It seems to be testing that the element found on the previous line using the class `js-class-toggle` has that class? Wouldn't the previous line fail if that wasn't the case?

andysellick (commented):
Nice work. Just a few small questions but doubt there's anything blocking this.

spec/javascripts/helpers/SpecHelper.js
andysellick: I reviewed this commit by commit so I think this has already been addressed later in the PR.

andysellick:
Seems to be some fairly big visual changes - worth rebasing this PR against `main`, that should hopefully resolve them.

AshGDS:
> Seems to be some fairly big visual changes - worth rebasing this PR against `main`, that should hopefully resolve them.

We're going to merge into `remove_jasmine_jquery_from_tests` so I think that's why Percy is unhappy, it should pass when we merge to main :+1:

spec/javascripts/components/cookie-banner-spec.js
AshGDS: @andysellick I can change it back - basically it was checking the element was hidden but the actual thing that gets hidden is it's parent element. Before I used `!this.isElementVisible` to check if something is hidden the test was failing but I reckon it will still work if I change it back to `.gem-c-cookie-banner__message`

spec/javascripts/helpers/SpecHelper.js
AshGDS: @andysellick Yeah sorry I had used this declaration and then James had used a different declaration, so it only got standardised after I rebased.

spec/javascripts/components/step-by-step-nav-spec.js
AshGDS: Thanks @andysellick - yeah I guess as our priority was removing `jasmine-jquery` instead of jQuery itself we were being cautious about only removing jQuery when we had to. I can update it so it uses vanilla JS though.

AshGDS: @andysellick I think it's better we tackle this if we look at removing jQuery from the tests in the future, as it's a bit unrelated to this PR (hence why you said it's non-blocking.) Thanks for raising it though!

AshGDS:
@andysellick this should be ready for an approval now if everything is OK :+1:

spec/javascripts/components/toggle-input-class-on-focus-spec.js
AshGDS: @andysellick Would you like me to remove it?

AshGDS: Whoops, I didn't realise I introduced it. Will remove it now.

spec/javascripts/components/step-by-step-nav-spec.js
andysellick: Fair enough!

andysellick (approved):
Great work both! 👍 👏

PR #3131, opened 2022-12-12
Use setAttribute to add GA4 JSONs to step by step nav

AshGDS:
Hi @andysellick , would you be able to approve this? Thanks :+1:

## What
Use `setAttribute` to add GA4 JSON attributes to step by step nav.

## Why
This fixes an issue when the step title has a quotation mark in. The HTML broke when a step title had a quote `'` in it.

## Visual Changes

None.


app/views/govuk_publishing_components/components/docs/step_by_step_nav.yml
```diff
@@ -83,7 +83,7 @@ examples:
       ga4_tracking: true
       steps: [
         {
-          title: "Do something",
+          title: "This title's got quotation marks \" in it.",
```
andysellick, line 86: Not sure this is needed?

PR #3143, opened 2022-12-15
Change brand colour for No. 10 to be govuk-black

patrickpatrickpatrick:
## What

Change brand colour for No. 10 to be govuk-black instead of bright-purple.

## Why

As part of the new redesign of No. 10's page, the link colour is to be changed to black. The links are styled using brand__color, so there this needs to be updated to the new colour.

### Before

Screenshot 2022-12-15 at 10 52 21

### After

Screenshot 2022-12-15 at 10 52 12


chao-xian (approved):
Hold on merging while we get confirmation from Product but LGTM - thanks

PR #3145, opened 2022-12-15
Remove dependency on .js.erb and puppet for loading our analytics

AshGDS:
Hi @andysellick / @JamesCGDS - would you be able to approve this if all is OK? Thanks :+1:

## What

This change creates a JavaScript file which will load in our analytics variables depending on the current domain that the user is on.  I have created `load-analytics.js`.  This contains a `loadUa()` and `loadGa4()` function (they're separate due to a suggestion from @andysellick  and it makes sense as it will be easier to remove UA.)

It runs both of these functions inside of a `DOMContentLoaded` event listener. This is because the `<body>` does not exist yet when `load-analytics.js` code initially runs in the browser, so we have to wait for the DOM to finish loading first before initialising our tracking code otherwise we get errors, because our tracking code tries to add event listeners on the body for things like event tracking. 

`loadUa()` and `loadGa4()` are quite similar. They essentially:

1. Grab the current domain that the code is running on using `window.location.hostname`.
2. Determine the environment you are on by comparing your domain against hardcoded categorised domains.
3. Set the correct environment variables depending on the environment you are on. These env vars are also hardcoded now.
4. Kick off the initialisation of analytics by running their respective `init()` functions, which will then run the `ga4-core.js` code and the UA tracking initialisation code.
5. The `ga4-core.js` / UA code then makes use of these environment variables. It can access them as they are set on the JavaScript `window` object.

I have moved grabbing the current version of publishing components (`gem_version` in the dataLayer) to a meta tag. Now that we can't use `.js.erb`, it's not possible to get the gem version using .js.erb. Therefore I've turned it into a `<meta` tag, and then I grab that using JavaScript. I did consider writing a `<script>` tag in the Ruby template which would be `window.GOVUK.analyticsGa4.vars.gem_version = "<%= GovukPublishingComponents::VERSION %>"` but I thought this was a bit messy writing JavaScript inside the template. Also it would be the first JavaScript that runs on the page, so I'd have to do things like `window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}` to prevent errors with `window.GOVUK` and `analyticsGa4` and `.vars` not yet existing.

### Testing the code

To test this repo, you will need to:
-  checkout the branch `remove-analytics-puppet-dependency` on `static`
- modify the `Gemfile` to use a local version of `govuk_publishing_components` in `static`
-  run `make static && govuk-docker up static-app` 

We need to use `static` to test because the analytics is executed inside the GOVUK `<head>` template, so you need an application which is using this GOVUK `<head>`. So you should be able to see the development GTM script load as well as the `dataLayer` on this page http://static.dev.gov.uk/templates/gem_layout.html.erb  

I'm having a caching issue with my `static` which kept making the code not run properly. If this happens to you then:
- stop static
- run `bundle exec rails assets:precompile` in static
- run `make static && govuk-docker up static-app` again

If you'd like to compare the environment variables  in `govuk-puppet` to check that I've hardcoded the right ones, these are the sources I used: 

- [Production](https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/production.yaml#L150)
- [Staging](https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/staging.yaml#L223)
- [Integration](https://github.com/alphagov/govuk-puppet/blob/main/hieradata_aws/integration.yaml#L31)
- [Universal Analytics' Production values](https://github.com/alphagov/static/blob/main/config/environments/production.rb#L87)
- [Universal Analytics' "UA-UNSET" values ](https://github.com/alphagov/static/blob/main/config/application.rb#L41)
- The dev GTM environment variables are from Ian. Moving forward we could use this to save time when testing, as our events will now get sent to the a development dashboard. Therefore we can show analysts the dataLayer pushes sent from our local environments, without having to put stuff on Integration first.

## Why

The previous implementation would use `static`. We are moving away from `static`. Also, it would require grabbing environment variables from `govuk-puppet`, and passing them through to JavaScript using a `.js.erb` file. We cannot use `govuk-puppet` anymore due to the new infrastructure and `.js.erb` is a deprecated filetype in Rails.

I am adding `loading-analytics.js` to the GOVUK `<head>`. Most of the other JavaScript files in this repo are initialised in `static` instead. Since the purpose of this PR is to completely remove our reliance on `static`, I'm not going to define the code in this repo and then execute it in `static` otherwise it kind of defeats the purpose of moving the code outside of `static` in the first place.

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+          gemVersion = 'not found'
+        }
+
+        window.GOVUK.analyticsGa4.vars.gem_version = gemVersion
+        window.GOVUK.analyticsGa4.init()
+      }
+    }
+  },
+
+  arrayContains: function (valueToFind, array) {
```
AshGDS, line 224: Do you think it is pointless having this function? It might be more readable if I just did

`if(productionDomains.indexOf(currentDomain) !== 1)` etc. in the code.

AshGDS:
Once this is approved I can modify the `static` PR https://github.com/alphagov/static/pull/2959/files to just be a PR that removes the code in `static` that we don't need anymore

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+    development: {
+      initialiseGA4: true,
+      auth: 'bRiZ-jiEHtw6hHpGd6dF9w',
+      preview: 'env-3'
+    }
+  }
+}
+
+// This needs to run in a DOMContentLoaded function otherwise tracking code crashes due to the <body> not existing.
+window.addEventListener('DOMContentLoaded', function () {
```
andysellick, line 252: I'm slightly hesitant over the use of `DOMContentLoaded` here as making this work reliably on older browsers is a bit of a juggling act. Having said that, this seems to work fine in IE9 and 11 (can't currently check 10 for some browserstack reason).

We do already have a DOMContentLoaded call in https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/dependencies.js so I'm wondering if it's worth combining the two? I have a long standing issue to resolve to remove that one, so putting them together would help with that. No worries if this is too complex.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+    } else if (this.arrayContains(currentDomain, this.integrationDomains)) {
+      environment = 'integration'
+    } else if (this.arrayContains(currentDomain, this.developmentDomains) || currentDomain.indexOf('.dev.gov.uk') !== -1) {
+      environment = 'development'
+    }
+
+    // If we recognise the environment (i.e. the string isn't empty), load in GA4
+    if (environment) {
+      // If analytics-ga4.js exists and our detected environment has 'initialiseGA4' set to true, load GA4.
+      if (typeof window.GOVUK.analyticsGa4.init !== 'undefined' && this.ga4EnvironmentVariables[environment].initialiseGA4) {
```
andysellick, line 203: I notice that all of the environments have `initialiseGA4` set to `true`, which seems like unnecessary code. I assume this `if` statement is basically checking before proceeding that the environment is valid? Could it not just do a check for `this.ga4EnvironmentVariables[environment]`?

andysellick: Looking at the code more, there's the initial check on line 201 that `environment` exists, are both of these checks necessary? Could they be combined somehow?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+        window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+        window.GOVUK.analyticsGa4.vars = window.GOVUK.analyticsGa4.vars || {}
+        window.GOVUK.analyticsGa4.vars.id = this.ga4EnvironmentVariables.id
```
andysellick, line 206: If `id` is consistent across all environments, and we've already checked that the environment is valid, can the ID not just be set here, rather than fetched from the object?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
andysellick: I think it's okay - it makes the code more readable. I was wondering if it could be renamed to be something even clearer, like maybe `if (theEnvironmentIs(this.staging, currentDomain))` but that might not be necessary.

docs/load-analytics.md
```diff
@@ -0,0 +1,17 @@
...
+5. The `ga4-core.js` / UA code then makes use of these environment variables. It can access them as they are set on the JavaScript `window` object.
+
+The Google Analytics 4 environment variables are:
+
+- `id` - this is the Google Tag Manager ID used to identify our organisation so that analytics data is sent to the right account.
+- `preview` - this is an optional variable that is used to determine what GTM environment the analytics data is sent to. If this is empty, analytics data is sent to the production dashboard. If you wanted to have your integration analytics to an 'integration' dashboard, you would populate this variable with the relevant value.
+- `auth` - this is used to authenticate analytics data being sent to another dashboard, such as an 'integration' dashboard. As it is called `auth` you may think this should be a secret environment variable. As this is just used to send data to another dashboard, the only risk when it is exposed is that someone could send fake analytics data to our integration/staging/development dashboards. We do not consider this issue big enough to make these secret as we only use these dashboards for testing/debugging purposes anyway.
```
andysellick, line 17: Nice documenting. I'd suggest not going into so much detail about the risks involved here (we don't want to give people ideas), I think it's probably enough to say that these variables can be determined through the browser by anyone smart enough so they're not classified as secret.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
+/* eslint-env jasmine */
+
+var linkedDomains = [
```
andysellick, line 3: Do we need all of these linkedDomains listed in the test? Is there not a way to fake them in the test for comparison purposes?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
...
+
+var expected
+var savedUaVars
+var savedGa4Vars
+
+describe('Analytics loading', function () {
+  beforeAll(function () {
+    savedUaVars = window.GOVUK.extendObject(window.GOVUK.analyticsVars)
+    window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+    savedGa4Vars = window.GOVUK.extendObject(window.GOVUK.analyticsGa4.vars)
```
andysellick, line 145: I'm wondering if there's an easier way of saving and switching around the stored variables here. Presumably some of this effort is to avoid impacting other tests? Could the spec helper by default call the analytics loading with a dud environment, to prevent it loading everywhere else? Could the DOMContentLoaded be wrapped in some logic relating to the environment instead of containing it? Could `window.GOVUK.loadAnalytics` be called only in these tests, and then the relevant window variables be deleted afterwards?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
...
+        var domain = domains[i]
+        window.GOVUK.loadAnalytics.loadGa4(domain)
+        expect(window.GOVUK.analyticsGa4.vars.environment).toEqual('production')
+      }
+    })
+
+    it('loads GA4 on staging domains', function () {
+      var domains = [
+        'www.staging.publishing.service.gov.uk',
+        'www.staging.publishing.service.gov.uk',
```
andysellick, line 182: Seems to be duplicate entries here again?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
...
+        window.GOVUK.analyticsGa4.vars = null
+        var domain = domains[i]
+        window.GOVUK.loadAnalytics.loadGa4(domain)
+        expect(window.GOVUK.analyticsGa4.vars).toEqual(null)
+      }
+    })
+
+    it('sets the correct production variables', function () {
+      window.GOVUK.loadAnalytics.loadGa4('www.gov.uk')
+      delete window.GOVUK.analyticsGa4.vars.internalDomains
```
andysellick, line 236: Why delete `internalDomains` here? Could you add `linkedDomains` to `expected`?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
...
+      expect(window.GOVUK.analyticsGa4.vars).toEqual(expected)
+    })
+
+    it('sets the correct staging variables', function () {
```
andysellick, line 240: These tests and the ones above (`loads GA4 on x domains`) seem to call the same bit of code, would it be worth combining them or is there a reason they're separated out?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
```diff
@@ -0,0 +1,323 @@
...
+      expected.gaProperty = 'UA-26179049-22'
+      window.GOVUK.loadAnalytics.loadUa('www.integration.publishing.service.gov.uk')
+      expect(window.GOVUK.analyticsVars).toEqual(expected)
+    })
+
+    it('loads the correct UA values on development', function () {
+      expected.gaProperty = 'UA-UNSET'
+      expected.gaPropertyCrossDomain = 'UA-UNSET'
+      window.GOVUK.loadAnalytics.loadUa('localhost')
+      expect(window.GOVUK.analyticsVars).toEqual(expected)
```
andysellick, line 308: This test has highlighted to me that the linkedDomains appear to be loaded on development and unrecognised domains - should we be preventing that?

spec/javascripts/helpers/SpecHelper.js
```diff
@@ -1,6 +1,7 @@
 /* eslint-env jasmine, jquery */
 
 beforeAll(function () {
+
```
andysellick, line 4: Nitpick: new empty line?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
+//= require govuk_publishing_components/analytics
+//= require govuk_publishing_components/analytics-ga4
+
+window.GOVUK.loadAnalytics = {
+
```
andysellick, line 5: Nitpick: empty line needed?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+  productionDomains: [
+    'www.gov.uk',
+    'www-origin.publishing.service.gov.uk'
+  ],
+  stagingDomains: [
+    'www.staging.publishing.service.gov.uk',
+    'www.staging.publishing.service.gov.uk',
```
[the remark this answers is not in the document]
AshGDS: Thanks - yep it's meant to be [www-origin.staging.publishing.service.gov.uk](http://www-origin.staging.publishing.service.gov.uk/) etc.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+    'www-origin.integration.publishing.service.gov.uk'
+  ],
+  developmentDomains: [
+    'localhost', '127.0.0.1', '0.0.0.0'
+  ],
+
+  loadUa: function (currentDomain) {
+    currentDomain = currentDomain || window.location.hostname
+
+    // Universal analytics data
```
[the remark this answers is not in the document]
AshGDS: I can make the comment more specific - it was more to signal that the domains are unrelated to determining what Environment we're on

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+      'view-immigration-status.service.gov.uk',
+      'visa-address-update.service.gov.uk',
+      'visas-immigration.service.gov.uk',
+      'your-defra-account.defra.gov.uk'
+    ]
+
+    // Universal Analytics variables
+    window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
+    window.GOVUK.analyticsVars.gaProperty = 'UA-UNSET'
+    window.GOVUK.analyticsVars.gaPropertyCrossDomain = 'UA-145652997-1'
```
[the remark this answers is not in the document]
AshGDS: `gaPropertyCrossDomain` is also used on Staging and Integration, so I had written it in this way to reduce duplication. Therefore I'd be repeating the value in the Production/Staging/Integration if statement. Should I still change it?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+    window.GOVUK.analyticsVars.primaryLinkedDomains = ['account.gov.uk']
+    window.GOVUK.analyticsVars.linkedDomains = linkedDomains
+
+    if (this.arrayContains(currentDomain, this.productionDomains)) {
+      window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-1'
+    } else if (this.arrayContains(currentDomain, this.stagingDomains)) {
+      window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-20'
+    } else if (this.arrayContains(currentDomain, this.integrationDomains)) {
+      window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-22'
```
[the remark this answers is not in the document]
AshGDS: @andysellick They're from puppet - I went onto Staging and Integration and did `console.log(window.GOVUK.analyticsVars` to grab them.

E.g. https://github.com/alphagov/govuk-puppet/blob/a60a4cadfae8b89fec1f0777397746c388be7788/hieradata_aws/staging.yaml#L222

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+    if (typeof window.GOVUK.analyticsInit !== 'undefined') {
+      window.GOVUK.analyticsInit()
+    }
+  },
+
+  loadGa4: function (currentDomain) {
+    currentDomain = currentDomain || window.location.hostname
+    var environment = ''
+
+    // Categorise current environment and set a couple UA variables
```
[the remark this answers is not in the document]
AshGDS: Thanks, will update - this is left over from when both were being set in one function.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: @andysellick Depending on the order that the tests are run, either `internalDomains` exists or it doesn't - basically it depends on when the linkTracker tests run. As it isn't relevant to these tests I've deleted it.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+        window.GOVUK.analyticsGa4.init()
+      }
+    }
+  },
+
+  arrayContains: function (valueToFind, array) {
+    return array.indexOf(valueToFind) !== -1
+  },
+
+  ga4EnvironmentVariables: {
```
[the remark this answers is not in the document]
AshGDS: Will do :+1:

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,261 @@
...
+        window.GOVUK.analyticsGa4.vars.environment = environment // Used for testing and debugging
+
+        var gemVersion = document.querySelector('meta[name="govuk:components_gem_version"]')
+        if (gemVersion) {
+          gemVersion = gemVersion.getAttribute('content')
+        } else {
+          gemVersion = 'not found'
+        }
+
+        window.GOVUK.analyticsGa4.vars.gem_version = gemVersion
```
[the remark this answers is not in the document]
AshGDS: Thanks @andysellick  :+1:

app/assets/javascripts/govuk_publishing_components/load-analytics.js
AshGDS: @andysellick I had added `initialiseGA4` so that we had a method in place to turn GA4 on and off on each environment which didn't require us to have to remove/append code. I'm guessing we might not need that anymore as that was for when we initially deployed GA4. But I was just thinking if we ever needed to turn GA4 off on one environment, having that boolean would be handy. What do you think?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
AshGDS: @andysellick I guess so - my thinking was it would be more readable to keep all the environment variables in one object.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: @andysellick I'd need to move the `linkedDomains` variable outside of `loadUa()` and directly onto `window.GOVUK.loadAnalytics` so that I can modify it in the test. I'll do that.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: Yep this code is to prevent it affecting the other tests - the tests were going haywire and the only way I could fix it was to preserve the previous analytics vars as the other tests seem to be dependent on them being structured a certain way.

I'm not sure what you mean with the suggestions - I don't think there's a way to determine if you're running jasmine tests which would stop `window.GOVUK.loadAnalytics` running as Jasmine is running on `localhost` so it would load the development UA/GTM variables onto `window.GOVUK`. 

If you called `window.GOVUK.loadAnalytics` and it set the variables, and then you delete them afterwards, the other tests will fail (mainly the UA ones) because they need something set on `window.GOVUK.analyticsVars` so if I deleted them they will fail. That's why I set them to their previous value.

Does that make sense? You might have to provide the solution you're proposing here as I'm not sure what the alternative would be. Thanks

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: Yep - meant to be [www-origin.staging.publishing.service.gov.uk](http://www-origin.staging.publishing.service.gov.uk/), good spot

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: @andysellick I believe it does that in our current implementation, which is why I kept it like that.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: @andysellick I think it's a bit more readable/clean having them in separate functions, especially because when the individual test fails you know exactly which part of the code is failing. Do you think it's worth combining them?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
AshGDS: @andysellick I believe `linkedDomains` is loaded on development/unrecognised domains in our current environment, which is why I've kept it the same, as I don't want to affect how universal analytics works.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
AshGDS: Thanks @andysellick - I've moved it there now.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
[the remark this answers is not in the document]
andysellick: Ah yes I see now. That's a good reason, but I think the difference would only be a couple of lines. Let me know what you think. For comparison, here's the suggestion:

```
// Universal Analytics variables
window.GOVUK.analyticsVars = window.GOVUK.analyticsVars || {}
window.GOVUK.analyticsVars.primaryLinkedDomains = ['account.gov.uk']
window.GOVUK.analyticsVars.linkedDomains = this.linkedDomains
window.GOVUK.analyticsVars.gaProperty = 'UA-UNSET'
window.GOVUK.analyticsVars.gaPropertyCrossDomain = 'UA-UNSET'

if (this.arrayContains(currentDomain, this.productionDomains)) {
  window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-1'
  window.GOVUK.analyticsVars.gaPropertyCrossDomain = 'UA-145652997-1'
} else if (this.arrayContains(currentDomain, this.stagingDomains)) {
  window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-20'
  window.GOVUK.analyticsVars.gaPropertyCrossDomain = 'UA-145652997-1'
} else if (this.arrayContains(currentDomain, this.integrationDomains)) {
  window.GOVUK.analyticsVars.gaProperty = 'UA-26179049-22'
  window.GOVUK.analyticsVars.gaPropertyCrossDomain = 'UA-145652997-1'
}
```

app/assets/javascripts/govuk_publishing_components/load-analytics.js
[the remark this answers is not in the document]
andysellick: Nice work 👍

app/assets/javascripts/govuk_publishing_components/load-analytics.js
andysellick: That seems like good reason enough. Could you add a comment to that section to highlight this feature?

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
andysellick: No, that makes sense, thanks.

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
andysellick: Okay, thanks.

AshGDS:
Thanks @andysellick - I've made the requested changes :+1:

spec/javascripts/govuk_publishing_components/load-analytics.spec.js
andysellick: It's been three weeks since I wrote that comment and I'm not sure what I was saying either! Seems good to me.

docs/analytics/analytics.md
```diff
@@ -170,7 +170,7 @@ named tracker, and sessions will persist to the other domain.
 
 Once a service starts sending analytics data to the shared GA property and has configured their system to include `www.gov.uk` in their cross domain linking, we must also configure `www.gov.uk` to include their domain in our cross linker configuration.
 
-Add the domain of the service to [`init.js.erb`](https://github.com/alphagov/static/blob/main/app/assets/javascripts/analytics.js.erb).
+Add the domain of the service to [load-analytics.js](https://github.com/alphagov/static/blob/main/app/assets/javascripts/govuk_publishing_components/load-analytics.js).
```
andysellick, line 173: I think this link is wrong?

spec/support/jasmine-browser.json
```diff
@@ -5,6 +5,7 @@
     "govuk_publishing_components/all_components-*.js",
     "govuk_publishing_components/modules-*.js",
     "govuk_publishing_components/analytics-*.js",
+    "govuk_publishing_components/load-analytics-*.js",
```
andysellick, line 8: This is interesting - does `load-analytics` need the `-*` on the end or were you just following the convention? Does it work without?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -0,0 +1,244 @@
...
+  integrationDomains: [
+    'www.integration.publishing.service.gov.uk',
+    'www-origin.integration.publishing.service.gov.uk'
+  ],
+  developmentDomains: [
+    'localhost', '127.0.0.1', '0.0.0.0'
+  ],
+
+  // For universal analytics - cross domain tracking
+  linkedDomains: [
```
andysellick, line 22: I'm wondering whether it would make this code easier to navigate by moving `linkedDomains` into its own separate file and importing it somehow? Non-blocking, we could do this later.

docs/analytics/analytics.md
AshGDS: Whoops, yeah it's linking to static

spec/support/jasmine-browser.json
AshGDS: Yeah, the tests fail without the `-*`, not sure why though.

docs/analytics/analytics.md
```diff
@@ -170,7 +170,7 @@ named tracker, and sessions will persist to the other domain.
 
 Once a service starts sending analytics data to the shared GA property and has configured their system to include `www.gov.uk` in their cross domain linking, we must also configure `www.gov.uk` to include their domain in our cross linker configuration.
 
-Add the domain of the service to [load-analytics.js](https://github.com/alphagov/static/blob/main/app/assets/javascripts/govuk_publishing_components/load-analytics.js).
+Add the domain of the service to [linked-domains.js](https://github.com/alphagov/govuk_publishing_components/tree/main/app/assets/javascripts/govuk_publishing_components/analytics/linked-domains.js).
```
AshGDS, line 173: This link won't work until the PR is merged as it's linking to the `main` branch.

AshGDS:
Thanks @andysellick - I pushed a new commit, it should be ready for another review :+1:

spec/support/jasmine-browser.json
andysellick: Fair enough. Some Rails magic, I guess.

andysellick (approved):
Niiiice. We should talk about how to deploy this - presumably there's a change in `static` that needs to be co-ordinated.

AshGDS:
Thanks @andysellick - I think merging this branch by itself wouldn't break anything as `ga4-core` is unchanged, so `ga4-core` would just check for `window.GOVUK.analyticsGa4.vars.id` regardless of where it was defined and load GA4 in the same way, so I don't think we'd need to coordinate a release with `static`. Hence why this isn't flagged a breaking change. 

With that being said, should we merge/deploy it?

This PR https://github.com/alphagov/static/pull/2959 will remove the stuff in `static` that we don't need anymore.

andysellick:
Yes, but wouldn't deploying this change without the corresponding change in `static` mean that the analytics code attempts to initialise itself twice? Well, we should definitely ensure your changes to `static` are deployed at the same time as this, I'll review.

andysellick:
Don't forget to squash your commits 👍

PR #3152, opened 2022-12-21
Add ga4 tracking to header links

AshGDS:
Hi @andysellick / @JamesCGDS - no rush due to xmas, but would you be able to approve this? Thanks :+1:

## What

Adds GA4 tracking to our header links using the `ga4-link-tracker`

## Why

It is next on our list of interactions to add.

## Visual Changes

None


app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -28,6 +28,14 @@
...
         id="logo"
         title="<%= logo_link_title %>"
+        data-ga4-link="<%= {
+          "event_name": "navigation",
+          "type": "header menu bar",
+          "external": "false",
+          "text": "GOV.UK",
+          "section": "Logo",
+          "index": 0
+        }.to_json %>"
```
JamesCGDS, line 38: Do we need to add the `index_total` property here?

AshGDS:
Thanks @JamesCGDS - yes it needs an `index_total` and we realised that we need to add the tracking to the links inside of the Search Bar as well. 

So I'll fix that - @andysellick hold off on reviewing it for now!

JamesCGDS (changes requested):
Looks good, just have one change request - would it be possible to add tracking to popular links as well?

AshGDS:
Thanks @JamesCGDS  - yep we need to add the tracking to the links inside of the Search Bar as well. 

So I'll fix that - @andysellick hold off on reviewing it for now!

AshGDS:
@JamesCGDS I've added the `index_total` to the `GOV.UK` logo, and have added tracking to the `Popular links` section in the header search dropdown.

I've also updated the indexes as I had implemented them incorrectly. I was using whole numbers e.g. `0-29`, When in fact the implentation guide calls for decimal indexes such as `1.1.1`

- The first number is the section number, which is basically the dropdown the link is in. So anything under "Menu" is a section (`1.X.X`) and then the links under "Search" are another section (`2.2.X`).
- The second number is the "sub section" number, which is basically which heading it's under. So in the Menu dropdown, anything under the Topics heading is `1.1.X` and anything under the Government Activity heading is `1.2.X`
- The third number is the link index within the sub section.

@andysellick @JamesCGDS Should be ready to review now. Thanks :+1:

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -158,6 +173,13 @@
                                   track_label: item[:href],
                                   track_dimension: item[:label],
                                   track_dimension_index: "29",
+                                  ga4_link: {
+                                      "event_name": "navigation",
+                                      "type": "header menu bar",
+                                      "index": '1.' + (column_index + 1).to_s + '.' + (index + 1).to_s,
```
andysellick, line 179: Instead of calling `.to_s` twice, I think you could do

```Ruby
"index": "1.#{column_index + 1}.#{index + 1}",
```

(similar for line 320)

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -292,6 +314,13 @@
                         track_label: popular_link[:href],
                         track_dimension: popular_link[:label],
                         track_dimension_index: "29",
+                        ga4_link: {
```
andysellick, line 317: Is there a test that can be written to check these attributes are set correctly?

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
AshGDS: Thanks :+1:

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
AshGDS: @andysellick Good spot, I've written a ruby test now :+1:

AshGDS:
Thanks @andysellick - I've made those changes

JamesCGDS (approved):
Looks good to me 👍

andysellick (approved):

andysellick:
I don't think there should be any visual changes with this PR - try rebasing against main to resolve this.

AshGDS:
Thanks @andysellick - It is actually already rebased with `main` - should we just approve the Percy change?

PR #3154, opened 2022-12-21
Add ga4-link-tracker to image card

JamesCGDS:
## What
This PR adds `ga4-link-tracker` to the `data-module` attribute on the `_image_card` component.

## Why
We are adding tracking to links on the homepage, which uses image cards.

[Trello card](https://trello.com/c/WAST3s6v/392-add-tracking-homepage)

## Visual Changes
N/A

### DIFF
--- a/app/views/govuk_publishing_components/components/_image_card.html.erb
+++ b/app/views/govuk_publishing_components/components/_image_card.html.erb
@@ -26,7 +26,7 @@
   extra_link_classes << brand_helper.color_class
 
   data_modules = %w[]
-  data_modules << "gem-track-click" if card_helper.is_tracking?
+  data_modules << "gem-track-click ga4-link-tracker" if card_helper.is_tracking?
   data_modules << "image-card" if card_helper.youtube_video_id
 %>
 <% if card_helper.href || card_helper.extra_details.any? %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Catch errors when modules initialised ([PR #3190](https://github.com/alphagov/govuk_publishing_components/pull/3190))
 * Add translation strings for cookie banner confirmation message ([PR #3191](https://github.com/alphagov/govuk_publishing_components/pull/3191))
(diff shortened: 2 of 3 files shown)


app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -25,9 +25,13 @@
 
 %>
 <% if card_helper.href || card_helper.extra_details.any? %>
-  <div class="<%= classes %> <%= brand_helper.brand_class %>"
-    <%= "data-module=gem-track-click" if card_helper.is_tracking? %>
-    <%= "lang=#{card_helper.lang}" if card_helper.lang %>>
+  <%
+    classes = "#{classes} #{brand_helper.brand_class}"
+    data_attributes = {}
```
andysellick, line 30: This is good. I wonder if it's worth making it more flexible while we're at it (but feel free to disagree) so we could pass anything as data attributes, something like this?

```
data_attributes ||= {}
((data_attributes[:module] ||= "") << " " << "gem-track-click ga4-link-tracker").strip! if card_helper.is_tracking?
```

I've lifted this from the accordion component. Looking at it again I think it might be possible to simplify it to `((data_attributes[:module] ||= "") << " gem-track-click ga4-link-tracker").strip! if card_helper.is_tracking?`.

Obviously this would require adding more tests, and a doc entry, so again, feel free to disagree with me.

andysellick (approved):

app/views/govuk_publishing_components/components/_image_card.html.erb
JamesCGDS: Has [3156](https://github.com/alphagov/govuk_publishing_components/pull/3156) handled this? If not, could I look at implementing this in a separate branch? Only because I've been holding this PR up for a while and it would be good to get it merged

andysellick: Don't worry about this suggestion, I'm putting together a spike on something that I think will be able to to deal with this.

PR #3156, opened 2022-12-23
Add Youtube Embed Support to the ImageCard Component

patrickpatrickpatrick:
## What

New variant of image card that supports embedding a Youtube video.

### More Details

ImageCardHelper modified to accept a youtube_video_id and a set of new functions for generating a Youtube embed and a placeholder link using a thumbnail of the Youtube embed.

YoutubeLinkEnhancement can accept an override class if you want an alternative class to be appended to the youtube embed that isn't `gem-c-govspeak__youtube-video`.

## Why

As part of the new design for No. 10 organisation page we need to be able to show promotional content with embeddable videos. Promotional content currently uses the image card component, so this PR adds support for embedding specifically Youtube videos into the image card component. Youtube embed functionality already exists in govspeak, therefore we are using that to generate the embed as it already has the functionality we need.

### Screenshot

Screenshot 2022-12-28 at 15 52 40


chao-xian:
@patrickpatrickpatrick lovely work. Small thing: I'd split this commit 1d83fb6d712a36c4d353f4daf168e5bf5f60f25f first into one where you refactor to split the partial, and then add in the youtube bits

andysellick:
Obviously still draft so I won't do a full review but one concern I'd like to raise. At the moment the component auditing relies on the fact that component files use the same name for themselves (i.e. the image card component has an image card template, image card scss file, etc.) in order to report accurate information, but this change looks like it might break that convention. Would it be possible to have the filenames for the various parts of this match the component names in order to preserve how the auditing works? Happy to discuss if this isn't clear.

patrickpatrickpatrick:
> Obviously still draft so I won't do a full review but one concern I'd like to raise. At the moment the component auditing relies on the fact that component files use the same name for themselves (i.e. the image card component has an image card template, image card scss file, etc.) in order to report accurate information, but this change looks like it might break that convention. Would it be possible to have the filenames for the various parts of this match the component names in order to preserve how the auditing works? Happy to discuss if this isn't clear.

I believe this should be the case in the way I've implemented it. Youtube card has a youtube_card.html.erb, yotube_card.js, youtube_card.scss; Image card has a image_card.html.erb, image_card.scss and media card has a media_card.html.erb and media_card.scss.

andysellick:
Thanks for the offline chat @patrickpatrickpatrick I think I understand this better now. As discussed, I've plugged this branch into an application to test how the auditing responds.

The auditing isn't detecting that the image card and youtube card components contain media card, and therefore doesn't correctly report problems to do with assets (e.g. not including the media card scss). It does allow for components that contain other components, but in this case the media card is considered a partial and not a top level component - it's importing `/components/media_card/media_card` (the pattern for a partial) and not `/components/media_card` (the pattern for a separate component). Unfortunately this distinction has to be made otherwise the auditing has no way of determining the difference between a component (where it should warn about missing assets) and a partial (where it shouldn't warn about missing assets) and of course we don't want to make the media card a separate component.

There's a similar problem with the 'suggested imports for this application' functionality (which appears on the front page of the component guide inside an application). 

Not sure what to do about this - we don't want to make the media card its own component, but image card and youtube card rely on the assets from media card. Some options:

- this could be solved both by an adjustment in this PR and a change to the auditing but I'd have to look into how it works, can't guarantee anything
- could the image card and youtube card import the relevant Sass directly? I'm not sure, there might be a problem of duplication
- totally crazy idea: what if there was just one component, but depending on what you pass it, it works either as the image card or the youtube card? You'd keep the helpers almost as they are, but the component would figure out which was being called then include the relevant partial (image card or youtube card). The Sass could be combined again and the whole problem could be avoided.

(I think this will be ultimately solved by the work being done where components include their own assets, but that's a little while off yet unfortunately)

lib/govuk_publishing_components/presenters/image_card_helper.rb
```diff
@@ -23,6 +23,23 @@ def initialize(local_assigns)
...
+        @youtube_video_id = local_assigns[:youtube_video_id] || nil
+      end
+
+      def href
+        if @youtube_video_id
+          youtube_href
+        else
+          @href
+        end
+      end
```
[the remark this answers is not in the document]
andysellick: (same for the `media` method)

lib/govuk_publishing_components/presenters/image_card_helper.rb
```diff
@@ -75,6 +100,45 @@ def context
...
+        figcaption_class << " brand__color" if @brand
+        thumbnail = content_tag(:div, class: "gem-c-image-card__youtube-thumbnail-image-container") do
+          image_tag(
+            youtube_thumbnail_url,
+            class: "gem-c-image-card__image gem-c-image-card__youtube-thumbnail-image",
+            alt: "",
+            loading: @image_loading,
+          )
+        end
+        thumbnail + content_tag(:figcaption, @image_alt, class: figcaption_class)
```
andysellick, line 123: Not 100% sure this will work, but I think you can avoid declaring thumbnail and streamline the end of this method by doing

```Ruby
def youtube_thumbnail
  figcaption_class = "gem-c-image-card__youtube-thumbnail-container-text govuk-link"
  figcaption_class << " brand__color" if @brand
  content_tag(:div, class: "gem-c-image-card__youtube-thumbnail-image-container") do
    image_tag(
      youtube_thumbnail_url,
      class: "gem-c-image-card__image gem-c-image-card__youtube-thumbnail-image",
      alt: "",
      loading: @image_loading,
    )
    content_tag(:figcaption, @image_alt, class: figcaption_class)
  end
end
```

app/views/govuk_publishing_components/components/docs/image_card.yml
```diff
@@ -205,6 +205,47 @@ examples:
       <div class="govuk-!-width-full">
         <%= component %>
       </div>
+  youtube_video_with_brand:
+    data:
+      brand: "department-for-work-pensions"
+      youtube_video_id: "XnP6jaK7ZAY"
+      href: "/still-not-a-page"
+      image_src: "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/91397/s712_SG_Swear_in_1_.jpg"
+      image_alt: "The vision for GOV.UK"
```
andysellick, line 214: For these examples I think `image_alt` is used but am I right in thinking `image_src` is ignored and could be removed?

app/assets/javascripts/govuk_publishing_components/components/image-card.js
```diff
@@ -0,0 +1,19 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function ImageCard ($module) {
```
andysellick, line 5: Nice addition, just wondering if we need to highlight when this gets released that this will need including in the assets for any application using the image card. I'm wondering if a breaking change is appropriate - what's the worst that can happen if this JS weren't included in an application using the image card with a Youtube video?

app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -27,6 +27,7 @@
 <% if card_helper.href || card_helper.extra_details.any? %>
   <div class="<%= classes %> <%= brand_helper.brand_class %>"
     <%= "data-module=gem-track-click" if card_helper.is_tracking? %>
+    <%= "data-module=image-card" if card_helper.youtube_video_id %>
```
andysellick, line 30: If tracking is enabled (on the previous line) then `data-module` is being declared twice for a youtube video. Will this work or will browsers ignore the second one? This might need to combine these two declarations.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
```diff
@@ -208,6 +208,46 @@
...
+  filter: brightness(.9);
+}
+
+.gem-c-image-card__youtube-thumbnail-container-text {
+  margin-top: 10px;
+}
+
+.gem-c-image-card__youtube-thumbnail-container:focus .gem-c-image-card__youtube-thumbnail-container-text {
+  @include govuk-focused-text;
+}
```
andysellick, line 249: Not sure where to raise this so not necessarily relevant to this file. 

This might not be something we have control over, but although getting the video to play using the keyboard is possible it is quite fiddly. You need to tab three times into the video element, and only the second focus state is visible (it's a blue box around the 'More' icon, see below). Is there anything we can do to make this more accessible for keyboard users? I've also yet to try this with a screen reader, which we should also consider - again, appreciate we're working with something embedded from another site, so it may not be possible (although that might raise questions about how we add youtube videos to GOV.UK).

Screenshot 2023-01-06 at 14 15 59

spec/components/image_card_spec.rb
```diff
@@ -5,7 +5,7 @@ def component_name
     "image_card"
   end
 
-  it "renders nothing when no link provided" do
+  it "renders nothing when no link or youtube video id provided" do
```
andysellick, line 8: I'd include these tests with the commit containing the relevant code changes.

lib/govuk_publishing_components/presenters/image_card_helper.rb
```diff
@@ -75,6 +100,45 @@ def context
...
+        "https://www.youtube.com/watch?v=#{@youtube_video_id}"
+      end
+
+      def youtube_thumbnail_url
+        "https://img.youtube.com/vi/#{@youtube_video_id}/maxresdefault.jpg"
+      end
+
+      def youtube_thumbnail
+        figcaption_class = "gem-c-image-card__youtube-thumbnail-container-text govuk-link"
+        figcaption_class << " brand__color" if @brand
```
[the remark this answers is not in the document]
patrickpatrickpatrick: The examples listed here are if the BrandHelper is being called in the partial. In this case, the `brand__color` needs to be applied to an element being generated inside CardHelper. Should CardHelper accept an instance of BrandHelper as an argument? Or I could put this element in the partial but then that would be adding logic to the partial.

app/assets/javascripts/govuk_publishing_components/components/image-card.js
patrickpatrickpatrick: If there's no Javascript included, it will still render the fallback link which will be rendered in the case where JS is disabled/doesn't run or cookies are disabled.

Screenshot 2023-01-06 at 14 58 10

app/views/govuk_publishing_components/components/_image_card.html.erb
patrickpatrickpatrick: Yep, this will overwrite 'gem-track-click' so I'll move it to another element.

lib/govuk_publishing_components/presenters/image_card_helper.rb
patrickpatrickpatrick: In this case I need to render the figcaption outside of the thumbnail image container for styling purposes. Unless I split thumbail into a seperate helper method and then just have `youtube_thumbnail_image + content_tag(:fig_caption...)`

app/views/govuk_publishing_components/components/_image_card.html.erb
patrickpatrickpatrick: actually https://github.com/alphagov/govuk_publishing_components/pull/3154 if this gets merge i guess i can use the same element

patrickpatrickpatrick: if gets merged in https://github.com/alphagov/govuk_publishing_components/pull/3154/files would use this implementation

lib/govuk_publishing_components/presenters/image_card_helper.rb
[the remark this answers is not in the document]
andysellick: Yeah I guess we don't want to call it both in the view and the helper. Would it be possible to move the view call into the helper?

app/assets/javascripts/govuk_publishing_components/components/image-card.js
andysellick: That's good to know. Will still need to check this gets deployed to applications correctly.

lib/govuk_publishing_components/presenters/image_card_helper.rb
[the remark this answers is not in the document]
patrickpatrickpatrick: Have added brand_helper as a parameter, so we can now use that helpers methods.

app/views/govuk_publishing_components/components/_image_card.html.erb
```diff
@@ -23,10 +23,13 @@
   ]
   extra_link_classes << brand_helper.color_class
 
+  data_modules = ""
+  data_modules << " gem-track-click" if card_helper.is_tracking?
+  data_modules << " image-card" if defined?(youtube_video_id)
```
andysellick, line 28: Non-blocking: I think you can use `%w[ ]` in order to not have to worry about leading whitespace, e.g.

```
data_modules = %w[]
data_modules << "gem-track-click" if card_helper.is_tracking?
data_modules << "image-card" if defined?(youtube_video_id)
```

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
andysellick: @patrickpatrickpatrick were you able to look into making this more accessible for keyboard users?

andysellick (commented):
Sorry, seem to have managed to get my review a bit muddled. Looking good, just a few minor questions, as well as the wider accessibility question.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
patrickpatrickpatrick: We can't style or change anything about the embed since it's an iframe. I suppose the only thing I could think of that would make the element more keyboard friendly would be making the iframe unfocusable, surrounding it in a focusable element that listens for keypresses and do the corresponding action for the keypresses on the youtube player. Even then we wouldn't be able to change any of the quite bad focus styles on the player. Alternatively, the embed could be made unfocusable by keyboard and we could display the link that appears when there's no embed.

patrickpatrickpatrick: https://accessibility.blog.gov.uk/2020/03/16/why-videos-on-gov-uk-use-the-youtube-video-player/ going from this blogpost YT has been quite extensively tested for accessibility and that's why we went for it over other players

andysellick: The Youtube video on this page - https://www.gov.uk/tour - seems to have slightly better focus states. They're obviously still part of the player (as they're blue outlines rather than our focus styles) but each control has some kind of visible focus state. I'm not sure what the difference between that player and this component is, but I think it's worth investigating.

app/views/govuk_publishing_components/components/_image_card.html.erb
patrickpatrickpatrick: Have rewritten this div to be a content_tag so we can just use `%w[]` for all the conditionally added attributes/classes.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
patrickpatrickpatrick: Apparently it's because the video on that page is larger. I guess at the size it's rendering currently Youtube assumes we are viewing it on mobile and so defaults to the mobile styling (which won't include the focus states). Going to see if there's a way to force the better desktop styling.

patrickpatrickpatrick: https://github.com/alphagov/govuk_publishing_components/pull/3156/commits/10c14d9825273e7c488b799cf827b38a1a0207fd have made an attempt at using this solution here, at desktop viewport the desktop youtube focus states will now be used

andysellick: Hmm sounds hacky. I have other ideas but they also seem hacky.

andysellick: (sorry that deleted comment was in response to an earlier comment)

andysellick: Latest deployment looks a little broken?

Screenshot 2023-01-10 at 11 02 18

patrickpatrickpatrick: Unfortunately I'm quite stumped. I'm able to get to use tablet styles but they aren't really that much better than the mobile styles. I think the only way to get it work would be to increase the width of the youtube video and decrease the size of the text container.

patrickpatrickpatrick: Have deployed something with the better focus/keyboard controls, let me know if you think that's OK or if we should think of alternative approach.

andysellick: That's better, well done! Shame the video in question has a blue bit behind the blue focus state for the play button, but that hopefully won't be a widespread problem. Did you have to do something complicated to achieve this?

patrickpatrickpatrick: It's a bit complex, yes.

```
.gem-c-image-card__youtube-video-embed iframe {
  z-index: 1;
  transform: scale(0.7);
  width: calc(100% + 42px);
}

.gem-c-image-card__youtube-video-embed.gem-c-govspeak__youtube-video iframe {
  top: -8px;
  left: -16px;
}

.gem-c-image-card__youtube-video-embed {
  transform: scale(1.3);
}
```

Basically here I scale down the iframe by 0.3, add the smallest magic number I can to the width of the iframe and then scale up the container of the iframe by 1.3.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
```diff
@@ -215,6 +215,17 @@
 
 .gem-c-image-card__youtube-video-embed iframe {
   z-index: 1;
+  transform: scale(0.7);
```
andysellick, line 218: This is a good solution, but I think it needs some tweaking. On both mobile and desktop it looks a little misaligned, although more noticeably on mobile.

On desktop it's not quite within the parent element (not sure if this is a problem, I haven't seen it alongside normal page content).

Screenshot 2023-01-11 at 14 22 20

On mobile it's too far to the right.

Screenshot 2023-01-11 at 14 22 37

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
```diff
@@ -215,6 +215,17 @@
...
+  width: calc(100% + 42px);
+}
+
+.gem-c-image-card__youtube-video-embed.gem-c-govspeak__youtube-video iframe {
+  top: -8px;
+  left: -16px;
```
andysellick, line 224: This is obviously a complex solution - would it be worth adding a comment to explain these numbers in particular?

patrickpatrickpatrick: have added comments to the all the rules pertaining the youtube embed

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
patrickpatrickpatrick: have fixed these two layout issues

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
andysellick: Well explained, good stuff.

app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
andysellick: Nice!

andysellick (approved):
Think this all looks good now, good job. One conflict to fix and might be worth squashing some of the commits together, but otherwise good 👍

PR #3166, opened 2022-12-30
[Help wanted] Pure CSS approach to numbering govspeak steps

kevindew:
Trello: https://trello.com/c/lxxx5XLZ/178-govuk-has-a-half-implemented-content-security-policy-csp

## Help wanted

I could really use some help from a frontend developer who is better informed on GOV.UK conventions. I made the changes needed to make this visually look as close as possible and wasn't sure if I was committing some sins with the approaches to sizings.

---

## What

This changes the approach for styling the custom list item markers for a Govspeak step list. Previously this was generated with a custom SVG image generated 30 times. Instead this uses a before pseudo-element with a content attribute of the number.

There were a few considerations made in putting this together:

1. I used the before pseudo-element rather than marker for IE support
2. I used the :before syntax rather than ::before as this seems to be the prevalent style in this codebase (presumably an IE legacy concern)
3. I couldn't seem to exactly match the weight of the font in the previous SVG, it seems it is somewhere between normal and bold
4. I had a fight with IE to avoid some extra padding in the list items, which led to the extra commented rules

I've tested it against the following browsers:

- Chrome
- Edge
- Firefox
- Safari
- Opera
- IE 11
- IE 10
- IE 9

## Why

The motivation for this change was the upcoming changes to the GOV.UK Content Security Policy that will disallow the use of data:image due to the increased security risk these can have [1].

This however has additional benefits. It reduces the size of CSS by approximately 13 KB and it resolves an ARIA warning in the component gem.

[1]: https://github.com/alphagov/govuk_app_config/blob/9b01e06d889be266501714e0dc0fc70a86e7e240/lib/govuk_app_config/govuk_content_security_policy.rb#L33-L37

## Visual Changes

There are some subtle differences compared to the previous version

### Before

Small screen:
Screenshot 2022-12-30 at 17 10 46

Large screen:
Screenshot 2022-12-30 at 17 10 16

### After

Small screen:
Screenshot 2022-12-30 at 17 12 30

Large screen:
Screenshot 2022-12-30 at 17 12 18


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
```diff
@@ -11,13 +11,32 @@
       list-style-type: decimal;
```
andysellick, line 11: I think you can safely remove the first three lines here as they no longer apply?

```
background-position: 0 .87em;
background-repeat: no-repeat;
list-style-type: decimal;
```

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
```diff
@@ -11,13 +11,32 @@
...
       margin-left: 0;
       padding: .75em 0 .75em 2.5em;
+      position: relative;
+
+      // IE doesn't render this correctly with this elemeent as display:block
+      // (it leaves a space for the :before pseudo element) by setting to
+      // display inline block and a width of 100% it works correctly.
+      display: inline-block;
+      width: 100%;
```
andysellick, line 20: This seems like a good solution. My only concern is that with padding to the left of the element, a width of 100% means it extends beyond its container, as shown below. I can't think of a situation where this might cause an actual problem, but it might be worth applying `box-sizing: border-box;` to be sure.

Screenshot 2023-01-03 at 09 42 36

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
```diff
@@ -11,13 +11,32 @@
       list-style-type: decimal;
       margin-left: 0;
       padding: .75em 0 .75em 2.5em;
+      position: relative;
+
+      // IE doesn't render this correctly with this elemeent as display:block
```
andysellick, line 16: Nitpick: `elemeent`.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
andysellick: Maybe keep the list style type, although I'm not sure what difference it makes.

andysellick (approved):
Looks good to me, great change, particularly pleased with the asset size reduction. I'm left to wonder what other legacy things like this we have kicking around that could do with a once over.

I've also checked this in Voiceover and it seems to be happy with the change, announcing the numbers.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
kevindew: Oops thanks

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
kevindew: Ah yes, I've removed background-position and background-repeat they're just things I missed.

I tried to do some history tracing of `list-style-type: decimal;` as it seemed a bit fishy (it relies on [overflow: hidden](https://github.com/alphagov/govuk_publishing_components/blob/5e682ed57f88cbedef61cafc0e62e4707170d38c/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss#L6) to hide them). I did some digging about the original approach but hit a dead end at https://github.com/alphagov/static/commit/1f41855b9e89690a3d5085e42bc0eef715a50801.

I wondered if it might be a way of still allowing numbers if someone disabled images, but that doesn't work (since they're still hidden by the overflow).

I've changed it to `list-style-type: none` since we've got no information recorded on why something quirky was done and it seems hard to imagine a case for it.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_steps.scss
kevindew: Ah yes, good shout on the border box. I've added that in.

Do you happen to have any experience with the problem I faced? It seems that in IE when I'm using the `:before` pseudo-element it continues to impact the layout of the li despite having `position: absolute`. I failed to find anything online of similar issues so don't know if it's something common or not. It was quite baffling.

kevindew:
Thanks so much for reviewing this @andysellick and for the accessibility checks, really appreciate it.

Is the use of the ems and font-size percentage ok? It felt a bit icky as the ems needed to be scaled down to match the font-size reduction but I assume we prefer em to rem still?

andysellick:
> Is the use of the ems and font-size percentage ok? It felt a bit icky as the ems needed to be scaled down to match the font-size reduction but I assume we prefer em to rem still?

I think it's fine. We definitely prefer em over rem as older browsers don't support that. There might be a better approach I haven't thought of but it looks and scales okay in various browsers so I think fixing it like this is okay.

kevindew:
Great thanks, are you happy for me to merge this?

PR #3167, opened 2023-01-03
Update popular links in the super navigation header

catalinailie:
## What
Update the list of popular links on the super navigation header component, so that it matches the list of popular links on the homepage. See https://github.com/alphagov/frontend/pull/3490
## Why
Updating this popular link is a request from content colleagues.


KludgeKML (approved):
LGTM

PR #3179, opened 2023-01-09
Add GA4 tracking to related navigation

andysellick:
## What
Adds Google Analytics 4 tracking to the sidebar and footer sections using the existing [ga4-link-tracker](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/ga4-link-tracker.md).

### Changes to the Related Navigation component

The [Related Navigation component](https://components.publishing.service.gov.uk/component-guide/related_navigation) shows a series of links. It is usually seen in the sidebar of a page but not exclusively. It includes some logic for which sections are displayed in different contexts, which is used when it is wrapped by other components.

- adds a `ga4_tracking` option that then adds relevant attributes to links as well as adding `ga4-link-tracker` into the `data-module` attribute
- changes how JS modules for tracking are enabled. UA tracking is enabled by default, so this change moves the `gem-track-click` `data-module` attribute from individual sections within the component to the top level, so that module is only initialised once (and combines it with the `ga4-link-tracker` `data-module` if GA4 tracking is enabled)
- changes how the title for each section is determined, as this is needed for the GA4 tracking. Content in each section has the title determined based on the content, except for the first section where the title defaulted to 'Related content'. Changes were needed since the attributes on all links when tracked needed to include the section title
- changes the aria-label configuration to fix an accessibility warning in the component guide (which I think was in error but this way seems better) by explicitly setting what was formerly the heading `id` to the `aria-label` instead, fixes the accessibility warning

### Changes to the Contextual Sidebar component

The [Contextual Sidebar component](https://components.publishing.service.gov.uk/component-guide/contextual_sidebar) is a wrapper for other components, including the Related Navigation component and the Step by Step Navigation component. It usually appears in the sidebar of a page.

- adds the `ga4_tracking` option to Contextual Sidebar, which is then passed to the Related Navigation component if set
- also passes any `ga4_tracking` option to the Step by Step Navigation component. This is slightly complicated by the fact that the step nav in this context only accepts one parameter, so we have to add the `ga4_tracking` flag to the `request.query_parameters` object, so it gets passed through various helpers and finally to the component itself correctly.

### Changes to the Contextual Footer component

The [Contextual Footer component](https://components.publishing.service.gov.uk/component-guide/contextual_footer) is a wrapper for the Related Navigation component. It appears in the footer of some pages. It contains logic that determines if a step by step navigation is already shown on the page - if this is false, the Related Navigation component is displayed.

- adds the `ga4_tracking` option, which is then passed to the Related Navigation component if set

## Why
We want to add tracking to links in the sidebar.

## Visual Changes
None.

Trello card: https://trello.com/c/IMjw2dHg/382-add-tracking-related-links-contextual-sidebar-and-contextual-footer

### DIFF
--- a/app/views/govuk_publishing_components/components/related_navigation/_section.html.erb
+++ b/app/views/govuk_publishing_components/components/related_navigation/_section.html.erb
@@ -3,20 +3,23 @@
      aria-labelledby="related-nav-<%= section_title %>-<%= random %>"
      data-module="gem-toggle">
 
-  <% unless section_title === "related_items" %>
+  <% if section_title === "related_items" %>
+   <% heading_text = related_nav_helper.component_title %>
+  <% else %>
     <%=
       heading_class = related_nav_helper.section_css_class("gem-c-related-navigation__sub-heading", section_title)
       heading_data = { 'track-count' => related_nav_helper.section_data_track_count(:related_item_section) }
       heading_id = "related-nav-#{section_title}-#{random}"
       heading_level = related_nav_helper.section_heading_level
+      heading_text = related_nav_helper.construct_section_heading(section_title)
 
       content_tag(heading_level, id: heading_id, class: heading_class, data: heading_data) do
-        related_nav_helper.construct_section_heading(section_title)
+        heading_text
       end
     %>
   <% end %>
 
-  <ul class="gem-c-related-navigation__link-list" data-module="gem-track-click">
+  <ul class="gem-c-related-navigation__link-list">
     <% constructed_link_array = [] %>
 
     <% section_link_limit = related_nav_helper.calculate_section_link_limit(links) %>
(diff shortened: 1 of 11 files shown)


app/views/govuk_publishing_components/components/related_navigation/_section.html.erb
```diff
@@ -3,27 +3,36 @@
...
       heading_data = { 'track-count' => related_nav_helper.section_data_track_count(:related_item_section) }
-      heading_id = "related-nav-#{section_title}-#{random}"
+      heading_label = "related-nav-#{section_title}-#{random}"
       heading_level = related_nav_helper.section_heading_level
+      heading_text = related_nav_helper.construct_section_heading(section_title)
 
-      content_tag(heading_level, id: heading_id, class: heading_class, data: heading_data) do
-        related_nav_helper.construct_section_heading(section_title)
+      content_tag(heading_level, class: heading_class, data: heading_data, aria: { label: heading_label }) do
+        heading_text
```
AshGDS, line 17: This should be `aria-labelledby` otherwise the assistive technology will read out the heading literally as `related-nav-related_items-dde33a2c`

AshGDS (changes requested):
@andysellick Looks good, I've requested a small change :+1:

app/views/govuk_publishing_components/components/related_navigation/_section.html.erb
andysellick: Actually it should be id, which is what it was in the first place. I think I went in a bit of a circle trying to fix an accessibility warning in the component guide, but I've concluded that the markup was fine as it was, I can't account for the warning. Have switched it back to id instead.

AshGDS: @andysellick Good spot, thanks

AshGDS (approved):
All good imo :+1:

PR #3180, opened 2023-01-10
Disable axe-core rules for issues highlighted in component guide

jon-kirwan:
## What

Disable axe-core rules for issues highlighted in component guide:

- `duplicate-id-aria`
- `duplicate-id`
- `landmark-unique`

## Why

The issues are highlighted because there are multiple instances of components, per page, in the component guide.

It's unlikely these issues would ever be seen because typically only one of these components would be deployed on any page e.g. **emergency banner**, **layout header**.


patrickpatrickpatrick (approved):
This all makes sense, I wasn't aware of there being an exception to rules before and makes sense to get rid of superfluous alerts.

PR #3190, opened 2023-01-16
Catch errors when modules initialised

andysellick:
## What
- wraps the initialisation of module JavaScript on a page with a try/catch block, so that if an error occurs in one module initialisation continues to the next without aborting and breaking all of the JS that follows

## Why
We've had some recent problems where an error in a single bit of JS caused all of the following JS to also fail.

## Visual Changes
None.

All credit goes to @AshGDS for thinking of this idea.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -38,8 +38,13 @@
           if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
             // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend Modules
             if (GOVUK.Modules[moduleName].prototype.init) {
-              new GOVUK.Modules[moduleName](element).init()
-              element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
+              try {
+                new GOVUK.Modules[moduleName](element).init()
+                element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/modules.js
```diff
@@ -38,8 +38,13 @@
           if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
             // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend Modules
             if (GOVUK.Modules[moduleName].prototype.init) {
-              new GOVUK.Modules[moduleName](element).init()
-              element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
+              try {
+                new GOVUK.Modules[moduleName](element).init()
+                element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
+              } catch (e) {
+                // if there's a problem with the module, don't start it
```
AshGDS, line 45: I'm wondering if this should this be more specific e.g.`if there's a problem with the module, catch its error so that other modules on the page will still start`

AshGDS (approved):
Looks good :+1: I added a non-blocking comment

app/assets/javascripts/govuk_publishing_components/modules.js
andysellick: Updated, thanks. I've also added a changelog entry as I think we should keep track of when this change was made.

PR #3191, opened 2023-01-17
Add translation strings for cookie banner confirmation message

jon-kirwan:
## What
Add translation strings for text toggled from JavaScript for the cookie banner confirmation message.

## Why
We should update the strings so that they can be translated.

## Anything else
- Updated missing translations [here](https://docs.google.com/spreadsheets/d/1Wips8w48mUoLE3UqY-TIeoHJcW9l371qUuRyCqbl5J0/edit#gid=0)


MartinJJones (approved):
Looks good to me 👍

PR #3195, opened 2023-01-18
Ensure our analytics load at the earliest point possible in page load

AshGDS:
## What

This fixes an issue with our eCommerce tracking no longer firing now that we have moved away from initialising our analytics in `static`. It fixes this issue by ensuring our analytics scripts exist as soon as `load-analytics.js` is found in the `<head>` of the page. Previously, our Google Analytics scripts would only exist after `DOMContentLoaded` fired, because the modules such as `ga4-event-tracker` required the `<body>` to exist first, so all of our analytics code was inside of `DOMContentLoaded`. This branch changes this, so that:

- At the point `load-analytics.js` loads, it will add the GA4 and UA Google scripts to the DOM `<head>` straight away (`gtm.js` and `google-analytics.com/analytics.js`) which will give us `window.dataLayer` and initialise UA.
-  Our GA4 modules such as `ga4-event-tracker` will wait until `DOMContentLoaded` has fired before initialising, as they have a dependency on the `<body>` of the page existing.
- Universal Analytics does not seem to need to wait for `DOMContentLoaded`, so all the GOVUK UA code will initialise in the `<head>` instead of inside of `DOMContentLoaded`.
- It achieves the above by moving the code `init-ga4.js` and `ga4.core.load()` into `load-analytics.js`

There's also some small refactoring to hopefully make `load-analytics.js` easier to understand now that it has more code in it. I also added some safeguards to prevent GA4 loading if certain things are missing (though that might be unnecessary and I'm being overly cautious.)

I've tested it on:
- `static`
- `collections` (browse page, coronavirus page, learn to drive page)
- `finder-frontend` (search page - eCommerce bug is fixed)
- `frontend` (GOVUK homepage)

## Why

- This ensures that the eCommerce tracking can run, as it relies on `window.dataLayer` and `window.GOVUK.analyticsGa4.vars.gem_version` in order to initialise itself, and push its pageview object.
- We weren't following best practice as we were adding the gtm script to the page after we got a `DOMContentLoaded` event, whereas it should exist in the `<head>` of the page as soon as possible.
- It mimics how our analytics was initialised initially in `static`.

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -179,11 +179,13 @@ window.GOVUK.loadAnalytics = {
   },
 
   startGa4Modules: function () {
-    var analyticsModules = window.GOVUK.analyticsGa4.analyticsModules
-    for (var property in analyticsModules) {
-      var module = analyticsModules[property]
-      if (typeof module.init === 'function') {
-        module.init()
+    if (!window.GOVUK.analyticsGa4.vars.disableModules) {
```
JamesCGDS, line 182: Would it be possible to put a comment here to explain what `disableModules` is used for?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -71,10 +73,9 @@ window.GOVUK.loadAnalytics = {
     }
   },
 
-  loadGa4: function (currentDomain) {
+  loadGa4Vars: function (currentDomain) {
     currentDomain = currentDomain || window.location.hostname
     var environment = ''
-
     // Categorise current environment
     if (this.arrayContains(currentDomain, this.productionDomains)) {
```
JamesCGDS: Nitpick: this code isn't part of the PR but I was wondering if there'd be any mileage in moving these parameters into an object so there's no reliance on passing them in the correct order? Happy to leave this as it is though as there are only 2 parameters, was just a thought that occurred to me.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -71,38 +73,119 @@ window.GOVUK.loadAnalytics = {
...
+    if (preview) {
+      preview = '&gtm_preview=' + preview + '&gtm_cookies_win=x'
+    }
+
+    this.googleSrc = 'https://www.googletagmanager.com/gtm.js?id=' + window.GOVUK.analyticsGa4.vars.id + auth + preview
+    newScript.src = this.googleSrc
+    firstScript.parentNode.insertBefore(newScript, firstScript)
+  },
+
   loadGa4: function (currentDomain) {
```
JamesCGDS, line 122: Again this code isn't part of the PR but I can't seem to find where we're calling `loadGa4` with parameters. Do we need to ultimately pass a parameter here? Although I think I've probably misunderstood something so apologies if this comment is redundant.

JamesCGDS:
Quick question: does [load-analytics.md](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/load-analytics.md) need updating to reflect these changes?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
AshGDS: @JamesCGDS That parameter is for our tests, it's used to override the domain we're on so that we can test that the right environment variables are loaded in the tests. Otherwise we couldn't test this code as jasmine always runs on `localhost` so it would always load the developer environment variables.

AshGDS:
> Quick question: does [load-analytics.md](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/load-analytics.md) need updating to reflect these changes?

@JamesCGDS Yep, good spot as @andysellick asked about this too :+1:

app/assets/javascripts/govuk_publishing_components/load-analytics.js
JamesCGDS: Cool understood, thanks @AshGDS. Do you think there'd be any mileage in putting a quick comment here saying as such? (Sorry I know this isn't part of the PR)

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -71,38 +73,119 @@ window.GOVUK.loadAnalytics = {
...
+    }
+
+    // If the current environment has initialiseGA4 = false, stop loading GA4
+    if (!this.ga4EnvironmentVariables[this.environment].initialiseGA4) {
+      return
+    }
+
+    // to be added: digital identity consent mechanism
+    var consentCookie = window.GOVUK.getConsentCookie()
+    if (consentCookie && consentCookie.usage) {
```
andysellick, line 156: Non blocking - I worry that this section is now slightly duplicating the equivalent section for UA i.e. checking for cookie consent before doing anything. Obviously we'll get rid of that other code at some point but I wonder if there's a way of combining them again to save checking for cookie consent twice.

andysellick (commented):
Mostly seems good, some small things, happy to discuss.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -71,38 +73,119 @@ window.GOVUK.loadAnalytics = {
     }
   },
 
+  loadGa4Vars: function () {
+    // If we recognise the environment, load the GA4 vars
+    if (this.ga4EnvironmentVariables[this.environment]) {
+      window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+      window.GOVUK.analyticsGa4.vars = window.GOVUK.analyticsGa4.vars || {}
+      window.GOVUK.analyticsGa4.vars.id = 'GTM-MG7HG5W'
+      window.GOVUK.analyticsGa4.vars.auth = this.ga4EnvironmentVariables[this.environment].auth
```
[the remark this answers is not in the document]
AshGDS: @andysellick Good idea - I may leave it for now as I think that will break the tests since they currently rely on being able to override those values. For example, if you wanted to test disabling GA4 on production works, in our tests we can currently do `window.GOVUK.loadAnalytics.ga4EnvironmentVariables['production'].initialiseGa4 = false` and `spyOn` is a `jasmine-jquery` thing so I have been trying to avoid using that.

app/assets/javascripts/govuk_publishing_components/load-analytics.js
AshGDS: @andysellick Good idea - the UA initialisation code looks a bit more complicated though so combining them may take a bit of time if we decided to do this.

AshGDS:
Thanks @andysellick / @JamesCGDS - I've made those changes now

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -6,37 +6,13 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
-    },
-
     sendData: function (data) {
       data.govuk_gem_version = this.getGemVersion()
       window.dataLayer.push(data)
     },
 
     getGemVersion: function () {
-      return window.GOVUK.analyticsGa4.vars.gem_version || 'not found'
+      return window.GOVUK.analyticsGa4.vars.gem_version
```
andysellick, line 15: This function seems redundant now that 'not found' is already defined elsewhere, maybe just update line 10 to `data.govuk_gem_version = window.GOVUK.analyticsGa4.vars.gem_version` and remove this function?

app/assets/javascripts/govuk_publishing_components/load-analytics.js
```diff
@@ -71,38 +73,119 @@ window.GOVUK.loadAnalytics = {
...
+      this.environment = 'development'
+    }
+
+    if (!this.environment) {
+      console.error('load-analytics.js: environment for GA4 not recognised - ' + this.environment + ' on ' + currentDomain)
+      return
+    }
+
+    // If the current environment has initialiseGA4 = false, stop loading GA4
+    if (!this.ga4EnvironmentVariables[this.environment].initialiseGA4) {
```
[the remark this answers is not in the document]
andysellick: This doesn't look like it's changed?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: @andysellick will do :+1:

JamesCGDS (approved):
Looks good, great work 👍

PR #3225, opened 2023-01-31
Add aria attributes to Details component

patrickpatrickpatrick:
## What

Add aria attributes to the Details component.

## Why

As part of the work to make the organisations page on GOV.UK more accessible. There is currently the following issue with the page:

> On the Departments, agencies and public bodies page, multiple buttons are present which began with the phrase "Works with" and contain the number of agencies or public bodies the organisation worked with. Many of these buttons have the same text occurring multiple times. For example, there are 23 occurrences of Works with 1 public body each of which relate to different department, agency or public body. Therefore, it's unclear in many instances which button related to which department, agency or public body which makes it difficult for screen reader users to differentiate between them.

@hannalaakso did some research and determined that the best way to fix the issue would be add an aria-label to the summary of the details.

[Relevant Trello Card](https://trello.com/c/NepoUk7x/1533-form-elements-on-departments-page-are-not-descriptive-of-their-function-or-purpose-9-m)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add component wrapper helper ([PR #3171](https://github.com/alphagov/govuk_publishing_components/pull/3171))
+* Add aria attributes to Details component ([PR #3225](https://github.com/alphagov/govuk_publishing_components/pull/3225))
 
 ## 34.7.1
 
--- a/app/views/govuk_publishing_components/components/_details.html.erb
+++ b/app/views/govuk_publishing_components/components/_details.html.erb
@@ -11,9 +11,11 @@
 
   data_attributes ||= {}
   data_attributes[:details_track_click] = ''
+
+  summary_aria_attributes ||= {}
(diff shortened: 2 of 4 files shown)


andysellick (commented):
Needs a test 😁

andysellick:
How urgent is this? Only I've got a PR ready to add this general functionality to all components if needed https://github.com/alphagov/govuk_publishing_components/pull/3171, which would making implementing this easier. However if it's urgent it's fine to go ahead, can rework it later.

patrickpatrickpatrick:
> How urgent is this? Only I've got a PR ready to add this general functionality to all components if needed #3171, which would making implementing this easier. However if it's urgent it's fine to go ahead, can rework it later.

It can probably wait for that to be merged first then, I don't think it's urgent.

patrickpatrickpatrick:
Have added a test to this now @andysellick

andysellick (approved):
Looks good. Please add a CHANGELOG entry before merging 👍

PR #3227, opened 2023-02-02
Allow a tracked ga4 link to be hardcoded via its ga4-link JSON

AshGDS:
## What

Allows a `url` to be hardcoded in a link's `ga4_link` JSON. This overrides the default functionality which automatically grabs the `href` of the link.

## Why

There has been a request to track a `<button>` click as a link click. Since button elements don't have a `href` attribute, the current JavaScript wouldn't capture a url if we added the link tracker to the button. Therefore if we can hardcode a `url` value to the button's `ga4_link` JSON, we can specify what we want the url to be.

## Visual Changes

None.


andysellick (approved):
Makes much more sense to me now! Thanks for adding the test.

PR #3238, opened 2023-02-06
Increase clickable area for links in navbar

patrickpatrickpatrick:
## What

Although some of the links on navbar use larger clickable areas on mobile viewports, this does not take into account users using touchscreens at larger viewports. This PR increases the touch areas for topic links in mobile viewport, topic links in desktop viewport and popular links in all viewports. It uses the already existing mixin in the layout-super-navigation-header.

## Why

For users who have trouble controlling their mouse, having the touch target limited to the link itself (for example the links in the Topics section of the menu) means it can be tricky to select with an unsteady hand. Could also make it difficult to click links if user a using a large touchscreen.

[Relevant Trello Card](https://trello.com/c/AMSVEJpN/1572-touch-targets-for-selecting-links-in-menu-could-be-bigger-s)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Increase clickable area for links in navbar ([PR #3238](https://github.com/alphagov/govuk_publishing_components/pull/3238))
+
+
 ## 34.10.1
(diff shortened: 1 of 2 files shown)


hannalaakso (approved):
This looks good to me 👍 Should we get a designer to review the interaction as well?

patrickpatrickpatrick:
also approved by a designer!

PR #3240, opened 2023-02-07
Add GA4 auto tracker

andysellick:
## What
Adds a GOVUK JavaScript module that when attached to a page with the right data attributes sends data to GA4 when the page loads. Initially for use on error pages in smart answers to send an event detailing the error.

## Why
Needed for the GA4 work with smart answers.

## Visual Changes
None.

Trello card: https://trello.com/c/xf2dDOBN/407-add-tracking-smart-answers-formerror

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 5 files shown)


JamesCGDS (approved):
Looks good 👍

PR #3241, opened 2023-02-08
Add documentation for testing a branch

andysellick:
## What / why
Adding documentation for testing a branch of the gem.

- was originally in dev docs, but multiple people expressed an inability to find it
- removing from dev docs (originally at https://docs.publishing.service.gov.uk/manual/test-a-branch-of-the-components-gem.html)
- adding it here to the gem will automatically include it in dev docs anyway
- copied with no changes

## Visual Changes
None.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,7 @@ yarn run jasmine:ci
 
 - [Install and use this gem](docs/install-and-use.md)
 - [Use a component in your application](docs/use-components.md)
+- [Test a branch of the gem](docs/test-a-branch-of-the-gem.md)
 
 ### Managing the gem
 
--- a/docs/test-a-branch-of-the-gem.md
+++ b/docs/test-a-branch-of-the-gem.md
@@ -1,0 +1,51 @@
+# Test a branch of the components gem
+
+This document explains how to test a github branch of `govuk_publishing_components` in an application, either locally or for testing a change in the gem on integration.
+
+## Prepare a branch of govuk_publishing_components
+
+Create a branch of `govuk_publishing_components` with the changes that you want to test.
+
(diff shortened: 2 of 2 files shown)


jon-kirwan (commented):
Should we include a section on deploying to Integration? (I seem to remember I found this quite useful when testing with Whitehall)

In which case we would need to add information on updating `Dockerfile` to replace (or add) `apt-get install -y build-essential nodejs` with `apt-get install -y build-essential nodejs git`

MartinJJones (dismissed):
Looks good to me, do we still need a changelog entry for this change?

andysellick:
> Should we include the following on deploying to Integration? (I seem to remember I found this quite useful when testing with Whitehall)
> 
> In which case we would need to add information on updating `Dockerfile` to replace (or add) `apt-get install -y build-essential nodejs` with `apt-get install -y build-essential nodejs git`

I've not got experience of this - deploying to integration has always been just like deploying a regular branch, but presumably whitehall is a bit different? If you're able to write this section @jon-kirwan I'd be happy to include it.

jon-kirwan:
> > Should we include the following on deploying to Integration? (I seem to remember I found this quite useful when testing with Whitehall)
> > In which case we would need to add information on updating `Dockerfile` to replace (or add) `apt-get install -y build-essential nodejs` with `apt-get install -y build-essential nodejs git`
> 
> I've not got experience of this - deploying to integration has always been just like deploying a regular branch, but presumably whitehall is a bit different? If you're able to write this section @jon-kirwan I'd be happy to include it.

Ah ok in which case if it's only for Whitehall maybe it's not necessary (since I think much of the Whitehall frontend has now been moved elsewhere to Collections or Government Frontend). Thanks, will re(approve)

jon-kirwan (approved):

PR #3243, opened 2023-02-08
Revert world locations base_path change

1pretz1:
## What

From raised Zendesk ticket: https://govuk.zendesk.com/agent/tickets/5207548

The [PR implementing the change](https://github.com/alphagov/govuk_publishing_components/pull/3102/commits/727cf81008226df642c832b0a11da14394538723) in how we fetch base_paths for world_locations caused the world location related links to redirect back to the same page. 

This occurred as the logic was changed to fetch the `base_path` from the links hash but the `base_path` isn't part of the included attributes for world locations and so the path (href) was blank [1] which causes the behaviour of redirecting back to the original page.

`base_paths` generated by parameterising the title is still flawed as it generates some broken links, however most links should now work after this revert. We should implement the original change when the base_path is included in the links. 

Example page of missing base_paths: /api/content/government/news/switzerland-hands-over-ukraine-recovery-conference-hosting-to-uk.

[1]: https://github.com/alphagov/govuk_publishing_components/blob/727cf81008226df642c832b0a11da14394538723/lib/govuk_publishing_components/presenters/related_navigation_helper.rb#L197.

### Before

Screenshot 2023-02-08 at 12 32 09

### After

Screenshot 2023-02-08 at 12 31 26

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Reverts release 34.7.1 concerning related World Location links ([PR #3243](https://github.com/alphagov/govuk_publishing_components/pull/3243))
+
 ## 34.8.0
 
 * Add GA4 Form tracker ([PR #3215](https://github.com/alphagov/govuk_publishing_components/pull/3215))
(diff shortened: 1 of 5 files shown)


hannako (approved):
LGTM

PR #3251, opened 2023-02-10
Cover all types of relative hrefs in hrefIsRelative function

AshGDS:
## What

- Previously we were checking if a URL was relative by the presence of a slash at the start of the href. We would also check if a URL was an anchor link by the presence of a `#` at the start of the href.
- This method was not accounting for all cases of relative hrefs. For example, a href starting with `?` is a valid relative href.
- Therefore, we now check if the href is relative or an anchor link by the lack of a domain name at the start of the href. In other words, if the href doesn't contain http://domain.example, https://domain.example, or //domain.example, then we assume the href is a relative href or anchor link.

## Why

This should cover all the types of relative hrefs and anchor links. Our tests pass against all these examples of relative URLs from RFC 1808 https://www.rfc-editor.org/rfc/rfc1808#section-5 apart from two on their list, but this is because: 
- `//g`, is a protocol relative URL, not a traditional relative URL. 
- Any href with colons at the start like `g:h` are detected as relative hrefs in this code, but browsers treat these as an external link going to a new domain. However, the RFC doesn't recommend people structure their links like this anyway: https://www.rfc-editor.org/rfc/rfc1808#section-5.3

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -40,6 +40,12 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
     },
 
     trackFunctions: {
+
+      getDomainRegex: function () {
+        // This regex matches a protocol and domain name at the start of a string such as https://www.gov.uk, http://gov.uk, //gov.uk
+        return /^(http:||https:)?(\/\/)([^\/]*)/ // eslint-disable-line no-useless-escape
+      },
+
       findTrackingAttributes: function (clicked, trackingTrigger) {
         if (clicked.hasAttribute('[' + trackingTrigger + ']')) {
           return clicked
@@ -82,12 +88,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
       },
 
       hrefIsRelative: function (href) {
-        // Checks that a link is relative, but is not a protocol relative url
-        return href[0] === '/' && href[1] !== '/'
-      },
+        // Checks that a href is relative by the lack of http:, https:// or // at the start of the href.
+        var domain = this.getDomainRegex().exec(href)
 
-      hrefIsAnchor: function (href) {
-        return href[0] === '#'
+        return !domain
       },
 
       isMailToLink: function (href) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -78,12 +78,11 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
-        // Checks that a link is relative, but is not a protocol relative url
-        return href[0] === '/' && href[1] !== '/'
-      },
+        // Checks that a href is relative by the lack of http:, https:// or // at the start of the href.
+        var domainRegex = /^(http:||https:)?(\/\/)([^\/]*)/ // eslint-disable-line no-useless-escape
+        var domain = domainRegex.exec(href)
 
-      hrefIsAnchor: function (href) {
-        return href[0] === '#'
+        return !domain
```
AshGDS, line 85: I had this return statement a bit more literal, but standardx keeps simplifying it.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -82,12 +82,11 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
       },
 
       hrefIsRelative: function (href) {
-        // Checks that a link is relative, but is not a protocol relative url
-        return href[0] === '/' && href[1] !== '/'
-      },
+        // Checks that a href is relative by the lack of http:, https:// or // at the start of the href.
+        var domainRegex = /^(http:||https:)?(\/\/)([^\/]*)/ // eslint-disable-line no-useless-escape
```
andysellick, line 86: We're already doing something similar in the `populateLinkDomain` function - is there any repetition that could be avoided here? (at the least this regex could be a single variable for both functions)

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
```diff
@@ -149,24 +149,47 @@ describe('GA4 core', function () {
...
+        '..',
+        '../',
+        '../g',
+        '../..',
+        '../../',
+        '../../g']
+
+      for (var i = 0; i < relativeURLs.length; i++) {
+        var href = relativeURLs[i]
+        expect(GOVUK.analyticsGa4.core.trackFunctions.hrefIsRelative(href)).toEqual(true)
```
andysellick, line 179: Can this also test that the function doesn't error when passed nothing?

AshGDS: Thanks @andysellick, I've added a check for an empty string, and nothing being passed :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
andysellick: Yep, that's good.

PR #3254, opened 2023-02-10
Expand use of component wrapper helper

andysellick:
**NOTE that this PR includes a breaking change, see below for details**

## What
Modify and improve the component wrapper helper to better suit the reality of how components are built. Specifically:

- changes the data attributes option name from `data` to `data_attributes` as most components are using the latter, which will avoid breaking changes
- adds a rendering test, to check the behaviour of the wrapper works in a real component environment
- adds a new attribute, role e.g. `role="navigation"` and associated helper methods
- fixes an issue where unspecified arguments were still being attached to a component e.g. `id="nil"`

Also use the component wrapper helper in the following components:

- accordion (also moves template code into a separate helper)
- cards
- checkboxes
- contents list (breaking change, see commit)

Miscellaneous changes:

- rename the breadcrumb helper, so that it is recognised by the component auditing

### Breaking changes

- contents list `aria_label` option must be updated to `aria` (now using the component wrapper) e.g. `aria_label: "something"` should change to `aria: { label: "something" }` (will need updates in `government-frontend` and `whitehall` - also used in `collections` but no instances use the `aria_label` option)

Related PRs to address this:

- https://github.com/alphagov/government-frontend/pull/2709
- https://github.com/alphagov/whitehall/pull/7412

## Why
The component wrapper helper is designed to improve consistency and reduce code across components, and to make adding new attributes easier. Also taking the opportunity to address some related issues.

## Visual Changes
None.


AshGDS (approved):
Looks good to me :+1:

PR #3257, opened 2023-02-13
Add GA4 tracking to our breadcrumbs

AshGDS:
Hi @JamesCGDS, would you be able to review this? Thanks :+1:

## What

- Adds the `ga4-link-tracker` to our Breadcrumbs component.

## Why

- It's one of the interactions on our to do list.
- I didn't create a boolean called `ga4_tracking: true` that toggles wether tracking is enabled. My thinking is that if we used a `ga4_tracking` boolean, we would have to add `ga4_tracking: true` to each frontend app, which might become a challenge to deploy and maintain, as some apps have multiple breadcrumb renders (e.g. `collections` has 6.) As well as this, the UA tracking doesn't have a toggler (it is always enabled by default on all breadcrumbs), so just adding the ga4 stuff directly into the component without a toggler replicates the existing functionality we have for tracking.

## Visual Changes

None.

### DIFF
--- a/lib/govuk_publishing_components/presenters/breadcrumbs.rb
+++ b/lib/govuk_publishing_components/presenters/breadcrumbs.rb
@@ -58,6 +58,12 @@ module GovukPublishingComponents
             dimension28: breadcrumbs_length.to_s,
             dimension29: crumb[:title],
           },
+          ga4_link: {
+            event_name: "navigation",
+            type: "breadcrumbs",
+            index: index.to_s,
+            index_total: breadcrumbs_length.to_s,
+          },
         }
 
         is_homepage = crumb[:url] == "/"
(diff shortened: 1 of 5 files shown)


lib/govuk_publishing_components/presenters/breadcrumbs.rb
```diff
@@ -58,6 +58,13 @@ def tracking_data(breadcrumbs_length)
             dimension28: breadcrumbs_length.to_s,
             dimension29: crumb[:title],
           },
+          ga4_link: {
+            event_name: "navigation",
+            type: "breadcrumbs",
+            text: crumb[:title],
```
JamesCGDS, line 64: Does the text property need to be set here or would it be handled automatically by the link tracker?

AshGDS: Good spot @JamesCGDS - I've removed that property now :+1:

JamesCGDS (approved):

PR #3259, opened 2023-02-15
Force English text on our GA4 related navigation section tracking

AshGDS:
## What

- Ensures that the `section` value on our Related Navigation tracking is always English so that we can understand the text in the analytics dashboard.

## Why

- The headings on our Related Navigation `section` elements get translated, however for the purpose of analysis we always want it to be English, so that everything is tracked under the same label even if they are translated. This keeps the analytics under one label instead of different labels for each language.

## Visual Changes

None.

### DIFF
--- a/lib/govuk_publishing_components/presenters/related_navigation_helper.rb
+++ b/lib/govuk_publishing_components/presenters/related_navigation_helper.rb
@@ -53,20 +53,25 @@ module GovukPublishingComponents
         end
       end
 
-      def construct_section_heading(section_title)
-        unless section_title == "related_items"
+      def construct_section_text(section, underscores_to_spaces)
+        unless section == "related_items"
+          defaults = [I18n.t("components.related_navigation.#{section}")]
+          defaults << section.tr("_", " ") if underscores_to_spaces
+
           I18n.t(
-            "components.related_#{@context}_navigation." + section_title,
-            default: [
-              I18n.t("components.related_navigation.#{section_title}"),
-              section_title.tr("_", " "),
-            ],
+            "components.related_#{@context}_navigation.#{section}",
+            default: defaults,
           )
         end
       end
 
-      def component_title
-        I18n.t("components.related_#{@context}_navigation.related_content", default: I18n.t("components.related_navigation.related_content"))
+      def construct_ga4_section_text(section)
+        # Force English so we can still understand what is being tracked if translated.
+        underscores_to_spaces = true
+        underscores_to_spaces = false if section == "related_content"
+        I18n.with_locale("en") do
+          construct_section_text(section, underscores_to_spaces)
+        end
       end
 
       def section_css_class(css_class, section_title, link: {}, link_is_inline: false)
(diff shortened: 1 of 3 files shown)


lib/govuk_publishing_components/presenters/related_navigation_helper.rb
```diff
@@ -65,10 +65,24 @@ def construct_section_heading(section_title)
...
+      def construct_ga4_section_heading(section_title)
+        # Force English so we can still understand what is being tracked if translated.
+        I18n.with_locale("en") do
+          construct_section_heading(section_title)
+        end
+      end
+
       def component_title
         I18n.t("components.related_#{@context}_navigation.related_content", default: I18n.t("components.related_navigation.related_content"))
       end
```
gclssvglx, line 77: I feel this method can be merged with `construct_section_heading` above. This should then mean you can merge `construct_ga4_section_heading` and `ga4_component_title`. This would be nice as both could then be re-used if ever we need/want to add another attribute to the event.

gclssvglx (commented):
Looking good @AshGDS - just have one refactor suggestion.

lib/govuk_publishing_components/presenters/related_navigation_helper.rb
AshGDS: Thanks @gclssvglx , I don't understand fully, would you be able to elaborate? I didn't interfere with the original methods as those are needed to display the translated version of the text to the user on the page, which is why forcing them to be English is in the separate methods.

gclssvglx: Hi @AshGDS sure, maybe something like this...

```ruby
def construct_section_text(section, underscores_to_spaces)
  unless section == "related_items"
    defaults = [I18n.t("components.related_navigation.#{section}")]
    defaults << section.tr("_", " ") if underscores_to_spaces

    I18n.t(
      "components.related_#{@context}_navigation.#{section}",
      default: defaults,
    )
  end
end

def construct_ga4_section_text(section)
  # Force English so we can still understand what is being tracked if translated.
  underscores_to_spaces = true
  underscores_to_spaces = false if section == "related_content"
  I18n.with_locale("en") do
    construct_section_text(section, underscores_to_spaces)
  end
end
```

Of course, you'll need to make sure any existing instances of the `component_title` call are converted.

AshGDS: @gclssvglx Ah OK I think I understand now - so in this case `heading_text = related_nav_helper.component_title` would become `heading_text = related_nav_helper.construct_section_text("related_content")` if I understand correctly? Nice idea!

gclssvglx: Yup, that's the idea 👍  Just removes some duplication and simplifies the code a bit

AshGDS: Thanks @gclssvglx - I've implemented that now, would you be able to re-review when you have a chance? :+1:

gclssvglx (approved):
Nice work @AshGDS 👍 :shipit:

PR #3261, opened 2023-02-21
Fix smart answer results tracking

JamesCGDS:
## What
This PR contains two small changes to the smart answers results tracking:

- Moves a call to `textContent()` a little later in the code as this was causing errors when calling it on a node that didn't exist
- Checks that elements with a `data-ga4-ecommerce-path` data attribute exist before tracking results and attaching event listeners

## Why
Prevents this code from erroring and running where it isn't needed

## Visual Changes
N/A


andysellick (approved):
Nice fix. Suggest you merge the two commits as they relate directly to each other.

PR #3262, opened 2023-02-21
Add indexing of links using JS

JamesCGDS:
## What
This PR adds indexing of links using JS for when the `index` and `index_total` properties cannot be retrieved from the template. It is enabled by adding a `data-ga4-set-indexes` attribute on the parent element.

## Why
Part of the GA4 migration.

## Visual Changes
N/A

Trello card: https://trello.com/c/4hvUQKAf/415-add-tracking-smart-answers-information-click


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -218,6 +218,28 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
         }
+      },
+
+      // The index and index_total properties are usually set by the application's templating engine however in instances where this isn't possible,
+      // we use JavaScript to calculate the index_total and append it to data-ga4-link and add an index property to each link
+      setIndexes: function (module) {
+        var linksToBeIndexed = []
+
+        // Only index links that are not search results
+        // querySelectorAll(a:not['data-ga4-ecommerce-path']) would have been used if not for lack of support in older browsers
```
andysellick, line 229: Not sure this comment is needed? I'd put it in the PR description if you want to document it.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -218,6 +218,28 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
+          }
+        }
+
+        var ga4LinkData = JSON.parse(module.getAttribute('data-ga4-link'))
+        ga4LinkData.index_total = linksToBeIndexed.length
+        module.setAttribute('data-ga4-link', JSON.stringify(ga4LinkData))
+
+        for (var j = 0; j < linksToBeIndexed.length; j++) {
+          linksToBeIndexed[j].setAttribute('data-ga4-index', j + 1)
+        }
```
andysellick, line 242: I think this second loop might not be necessary. I've rewritten this function a little, how does this look?

```Javascript
setIndexes: function (module) {
  var links = module.querySelectorAll('a')
  var total_links = 0
  for (var i = 0; i < links.length; i++) {
    var link = links[i]
    // Only index links that are not search results
    if (!link.getAttribute('data-ga4-ecommerce-path')) {
      total_links++
      link.setAttribute('data-ga4-index', i + 1)
    }
  }

  var ga4LinkData = JSON.parse(module.getAttribute('data-ga4-link'))
  ga4LinkData.index_total = total_links
  module.setAttribute('data-ga4-link', JSON.stringify(ga4LinkData))
}
```

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
```diff
@@ -275,6 +275,48 @@ describe('GA4 core', function () {
...
+
+    describe('when the data-ga4-set-indexes property exists on the module', function () {
+      var module
+
+      beforeEach(function () {
+        module = document.createElement('div')
+        module.setAttribute('data-ga4-link', '{"someData": "blah"}')
+        module.innerHTML = '<a id="example1" href="www.example1.com">Example link 1</a>' +
+
+        '<a id="example2" href="www.example2.com">Example link 2</a>' +
```
andysellick, line 287: Nitpick: don't think we need empty lines between each bit of markup here.

andysellick (commented):
Interesting change. I've made a rewrite suggestion, but otherwise looks good. I'd also squash the commits together since they relate to the same thing.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
JamesCGDS: Thanks Andy, I've just tested this and I think I might need both loops - the problem is that if I only have the one loop, the `index` property will be set incorrectly. So in the code above, it sets the `data-ga4-index` attribute correctly but the index would’ve kept incrementing as it’s taking all links into account. For example, if there are 10 links but the first 7 are search results, clicking on the first non-search result link would look like:

```
index: 8
index_total: 3 
```

andysellick: Ah of course. Would it fix it to amend `link.setAttribute('data-ga4-index', i + 1)` to `link.setAttribute('data-ga4-index', total_links + 1)`?

JamesCGDS: Very good point, thanks Andy - although I think we might be able to just set it to total_links

andysellick: Ah of course it's already being incremented on the previous line. Excellent.

andysellick (approved):
Nice.

PR #3263, opened 2023-02-21
Add helpers by component to auditing

andysellick:
## What
Expands the component auditing to audit which components use some of our shared helper code, specifically the brand helper, component wrapper helper, heading helper and shared helper. 

You can find this by running the component guide locally and following the 'audit' link from the main page.

## Why
The component wrapper helper is probably going to take on some of the work of the shared helper, so we'll eventually be migrating code from one to the other, so it's helpful to know which components use it.

Also generally useful to know which components rely on different bits of code.

## Visual Changes
New section added to the auditing page:

Screenshot 2023-02-21 at 11 46 21

### DIFF
--- a/app/models/govuk_publishing_components/audit_components.rb
+++ b/app/models/govuk_publishing_components/audit_components.rb
@@ -34,12 +34,19 @@ module GovukPublishingComponents
   private
 
     def compile_data(path)
+      component_file_details = get_component_file_details(path)
+      helper_usage = get_helper_usage(component_file_details)
+
+      # remove the template file reference as not needed
+      component_file_details.map { |detail| detail.delete(:template_file) }
+
       {
         gem_found: true,
         component_code: clean_files(@all_templates, @templates_full_path).sort,
         components_containing_components: find_all_partials_in(@all_templates),
-        component_file_details: get_component_file_details(path),
+        component_file_details: component_file_details,
         component_numbers: @component_numbers,
+        helper_usage: helper_usage,
       }
     end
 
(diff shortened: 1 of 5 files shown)


app/models/govuk_publishing_components/audit_components.rb
```diff
@@ -215,5 +225,50 @@ def get_asset_link(a_thing, component)
...
+
+        helpers.each do |helper|
+          next if File.foreach(file).grep(helper[:match]).empty?
+
+          helper[:used_by] << {
+            name: component[:name],
+            link: component[:template_file].split("/components/")[1],
+          }
+        end
+      end
```
gclssvglx, line 269: If the files are small, something like this maybe more efficient...

```ruby
components.each do |component|
  next unless File.file?(component[:template_file])

  # This reads the entire file into memory, so not good for large files. Also, you need to remember to `.close` it to avoid leaks
  file = File.open(component[:template_file])
  helpers.each do |helper|
    # I've reversed the logic a little, think it's easier to read given the `next` above
    next unless file.grep(helper[:match]).present?  

    helper[:used_by] << {
      name: component[:name],
      link: component[:template_file].split("/components/")[1],
    }
  end
  file.close
end
```
Otherwise, I think what you've done is fine.

But, the performance will also depend on which is the shorter list `components` or `helpers`. As a general rule, whichever is predictably the shorter put that on the outside loop.

andysellick: Thanks. I deliberately avoided `File.open` on the basis that some are large files, because I read that `File.foreach(file).grep` reads the file line by line, so doesn't load it all into memory at once. Most of the time the lines we're looking for will be in the first 10 lines or so of the template, so that seemed like an efficient way to go.

PR #3270, opened 2023-02-22
Enable strip PII override in GA4 form tracker

andysellick:
## What
- the form tracker (when configured to collect values from text inputs) strips PII from that value
- this only included the default things (see PII doc for list) and not dates or postcodes
- changing this to use PII strip override, which redacts everything possible

## Why
Part of the GA4 work, plus we don't want to collect any PII.

## Visual Changes
None.

Trello card: https://trello.com/c/LNE9OTXb/427-add-tracking-search-box-on-the-header-menu-bar


JamesCGDS (approved):
Looks good! 👍

PR #3277, opened 2023-02-24
Change how GA4 index parameter works

andysellick:
## What
Changes how the index parameter is handled, from being a string of the form `1.2.3` or `2.1` to an object with sub-parameters. 

This is handled by a function in `ga4-core.js`. Trackers that could receive an index parameter are also modified in this PR to call the function. This means that we can automatically update our index data without having to change all the places in applications where index is declared. The function is written so that if we decided to switch an application to pass the correct `index` parameter, it shouldn't cause a problem - so if we wanted to later update all our applications to pass the correct data in the first place, we could do that.

The ecommerce tracker has not been modified as the `index` parameter works differently there.

## Why
The data structure is being changed to help analysts understand the data more easily. This will involve future work to fix existing instances of the index not being quite right, and to add a third sub-parameter to this object.

This is being handled centrally in the gem code to avoid having to change all the existing instances of how index is handled across applications.

## Visual Changes
None.

Trello card: https://trello.com/c/8tNN5gGX/426-change-the-index-parameter

### DIFF
--- a/lib/govuk_publishing_components/presenters/breadcrumbs.rb
+++ b/lib/govuk_publishing_components/presenters/breadcrumbs.rb
@@ -61,7 +61,9 @@ module GovukPublishingComponents
           ga4_link: {
             event_name: "navigation",
             type: "breadcrumbs",
-            index: index.to_s,
+            index: {
+              index_link: index.to_s,
+            },
             index_total: breadcrumbs_length.to_s,
           },
         }
--- a/spec/components/breadcrumbs_spec.rb
+++ b/spec/components/breadcrumbs_spec.rb
@@ -73,7 +73,9 @@ describe "Breadcrumbs", type: :view do
     expected_tracking_options = {
       event_name: "navigation",
       type: "breadcrumbs",
-      index: "1",
+      index: {
+        index_link: "1",
+      },
       index_total: "1",
     }
 


andysellick:
@JamesCGDS @AshGDS thanks for your feedback and help on this, have now amended the `ga4-core.js` code to fix the identified problem 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -237,6 +237,37 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
+          return index
+        }
+
+        index = index.toString().split('.')
+        // this will soon include 'index_section_count'
+        // number of sections overall in the thing being tracked (but not yet in use)
+        switch (index.length) {
+          case 1:
+            return {
+              index_section: parseInt(index[0])
```
JamesCGDS, line 258: Quick question: I remember us saying that the `index` attribute on accordions would only ever contain one digit and that it would need to be mapped to `index_section`, which is what the above code would do. I'm just wondering about what would happen with links that only have one digit, like the breadcrumbs; would mapping the breadcrumb indexes to `index_section` be okay? Or would it need to be `index_link`?

andysellick: That's concerning - if the indexes are that inconsistent this may prevent us from doing this entire approach - we'll have to manually update all the existing indexes in every component and component call across all our applications. Which I was hoping to avoid. Will raise.

AshGDS: The Breadcrumbs GA4 stuff is only defined in one place, so maybe it could have the correct object added straight into it's `ga4-link` attribute? Or is that complicating things because then we'll have `1.2.X` in some components and `index: {index_link: 1}` in others

andysellick: I've written the code to be future proof - if the parameter passed to the function is already an object, it'll simply return the object, so for those things that are an edge case we could update them manually. Good shout.

JamesCGDS: Cool, sounds like we've got a solution (manually updating the edge cases) so I'll mark this as resolved 👍

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
```diff
@@ -313,6 +313,45 @@ describe('GA4 core', function () {
...
       })
     })
+
+    describe('when handling the index parameter', function () {
+      it('converts a three digit index into sub-parameters as expected', function () {
+        var expected = {
+          index_section: 2,
+          index_link: 3
+        }
+        // console.log(GOVUK.analyticsGa4.core.trackFunctions.createIndexObject('1.2.3'))
```
JamesCGDS, line 323: Rogue comment

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -96,12 +96,16 @@ describe('GA4 click tracker', function () {
     beforeEach(function () {
       attributes = {
         event_name: 'navigation',
-        type: 'a link'
+        type: 'a link',
+        index: 1
```
JamesCGDS, line 100: Should this test reflect the single digit edge case for links? I.e. should we change this to `index: {index_link: 1}`?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
andysellick: Darn it. Thanks for spotting, fixed.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
andysellick: Have updated now, thanks.

JamesCGDS (approved):

AshGDS (commented):
Looks good, I have a couple questions:

- Will you be fixing the breadcrumbs index in this PR, or will this be done in a separate PR?
- Will the index functionality be documented anywhere?

andysellick:
> * Will you be fixing the breadcrumbs index in this PR, or will this be done in a separate PR?
> * Will the index functionality be documented anywhere?

Of course, I'd forgotten that the breadcrumbs stuff is here! Yes, will definitely do that.

And yes, documentation is also a good idea. Will check.

AshGDS:
Thanks @andysellick - for reference, here's the breadcrumb code: https://github.com/alphagov/govuk_publishing_components/blob/ca68f50ea4641f2f09f23d3e83b51ec19d970efc/lib/govuk_publishing_components/presenters/breadcrumbs.rb#L61

andysellick:
@AshGDS have updated the breadcrumb tracking as suggested, thanks for that. I had a look at the documentation but aside from updating what's already there I'm not sure I want to add anything specific for the new index parameters - that feels like something more suitable for the implementation guide/record. Also I'm hoping that eventually we can phase out the need for this code, by updating GA4 tracking data in applications when there's time and opportunity to do so. Does all that sound okay or do you have a specific thing you think we should be documenting here?

AshGDS (approved):
All good, happy for the documentation to just be in the implementation record @andysellick

PR #3289, opened 2023-03-07
Add GA4 tracking to step by step links

AshGDS:
## What

Adds link tracking to the step by step navigation links. Uses the `ga4-link-tracker`.

You can test it by running `collections` with a local version of publishing components and going to http://127.0.0.1:3070/learn-to-drive-a-car. You will then see the `ga4-link` attribute on the links within each step.

I've used the new `index` way of tracking, so let me know if I've implemented it incorrectly.

## Why

It's on our list of things to implement.

Trello card: https://trello.com/c/TI06lTIG/431-add-tracking-step-by-step-links

## Visual Changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
@@ -22,7 +22,7 @@
 %>
 <% if steps %>
   <div
-    data-module="gemstepnav<% if ga4_tracking %> ga4-event-tracker<% end %>"
+    data-module="gemstepnav<% if ga4_tracking %> ga4-event-tracker ga4-link-tracker<% end %>"
     <%= "data-ga4-expandable" if ga4_tracking %>
     class="gem-c-step-nav js-hidden <% if small %>govuk-!-display-none-print<% end %> <% unless small %>gem-c-step-nav--large<% end %>"
     <%= "data-remember" if remember_last_step %>
@@ -77,8 +77,10 @@
             <%
               in_substep = false
               options[:step_nav_content_id] = step_nav_content_id
+              options[:step_title] = step[:title]
               options[:step_index] = step_index
               options[:link_index] = 0
+              options[:ga4_tracking] = ga4_tracking
             %>
             <% step[:contents].each do |element| %>
               <%= step_nav_helper.render_step_nav_element(element, options) %>
--- a/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
+++ b/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
@@ -46,6 +46,7 @@ module GovukPublishingComponents
                 :li,
                 class: "gem-c-step-nav__list-item js-list-item #{link_active(contents[:active])}",
               ) do
+                @options[:ga4_index_total] = element[:contents].length if @options[:ga4_tracking]
                 create_list_item_content(contents)
               end,
             )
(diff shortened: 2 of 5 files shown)


spec/components/step_by_step_nav_spec.rb
```diff
@@ -257,7 +257,25 @@ def stepnav
...
+    }
+
     assert_select ".gem-c-step-nav[data-ga4-expandable]"
     assert_select ".gem-c-step-nav[data-module='gemstepnav ga4-event-tracker']"
+    assert_select "#{step1} .gem-c-step-nav__link[href='/link1'][data-ga4-link='#{expected.to_json}']"
+
+    expected = {
+      "event_name": "navigation",
+      "type": "step by step",
+      "index": { "index_section": "3", "index_link": "1" },
```
AshGDS, line 275: This is `index_section: 3` even though it's Step 2 because I think Step 1 has two sections, so Step 2 is section 3.

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -77,8 +77,10 @@
             <%
               in_substep = false
               options[:step_nav_content_id] = step_nav_content_id
+              options[:step_title] = step[:title]
```
AshGDS, line 80: This `options` object gets passed to a function which builds all the `<li>` elements. Therefore I needed to add the `step_title` and `ga4_tracking` boolean to this object, so that I can build the `ga4-link` object on each list element that has a link.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
```diff
@@ -46,6 +46,7 @@ def list(element)
                 :li,
                 class: "gem-c-step-nav__list-item js-list-item #{link_active(contents[:active])}",
               ) do
+                @options[:ga4_index_total] = element[:contents].length
```
andysellick, line 49: Does this need a `if @options[:ga4_tracking]` at the end?

spec/components/step_by_step_nav_spec.rb
andysellick: > I believe this is `index_section: 3` even though it's Step 2 because in the test data Step 1 has two sections, so Step 2 is section 3.

Thanks for the explanation. It's slightly laborious, but could you make this clearer in the tests? e.g. have one test that checks the behaviour with simple data (two steps) and one with this slightly more complex scenario of one step containing two?

app/views/govuk_publishing_components/components/_step_by_step_nav.html.erb
```diff
@@ -77,8 +77,10 @@
...
               options[:step_index] = step_index
               options[:link_index] = 0
+              options[:ga4_tracking] = ga4_tracking
```
andysellick, line 83: Not sure where to put this comment. Can you expand the YML file example of with GA4 tracking to include links? That would show the extra attributes that are added to links when this option is enabled.

AshGDS: Thanks @andysellick, I've done this now.

lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb
AshGDS: @andysellick Good spot, done.

AshGDS:
Thanks @andysellick I've made all the suggested changes and spotted that I forgot to add the `ga4-link-tracker` data module.

PR #3290, opened 2023-03-08
Update accordion index parameter

andysellick:
## What
- in line with changes to the index parameter, explicitly define index as index_section, remove index_total and use that value for index_section_count

## Why
Part of the GA4 migration work.

## Visual Changes
None.

Trello card: https://trello.com/c/8tNN5gGX/426-change-the-index-parameter

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix smart answer PII issue ([PR #3291](https://github.com/alphagov/govuk_publishing_components/pull/3291))
+* Update accordion index parameter ([PR #3290](https://github.com/alphagov/govuk_publishing_components/pull/3290))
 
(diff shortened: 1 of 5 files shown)


JamesCGDS:
Does the `Show/Hide all sections` button need updating too? I'm just doing some testing locally and it looks like it's using the previous `index` approach.

If so, I think [this line](https://github.com/alphagov/govuk_publishing_components/blob/74e8a5e94cb7138a5dc021c4b8fde9d73e52c7b7/app/assets/javascripts/govuk_publishing_components/components/accordion.js#L65) might need updating to something like:

`var showAllAttributesGa4 = { event_name: 'select_content', type: 'accordion', index: {index_section: 0, index_section_count: indexTotal}}`

JamesCGDS (approved):

PR #3291, opened 2023-03-08
Fix smart answer PII issue

JamesCGDS:
Dates were being recorded on change response events - this PR redacts them

## What
When changing a form response on smart answers, dates that are in string format are being recorded and this is a PII issue. Therefore, this PR redacts the dates and replaces them with `[date]` (e.g. `1 January 1990` is replaced with `[date]`).

## Why
Addresses a PII issue.

## Visual Changes
N/A

Trello card: https://trello.com/c/dWFDOLAV/433-smart-answer-bugs-formchangeresponse

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


andysellick (approved):
Approving for now as need to implement a solution quickly, but have agreed that there are a few alternative approaches to add to our backlog for future investigation that could provide a more flexible approach.

PR #3292, opened 2023-03-08
Fix component auditing

andysellick:
## What / why
Resolves a few issues with the component auditing. Specifically:

- fixes a silent problem introduced when the data structure changed in a recent update
- removes print stylesheet auditing, as we no longer have separate print styles for components
- adds a count to the component references found in applications

## Visual Changes
Mentions of print stylesheets have been removed from the auditing pages.

Counts have been added to the lists of components in applications:

Screenshot 2023-03-09 at 10 46 37

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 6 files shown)


MartinJJones (approved):
Looks good to me 👍

PR #3295, opened 2023-03-10
Add ga4 link tracking to footer links

AshGDS:
## What

Adds GA4 Link tracking to the Footer Links.

## Why

It is on our list of interactions to implement. Trello card: https://trello.com/c/3s9DA2dY/434-add-tracking-footer-links

The code for this might be a bit more complicated than usual but I have tried my best to explain it. Essentially:

- There are 5 sections in the Footer. These are:
    - Topics section
    - Government Activity section
    - The "Meta" section (contains link such as: Help, Privacy, Cookies etc.)
    - The OGL Link section
    - The Crown Copyright link section.
- Of these 5 sections:
    - Topics and Government Activity have their data attributes generated in the same place in Ruby
    -  The "Meta" section has its data attributes generated in a similar place in Ruby as the above (with a slight difference)
    - The OGL Link is hard coded in the HTML
    - The Crown Copyright link is hard coded in the HTML

Therefore the GA4 stuff is either defined in `public_layout_helper.rb` for the Ruby generated sections, or `_layout_footer.html.erb` for the hard coded sections. The `index-total` value is shared across both files. For the OGL/Crown Copyright link all the values are hardcoded apart from `index-total`. For the other three sections, it's a mix of hardcoded values and code. 

I will comment the lines of code to try and make it make more sense as well.

## Visual Changes

None.


app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -102,6 +115,17 @@
...
         data-track-links-only
+        data-ga4-link="<%= {
+            "event_name": "navigation",
+            "section": "Copyright",
+            "index": {
+              "index_section": "5",
+              "index_link": "1",
+              "index_section_count": "5",
+            },
+            "text": "© Crown copyright",
```
AshGDS, line 126: The text value is hardcoded because this link gets translated (line 130.)

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -158,6 +158,7 @@
         with_border: !omit_footer_border,
         navigation: omit_footer_navigation ? nil : layout_helper.footer_navigation,
         meta: layout_helper.footer_meta,
+        ga4_index_total: layout_helper.get_footer_length
```
AshGDS, line 161: This grabs the `@footer_length` value from `public_layout_helper.rb` so that the OGL and Crown Copyright link can access it.

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -158,6 +158,7 @@
         with_border: !omit_footer_border,
         navigation: omit_footer_navigation ? nil : layout_helper.footer_navigation,
         meta: layout_helper.footer_meta,
+        ga4_index_total: layout_helper.get_footer_links_count
```
AshGDS, line 161: This passes the amount of footer links down to where the OGL and Crown Copyright links are hardcoded, so that they can access the value.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
       attr_reader :footer_navigation, :footer_meta, :cookie_banner_data
 
       def initialize(local_assigns)
+        @footer_links_count = generate_ga4_footer_links_count(local_assigns)
```
AshGDS, line 49: We must calculate the amount of footer links as the first thing in this function. This is because the two lines beneath this are where the links are actually generated and have their data attributes added to them. Therefore the amount of links must be calculated here otherwise the data attributes for everything will already be set if we do it at a later time in this function.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
         @footer_navigation = local_assigns[:footer_navigation] || navigation_link_generation_from_locale(I18n.t("components.layout_footer.navigation_links"))
-        @footer_meta = local_assigns[:footer_meta] || { items: add_data_attributes_to_links(FOOTER_META[:items], "supportLink") }
+        @footer_meta = local_assigns[:footer_meta] || { items: add_data_attributes_to_links(FOOTER_META[:items], "supportLink", { section: "Support links", index_section: 3 }) }
```
AshGDS, line 51: This is where the `Help, Privacy, Cookies` links are generated.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
+
+        footer_links_count += extra_footer_links
+        footer_links_count
+      end
+
+      def get_footer_links_count
+        @footer_links_count
+      end
+
       def navigation_link_generation_from_locale(links)
```
AshGDS, line 81: This is where the `Topics` and `Government Activity` links get generated.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
         links.each_with_index.map do |menu, i|
+          ga4_data = {
+            section: menu[:title],
+            index_section: i + 1,
+          }
           {
             title: menu[:title],
             columns: footer_navigation_columns[i],
-            items: add_data_attributes_to_links(menu[:menu_contents], footer_track_actions[i]),
+            items: add_data_attributes_to_links(menu[:menu_contents], footer_track_actions[i], ga4_data),
```
AshGDS, line 90: We are passing the `section` and `index_section` which we calculated in this function downwards towards the `add_data_attributes_to_links` function.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
         FOOTER_TRACK_ACTIONS
       end
 
-      def generate_data_attribute(link, track_action)
+      def generate_data_attribute(link, track_action, ga4_attributes)
```
AshGDS, line 103: This is where all the data attributes for a link are consolidated.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
           track_label: link[:href],
           track_options: {
             dimension29: link[:text],
           },
+          ga4_link: ga4_link,
         }
       end
 
-      def add_data_attributes_to_links(items, track_action)
+      def add_data_attributes_to_links(items, track_action, ga4_data)
```
AshGDS, line 130: This is where all the links get looped through, and then it calls the `generate_data_attribute` and adds the result of that function to the link.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -6,8 +6,9 @@
   with_border ||= false
   classes = %w(gem-c-layout-footer govuk-footer)
   classes << "gem-c-layout-footer--border" if with_border
+  ga4_index_total ||= nil
```
AshGDS, line 9: This is for the OGL / Crown Copyright link to know what the `index-total` value is.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
       attr_reader :footer_navigation, :footer_meta, :cookie_banner_data
 
       def initialize(local_assigns)
+        @footer_links_count = generate_ga4_footer_links_count(local_assigns)
         @footer_navigation = local_assigns[:footer_navigation] || navigation_link_generation_from_locale(I18n.t("components.layout_footer.navigation_links"))
```
AshGDS, line 50: This is where the `Topics / Government Activity` links are generated. Although there is a translation being done here, it looks like there's only an English locale file in the repo, so I haven't had to do any work to ensure things are tracked in English, which would've been a faf!

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
-        @footer_meta = local_assigns[:footer_meta] || { items: add_data_attributes_to_links(FOOTER_META[:items], "supportLink") }
+        @footer_meta = local_assigns[:footer_meta] || { items: add_data_attributes_to_links(FOOTER_META[:items], "supportLink", { section: "Support links", index_section: 3 }) }
         @cookie_banner_data = local_assigns[:cookie_banner_data] || {}
       end
 
+      def generate_ga4_footer_links_count(local_assigns)
```
AshGDS, line 55: Just looping through the `Topics`, `Government Activity`, and `Meta` links and totalling the count of them. I then `add 2` at the end to account for the `OGL` and `Crown Copyright` link.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
andysellick: Is there any scenario where this wouldn't be passed, and it would default to `nil`?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -89,6 +90,19 @@
...
           data-track-options='{"dimension29": "Open Government Licence v3.0"}'
           data-track-links-only
+          data-ga4-track-links-only
+          data-ga4-link="<%= {
+            "event_name": "navigation",
+            "section": "Licence",
+            "index": {
+              "index_section": "4",
+              "index_link": "1",
+              "index_section_count": "5",
```
andysellick, line 100: Nitpick: do these numbers need to be quoted?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -102,6 +116,18 @@
         data-track-label="https://www.nationalarchives.gov.uk/information-management/re-using-public-sector-information/uk-government-licensing-framework/crown-copyright/"
         data-track-options='{"dimension29": "© Crown copyright"}'
         data-track-links-only
+        data-ga4-link="<%= {
+            "event_name": "navigation",
```
andysellick, line 120: Nitpick: indentation looks like 4 chars instead of 2?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
+            footer_links_count += column[:items].length
+          elsif column[:menu_contents]
+            footer_links_count += column[:menu_contents].length
+          end
+        end
+
+        footer_meta = local_assigns[:footer_meta] || FOOTER_META
+        footer_links_count += footer_meta[:items].length
+
+        extra_footer_links = 2 # Amount of links to add to @footer_links_count. This accounts for the OGL link and Crown Copyright link.
```
andysellick, line 71: Not sure you need to declare this as a variable as it only gets used on the next line?

```Ruby
# add 2 to the links count to account for the OGL link and Crown Copyright link.
footer_links_count += 2
```

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -46,17 +46,48 @@ class PublicLayoutHelper
...
+
+        footer_links_count += extra_footer_links
+        footer_links_count
+      end
+
+      def get_footer_links_count
+        @footer_links_count
+      end
```
andysellick, line 79: I don't think you need to have this method to return `@footer_links_count`. If you want to access an instance variable like this from outside the class, you can use `attr_reader` e.g. https://github.com/alphagov/govuk_publishing_components/blob/main/lib/govuk_publishing_components/presenters/shared_helper.rb#L4

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
-      def generate_data_attribute(link, track_action)
+      def generate_data_attribute(link, track_action, ga4_attributes)
+        if ga4_attributes
+          ga4_link = {
+            "event_name": "navigation",
+            "type": "footer",
+            "index": {
+              "index_link": ga4_attributes[:index_link].to_s,
+              "index_section": ga4_attributes[:index_section].to_s,
+              "index_section_count": "5",
```
andysellick, line 111: Is `index_section_count` always 5? We've got examples in the component guide where the footer doesn't have all the links, does this line need to call the counting method?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
+            }, # //Section.link number notation
```
andysellick, line 112: Nitpick: I'm confused by this comment, which line does it refer to?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
+            "index-total": @footer_links_count.to_s, # //Across all sections
+            "section": ga4_attributes[:section],
+            # //Use 'Support links', 'Licence' and 'Copyright' for sections 3, 4 & 5
```
andysellick, line 115: Nitpick: what does this comment refer to?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
+        if ga4_attributes
+          ga4_link = {
+            "event_name": "navigation",
+            "type": "footer",
+            "index": {
+              "index_link": ga4_attributes[:index_link].to_s,
+              "index_section": ga4_attributes[:index_section].to_s,
+              "index_section_count": "5",
+            }, # //Section.link number notation
+            "index-total": @footer_links_count.to_s, # //Across all sections
```
andysellick, line 113: Nitpick: `//` in these comments seems unnecessary?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
         FOOTER_TRACK_ACTIONS
       end
 
-      def generate_data_attribute(link, track_action)
+      def generate_data_attribute(link, track_action, ga4_attributes)
+        if ga4_attributes
```
andysellick, line 104: Is this check needed? I can't see where this method is called without a `ga4_attributes` parameter.

andysellick (commented):
Looking at the component in the preview app, the links in 'Services and information' and 'Departments and policy' don't appear to have any ga4 attributes on (although they are wrapped in the link tracker module). I thought these attributes were needed for the link tracking to function, although I haven't tested this here yet.

Also thanks for the comments you added throughout the PR to explain the code, however I think some of that information would be better written into the PR description or commit descriptions, particularly for future reference.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick Whoops, was meant to remove this - I copied the object straight from the implementation guide then modified it, so those comments are from the implementation guide!

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick Whoops, was meant to remove this - I copied the object straight from the implementation guide then modified it, so those comments are from the implementation guide!

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: @andysellick I think in our docs we had the data type of all our numbers as a string. Our code does work without them being a string but I thought I would be consistent with our documentation even though we do use a regular number in a lot of places. Is this OK?

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick Cool, wasn't sure which method to use for this but I will do that one instead :+1:

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick I think for GOV.UK it is. I guess I could do a counting method but that method will still involve hard coding so I think the only benefit that would give us is that if the Topics / Government activity was given an extra section then this value would auto change, but for the other sections we'd probably have to manually change it.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick Good point, I think the UA tests were failing when I started writing the tests, so I put this in, but then realised the UA tests need the GA4 stuff anyway so this can probably be removed.

AshGDS:
Thanks @andysellick - yeah I wasn't sure where to document the complexities since I'm sort of explaining parts of the code I didn't write. I appreciate that putting them on each line in the PR disrupts the flow of reviewing the code, apologies.

I think something weird is going on where the code that populates the footer with the GOV.UK content is different from the code which lets you build a footer using the component guide examples.

I think this is the line of code that the component guide is using to build links: https://github.com/alphagov/govuk_publishing_components/blob/1d832e049e707eae0217ed98766bdabbeee0fb89/app/views/govuk_publishing_components/components/_layout_footer.html.erb#L44

Whereas I think the GOV.UK footer links are built before it gets to the `.html.erb` file, and then it adds them to this template. (See the `navigation` and `meta` values.) https://github.com/alphagov/govuk_publishing_components/blob/1d832e049e707eae0217ed98766bdabbeee0fb89/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L159 

So it looks like there's the "GOV UK" way to add footer links, and the "Generic" way to add Footer Links which the component guide uses. And I've added tracking to the GOV UK links but not the generic way. 

I'm guessing I should add some more generic link tracking to those default links too? I guess just build the indexes / sections within that loop inside the `.html.erb` and not do anything GOVUK specific.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: @andysellick I think it can be removed, though if the footer ever gets rendered without `ga4_index_total` I think the template would crash without the `||= nil`. Should I remove it?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: @andysellick Indentation is definitely a weakness of mine!

lib/govuk_publishing_components/presenters/public_layout_helper.rb
```diff
@@ -69,20 +100,43 @@ def footer_track_actions
...
             dimension29: link[:text],
           },
+          ga4_link: ga4_link,
         }
       end
 
-      def add_data_attributes_to_links(items, track_action)
+      def add_data_attributes_to_links(items, track_action, ga4_data)
+        ga4_index = 0
         items.map do |item|
```
JamesCGDS, line 132: Quick question -  would using `items.map.with_index do |item, index|` be an option here? I did some quick local testing and it seemed to work fine but no worries if there's a reason why we're not using it - just wanted to check.

AshGDS: @JamesCGDS Good spot - I tried `map_with_index` but didn't realise it was actually `map.with_index` so I'll update it to use that :+1:

AshGDS:
Thanks @andysellick and @JamesCGDS - I've made your changes @JamesCGDS  and have done most of the changes you suggested @andysellick . 

@andysellick I put a rocket emoji 🚀 on the issues that should be ready for you to check and resolve. The rest of the things you identified will probably need a response to what I wrote.

Thanks :+1:

lib/govuk_publishing_components/presenters/public_layout_helper.rb
andysellick: Good stuff. Also I've checked and you don't need to explicitly return the variable by itself as the last line in the method, whatever the last line equates to is what is returned, i.e.

```Ruby
def test
  thing = 5
  thing + 5
  thing
end
```

is equivalent to

```Ruby
def test
  thing = 5
  thing + 5
end
```

(they both return `10`)

lib/govuk_publishing_components/presenters/public_layout_helper.rb
andysellick: In the past we have added more sections to the footer, which would change this number (for example there was a coronavirus section I think, and definitely a Brexit one). 

Rather than block this PR (because no change to these sections is planned in the immediate future) can you record this on the tech debt card as something to either come up with a test that would break or a bit of documentation we could insert somewhere to highlight to future developers that this should change if they add new sections?

andysellick:
> So it looks like there's the "GOV UK" way to add footer links, and the "Generic" way to add Footer Links which the component guide uses. And I've added tracking to the GOV UK links but not the generic way.
> 
> I'm guessing I should add some more generic link tracking to those links populated using the component guide method too? I guess just build the indexes / sections within that loop inside the `.html.erb` .

@AshGDS yeah it's a bit unclear to me too. Presumably you've tested this locally with a local gem + static + application to see if the tracking appears?

AshGDS:
@andysellick Yep, it works in `static`. I think `public_layout_helper.rb` is taking the data required for a footer render, and then adding data attributes to it. Then, it passes that data back to the footer render. Then those links are rendered at that same line of code that the component guide example links would, but these links also have ga4 data attributes added to them because of the transformations that `public_layout_helper.rb` did. That's why the tracking appears fine in `static` because the actual render of the GOVUK footer grabs the link data attributes from `public_layout_helper.rb` and renders them. Whereas in the component guide, the links are just being passed straight to the footer render code without going through `public_layout_helper.rb` so they don't have any GA4 attributes. 

I can't define all the GA4 attributes straight inside the `.html.erb` file instead because I think there is an issue. I think the footer links from `public_layout_helper.rb` aren't merging their data attributes with any attributes defined in the `.html.erb` file. If I'm correct, the `.html.erb` file has this code which should merge the attributes together:

```
 attributes = {
    class: "govuk-footer__link",
    data: {
    "track-category": "footerClicked",
    "ash-says": "hello"
    }
}.merge(item.fetch(:attributes, {}))
```

But it looks like it doesn't do that. Therefore I defined all the GA4 stuff in `public_layout_helper.rb`. Whereas I could've potentially done all the GA4 tracking in the `.html.erb` file instead if the data-attributes merged properly with the existing ones.

lib/govuk_publishing_components/presenters/public_layout_helper.rb
AshGDS: @andysellick Thanks, I've added it to the tech debt doc.

andysellick:
@AshGDS that was very detailed but I'm still struggling to understand. Shall we have a chat about this in dev chat tomorrow?

AshGDS:
@andysellick Sounds good :+1:

PR #3296, opened 2023-03-13
Add GA4 tracking option to tabs component

andysellick:
## What
- previously relied on manually passing data attributes per tab from the application to the component
- but no actual need for this, attributes can be determined inside the component, and simply activated by setting ga4_tracking to true
- does not represent a breaking change as options can be switched over later once this is deployed

Example of current tracking on tabs for comparison: https://www.gov.uk/bank-holidays

## Why
Should make enabling tracking on tabs simpler.

## Visual Changes
None.

Trello card: https://trello.com/c/OSeC4N43/261-epic-technical-debt

### DIFF
--- a/app/views/govuk_publishing_components/components/_tabs.html.erb
+++ b/app/views/govuk_publishing_components/components/_tabs.html.erb
@@ -6,15 +6,33 @@
   panel_css_classes = %w(govuk-tabs__panel)
   panel_css_classes << "gem-c-tabs__panel--no-border" if panel_border == false
   panel_css_classes = panel_css_classes.join(" ")
+
+  ga4_tracking ||= false
+  data_module = "govuk-tabs"
+  data_module << " ga4-event-tracker" if ga4_tracking
 %>
 <% if tabs.count > 1 %>
-  <div class="govuk-tabs gem-c-tabs" data-module="govuk-tabs">
+  <div class="govuk-tabs gem-c-tabs" data-module="<%= data_module %>">
     <h2 class="govuk-tabs__title">
       <%= t("components.tabs.contents") %>
     </h2>
     <ul class="govuk-tabs__list">
-      <% tabs.each do |tab| %>
+      <% tabs.each_with_index do |tab, index| %>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_tabs.html.erb
```diff
@@ -6,15 +6,33 @@
...
+          tab[:tab_data_attributes] ||= {}
+          if ga4_tracking
+            tab[:tab_data_attributes][:ga4_event] = {
+              event_name: "select_content",
+              type: "tabs",
+              text: tab[:label],
+              index: {
+                index_section: index + 1,
+              },
+              index_total: tabs.length,
```
JamesCGDS, line 32: Quick question: should this moved inside the `index` object and renamed to `index_section_count`? If I'm reading this [card](https://trello.com/c/8tNN5gGX/426-change-the-index-parameter) correctly, `index_total` would then be left as `undefined`.

andysellick: I've checked with the analysts, and you are correct. Have updated, thanks.

JamesCGDS (approved):
Looks good! Happy to approve 👍 Although just want to note that when the time comes to deploy this, I’m guessing we’ll also need to raise a corresponding PR in [frontend](https://github.com/alphagov/frontend) to remove `ga4-event-tracker` [here](https://github.com/alphagov/frontend/blob/9412f16ffb962f8b3066e3aeaddbc9251733265e/app/views/calendar/bank_holidays.html.erb#L21) and [here](https://github.com/alphagov/frontend/blob/9412f16ffb962f8b3066e3aeaddbc9251733265e/app/views/transaction/_additional_information_tabbed.html.erb#L1) as I believe this would cause the tracking to be triggered twice since we're now adding `ga4-event-tracker` here too.

andysellick:
@JamesCGDS yes, there'll be a corresponding PR in frontend, but we don't need to worry about that now as frontend won't set `ga4_tracking` to true until we make that change, so no risk of duplicate events.

PR #3300, opened 2023-03-15
Fix redaction bug in GA4 form tracker

andysellick:
## What
- if there were multiple text inputs in an answer we would redact each of them, resulting in '[REDACTED],[REDACTED]' if there were two inputs, and '[REDACTED],[REDACTED],[REDACTED]' if there were three
- better to have it just say it once
- refactoring the code to do this, now easier to not collect these answers in the first place (inside getInputValues()) and instead set a global flag that can be later used to append '[REDACTED]' onto the end of the collected answers
- note that this is necessary because there is a smart answer question that contains a select dropdown and a text input, and we also have some situations where an optional flag is used to not redact text input (`includeTextInputValues`)

## Why
Making the data easier to navigate for analysts.

## Visual Changes
None.

Trello card: https://trello.com/c/JeOLONF8/435-smart-answer-bugs-tidy-up-redacted-value-passed-in-form-events


gclssvglx (approved):
Looks great @andysellick - nice one 👍

PR #3306, opened 2023-03-16
Add GA4 tracking to footer links

AshGDS:
Hi @andysellick / @JamesCGDS - would you be able to approve this if all is OK? Thanks :+1:

Trello card: https://trello.com/c/3s9DA2dY/434-add-tracking-footer-links

Hopefully my code comments are OK, happy to change/reduce them. I guess it's better to over comment and remove them, than under comment and have to explain things!

## What

- Version 2 of the GA4 Link Tracking which only touches the template (`.html.erb` file) instead of any `.rb` files.
- There are 5 sections of links to track in the Footer. These are:
    - Topics section
    - Government Activity section
    - The "Meta" (or Support Links) section (contains link such as: Help, Privacy, Cookies etc.)
    - The Open Government Licence (OGL) Link section
    - The Crown Copyright link section.

## Why

- The previous version was adding the tracking via Ruby to the GOVUK footer links and the passing them to the template, so the links were being passed to the template with the GA4 attributes already on them. In the template, if a link already contains data attributes, then any data attributes that were being set in the template were overwritten, regardless if the keys of the data attributes were unique to the template. It just does a whole "replace" of the data attributes if the link already has one.
    - This led to an issue where none of the component guide example links had any GA4 tracking on them. This was because when I tried adding the GA4 tracking directly on the template, it wouldn't work for any of the GOVUK links. This was because again if there are existing data attributes on a link (and all the GOVUK links already had data attrs on them), then any data attributes in the template were overwritten, as the code would assume the pre-existing data attributes object on the link meant you didn't want to set any new attributes in the template. So to solve this initially, I had just put the GA4 tracking in the Ruby code on the GOVUK links, since that data attributes object in Ruby was given priority over anything set in the template. But I didn't realise this prevented the component guide examples from having any tracking on them, since the component guide links just use the data attributes object set in the template itself, which the GOVUK links aren't using.
- Therefore, I wrote a fix so that any existing data attributes on a link passed to the template will merge with the data attributes defined in the template. This has allowed me to put all the GA4 tracking in the template as the attributes the links have already won't override them. Hopefully this also makes the code easier to understand.

## Expected dataLayer object

image

image

## Visual Changes
None.


app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -26,24 +48,51 @@
...
+                          }
+
+                          # The below code merges the existing data attributes on a link with any set above.
+                          # If a link already contains any of the same data attributes,
+                          # that existing attribute will take priority over any defined above.
+                          # This is so that more complex values can be set with Ruby helpers,
+                          # whereas defaults and simple attributes can be set above.
+
+                          if item[:attributes]
+                            if item[:attributes][:data]
```
andysellick, line 88: Could these two `if` statements be combined?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -26,24 +48,51 @@
...
+                              # Merge the keys of both data attribute objects together
+                              attributes[:data] = attributes[:data].merge(item[:attributes].fetch(:data, {}))
+                              # Delete as there is a parent-level merge after this, and we don't want item[:attributes][:data] overwriting attributes[:data] which now contains the merged key/values.
+                              item[:attributes].delete(:data)
+                            end
+                          end
+                          attributes = attributes.merge(item.fetch(:attributes, {}))
```
andysellick, line 95: I wonder if there's a simpler option here? If we want to have default GA4 attributes but also allow some to be passed to be overridden, could we instead just check to see if they exist and if not, add them? Something like

```Ruby
<% if item[:href] && item[:text] %>
  <li class="govuk-footer__list-item">
    <%
      attributes = {
        class: "govuk-footer__link",
        data: {
          "track-category": "footerClicked",
        }
      }
      attributes = attributes.merge(item.fetch(:attributes, {}))

      unless attributes[:data][:ga4_link]
        attributes[:data][:ga4_link] = {
          "event_name": "navigation",
          "type": "footer",
          "index": {
            "index_link": (index + 1).to_s,
            "index_section": (ga4_section_index + 1).to_s,
            "index_section_count": ga4_index_section_count.to_s
          },
          "index-total": ga4_index_total.to_s,
          "section": section[:title]
        }
      end
    %>
  </li>
<% end %>
```

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -6,15 +6,37 @@
...
+
+  if meta.any?
+    amount_of_meta_sections = 1
+    ga4_index_section_count += 1
+    ga4_index_total += meta[:items].length
+  end
+
+  # For our GOVUK footer links, we need to add 2 to the index_total and index_section_count. This is due to the "Open Government License" and "Crown Copyright" links being defined separately to the rest of our footer links.
+  ga4_index_total += 2
+  ga4_index_section_count += 2
```
andysellick, line 29: There's quite a bit of code being introduced into the view here (also some repetition of code around lines 80 and 135). I think this would be better in a separate helper file.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
andysellick: Having thought about this, I guess you're thinking that we'd maybe pass only part of this, say if we wanted to override just one attribute. That's probably sensible, although what's the case for needing to override any of it? Is this functionality needed at all?

andysellick (commented):
Looking good, just a few questions and nitpicks. Also you probably want to squash the commits together, since this feels like a single change.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: @andysellick Thanks, I can change it to that. The reason I wrote it the previous way is because
I was trying to remove some of the weird behaviour. Your example is simpler to understand, but if someone goes in and tries to add a new data attribute to the `attributes` object in the template, it won't be reflected on any GOVUK footer link. This is because the merge line replaces the _whole_ [:data] object or **whole** [:class] string defined in the template if one already exists on the link already. It doesn't "merge" the values together, it merges keys together, and the keys are [:data] and [:class] so the whole data attributes or class gets overwritten if one already exists on the link, instead of actual values all getting merged together from both objects. Hope that makes sense!

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: @andysellick Thanks, I've moved this to a helper file.

AshGDS:
Thanks @andysellick - it should be ready to re-review again :+1:

app/views/govuk_publishing_components/components/_layout_footer.html.erb
andysellick: Ah okay. So maybe something like this would be helpful? (example taken from the component wrapper helper)

```Ruby
      def extend_object(option, object)
        @options[option] ||= {}
        object.each_key do |key|
          @options[option][key] =
            if @options[option].key?(key)
              "#{@options[option][key]} #{object[key]}"
            else
              object[key]
            end
        end
      end
```

But the main question I think we need to answer is do we need to worry about extending these objects, or can we just assume that our tracking will be in here and not passed in from outside the component?

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -26,25 +29,40 @@
...
                             data: {
-                              "track-category": "footerClicked"
+                              "track-category": "footerClicked",
                             }
-                          }.merge(item.fetch(:attributes, {}))
+                          }
+                          attributes = attributes.merge(item.fetch(:attributes, {}))
                           attributes[:rel] = "noopener" if attributes[:target] == "_blank" && !attributes[:rel]
+
+                          unless attributes[:data][:ga4_link]
```
andysellick, line 53: This section seems to be nearly repeated within the template, could it be moved into the helper? You could write a method to call it with the varying parameters e.g. `footer_helper.get_ga4_attributes((index + 1).to_s, (ga4_section_index + 1).to_s)`.

lib/govuk_publishing_components.rb
```diff
@@ -20,6 +20,7 @@
 require "govuk_publishing_components/presenters/step_by_step_nav_helper"
 require "govuk_publishing_components/presenters/page_with_step_by_step_navigation"
 require "govuk_publishing_components/presenters/public_layout_helper"
+require "govuk_publishing_components/presenters/footer_helper"
```
andysellick, line 23: Can you rename the footer helper to match the component name? One of the things the component auditing does is list if a component has a helper, but it bases its search on what the component is named and currently it's not finding this file. Should work if it's `layout_footer_helper.rb`, I think.

lib/govuk_publishing_components/presenters/footer_helper.rb
```diff
@@ -0,0 +1,33 @@
...
+
+        # For our GOVUK footer links, we need to add 2 to the index_total and index_section_count. This is due to the "Open Government License" and "Crown Copyright" links being defined separately to the rest of our footer links.
+        @ga4_index_total += 2
+        @ga4_index_section_count += 2
+
+        @ga4_ogl_link_index_section = footer_navigation_section.length + @amount_of_meta_sections + 1
+        @ga4_copyright_link_index_section = footer_navigation_section.length + @amount_of_meta_sections + 2
+      end
+
+      attr_reader :ga4_index_total, :ga4_index_section_count, :ga4_ogl_link_index_section, :ga4_copyright_link_index_section
```
andysellick, line 30: Nitpick: I think normally our convention is to put the `attr_reader` line at the top of the file, for more visibility.

andysellick (commented):
Nice helper addition, just a couple more questions/suggestions.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -63,11 +81,31 @@
...
           <ul class="govuk-footer__inline-list govuk-!-display-none-print" data-module="gem-track-click" data-track-links-only>
-            <% meta[:items].each do |item| %>
+            <% meta[:items].each_with_index do |item, index| %>
               <li class="govuk-footer__inline-list-item">
                 <%
-                  attributes = { class: "govuk-footer__link" }.merge(item.fetch(:attributes, {}))
+                  attributes = {
+                    class: "govuk-footer__link",
+                    data: {
+                      "track-category": "footerClicked",
```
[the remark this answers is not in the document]
AshGDS: @andysellick Good spot, will remove.

app/views/govuk_publishing_components/components/_layout_footer.html.erb
AshGDS: Thanks, moved to a helper method.

lib/govuk_publishing_components.rb
```diff
@@ -20,6 +20,7 @@
 require "govuk_publishing_components/presenters/step_by_step_nav_helper"
 require "govuk_publishing_components/presenters/page_with_step_by_step_navigation"
 require "govuk_publishing_components/presenters/public_layout_helper"
+require "govuk_publishing_components/presenters/layout_footer_helper"
```
andysellick, line 23: Nitpick: could this be positioned alphabetically? I know not all of the others are, but they're sort of mostly alphabetical...

andysellick (commented):
Two little nitpicks, and I then I think we're good to go 🚀 

One other thing I'm not sure we've answered yet - do we need to be able to pass in data attributes to override the GA4 data attributes? The footer is only used in one place, if we need to change the attributes could they not just be changed in the code here?

AshGDS:
@andysellick Thanks Andy - added both of those changes. I don't think we'd need to override the GA4 attributes, so we can leave it how it is for now.

andysellick (approved):
Remember to squash your commits 👍  🚀

PR #3308, opened 2023-03-17
Update error summary to use component wrapper helper

jon-kirwan:
## What
Update error summary to use [component wrapper helper](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component-wrapper-helper.md)

## Why

Resolves an issue whereby the data attribute for `module: "govuk-error-summary"` is overwritten when the same option is passed to the component. For example:

```
<%= render "govuk_publishing_components/components/error_summary", {
  data_attributes: {
    module: "auto-track-event"
  },
} %>
```
## Visual Changes
None

## Anything else
See https://github.com/alphagov/frontend/pull/3562

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Always require AssetHelpers so gem can operate without Rails ([PR #3309](https://github.com/alphagov/govuk_publishing_components/pull/3309))
 * Update AssetHelper documentation ([PR #3298](https://github.com/alphagov/govuk_publishing_components/pull/3298))
 * Ensure color contrast in code examples meets accessibility requirements ([PR #3311](https://github.com/alphagov/govuk_publishing_components/pull/3311))
+* Update error summary to use component wrapper helper ([PR #3308](https://github.com/alphagov/govuk_publishing_components/pull/3308))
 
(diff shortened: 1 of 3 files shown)


jon-kirwan:
> @jon-kirwan so great to see someone else implementing this! If you have time, can you add an entry to the [audit spreadsheet](https://docs.google.com/spreadsheets/d/1J-0IPyr2ycPuw1IWOIoX0o9Tk9nF35D0DJdsj_ylLQE/edit?pli=1#gid=1458513074)?

Sure no problem - now updated! 👍

AshGDS (changes requested):
Looks good, but I think you need to add `uses_component_wrapper_helper: true` to the `.yml` doc?

jon-kirwan:
> Looks good, but I think you need to add `uses_component_wrapper_helper: true` to the `.yml` doc?

Good spot! Thank you. I've now updated the yaml docs to include the line

andysellick (approved):
Great to see this being implemented in other components. In theory you could remove the example from the component guide that demonstrates data attributes, now that it's included as part of the 'standard' options.

AshGDS (approved):

PR #3318, opened 2023-03-23
Make auditing aware of new asset loading model

andysellick:
## What
Update the auditing so that it suppresses warnings relating to missing stylesheets for applications that use the new per page component asset loading model. This model is currently only implemented in `frontend`, but will be rolled out to other applications soon.

Also do some minor cleaning up of the UI, including adding a notice for the above where relevant.

## Why
To remove misleading auditing warnings.

## Visual Changes
Example of what `frontend` looks like in the auditing now:

Screenshot 2023-03-23 at 10 07 17

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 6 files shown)


MartinJJones (approved):
Tested locally and the changes look good to me, thanks for implementing this 👍

PR #3319, opened 2023-03-24
Fix abbr element spacing in govuk-button--start 

patrickpatrickpatrick:
## What

Add spacing to `abbr` so that in a link with `govuk-button--start` applied to it, there is spacing between the inline element and the text of the link.

## Why

If using an inline element (such as `<abbr>`, `<address>` or `<telephone>`) in a link with the `govuk-button--start` class applied to it, there will not be any spacing between the inline element and the text of the link. This issue came to our attention on this [page](https://www.gov.uk/government/organisations/department-for-work-pensions/about/accessible-documents-policy) of GOV.UK which was authored in Whitehall. 

I think this change in govspeak CSS should be temporary. Have raised an issue in [govuk-frontend](https://github.com/alphagov/govuk-frontend/issues/3406) but not sure what timeline of that fix will be.

### Before

Screenshot 2023-03-24 at 09 42 45

### After

Screenshot 2023-03-24 at 09 43 21

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
@@ -24,4 +24,11 @@
   .govuk-button--start {
     @extend .govuk-button--start;
   }
+
+  // Temporary fix for issue where there is no spacing between text
+  // and abbr elements in the govuk-button--start styled links
+
+  .govuk-button--start abbr {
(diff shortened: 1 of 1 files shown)


patrickpatrickpatrick:
Closing because there's an alternative approach to the issue which doesn't require changes to govuk_publishing_components.

matthillco:
Cool, I'd be interested to know what the alternative solution is?

PR #3320, opened 2023-03-24
Update readme to include info on using AssetHelper

MartinJJones:
Updated the documentation in the readme file to include further information loading stylesheets individually.


docs/set-up-individual-component-css-loading.md
```diff
@@ -166,6 +166,12 @@ Use `yield` to add body content between `<body>` and `</body>`:
   </head>
 ```
 
+### Check stylesheets are loading individually
+
+When the setup is complete, only the component stylesheets required to render the page are requested. For example, the screenshot below shows separate network requests for the `browse`, `lead-paragraph` and `cards` stylesheets:
+
+![Network requests for individual stylesheets](./images/individual-stylesheets.png)
+
```
jon-kirwan, line 174: Perhaps it could be helpful to say where the example screenshot is taken from e.g. 

> for example, the screenshot below _displays network activity for a [browse page e.g. benefits](https://www.gov.uk/browse/benefits) and_ shows separate network requests for the `browse`, `lead-paragraph` and `cards` stylesheets.

jon-kirwan (commented):
Sorry for the delay! It looks good and I think it's helpful to include. I've added one small suggestion below. I also wondered whether displaying the full Network panel in the screenshot which would show the "Network" tab and the "CSS" filter selected would be helpful too?

docs/set-up-individual-component-css-loading.md
MartinJJones: Thanks, makes sense to me, I've made the suggested changes.

jon-kirwan (approved):
A helpful addition to the docs. Thanks for the changes!

PR #3322, opened 2023-03-24
Fix search icon cut-off in navbar

patrickpatrickpatrick:
## What

Adjust width of the search icon. As the width and height of the svg are different and the icon-size variable is not used anywhere else, have removed icon-size variable.

## Why

Stem of icon was being slightly cut-off.

### Before

Screenshot 2023-03-24 at 15 33 16

### After

Screenshot 2023-03-24 at 15 32 43


andysellick (approved):
Nice detail.

PR #3324, opened 2023-03-28
Make the component wrapper more robust

andysellick:
## What
Improve the component wrapper helper in the following ways.

- let it recognise if blank or nil values are passed for certain things. Previously it was possible to pass `nil` as an ID attribute, at which point the wrapper would error because `nil` is not a valid HTML ID, rather than simply ignoring it
- improve the error messages to include the thing that is causing the error, in order to make troubleshooting issues easier

## Why
`Frontend` had a bit of code where `nil` was being [passed to a component as an ID](https://github.com/alphagov/frontend/commit/44a16abd2888cb0dddbb4a2d01f5a75477bbb089), which was causing the build to fail. This change should prevent similar errors in the future.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Update to Lux 307 ([PR #3329](https://github.com/alphagov/govuk_publishing_components/pull/3329))
 * Drop support for Ruby 2.7 ([PR #3310](https://github.com/alphagov/govuk_publishing_components/pull/3310))
 * Update GA4 index parameter ([PR #3297](https://github.com/alphagov/govuk_publishing_components/pull/3297))
+* Make the component wrapper helper more robust ([PR #3324](https://github.com/alphagov/govuk_publishing_components/pull/3324))
 * Make auditing aware of new asset loading model ([PR #3318](https://github.com/alphagov/govuk_publishing_components/pull/3318))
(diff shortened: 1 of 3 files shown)


jon-kirwan (approved):
All seems good and works fine for me (I've tested the same `frontend` page).

The only comment I have, which would be a separate issue, is whether we should add another method to check for valid or accepted data attributes (for things like "track-" or "module") but an invalid data attribute is going to to have less impact than an invalid class or ARIA attribute so maybe not necessary.

andysellick:
> The only comment I have, which would be a separate issue, is whether we should add another method to check for valid or accepted data attributes (for things like "track-" or "module") but an invalid data attribute is going to to have less impact than an invalid class or ARIA attribute so maybe not necessary.

Thanks @jon-kirwan. Very interesting question! I'd omitted checking this so far because data attributes can be anything, so long as they start with `data-`, but thinking about it now I realise you could check for some invalid input at least (e.g. with spaces in) so will have a think.

andysellick:
@jon-kirwan I've had a go at validating data attributes - have added an extra commit if you'd like to take a look when you have time, thanks.

jon-kirwan:
Nice one! All looks good to me. I've tested from the component guide and an application (`feedback`) entering various invalid data attribute combinations (e.g. `moDule`, `mo dule`) which are correctly raise an ArgumentError with a warning message: `Data attributes (moDule) cannot contain capitals, spaces or colons, or start with 'xml'"`

PR #3328, opened 2023-03-30
Ga4 accordion link tracking

JamesCGDS:
## What
This PR adds GA4 tracking to accordion links. 

This PR also includes a change to the `setIndex` function and it will now combine `data-ga4-index` (our method of adding `index_link` to links that are generated by content publishers ([further details here)](https://github.com/alphagov/govuk_publishing_components/pull/3262)) with the `index` property if it exists on a parent element.

## Why
Part of the GA4 migration.

The change to `setIndex` has been made because in some scenarios, such as accordion links, we need to set `data-ga4-link` on a parent element due to the links being generated by a content publisher. Within `data-ga4-link`, an `index` property may already exist and contain values such as `index_section` and `index_section_count`. In order to prevent this data from being overwritten with the value in `data-ga4-index` (the previous implementation set `index` to the value of `data-ga4-index` if it existed), `setIndex` combines them together.

[Trello card](https://trello.com/c/F8LlOd07/490-add-tracking-accordion-links)

## Visual Changes
N/A

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -108,28 +108,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   Ga4LinkTracker.prototype.setIndex = function (indexData, target) {
-    var index
+    var index = window.GOVUK.analyticsGa4.core.trackFunctions.createIndexObject(indexData)
 
     if (target.getAttribute('data-ga4-index')) {
       try {
-        index = JSON.parse(target.getAttribute('data-ga4-index'))
+        var indexLink = JSON.parse(target.getAttribute('data-ga4-index'))
 
         // Check whether the index object already exists on a parent element, as is the case with tracking accordion links.
-        // If true, combine data-ga4-index with the index object
-        if (typeof indexData === 'object') {
-          for (var i = 0; i < Object.keys(indexData).length; i++) {
-            index[Object.keys(indexData)[i]] = indexData[Object.keys(indexData)[i]]
-          }
-        }
+        // If true, combine data-ga4-index with the index object. Otherwise, just return indexLink
+        index = index ? window.GOVUK.extendObject(index, indexLink) : indexLink
       } catch (e) {
         console.error('Unable to parse index as JSON: ' + e.message, window.location)
         return
       }
-    } else {
-      index = indexData || undefined
     }
 
-    return window.GOVUK.analyticsGa4.core.trackFunctions.createIndexObject(index)
+    return index
   }
 
   Modules.Ga4LinkTracker = Ga4LinkTracker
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -113,6 +113,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     if (target.getAttribute('data-ga4-index')) {
       try {
         index = JSON.parse(target.getAttribute('data-ga4-index'))
+
+        // Check whether the index object already exists on a parent element, as is the case with tracking accordion links.
+        // If true, combine data-ga4-index with the index object
+        if (typeof indexData === 'object') {
```
AshGDS, line 119: Could this be done when the indexes are first set (the `setIndexes` function) instead? Just because we set create `data-ga4-index` ourselves, but then it seems like we're fixing it here, when we have control over both functions. Or would doing it in there instead affect other components?

app/views/govuk_publishing_components/components/_accordion.html.erb
```diff
@@ -49,6 +50,24 @@
               index_section_count: items.length,
             },
           }.to_json
+
+          # Data attributes for accordion links. These have been created
```
AshGDS, line 54: I think you could get away with not having this comment, I think it's clear from the code that it's separate from the other data_attributes.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS, line 1: I guess this is something for a separate PR but as we spoke about we need to make sure we always push `index_link: undefined, index_section: undefined, index_section_count: undefined` etc. when only one or two of the index keys have values, as I believe GA4 persists values from previous pushes so we need to make sure empty values are still pushed with their value set to `undefined`. (It's fine to just do `index: undefined` when none of the index keys exist though)

AshGDS (commented):
Looking good, had some comments which are more questions than requested changes.

app/views/govuk_publishing_components/components/_accordion.html.erb
JamesCGDS: Cool, I'll remove it

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
JamesCGDS: I think this might be for another PR - I did discuss this with Andy and we're not sure if `index_section` etc would persist because we'd be overwriting the `index` object itself with a new `index` object if that makes sense. Although we're not 100% sure so I'm going to pick this up with the PA's next week.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
JamesCGDS: Potentially yeah, I'll need to have a quick think about it

andysellick (commented):
Good improvements, nice job, that function is a lot more understandable and simpler now. I'm happy with this but I think Ash has some comments outstanding, so I'll leave him to approve it when he's happy.

AshGDS (approved):
All good :+1:

JamesCGDS:
Thanks @AshGDS @andysellick - I'll squash the commits together but I'm going to hold off on merging this as the PA's have asked for this not to be deployed yet.

PR #3330, opened 2023-03-30
Add data attributes to attachment links using JS

AshGDS:
## What

Sets the GA4 data attributes for attachment links via JavaScript, instead of setting them directly on the attachment links' HTML.

## Why

Initially we were going to set our GA4 data attributes directly on each attachment links' HTML - as shown in https://github.com/alphagov/whitehall/pull/7481. However, as the HTML lives in `whitehall`, deploying the new HTML to our servers requires us to republish 45k+ documents which takes a bit of time.

This could become an issue if we constantly need to make changes to the data attributes, as we'd need to republish 45k documents every time. It also feels a bit like technical debt to implement something that would require 45k republishes to update.

Therefore, in `whitehall` we are going to set an attribute called `data-ga4-attachment-links` on the HTML of each attachment link once. Then, we can add the actual data attributes using JS by grabbing every instance of `data-ga4-attachment-link`. This will stop us having to republish 45k+ documents if the GA4 attributes ever need changing/updating. With this technique, we will only need to do the 45k+ republish once, just to add the tag `data-ga4-attachment-link` on everything.

I put the function call in `init-ga4.js` as it needs to run before the `ga4-link-tracker` module runs, otherwise the module wouldn't get initialised on the HTML if `data-module='ga4-link-tracker'` was added to each link after the tracker was initialised.

## Testing

The JS tests should demonstrate that it works, however if you want to see it for yourself:

1. Run `static` locally with a local version of `govuk_publishing_components`
2. Run `government-frontend`, pointing to your local `static`
3. Go to `init-ga4.js` in `govuk_publishing_components` and on `line 15` change the CSS selector we're using from `[data-ga4-attachment-link]` to `section.attachment.embedded`.
4. If you go to http://127.0.0.1:3090/government/publications/equality-act-guidance you will see that every attachment link has the GA4 link attributes - this is because they are section elements with the classes `attachment embedded`, so our GA4 code has added the attributes to every element matching `section.attachment.embedded`. Of course in our real solution we are going to be adding a GA4 specific selector `[data-ga4-attachment-link]` to the links we need via `whitehall`.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -273,6 +273,31 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
               index_link: parseInt(index[2])
             }
         }
+      },
+
+      addAttributesToElements: function (selector, dataAttributes) {
+        var targetElements = document.querySelectorAll(selector)
+
+        for (var i = 0; i < targetElements.length; i++) {
+          var element = targetElements[i]
+
+          for (var j = 0; j < dataAttributes.length; j++) {
+            var key = dataAttributes[j].key
+            var value = dataAttributes[j].value
+
+            // value must check for undefined only as it would return false on an empty string, the number 0, etc.
+            if (key && value !== undefined) {
+              var existingAttributeValue = element.getAttribute(key)
+
+              if (key === 'data-module' && existingAttributeValue) {
+                // Combines values to prevent replacing any existing data-module values.
+                element.setAttribute(key, existingAttributeValue + ' ' + value)
+              } else {
+                element.setAttribute(key, value)
+              }
+            }
+          }
+        }
       }
     },
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add data attributes to attachment links using JS [(PR #3330)](https://github.com/alphagov/govuk_publishing_components/pull/3330)
+
 ## 35.2.0
 
 * Update to Lux 307 ([PR #3329](https://github.com/alphagov/govuk_publishing_components/pull/3329))
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -8,6 +8,11 @@ var initFunction = function () {
     window.GOVUK.analyticsGa4.vars.internalDomains = []
     window.GOVUK.analyticsGa4.vars.internalDomains.push(window.GOVUK.analyticsGa4.core.trackFunctions.getHostname())
     window.GOVUK.analyticsGa4.core.trackFunctions.appendDomainsWithoutWWW(window.GOVUK.analyticsGa4.vars.internalDomains)
+    var attachmentLinkData = [
+      { key: 'data-module', value: 'ga4-link-tracker' },
+      { key: 'data-ga4-track-links-only', value: '' },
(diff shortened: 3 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -273,6 +273,27 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
               index_link: parseInt(index[2])
             }
         }
+      },
+
+      addAttributesToElements: function (targetElementsSelector, dataAttributes) {
+        var targetElements = document.querySelectorAll(targetElementsSelector)
```
JamesCGDS, line 279: Quick question: would we want to wrap the code in this function in a guard condition to check that `targetElementsSelector` exists so that the code doesn't run when it doesn't need to? Having said that, I believe `querySelectorAll` will just return an empty `nodeList` on elements that don't exist so I don't think this code would get very far anyway.

Also sidenote nitpick: what do you think about renaming `targetElementsSelector` to just `selector`? It made it a bit more readable for me but I appreciate this very much comes down to personal preferences so happy to leave it as it is!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -273,6 +273,27 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
+
+        for (var i = 0; i < targetElements.length; i++) {
+          var element = targetElements[i]
+
+          for (var j = 0; j < dataAttributes.length; j++) {
+            var key = dataAttributes[j].key
+            var value = dataAttributes[j].value
```
JamesCGDS, line 286: If I'm understanding this code correctly, it looks like we're depending on `key` and `value` existing on `dataAttributes[j]` - I've been (possibly over) thinking this and I'm wondering if setting `var key` and `var value` to something like `dataAttributes[j][Object.keys(dataAttributes[j])[0]]` and `dataAttributes[j][Object.keys(dataAttributes[j])[1]]` respectively would make things more robust as it wouldn't matter what the properties on `dataAttributes[j]` are called.

Having said that, I don't think it's a very readable solution and it could be hard to understand. Alternatively, do you think another option could be wrapping this code in another guard condition - something like `if(dataAttributes[j].hasOwnProperty('key') && dataAttributes[j].hasOwnProperty('value')) { ... }`?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: Yeah, if the list is empty the for loop just won't execute so I think it should be fine to stay how it is.

I can change it to `selector` instead :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: Thanks - I think the second alternative you suggested is probably worth adding. 

I am wondering about our wider approach to "checking data passed to a function is correct" though, because the function would also crash if someone passed a number or object to the `selector` value - and if we were checking for that as well to prevent crashes, functions would start getting big.

My personal view is that when the data passed into a function is fully controlled by developers, it should be OK not to do any strict checking. This is because I think a developer should be responsible enough to ensure that what they're passing into a function is correct. However, when the data being passed to a function is unreliable (e.g. if the data passed into the function is coming from the DOM or an API), then I think there should be stricter checks. What do you think?

@andysellick Do you have any views on strict input checking on functions? I couldn't find one in our JS coding style guide.

andysellick: I don't think we've ever written down anything to this effect, but your approach of the level of checking being dependent on where the data is coming from sounds sensible.

AshGDS:
Thanks @JamesCGDS - I've made the requested changes.

PR #3334, opened 2023-04-05
Fix GA4 Step Nav link tracking bug

AshGDS:
Hi @JamesCGDS - would you be able to approve this if all is OK? Thanks :+1: 

## What

Refactors our GA4 Step by Step nav link tracking to use `data-ga4-track-links-only` and `data-ga4-set-indexes` on the parent container of each step, instead of adding a separate `data-ga4-link` JSON on each link within a step.

## Why

In the current implementation of the step nav link tracking, we were adding `data-ga4-link` to every link via Ruby. This caused an issue where the `index_total` was incorrect. This bug occurs because links can be added via content, or via list items. Therefore when calculating the `index_total` in Ruby, the maths for the content links `index_total` and the maths for the lists `index_total` were separate and couldn't be combined to make the `index_total` correct.  Therefore in our current implementation if there was a link added via content, that link would have `index_total: 1`. If there was a list immediately after that content link, which contained four links, those links in the list would have `index_total: 4`. The correct value would be `index_total: 5` as they are all links under the same step.

You can see an example of the bug here: https://www.gov.uk/when-someone-dies - on Step 1, the first link has `index_total: 1` but the links in the list have `index_total: 4`.

Therefore, if we use the `data-ga4-track-links-only` and `data-ga4-set-indexes` technique instead, we can be sure that the link indexes and index_total will be correct, as JavaScript will set the indexes for us so it won't matter if the links came from code or content.

## Visual Changes

None.


JamesCGDS:
Quick question: I've noticed that `data-ga4-link` is duplicated on individual step-by-step section elements and the links contained within that section. In the latter case, the incorrect `index_total` is still being pulled through. Just wanted to check if this is expected behaviour/if my local setup is doing something weird. I believe it's caused by this bit of code on line 103:

`options[:ga4_tracking] = ga4_tracking`

Hope I've made sense here, happy to jump on a call if I haven't!

JamesCGDS:
@AshGDS - please disregard the above, seems my local setup was doing something weird.

JamesCGDS (approved):
Looks good 👍

PR #3346, opened 2023-04-17
Update GA4 index parameter on related navigation

andysellick:
## What
Update the GA4 `index` parameter on the related navigation and contextual sidebar components. Specifically:

- break the `index` parameter into `index_section` and `index_link` sub parameters
- add a new `index_section_count` sub parameter
- add the `index_total` parameter

## Why
Part of the GA4 migration. We've changed how the index parameters should work, this is to reflect that.

## Visual Changes
None.

Trello card: https://trello.com/c/16ngDhBP/493-change-the-index-parameter-part-2-indexsectioncount

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,11 +9,12 @@
 
 ## Unreleased
 
-* Add brand colour for DSIT [(PR #3349)](https://github.com/alphagov/govuk_publishing_components/pull/3349)
+* Add brand colour for DSIT ([PR #3349](https://github.com/alphagov/govuk_publishing_components/pull/3349))
+* Update GA4 index parameter on related navigation ([PR #3346](https://github.com/alphagov/govuk_publishing_components/pull/3346))
 
 ## 35.3.0
 
-* Add data attributes to attachment links using JS [(PR #3330)](https://github.com/alphagov/govuk_publishing_components/pull/3330)
+* Add data attributes to attachment links using JS ([PR #3330](https://github.com/alphagov/govuk_publishing_components/pull/3330))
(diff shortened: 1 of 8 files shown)


AshGDS (approved):
Looks good. Question, what's the benefit of using `OpenStruct` over a regular hash?

CHANGELOG.md
```diff
@@ -9,7 +9,8 @@
 
 ## Unreleased
 
-* Add data attributes to attachment links using JS [(PR #3330)](https://github.com/alphagov/govuk_publishing_components/pull/3330)
+* Update GA4 index parameter on related navigation ([PR #3346](https://github.com/alphagov/govuk_publishing_components/pull/3346))
+* Add data attributes to attachment links using JS ([PR #3330](https://github.com/alphagov/govuk_publishing_components/pull/3330))
```
AshGDS, line 13: Just noticed, this is in the wrong place, this was already released.

andysellick:
> Looks good. Question, what's the benefit of using `OpenStruct` over a regular hash?

Great question. Apparently it's a bit of Rails magic, when it gets created in the parent component (contextual sidebar) and then updated in the child component (related nav) any changes in the child are reflected in the parent, i.e. it persists, but a regular hash wouldn't do that. This means we can count the sections in related nav then reflect that in the ukraine sidebar (part of the contextual sidebar).

CHANGELOG.md
andysellick: Good spot, something weird happened there. Fixed now, thanks!

AshGDS (approved):

PR #3347, opened 2023-04-18
Fix component guide previews for application components

MartinJJones:
## What

- Call `render_component_stylesheets` to load stylesheets individually
- Update application.css used in the component guide to request only the minimum number of stylesheets required to render components correctly
- Remove `config.assets.digest = false` setting in `test.rb`, this allows the default value of `true` to be used - https://guides.rubyonrails.org/configuring.html#config-assets-digest
- Check if the frontend rendering application is using an application.css stylesheet before including it in the component guide view templates

## Why

- Ensure component previews work in rendering applications that are loading stylesheets individually
- Removing `config.assets.digest = false` fixes a failing test when using the individual loading of stylesheets feature in the gem and matches the default behaviour for development and production environments
- Some applications no longer have an application.css file. For these applications, `application_stylesheet` should be set to `false` to avoid the error below:
```
Sprockets::Rails::Helper::AssetNotPrecompiledError at /taxon-list
Asset `application.css` was not declared to be precompiled in production.
Declare links to your assets in `app/assets/config/manifest.js`.
  //= link application.css
and restart your server
```

Trello: https://trello.com/c/hisIwM5O/1966-fix-component-previews-in-component-guide

### No application.css file

When viewing the component guide in a rendering application, the application.css is always requested for that application. However, some applications are not using an application.css file, which then causes a server error. We need to add a check to ensure that we only try to load this file when required.

To reproduce this issue:

- run the collections app locally
- visit `/component-guide`
- click on the link for any of the application components, you will then get an error similar to that below:

```
Sprockets::Rails::Helper::AssetNotPrecompiledError at /taxon-list
Asset `application.css` was not declared to be precompiled in production.
Declare links to your assets in `app/assets/config/manifest.js`.
  //= link application.css
and restart your server
```

**Note**: if you cannot reproduce the error, try running bundle exec rails assets:clobber assets:precompile to remove any existing assets and compile them again

### App levels components not rendered correctly

This issue is caused because the component stylesheet is not loaded for app level components in the component guide, we just need to add support for this.

To reproduce this issue:

- run the finder-frontend app locally
- visit `/component-guide`
- click on any of the application components, date-filter, expander or option-select.

## Percy - Visual changes

Percy is reporting the 2 visual changes below, these only impact the component guide, not the live rendering applications.

### Character count component

This is because the `textarea` element now has `margin-bottom: 5px` instead of `0px`, this is correct and is what is used in both the Design System and GOV.UK

- [Design System Character Count example](https://design-system.service.gov.uk/components/character-count/)
- [GOV.UK Contact page with `texarea` example](https://www.gov.uk/contact/govuk)

### Label component

The visual difference is because the change in how stylesheets are loaded in the component guide, the required stylesheet from the radio component are not included, this is mentioned in the component guide:

> Note that this example will not render correctly - do not be alarmed. It only works properly when inside the radio component.

https://components.publishing.service.gov.uk/component-guide/label/inside_a_radio_component

### DIFF
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -1,6 +1,42 @@
 // This file contains the styles for the Component Guide.
 
-@import "govuk_publishing_components/all_components";
+// feature flag for accessible link styles
+$govuk-new-link-styles: true;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
+
+// Import the same stylesheets used in static
+// https://github.com/alphagov/static/blob/198a598682df40ce4a2c3c286c06244297c18cf0/app/assets/stylesheets/application.scss
+// Although the component guide does not use static, we still need to import the same stylesheets used in static
+// to avoid any rendering issues
+// By following the same approach as frontend rendering applications,
+// we ensure that stylesheets are loaded in the expected order and components render correctly
+@import "govuk_publishing_components/components/breadcrumbs";
+@import "govuk_publishing_components/components/button";
+@import "govuk_publishing_components/components/error-message";
+@import "govuk_publishing_components/components/heading";
+@import "govuk_publishing_components/components/hint";
+@import "govuk_publishing_components/components/input";
+@import "govuk_publishing_components/components/label";
+@import "govuk_publishing_components/components/search";
+@import "govuk_publishing_components/components/skip-link";
+@import "govuk_publishing_components/components/textarea";
+@import "govuk_publishing_components/components/title";
+
+@import "govuk_publishing_components/components/cookie-banner";
+@import "govuk_publishing_components/components/feedback";
+@import "govuk_publishing_components/components/layout-footer";
+@import "govuk_publishing_components/components/layout-for-public";
+@import "govuk_publishing_components/components/layout-header";
+@import "govuk_publishing_components/components/layout-super-navigation-header";
+
+// Imported to ensure inverse-header component examples render correctly
+// this helps avoid any issues with visual regressions tests
+@import "govuk_publishing_components/components/lead-paragraph";
+
+// Include required helpers
+@import "../../stylesheets/govuk_publishing_components/components/helpers/markdown-typography";
 
 $gem-guide-border-width: 1px;
 
--- a/app/views/govuk_publishing_components/component_guide/example.html.erb
+++ b/app/views/govuk_publishing_components/component_guide/example.html.erb
@@ -1,6 +1,6 @@
 <% content_for :title, "#{@component_example.name} example - #{@component_doc.name} component" %>
 <% content_for :application_stylesheet do %>
-  <% if @component_doc.source == "application" %>
+  <% if @component_doc.source == "application" && application_stylesheet_in_use? %>
     <%= render 'application_stylesheet' %>
   <% end %>
 <% end %>
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix some GA4 index bugs ([PR #3375](https://github.com/alphagov/govuk_publishing_components/pull/3375))
 * Rename type breadcrumbs to breadcrumb ([PR #3373](https://github.com/alphagov/govuk_publishing_components/pull/3373))
 * Update AssetHelper documentation ([PR #3378](https://github.com/alphagov/govuk_publishing_components/pull/3378))
+* Fix component guide previews for application components ([PR #3347](https://github.com/alphagov/govuk_publishing_components/pull/3347))
 
 ## 35.3.3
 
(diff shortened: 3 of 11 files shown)


app/assets/stylesheets/component_guide/application.scss
```diff
@@ -1,6 +1,43 @@
...
+// the components.
+
+// feature flag for accessible link styles
+$govuk-new-link-styles: true;
+
+@import "govuk_publishing_components/govuk_frontend_support";
+@import "govuk_publishing_components/component_support";
+
+// Include all govuk frontend components
+@import "govuk/components/all";
```
andysellick, line 13: Why do we need to include all govuk-frontend components? I tried removing this line and couldn't see any immediate problems, but that's not to say there weren't any.

app/assets/stylesheets/component_guide/application.scss
```diff
@@ -1,6 +1,43 @@
...
+
+// Include stylesheets used in static
```
andysellick, line 15: Why do we need stylesheets that are used in `static` specifically? I think it would be good to have the answer in the commit message for this file, to explain "required stylesheets".

app/assets/stylesheets/component_guide/application.scss
```diff
@@ -1,6 +1,43 @@
 // This file contains the styles for the Component Guide.
 
-@import "govuk_publishing_components/all_components";
+// This is the file that the application needs to include in order to use
+// the components.
```
andysellick, line 4: Is this comment necessary?

andysellick (commented):
Generally looks good, just a couple of comments. As I suggested in my previous review, it would be good to discuss/document what alternative approaches there are, as like you said repurposing the config variable feels a little strange.

app/assets/stylesheets/component_guide/application.scss
MartinJJones: No, I'll remove this now

app/assets/stylesheets/component_guide/application.scss
MartinJJones: Good spot, I included this as it is [currently included in `all_components`](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/_all_components.scss#L10-L11), after testing this change locally, I don't think this is required, will do some further testing and see if Percy picks up on any visual changes.

MartinJJones: After removing this import, [Percy shows a change in the `label` component](https://percy.io/55ba5e1a/govuk_publishing_components/builds/26892110/changed/1496462152?browser=firefox&browser_ids=39&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C1280), with the radio input circle no longer being displayed.

There appears to be an existing rendering issue in the component guide for [label component](https://components.publishing.service.gov.uk/component-guide/label/preview):

> Note that this example will not render correctly - do not be alarmed. It only works properly when inside the radio component.

I can at least keep the radio input circle by also ensuring the radio stylesheet is imported into application.css

app/views/govuk_publishing_components/component_guide/example.html.erb
```diff
@@ -1,6 +1,6 @@
 <% content_for :title, "#{@component_example.name} example - #{@component_doc.name} component" %>
 <% content_for :application_stylesheet do %>
-  <% if @component_doc.source == "application" %>
+  <% if @component_doc.source == "application" && GovukPublishingComponents::Config.application_stylesheet != false %>
```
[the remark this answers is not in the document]
andysellick: Those are good points, but I'm still uncertain on the best approach. Maybe worth asking some BE devs for their opinion?

app/assets/stylesheets/component_guide/application.scss
andysellick: Oh, is it because the component guide (unlike pages served from the application) isn't built using static, and therefore the styles from static aren't available? That makes sense, thanks. Maybe worth expanding the comment to explain that?

MartinJJones: Yes, although it is possible to make the styles from static available and not import them into application.css, some components do not render correctly using this approach.

This appears to be because we always expect stylesheets to be loaded in a certain order, those included in static, followed by those in the rendering app. This can cause rendering issues in apps that do not use this approach, an example of this is shown in the visual regression test for the `textarea` component in this PR.

The component-guide currently loads the stylesheets in the order below:

```SCSS
...
- character-count
- ...
- textarea
```

Applies the CSS below:

`.gem-c-textarea .govuk-textarea {margin-bottom: 0;}`

The rendering apps, load them in the order below:

```SCSS
- textarea
- ...
- character-count
```

Applies the CSS below:
`.govuk-character-count .govuk-textarea {margin-bottom: 5px;}`

Importing the `character-count` stylesheet after the `textarea` stylesheet ensures any overrides work as expected.

Overall there is more work to do in this area, I have created a separate card to fully implement the Per Page Asset Loading feature in the component guide, as part of this we will need to identify the correct order for stylesheets to be loaded - https://trello.com/c/Hhs2H5S7/2056-fully-implement-individual-loading-of-assets-in-the-component-guide.

I've updated the comment in the `application.css` to provide further information on why this approach is used.

andysellick (approved):

app/assets/stylesheets/component_guide/application.scss
andysellick: Good stuff. Alarming that our CSS is apparently so brittle! We should definitely look into fixing this at some point.

PR #3361, opened 2023-04-21
Add organisation brand colour for Department for Business & Trade

baisa:
[Trello card](https://trello.com/c/rtK4IWUx/1206-update-department-of-business-trade-branding-labels-in-publishing-components)
This change should be removed after relevant govuk-frontend release.

## What
We are updating the labelling of Department of International Trade to Department for Business & Trade.

## Why
This is part of the Supporting Government changes work (departments updates).

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
@@ -96,3 +96,27 @@
     border-color: #045f71;
   }
 }
+
+// This change should be removed after relevant govuk-frontend release
+
+.brand--department-for-business-and-trade {
+  .brand__color {
+    color: #cf102d;
+
+    &:link,
+    &:visited,
+    &:active {
+      color: #cf102d;
+    }
+
+    &:hover,
+    &:focus {
+      color: $govuk-focus-text-colour;
+    }
+  }
+
+  &.brand__border-color,
+  .brand__border-color {
+    border-color: #cf102d;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,7 @@
 
 ## 35.3.1
 
+* Add organisation brand colour for Department for Business & Trade ([#3361](https://github.com/alphagov/govuk_publishing_components/pull/3361))
 * Add brand colour for DSIT ([PR #3349](https://github.com/alphagov/govuk_publishing_components/pull/3349))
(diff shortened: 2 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
```diff
@@ -96,3 +96,27 @@
...
+
+    &:link,
+    &:visited,
+    &:active {
+      color: #005ea5;
+    }
+
+    &:hover,
+    &:focus {
+      color: #005ea5;
```
andysellick, line 114: I think this line needs to change to match the example on line 70. We only want the colour to be applied to the link when it is not focussed, because the focus state has a different background colour (yellow) and therefore the contrast is different - for example, if the brand colour was yellow, the link when focussed would be yellow on yellow, and therefore unreadable. This is what it looks like now:

Screenshot 2023-04-24 at 11 19 26

and this is what it should look like:

Screenshot 2023-04-24 at 11 22 28

app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
```diff
@@ -96,3 +96,27 @@
     border-color: #045f71;
   }
 }
+
+// This change should be removed after relevant govuk-frontend release
+
+.brand--department-for-business-and-trade {
+  .brand__color {
+    color: #005ea5;
```
andysellick, line 104: A small and probably not blocking comment - this colour doesn't pass WCAG AAA. It passes AA, which is good, but it would be better if it could also pass AAA. If it was adjusted to `#005A9E` it would pass AAA, but I appreciate that this may involve some back and forth.

Screenshot 2023-04-24 at 11 25 07

See https://webaim.org/resources/contrastchecker/

andysellick (changes requested):
Sorry, one small and easily fixed problem, plus a non-blocking comment to consider.

app/assets/stylesheets/govuk_publishing_components/components/helpers/_brand-colours.scss
baisa: @andysellick I have pushed the recent change as discussed - thank you!

andysellick (approved):

andysellick:
@baisa the failures in the visual regression tests should disappear if you rebase this branch against `main` - it uses the main branch as the baseline for testing, and I think another PR with some visual changes got merged ahead of this one.

PR #3362, opened 2023-04-24
[DO NOT MERGE] Create GA4 "change" event tracker for finder-frontend finders

AshGDS:
Hi @andysellick / @JamesCGDS - would you be able to review this when you get a chance? I've tried to explain it as best as I could, but let me know if something doesn't make sense. Thanks. 

## What
- Adds a new JavaScript "change" event tracker for the finders in `finder-frontend`. Related to this PR: https://github.com/alphagov/finder-frontend/pull/3042

Related trello cards: https://trello.com/c/zCIwJ8pP/515-add-tracking-search-box-on-search-finder-pages https://trello.com/c/aEXGK5cE/527-add-tracking-select-filter-on-search-finders-selectcontent
https://trello.com/c/sH9vKqA4/528-add-tracking-remove-filter-on-search-finders-selectcontent
https://trello.com/c/ZQeyh2qb/530-add-tracking-sort-by-search-finders-selectcontent

So this adds tracking to the search box, adding and removing the filters, changing the sort order, and clicking "clear all filters" on mobile.

## Why

- This is needed because the finders are updated with client side JavaScript. They fire a `change` event whenever a keyword/filter/sort option is changed. Therefore we need to hook into this to track what was changed for GA4.
- As each finder is one big form, it fires one big change event if anything in the search changes. Therefore we need a way to know which element triggered the change, so that we can send the correct GA4 event. In `finder-frontend`, we grab the `event.target` as well as `data-` attributes on the input or input parent, which gives us a simple way to categorise the change and build our GA4 object. We then pass that information through to this `Ga4FinderTracker` which processes it into a GTM object.

## How to test this
1. Run `static` locally, pointing to your local `govuk_publishing_components` with this branch, `ga4-finder-tracker`.
2. In `finder-frontend`, checkout the branch `ga4-finder-tracker`  https://github.com/alphagov/finder-frontend/pull/3042
3. Run `finder-frontend` pointing to your local `static`
4. Go to http://127.0.0.1:3062/search/all and change something in the search (e.g. a keyword, add a filter, remove a filter, or click "Clear all filters" in mobile view. Check the `dataLayer` for an `event_data` event (it's usually index 7 in the dataLayer). Also check the finders without any filters e.g. http://127.0.0.1:3062/government/people and ones with radio buttons https://www.gov.uk/official-documents.

## Brief explanation of how it works

1. In finder-frontend, we define some data-attributes on the finder `<form>` elements. 
    a. Firstly, we put a `data-ga4-filter-container` on the element that wraps all the filters. This will be needed to set `index_section` on each filter.
    b. Next, we add a `data-ga4-section` on each filter with the `data-ga4-filter-container` element. This is needed to identify where we need to add `data-ga4-index` for our `index_section`s. This key is also used to identify the `section` value to add to the GTM object. E.g. for a Topics filter, the attribute would be `data-ga4-section="Topics"`
   c. Next, on each element that can cause a `change` event on the `<form>` we add an attribute called `data-ga4-change-category` which gives us 1. metadata on what that element changes about the search results and 2. the HTML type that the element is. For example, on a `<select>` element that changes the filters, we would add `data-ga4-change-category="update-filter select"`. On the search box we would add `data-ga4-change-category="update-keyword text".

2. On the loading of a `finder`, we call `window.GOVUK.analyticsGa4.Ga4FinderTracker.setFilterIndexes()` in `finder-frontend`. This will grab the `data-ga4-filter-container`, look for any `data-ga4-section` elements inside it, and set an `index_section` on each section it finds.
3. In `finder-frontend`, when the search results are changed, we pass through the `event.target` and `data-ga4-change-category` to `Ga4FinderTracker.trackFormChange()`.
4. `trackFormChange()` will grab the "value" of the Filter that caused the change. For example the name of the checkbox, or the new keyword that was just searched for.
5. If it was a filter that caused the `change` event, then `trackFormChange()` will try and determine wether the filter that changed was a `Add filter` or `Remove filter` event.
6. It will then use a `switch` statement to build the GTM object based on which `data-ga4-change-category` was passed. 

I've also written a markdown documentation file. Hope how I've built this makes sense.

## Visual Changes

 
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
@@ -6,6 +6,17 @@
   GOVUK.analyticsGa4 = GOVUK.analyticsGa4 || {}
 
   GOVUK.analyticsGa4.Ga4FinderTracker = {
+
+    setFilterIndexes: function () {
+      var filterContainer = document.querySelector('[data-ga4-filter-container]')
+      var filterSections = filterContainer.querySelectorAll('[data-ga4-section]')
+
+      for (var i = 0; i < filterSections.length; i++) {
+        var section = filterSections[i]
+        section.setAttribute('data-ga4-index', JSON.stringify({ index_section: i + 1, index_section_count: filterSections.length }))
+      }
+    },
+
     trackChangeEvent: function (eventTarget, changeEventMetadata) {
       var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
 
@@ -27,46 +38,42 @@
         // The "value" we need for a checkbox is the label text that the user sees beside the checkbox.
         elementValue = document.querySelector("label[for='" + checkboxId + "']").textContent
 
-        if (eventTarget.checked === false) {
-          // Checkbox unchecked = filter was removed
-          console.log('Filter was removed because it was unchecked')
-          wasFilterRemoved = true
-        }
+        // If the checkbox unchecked is unchecked, the filter was removed.
+        wasFilterRemoved = !eventTarget.checked
       } else if (elementType === 'select') {
         // The value of a <select> is the value attribute of the selected <option>, which is a hyphenated key. We need to grab the human readable label instead for tracking.
         elementValue = eventTarget.querySelector("option[value='" + eventTarget.value + "']").textContent
         var defaultValue = eventTarget.querySelector('option:first-of-type').textContent
-        console.log('Default value:', defaultValue)
 
         if (elementValue === defaultValue) {
           // <select> elements being reverted to their first option (i.e. their default value) count as a "removed filter". (This will be used on the filter <select>s but not the sort by <select>, as you can't "remove" the sort by filter.)
-          console.log('Filter was removed because it matches a select default value')
           wasFilterRemoved = true
         }
       } else if (elementType === 'text') {
         elementValue = eventTarget.value
         if (elementValue === '') {
-          console.log('Filter was removed because the value is an empty string')
           // If our custom date filters are reset, they become an empty text box, so we count this as a "removed filter". This boolean won't be used for the keyword search box, as deleting the keyword isn't considered removing a filter.
           wasFilterRemoved = true
         }
       }
 
-      console.log('Change type is', changeType)
-
       switch (changeType) {
         case 'update-filter':
-          console.log('Was the filter removed?', wasFilterRemoved)
+          schema.event_data.event_name = 'select_content'
+
+          var section = eventTarget.closest('[data-ga4-section]')
+          schema.event_data.section = section ? section.getAttribute('data-ga4-section') : undefined
+
           if (wasFilterRemoved) {
-            console.log('TODO: Add GA4 schema for "filter removed" event with value', elementValue)
+            schema.event_data.action = 'remove'
           } else {
-            console.log('TODO: Add GA4 schema for "filter added event" with value', elementValue)
+            schema.event_data.action = elementType === 'text' ? 'search' : 'select'
+            schema.event_data.index = this.getSectionIndex(section)
           }
           break
 
         case 'update-keyword':
           schema.event_data.event_name = 'search'
-          schema.event_data.text = PIIRemover.stripPIIWithOverride(elementValue, true, true)
           schema.event_data.url = window.location.pathname
           schema.event_data.section = 'Search'
           schema.event_data.action = 'search'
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -0,0 +1,137 @@
...
+
+      var wasFilterRemoved = false
+
+      var section = eventTarget.closest('[data-ga4-section]')
+
+      var schema = new GOVUK.analyticsGa4.Schemas().eventSchema()
+      schema.event = 'event_data'
+      schema.event_data.type = 'finder'
+
+      if (elementType === 'checkbox') {
```
andysellick, line 43: This section might be cleaner split into multiple functions.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -0,0 +1,137 @@
...
+        elementValue = eventTarget.value
+        if (elementValue === '') {
+          // If our custom date filters are reset, they become an empty text box, so we count this as a "removed filter". This boolean won't be used for the keyword search box, as deleting the keyword isn't considered removing a filter.
+          wasFilterRemoved = true
+        }
+      }
+
+      schema.event_data.text = elementValue
+
+      switch (changeType) {
```
andysellick, line 81: Maybe clearer to declare `changeType` immediately prior to this line?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -0,0 +1,137 @@
...
+      for (var i = 0; i < filterSections.length; i++) {
+        var section = filterSections[i]
+        section.setAttribute('data-ga4-index', JSON.stringify({ index_section: i + 1, index_section_count: filterSections.length }))
+      }
+    },
+
+    trackChangeEvent: function (eventTarget, changeEventMetadata) {
+      var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
+
+      changeEventMetadata = changeEventMetadata.split(' ')
```
andysellick, line 28: Did you consider using more than one attribute to save some processing here? Or using a JSON string?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -0,0 +1,137 @@
...
+          break
+
+        case 'update-sort':
+          schema.event_data.event_name = 'select_content'
+          schema.event_data.action = 'sort'
+          schema.event_data.section = 'Sort by'
+          break
+
+        default:
+          break
```
andysellick, line 119: Not sure this needs a default if all it does is `break`?

andysellick (commented):
Have had a look through the code parts, seems mostly sensible although it's slightly disappointing how (necessarily) long this is. I'm left wondering whether moving some of this into `finder-frontend` would be more efficient in terms of asset size for the site overall.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
AshGDS: My thinking was it would be less bytes added to the HTML if they were in one string, otherwise you'd have to add another `data-ga4-something` to each input in the HTML. Should I split it out?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
AshGDS: Good spot, I think JS was complaining when I was building the switch statement so I added it while debugging!

AshGDS:
@andysellick I'm happy to move into `finder-frontend` - should I do that and close this PR? I can make all the changes you requested in this PR, except for making it part of `analyticsModules`.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -0,0 +1,137 @@
...
+        var index = sectionElement.getAttribute('data-ga4-index')
+        index = JSON.parse(index)
+        return index
+      } catch (e) {
+        console.error('GA4 configuration error: ' + e.message, window.location)
+      }
+    }
+  }
+
+  global.GOVUK = GOVUK
```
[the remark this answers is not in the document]
AshGDS: I think it would need to remain outside of analytics modules, since the track function needs to be called when the search results update, so having it as a self-contained module wouldn't give the finder-frontend search code access to the `trackFormChange` function at the right time.

andysellick:
> @andysellick I'm happy to move into `finder-frontend` - should I do that and close this PR? I can make all the changes you requested in this PR, except for making it part of `analyticsModules`.

Yes, I think that's sensible. Do you agree? I have less visibility of the code so if there's a reason to keep it here (e.g. code reuse) I'm happy to discuss.

AshGDS:
@andysellick it should be ready for review in `finder-frontend` now: https://github.com/alphagov/finder-frontend/pull/3042

I've made the requested changes.

PR #3365, opened 2023-04-27
Add missing background colour style for metadata block "inverse" option

jon-kirwan:
## What
Add missing background colour style for metadata block "inverse" option

## Why
Currently the style is missing and the inverse option has no effect.

### After


app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -23,16 +23,20 @@
...
-  a:active,
-  a:visited {
-    color: govuk-colour("white");
+  a {
+    @include govuk-link-style-inverse;
   }
 
-  a:focus {
-    color: govuk-colour("black");
+  .gem-c-metadata__list {
```
MartinJJones, line 34: I'm wondering if we should move `.gem-c-metadata__list` out of `.gem-c-metadata--inverse`, allowing it to be used in other contexts in the future.

Not sure how likely this is, what do you think?

jon-kirwan: Good point - I was thinking about that but, at least for now, we only need the spacing changes with inverse enabled since it allows for consistent spacing. I think if I try to move it outside of `gem-c-metadata--inverse` I might end up causing spacing issues (without the inverse option).

MartinJJones: Thanks, make sense, I'll approve the changes now 👍

MartinJJones (approved):

MartinJJones (commented):
Just need to rebase

PR #3373, opened 2023-05-03
Rename type breadcrumbs to breadcrumb

JamesCGDS:
## What
This PR makes a very small change to the GA4 event `type` parameter when tracking breadcrumbs and changes it to `breadcrumb` rather than `breadcrumbs`.

## Why
Requested by PA's

[Trello card](https://trello.com/c/Fwv2tyhJ/514-rename-breadcrumb)

## Visual Changes
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


JamesCGDS:
Hi @AshGDS @andysellick - FYI I'm not sure if this PR would warrant a changelog entry as it's so small so I've left it out (but happy to add one if you feel it's warranted).

andysellick (approved):
Yeah, seems like a small change but I think I'd still put a changelog entry in, just so we can keep track of when it's been deployed more easily.

PR #3374, opened 2023-05-03
Add auditing of application components to component auditing

andysellick:
## What
Update the component auditing to include details of any components found in applications. This uses the existing code and templates to provide a similar component file listing for components within applications.

## Why
We want to improve our auditing tools so that we can get a better view of what components exist, and whether there is any duplication or other problems that we should be addressing.

## Visual Changes
Shows components found in `collections`, for example:

Screenshot 2023-05-03 at 11 15 53

Also now includes all components in the main list of components, including which application and a control to filter by application:

Screenshot 2023-05-10 at 09 12 52


app/models/govuk_publishing_components/audit_applications.rb
```diff
@@ -55,6 +55,19 @@ def initialize(path, name)
             components: components_in_ruby,
           },
         ]
+
+        # applications might not all of these things for all components
```
MartinJJones, line 59: Small nitpick, should this read "# applications might not use all of these things for all components"

MartinJJones (approved):
Can't wait to start using the new component auditing features, looks very useful!

I've approved, looks like a rebase is required to resolve the conflict in the Changelog, I've also added one small nitpick, I still understand the comment, so non-blocking.

PR #3379, opened 2023-05-05
Fully implement loading of individual stylesheets in component guide

MartinJJones:
## What

Fully implement loading of individual stylesheets in component guide.

- Remove import of stylesheets used in static from application.scss
- Update asset_helper.rb to ensure all component stylesheets are requested when viewing the component guide
- Ensure stylesheets are requested in the expected order, for example, the `button` stylesheet should always be requested before the `feedback` stylesheet in the [feedback component](https://github.com/alphagov/govuk_publishing_components/pull/3379/files#diff-467712c9048ef82f631d6018dc019d21a911b8c5dedbfe37a512b84cea69045eR2)

## Why

The components in the gem should render correctly when using the individual loading of stylesheets feature in any application.

Without this change, applications would need to import all the stylesheets we import in static - https://github.com/alphagov/static/blob/main/app/assets/stylesheets/application.scss#L11-L30 into their application.css file, in order for components to render correctly.

Trello: https://trello.com/c/Hhs2H5S7/1993-fully-implement-individual-loading-of-assets-in-the-component-guide

## Visual changes

Percy is showing 1 visual change for the "Form checkboxes - Checkbox items with conditional reveal checked" component. This appears to be expected given the [note included in the documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/checkboxes.yml#L303):

> Note: ... No styling will be applied to the inserted content by the component.

### DIFF
--- a/app/views/govuk_publishing_components/components/_character_count.html.erb
+++ b/app/views/govuk_publishing_components/components/_character_count.html.erb
@@ -1,6 +1,4 @@
 <%
-  add_gem_component_stylesheet("character-count")
-
   id ||= "character-count-#{SecureRandom.hex(4)}"
   maxlength ||= nil
   maxwords ||= nil
@@ -24,3 +22,7 @@
     </div>
   <% end %>
 <% end %>
+<%
+  add_gem_component_stylesheet("error-message")
+  add_gem_component_stylesheet("character-count")
+%>
--- a/lib/govuk_publishing_components/app_helpers/asset_helper.rb
+++ b/lib/govuk_publishing_components/app_helpers/asset_helper.rb
@@ -79,11 +79,15 @@ module GovukPublishingComponents
     private
 
       def is_already_used?(component)
-        if GovukPublishingComponents::Config.exclude_css_from_static
+        if GovukPublishingComponents::Config.exclude_css_from_static && !viewing_component_guide?
           all_component_stylesheets_being_used.include?(component) || STATIC_STYLESHEET_LIST.include?(component)
         else
           all_component_stylesheets_being_used.include?(component)
         end
+      end
+
+      def viewing_component_guide?
+        request.path.include?("/component-guide")
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fully implement loading of individual stylesheets in component guide ([PR #3379](https://github.com/alphagov/govuk_publishing_components/pull/3379))
+
 ## 35.6.0
 
 * Remove GA4 index parsing code ([PR #3434](https://github.com/alphagov/govuk_publishing_components/pull/3434))
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -6,31 +6,6 @@ $govuk-new-link-styles: true;
 @import "govuk_publishing_components/govuk_frontend_support";
 @import "govuk_publishing_components/component_support";
 
-// Import the same stylesheets used in static
-// https://github.com/alphagov/static/blob/198a598682df40ce4a2c3c286c06244297c18cf0/app/assets/stylesheets/application.scss
-// Although the component guide does not use static, we still need to import the same stylesheets used in static
-// to avoid any rendering issues
-// By following the same approach as frontend rendering applications,
-// we ensure that stylesheets are loaded in the expected order and components render correctly
-@import "govuk_publishing_components/components/breadcrumbs";
-@import "govuk_publishing_components/components/button";
-@import "govuk_publishing_components/components/error-message";
-@import "govuk_publishing_components/components/heading";
-@import "govuk_publishing_components/components/hint";
-@import "govuk_publishing_components/components/input";
-@import "govuk_publishing_components/components/label";
-@import "govuk_publishing_components/components/search";
-@import "govuk_publishing_components/components/skip-link";
-@import "govuk_publishing_components/components/textarea";
-@import "govuk_publishing_components/components/title";
-
-@import "govuk_publishing_components/components/cookie-banner";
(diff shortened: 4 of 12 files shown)


lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -79,12 +79,22 @@ def application_stylesheet_in_use?
...
+
+      def in_component_guide_path?
+        request.path.include?("/component-guide")
+      end
+
+      def viewing_component_guide?
+        return @viewing_component_guide if defined?(@viewing_component_guide)
+
+        @viewing_component_guide ||= in_component_guide_path?
+      end
```
KludgeKML, line 97: Is `@viewing_component_guide` likely to be defined outside this method? (ie is there a reason we're memoizing the value in this slightly non-idiomatic way)

MartinJJones: The main reason for using this approach was from following this blog post - https://blog.appsignal.com/2022/12/20/a-guide-to-memoization-in-ruby.html#ignoring-the-false-or-nil-return

In particular:

> Memoization using ||= doesn't consider false/nil return values, so such cases should be handled.

I also observed this when testing the changes in the `frontend` rendering app. In this case, `in_component_guide_path?` returns false as expected, but this method would be called several times, each returning a false value.

There could (is likely) a better way to do this, and one that makes sense to others reading the code.

KludgeKML: Actually, I think we might consider not memoizing at all? request.path.include? is just doing a string call, it probably doesn't need to be incredibly performant. It's not as necessary as memoizing a database or network call, or a complex calculation.

MartinJJones: Thanks Keith, good point. I've updated the code now and removed the memoization.

KludgeKML (approved):
Looks good to me!

MartinJJones:
@andysellick this PR will fully implement the individual loading of stylesheets in the component guide.

One thing I wanted to bring to your attention is the visual change for "Form checkboxes - Checkbox items with conditional reveal checked", I've added further detail regarding this in the PR description. Just wanted to check if this is OK?

jon-kirwan (approved):
All seems ok to me. (I've compared viewing the government frontend component guide and can see all component stylesheets, including those from Static, are correctly loaded).
Other changes look fine to me too 🚀

jon-kirwan (approved):
All seems ok to me. (I've compared viewing the government frontend component guide and can see all component stylesheets, including those from Static, are correctly loaded).
Other changes look fine to me too 🚀

jon-kirwan (approved):
All seems ok to me. (I've compared viewing the government frontend component guide and can see all component stylesheets, including those from Static, are correctly loaded).
Other changes look fine to me too 🚀

andysellick:
> One thing I wanted to bring to your attention is the visual change for "Form checkboxes - Checkbox items with conditional reveal checked", I've added further detail regarding this in the PR description. Just wanted to check if this is OK?

@MartinJJones that visual change makes sense, yes.

app/views/govuk_publishing_components/components/_character_count.html.erb
```diff
@@ -24,3 +22,7 @@
     </div>
   <% end %>
 <% end %>
+<%
+  add_gem_component_stylesheet("error-message")
+  add_gem_component_stylesheet("character-count")
```
andysellick, line 27: Don't these normally go at the top? What's the reason for having them at the bottom?

andysellick (commented):
Looks generally good to me, sensible approach on including stylesheets for components inside components. Looking at the component auditing there seems to be quite a few more components that contain components than updated in this PR - are you aware of them or have they already been done?

MartinJJones:
> Looks generally good to me, sensible approach on including stylesheets for components inside components. Looking at the component auditing there seems to be quite a few more components that contain components than updated in this PR - are you aware of them or have they already been done?

Good point, this approach of requesting component stylesheets within other components was only required where Percy was showing an issue when individual stylesheets for all components in the component guide. This always happened when stylesheets were not loaded in the expected order, in most cases this is not an issue and no action is required.

app/views/govuk_publishing_components/components/_character_count.html.erb
MartinJJones: This example is a little more complex than the feedback component given in the PR description.

The `character-count` component also renders the [textarea component](https://github.com/alphagov/govuk_publishing_components/pull/3379/files#diff-5ca2449113fd07595d5a64fffc87b170dde4f2feb677f8cfb80cb890ab02e844R18)

By including the `character-count` component stylesheet at the top of the page, we would load the stylesheets in the order below:

- character-count
- textarea

The textarea for margin-bottom is then set to 0, from textarea.scss

```CSS
.gem-c-textarea .govuk-textarea {
    margin-bottom: 0;
}
```

Percy highlighted this visual change, which was fixed by moving the `character-count` component stylesheet to the bottom.

This would then apply the CSS below, and margin-bottom is set to as required.

```CSS
.govuk-character-count .govuk-form-group,
.govuk-character-count .govuk-textarea {
    margin-bottom: 5px;
}
```

This was a similar issue for the error-message stylesheet and other components in the PR where other stylesheets are imported.

We normally don't have to worry about this in our applications as the styles for textarea are included in static and always imported first - https://github.com/alphagov/static/blob/main/app/assets/stylesheets/application.scss#L20.

This would be an issue for other apps using the gem that do not use static.

andysellick: Oooh complex! Good explanation, thanks. Is this something that can be fixed (or even a problem that needs solving?) - not here, but maybe worth raising an issue about it?

MartinJJones: I think this would only be an issue if you are using the gem in an app that does not use static and not using the individual loading of stylesheets feature. I will raise an issue for this :)

PR #3380, opened 2023-05-05
Add option for blue bar background to public_layout

patrickpatrickpatrick:
## What

Add a new option `blue_bar_background` for public_layout. Add new option `blue_bar` for public_layout. Rework logic for blue bar appearance in the public_layout. Add tests for the blue bar background and the blue bar logic.

## Why

There are [instances of pages on collections](https://www.gov.uk/browse) that have a coloured background in the heading section. This meant that the blue bar (which is rendered in the public_layout) had to be rendered by collections otherwise it would result in a gap between the blue bar and coloured section. This solution seemed fine until we had a recent deployment of the global bar. As the global bar includes a blue bar, it meant that the Browse pages had two blue bars. They therefore needed to be removed while the global bar was present and then added back again when the global bar was removed. 

To avoid this needing to happen the next time the bar is deployed, changes have been made to the `public_layout` component. It means that if a page has a coloured background heading it can be specified using the attributes and then it will be applied to the blue bar. This means the blue bar can always be rendered in public_layout and so applications don't need to render the blue bar if they have a heading section with a background colour.

[Relevant Trello Card](https://trello.com/c/pT7Sh8Yn/1793-update-layoutforpublic-template-so-that-global-bar-is-independent-of-requiring-changes-in-the-browse-template-l)

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -3,9 +3,11 @@
 
   emergency_banner ||= nil
   full_width ||= false
+  blue_bar ||= local_assigns.include?(:blue_bar) ? local_assigns[:blue_bar] : !full_width
   global_bar ||= nil
   html_lang ||= "en"
   layout_helper = GovukPublishingComponents::Presenters::PublicLayoutHelper.new(local_assigns)
+  blue_bar_background_colour = layout_helper.blue_bar_background_colours.include?(blue_bar_background_colour) ? blue_bar_background_colour : nil
   logo_link ||= "/"
   navigation_items ||= []
   omit_feedback_form ||= false
@@ -44,6 +46,9 @@
   blue_bar_dedupe = !full_width && global_bar.present?
   body_css_classes = %w(gem-c-layout-for-public govuk-template__body)
   body_css_classes << "draft" if draft_watermark
+
+  blue_bar_wrapper_classes = %w()
+  blue_bar_wrapper_classes << "gem-c-layout-for-public__blue-bar-wrapper--#{blue_bar_background_colour}" if blue_bar_background_colour
 -%>
 <!DOCTYPE html>
   <!--[if lt IE 9]><html class="lte-ie8 govuk-template" lang="<%= html_lang %>"><![endif]-->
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -1,8 +1,11 @@
 <%
-  add_gem_component_stylesheet("layout-for-public")
+  blue_bar_background_colours = ["browse"].freeze
 
+  add_gem_component_stylesheet("layout-for-public")
+  blue_bar_background_colour ||= blue_bar_background_colours.include?(blue_bar_background_colour) ? blue_bar_background_colour : nil
```
chao-xian, line 5: I don't understand what's going on here... sorry

patrickpatrickpatrick: This line checks to see if the background colour specified is in the list of blue_bar_background_colours. If it isn't then it sets blue_bar_background_colour to nil otherwise it sets it to the background colour specified.

chao-xian: OK I think the naming is _confusing_ so can we rename them pls? One as allowed colours maybe? The other as specified colours? Trying to spot one letter difference isn't user friendly

patrickpatrickpatrick: Ah actually since it's a constant, the naming convention is that it should be in all caps. Have changed that.

app/views/govuk_publishing_components/components/docs/layout_for_public.yml
```diff
@@ -27,6 +27,12 @@ examples:
     description: This allows the header to be omitted which is currently used when rendering CSV previews from Whitehall
     data:
       omit_header: true
+  blue_bar_background:
+    description: For use when a page has a heading component with a background colour.
+    data:
+      full_width: true
+      blue_bar: true
+      blue_bar_background: "explore"
```
MartinJJones, line 35: There does not appear to be an input for `blue_bar_background`, should this be `blue_bar_background_colour`?

If so, does the string also need to be changed from "explore" to "browse" so the [grey background color](https://github.com/alphagov/govuk_publishing_components/pull/3380/files#diff-841f2eaa57914d8e9aabb60cae32faa0fb1d09fcfd485e89e60f4530d73d1f92R12) is applied?

MartinJJones (commented):
The changes look good to me overall 👍

I've added some comments for things to look at, happy to help further on these if needed.

MartinJJones (approved):
Changes looked good to me 👍, I've also approved the 1 visual regression test showing in Percy.

MartinJJones (approved):

PR #3382, opened 2023-05-09
Change type 'html attachment' to just 'attachment'

AshGDS:
https://trello.com/c/IsRPgFh9/559-change-html-attachment-type-to-attachment-on-attachment-navigation-clicks

## What

Change type 'html attachment' to just 'attachment'

## Why

This change is being made because we're going to add data-ga4-attachment-link to every attachment link, so soon it won't just be html attachments being tracked with this type.

## Visual Changes

None.


JamesCGDS (approved):
Looks good to me 👍

PR #3383, opened 2023-05-09
Bump version to 35.3.4

baisa:
Releases version 35.3.4 of the gem.


andysellick (approved):
It's possible this should be a minor release since the organisation logo I think counts as a change to the API, but it's probably fine.

baisa:
> It's possible this should be a minor release since the organisation logo I think counts as a change to the API, but it's probably fine.

Thank you @andysellick Will merge it now 👍

PR #3390, opened 2023-05-11
Conditionally set GA4 type in related navigation

JamesCGDS:
## What
This PR addresses a bug where the `index` parameters were not aggregating when a `contextual sidebar` and a `contextual footer` were rendered together on a page. Combining the index properties of 2 different components could be complex and so the simplest fix was to provide a differentiation between `contextual sidebar` and `contextual footer` components in the `type` parameter so that the `index` properties need not be aggregated.

## Why
Requested by PA's.

[Trello card](https://trello.com/c/dYtGpIpv/502-indexcount-is-different-for-related-content-and-contextual-footer)

## Visual Changes
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
lgtm :+1:

PR #3396, opened 2023-05-11
[Fixes] World Locations Links From Base Path

minhngocd:
## What
Navigation helper: Within related navigation links, world Locations paths are now generated based on the base path if it is provided in the content item. Not a breaking change since if there is no base path provided, it will fall to previous behaviour of generating the slug based on page title. The special cases list has been removed since it is out of date.

## Why
Making it more flexible to deal with cases in related navigation links where there are pages where the name does not match the slug. e.g. "UK and NATO" / "uk-joint-delegation-to-nato".
The special cases list was out of date, with the new update now we should just default to using the base path if it is already being provided. This should fix some broken links

## Visual Changes
No visual changes.
No breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@
 * Update to LUX 308 ([PR #3394](https://github.com/alphagov/govuk_publishing_components/pull/3394))
 * Ensure PIIRemover is running on GA4 link clicks ([PR #3402](https://github.com/alphagov/govuk_publishing_components/pull/3402))
 * Update hint component ([PR #3405](https://github.com/alphagov/govuk_publishing_components/pull/3405))
+* [Fixes] World Locations Links From Base Path ([PR #3396](https://github.com/alphagov/govuk_publishing_components/pull/3396))
 
 ## 35.3.5
 
--- a/lib/govuk_publishing_components/presenters/related_navigation_helper.rb
+++ b/lib/govuk_publishing_components/presenters/related_navigation_helper.rb
@@ -12,9 +12,6 @@ module GovukPublishingComponents
         world_locations
         statistical_data_sets
       ].freeze
-      WORLD_LOCATION_SPECIAL_CASES = {
-        "UK Mission to the European Union" => "uk-mission-to-the-eu",
-      }.freeze
 
       attr_reader :related_navigation, :index_section_count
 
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -10,6 +10,8 @@
 ## Unreleased
 
 * Update to LUX 308 ([PR #3394](https://github.com/alphagov/govuk_publishing_components/pull/3394))
+* World Locations paths are now generated based on the base path if it is provided. If there is no base path provided, it will fall to previous behaviour of generating the slug based on page title. The special cases list has been removed.
```
dnkrj, line 13: I think that this should be reframed from the perspective of the component - related navigation - both here and an in the PR

spec/dummy/app/assets/config/manifest.js
```diff
@@ -2,6 +2,7 @@
 //= link_tree ../images
 //= link_directory ../javascripts .js
 //= link_directory ../stylesheets .css
+//= link application.css
```
dnkrj, line 5: Is this intentional? / Where does this come from?

dnkrj (commented):
I think that this looks good, just needs to be reframed a bit I think.
Is this going to be a breaking change for any applications?
Maybe reference who uses this in the PR description.

spec/dummy/app/assets/config/manifest.js
minhngocd: The unit tests don't run locally for me unless I put this in. 
It's under spec folder for a dummy app so should not be impacting any production code

CHANGELOG.md
minhngocd: Updated ! let me know what you think

dnkrj (approved):
LGTM 👻

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Add auditing of application components to component auditing ([PR #3374](https://github.com/alphagov/govuk_publishing_components/pull/3374))
 * Fix GA4 index parameters on step nav show/hide all control ([PR #3397](https://github.com/alphagov/govuk_publishing_components/pull/3397))
 * Update to LUX 308 ([PR #3394](https://github.com/alphagov/govuk_publishing_components/pull/3394))
+* [Fixes] Navigation helper: Within related navigation links, world Locations paths are now generated based on the base path if it is provided in the content item. Not a breaking change since if there is no base path provided, it will fall to previous behaviour of generating the slug based on page title. The special cases list has been removed since it is out of date. ([PR #3396](https://github.com/alphagov/govuk_publishing_components/pull/3396))
```
andysellick, line 15: Can this match the PR name please?

PR #3397, opened 2023-05-12
Fix GA4 index on step nav

andysellick:
## What
Fixes the index parameters on the step by step navigation 'show/hide all' element. They were set to e.g.

```
index: 0, index_section_count: 8
```

but should be

```
index: { index_section: 0, index_section_count: 8 }
```

## Why
Part of GA4 bug fixing.

## Visual Changes
None.

Trello card: https://trello.com/c/6hEwEVkq/526-fix-indextotal-parameters


gclssvglx (approved):
Great stuff @andysellick

PR #3405, opened 2023-05-17
Update hint component

MartinJJones:
## What

- Replace hardcoded hint component by using the render method, instead of hardcoding it in the label component view template
- Add a new `is_radio_label_hint` input, setting this to `true` will add the `govuk-radios__hint` class to apply the correct styling when the label is used on a radio input
- margin-bottom will now only be applied if it is passed as an input, it will not default to margin-bottom-3, instead it will default to the margin used in the design system

## Why

- To ensure that any future changes to the hint component are reflected when used in the `label` component view template
- Allow the `hint` component stylesheet to be included when used within the `label` component view template, this happens automatically when using the render method

Trello: https://trello.com/c/Hhs2H5S7/1993-fully-implement-individual-loading-of-assets-in-the-component-guide

## Visual Changes

Percy is showing several visual changes, these are all related to how `margin-bottom` is now applied. Previously, the shared_helper would default to `margin-bottom-3` (15px), regardless of context.

The styling now matches the margin-bottom set in the Design System, for this reason I believe the visual changes should be approved. You can still pass in a margin-bottom value if you want to override margin-bottom from the design system.

I have included a before and after image for the [input component when used with a hint](https://components.publishing.service.gov.uk/component-guide/input/with_hint/preview).

### Input with hint - Before

CSS applied
```CSS
.govuk-\!-margin-bottom-3 {
    margin-bottom: 15px !important;
}
```

Screenshot 2023-05-17 at 10 26 19

### Input with hint - After

CSS applied

```CSS
.govuk-label:not(.govuk-label--m):not(.govuk-label--l):not(.govuk-label--xl) + .govuk-hint {
    margin-bottom: 10px;
}
```

Screenshot 2023-05-17 at 11 45 04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix GA4 index parameters on step nav show/hide all control ([PR #3397](https://github.com/alphagov/govuk_publishing_components/pull/3397))
 * Update to LUX 308 ([PR #3394](https://github.com/alphagov/govuk_publishing_components/pull/3394))
 * Ensure PIIRemover is running on GA4 link clicks ([PR #3402](https://github.com/alphagov/govuk_publishing_components/pull/3402))
+* Update hint component ([PR #3405](https://github.com/alphagov/govuk_publishing_components/pull/3405))
 
 ## 35.3.5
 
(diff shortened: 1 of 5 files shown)


AshGDS (approved):
Looks good to me :+1:

PR #3413, opened 2023-05-23
Remove list-style for govspeak ordered lists

kevindew:
This removes setting list-style on ordered lists to instead allow the
use of the type attribute on an ordered list to specify the list-style.

I couldn't determine why this style was needed. Current browsers style
this as list-style decimal by default and I can't find any code this
undoes. It could be that previously this was to undo an earlier
override, but it would be very unconventional for current GOV.UK to
style an `ol` without a class so this seems unlikely.

The motivation for making this change is to support some niche behaviour
used by publishers that will become disabled. Currently some advanced
published have managed to published ordered lists that are alphabetical
or roman numeral prefixed by use of an `ol` element with a style
attribute. This will not be available with the role out of the GOV.UK
Content Security Policy which forbids unsafe inline styles [1].

[1]: https://github.com/alphagov/govuk_app_config/pull/291

### Before
Screenshot 2023-05-23 at 17 50 59

### After
Screenshot 2023-05-23 at 17 50 38

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove hardcoded examples in the component guide ([3418](https://github.com/alphagov/govuk_publishing_components/pull/3418))
+* Remove list-style for govspeak ordered lists ([PR #3413](https://github.com/alphagov/govuk_publishing_components/pull/3413))
 
 ## 35.5.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
@@ -102,7 +102,6 @@
 
   ul {
     list-style: disc;
-    list-style-position: outside;
   }
 
   li {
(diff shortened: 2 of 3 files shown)


MartinJJones:
The changes look good to me.

I was reading further into the use of the type attribute for ordered lists out of interest on MDN - https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ol#type, which advises:

> Note: Unless the type of the list number matters (like legal or technical documents where items are referenced by their number/letter), use the CSS [list-style-type](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type) property instead.

and also in the W3C spec - https://www.w3.org/TR/2012/WD-html-markup-20120329/ol.html

> Note: The type attribute on the ol element was deprecated in a previous version of HTML, but is no longer deprecated, as it has meaning and is not simply presentational.

I imagine the type does matter for the documents publishers have used this inline style approach on previously. Either way, looks like an improvement to me 👍

MartinJJones (approved):
Changes look good to me 👍

kevindew:
Thanks @MartinJJones, yup I read the same spec and figured this was appropriate usage of the attribute as it used to add semantics to the copy.

PR #3423, opened 2023-05-31
Component auditing improvements

andysellick:
## What
Couple of new bits on the component audits, specifically:

- stop using the tag component on the component details list, not appropriate
- add JS generated numbers to indicate totals on the component details list

## Why
Improve our component auditing.

## Visual Changes
New column totals on the component details. These are updated dynamically if the selection is changed.

Screenshot 2023-05-31 at 16 20 57


jon-kirwan (approved):
Nice work. A handy addition and all works as expected.

The only comment I have, perhaps for future changes, is whether we should consider using `forEach` ([IE 9,10,11](https://caniuse.com/mdn-javascript_builtins_array_foreach) up) rather than `for` loops when iterating over nested items - only because it might be easier to read? Something a bit like:

```
this.headingLabels.forEach(function (heading, i) {
  headingCounts[i] = 0
})

visibleRows.forEach(function (row) {
  this.headingLabels.forEach(function (heading, i) {
    var cell = row.querySelector('[data-component-type=' + heading + ']')

    if (cell.textContent.trim().length) {
      headingCounts[i] += 1
    }
  })
}, this)
```

andysellick:
Thanks for the review @jon-kirwan! Great shout on the `forEach` loops, didn't realise they were supported as far back as IE9 - I seem to always instinctively write JS for the lowest common denominator...

PR #3426, opened 2023-06-02
Fix some issues with GA4 indexes

AshGDS:
## What
- Sets any unused index keys in the `index` object to `undefined` if they aren't being used in the `index: {}` object.
- Updates the GOVUK Logo index values

## Why

- The PAs noticed that sometimes index values were being persisted between dataLayer pushes. This was because we aren't setting the indexes to `undefined` if they don't exist. Therefore, an `index_link`, `index_section` or `index_section_count` from a previous event could be included with the current event.
- Updates the GOVUK Logo index in accordance with this card: https://trello.com/c/SRFZ68d4/566-fix-index-on-govuk-link-click-nav-event-in-header-menu-bar

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -30,7 +30,32 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
       firstScript.parentNode.insertBefore(newScript, firstScript)
     },
 
+    ensureIndexesArePopulated: function (data) {
+      if (!data.event_data) {
+        return data
+      }
+
+      if (!data.event_data.index) {
+        return data
+      }
+
+      var indexKeys = ['index_link', 'index_section', 'index_section_count']
+
+      for (var i = 0; i < indexKeys.length; i++) {
+        var indexKey = indexKeys[i]
+
+        // If the index key isn't in the object, populate it. However if it's set to 0, leave it as 0. 0 is falsy so we have to add this extra check.
+        if (!data.event_data.index[indexKey] && data.event_data.index[indexKey] !== 0) {
+          data.event_data.index[indexKey] = undefined
+        }
+      }
+
+      return data
+    },
+
     sendData: function (data) {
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -30,7 +30,32 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
...
     },
 
+    ensureIndexesArePopulated: function (data) {
+      if (!data.event_data) {
+        return data
+      }
+
+      if (!data.event_data.index) {
+        return data
+      }
```
andysellick, line 40: Any chance this if and the previous if could be combined? I suspect not but wanted to check.

andysellick (commented):
Wow this is really interesting - @JamesCGDS and I had a conversation about this recently where we assumed that this problem wouldn't happen - because now the `index` parameter is  sub parameters it would completely override the previous value - but apparently not!

Code looks sensible to me but I'm just holding off approving as I think it might be worth a quick chat about this to check we fully understand the implications of this problem.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: It can't as if `event_data` is `undefined`, and you try and call `.index` on it, you will get an error essentially saying `can't call .index on something that's undefined`.

AshGDS:
Thanks @andysellick - yeah, unfortunately it doesn't seem to work like that. I think if it was like how you described then we wouldn't have to worry about overriding values at all, as our `event_data` object would just override everything each time!

andysellick:
> Thanks @andysellick - yeah, unfortunately it doesn't seem to work like that. I think if it was like how you described then we wouldn't have to worry about overriding values at all, as our `event_data` object would just override everything each time!

I guess the question now is do we need to do this for the `event_data` object as well? Or just do some kind of empty push like we do for the ecommerce stuff?

andysellick (approved):

PR #3430, opened 2023-06-04
Fix PII on referrer parameter

JamesCGDS:
## What
When conducting a search on a finder application (e.g. https://www.gov.uk/search/all?), the `referrer` parameter on the `pageView` object does not redact PII. (Instructions to recreate the issue have been left on this [trello card](https://trello.com/c/hs46dfAY/591-pii-coming-through-in-referrer-property-on-pageview-object))

## Why
PII issue fix.

[Trello card](https://trello.com/c/hs46dfAY/591-pii-coming-through-in-referrer-property-on-pageview-object)

## Visual Changes
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Change GA4 share values ([PR #3407](https://github.com/alphagov/govuk_publishing_components/pull/3407))
 * Add GA4 index_section_count to step by step links ([PR #3410](https://github.com/alphagov/govuk_publishing_components/pull/3410))
 * Set GA4 link text to 'image' if there's only an image and no link text ([PR #3404](https://github.com/alphagov/govuk_publishing_components/pull/3404))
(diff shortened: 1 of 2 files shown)


JamesCGDS:
Hi @AshGDS @andysellick, this is a quick fix and I'm sure there's a better way to do this. I was thinking of passing through the `referrer` parameter to the `getReferrer` function and the just setting a variable to either the `referrer` parameter if it existed or `document.referrer` and then just calling `this.PIIRemover.stripPIIWithOverride` once instead of twice as I've done here. I was getting some errors though so I pushed up the simplest fix. Hope I've made sense here, happy to discuss.

andysellick (approved):
Looks good, is it worth squashing these two commits together?

andysellick:
Also I think the visual regression test failures can be resolved by rebasing against `main`.

PR #3434, opened 2023-06-06
Remove GA4 index parsing code

andysellick:
## What
This code was used to convert string values for the GA4 index parameter into sub attributes such as index_section, index_link. It was originally written to be future proof i.e. if passed an object that already contained these sub attributes, it just returned it. At this point nothing is using it, so it's redundant.

## Why
This code is no longer needed as all instances of code that uses an index parameter now has it as sub attributes already.

## Visual Changes
None.

Trello card: https://trello.com/c/TDGHW9cX/438-remove-index-processing-code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove GA4 index parsing code ([PR #3434](https://github.com/alphagov/govuk_publishing_components/pull/3434))
 * Fix GA4 tracking on step nav related ([PR #3431](https://github.com/alphagov/govuk_publishing_components/pull/3431))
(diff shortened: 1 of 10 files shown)


JamesCGDS:
Looks good to me 👍 Spotted a small thing and it's not part of this PR but in `ga4-link-tracker.md`, we are still referring to `index` as a string/number (e.g. `"index: "1"`). As part of this PR, would we want to change those instances to `"index": { "index_link": 1 }`?

andysellick:
@JamesCGDS thanks for the review, have updated the doc as per your suggestion, does that look okay?

JamesCGDS (approved):
Looks good 👍

PR #3435, opened 2023-06-06
Update setIndexes function

JamesCGDS:
## What
This PR adds a condition to the `setIndexes` function where links with a `data-ga4-do-not-index` attribute are ignored/not indexed.

## Why
This change was made because there are occasions where links generated by content publishers (which need indexing via JS as we can't amend the HTML directly) appear in the same module as links that are accessible in a template but we don't want to index them (e.g. `change_response` and `form_start_again` links as per the Implementation Guide).

For example, on [this page ](https://www.gov.uk/food-business-registration/westminster), the links that appear within the results body (e.g. [starting a food business](http://www.food.gov.uk/business-industry/caterers/startingup/)) are generated by govspeak however the `change location` link is accessible in the template and is a link that we don't want to index.

Since the `setIndexes` function works on all links within a container with a `data-ga4-set-indexes` attribute, the `change location` link was being indexed and interfering with the calculation of the remaining indexes (e.g. including the `change location` link in the `index_total` when it shouldn't). By adding a `data-ga4-do-not-index` attribute to the `change location` link, we are able to flag the link as one to ignore by the `setIndexes` function.

## Visual Changes
N/A


JamesCGDS:
Hi @AshGDS @andysellick, I've tried to explain this issue as best I could in the PR description but I appreciate it could be quite confusing - if demonstrating the problem over a call would be better just give me a shout 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -256,8 +256,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
         var totalLinks = 0
         for (var i = 0; i < links.length; i++) {
           var link = links[i]
-          // Only index links that are not search results
-          if (!link.getAttribute('data-ga4-ecommerce-path')) {
+          // Only index links that are not search results or do not have a data-ga4-do-not-index attribute
+          if (link.getAttribute('data-ga4-ecommerce-path') || link.getAttribute('data-ga4-do-not-index') !== null) {
+            continue
+          } else {
```
AshGDS, line 262: Don't have to make this change if you don't feel it's necessary, but I believe the `else {` is redundant here and you could just have the code standalone, after the if statement.

```JavaScript
if (link.getAttribute('data-ga4-ecommerce-path') || 
  link.getAttribute('data-ga4-do-not-index') !== null) {
  continue
}
  
totalLinks++
link.setAttribute('data-ga4-index', '{"index_link": ' + totalLinks + '}')
  
```

docs/analytics-ga4/ga4-link-tracker.md
```diff
@@ -69,7 +69,11 @@ The `data-ga4-track-links-only` attribute ensures that only link clicks are trac
 </div>
 ```
 
-Note: the addition of the `data-ga4-set-indexes` data attribute on the parent element will allow the `index` and `index_total` properties to be set.
+Note: the addition of the `data-ga4-set-indexes` data attribute on the parent element will allow the `index` and `index_total` properties to be set. If links that we can amend directly appear alongside links generated by content items or publishing tools and we do not want to index the former, adding a `data-ga4-do-not-index` attribute will prevent those links from being indexed. For example:
+
+```html
+<a href="/link1" data-ga4-do-not-index>This link will not be indexed</a>
```
AshGDS, line 75: I think it's worth putting this example within a `<div data-ga4-set-indexes>` element, and including a link that would get indexed, to make it clearer.

AshGDS (commented):
Looks good, have some minor comments but can approve if you disagree with them.

docs/analytics-ga4/ga4-link-tracker.md
JamesCGDS: Good point, will do 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
andysellick: Wouldn't having the code outside of the `if` statement make it always run, regardless of the outcome of the `if`? I think you actually want

```Javascript
if (link.getAttribute('data-ga4-ecommerce-path') || link.getAttribute('data-ga4-do-not-index') !== null) { 
   totalLinks++ 
   link.setAttribute('data-ga4-index', '{"index_link": ' + totalLinks + '}')
}
```

andysellick: Ah wait no I've misunderstood what `continue` does, ignore me.

andysellick (approved):
All looks good to me (although I'd squash the commits together)

andysellick (approved):
All looks good to me (although I'd squash the commits together)

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: We want the link to not have both `data-ga4-ecommerce-path` and `data-ga4-do-not-index` on it in order for an index to be set on the link. 

I might be overthinking it, but if you tried:
```JavaScript
if (!link.getAttribute('data-ga4-ecommerce-path') || !link.getAttribute('data-ga4-do-not-index')) { 
   totalLinks++ 
   link.setAttribute('data-ga4-index', '{"index_link": ' + totalLinks + '}')
}
``` 

the issue is that it would still set the index if one attribute existed. For example, if `data-ga4-do-not-index` existed but `data-ga4-ecommerce-path` didn't, it would still set the index. 

And if you did 
```JavaScript
if (!link.getAttribute('data-ga4-ecommerce-path') && link.getAttribute('data-ga4-do-not-index' === null) {
   totalLinks++ 
   link.setAttribute('data-ga4-index', '{"index_link": ' + totalLinks + '}')
}
```
 then the link would need both attributes to not exist in order for an index to be set on it. This makes sense, but then if the link has `'data-ga4-do-not-index` only, it would still set the index since `data-ga4-ecommerce-path` would also need to be on the link in order to stop the index being set.

This is why I thought the method of using `continue` made sense, since it skips the link if one of the attributes exists, instead of trying to do stuff based on things existing or not existing which was getting complicated!

Please correct me if i'm wrong!

docs/analytics-ga4/ga4-link-tracker.md
```diff
@@ -72,7 +72,9 @@ The `data-ga4-track-links-only` attribute ensures that only link clicks are trac
 Note: the addition of the `data-ga4-set-indexes` data attribute on the parent element will allow the `index` and `index_total` properties to be set. If links that we can amend directly appear alongside links generated by content items or publishing tools and we do not want to index the former, adding a `data-ga4-do-not-index` attribute will prevent those links from being indexed. For example:
 
 ```html
-<a href="/link1" data-ga4-do-not-index>This link will not be indexed</a>
+<div data-ga4-set-indexes>
+  <a href="/link1" data-ga4-do-not-index>This link will not be indexed</a>
```
AshGDS, line 76: Nitpick: Could you put a regular link as well e.g. `<a href="/link2" data-ga4-index="{index: {index_link: 1}}" >This link got indexed</a>`

AshGDS (approved):
All good, left a minor nitpick comment

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: Ah, just saw you mentioned you misunderstood - ignore me as well 😃

PR #3436, opened 2023-06-07
Change how GA4 schema construction works

andysellick:
## What
Currently we pass nested data structures from markup into the analytics code, where it is checked against a schema and then sent to GTM. For example:

```
<div data-ga4-attribute="{ index: { index_link: 1, index_section: 1 }, index_total: 3 }">

(translates to)

{
  index: {
    index_link: 1,
    index_section: 1
  },
  index_total: 3
}
```

This change will allow the above as well as a flattened structure to be passed, such as:

```
<div data-ga4-attribute="{ index_link: 1, index_section: 1, index_total: 3  }">

(will still translate to exactly the same)

{
  index: {
    index_link: 1,
    index_section: 1
  },
  index_total: 3
}
```

The schema remains the source of truth, but since our attribute names are unique we can find them in the schema and insert the passed values accordingly, preserving the intended object data structure.

This change is also backwards compatible, so we don't need to change any of the attributes in markup/components immediately. Once all of the attributes we currently have are flattened part of this code can be removed, as it currently checks whether the data passed to it is nested.

### Testing
I've tested this locally in `frontend` and it seems to work - changes to the schema are reflected automatically in events (e.g. moving `index_total` inside `index`), and changes to the tracking in the application are likewise updated to match the schema (e.g. adding `index_section_count` outside of the `index` attribute still shows it inside the attribute as per the schema).

## Why
Several reasons:

- this will slightly simplify our adding of data attributes to markup
- this will also make the code more flexible in future, e.g. if we wanted to move an existing attribute under another (say `index_total` inside the `index` object) this could be done with a minimal change to the schema itself

Having said that, this code is not as flexible as it could be as it doesn't do true object recursion - it only checks for sub attributes at one level more down. That is, it currently supports the structure of `index: { index_link: 1 }` but would not work if further depth were applied e.g. `index: { index_link: { example: 1 }}`.

## Visual Changes
None.

Trello card: https://trello.com/c/WO2to9JG/562-investigate-whether-we-need-to-nest-attributes-in-data-attributes


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js
```diff
@@ -45,19 +49,60 @@
...
     for (var property in data) {
-      if (property in schema.event_data) {
-        schema.event_data[property] = data[property]
+      // some passed data might be undefined, don't want it to overwrite e.g. the index sub parameters
+      if (data[property] !== undefined) {
+        if (exceptions.indexOf(property) >= 0 || !this.isAnObject(data[property])) {
+          schema.event_data = this.addToObject(schema.event_data, property, data[property])
+        } else {
+          // we check for one level of nesting in the data attributes data
+          // this check can be remove once nesting is removed from all data attributes
```
JamesCGDS, line 70: Possible typo - should `this check can be remove` be `this check can be removed`?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js
```diff
@@ -45,19 +49,60 @@
...
+            schema.event_data = this.addToObject(schema.event_data, subproperty, data[property][subproperty])
+          }
+        }
       }
     }
     return schema
   }
 
+  // might be easier to check if it's not a string or a number?
+  Schemas.prototype.isAnObject = function (item) {
```
JamesCGDS, line 81: Not sure if this is overkill but would we want to write a unit test for this function? I've noticed that tests have been written for `addToObject` (although admittedly `addToObject` is a more complex function).

JamesCGDS (commented):
Looks good 🚀  Just a couple of minor/non-blocking comments.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.spec.js
```diff
@@ -110,6 +110,74 @@ describe('Google Analytics schemas', function () {
     expect(returned).toEqual(expected)
   })
 
+  it('does not overwrite the schema is an undefined value is passed', function () {
```
JamesCGDS, line 113: Possible typo: should `does not overwrite the schema is an undefined value is passed` be `does not overwrite the schema if an undefined value is passed`

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js
andysellick: Fixed, thanks.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.spec.js
andysellick: Fixed, thanks.

JamesCGDS (approved):
🚀

PR #3442, opened 2023-06-09
Add support for HTML and external attachment types

ollietreend:
## What

This commit adds support for HTML and external attachments to the Attachment component.

So far, the Attachment component has only been able to represent file attachments – i.e. attachments which have a 'content type' and a thumbnail that changes depending on the type of file it is.

This change acts as an enabler for us to switch off custom rendering of embedded attachments in the Whitehall Publisher app. It also ports across the HTML attachment thumbnail SVG which is currently [held in Whitehall][1].

I've also taken the opportunity to tighten up the tests to specify which thumbnail should appear for the various different types of attachments. This just supplements existing tests rather than explicitly testing them independently.

I've also fixed a minor styling bug which was present, where custom thumbnail images didn't receive a grey border because the `` element was missing the required CSS class name.

## Why

So that we can stop Whitehall from rendering attachments using its own in-app partials. Instead, it'll be able to use the Govspeak gem to render attachments, which in turn uses this Attachment component. It'll bring Whitehall more in line with other publishing apps such as Content Publisher.

## Visual Changes

The styling of HTML and external attachments replicates what is currently rendered by Whitehall.

For example, see this publication which includes File, HTML and External attachment types (all currently rendered by Whitehall): [Traineeship employer incentive registration form](https://www.gov.uk/government/publications/traineeship-employer-incentive-registration-form)

| Description | Before | After |
| --- | --- | --- |
| Custom thumbnail | custom_thumbnail_before | custom_thumbnail_after |
| HTML attachment | N/A | html_attachment |
| External attachment | N/A | external_attachment |

Trello: https://trello.com/c/Z5IwGrAw/1177-remove-custom-attachment-rendering-from-whitehall

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fully implement loading of individual stylesheets in component guide ([PR #3379](https://github.com/alphagov/govuk_publishing_components/pull/3379))
+* Add support for HTML and external attachment types ([PR #3442](https://github.com/alphagov/govuk_publishing_components/pull/3442))
 
 ## 35.6.0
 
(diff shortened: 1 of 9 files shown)


kevindew:
This is great, will be really exciting to remove the rendering from Whitehall 👍

dnkrj (approved):
Looks great 🌞

PR #3443, opened 2023-06-09
Add ga4 tracking to single page notifications button

JamesCGDS:
## What
This PR adds GA4 tracking to the `single_page_notification_button` component (see example screenshot below).

image

The `single_page_notification_button` component can appear twice on a page - once at the top and once in the contextual footer ([example](https://www.gov.uk/guidance/travel-to-england-from-another-country-during-coronavirus-covid-19)). Whilst the component consists of a form and a submit button, it has been treated as a link in GA4 tracking terms as certain parameters such as `link_path_parts` and `link_domain` rely on functions that are only present in the `linkTracker` module (and therefore is unsuitable for the `eventTracker`).

## Why
Part of the GA4 migration.

[Trello card](https://trello.com/c/iP21pmuk/565-subscribe-get-emails-about-this-page-stop-getting-emails-about-this-page-button)

## Visual Changes
N/A

### DIFF
--- a/app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
+++ b/app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
@@ -1,25 +1,37 @@
 <%
   add_gem_component_stylesheet("single-page-notification-button")
 
-  component_helper = GovukPublishingComponents::Presenters::SinglePageNotificationButtonHelper.new(local_assigns)
+  spnb_helper = GovukPublishingComponents::Presenters::SinglePageNotificationButtonHelper.new(local_assigns)
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
 
   wrapper_classes = %w(govuk-!-display-none-print)
   wrapper_classes << shared_helper.get_margin_bottom
+
+  ga4_data_attributes = ga4_data_attributes ||= nil
+  ga4_link_data_attributes = ga4_data_attributes[:ga4_link] if ga4_data_attributes
+  ga4_link_data_attributes[:url] = spnb_helper.form_action if ga4_link_data_attributes
+
+  component_helper.add_class(wrapper_classes.join(" "))
+  component_helper.add_data_attribute({ module: "gem-track-click" })
+  component_helper.add_data_attribute({ module: ga4_data_attributes[:module] }) if ga4_data_attributes
 %>
 <% button_text = capture do %>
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
```diff
@@ -1,25 +1,37 @@
 <%
   add_gem_component_stylesheet("single-page-notification-button")
 
-  component_helper = GovukPublishingComponents::Presenters::SinglePageNotificationButtonHelper.new(local_assigns)
+  spnb_helper = GovukPublishingComponents::Presenters::SinglePageNotificationButtonHelper.new(local_assigns)
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
```
andysellick, line 6: 👍 for implementing the component wrapper helper 👍

app/views/govuk_publishing_components/components/_single_page_notification_button.html.erb
```diff
@@ -1,25 +1,37 @@
...
   wrapper_classes = %w(govuk-!-display-none-print)
   wrapper_classes << shared_helper.get_margin_bottom
+
+  ga4_data_attributes = ga4_data_attributes ||= nil
+  ga4_link_data_attributes = ga4_data_attributes[:ga4_link] if ga4_data_attributes
+  ga4_link_data_attributes[:url] = spnb_helper.form_action if ga4_link_data_attributes
+
+  component_helper.add_class(wrapper_classes.join(" "))
+  component_helper.add_data_attribute({ module: "gem-track-click" })
+  component_helper.add_data_attribute({ module: ga4_data_attributes[:module] }) if ga4_data_attributes
```
andysellick, line 17: This looks good but I think it would help to have an entry in the doc file to demonstrate this functionality.

I created one myself and noticed something odd - the GA4 data module is being added to both the wrapping element and the form, although I'm not quite sure how.

andysellick (commented):
Looks good, although one possible problem (also would squash the two commits).

andysellick:
@JamesCGDS just had another thought about this. Is it better to have the ga4 data attributes as their own separate thing that gets passed to the component, or should we use the existing `data_attributes` attribute? The latter seems tidier but you'd have to switch around the markup slightly, and other than tidiness I'm not sure what the benefit overall is - but something to consider perhaps.

JamesCGDS:
Thanks for the review @andysellick 👍 Re the `data_attributes` - following on from our call earlier, I'm thinking of keeping it the same as I believe using `data_attributes` was causing the duplicating behaviour

andysellick (approved):

PR #3449, opened 2023-06-13
Add 'View online' preview link to Attachment

ollietreend:
## What

This adds a new `preview_url` parameter that can be passed to the Attachment component to show a "View online" preview link.

This change acts as an enabler for us to switch off custom rendering of embedded attachments in the Whitehall Publisher app.

Whitehall uses the 'View online' link to allow users to preview the contents of CSV files online, without having to download the file to their device and open it in a compatible application.

Unforuntately we don't have translations for the "View online" label, so I've left it blank in all non-English locales. We may choose to add these at a later date.

## Why

So that we can stop Whitehall from rendering attachments using its own in-app partials. Instead, it'll be able to use the Govspeak gem to render attachments, which in turn uses this Attachment component. It'll bring Whitehall more in line with other publishing apps such as Content Publisher.

## Visual Changes

The govuk_publishing_components Attachment differs slightly from the custom-rendered Whitehall partial when rendering a previewable CSV. We are deliberately following the existing pattern established by this Attachment component, such as making the attachment title and thumbnail the link to 'download' the file.

### Before (rendered by Whitehall)

Example page: https://www.gov.uk/government/statistics/uks-carbon-footprint

before

### After (rendered by govuk_publishing_components)

after

---

Trello: https://trello.com/c/Z5IwGrAw/1177-remove-custom-attachment-rendering-from-whitehall


dnkrj (approved):
Nice!

PR #3454, opened 2023-06-20
Refactor document list component

andysellick:
**NOTE** this PR is linked to [changes in collections](https://github.com/alphagov/collections/pull/3301), which relies on this change.

## What
- remove top and bottom margin options (undocumented and soon to be not used)
- add shared helper for margin bottom, and document
- change CSS to prevent last list item from setting the bottom margin of the component

## Why
The document list component had no bottom margin on the parent item by default. However, every list item in it had a bottom margin of 25px, which meant that the 25px of bottom margin extended out of the parent list, effectively setting the bottom margin of the component.

(for reference, this is a weird CSS thing where if a child has bottom margin it will escape from the parent unless the parent has padding (or any number of other things, turns out margin is complex) and the parent in this case didn't have padding)

This seemed like a brittle situation, so this changes the list items to have the same margin top instead, and setting the first-child to margin top zero. At the same time we add the shared_helper margin bottom options, setting by default a bottom margin of 25px, so there should be no visual change.

Turns out the document list component also had options to set the margin top and bottom but this was not documented. Also the margin bottom setting was less than the escaped margin from the last list item, so it wouldn't have had any effect anyway. I've checked our applications and the only place where this component is used with those options is `collections`, so a separate PR will be raised to deal with this.

**Further note**: there's another undocumented feature of this component, which is to have it not be wrapped in a `<ul>` by passing `within_multitype_list` as an option. Does anyone know if this is being used anywhere? I couldn't see any.

## Visual Changes
None.

The visual changes in Percy are because I added a new documentation entry for the new margin bottom option.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Refactor document list component ([PR #3454](https://github.com/alphagov/govuk_publishing_components/pull/3454))
 * Add option for blue bar background to public_layout ([PR #3380](https://github.com/alphagov/govuk_publishing_components/pull/3380))
 * Add ga4-link attribute for other 'see all updates' link ([PR #3451](https://github.com/alphagov/govuk_publishing_components/pull/3451/))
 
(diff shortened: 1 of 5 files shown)


MartinJJones (approved):
Nice work, an interesting issue.

I've approved the visual regression change Percy has highlighted as well, as this is expected.

PR #3455, opened 2023-06-22
Use safe navigation operator in asset_helper.rb

MartinJJones:
## What
Update the `viewing_component_guide?` method to use the safe navigation operator in asset_helper.rb

## Why
This will fix the issue below when using the AssetHelper in the smart answers rendering application.

```
undefined method `path' for nil:NilClass
```

## Further info
The issue in smart answers is that the AssetHelper tries to check if we are viewing the component guide, it does this by checking if the request path contains "component-guide", however the "request" object is not available when trying to render govspeak outside of the component-guide.

This only appears to happen in smart answers, using govspeak in another rendering application does not cause any issues.

Using this approach, calling `!viewing_component_guide?` will return `false` if the request object is not available, even if the component guide is being viewed. The impact here would be low, and limited to stylesheets from static not being loaded in the component guide. For this to happen, the request object would need to be nil in the component guide.

It is worth nothing that when testing release 35.7.0 of the gem in smart answers, we had access to the request object when viewing the component guide, so we are still able to perform this check.

### DIFF
--- a/lib/govuk_publishing_components/app_helpers/asset_helper.rb
+++ b/lib/govuk_publishing_components/app_helpers/asset_helper.rb
@@ -87,7 +87,7 @@ module GovukPublishingComponents
       end
 
       def viewing_component_guide?
-        request.path.include?("/component-guide")
+        request&.path&.include?("/component-guide")
       end
     end
   end
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Add option for blue bar background to public_layout ([PR #3380](https://github.com/alphagov/govuk_publishing_components/pull/3380))
 * Add ga4-link attribute for other 'see all updates' link ([PR #3451](https://github.com/alphagov/govuk_publishing_components/pull/3451/))
 * Change GA4 type ([PR #3456](https://github.com/alphagov/govuk_publishing_components/pull/3456))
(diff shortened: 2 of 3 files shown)


lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -87,7 +87,7 @@ def is_already_used?(component)
       end
 
       def viewing_component_guide?
-        request.path.include?("/component-guide")
+        request&.path&.include?("/component-guide")
```
andysellick, line 90: I'm not familiar with this syntax but I assume it does what it's supposed to! Is there any way there could be a test written for this?

MartinJJones: This was new to me in Ruby/Rails as well, there is further documentation on this here - https://ruby-doc.org/core-2.6/doc/syntax/calling_methods_rdoc.html#:~:text=Safe%20navigation%20operator%C2%B6%20%E2%86%91,if%20the%20call%20is%20skipped.

I should be able to add a test for this, will add this now.

MartinJJones: I've added a test to check what happens when the request object is not available - https://github.com/alphagov/govuk_publishing_components/pull/3455/commits/7725991e23b74ddcc9b298475752ad5706111b31

andysellick (approved):
👏

PR #3456, opened 2023-06-22
Change GA4 type

JamesCGDS:
## What
This PR makes a change to the `type` parameter of elements that appear at the top and bottom of a `Part of` sidebar. Previously, the `type` was set to `related content` but this has now changed to `part of`.

Example (top):
image

Example (bottom):
image

## Why
GA4 bug fix.

[Trello card](https://trello.com/c/oXi9yaRO/598-part-of-also-part-of-showing-event-type-related-content-rather-than-part-of)

## Visual Changes
N/A


JamesCGDS:
As this change is very small, I'm not sure it would warrant an entry in the CHANGELOG. Happy to discuss though.

andysellick (approved):
It is a small change, but I think a CHANGELOG entry would be useful for keeping track of which changes are in which release of the gem.

PR #3458, opened 2023-06-26
Upgrade LUX to version 309

andysellick:
## What / why
Updating to the latest version of LUX (from 308 to 309). Has been checked and tested [according to the docs](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#testing).

See https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md for full details.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
The changes look good to me, after following the documentation I can see that the version number is now set to 309 and the `customerid` and `samplerate` are set as expected 👍

Issue #3439, opened 2023-06-09
Notice component fails 2.5.8 Target Size (Minimum)


selfthinker:
The [Notice component](https://components.publishing.service.gov.uk/component-guide/notice) fails [2.5.8 Target Size (Minimum)](https://w3c.github.io/wcag/understanding/target-size-minimum.html).

Note: 2.5.8 is part of the new draft WCAG 2.2 criteria. The latest standard is WCAG 2.1. WCAG 2.2 isn't published yet, so this is potentially subject to change.

The Notice component fails when links are on separate lines (like in the ‘Without title’ example). That's because those links are only 23.5px high and have no further spacing between them. They need to be at least 24px high or have additional spacing that adds up to 24px in order to pass.

Screenshot showing the dimensions of the link:

<img width="421" alt="Screenshot of notice component without title" src="https://github.com/alphagov/govuk_publishing_components/assets/108893/29ce601d-9728-4e79-bd71-dd25e3a2ab15">

jon-kirwan:
@selfthinker I wonder if we need to make any changes here.

The Notice component example includes the following markup (`<br>` elements are included to produce a line breaks).
```
<p>
Scheduled to publish at 8am on 25 April 2019<br />
<a href="change-date">Change date</a><br />
<a href="stop-scheduled-publishing">Stop scheduled publishing</a>
</p>
```

However, from several instances I've looked at, publishers often seem to either add links in inline text or format them with a list e.g. -

- [Code of practice for the exercise of stop and search powers ](https://www.gov.uk/government/publications/code-of-practice-for-the-exercise-of-stop-and-search-powers)
- [About the Document Checking Service pilot scheme](https://www.gov.uk/guidance/apply-for-the-document-checking-service-pilot-scheme)
- [VAT Mini One Stop Shop: Non-Union Return](https://www.gov.uk/government/publications/vat-mini-one-stop-shop-non-union-return)
- [NHS COVID-19 app](https://www.gov.uk/government/collections/nhs-covid-19-app)
- [How to classify trade marks](https://www.gov.uk/government/publications/how-to-classify-trade-marks)
- [Withdrawn: School inspection handbook](https://www.gov.uk/government/publications/school-inspection-handbook-from-september-2015)
- [Legal definition of waste guidance](https://www.gov.uk/government/publications/legal-definition-of-waste-guidance)
[...]

PR #3465, opened 2023-06-29
Update notice component example

jon-kirwan:
## What

Update HTML in Notice example to remove line break elements and replace with unordered list (to represent the two links).

## Why

WCAG 2.2 2.5.8 fail (Target Size (Minimum) (Level AA)). See https://w3c.github.io/wcag/understanding/target-size-minimum.html.

### Before

components-gem-pr-3464 herokuapp com_component-guide_notice_without_title

### After

components-gem-pr-3465 herokuapp com_component-guide_notice_without_title

## Anything else

See https://github.com/alphagov/govuk_publishing_components/issues/3439


maxgds (approved):
I think in reality we wouldn't mark up links as a list in that way but for a quick sample text and to fix the issue this seems sensible.

PR #3466, opened 2023-06-29
Improve the govspeak table layout when text direction is right-to-left

KludgeKML:
## What

- Improve the govspeak table layout when text direction is right-to-left
- Add example in the component guide docs

## Why

Currently table headings and captions in govspeak elements are set to left align, which ignores the direction set by the direction-rtl class and means that (eg) [Arabic-language pages with tables](https://www.gov.uk/government/people/cecille-el-beleidi.ar) in them have the table headings appear left-aligned while the table cells are right-aligned.

Lastly, even though the table cells are right-aligned, the padding around the table cells was still set based on a text direction of left-to-right being used. This meant that the text was not aligned to the right of the column and some text would appear very close together.

### Before (headers and table contents left-aligned)

govspeak-table

### After (headers and table contents right-aligned)

govspeak-table-rtl


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix Govspeak th and caption items setting text-align to left. Replace with start, which works correctly with RTL language documents ([PR #3466](https://github.com/alphagov/govuk_publishing_components/pull/3466))
```
andysellick, line 12: Can this match the PR name please?

andysellick (approved):
Looks good, one minor quibble. There's a release going out today if you'd like to get it in?

CHANGELOG.md
MartinJJones: Done 👍

andysellick (approved):

PR #3471, opened 2023-06-30
Improve single page notification button helper

hannako:
## What

Prior to this PR, if `skip_account` is present, the form action for the single page notification button is the non-govuk-acount signup endpoint, and the params necessary for that route are added to the request.

If `skip_account` is `nil`, the default form action is set that enforces the govuk_account 

New changes:
Instead of [checking if `skip_account` is present](https://github.com/alphagov/govuk_publishing_components/blob/7aecb57136a75439e5d8262cf7b5950ef2e1d3bd/lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb#L64), [check if `skip_account == "true"`](https://github.com/alphagov/govuk_publishing_components/blob/b36cc86680060aabc503a2be97239f56773a104d/lib/govuk_publishing_components/presenters/single_page_notification_button_helper.rb#L83-L85)

## Why

When this component is called by the frontend, we can pass anything into the the skip_account field, so we need to make the code here stricter. For example passing `"false"` to skip_account in Frontend, we would expect the component to implement the default non-govuk-account sign up path for the button. But in fact that would not have been the case, because `"false".present?` returns true.

## Visual Changes
None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update single page notification button ([PR #3471](https://github.com/alphagov/govuk_publishing_components/pull/3471))
 * Fix GA4 accordion tracking bugs ([PR #3461](https://github.com/alphagov/govuk_publishing_components/pull/3461))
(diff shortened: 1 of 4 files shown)


BeckaL (approved):
LGTM 👍

PR #3475, opened 2023-07-03
Add margin_bottom param to Attachment component

ollietreend:
## What

The Attachment component defaults to having no margin-bottom.

However sometimes we need it to have a margin-bottom value, for example when rendering a list of attachments on a document. Otherwise the components look too 'squashed up'.

This has previously been worked around by [wrapping the component in a container element][1] with margin bottom. But this feels a bit hacky and becomes trickier when we don't have so much control over the way the markup around the component (for example, when it's been embedded within a Govspeak document).

[1]: https://github.com/alphagov/government-frontend/blob/0bec9a842902db2a7353d28400fa7bf5d7008c8b/app/views/content_items/_attachments.html.erb#L18-L23

## Why

To make the attachment usable within Whitehall and Govspeak, without needing extra wrapper `<div>`s to apply a margin. This will enable us to [remove custom attachment rendering from Whitehall](https://trello.com/c/Z5IwGrAw/1177-remove-custom-attachment-rendering-from-whitehall).

### Before

N/A

### After

Attachment with margin bottom

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add margin_bottom param to Attachment component ([PR #3475](https://github.com/alphagov/govuk_publishing_components/pull/3475))
+
 ## 35.9.0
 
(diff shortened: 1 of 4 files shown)


andysellick (approved):
Nicely done. Excellent use of the shared helper 👏

PR #3478, opened 2023-07-05
Wrap the text of start buttons in a span

CristinaRO:
When using an inline element (such as `<abbr>`, `<address>` or `<telephone>`) in a link with the `govuk-button--start` class applied to it, there is no spacing between the inline element and the text of the link.

The reason that this is occurring is the `display: inline-flex;` rule in the `govuk-button--start` class.

The suggested solution was to wrap the text of `start` buttons in a `<span>` tag when rendering a start button:
https://github.com/alphagov/govspeak/pull/277#issuecomment-1620327662

This is the solution I'm implementing here.


maxgds (commented):
Unfortunately I don't think this is a good idea. Abbreviation tags are a bit problematic in the way they work generally as they tend to generate text to be seen on hover that isn’t really stylable unless you do a load of extra work. That aside, the hover just won’t work within the button space reliably because the button has a hover state of its own. I put an abbr into the button using dev tools and the result is that the abbreviation gains a dotted underline that looks weird and hovering over it does nothing. You mentioned other examples such as telephone numbers and addresses, we generally advise against marking telephone numbers up as it can lead to unpredictable behaviour, and this would be magnified if on a button. Putting an address on a button seems like an odd thing to want to do, too. I think it would be better to avoid the input for a button being anything other than plain text, and if that’s going to be unpredictable do some cleansing to remove extra markup before passing it in. Given all this there seems to be no reason to add the span unless we were to come up with a must-have use case for markup, which seems unlikely as the button should be a relatively simple component with a single purpose.

CristinaRO:
@patrickpatrickpatrick Looping you in as the [original reporter](https://github.com/alphagov/govuk-frontend/issues/3406) of this issue, and @owenatgov as someone with Opinions! :)

maxgds (approved):
After a few discussions we have decided that it does make sense to push this fix. Eventually we should see a fix from the Design System team, but this local patch makes sense in the interim. 

That said, the behaviour of abbreviations is often confusing, and this does stack two interactions on top of each other in a less than ideal way. It should be discouraged where possible, but we will handle it better when it does happen.

PR #3479, opened 2023-07-05
Release 35.10.0

ollietreend:
## 35.10.0

* Add margin_bottom param to Attachment component ([PR #3475](https://github.com/alphagov/govuk_publishing_components/pull/3475))


andysellick:
@ollietreend for a change to the API (in this case adding an option to a component) we'd normally count this as a minor release, rather than a patch.

ollietreend:
Ah, fair point. Thanks @andysellick! I'll change that now.

andysellick (approved):

PR #3481, opened 2023-07-06
Fix links to github in component auditing

andysellick:
## What / why
Some of the links to component files in github in the component auditing weren't correct.

This only happened when viewing a component guide for an application when it was [rendered remotely on heroku](https://govuk-collections.herokuapp.com/component-guide/audit). The links to component files (in component details) were being incorrectly generated as follows, using `app` as the name of the application, when it should have been e.g. `collections`.

- **WRONG** github.com/alphagov/app/blob/main/`app`/views/components/_taxon_list.html.erb
- **RIGHT** github.com/alphagov/app/blob/main/`collections`/views/components/_taxon_list.html.erb

This was happening because an application doesn't actually know its own name, so the application name was being inferred from the path. On a local system that works (you'd have the repo for an app checked out somewhere like `/home/me/govuk/collections`) but when this was being built remotely on heroku the path was ending in `app` instead, so the links weren't correct.

For reference, the component auditing has to work in several locations:

- running locally inside `govuk_publishing_components` (can access itself and other apps)
- running locally inside an application such as `finder-frontend` (can access itself and only that app)
- running remotely inside `govuk_publishing_components` (can access only itself)
- running remotely inside an application such as `finder-frontend` (can access itself and only that app)

## Visual Changes
None.


gclssvglx (approved):
Looks good 👍  One suggestion though - instead of manually adding the app name, for Rails apps, we could do either of these...

```ruby
Rails.application.class.module_parent_name.underscore.dasherize
File.basename(Rails.root.to_s)
```
I personally prefer the first one as I feel it'll play nicer than the second the app is deployed.

Of course if the app is Sinatra or something else, then hard-coding the name seems the best option.

MartinJJones (approved):
Changes look good to me 👍

andysellick:
@gclssvglx that was so helpful, thank you! I knew there should be a way of doing this without resorting to adding extra config. Thanks!

PR #3486, opened 2023-07-06
Add link option to tabs component

andysellick:
## What
Adds an option to the tabs component to make the tabs act as links, instead of switching between hidden elements on the same page.

Note that for the purposes of tracking there has to be a bit of logic involved to switch from the GA4 event tracker to the link tracker. We're not actually tracking tabs as links anywhere on GOV.UK so this tracking is a bit unnecessary but it feels like more of a complete change if this is included now.

## Why
We have a tabs component in `collections-publisher` that looks exactly the same as this tabs component, but acts as links. If we add this option to the gem component we can reuse it in that application and remove a redundant component.

Related to issue https://github.com/alphagov/govuk_publishing_components/issues/3487

## Visual Changes
Screenshot 2023-07-06 at 14 36 11

### DIFF
--- a/spec/components/tabs_spec.rb
+++ b/spec/components/tabs_spec.rb
@@ -31,7 +31,29 @@ describe "Tabs", type: :view do
     assert_select "#tab1"
   end
 
-  it "renders with GA4 tracking" do
+  it "renders tabs as links" do
+    render_component(
+      as_links: true,
+      tabs: [
+        {
+          label: "First section",
+          href: "/page1",
+        },
+        {
+          label: "Second section",
+          href: "/page2",
+        },
+      ],
+    )
+
+    assert_select ".govuk-tabs"
+    assert_select ".govuk-tabs__tab", 2
+    assert_select ".govuk-tabs__list-item", 2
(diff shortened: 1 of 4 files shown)


spec/components/tabs_spec.rb
```diff
@@ -31,7 +31,29 @@ def component_name
...
+          href: "/page2",
+        },
+      ],
+    )
+
+    assert_select ".govuk-tabs"
+    assert_select ".govuk-tabs__tab", 2
+    assert_select ".govuk-tabs__list-item", 2
+    assert_select ".govuk-tabs__tab[href='/page1']"
+    assert_select ".govuk-tabs__tab[href='/page2']"
```
AshGDS, line 53: Nitpick: Should this test check that they're actually `<a>` tags? E.g. `assert_select "a.govuk-tabs__tab[href='/page1']"`

AshGDS (approved):
Looks good, added a minor nitpick comment :+1:

spec/components/tabs_spec.rb
andysellick: Good shout, have updated, thanks.

PR #3495, opened 2023-07-11
GA4 content navigation fixes

AshGDS:
## What

- Instead of using `ga4-track-links-only` for the contents link tracking, I'm setting the GA4 attributes on each link instead. This is because using `ga4-track-links-only` didn't work properly for our requirement of trying to merge the indexes but also merging `index_total`. So it's easier to add the attributes to each link instead.

- Adds `action` to the `See all updates` link tracking, which was missed.

## Why

- Changes requested in https://github.com/alphagov/government-frontend/pull/2801

## Visual Changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_contents_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_contents_list.html.erb
@@ -13,20 +13,16 @@
   link_classes << "govuk-link--no-underline" unless underline_links
 
   ga4_tracking ||= false
-  ga4_data = nil
-  if ga4_tracking
-    ga4_data = {
+  ga4_data = {
       event_name: "navigation",
       type: "content",
       section: t("components.contents_list.contents", locale: :en) || ""
-    }.to_json
-  end
+  } if ga4_tracking
   local_assigns[:aria] ||= {}
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-contents-list #{brand_helper.brand_class}")
   component_helper.add_data_attribute({ module: "gem-track-click" })
   component_helper.add_data_attribute({ module: "ga4-link-tracker" }) if ga4_tracking
-  component_helper.add_data_attribute({ ga4_link: ga4_data, ga4_track_links_only: "" }) if ga4_tracking
   component_helper.add_aria_attribute({ label: t("components.contents_list.contents") }) unless local_assigns[:aria][:label]
   component_helper.add_role("navigation")
 -%>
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_contents_list.html.erb
```diff
@@ -19,14 +19,13 @@
       event_name: "navigation",
       type: "content",
       section: t("components.contents_list.contents", locale: :en) || ""
-    }.to_json
+    }
```
JamesCGDS, line 22: Minor refactoring suggestion:

```
ga4_data = {
    event_name: "navigation",
    type: "content",
    section: t("components.contents_list.contents", locale: :en) || ""
} if ga4_tracking
```

What do you think?

app/views/govuk_publishing_components/components/_contents_list.html.erb
```diff
@@ -44,8 +43,12 @@
       <% contents.each.with_index(1) do |contents_item, position| %>
         <li class="<%= cl_helper.list_item_classes(contents_item, false) %>" <%= "aria-current=true" if contents_item[:active] %>>
           <% link_text = format_numbers ? cl_helper.wrap_numbers_with_spans(contents_item[:text]) : contents_item[:text]
-             ga4_link_data = nil
-             ga4_link_data = { "index": { "index_link": position }, "index_total": contents.length }.to_json if ga4_tracking
+             if ga4_data
```
JamesCGDS, line 46: What do you think about putting `if ga4_tracking` here instead of `if ga4_data`? And then I think we could remove the declaration `ga4_data = nil` (currently on line 16)

app/views/govuk_publishing_components/components/_contents_list.html.erb
```diff
@@ -56,7 +59,7 @@
               track_options: {
                 dimension29: contents_item[:text]
               },
-              ga4_link: ga4_link_data
+              ga4_link: ga4_data.to_json
```
JamesCGDS, line 62: Quick question: I think the current implementation will always add a `data-ga4-link` attribute. If `ga4_data` is `nil`, then `data-ga4-link` will be present and set to `null` because it's still calling `to_json` on a `nil` value. I've had a quick tinker around and this seems to work as needed: `ga4_link: (ga4_tracking ? ga4_data.to_json : nil)`

What do you think?

JamesCGDS (commented):
Looks good!  I've left a few comments/suggestions - happy to discuss 👍

AshGDS:
@JamesCGDS Thanks - I've made all the requested changes :+1:

JamesCGDS (approved):
🚀

PR #3500, opened 2023-07-17
GA4 HTML attachment tracking

AshGDS:


spec/components/attachment_spec.rb
```diff
@@ -268,4 +268,20 @@ def assert_thumbnail(type, src: nil)
...
+  it "adds GA4 tracking on HTML attachment links" do
+    render_component(
+      attachment: {
+        title: "Test",
+        url: "https://assets.publishing.service.gov.uk/government/uploads/system/uploads/attachment_data/file/687542/February_2018_Consular_MI.csv",
+        filename: "test",
+        type: "html",
+      },
+    )
+    puts rendered
```
JamesCGDS, line 281: Does this need removing?

JamesCGDS (changes requested):
Looks good, just one small thing I noticed 👍 Also I think the docs may also need updating because [`analytics.md`](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/analytics.md) makes reference to `init-ga4` adding attributes for attachment link tracking

spec/components/attachment_spec.rb
AshGDS: Good spot, removed :+1:

AshGDS:
Thanks @JamesCGDS, I've made the changes :+1:

JamesCGDS (approved):

PR #3501, opened 2023-07-17
Fix attachment metadata styling issue

MartinJJones:
## What
Fixes a styling issue where long strings of text for external links do not wrap onto a new line.

The approach used to fix the issue is the same as that used in the Design System - https://github.com/alphagov/govuk-frontend/blob/49a3eb389eec89e004873ddfeb8c1f9aa1d572e4/packages/govuk-frontend/src/govuk/components/summary-list/_index.scss#L64-L68

## Why
The long string of text was not contained and overlapped with other elements on the page.

## Further info
Tested in the browsers below using BrowserStack:

- Chrome
- FireFox
- Safari
- Safari iOS
- IE11

The issue still exists in IE11, but it is worth noting that this issue existed before the recent change to use the attachment component from the gem.

I will create a separate issue for IE11, so this can be looked into separately.

### Before
before-text

#### Before - If the 1st link does not contain a hyphen
before-text-no-hyphen

### After
after-text

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * GA4 content navigation fixes ([PR #3495](https://github.com/alphagov/govuk_publishing_components/pull/3495))
 * Fix feedback component spacing ([PR #3470](https://github.com/alphagov/govuk_publishing_components/pull/3470))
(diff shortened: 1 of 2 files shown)


maxgds (approved):
Nice.

MartinJJones:
Github issue created to review the attachment component further in the future, includes further info on IE11.

PR #3502, opened 2023-07-17
Render an outline for search button and input in Windows high contrast mode

hannalaakso:
## What
Add a transparent outline to the search button in all modes, and to form inputs in the blue background mode. This way Windows high contrast mode renders an outline for the elements, which helps to distinguish them from their background and makes them easier to identify. 

This does not result in any visual changes in the standard contrast mode.

## Why
We remove the default border from the search button in all modes, and the default border from from form input elements in the blue background mode, in order to comply with the GOV.UK design. 

However, this meant that these elements didn't have a visible outline in Windows high contrast mode which could make them difficult to distinguish from their background, or to identify. (This wasn't an issue in Firefox when users changes the default browser colours there because Firefox renders a distinct colour background inside buttons and form inputs.)

## Visual Changes

In Edge with Windows high contrast mode enabled.

### Default

[Example](https://components.publishing.service.gov.uk/component-guide/search/default)

#### Before

Screenshot 2023-07-24 at 15 55 23

#### After

Screenshot 2023-07-24 at 15 55 16

### Blue background

[Example](https://components.publishing.service.gov.uk/component-guide/search/for_use_on_govuk_blue_background)

### Before

Screenshot 2023-07-24 at 15 55 39

### After

Screenshot 2023-07-24 at 15 55 02

# Tested with

## High Contrast mode
✅ Windows High Contrast mode
✅  Firefox when uses changes the default colours in the browser

## Windows	
✅  Edge 114
✅  Google Chrome, latest
✅  Mozilla Firefox, latest
✅  Internet Explorer 11	

## macOS	
✅  Safari 16
✅  Google Chrome, latest
✅  Mozilla Firefox, latest

## iOS	
✅  Safari, iOS 16

## Android	
✅  Samsung Galaxy, Android 13

Fixes https://trello.com/c/bqURvY4r/1900-form-controls-may-not-be-identified-when-in-high-contrast-mode-45-m

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Extend component wrapper helper and use on heading component ([PR #3519](https://github.com/alphagov/govuk_publishing_components/pull/3519))
 * Remove licence-finder from list of audited applications ([PR #3518](https://github.com/alphagov/govuk_publishing_components/pull/3518))
 * Fix inconsistent focus state on document list component ([PR #3468](https://github.com/alphagov/govuk_publishing_components/pull/3468))
+* Render an outline for search button and input in Windows high contrast mode ([PR #3502](https://github.com/alphagov/govuk_publishing_components/pull/3502))
 
 ## 35.12.0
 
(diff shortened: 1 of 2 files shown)


matthillco (approved):
I've tested this in Windows 10 and 11 under Edge, Chrome and Firefox and the outline appears as expected so I'd say this is fixed. Nice job.

PR #3509, opened 2023-07-20
Add functionality for preventing the redaction of publicly available information

JamesCGDS:
## What
This PR adds functionality for preventing redaction by detecting a `data-ga4-do-not-redact` attribute on elements that contain information that we don't want redacted. For example, all email addresses are currently redacted however, in some cases, we don't want to redact this information e.g. when the email address is publicly available contact information and surfaced within a set of results [(example)](https://www.gov.uk/disabled-students-allowances-assessment-centre#results).

## Why
Requested by PA's.

[Trello card](https://trello.com/c/nDZmuHH1/622-public-email-address-redacted-on-place-forms)

## Visual Changes
N/A

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -272,6 +272,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
           // if there's a problem with the config, don't start the tracker
           console.error('Unable to JSON.parse or JSON.stringify index: ' + e.message, window.location)
         }
+      },
+
+      applyRedactionIfRequired: function (PIIRemover, element, data) {
+        return element.closest('[data-ga4-do-not-redact]') ? data : PIIRemover.stripPIIWithOverride(data, true, true)
       }
     },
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -297,6 +297,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
             }
           }
         }
+      },
+
+      applyRedactionIfRequired: function (PIIRemover, element, data) {
+        return element.hasAttribute('data-ga4-do-not-redact') ? data : PIIRemover.stripPIIWithOverride(data, true, true)
```
AshGDS, line 303: Should we use `element.closest` instead? This would allow the element to check itself and it's parent elements for the attribute. This would account for clicking on a child link within content, and would allow you to put the attribute on the parent div instead of on every individual link on a page.

AshGDS (commented):
LGTM :+1: I've suggested a potential improvement

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
JamesCGDS: Thanks @AshGDS, good idea - I've made that change and I've also added another test to check that this is working when `data-ga4-do-not-redact` is applied directly on an element and on a parent element.

AshGDS (approved):

PR #3514, opened 2023-07-21
Fix empty input validation issues

jon-kirwan:
## What

Fix empty input validation issues for:

- is this page useful? / email survey sign up form
- report a problem with this page form

Added the Accept request HTTP header (with the XHR request headers) to indicate the the content type of JSON. Rails determines the desired response format from the HTTP Accept header submitted by the client. See https://api.rubyonrails.org//classes/ActionController/MimeResponds.html#method-i-respond_to.

**Depends on:**
- https://github.com/alphagov/feedback/pull/1678

## Why

To provide error feedback to the user when submitting an empty form (or with an incorrectly formatted email).

#### Before

www gov uk_browse

#### After

collections dev gov uk_browse(iPad Air)

#### Before

www gov uk_browse(iPad Air) (1)

#### After

collections dev gov uk_browse(iPad Air) (1)

#### Before

www gov uk_browse(iPad Air) (2)

#### After

collections dev gov uk_browse(iPad Air) (2)

#### Before

www gov uk_browse(iPad Air) (3)

#### After

collections dev gov uk_browse(iPad Air) (3)

#### Before

www gov uk_browse(iPad Air) (8)

#### After

collections dev gov uk_browse(iPad Air) (5)

#### Before

www gov uk_browse(iPad Air) (6)

#### After

collections dev gov uk_browse(iPad Air) (6)

#### Before

www gov uk_browse(iPad Air) (7)

#### After

collections dev gov uk_browse(iPad Air) (7)

## Anything else

- report a problem with this page form - the empty validation issue isn't captured elsewhere as an issue
- is this page useful? / email survey sign up form - I've changed the HTML input type from `email` to `text` since controller level validations ensure that every user provides a valid email address (also input `type='email'` activates built-in form validation which i've not seen used anywhere else on GOV.UK)
- there is a test for "**submitting a form that fails because `email_survey_signup[survey_source]` is missing**" which expects a response message of "email survey sign up failure" and then applies a generic error message. However, I think this is not correct since the response message was amended [here](https://github.com/alphagov/feedback/commit/b9ae4f49fd63a357a53764df983fdd0f682f05cc). As a temporary fix I've updated the JS to handle this type of response message but will need further investigation to decide if the controller should be updated to return this type of response message (for a missing `survey_source`) or should the test be removed.

https://github.com/alphagov/feedback/pull/1678


jon-kirwan:
Unfortunately, the changes included in the pull request fail to resolve the issue outside of development environments due to issues described here: https://trello.com/c/0xzDRPPl/27-nginx-clobbers-error-http-responses-of-frontend-applications.

PR #3515, opened 2023-07-24
Move GA4 attachment link tracking to all attachment links

AshGDS:
## What

- Moves the GA4 link tracking from just being on HTML attachment links, to all attachment links

## Why

- The PAs requested this happen after the HTML attachment tracking is deployed. Therefore this can be merged/deployed after the HTML attachment link tracking is live.
- [Trello card](https://trello.com/c/MHFXw6H2/552-change-filedownloads-on-attachment-pages-to-have-attachment-instead-of-generic-download-type)

## Visual Changes
None.


spec/components/attachment_spec.rb
```diff
@@ -269,19 +269,18 @@ def assert_thumbnail(type, src: nil)
     assert_select '.gem-c-attachment:not([class*="govuk-!-margin-bottom-"])'
   end
 
-  it "adds GA4 tracking on HTML attachment links" do
+  it "adds GA4 tracking on attachment links" do
```
andysellick, line 272: Nitpick: maybe rename this test to it `includes GA4 tracking on attachment links by default`

andysellick (approved):
Looks good. Maybe worth getting @JamesCGDS to give it a second glance, as I think he added this in the first place?

AshGDS:
Thanks @andysellick - I added this initially, so it shouldn't need a second look from @JamesCGDS :+1:

PR #3521, opened 2023-07-27
Fix 100% scroll problem

andysellick:
## What
Fix a problem with the scroll tracker, specifically:

- in some cases the percentage scroll mode doesn't work for the 100% scroll position
- I think this is being caused by the potential for a horizontal scrollbar, so the cleanest solution is to subtract 20px from the overall calculated page height prior to percentage position calculation, making all of the percentage positions very slightly further up the page, but allowing the 100% position to be reachable in all tested scenarios

Thanks to @AshGDS for his invaluable help in figuring out this problem 👏 

## Why
Sometimes the 100% scroll track event doesn't fire, because even if you're at the bottom of the page the calculated 100% position is slightly further away, and unreachable.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix 100% scroll problem ([PR #3521](https://github.com/alphagov/govuk_publishing_components/pull/3521))
 * Extend component wrapper helper and use on heading component ([PR #3519](https://github.com/alphagov/govuk_publishing_components/pull/3519))
 * Remove licence-finder from list of audited applications ([PR #3518](https://github.com/alphagov/govuk_publishing_components/pull/3518))
 * Fix inconsistent focus state on document list component ([PR #3468](https://github.com/alphagov/govuk_publishing_components/pull/3468))
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-scroll-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-scroll-tracker.js
@@ -199,14 +199,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   Ga4ScrollTracker.Percentage.prototype.getTrackingNodes = function (trackedNodes) {
     var body = document.body
     var html = document.documentElement
-    var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight)
+    // remove 20px from the calculated page height to allow for a possible horizontal scrollbar
+    var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight) - 20
 
     var percentDetails = []
 
     for (var i = 0; i < this.config.percentages.length; i++) {
       var percent = this.config.percentages[i]
-      // subtract 1 pixel to solve a bug where 100% can't be reached in some cases
-      var pos = ((pageHeight / 100) * percent) - 1
+      var pos = ((pageHeight / 100) * percent)
       var alreadySeen = false
       if (trackedNodes.length) {
         alreadySeen = trackedNodes[i].alreadySeen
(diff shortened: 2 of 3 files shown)


AshGDS (changes requested):
Hi @andysellick, thanks for this, but unfortunately the problem still persists with this change. 

You might be able to replicate the issue by trying various zoom levels in your browser if you zoom in, as the 100% scroll event seems to work at certain zoom levels but break at other zoom levels. I'm console logging `percentDetails` and `window.scrollY + window.innerHeight` to see the differences.

I think the problem stems from these values:

```JavaScript
var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight)
```

and

```JavaScript
  Ga4ScrollTracker.prototype.isVisible = function (top, bottom) {
    var scroll = window.scrollY || document.documentElement.scrollTop // IE fallback
    return scroll <= top && (scroll + this.windowHeight) >= bottom
  }
```


`pageHeight` for 100% scroll is this. note the `bottom` value:

```JavaScript
[
    {
        "alreadySeen": false,
        "top": 3993,
        "bottom": 3993,
        "eventData": {
            "type": "percent",
            "percent_scrolled": "100"
        }
    }
]
```

But `scroll + this.windowHeight` is `3992.5`, so the value from calculating the scroll position is `0.5` pixels away from firing. 

So I think either pageHeight needs to be reduced by 1, or the scroll value needs to be rounded up.

E.g. either

```JavaScript
var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight) - 1
```

or 

```JavaScript
var scroll = Math.ceil(window.scrollY || document.documentElement.scrollTop) // IE fallback
```


Thanks!

AshGDS:
It might be safer to both minus 1 from pageHeight and round up the scroll value, as I've encountered a scenario where the `bottom` for 100% is `3981` but the `scroll + this.windowHeight` value is `3979.888916015625`

AshGDS:
Thanks @andysellick - it's a lot more reliable now. I'll approve it but there's still some rare cases where it goes wrong.

I think using `ceil` in your last commit may still result in scenarios of not hitting 100%. For example I encountered a scenario (albeit they're very rare now, it seems to be very specific page widths/zooms) where scroll was `7601` but bottom was `7602` because bottom was getting rounded up. If it was rounded down they would both be `7601`. Annoyingly though, even when doing `Math.floor`, I zoomed out my browser in one scenario and that also results in a mismatch between `scroll` and `bottom`. I got `scroll: 7265 vs bottom: 7266`.

The main reason for all of this inconsistency is because `var pageHeight = Math.max(body.scrollHeight, body.offsetHeight, html.clientHeight, html.scrollHeight, html.offsetHeight) - 1 ` is trying 5 different ways to calculate the browser height. Each one factors different things into how it calculates height. See this: https://stackoverflow.com/revisions/46429828/3

Other height calculations such as `window.innerHeight` factor in different things too. For example, in the test that you need to subtract 15px from, you have to do that because `window.innerHeight` includes the horizontal scrollbar in its calculation, which is 15px high. It's possible that Jasmine has a horizontal scrollbar on its test page. So the horizontal scrollbar is being added to the `body` height when you do `setPageHeight`. Then the scroll tracker is grabbing the body height which has that added extra 15px on it, but is also grabbing the 15px scrollbar from the browser, resulting in a 15px mismatch.

And for the 100% scroll not firing issue, it seems that Math.max favours `body.scrollHeight`,  `body.offsetHeight` or `html.offsetHeight` when the 100% scroll calculation goes wrong. The problem is likely that these calculations don't include `margin` values in their height calculation. There is probably some `margin: [negative number]px` in our CSS at certain widths on certain pages which is making the scroll height of the page smaller by 1 or 2 pixels, but those other height calculations don't include margin so it doesn't get factored in. 

The safest thing to do might be to give us a lot more buffer for 100%, e.g. psuedocode

```
when making percentDetails for 100%:
      minus 10px from the 'bottom' value
```

AshGDS (approved):

AshGDS (approved):
Thanks @andysellick - this technique should be a lot more reliable :+1:

Issue #3525, opened 2023-07-27
Overflowing inputs over GOV.UK Search results


nnagewad:
## Issue

- There is an issue on GOV.UK Search where the input fields from the filter overflows on the page's search results

### Screenshot
![Screenshot 2023-07-27 at 15 28 06 (2)](https://github.com/alphagov/govuk_publishing_components/assets/8485154/23bca89d-19a5-4342-87ab-8a57ed32503d)

## Why it's an issue

- Although the chances of users experiencing this could be slim, we could handle this a bit better
- This could also qualify as an accessibility issue since the input field is obstructing the content below

PR #3538, opened 2023-08-03
Fix select width overlap bug

andysellick:
## What / why
Adjust the select component 'full width' styling to fix a problem.

- where the full width option is used in a confined space, the select element can extend outside of the width of the containing element
- noticed on [topic filters on search pages](https://www.gov.uk/search/all?keywords=minister&level_one_taxon=d6c2de5d-ef90-45d1-82d4-5f2438369eea&order=relevance)

Tested in Edge, IE11, Chrome, Firefox, Opera, Safari.

## Visual Changes

Before | After
------ | -------
Screenshot 2023-08-03 at 15 50 34 | Screenshot 2023-08-03 at 15 50 41

Fixes https://github.com/alphagov/govuk_publishing_components/issues/3525

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix select width overlap bug ([PR #3538](https://github.com/alphagov/govuk_publishing_components/pull/3538))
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Looks good to me 👍

PR #3540, opened 2023-08-07
Fix row spacing on topics browse pages

matthillco:
## What
The existing grid setup forced every row to be the same height as the tallest row. But this causes large gaps of whitespace if one content item is larger than all the others, and a peculiar behaviour for the first row where it would always collapse, leaving row heights inconsistent.

Removing `grid-template-rows: auto` prevents the first row collapsing and removes the whitespace anomalies for the other rows.

Note that this change also requires guidance to be provided for content authors to ensure that each content item is of similar length which helps keeps the row heights consistent.

[Trello Link](https://trello.com/c/z4SoPSlt/1883-spacing-bug-on-some-browse-pages)

## Why
After reviewing the usage of the grid, designers and content authors preferred to fix the presentation of rows to be consistently the same height., rather than the current behaviour.

## How

I've tested this using my local instance of the gem running in my local `collections` app and the changes look fine. 

### Before

image

### After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Redact GA params from pageview data ([PR #3568](https://github.com/alphagov/govuk_publishing_components/pull/3568))
 * Add GA4 tracking to devolved nations banners ([PR #3556](https://github.com/alphagov/govuk_publishing_components/pull/3556))
 * Add GA4 tracking to the cookie banner ([PR #3564](https://github.com/alphagov/govuk_publishing_components/pull/3564))
+* Fix row spacing on topics browse pages ([PR #3540](https://github.com/alphagov/govuk_publishing_components/pull/3540))
 
 ## 35.14.0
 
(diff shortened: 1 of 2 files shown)


hannalaakso (approved):
LGTM, thanks @matthillco  👍  I've done some cross browser testing in our usual browsers and didn't see any regressions. The fix also makes sense semantically as using [grid-template-rows](https://developer.mozilla.org/en-US/docs/Web/CSS/grid-template-rows) seems unnecessary here since we're not setting different row heights.

(For anyone else reviewing, note that the bug doesn't currently appear on live as our content designers removed the unusually tall content which caused the bug to surface; however, I've tested this fix by duplicating the [content as it was when the bug was reported](https://webarchive.nationalarchives.gov.uk/ukgwa/20230620154006/https://www.gov.uk/browse/housing-local-services) and can confirm that this PR fixes the original issue).

PR #3541, opened 2023-08-08
Adjust core functions setIndexes

andysellick:
## What
- allow an index_total to be set and collected, to override a calculated index_total, normally determined by the number of links found in the element
- used in a rare situation on taxon pages where we want to use the setIndexes functionality but also need to adjust the index_total to account for an extra link

## Why
See related PR.

## Visual Changes
None.

Trello card: https://trello.com/c/DqMNdDNz/652-change-type-of-explore-sub-topics-links-to-subtopic-list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


JamesCGDS (approved):
Looks good 👍

PR #3544, opened 2023-08-10
Add a reset option into the GA4 scrolltracker

andysellick:
## What
- code adds an event listener for 'dynamic-page-update' event, which resets all of the nodes to alreadySeen = false
- this event can then be triggered by code elsewhere, in order to 'reset' scroll tracking on any given page
- initial use is for search pages, where users change their search filters, page dynamically updates and so scroll tracking only works the first time, now if the search code triggers that event, the scroll tracker will reset

## Why
Scroll tracking works correctly on search pages until users change their filters. At that point the user is effectively on a new page (content and URL change) and we want the scroll tracking to 'reset' (i.e. if the user has scrolled to the bottom of the page previously, we want to also record that happened after the search filters are changed). Unfortunately because the page was updated dynamically using JS, the scroll tracker still thinks its on the same page, and therefore doesn't register GA4 events when users scroll. This PR fixes that - a related PR will be raised in `finder-frontend` to add to required event trigger.

## Visual Changes
None.

Trello card: https://trello.com/c/fhnLqnKw/669-scroll-tracker-to-refresh-on-dynamic-pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add a reset option into the GA4 scrolltracker ([PR #3544](https://github.com/alphagov/govuk_publishing_components/pull/3544))
 * Adjust core functions setIndexes ([PR #3541](https://github.com/alphagov/govuk_publishing_components/pull/3541))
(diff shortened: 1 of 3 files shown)


JamesCGDS (approved):
🚀

PR #3546, opened 2023-08-11
Expand GA4 form tracker

andysellick:
## What
- when handling text inputs, form tracker didn't know about some input type values that it is required to recognise, specifically 'email' and 'number'
- other input types exist but are currently not in use on GOV.UK, but can now be more easily added if required

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/TXQOqtZ0/628-email-subscriptions-manage-subscriptions-sign-in-change-unsubscribe-forms

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
@@ -84,12 +84,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var labelText = input.label.innerText || input.label.textContent
       var inputType = elem.getAttribute('type')
       var inputNodename = elem.nodeName
+      var inputTypes = ['text', 'search', 'email', 'number']
 
       if (inputType === 'checkbox' && elem.checked) {
         input.answer = labelText
       } else if (inputNodename === 'SELECT' && elem.options[elem.selectedIndex].value) {
         input.answer = elem.options[elem.selectedIndex].text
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -84,12 +84,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
       var inputType = elem.getAttribute('type')
       var inputNodename = elem.nodeName
+      var inputTypes = ['text', 'search', 'email', 'number']
 
       if (inputType === 'checkbox' && elem.checked) {
         input.answer = labelText
       } else if (inputNodename === 'SELECT' && elem.options[elem.selectedIndex].value) {
         input.answer = elem.options[elem.selectedIndex].text
-      } else if ((inputType === 'text' || inputType === 'search') && elem.value) {
+      } else if (inputTypes.includes(inputType) && elem.value) {
```
JamesCGDS, line 93: Quick question - instead of `inputTypes.includes(inputType)`, would we want to use `inputTypes.indexOf(inputType) !== -1`? Only because I've seen it elsewhere in our GA4 code and I guess would be beneficial if we want to gather data from any remaining IE11 users.

andysellick: Great shout, thanks @JamesCGDS, have updated 👍

JamesCGDS (approved):
Looks good! 🚀

PR #3549, opened 2023-08-14
Add GA4 tracking to the emergency banner

AshGDS:
## What

- Adds GA4 tracking to the emergency banner component.
- More specifically, it adds `emergency_banner: true` to the `pageview` if the emergency banner is rendered. It also adds the `ga4-link-tracker`. Both of these rely on `ga4_tracking: true` so we can toggle the tracking on and off on the banner.
- I've checked it in a local version of `static`, the `pageview` attribute is added correctly.

## Why

- https://trello.com/c/yH88Twnv/661-banners-emergency-banner-element-visibility-and-navigation

## Visual Changes

- It adds a "With GA4 tracking" to the component examples

### DIFF
--- a/app/views/govuk_publishing_components/components/_emergency_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_emergency_banner.html.erb
@@ -7,6 +7,7 @@
   link_text ||= "More information"
   campaign_class ||= nil
   homepage ||= false
+  ga4_tracking ||= false
 
   emergency_banner_helper = GovukPublishingComponents::Presenters::EmergencyBannerHelper.new()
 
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/docs/emergency_banner.yml
```diff
@@ -45,3 +45,13 @@ examples:
       short_description: "1491 to 1547"
       link: "https://www.gov.uk/"
       homepage: true
+  with_ga4_tracking:
+    description: |
+      Toggles emergency_banner: true on the GA4 pageview object, and enables the GA4 link tracker on the banner.
```
andysellick, line 50: Nitpick: I'd maybe reword this as `Enables GA4 tracking on the banner. This includes link tracking on the component itself, and allows pageviews to record the presence of the banner on page load.`

app/views/govuk_publishing_components/components/_emergency_banner.html.erb
```diff
@@ -7,6 +7,7 @@
   link_text ||= "More information"
   campaign_class ||= nil
   homepage ||= false
+  ga4_tracking ||= false
```
andysellick, line 10: Presumably there'll be a separate PR in `static` to set this to `true`?

andysellick (commented):
Looks good, just two very small things.

app/views/govuk_publishing_components/components/_emergency_banner.html.erb
AshGDS: Yep, I'll do this as part of `To deploy` on the card.

AshGDS:
Thanks @andysellick - I've made the requested changes.

andysellick (approved):

PR #3552, opened 2023-08-15
Add GA4 tracking to the phase banner

AshGDS:
## What

- Adds GA4 tracking to the phase banner component.
- More specifically, it adds `phase_banner: alpha/beta/undefined` to the `pageview` if the phase banner is rendered. It grabs `alpha` `beta`  based on the value of the `data-ga4-phase-banner` attribute, for example: `data-ga4-phase-banner=beta`. This PR also adds the `ga4-link-tracker` to the banner. 
- Both of these rely on `ga4_tracking: true` so we can toggle the tracking on and off on the banner.
- I've checked it in a local version of `static`. The `pageview` attribute is added correctly fortunately (there was speculation in the implementation guide that this wouldn't be possible.) I believe this is working because our pageview fires after DOMContentLoaded so it doesn't matter where a banner is in the DOM.

## Why

- https://trello.com/c/d6CHHgco/663-banners-phase-banner-element-visibility-and-navigation

## Visual Changes

- It adds a "With GA4 tracking" to the component examples

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -54,7 +54,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
             and therefore we can use it to set the dynamic property appropriately. This value is used by PA's to differentiate
             between fresh page loads and dynamic page updates. */
             dynamic: referrer ? 'true' : 'false',
-            emergency_banner: document.querySelector('[data-ga4-emergency-banner]') ? 'true' : undefined
+            emergency_banner: document.querySelector('[data-ga4-emergency-banner]') ? 'true' : undefined,
+            phase_banner: this.getElementAttribute('data-ga4-phase-banner') || undefined
           }
         }
         window.GOVUK.analyticsGa4.core.sendData(data)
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -92,6 +93,10 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       }
     },
 
+    getElementAttribute: function (attributeName) {
+      return document.querySelector('[' + attributeName + ']').getAttribute(attributeName)
```
andysellick, line 97: I wonder if moving more of the logic above into this function would make this a little clearer, something like this:

```JavaScript
phase_banner: this.getElementAttribute('data-ga4-phase-banner') || undefined
```

```JavaScript
getElementAttribute: function (attributeName) {
  var el = document.querySelector('[' + attributeName + ']')
  if (el) {
    return el.getAttribute(attributeName)
  }
},
```

andysellick (approved):
Looks good, just one small suggestion.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
AshGDS: @andysellick Good idea - I've added this, thanks.

PR #3564, opened 2023-08-18
Add GA4 tracking to the cookie banner

AshGDS:
## What

- Adds GA4 tracking to the cookie banner component, but only on the "You have accepted cookies" HTML.
- More specifically, it adds `cookie_banner: 'true'/undefined` to the `pageview` if the "You have accepted cookies" section of the banner is rendered. It grabs this value based on the presence of `data-ga4-cookie-banner` in the DOM. The GA4 Pageview JS should fire at the same time that this message is displayed as they've just accepted cookies, so our analytics should load, trigger a pageview, and track the visibility of the cookie banner.
- The `ga4-link-tracker` is also added to this "cookies accepted" section.
- Both of these rely on `ga4_tracking: true` so we can toggle tracking on and off on the banner.
- I also fixed a bug with the examples in the component guide. Because the cookie banner used `document.querySelector`, the banners would break if more than one banner existed. This is because they would search for elements using the whole DOM. Therefore only the first banner in the component example would work, as document.querySelector always returns the first instance of an element. I changed it to use `this.$module.querySelector` instead so that the element searching is only within the module, which probably fits our module pattern anyway since the JS is now self contained.
- I've checked it in a local version of `static` and the pageview is sending correctly.

## Why

- https://trello.com/c/bfDEtTyR/665-banners-cookie-banner-navigation-and-hide-this-message-user-interaction

## Visual Changes

- Added a `with_ga4_tracking` example to the docs for the component.


app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -39,7 +41,14 @@
         <h2 class="govuk-cookie-banner__heading govuk-heading-m"><%= title %></h2>
         <div tabindex="-1" class="govuk-cookie-banner__content gem-c-cookie-banner__confirmation">
           <span class="gem-c-cookie-banner__content"><%= text %></span>
-          <p class="gem-c-cookie-banner__confirmation-message--accepted govuk-body" hidden><%= t("components.cookie_banner.confirmation_message.accepted") %>. <span class="gem-c-cookie-banner__confirmation-message"><%= confirmation_message %></span></p>
+          <p class="gem-c-cookie-banner__confirmation-message--accepted govuk-body" hidden
+          <% if ga4_tracking %>
+            data-ga4-cookie-banner
```
andysellick, line 46: Maybe worth a short comment here to explain what this attribute is for? e.g. `# page view JS looks for data-ga4-cookie-banner`

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
andysellick, line 1: Interested by the commit comment - are you saying the component guide for the cookie banner has been broken since before now?

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -94,7 +94,18 @@
...
+        button_data_module << " ga4-event-tracker" if ga4_tracking
+
+        ga4_event = {
+          event_name: "select_content",
+          type: "cookie banner",
+          action: "closed",
+          section: t("components.cookie_banner.confirmation_message.accepted", locale: :en)
+        }.to_json if ga4_tracking
+      %>
+      <button class="gem-c-cookie-banner__hide-button govuk-button" data-hide-cookie-banner="true" data-module="<%= button_data_module %>" data-track-category="cookieBanner" data-track-action="Hide cookie banner" <% if ga4_tracking %> data-ga4-event="<%= ga4_event %>" <% end %>><%= t("components.cookie_banner.hide") %></button>
```
andysellick, line 108: Maybe space this out for clarity?

```html
<button 
  class="gem-c-cookie-banner__hide-button govuk-button" 
  data-hide-cookie-banner="true" 
  data-module="<%= button_data_module %>" 
  data-track-category="cookieBanner" 
  data-track-action="Hide cookie banner" 
  <% if ga4_tracking %> data-ga4-event="<%= ga4_event %>" <% end %>>
    <%= t("components.cookie_banner.hide") %>
</button>
```

andysellick (commented):
Looks good, just a few comments/questions.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
AshGDS: @andysellick Yep - if you look on the live site: https://components.publishing.service.gov.uk/component-guide/cookie_banner

Try and accept cookies on the second cookie banner. It won't manipulate the DOM properly as it ends up interacting with the first cookie banner example. So only the first example works in the component guide at the moment.

AshGDS:
Thanks @andysellick - I've made the requested changes :+1:

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
andysellick: Wow that must have been there for a while. Great fix 👍

andysellick (approved):
Nice - squash the related commits together and we're good to go 👍

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
AshGDS: Thanks @andysellick

PR #3566, opened 2023-08-22
Add homepage variant of navbar for new design of homepage

patrickpatrickpatrick:
## What

This commit contains several small changes that are part of the new design of the homepage. They are all optional parameters for layout_super_navigation_header and are the following:

* toggle the logo text off
* toggle the border left of the toggle button off
* change the background colour of the navbar to blue
* make the links/buttons larger for desktop only

Created a new partial `layout_super_navigation_header_home` which has all these changes toggled on. Also toggles some additional classes on elements when the navbar is blue to make sure the contrast between hover state styles is OK.

## Why

As part of several small changes that are going to be made to the homepage, we need to override some default styling of the layout super navigation header. As these variations will only be rendered on the homepage, a new partial `layout_super_navigation_header_home` has been created. This partial will render the specific homepage differences and it will mean that the `layout_for_public` will only need one variable to change between homepage styled layout super navigation header and the regular styled layout super navigation header.

## Screenshot

Screenshot 2023-09-29 at 16 06 07

[Relevant Trello Link](https://trello.com/c/qTmNgPwV/2019-look-and-feel-homepage-top-menu-navigation-header-m)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -8,6 +8,10 @@ $search-width-or-height: $black-bar-height;
 $pseudo-underline-height: 3px;
 $button-pipe-colour: darken(govuk-colour("mid-grey"), 20%);
 
+$large-navbar-height: 86px;
+
+$pale-blue-colour: #d2e2f1;
+
 $after-link-padding: govuk-spacing(4);
 $after-button-padding-right: govuk-spacing(4);
 $after-button-padding-left: govuk-spacing(4);
@@ -89,6 +93,11 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+.gem-c-layout-super-navigation-header--blue-background {
+  background: $govuk-brand-colour;
+  border-top: 1px solid $govuk-brand-colour;
+}
+
 .gem-c-layout-super-navigation-header__container {
   position: relative;
 }
@@ -128,7 +137,6 @@ $after-button-padding-left: govuk-spacing(4);
 
 .gem-c-layout-super-navigation-header__navigation-item,
 .gem-c-layout-super-navigation-header__search-item {
-  background: govuk-colour("black");
   display: block;
   float: left;
   margin: 0;
@@ -205,6 +213,17 @@ $after-button-padding-left: govuk-spacing(4);
       }
     }
 
+    &.gem-c-layout-super-navigation-header__navigation-item-link--blue-background,
+    &.gem-c-layout-super-navigation-header__search-item-link--blue-background {
+      &:hover {
+        color: govuk-colour("white");
+
+        &::after {
+          background: govuk-colour("white");
+        }
+      }
+    }
+
     @include focus-and-focus-visible {
       .gem-c-layout-super-navigation-header__navigation-item-link-inner {
         border-color: $govuk-focus-colour;
@@ -267,6 +286,17 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+// to stop the black underline appearing underneath the menu
+// after it has been toggled closed if the blue background has
+// been enabled
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background {
+  &:focus:not(:focus-visible) {
+    &::after {
+      background: none;
+    }
+  }
+}
+
 .gem-c-layout-super-navigation-header__navigation-item-link {
   @include govuk-media-query($from: "desktop") {
     padding: govuk-spacing(3) 0;
@@ -279,9 +309,13 @@ $after-button-padding-left: govuk-spacing(4);
 }
 
 .gem-c-layout-super-navigation-header__navigation-item-link-inner {
-  background-color: govuk-colour("black");
   border-left: 1px solid $button-pipe-colour;
   padding: govuk-spacing(1) $after-link-padding;
+}
+
+.gem-c-layout-super-navigation-header__navigation-item-link-inner--blue-background {
+  border-left: 0;
+  border-right: 1px solid $pale-blue-colour;
 }
 
 // Search link and dropdown.
@@ -297,8 +331,6 @@ $after-button-padding-left: govuk-spacing(4);
     background: $govuk-brand-colour;
 
     &:hover {
-      background: govuk-colour("black");
-
       &::before {
         left: 0;
         right: 0;
@@ -321,10 +353,6 @@ $after-button-padding-left: govuk-spacing(4);
 
     @include focus-not-focus-visible {
       background: $govuk-link-colour;
-
-      &:hover {
-        background: govuk-colour("black");
-      }
     }
 
     @include focus-and-focus-visible {
@@ -345,6 +373,10 @@ $after-button-padding-left: govuk-spacing(4);
   height: 20px;
   pointer-events: none;
   width: 21px;
+}
+
+.gem-c-layout-super-navigation-header__open-button .gem-c-layout-super-navigation-header__search-toggle-button-link-icon {
+  display: none;
 }
 
 // Search and popular content dropdown.
@@ -472,15 +504,23 @@ $after-button-padding-left: govuk-spacing(4);
     .gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner {
       border-color: $button-pipe-colour;
 
+      &.gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner--blue-background {
+        border-color: $pale-blue-colour;
+      }
+
       @include govuk-media-query($from: 360px) {
         &::before {
           @include chevron(govuk-colour("white"), true);
         }
       }
     }
+
     // stylelint-enable max-nesting-depth
   }
+}
 
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background,
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button {
   // Open button modifier
   &.gem-c-layout-super-navigation-header__open-button {
     // stylelint-disable max-nesting-depth
@@ -528,6 +568,24 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background {
+  @include focus-not-focus-visible {
+    &:hover {
+      &::after {
+        background: govuk-colour("white");
+      }
+
+      .gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner {
+        color: govuk-colour("white");
+
+        &::before {
+          border-color: govuk-colour("white");
+        }
+      }
+    }
+  }
+}
+
 .gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner {
   display: inline-block;
   border-left: 1px solid govuk-colour("white");
@@ -540,6 +598,10 @@ $after-button-padding-left: govuk-spacing(4);
       @include chevron(govuk-colour("white"));
     }
   }
+}
+
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button-inner--no-left-border {
+  border-left: 0;
 }
 
 // Styles for search toggle button.
@@ -580,9 +642,30 @@ $after-button-padding-left: govuk-spacing(4);
     }
 
     &:hover {
-      background: govuk-colour("black");
       border-bottom: $pseudo-underline-height solid govuk-colour("mid-grey");
       color: govuk-colour("mid-grey");
+    }
+
+    &.gem-c-layout-super-navigation-header__search-toggle-button--blue-background {
+      &:hover {
+        border-bottom: $pseudo-underline-height solid govuk-colour("white");
+        color: govuk-colour("white");
+      }
+
+      // stylelint-disable max-nesting-depth
+      @include focus-not-focus-visible {
+        border-bottom: none;
+      }
+      // stylelint-enable max-nesting-depth
+    }
+
+    &:focus-visible {
+      &.gem-c-layout-super-navigation-header__search-toggle-button--blue-background {
+        &:hover {
+          border-bottom: $pseudo-underline-height solid $govuk-focus-colour;
+          color: govuk-colour("black");
+        }
+      }
     }
 
     @include focus-and-focus-visible {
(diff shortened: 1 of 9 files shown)


app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -17,8 +22,41 @@
...
+  search_item_link_classes << "gem-c-layout-super-navigation-header__search-item-link--blue-background" if blue_background
 %>
-<header role="banner" class="gem-c-layout-super-navigation-header" data-module="gem-track-click ga4-event-tracker ga4-link-tracker" data-track-links-only data-ga4-expandable>
+<%= content_tag(:header, {
+    role: "banner",
+    class: layout_super_navigation_header_classes,
+    data: {
+      module: "gem-track-click ga4-event-tracker ga4-link-tracker",
+      "track-links-only": true,
+      "ga4-expandable": true,
```
hannalaakso, line 67: `data-track-links-only` and `data-ga4-expandable` don't have a value set on the current homepage navbar, whereas these will set the values to `true`. It might be worth checking that this doesn't change the tracking behaviour.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -267,6 +284,14 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background {
+  &:focus:not(:focus-visible) {
```
hannalaakso, line 298: Not a blocking but would it be possible to leave a comment here explaining what scenario this bit of code is for? 🙏

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -267,6 +284,14 @@ $after-button-padding-left: govuk-spacing(4);
...
+    &::after {
```
hannalaakso, line 299: There are a few places in the PR where we nest things quite deeply. I'm wondering if it might be possible to write the out the full class instead? The Design System coding standards which we mostly try to follow in the gem discourage the use of the ambersand and deep nesting: https://github.com/alphagov/govuk-frontend/blob/main/docs/contributing/coding-standards/css.md#nesting It might not be possible to do that in all the places but I think it would be good to do where possible to make the code easier to read.

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -68,9 +106,11 @@
             <!--[if IE 8]>
             <img src="<%= asset_path('govuk-logotype-crown.png') %>" alt="" class="govuk-header__logotype-crown-fallback-image" width="36" height="32">
             <![endif]-->
-            <span class="govuk-header__logotype-text">
-              <%= logo_text %>
-            </span>
+            <% unless hide_logo_text %>
```
hannalaakso, line 121: Since we're removing the link text, I think we should add an `aria-label` to the link. It's got a `title` but I think `aria-label` tests better for accessibility eg. https://www.deque.com/blog/text-links-practices-screen-readers/. Worth researching at least. Another alternative might be including visually hidden link text.

hannalaakso (commented):
This is looking good 👍 I haven't done a thorough review of the Sass yet as I know it will change with the focus/hover states that are being worked on. I'll also do some more browser/device testing once those states have been finalised.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
patrickpatrickpatrick: There is so much nesting going on already in this .scss that I think it's a whole other issue to refactor the entire stylesheet to get into line with the Design System coding standards. 

Personally I'm also not sure if it would make it easier to read as this line would then be:

`.gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background:focus:not(:focus-visible)::after {`

which I myself don't find easier to read, I think the nesting makes more sense here.

patrickpatrickpatrick:
Am aware that the extra stylesheets will not be picked up by the component auditor. For now these styles are going to be used on one page though. If they are going to be rolled out pages then that would be a good time to see about implementing this functionality into the auditor.

patrickpatrickpatrick:
Going to discuss how to load the css today with Martin, I think it is worth having some mechanism to separately load it as it is a bespoke change to a globally used component and ideally we wouldn't want to include that CSS on all pages.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
hannalaakso: The gem docs address the issue of the long nested class names [here](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md#bem) with:
> To avoid long and complicated class names, we follow the [BEM guidance](http://getbem.com/faq/#css-nested-elements) that classes do not have to reflect the nested nature of the DOM.

But it's true that there is a lot of existing nesting in this component already so maybe it's not worth trying to overhaul the standard in this PR.

patrickpatrickpatrick: I think if we were to do it in this instance then we would/should do it for all instances and it is 1000 lines of css. It's probably a task for when we don't have an imminent deadline.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -472,17 +514,40 @@ $after-button-padding-left: govuk-spacing(4);
...
+
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button--large-navbar {
+  height: $large-navbar-height;
+}
 
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background,
+.gem-c-layout-super-navigation-header__navigation-top-toggle-button {
   // Open button modifier
   &.gem-c-layout-super-navigation-header__open-button {
+    &::before {
```
hannalaakso, line 540: There's a white block caused by this pseudo element when the button is focused and the menu is open. Not sure if that's part of the design?  
Screenshot 2023-10-03 at 14 10 51

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -618,16 +719,23 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+.gem-c-layout-super-navigation-header__search-toggle-button--large-navbar {
+  height: $large-navbar-height;
+}
+
+// to stop the search icon moving on hover
```
hannalaakso, line 726: I'm still seeing a slight "jump" when I hover over the search icon and the close X icon. This is on Chrome desktop 117.0.5938.132 and Mac.

hannalaakso (commented):
Thanks for addressing my comments @patrickpatrickpatrick! 🙏 I still need to do a bit more reviewing and testing on devices (including checking how the new `aria-label` behaves with screen readers).

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -17,63 +23,123 @@
...
+          id: "logo",
+          title: logo_link_title,
+        } do %>
           <span class="govuk-header__logotype">
             <!--[if gt IE 8]><!-->
-            <svg
-              aria-hidden="true"
-              class="govuk-header__logotype-crown gem-c-layout-super-navigation-header__logotype-crown"
-              focusable="false"
-              height="30"
```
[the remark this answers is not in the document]
patrickpatrickpatrick: Have added this back to SVG

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -788,3 +896,31 @@ $after-button-padding-left: govuk-spacing(4);
...
+// isn't a multiple of 5 :(
+.gem-c-layout-super-navigation-header__navigation-item-link--large-navbar,
+.gem-c-layout-super-navigation-header__search-item-link--large-navbar {
+  padding-top: 33px;
+  padding-bottom: 33px;
+}
+
+.gem-c-layout-super-navigation-header__logotype-crown {
+  height: 30px;
+  width: 36px;
```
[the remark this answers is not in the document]
patrickpatrickpatrick: Have set them on the SVG itself

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
patrickpatrickpatrick: Have removed this

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -574,17 +650,31 @@ $after-button-padding-left: govuk-spacing(4);
...
+      &:hover {
+        border-bottom: $pseudo-underline-height solid govuk-colour("white");
+        color: govuk-colour("white");
+      }
+    }
+
+    &:focus-visible {
+      &.gem-c-layout-super-navigation-header__search-toggle-button--blue-background {
+        &:hover {
+          border-bottom: $pseudo-underline-height solid $govuk-focus-colour;
```
hannalaakso, line 672: The focus state overhangs a bit on mobile, this is on Android 12 on Galaxy S23. The focus isn't visible much on touch screens but probably worth tweaking still.
Screenshot 2023-10-03 at 19 33 48

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -17,63 +23,125 @@
...
-              viewBox="0 0 132 97"
-              width="36"
-              xmlns="http://www.w3.org/2000/svg"
-            >
+            <%= content_tag(:svg, {
+              aria: {
+                hidden: '',
```
hannalaakso, line 117: I think that this needs to be set to `true`. The HTML validator says `Error: Bad value for attribute aria-hidden on element svg.`

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
hannalaakso: This works well on NVDA and JAWS on Mac desktop but unfortunately Mac Voiceover doesn't read the label and skips the link entirely 😞 I think we should go with visually hidden text instead.

patrickpatrickpatrick:
https://github.com/alphagov/govuk_publishing_components/pull/3566#discussion_r1344416223

(I can't reply to this for some reason)

We need to override the height and width in CSS for large navbar styles. If the CSS styles aren't applied in a small subset of browsers, the page is still very usable and doesn't have any adverse effects (because we define the 'back-up' sizes on the svg itself). So I would say I am willing to take the risk.

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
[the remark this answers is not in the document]
patrickpatrickpatrick: We need to override the height and width in CSS for large navbar styles on Desktop only. If the CSS styles aren't applied in a small subset of browsers, the page is still very usable and doesn't have any adverse effects (because we define the 'back-up' sizes on the svg itself). So I would say I am willing to take the risk.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -574,17 +650,31 @@ $after-button-padding-left: govuk-spacing(4);
...
 
     &:hover {
-      background: govuk-colour("black");
       border-bottom: $pseudo-underline-height solid govuk-colour("mid-grey");
       color: govuk-colour("mid-grey");
     }
 
+    &.gem-c-layout-super-navigation-header__search-toggle-button--blue-background {
+      &:hover {
+        border-bottom: $pseudo-underline-height solid govuk-colour("white");
```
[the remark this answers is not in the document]
patrickpatrickpatrick: should be fixed now

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
patrickpatrickpatrick: this shouldn't be an issue now (large styles not on mobile)

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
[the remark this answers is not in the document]
hannalaakso: Okay, that sounds like a sensible assessment with regards to what the risk is 👍  And it feels a bit more semantic too not to change the size in the markup since we're not changing the SVG size on all viewports.

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
hannalaakso: Just realised that Mac Safari skips the link too (with or without Voiceover enabled).

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -17,63 +23,125 @@
...
-            "index_total": 1
-          }.to_json %>"
-        >
+      <%= content_tag(:div, {
+        class: header_logo_classes
+      }) do %>
+        <%= link_to logo_link, {
+          class: header_link_classes,
+          aria: {
+            label: hide_logo_text ? logo_link_title : nil
```
hannalaakso, line 88: (Follow on from https://github.com/alphagov/govuk_publishing_components/pull/3566#discussion_r1340534201)

The aria-label works well on NVDA and JAWS on Mac desktop but unfortunately Mac Voiceover doesn't read the label and skips the link entirely 😞 This is also happens just in Mac Safari without VoiceOver enabled, the link is just skipped. I think we try to go with visually hidden text instead to make Safari announced the link text.

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
hannalaakso: (Continued conversation in https://github.com/alphagov/govuk_publishing_components/pull/3566/files#r1345700075 as this current conversation was getting lost as it's so old)

app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
hannalaakso: Thanks @patrickpatrickpatrick for making the change. I've just tested it on Safari and tabbing still doesn't navigate to the lin but I've now realised Mac Safari doesn't navigate to it even on the current navigation. So t's in fact an existing issue! We can ask Anika about it next week. 

Would you mind adding back the aria-label since that tested well with JAWS and NVDA? 🙏 We should also remove the `title` from the link since it defeats the point of `aria-label.

hannalaakso (approved):
Thanks for addressing all my comments @patrickpatrickpatrick! This is really nice work and it's good to have test coverage for it too. Exciting to see how it tests with design and users  ✨

PR #3568, opened 2023-08-22
Redact GA params from pageview data

andysellick:
## What
- replaces Google Analytics tracking parameters in URLs with `_ga=[id]` or `_gl=[id]` when creating GA pageviews
- has to happen prior to PII being stripped as they occasionally get mistaken for dates, which would break this redaction
- vaguely related: refactor the get referrer code to make it easier to read and insert the GA param redaction

## Why
To clean up the location data in GA4.

## Visual Changes
None.

Trello card: https://trello.com/c/2hJRUoDM/660-redact-strip-ga-and-gl-trackers-from-page-location-and-page-referrer-values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Redact GA params from pageview data ([PR #3568](https://github.com/alphagov/govuk_publishing_components/pull/3568))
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
LGTM :+1:

PR #3577, opened 2023-08-30
Fix GA4 bug - all attachments links tracked with the same JSON

AshGDS:
## What

- Previously, we had our GA4 JSON attribute on the parent element of the attachment links (the `<section>` tag.)
- This meant that every link within this element would be tracked with the same attributes listed in the JSON.
- We don't want to do this anymore, as we realised there are links that we don't want tracked in this way (e.g. email links can be nested in the attachments.)
- Therefore, we move the `ga4_link` JSON onto the `data_attributes` hash. This hash is used by the attachment thumbnail link, and the attachment title link. And this is good as we only want the tracking on those links.

http://0.0.0.0:3212/component-guide/attachment

## Why

- https://trello.com/c/MHFXw6H2/552-change-filedownloads-on-attachment-pages-to-have-attachment-instead-of-generic-download-type

## Visual Changes
None.


JamesCGDS (approved):
LGTM 👍

PR #3582, opened 2023-09-01
Add new rule to fix accordion print style

patrickpatrickpatrick:
## What

Add new rule to override content-visibility in print styles.

## Why

The accordions should be expanded when printed. On browsers that support content visibility, the content was still hidden. The addition of this rule fixes this and ensures that printed pages will show all accordion sections.

### Before

Screenshot 2023-09-01 at 12 38 39

### After

Screenshot 2023-09-01 at 12 38 33

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add new rule to fix accordion print style ([PR #3582](https://github.com/alphagov/govuk_publishing_components/pull/3582))
+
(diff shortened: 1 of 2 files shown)


querkmachine:
Copied over from a Slack discussion, in case it was missed there: 

> To answer from the Design System team perspective, we consider this [not expanding accordions when printing] correct and is an intentional choice.
> 
> We maintain the existing state of accordions/tabs/details so that users have control over what will be printed and aren’t forced to waste paper and ink printing content they don’t want or need.

maxgds (approved):
Thanks for looking at this!

PR #3590, opened 2023-09-05
Prepend page path when tracking anchor links in GA4

AshGDS:
## What

- Ensures the page path is added to anchor links tracked by the `ga4-link-tracker` and `ga4-event-tracker`
- For example, `#england-and-wales` will become `/bank-holidays#england-and-wales` in the GA4 dashboard.
- I also cleaned up the link tracker slightly as we kept repeating `window.GOVUK.analyticsGa4.core.trackFunctions`, let me know your thoughts.
- Will add CHANGELOG when the pending release PR is live

## Why

https://trello.com/c/7thzJTbm/672-include-page-path-in-anchor-links

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -205,6 +205,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
         }
       },
 
+      getPathname: function () {
+        return window.location.pathname
+      },
+
       getProtocol: function () {
         return window.location.protocol
       },
@@ -252,6 +256,13 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
 
       applyRedactionIfRequired: function (PIIRemover, element, data) {
         return element.closest('[data-ga4-do-not-redact]') ? data : PIIRemover.stripPIIWithOverride(data, true, true)
+      },
+
+      appendPathToAnchorLinks: function (url) {
+        if (!this.stringStartsWith(url, '#') || this.getPathname() === '/') {
+          return url
+        }
+        return this.getPathname() + url
       }
     },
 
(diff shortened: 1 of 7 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -205,6 +205,10 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
         }
       },
 
+      getPathname: function () {
+        return window.location.pathname
+      },
+
```
AshGDS, line 211: This is wrapped in a function so that it can be mocked in testing

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -62,13 +62,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   Ga4LinkTracker.prototype.trackClick = function (event) {
     var element = event.target
-
+    var trackFunctions = window.GOVUK.analyticsGa4.core.trackFunctions
```
andysellick, line 65: I'm in two minds about this. It does make the code cleaner and easier to read, but at the same time it slightly obfuscates what is being called by this variable - you have to read this bit of the code to know what `trackFunctions` refers to (although it is named the same thing, so that's good).

It might make a difference to the asset size on the public site, but only fractionally, and it's possible any saving would be negated by gzip/minifying anyway.

In conclusion... I don't know! It's probably fine 😁

andysellick (approved):
Seems good. Don't forget to include a changelog entry.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: Thanks @andysellick - I've asked another dev and they liked the refactor, so will keep it for now :+1:

PR #3592, opened 2023-09-06
Fix error in LUX script

andysellick:
## What
Fix an error in the live real user monitoring script relating to the unusual way in which we use this script.

- to reproduce, clear cookies, reload page then accept cookies, error of `Uncaught TypeError: Cannot read properties of null (reading 'pop')` will occur
- this refers to the `getCustomerId()` function, which if `LUX.customerid` is not defined will attempt to determine it from the source URL of the LUX script, something like `lux.js?id=1234`
- we don't get our customerid this way, instead we hard code it into this file, because we host our own copy of the LUX script (see the documentation for more detail and background on this), so the script failed, because there was nothing to find
- instead we remove this part of the `getCustomerId()` function and simply return `LUX.customerid`. We also now need to set this inside the LUX object itself, because otherwise it isn't set properly in the steps described above
- documentation updated to reflect this for future updates

## Why
Was causing an error on the live site. This is bad.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix error in LUX script ([PR #3592](https://github.com/alphagov/govuk_publishing_components/pull/3592))
 * Prepend page path when tracking anchor links in GA4 ([PR #3590](https://github.com/alphagov/govuk_publishing_components/pull/3590))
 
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Looks good on my local static :+1:

PR #3594, opened 2023-09-07
Fix bug with GA4 search results count

AshGDS:
## What

The results count heading contained new line characters and extra spaces, so splitting via the space character to parse the integer broke. This was fixed by using our `removeLinesAndExtraSpaces` function. 

I checked it on a local version of finder-frontend and the search results count is coming through now.

## Why

https://trello.com/c/MXhpmk9c/659-fix-search-results-count-on-viewitemlist-event

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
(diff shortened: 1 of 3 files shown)


JamesCGDS (approved):
Looks good! 👍

PR #3596, opened 2023-09-08
Add large dark icon for action link

1pretz1:
## What

Uses the existing dark icon but makes it slightly larger as per the
design recommendation for the new benefits calculator link being added
to the browse benefit page.

Related PR: https://github.com/alphagov/collections/pull/3373

## Why

Current dark icon is too small and doesn't fit the recommended design.

## Visual Changes

Large dark icon used for the new action link:

Screenshot 2023-09-13 at 10 39 04
Screenshot 2023-09-13 at 10 39 39

Trello:
https://trello.com/c/mwQET0e0/2153-add-link-to-the-benefits-smart-answer-to-the-browse-benefits-page

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
@@ -150,6 +150,22 @@
   }
 }
 
+.gem-c-action-link--dark-large-icon {
+  &:before {
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.png");
+    background: image-url("govuk_publishing_components/action-link-arrow--dark.svg"), linear-gradient(transparent, transparent);
+    height: 34px;
+    width: 40px;
+    background-repeat: no-repeat;
+    background-size: 32px auto;
+    background-position: 0 2px;
+  }
+
+  @include govuk-media-query($until: tablet) {
+    margin-bottom: govuk-spacing(2);
+  }
+}
+
 .gem-c-action-link--dark-icon,
 .gem-c-action-link--small-icon {
   max-width: none;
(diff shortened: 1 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -150,6 +150,24 @@
   }
 }
 
+.gem-c-action-link--dark-large-icon {
+  max-width: none;
+
```
jon-kirwan, line 155: I'm not quite sure why `max-width: none;` is needed @MartinJJones - any idea?

jon-kirwan (commented):
All looks fine to me. I've added one comment below with regards to a style that might not be needed but have asked for a second opinion.

app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
MartinJJones: It appears the component sets a `max-width` of `410px` for screen sizes up to `$tablet` (641px).

`gem-c-action-link` - https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss#L19

When testing locally, this would not make a difference given the current length of text, but if more text was added in the future it would have a slight impact, for example:

### Screen width 640px

#### max-width set to none
Screenshot 2023-09-15 at 10 44 43

#### max-width left as default - 410px
Screenshot 2023-09-15 at 10 44 21

So I think this adding or removing it would come down to preference of how it should appear on tablet when we have a lot of text.

1pretz1: Thanks for the comment Jon and for digging into it Martin! I've removed as from your investigation, it only impacts longer text than we're currently proposing so not value in keeping `max-width`

1pretz1:
Addressed the `max-width` comment @jon-kirwan @MartinJJones (thanks for pointing it out and investigating)- could this be re-reviewed please? Thank you!

MartinJJones (approved):
Looks good to me 👍

PR #3598, opened 2023-09-11
Check HTML attribute when populating GA4 language

AshGDS:
## What

- If the `lang` attribute doesn't exist on the `#content` element, use the `<html lang=` value instead.
- Unfortunately I couldn't write a test for this - for some reason, I can't set the `lang` attribute on the `<html>` element within a Jasmine test even though it should be allowed. It might be a jasmine bug since setting other names for attributes on works on the same element.

## Why

https://trello.com/c/ljF64JLm/377-page-view-enhancement-content-language-not-consistently-tracked

## Visual Changes

 None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -112,11 +112,12 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
     getLanguage: function () {
       var content = document.getElementById('content')
+      var html = document.querySelector('html')
       if (content) {
         return content.getAttribute('lang') || this.nullValue
-      } else {
-        return this.nullValue
       }
+
+      return html.getAttribute('lang') || this.nullValue
```
andysellick, line 120: I've been having a think about this. I think what we want to do is capture the first instance of the `lang` attribute, wherever that may be in the DOM. If we use `document.querySelector` it will find that, because it searches each element and its children in turn before moving onto the next one. So something like this should be possible:

```JavaScript
getLanguage: function () {
  var langEl = document.querySelector('[lang]')
  if (langEl) {
    return langEl.getAttribute('lang') || this.nullValue
  }
  return this.nullValue
}
```

and the test:

```JavaScript
  it('finds the first lang attribute on the page', function () {
    var div = document.createElement('div')
    div.innerHTML =
      '<div>' +
        '<div lang="fr">' +
          '<div lang="es"></div>' +
        '</div>' +
      '</div>' +
      '<div lang="en"></div>'
    document.body.appendChild(div)
    GOVUK.analyticsGa4.analyticsModules.PageViewTracker.init()
    expect(window.dataLayer[0].page_view.language).toEqual('fr')
    document.body.removeChild(div)
  })
```

Also you can test the `lang` attribute on the `html` element if you want using `document.documentElement.setAttribute('lang', 'en')`.

AshGDS: @andysellick thanks for the suggestion and for testing it. :+1:

Unfortunately I think the way language works on our pages would make this less useful to us. It seems that the `lang` value in `main#content` is what is correct on foreign pages, and then `lang=en` in the `<html>` is like a secondary fallback, even though it's the first one to appear in the DOM.

For example, on this page: https://www.gov.uk/guidance/support-for-family-members-of-british-nationals-in-ukraine-and-ukrainian-nationals-in-ukraine-and-the-uk.uk

You essentially have this:

```HTML
<html lang="en">
     <main id="content" lang="uk">
        [Ukranian content]  
    </main>
</html>
```

So it's more accurate for us to check the value of the `<main>` element first, and then fallback to `<html lang=en>` when a language doesn't exist on `<main>` (for example on the GOVUK homepage)

Other examples where `<main>` is more accurate than `<html>`: [Welsh consultation] |(https://www.gov.uk/government/consultations/ymgynghoriad-ar-y-cynllun-rheoli-pysgodfa-arfaethedig-ar-gyfer-cregyn-y-brenin.cy) [Finnish news](https://www.gov.uk/government/news/platinum-pudding-jalkiruokakilpailu.fi) |[Portugese news](https://www.gov.uk/government/news/529627.pt) | [Welsh people profile](https://www.gov.uk/government/people/robert-devereux.cy)

Thanks

andysellick: Ah okay fair enough then. You should be able to write a test for this using the line I included above.

AshGDS:
Thanks @andysellick - unfortunately `document.documentElement.setAttribute('lang', 'en')` does not work. I can set attributes on the `html` element, it just doesn't accept `lang` as the attribute name for some reason.

I've pushed the commit with the failing test, in case you want to try and look at why it's failing. :+1:

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.spec.js
```diff
@@ -242,14 +242,23 @@ describe('Google Tag Manager page view tracking', function () {
...
+    it('set correctly using the #content element', function () {
       content.setAttribute('lang', 'wakandan')
       expected.page_view.language = 'wakandan'
 
       GOVUK.analyticsGa4.analyticsModules.PageViewTracker.init()
       expect(window.dataLayer[0]).toEqual(expected)
     })
 
+    it('set correctly using the html lang attribute', function () {
+      document.documentElement.setAttribute('lang', 'cy')
```
andysellick, line 254: Yeah okay this doesn't work. I nearly got it to work with a `setTimeout` but the expectation then happened before it, so it didn't work, and you can't include the expects in a timeout as they don't get evaluated properly. You can remove this bit, but maybe make a note somewhere that setting the lang on the HTML element in Jasmine and therefore testing it isn't possible.

AshGDS:
Thanks @andysellick - I've done this, I added a comment in the `getLanguage` function (see last commit) :+1:

andysellick (approved):

PR #3602, opened 2023-09-11
Add new homepage action link variant

patrickpatrickpatrick:
## What

Add new variant of action link for the homepage. Add new svg `app/assets/images/govuk_publishing_components/action-link-arrow--homepage.svg`.

## Why

As part of the changes being made to the homepage.

## Visual Changes

TBA


maxgds:
No criticism on you, Patrick, but I don't understand why another variant is needed? We have multiple options already (and I'm unsure how many are actually used). The classname is treating "homepage" as a state but being on a homepage isn't really a state as such. What is the differentiating factor that leads to this variation being needed? It seems purely decorative - are we at a point where we should just be refactoring the component and passing arrow colours and background colours in instead of making a new one for each use case - which is somewhat against the principles of components being reusable.

maxgds (approved):
I think in the interests of serving the instant need this is ok to go in, but please raise an issue to the effect of the previous conversation - this component is in severe need of rationalising, probably something to do in conjunction with design.

PR #3615, opened 2023-09-18
Add 'search_term' to GA4 pageview object

AshGDS:
## What

- Adds the `search_term` query string to the pageview.
- Fixes an issue with the email PII redaction but it is not perfect, so it would need your opinion.
    - Currently, when this string is put through the PIIRemover: `hello+world+email@example.com+SW12AA+1990-01-01`, the PII remover will just return it as `[email]`. It should be `hello+world+[email]+[postcode]+[date]`.
    -  This happens because the PIIRemover treats `+` as part of the email address, so the whole string ends up being included as the email address to be redacted. 
    - I fixed this by making `+` a character to be excluded for pattern matching an email. 
    - This works, however I know that in gmail technically users can put a `+` in their email address. [See here](https://eit.ces.ncsu.edu/2023/02/gmail-plus-addressing-the-hidden-feature-that-can-help-you-get-more-out-of-your-inbox/#:~:text=Here's%20how%20it%20works%3A,%2Banythingyouwant%40gmail.com.)
    - Is there a way to stop the `+` issue, but still have email addresses with `+` work? 
    - We could solve it by replacing `+` with a space before it gets put through the PII remover for this `search_query` change. We would probably need to do something similar on the `referrer` `location`, `query_string` pageview values though.
- CI is failing but I believe it's an unrelated issue
## Why

https://trello.com/c/ZYon1wog/688-search-enhancement-new-attribute-search-term
## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -59,7 +59,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
             devolved_nations_banner: this.getElementAttribute('data-ga4-devolved-nations-banner') || undefined,
             cookie_banner: document.querySelector('[data-ga4-cookie-banner]') ? 'true' : undefined,
             intervention: this.getInterventionPresence(),
-            query_string: this.getQueryString()
+            query_string: this.getQueryString(),
+            search_term: this.getSearchTerm()
           }
         }
         window.GOVUK.analyticsGa4.core.sendData(data)
@@ -72,6 +73,23 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
 
     getSearch: function () {
       return window.location.search
+    },
+
+    getSearchTerm: function () {
+      var queryString = this.getSearch()
+
+      if (!queryString) {
+        return undefined
+      }
+
+      var searchTerm = queryString.match(/keywords=([^&]*)/)
+      if (!searchTerm) {
+        return undefined
+      }
+
+      searchTerm = searchTerm[0].replace('keywords=', '')
+      searchTerm = this.PIIRemover.stripPIIWithOverride(searchTerm, true, true)
+      return searchTerm
     },
 
     getQueryString: function () {
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
@@ -2,7 +2,7 @@
   'use strict'
 
   var GOVUK = global.GOVUK || {}
-  var EMAIL_PATTERN = /[^\s=/?&#]+(?:@|%40)[^\s=/?&]+/g
+  var EMAIL_PATTERN = /[^\s=/?&#+]+(?:@|%40)[^\s=/?&+]+/g
   var POSTCODE_PATTERN = /\b[A-PR-UWYZ][A-HJ-Z]?[0-9][0-9A-HJKMNPR-Y]?(?:[\s+]|%20)*[0-9](?!refund)[ABD-HJLNPQ-Z]{2,3}\b/gi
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -74,6 +75,21 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
     },
 
+    getSearchTerm: function () {
+      var queryString = this.getSearch()
+
+      if (queryString.indexOf('&keywords=') === -1 && queryString.indexOf('?keywords=') === -1) {
+        return undefined
+      }
+
+      var searchTerm = queryString.match(/keywords=([^&]*)/)
```
andysellick, line 85: I think some checking could make this code end quicker on non-relevant pages, as well as cut down some of the checks. Could something like this work?

```JavaScript
getSearchTerm: function () {
  var queryString = this.getSearch()

  if (!queryString) {
    return undefined
  }

  var searchTerm = queryString.match(/keywords=([^&]*)/)
  if (!searchTerm) {
    return undefined
  }

  searchTerm = searchTerm[0].replace('keywords=', '')
  searchTerm = this.PIIRemover.stripPIIWithOverride(searchTerm, true, true)
  return searchTerm 
}
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -74,6 +75,21 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       return window.location.search
     },
 
+    getSearchTerm: function () {
+      var queryString = this.getSearch()
```
andysellick, line 79: A bit unrelated (maybe more something to think about for tech debt) but I'm wondering if we should redact PII directly in the `getSearch` function?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -2,7 +2,7 @@
   'use strict'
 
   var GOVUK = global.GOVUK || {}
-  var EMAIL_PATTERN = /[^\s=/?&#]+(?:@|%40)[^\s=/?&]+/g
+  var EMAIL_PATTERN = /[^\s=/?&#+]+(?:@|%40)[^\s=/?&+]+/g
```
andysellick, line 5: Not sure I fully understand the implications of this change, might be worth discussing directly.

Also should there be a test for this change?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
AshGDS: Thanks, will use this instead!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
AshGDS: @andysellick That's probably a good idea :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
andysellick: Thanks, that was helpful. I think this change is good, because it's better to prevent the regex from escaping from the email address as you describe.

The gmail address problem is a concern, but at least we're redacting the majority part of the email, which I think should make it anonymous enough. For example `john+work@gmail.com` (stemming from the original `john@gmail.com`) would be redacted to `john+[email]`.

So I think we're okay, add a test and we should be good 👍

AshGDS:
Thanks @andysellick - should be ready for another review :+1:

andysellick (approved):
Good stuff. If you can squash the relevant commits together and make a note of the tech debt thing on our board somewhere (I think there's a tag for tech debt) that'd be great 👍

AshGDS:
@andysellick Done at https://trello.com/c/FP9j5Zy1/635-put-piiremover-in-getsearch-function :+1:

PR #3620, opened 2023-09-18
Updates placeholder schema references

1pretz1:
The placeholder schema was removed in [1], causing CI to fail. Swaps over to using `guide`.

[1]: https://github.com/alphagov/publishing-api/pull/2493

### DIFF
--- a/spec/features/contextual_navigation_spec.rb
+++ b/spec/features/contextual_navigation_spec.rb
@@ -325,7 +325,7 @@ describe "Contextual navigation" do
   end
 
   def not_step_by_step_content
-    not_step_by_step_content = random_item("placeholder")
+    not_step_by_step_content = random_item("guide")
     not_step_by_step_content["links"].delete("part_of_step_navs")
     not_step_by_step_content["links"].delete("secondary_to_step_navs")
     not_step_by_step_content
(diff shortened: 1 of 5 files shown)


spec/features/contextual_navigation_spec.rb
```diff
@@ -325,7 +325,7 @@ def given_there_is_a_non_step_by_step_parent_page
   end
 
   def not_step_by_step_content
-    not_step_by_step_content = random_item("placeholder")
+    not_step_by_step_content = random_item("generic")
```
leenagupte, line 328: Not a blocker, but I feel like we might end up having a similar problem with "generic" at some point. The only [example in publishing-api](https://github.com/alphagov/publishing-api/blob/main/content_schemas/examples/generic/frontend/services_and_information.json#L10-L12) lists "whitehall-frontend" as the rendering app, which is no longer true of anything.

leenagupte (approved):
Thanks for fixing this 🏅 

One non-blocking comments, but otherwise. 👍

KludgeKML (approved):
LGTM (+ agree with Leena's comment)

spec/features/contextual_navigation_spec.rb
1pretz1: Good point 👍 I've switched all the references to `guide` now

PR #3626, opened 2023-09-21
Fix various bugs with the GA4 pageview tracker

AshGDS:
## What

Fixes various bugs with the GA4 pageview tracker. These are:

1. In the `query_string` parameter, I was redacting PII before stripping `_ga` and `_gl`. This breaks the _ga/_gl stripping code, as the PIIRemover detects the random characters in these values as `[date]`. And therefore the regex for stripping the values would break, as correctly it doesn't count `[` and `]` as part of the _ga/_gl values.
2. Fixes `html lang='en'` never being grabbed in certain scenarios. Previously, if the `#content` element existed, it would either grab the language value from it, or return `this.nullValue` straight away. This isn't correct, as we want to check the HTML language value as well before returning `this.nullValue`. Ironically, we couldn't figure out how to test the `<html>` element, and if we could we would've spotted this sooner!
3. The highest level of PII redaction was removed by accident for `getLocation` in #3568 and `getTitle` in #2842. The default `stripPII` function only does extra redaction if certain meta tags exist. We usually use `stripPIIWithOverride` so that we can enable date and postcode redaction ourselves.
 This is a consequence of porting the PIIRemover directly from UA, and not really knowing what we wanted to redact at the time. Eventually we ended up needing to always use the highest level of redaction, so we created `stripPIIWithOverride` to allow this, whereas UA would use `stripPII` which would only redact emails, state, password tokens and unlock tokens. `stripPII` would only be stricter and redact dates/postcodes if certain meta tags existed on the page. Fortunately on `finder-frontend`, one of the meta tags existed `<meta name="govuk:static-analytics:strip-postcodes" content="true">` so postcodes weren't being tracked in searches. Dates however are coming through so this will fix that. This change will also ensure the data/postcode redaction is site-wide.

## Why

1. Will fix a bug with this card https://trello.com/c/Rb45fH1B/633-search-enhancement-new-attribute-query-string
2. Will fix a bug with this card https://trello.com/c/ljF64JLm/377-page-view-enhancement-content-language-not-consistently-tracked
3. Will prevent PII coming through/keep our PII consistent

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -95,8 +95,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
     getQueryString: function () {
       var queryString = this.getSearch()
       if (queryString) {
-        queryString = this.PIIRemover.stripPIIWithOverride(queryString, true, true)
         queryString = this.stripGaParam(queryString)
+        queryString = this.PIIRemover.stripPIIWithOverride(queryString, true, true)
         queryString = queryString.substring(1) // removes the '?' character from the start.
         return queryString
       }
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -148,8 +148,9 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       var content = document.getElementById('content')
       var html = document.querySelector('html')
       if (content) {
-        if (content.getAttribute('lang')) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -148,7 +148,9 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       var content = document.getElementById('content')
       var html = document.querySelector('html')
       if (content) {
-        return content.getAttribute('lang') || this.nullValue
+        if (content.getAttribute('lang')) {
+          return content.getAttribute('lang')
```
andysellick, line 152: Since you're calling this twice, you should probably store it in a variable.

AshGDS: Whoops, good spot :+1:

andysellick (approved):

PR #3627, opened 2023-09-21
[DO NOT MERGE] Add prettier

andysellick:
## What
Adds Prettier. Maybe.

## Why
Stylelint was [recently upgraded](https://github.com/alphagov/govuk_publishing_components/pull/3522) to v1, which [removes all stylistic rules](https://github.com/alphagov/stylelint-config-gds/blob/main/CHANGELOG.md#100) meaning that we're no longer linting for loads of things such as rogue tabs vs spaces, empty space inside parentheses, single quotes, braces on new lines, etc.

Recommended solution is to adopt Prettier. However Prettier is very opinionated and allows little configuration, which means we'd have to change a lot of our code to match the expectations of the linter. Problems so far:

- it insists on removing spaces after function names e.g. `function example ()` becomes `function example()`
- it insists on adding zeroes to the start of numbers less than 1, e.g. `.2em` becomes `0.2em`
- you can choose between single or double quotes but apparently we're massively inconsistent so either approach will involve loads of changes
- it insists on flat indenting markup snippets in JS files e.g.

```
var html = 
  '<div>' +
    '<p>child</p>' +
  '</div>'
```

becomes

```
var html = 
  '<div>' +
  '<p>child</p>' +
  '</div>'
```

## Visual Changes
 None.

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,0 +1,13 @@
+app/assets/javascripts
+bin
+config
+docs
+lib
+spec/javascripts/helpers
+spec/javascripts/vendor
+spec/support
+CHANGELOG.md
+
+**/*.erb
+**/*.rb
+**/*.yml
--- a/.prettierrc
+++ b/.prettierrc
@@ -1,0 +1,6 @@
+{
+  "semi": false,
+  "singleQuote": true,
+  "singleAttributePerLine": false,
+  "printWidth": 160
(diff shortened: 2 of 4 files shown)


andysellick:
Closing in favour of https://github.com/alphagov/govuk_publishing_components/pull/3629

PR #3628, opened 2023-09-21
Add 'ga4-browse-topic' meta tag to track the mainstream browse topic

AshGDS:
## What

The PAs were previously using the `govuk:section` meta tag to populate `browse_topic` in our pageview tracking. If you look at `meta_tags.rb` in this repo, it looks like this value is populated by grabbing a `section` value from the `content_item`. However, in reality `govuk:section` was being added to the page by digging through the `content_item` in `frontend` and pulling a deeply nested value, rather than just using the `section` value in the `content_item`, which doesn't seem to exist where we'd want it. (For the code that's in `frontend`, [click here](https://github.com/alphagov/frontend/blob/542e53d60f12b21d3299424025cdb7905ede4303/app/controllers/content_items_controller.rb#L8)) This was why the meta tag looked like it was fine for our purpose on certain pages, as these were pages in `frontend`. 

So, `browse_topic` was only being populated in certain places like `frontend`, and not on other 'mainstream browse' pages in other rendering apps, as the `section` value doesn't seem to come through in the `content_item` where the PAs would want it to.

Instead of coding a `govuk:section` meta tag in each frontend app which digs through the content item, we can add one in our global meta tags section in this repo. I've called it `ga4-browse-topic`. This should populate our pageview `browse_topic` wherever it exists.

I guess the only downside of this is that if the user has rejected GA4, an extra meta tag is being sent which isn't used. But I guess that applies for other meta tags anyway that are only used for analytics/debugging purposes - (e.g. `govuk:rendering-app` isn't relevant to the public but is always sent in the page.)

## Why

https://trello.com/c/DDaR9u6C/342-page-view-enhancement-fix-missing-govuksection-meta-tag-on-content-items-that-are-tagged-to-more-than-one-mainstream-browse-topi

## Testing

If you want to see the meta tag working, you can:
- Run `static` with this branch of `govuk_publishing_components`
- Run `government-frontend` with this branch of `govuk_publishing_components` and visit http://127.0.0.1:3090/universal-credit and you will see `<meta name="govuk:ga4-browse-topic" content="benefits">`
- Run `government-frontend` with this branch of `govuk_publishing_components` and visit http://127.0.0.1:3010/check-uk-visa and you will see `<meta name="govuk:ga4-browse-topic" content="visas and immigration">`
- Run `frontend` with this branch of `govuk_publishing_components` and visit http://127.0.0.1:3005/renew-driving-licence and you will see `<meta name="govuk:ga4-browse-topic" content="driving and transport">`

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,8 @@
 ## Unreleased
 
 * Remove option select GA4 attributes ([PR #3625](https://github.com/alphagov/govuk_publishing_components/pull/3625))
+* Fix various bugs with the GA4 pageview tracker ([PR #3626](https://github.com/alphagov/govuk_publishing_components/pull/3626))
+* Add 'ga4-browse-topic' meta tag to track the mainstream browse topic ([PR #3628](https://github.com/alphagov/govuk_publishing_components/pull/3628))
 
 ## 35.16.0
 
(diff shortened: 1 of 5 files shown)


andysellick (approved):
👍

PR #3629, opened 2023-09-21
Add stylistic plugin for stylelint

andysellick:
## What
Adds [stylelint-stylistic plugin](https://github.com/elirasza/stylelint-stylistic). This includes adding the plugin to package.json, along with a few configuration rules to match our coding style.

Some changes to the code are inevitable, but limited, and are included in a separate commit.

## Why

Stylelint was [recently upgraded](https://github.com/alphagov/govuk_publishing_components/pull/3522) to v1, which [removes all stylistic rules](https://github.com/alphagov/stylelint-config-gds/blob/main/CHANGELOG.md#100) meaning that we're no longer linting for loads of things such as rogue tabs vs spaces, empty space inside parentheses, single quotes, braces on new lines, etc.

Recommended solution is to adopt Prettier (attempted in https://github.com/alphagov/govuk_publishing_components/pull/3627). However Prettier is very opinionated and allows little configuration, so this alternative approach is to use a stylelint plugin.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Change prev and next GA4 type ([PR #3631](https://github.com/alphagov/govuk_publishing_components/pull/3631))
 * Remove timestamps from GA4 video urls ([PR #3632](https://github.com/alphagov/govuk_publishing_components/pull/3632))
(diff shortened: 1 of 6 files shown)


maxgds (approved):
Happy for this to go in - whether it winds up as a stepping stone or the final solution is a discussion worth having, but this is good for now.

PR #3630, opened 2023-09-21
Add an `aria-label` to the attachment preview link

### LINKED ISSUE #3560: Link text for CSV attachments is not descriptive
## What

When attaching a CSV file via the [Whitehall block attachment component](https://components.publishing.service.gov.uk/component-guide/govspeak#whitehall_block_attachments) it produces two links: "View online" and "Download CSV 21.7 KB" (for example). The "View online" link is not descriptive when out of context as it doesn't mention what you can view online. The "Download" link is better but would also be confusing if there was more than one of them on the same page.

Example URL:
https://www.gov.uk/government/publications/queen-elizabeth-ii-conference-centre-expenditure-april-2016

## Why

Many screen reader users navigate by links and therefore listen to link text out of context. They can still get the context but would need to search for it before or after the link. It makes it much easier for them if all link texts describe their topic or purpose when viewed out of context.

## More context

This was found in the 2022 WCAG audit of GOV.UK and is a fail of 2.4.4 Link Purpose (In Context).
This issue is not entirely due to the publishing components as some of the markup is coming from Whitehall.

dnkrj:
# What

Add an `aria-label` of "View [attachment title] online" to attachment preview links

# Why

Fixes #3560 

When attaching a CSV file via the [Whitehall block attachment component](https://components.publishing.service.gov.uk/component-guide/govspeak#whitehall_block_attachments) it produces two links: "View online" and "Download CSV 21.7 KB" (for example). The "View online" link is not descriptive when out of context as it doesn't mention what you can view online.

This was found in the 2022 WCAG audit of GOV.UK and is a fail of 2.4.4 Link Purpose (In Context).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -18,6 +18,7 @@
 * Add 'ga4-browse-topic' meta tag to track the mainstream browse topic ([PR #3628](https://github.com/alphagov/govuk_publishing_components/pull/3628))
 * Ensure cookie banner isn't tracked as visible when it is hidden via JS ([PR #3612](https://github.com/alphagov/govuk_publishing_components/pull/3612))
 * Add spelling_suggestion meta tag to the GA4 pageview object ([PR #3633](https://github.com/alphagov/govuk_publishing_components/pull/3633))
+* Add an `aria-label` to the attachment preview link ([PR #3630](https://github.com/alphagov/govuk_publishing_components/pull/3630))
 
 ## 35.16.0
 
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -99,7 +99,7 @@
     <% end %>
 
     <% if attachment.preview_url.present? %>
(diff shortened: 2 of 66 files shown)


ryanb-gds (approved):
That required changes to a surprising number of files

selfthinker:
Although I saw that this is working on www.gov.uk (for example, on [this page](https://www.gov.uk/government/publications/queen-elizabeth-ii-conference-centre-expenditure-april-2016)), it is not working in the [component guide](https://components.publishing.service.gov.uk/component-guide/govspeak#whitehall_block_attachments). What's the reason for that?
Although it's not necessarily an issue if it doesn't work in the component guide as long as it works on the website, it is confusing and might make people believe there is a bug.

selfthinker:
@dnkrj, thanks for this. I'm still not sure I understand it correctly.
Does that mean it is fixed in attachement/with_preview_link but not in govspeak#whitehall_block_attachments?

dnkrj:
It's fixed everywhere, it's just that that demo page is a demo of really only the govspeak styling and uses hardcoded HTML that is now out of date

selfthinker:
Okay, thank you for the clarification. :+1:

PR #3632, opened 2023-09-22
Remove timestamps from GA4 video urls

andysellick:
## What
- tracking videos includes the URL of the video, retrieved from the YouTube API
- the URL changes during playback to include the time position of the video, as a URL parameter of the form `t=[0-9]`
- use regexes to remove these parameters and preserve the original URL

## Why
It's quite hard to track videos being watched in GA4 when the URL is inconsistent - removing the timestamp should make the video URL always the same for a single video.

## Visual Changes
None.

Trello card: https://trello.com/c/4r5QApUs/695-placeholder-video-url-remove-extra-query-strings

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-video-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-video-tracker.js
@@ -70,7 +70,7 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       var data = {}
       data.event_name = 'video_' + event
       data.type = 'video'
-      data.url = player.getVideoUrl()
+      data.url = this.cleanVideoUrl(player.getVideoUrl())
       data.text = player.videoTitle
       data.action = event
       data.video_current_time = Math.round(player.getCurrentTime())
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-video-tracker.js
```diff
@@ -81,6 +81,13 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       var schema = schemas.mergeProperties(data, 'event_data')
 
       window.GOVUK.analyticsGa4.core.sendData(schema)
+    },
+
+    cleanVideoUrl: function (url) {
+      console.log('cleanVideoUrl', url)
```
AshGDS, line 87: Accidentally left in!

AshGDS (changes requested):
Looks good to me, small change requested :+1:

Also worth nothing that I think for youtube URLs,`t` can have different types of time values, such as `01h15m05s`, `200s`, `15m`, `01h` etc. but I can see in the trello card that only whole numbers are coming through for `t`, so I don't think you need to account for this anyway. Just thought it's worth mentioning in case it comes up as an issue later!

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-video-tracker.js
andysellick: D'oh! Sorry 🤦 

Fixed.

andysellick:
Thanks for the review @AshGDS will make a note of your observation on the card as well.

AshGDS (approved):
:+1:

PR #3635, opened 2023-09-22
Change GA4 contents list type

andysellick:
## What
Change GA4 attributes for the contents list component from type `content` to type `contents list`.

## Why
We want more specific types when tracking different kinds of interactions.

## Visual Changes
None.

Trello card: https://trello.com/c/kPQBqdSw/675-change-contents-navigation-selectcontent-event-types-to-contents-list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Change GA4 contents list type ([PR #3635](https://github.com/alphagov/govuk_publishing_components/pull/3635))
 * Change prev and next GA4 type ([PR #3631](https://github.com/alphagov/govuk_publishing_components/pull/3631))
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
LGTM 👍 

Just to let you know, there's one instance of `type: "content"` in `government-frontend`. Maybe it can be changed to use the component? 

Code: https://github.com/alphagov/government-frontend/pull/2801/commits/63d63591942d8b8e8a1160f5f3251d81030f1e3c#diff-1735f0a94e9996bdb30e447e6a9f0af5033f02857cc59c2184f7ceb64b680f91R66

I _think_ it's for these types of pages: https://www.gov.uk/service-manual/helping-people-to-use-your-service/making-your-service-accessible-an-introduction

andysellick:
@AshGDS good spot. I don't think that list could be easily switched to using the gem component as it has some custom JS associated with it to highlight where you are in the page as you scroll. Since it's not technically the gem component I'll ask the PAs if they want this also changed to match, and raise a separate PR if so.

PR #3638, opened 2023-09-25
Add GA4 to nested contents item links

andysellick:
## What
- include GA4 attributes for tracking on nested contents item links
- includes new code to calculate the index_total (total number of links) and the index_link value for each link
- note that where a contents list item is 'active' (i.e. not a link) this should not alter the overall index_total or index_link values, it is assumed that this non-link item is included in those counts

Index values should be as follows:

```
Parent link // index_link 1
- nested link // index_link 2
- nested link // index_link 3
Parent link // index_link 4
- nested link // index_link 5
- active (not a link)
- nested link // index_link 7

index_total = 7
```

## Why
We weren't tracking clicks on nested contents items links, only on top level items.

## Visual Changes
One tiny change to the doc page for this component, under GA4 tracking include an `active` item to help with testing.

Trello card: https://trello.com/c/ZD50PFtM/693-add-tracking-to-dashed-child-content-items

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add GA4 to nested contents item links ([PR #3638](https://github.com/alphagov/govuk_publishing_components/pull/3638))
 * Change GA4 contents list type ([PR #3635](https://github.com/alphagov/govuk_publishing_components/pull/3635))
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
Looks good - just needs a CHANGELOG update :+1:

PR #3639, opened 2023-09-26
Fix option select arrow when CSS missing

andysellick:
## What
- if CSS is disabled or fails to load, this component's SVG arrows (for expand/collapse) appear at a huge size
- this not only looks terrible but advertises that something is broken and gets in the way of the user using the component
- since the CSS sets the width and height of the SVG already, solution is to set width and height inline attributes on the SVG to zero, so if CSS fails to load the SVG doesn't appear

## Why
The component should be mostly usable without styles or JS enabled. Recently there was a problem with this component's styles and the massive SVG arrow was briefly visible on the live site. This has also happened a few times before.

## Visual Changes

Before | After
------ | -------
Screenshot 2023-09-26 at 11 01 18 | Screenshot 2023-09-26 at 11 03 47


MartinJJones (approved):
Nice idea!, tested locally and blocked the option-select.css stylesheet in the network tab and it works perfectly.

PR #3642, opened 2023-09-27
Fix share links styling if CSS disabled

andysellick:
## What
- share links component uses an SVG for social media icons
- if styles for the component are missing or disabled, these SVGs appear at their natural size, which is pretty much the size of the page
- fixing with a width and height attribute of 0 for each SVG, then correcting in the stylesheet when it loads

Also tested in a few older IE versions.

## Why
The site should be broadly usable even if styles and JS are unavailable.

## Visual Changes

Before | After
------ | ------
Screenshot 2023-09-27 at 09 50 57 | Screenshot 2023-09-27 at 09 51 15


MartinJJones (approved):
Changes look good to me

PR #3649, opened 2023-10-04
Flatten GA4 data attributes

andysellick:
## What / why
In a [PR a while back](https://github.com/alphagov/govuk_publishing_components/pull/3436) we removed the need for attributes such as `index_link` to be nested beneath an `index` attribute, as the JS could now find each attribute name and insert it into a copy of the schema, where structure was defined.

This PR updates some of our code to make use of this functionality, by de-nesting nested attributes. Note that this change is internal to the gem and should not break anything - another PR will be raised later to remove the code that currently checks for nested attributes, once it is safe to do so.

## Visual Changes
None.

Trello card: https://trello.com/c/IseA05Fd/644-remove-nesting-of-ga4-attributes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Flatten GA4 data attributes ([PR #3649](https://github.com/alphagov/govuk_publishing_components/pull/3649))
 * Add GA4 'print intent' tracker ([PR #3652](https://github.com/alphagov/govuk_publishing_components/pull/3652))
 
 ## 35.19.0
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -95,6 +95,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       data.external = trackFunctions.isExternalLink(data.url) ? 'true' : 'false'
       data.index = this.setIndex(data.index, event.target)
 
+      // flatten the attributes in index into the main data
+      if (data.index) {
+        for (var prop in data.index) {
+          data[prop] = data.index[prop]
+        }
+        delete data.index
+      }
+
       if (data.type === 'smart answer' && data.action === 'change response') {
         data.section = this.PIIRemover.stripPIIWithOverride(data.section, true, true)
       }
@@ -117,7 +125,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     if (target.getAttribute('data-ga4-index')) {
       try {
         var indexLink = JSON.parse(target.getAttribute('data-ga4-index'))
-
         // Check whether the index object already exists on a parent element, as is the case with tracking accordion links.
         // If true, combine data-ga4-index with the index object. Otherwise, just return indexLink
         index = index ? window.GOVUK.extendObject(index, indexLink) : indexLink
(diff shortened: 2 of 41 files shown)


AshGDS (commented):
Looks good, thanks for doing this large amount of work :+1: 

I've included a small change request.

I have a question - will applications that still use `data-ga4-index` with the nested attributes work? We still use it in `finder-frontend` for the email signup links and the finder tracker so will these need to be updated before this is merged?

andysellick:
> I have a question - will applications that still use `data-ga4-index` with the nested attributes work? We still use it in `finder-frontend` for the email signup links and the finder tracker so will these need to be updated before this is merged?

Probably. The next step is to try this branch out with each application and make sure the tests pass and the tracking behaves broadly as expected. Good to have the details of the index usage though.

JamesCGDS (commented):
Looks good to me, nice work! 👍 The only thing I noticed was that the [Intervention component](https://github.com/alphagov/govuk_publishing_components/blob/8010b1321d3747f73d39e0616dfe0d781d8e43b1/app/views/govuk_publishing_components/components/_intervention.html.erb#L37) and [associated test](https://github.com/alphagov/govuk_publishing_components/blob/8010b1321d3747f73d39e0616dfe0d781d8e43b1/spec/components/intervention_spec.rb#L72C1-L72C1) seem to be using `index` still - would we want to update this as part of this PR?

AshGDS:
Thanks @JamesCGDS - @andysellick  yep that should have always been `index_link` but I accidentally merged it with `index:`. Random aside: Did we used to use `index: 1` in the early days of GA4? Maybe my brain was in early GA4 mode when I wrote it!

andysellick:
Ah cool thanks that makes sense now. Yeah we did used to have a far simpler index strategy, I wondered if it was a holdover from the early days. I'll change it to `index_link` 👍

andysellick:
So I _think_ this is safe to merge - all of the changes are internal to the gem, and shouldn't impact applications. @AshGDS @JamesCGDS do you agree? Have I overlooked anything?

JamesCGDS:
@andysellick I agree, I think this is safe to merge. I'll approve this but if you'd like Ash to also give the green light to be doubly sure then that's cool 👍

AshGDS (approved):
@andysellick If deployed it will break the "Subscribe to feed" and "Get emails" indexes in finder-frontend as I've just tested them with your branch. So these would need to be changed in `finder-frontend` before it is deployed. https://www.gov.uk/administrative-appeals-tribunal-decisions

The finder tracker filter heading indexes still seem to work, but I'll guess you'll just need to flatten them like everything else in this repo has been flattened.

Will approve as the code is fine, and I guess you or I can fix the `finder-frontend` stuff before another release of the gem.

(And again, thanks for putting in the effort to make all these changes!)

andysellick:
Thanks for looking into this @AshGDS. You're right, this definitely breaks the indexes in those links. I think this PR might be a little over-zealous and my approach to it has changed slightly. I think we want to remove nesting where it occurs in attributes that are read directly by trackers, but in other cases (such as this `data-ga4-index` attribute) we don't necessarily need to remove the nesting.

This is because that `data-ga4-index` attribute isn't read directly by a tracker, it's parsed and processed first and combined with the read attributes and then fed to a tracker. So it doesn't matter if it's nested or not, and actually having it as JSON and a single attribute has advantages - it's simpler, we only have to read one attribute, and combining it with the read attribute data is easier.

I've added another commit to this PR to undo some of the earlier stuff. I've checked and it works with finder-frontend.

Also, if it turns out I'm wrong about the above, and we do need to flatten this stuff, we can always address it in a separate PR (this one is big enough) before we do the actual final removal of the nesting code.

AshGDS (approved):
Thanks @andysellick - LGTM :+1:

PR #3650, opened 2023-10-05
Version 35.18.0

patrickpatrickpatrick:
* Change GA4 type on contents lists ([PR #3647](https://github.com/alphagov/govuk_publishing_components/pull/3647))
* Bump Ruby version and use floating patch version ([PR #3646](https://github.com/alphagov/govuk_publishing_components/pull/3646))
* Add homepage variant of navbar for new design of homepage ([PR #3566](https://github.com/alphagov/govuk_publishing_components/pull/3566))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 35.18.0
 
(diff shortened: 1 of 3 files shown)


hannalaakso (approved):
LGTM 👍 

(For reference, there was a difference flagged by Percy here where a "new" snapshot seems to have been generated. This seems to because Percy failed to include that unrelated snapshot in an earlier Percy build (4562) so it got included in this one instead. A bit strange but seems like a Percy thing rather than any kind of issue for us.)

PR #3652, opened 2023-10-05
Add GA4 'print intent' tracker

AshGDS:
## What

- Adds a new tracker called `ga4-print-intent-tracker`
- This runs on pageload, adding a `beforeprint` listener to the `window`
- The event listener fires if the user triggers a "Print page" via their browser (e.g. with `Ctrl/Cmd + P`, or by going to `File > Print`)

## Why

- This is tracked in UA https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/analytics/print-intent.js . I'm not sure why the code is a lot more complicated, I guess the UA requirements were different. I didn't reuse this code as well because `mediaQueryList.addListener` is deprecated JS.
- https://trello.com/c/EpTcwew0/684-print-page-event-to-be-sent-with-ctrl-cmdp

## How to test
- Run `static` with a local version of `govuk_publishing_components` pointing to this branch
- Trigger a print page on static, and you should see the object in the data layer

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.js
@@ -1,0 +1,24 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analyticsModules || {};
+
+(function (analyticsModules) {
+  'use strict'
+
+  var Ga4PrintIntentTracker = {
+    init: function () {
+      window.addEventListener('beforeprint', function () {
+        var data = {
+          event_name: 'print_page',
+          type: 'print page',
+          method: 'browser print'
+        }
+        var schema = new window.GOVUK.analyticsGa4.Schemas()
+        schema = schema.mergeProperties(data, 'event_data')
+        window.GOVUK.analyticsGa4.core.sendData(schema)
+      })
+    }
+  }
+
+  analyticsModules.Ga4PrintIntentTracker = Ga4PrintIntentTracker
+})(window.GOVUK.analyticsGa4.analyticsModules)
--- a/docs/analytics-ga4/ga4-print-intent-tracker.md
+++ b/docs/analytics-ga4/ga4-print-intent-tracker.md
@@ -1,0 +1,20 @@
+# Google Analytics 4 print intent tracker
+
+This script tracks when the user opens the print prompt in their browser, e.g. when pressing Ctrl/Cmd + P or going to File > Print.
+
+## How it works
+
+Assuming that consent has been given, the print intent tracker is launched automatically by `init-ga4.js`, as there is a function in that file which automatically runs the `init()` function on anything namespaced under `window.GOVUK.analyticsGa4.analyticsModules`.
+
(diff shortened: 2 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.js
```diff
@@ -0,0 +1,24 @@
...
+  var Ga4PrintIntentTracker = {
+    init: function () {
+      window.addEventListener('beforeprint', function () {
+        var data = {
+          event_name: 'print_page',
+          type: 'print page',
+          method: 'browser print'
+        }
+        var schemas = new window.GOVUK.analyticsGa4.Schemas()
+        var schema = schemas.mergeProperties(data, 'event_data')
```
andysellick, line 17: Thought for a moment that you'd declared the same variable twice! But you hadn't. Is it worth reusing the schemas variable here?

Side note - possible bit of tech debt. I think always now we do these three lines (declare the schema, merge the schema, call sendData) and maybe we could move all of that into sendData and just call that instead. I'll look into it.

docs/analytics-ga4/ga4-print-intent-tracker.md
```diff
@@ -0,0 +1,20 @@
+# Google Analytics 4 print intent tracker
+
+This script tracks when the user opens the print prompt in their browser, e.g. when pressing Ctrl/Cmd + P or going to File > Print.
+
+## How it works
+
+Assuming that consent has been given, the print intent tracker is launched automatically by `init-ga4.js`, as there is a function in that file which automatically runs the `init()` function on anything namespaced under `window.GOVUK.analyticsGa4.analyticsModules`.
+
+When intialised, a `beforeprint` event listener is added to the JavaScript `window`. If the `beforeprint` event is fired, the following dataLayer object is sent to GA4:
```
andysellick, line 9: Typo: `initialised`

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.spec.js
```diff
@@ -0,0 +1,31 @@
...
+  var GOVUK = window.GOVUK
+  var expected
+
+  beforeAll(function () {
+    window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+    window.GOVUK.analyticsGa4.vars = window.GOVUK.analyticsGa4.vars || {}
+    window.GOVUK.analyticsGa4.vars.gem_version = 'aVersion'
+  })
+
+  describe('when the user prints the page', function () {
```
andysellick, line 13: Since there's only one test here the `describe` block is perhaps unnecessary?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.js
andysellick: Have added a card: https://trello.com/c/sI76EaNX/655-investigate-shortening-calls-to-schema-code

andysellick (commented):
Looks good overall, just a few minor things.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-print-intent-tracker.js
AshGDS: Sounds good :+1:

andysellick (approved):

PR #3659, opened 2023-10-12
Implement One Login "cross service header"

jon-kirwan:
## What

https://trello.com/c/9o9pvE3a/2146-implement-one-login-cross-service-header-on-the-email-subscriptions-service

## Why

> - To help us here in One Login improve the component
> - To provide email subscriptions service users with the navigation improvements we’re already confident about following our testing

## Visual changes
See: https://components-gem-pr-3659.herokuapp.com/component-guide/cross_service_header

## Anything else
- I've added the naming convention of '**cross service header**' for both component and file names, aligning them with the JavaScript module name. Additionally, CSS classes use `gem-c-cross-service-header`, `gem-c-one-login-header`, and `gem-c-service-header`, maintaining concise and clear naming.
- See spike notes [here](https://docs.google.com/document/d/1PZMk0DA86QQLdJLKAAEH6Fwa75sxvEcs3RcXdMTbbQE/edit?pli=1) which details why it's not currently possible to replace the **layout header** component (and hence the need to add new properties such as `show_cross_service_header` and `one_login_navigation_items`.

Also, see:
- https://github.com/alphagov/email-alert-frontend/pull/1634
- https://github.com/alphagov/static/pull/3170

### DIFF
--- a/app/views/govuk_publishing_components/components/cross_service_header/_service_header.html.erb
+++ b/app/views/govuk_publishing_components/components/cross_service_header/_service_header.html.erb
@@ -1,0 +1,49 @@
+<% service_navigation_aria_label ||= "Service menu" %>
+
+<% if product_name %>
+<div class="gem-c-service-header" data-one-login-header-nav>
+  <div class="govuk-width-container">
+    <div class="gem-c-service-header__container">
+      <!-- The name of your service goes here -->
+      <h2 class="gem-c-service-header__heading"><%= product_name %></h2>
+      <% if service_navigation_items.any? %>
+        <div>
+          <button type="button" 
+            aria-controls="service-header__nav" 
+            data-open-class="gem-c-cross-service-header__button--open" 
+            aria-label="Show service navigation menu" 
+            data-label-for-show="Show service navigation menu" 
+            data-label-for-hide="Hide service navigation menu" 
+            data-text-for-show="Menu" 
+            data-text-for-hide="Close" 
+            aria-expanded="false" 
+            class="gem-c-cross-service-header__button gem-c-cross-service-header__button--gem-c-service-header js-x-header-toggle">
+            <span class="gem-c-service-header__button-content">Menu</span>
+          </button>
+          <!-- Important! Label your navigation appropriately! When there are multiple navigation landmarks on a page, additional labelling is required to provide support for screen reader users -->
+          <nav aria-label="<%= service_navigation_aria_label %>">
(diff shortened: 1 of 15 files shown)


app/views/govuk_publishing_components/components/cross_service_header/_service_header.html.erb
```diff
@@ -0,0 +1,50 @@
...
+            aria-label="Show service navigation menu" 
+            data-label-for-show="Show service navigation menu" 
+            data-label-for-hide="Hide service navigation menu" 
+            data-text-for-show="Menu" 
+            data-text-for-hide="Close" 
+            aria-expanded="false" 
+            class="gem-c-cross-service-header__button gem-c-cross-service-header__button--gem-c-service-header js-x-header-toggle">
+            <span class="gem-c-service-header__button-content">Menu</span>
+          </button>
+          <%# TODO: "govuk-header__navigation" is required by slimmer, set margin 0 for now to ensure design is consistent %>
```
MartinJJones, line 23: We should be able to remove this TODO: comment now, in regard to the comment "govuk-header__navigation is required by Slimmer", this is used by the AccountsShower processor in Slimmer, but the CSS class names are not required in the new service header template.

There is no documentation for the AccountsShower processor in Slimmer (I'll see if I can add anything based on what knowledge I have), but the class names is used as below:

- [AccountsShower is_navigation_empty?](https://github.com/alphagov/slimmer/blob/main/lib/slimmer/processors/accounts_shower.rb#L57) - references the `govuk-header__navigation` class used in the [navigation_items template](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb#L4)

- if [is_navigation_empty?](https://github.com/alphagov/slimmer/blob/main/lib/slimmer/processors/accounts_shower.rb#L28-L31) is true, then the div element with the `govuk-header__content` class is removed - referenced in the [layout_header template](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_header.html.erb#L53)

MartinJJones (commented):
Changes look good to me overall, just one small nitpick to remove one line comment, I've added some further detail as to why this can be removed as well.

app/views/govuk_publishing_components/components/cross_service_header/_service_header.html.erb
jon-kirwan: Thanks, that's good to know. I've removed the comment and merged the change in with the "add 'cross service header' component" commit.

MartinJJones (approved):
Great work on this!

PR #3662, opened 2023-10-13
Limit GA4 ecommerce arrays to 200 items

AshGDS:
## What

- Limits the amount of items in the GA4 ecommerce items array to a maximum of 200 items

## Why

- There are a couple search finders that contain more than 200 search results on one page, for example https://www.gov.uk/world/organisations and https://www.gov.uk/government/groups
- The dataLayer push for these was fine, but the PAs were having issues with the data when it was in their dashboard
- We think this is due to GA4 limiting how big an array sent to it can be
- Therefore we are limiting the array to the GA4 limit in their documentation

https://trello.com/c/wLz7uULw/679-fix-world-organisations-finder-ecommerce-tracking
https://trello.com/c/jHYCtOA8/678-fix-groups-finder-ecommerce-tracking
## Visual Changes

None.


JamesCGDS (approved):
LGTM! 🚀

PR #3663, opened 2023-10-13
Allow GA4 Form text values to be undefined instead of 'No answer given'

AshGDS:
## What

- Adds detection of a data attribute on forms called `data-ga4-form-undefined-if-text-is-empty` 
- When this is present on a form, and the submitted text by the user on the form is empty, it will push `undefined` to GA4 instead of `No answer given`
- Let me know if `data-ga4-form-undefined-if-text-is-empty` is too long of a name

## Why

https://trello.com/c/WtvEWiS3/592-placeholder-replace-no-answer-given-with-undefined-on-homepage-search-event

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
@@ -9,6 +9,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.trackingTrigger = 'data-ga4-form' // elements with this attribute get tracked
     this.includeTextInputValues = this.module.hasAttribute('data-ga4-form-include-text')
     this.redacted = false
+    this.useFallbackValue = this.module.hasAttribute('data-ga4-form-no-answer-undefined') ? undefined : 'No answer given'
   }
 
   Ga4FormTracker.prototype.init = function () {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -123,5 +123,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     return answers
   }
 
+  Ga4FormTracker.prototype.useFallbackValue = function (formElement) {
+    if (formElement.closest('[data-ga4-form-undefined-if-text-is-empty]')) {
```
andysellick, line 127: I'd suggest moving this into the constructor function at the start, along with the other main variables. It would then be something like

```Javascript
this.useFallbackValue = this.module.hasAttribute('data-ga4-form-undefined-if-text-is-empty') ? undefined : 'No answer given'

...

data.text = data.text || (this.combineGivenAnswers(formData) || this.useFallbackValue)
```

Slightly unrelated (and it's weird I'm asking since I wrote this originally) do you think the `()` in the line above are needed? Would `data.text || this.combineGivenAnswers(formData) || this.useFallbackValue` be any different?

andysellick, line 127: Maybe `data-ga4-form-no-answer-undefined`? It's still a bit of a mouthful though.

AshGDS: Thanks, i'll use this instead

AshGDS: Thanks, have updated both :+1:

AshGDS:
Should be ready for a re-review, thanks @andysellick :+1:

andysellick (approved):
Nice 👏

PR #3664, opened 2023-10-13
Return undefined for meta tags if content is an empty string

AshGDS:
## What

- On search pages, when there's no spelling suggestion, https://www.staging.publishing.service.gov.uk/search/all?keywords=test&order=relevance the `govuk:spelling-suggestion` meta tag exists, but its content is empty. This pushes an empty string for the `spelling_suggestion` pageview value to GA4, when it should be `undefined`.
- This PR fixes that. Although I'm wondering if the meta tag should be fixed in `finder-frontend` instead.

## Why

https://trello.com/c/Nq2h1Kdl/708-spelling-suggestion-should-be-undefined-when-not-present-instead-of

## Visual Changes

None.


andysellick (approved):
This is good, although I wonder if we should bake this directly into the schema stuff - so that anything passed to the schema that is an empty string like this just gets discarded/returned as undefined. Worth adding a tech debt card?

PR #3665, opened 2023-10-16
Add viewport size to GA4 page view

andysellick:
## What
Adds browser viewport width and height to the GA4 page view.

Uses a fairly clean approach that appears to be reasonably backwards compatible for browsers.

In the tests I'm using the function I'm testing to get the viewport size, which seems a bit redundant but otherwise I think we'd need to be setting the viewport size in the test, which could get complicated and messy. Although now I write this that sounds useful. Perhaps something to consider while reviewing. EDIT: done some digging, this indeed looks difficult and messy. Maybe best avoided.

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/9J21E5KI/698-viewport-size-custom-dimension

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add viewport size to GA4 page view ([PR #3665](https://github.com/alphagov/govuk_publishing_components/pull/3665))
 * Replace '+' in GA4 search term values with an actual space ([PR #3653](https://github.com/alphagov/govuk_publishing_components/pull/3653))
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
@andysellick Looks good. In terms of testing, maybe you could move the grabbing of `document.documentElement.clientWidth` and `window.innerWidth` into their own getter functions, and then these functions could be spied on to return necessary values for testing?

andysellick:
@AshGDS that's a good shout, but I'm still wondering whether it's worth doing - the test is that we intercept a request for a number, replace it with a known number, and check that we get that number. I'm not certain that's worth the extra lines of code, but it does feel like this should test _something_.

PR #3667, opened 2023-10-16
Fix GA4 contents list events

andysellick:
## What
Set the GA4 attributes for contents list events to be the correct values. Hopefully. This time.

## Why
I got them wrong. I don't know how.

## Visual Changes
None.

Trello card: https://trello.com/c/pWP26ZKI/691-add-contents-list-selectcontent-events-to-taxon-pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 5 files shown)


AshGDS (approved):
Looks good, does it need a CHANGELOG?

PR #3668, opened 2023-10-17
Improve attachment details styles

andysellick:
## What
Fix some styling issues with the details component where it appears inside the attachment component.

- where the details component is included in the attachment component inside govspeak it was not being styled, because the details stylesheet was not being included
- also increases font size of this component in this context, and fixes a spacing issue

## Why
Improve accessibility and visual appearance.

## Visual Changes
The problem can be seen on [pages like this](https://www.gov.uk/government/groups/secretary-of-state-for-transports-honorary-medical-advisory-panel-on-driving-and-psychiatric-disorders) rendered using govspeak inside `government-frontend`. The details component lacks focus and other styles. In addition, the details component had been shrunk so that it was smaller than its contents.

Spacing has also been added to the top of the details component in this context as the paragraph above has a 'last of type' zero bottom margin. This lack of a gap was less noticeable with the former smaller font size.

Before | After
------- | ------
Screenshot 2023-10-17 at 14 09 46 | Screenshot 2023-10-17 at 14 11 01

### DIFF
--- a/app/views/govuk_publishing_components/components/_govspeak.html.erb
+++ b/app/views/govuk_publishing_components/components/_govspeak.html.erb
@@ -1,5 +1,6 @@
 <%
   add_gem_component_stylesheet("govspeak")
+
   inverse ||= false
   direction_class = "direction-#{direction}" if local_assigns.include?(:direction)
   disable_youtube_expansions = local_assigns.fetch(:disable_youtube_expansions) if local_assigns.include?(:disable_youtube_expansions)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Improve attachment details styles ([PR #3668](https://github.com/alphagov/govuk_publishing_components/pull/3668))
 * Fix GA4 contents list events ([PR #3667](https://github.com/alphagov/govuk_publishing_components/pull/3667))
 * Add viewport size to GA4 page view ([PR #3665](https://github.com/alphagov/govuk_publishing_components/pull/3665))
 * Replace '+' in GA4 search term values with an actual space ([PR #3653](https://github.com/alphagov/govuk_publishing_components/pull/3653))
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/_govspeak.html.erb
```diff
@@ -1,5 +1,6 @@
 <%
   add_gem_component_stylesheet("govspeak")
+  add_gem_component_stylesheet("details")
```
andysellick, line 3: @MartinJJones this solves the problem for now but isn't required under normal circumstances for the attachment component to correctly render the details component. Is this something better addressed in govspeak?

MartinJJones: Yeah, I think it would be best to fix this in govspeak.

Instead of including the additional stylesheet here, we could always import the `details` stylesheet into `/govspeak/attachment.scss` - https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss#L13-L14 which appears to follow the current convention, for example:

```
  @import "../attachment-link";
  @import "../attachment";
  @import "../details";
```

This would mean that we do not request an additional stylesheet when the details component is not used, however I would expect this to slightly increase the overall size of the govspeak.css

I still think this approach makes sense however, for example if an application does not use the individual loading of stylesheets feature but needs to render the `attachment` component, we would likely face the same issue with styling, in that we would still need to remember to include the `details` component styles as well, we effectively would do the same here but for Govspeak.

MartinJJones (commented):
Nice work on identifying the issue and also increasing the font-size, which looks much better (getting rid of 14px fonts 👍)

I do think we should look to include the details component in Govspeak and have added a suggestion around that as well.

app/views/govuk_publishing_components/components/_govspeak.html.erb
andysellick: Thanks @MartinJJones. I've made and tested your suggestion, works fine, makes sense 👍

MartinJJones (approved):
Changes look good to me, thanks 👍

PR #3671, opened 2023-10-18
Update the styling of image-card two thirds variation

MartinJJones:
## What
Update the styling of the image-card component (two thirds variation), the changes include:

- Remove the `display: block` rule, using the default value set of `display: flex` instead
- Use flex to align the card image and card text in the same row
- Wrap the image card text onto a new line if the width would go below 70%
- Ensure the card image and card text align as intended

## Why
As part of the changes for the new homepage design and to help ensure that the image card text does not appear too squashed at smaller Desktop screen sizes.

[Trello card](https://trello.com/c/LbnpQxg3/2136-iterate-featured-section-images-m)

## Visual Changes
For larger desktop screen sizes, there is no change to the layout of the featured section, it will remain as shown in the screenshot below:

desktop-new-featured-section

### Featured section new homepage before - 768px
desktop-769-before

### Featured section new homepage after - 768px
desktop-769-after

### IE11 support
When testing in browser stack, IE11 will render the card image above the card text as shown below:

IE11-layout

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/image_card.yml
+++ b/app/views/govuk_publishing_components/components/docs/image_card.yml
@@ -205,8 +205,8 @@ examples:
       <div class="govuk-!-width-full">
         <%= component %>
       </div>
-  two_thirds_column:
-    description: This variant is used for the featured section on the homepage, the aspect ratio used is 1:1
+  two_thirds:
+    description: This variant is used for the featured section on the homepage. The aspect ratio used is 1:1 and the width of the image is constrained to 80px wide.
     data:
       two_thirds: true
       href: "/still-not-a-page"
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Add GA4 'print intent' tracker ([PR #3652](https://github.com/alphagov/govuk_publishing_components/pull/3652))
 * Update LUX to version 312 ([PR #3672](https://github.com/alphagov/govuk_publishing_components/pull/3672))
 * Grab data-ga4-ecommerce-content-id in GA4 ecommerce tracking ([PR #3676](https://github.com/alphagov/govuk_publishing_components/pull/3676))
+* Update the styling of image-card two thirds variation ([PR #3671](https://github.com/alphagov/govuk_publishing_components/pull/3671))
 
 ## 35.19.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_image-card.scss
@@ -41,13 +41,6 @@
   }
 }
 
-@include govuk-media-query($from: tablet) {
-  .gem-c-image-card.gem-c-image-card--two-thirds {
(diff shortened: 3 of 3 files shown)


MartinJJones:
@hannalaakso @patrickpatrickpatrick would be interested to get your thoughts on the approach used here, do you think we would also need to update the component variation name given this new approach?

patrickpatrickpatrick:
This looks good to me. I personally don't think the name needs to change because it's using two-thirds instead of half and half which is what the the image cards use.

MartinJJones:
> Perhaps 'column' could be removed from the name of the variant in the documentation now though, since it's not going to be a column at all viewports

Thanks @patrickpatrickpatrick, good suggestion, I've made this change in the latest commit.

I've also added further detail on the flex styles used and included a screenshot of the rendering in IE11 from BrowserStack.

app/views/govuk_publishing_components/components/docs/image_card.yml
```diff
@@ -205,8 +205,8 @@ examples:
       <div class="govuk-!-width-full">
         <%= component %>
       </div>
-  two_thirds_column:
-    description: This variant is used for the featured section on the homepage, the aspect ratio used is 1:1
+  two_thirds:
+    description: This variant is used for the featured section on the homepage. The aspect ratio used is 1:1 and the width of the image is constrained to 90px wide.
```
hannalaakso, line 209: A tiny thing, but should this say `95px wide`?

hannalaakso (changes requested):
The PR looks good 👍  However, I noticed looking at the changes flagged by Percy that something seems to have changed with the Youtube video variant so I think that just needs checking. Attaching a screenshot from the PR preview: 
Screenshot 2023-10-19 at 18 21 22

app/views/govuk_publishing_components/components/docs/image_card.yml
MartinJJones: Thanks Hanna, good spot, I've just updated this, it should actually read `80px wide`, as the flex-basis for the image card here is set to 95px, padding is 15px, leaving 80px for the image. I've also updated the related comment in the SCSS file around this.

MartinJJones:
> The PR looks good 👍 However, I noticed looking at the changes flagged by Percy that something seems to have changed with the Youtube video variant so I think that just needs checking. Attaching a screenshot from the PR preview:

I've just had a further look at Percy and I think the visual changes are because the image card two thirds variation now uses a smaller imager, so all the rest of the content on the page appears to move up when looking and the before and after screenshots, for example:


The YouTube video variant does appear to be an issue, but I don't think it is related to this PR as it currently appears this way on the live component guide example page - https://components.publishing.service.gov.uk/component-guide/image_card

hannalaakso (approved):
Thanks @MartinJJones! As discussed, the video variant on the current live gem app looks correct to me but the PR preview doesn't. Very strange but another dev has just confirmed that they're seeing the same as you and it has been like that for a while. I'll try and have a look separately on what's going on, but don't want to block this PR. Thanks for raising the ticket too.

PR #3673, opened 2023-10-19
Use custom SVG for search 'cancel' icon

### LINKED ISSUE #3086: Native implementation of search cancel button doesn't response to browser zoom
Hi there!

When you do browser zoom, the search cancel button doesn't scale properly and it looks tiny (see screenshot). The issue happens in Chrome and Safari.
Screenshot 2022-11-25 at 15 47 11

Since we are looking at this, could we consider overriding the cancel button style so it looks like the close icon used in the[ layout super navigation header ](https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header)

For example, I see that Google uses their own style on their search bar. Here's an article that mentions how we could potentially do this: https://stackoverflow.com/questions/22928819/override-the-webkit-search-cancel-button

Thanks

AshGDS:
## What

- Replaces the default search 'cancel' button with our own custom SVG provided by one of our designers. The default browser icon does not scale well when zoomed in, so we are using our own SVG for this icon which scales when zoomed in.
- I have run the SVG through a minifier to reduce it's filesize.
- I have made it bigger than the default icon, as a bigger hitbox is better for accessibility, especially on mobile. But let me know if it's too big.
- This doesn't apply to Firefox as Firefox doesn't support search cancel buttons. I have been asked by our GOVUK Lead FE Dev to raise an issue in this repo that we're using this non standard CSS, just so we're kept aware of it once it's merged.
- Visual changes have been approved by the designer 

## Why

- The default cancel button used by browsers does not scale well when zoomed in
- Closes #3086

## Testing
- Chrome on Desktop & Mobile ✅ 
- Edge ✅ 
- IE 11 ✅ 
- Safari on Mac & iPhone ✅ 
- Samsung Browser ✅ 
- Firefox ❌ - It doesn't support `::-webkit-search-cancel-button` so there are never any cancel buttons present on Firefox. [See the MDN page for ::webkit-search-cancel-button](https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-search-cancel-button)

### Before

image

image

### After

image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Flatten GA4 data attributes ([PR #3649](https://github.com/alphagov/govuk_publishing_components/pull/3649))
+* Use custom SVG for search 'cancel' icon ([PR #3673](https://github.com/alphagov/govuk_publishing_components/pull/3673))
 * Add GA4 'print intent' tracker ([PR #3652](https://github.com/alphagov/govuk_publishing_components/pull/3652))
 * Update LUX to version 312 ([PR #3672](https://github.com/alphagov/govuk_publishing_components/pull/3672))
 * Grab data-ga4-ecommerce-content-id in GA4 ecommerce tracking ([PR #3676](https://github.com/alphagov/govuk_publishing_components/pull/3676))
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Looks good 👍 

Tested on a few random browsers and devices, all seems good.

PR #3676, opened 2023-10-20
Grab data-ga4-ecommerce-content-id in GA4 ecommerce tracking

AshGDS:
## What

- Allows the GA4 ecommerce tracker to grab a data attribute called `data-ga4-ecommerce-content-id` and push it to the dataLayer.  We want to track the content id of results, as the URLs get cut off by GA4s 100 character limit.
- We are using this for finders only, so any of the tests referring to smart answers have the value set to `undefined`

## Why

- High priority bug fix https://trello.com/c/7Snj5Lx1/710-search-enhancement-add-content-id-to-searchresults-ecommerce-items-because-link-paths-are-truncated-to-100-characters
- Related to https://github.com/alphagov/finder-frontend/pull/3192 but either PR can be merged in any order, as nothing should crash if one is merged and the other isn't
## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -325,6 +325,7 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
           var target = data.event.target
           ecommerceSchema.search_results.ecommerce.items.push({
             item_id: target.getAttribute('data-ga4-ecommerce-path'),
+            item_content_id: target.getAttribute('data-ga4-ecommerce-content-id') || undefined,
             item_name: target.textContent,
             item_list_name: listTitle,
             index: window.GOVUK.analyticsGa4.core.ecommerceHelperFunctions.getIndex(target, startPosition)
@@ -351,6 +352,7 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
 
             ecommerceSchema.search_results.ecommerce.items.push({
               item_id: path,
+              item_content_id: item.getAttribute('data-ga4-ecommerce-content-id') || undefined,
               item_list_name: listTitle,
               index: window.GOVUK.analyticsGa4.core.ecommerceHelperFunctions.getIndex(item, startPosition)
             })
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -325,6 +325,7 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
           var target = data.event.target
           ecommerceSchema.search_results.ecommerce.items.push({
             item_id: target.getAttribute('data-ga4-ecommerce-path'),
+            item_content_id: target.getAttribute('data-ga4-ecommerce-content-id') || undefined,
```
andysellick, line 328: Slightly confused by the structure of this code but assuming there's a simple explanation. Further down on line 355 we're adding the new id for each of the items in the search results, but what's this one here? I can't make out if it's the first one, or a set of parent attributes? From the test changes it looks like we're just checking for the ones further down, so can you clarify what this one is?

andysellick (commented):
Looks fine, just one question.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: @andysellick One of them is for the object that is pushed when a search result is clicked, and the other one is for the object that is pushed on page load/the results updating with an array of all the search results

andysellick: Thanks, that makes sense. And do the changes to the tests cover both instances?

AshGDS: @andysellick I believe so - there's one for the click of a search result here: https://github.com/alphagov/govuk_publishing_components/blob/6da1e132b7f0d15a1aad4d3e3dd78436d2b1014d/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js#L241 

and one for the view_item_list event here:

https://github.com/alphagov/govuk_publishing_components/blob/6da1e132b7f0d15a1aad4d3e3dd78436d2b1014d/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js#L152

Thanks

andysellick: Great, thanks 👍

andysellick (approved):

PR #3680, opened 2023-10-23
Prevent GA4 data sending during Smokey tests

AshGDS:
This prevents the GA4 dashboards being polluted with Smokey test data

## What

- Our GA4 dashboards are being polluted with test data coming from Smokey
- It seems that smokey sets it's user agent to `Smokey Test / Ruby`. As we can grab user agents in JS, we can use this to prevent data being sent if `Smokey Test / Ruby` is the user agent. See https://github.com/alphagov/smokey/blob/32678764d2dfdf18615b2e24d30c615b6cfec644/features/support/env.rb#L33 and https://github.com/alphagov/smokey/blob/32678764d2dfdf18615b2e24d30c615b6cfec644/features/support/http_requests.rb#L77
- I am preventing Smokey data from being sent to GA4 in the `sendData` function. I considered just disabling analytics entirely if it was a Smokey test, but I thought this was a bad idea. Mainly because then we couldn't write new Smokey tests which check our GA4 code is actually working. By doing it this way, in the future we could modify `sendData` so that if the user is a smokey test, it pushes to a different array such as `window.smokeyLayer` so that we can have smokey tests that check our dataLayer is working without actually pushing any data to GA4.

## Why

- https://trello.com/c/TnMU9ZPm/709-investigate-how-to-filter-smokey-tests

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add metadata inverse no padding option ([PR #3689](https://github.com/alphagov/govuk_publishing_components/pull/3689))
+* Prevent GA4 data sending during Smokey tests ([PR #3680](https://github.com/alphagov/govuk_publishing_components/pull/3680))
 
 ## 35.20.1
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -30,7 +30,16 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
       firstScript.parentNode.insertBefore(newScript, firstScript)
     },
 
+    getUserAgent: function () {
+      return navigator.userAgent
+    },
+
     sendData: function (data) {
+      // Prevent any GA4 data being sent during Smokey tests
+      if (this.getUserAgent() === 'Smokey Test / Ruby') {
+        return
+      }
+
       data.govuk_gem_version = this.getGemVersion()
       // set this in the console as a debugging aid
       if (window.GOVUK.analyticsGa4.showDebug) {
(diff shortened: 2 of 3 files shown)


AshGDS:
I used a [user agent faker](https://addons.mozilla.org/en-GB/firefox/addon/user-agent-string-switcher/) in Firefox (Chrome wasn't letting me change the user agent) and it seems to work. With `Smokey Test / Ruby` as the user agent, the `dataLayer` in `static` was just

``` JSON
[
  {
    "gtm.blocklist": [
      "customPixels",
      "customScripts",
      "html",
      "nonGoogleScripts"
    ]
  },
  {
    "gtm.start": 1698330200088,
    "event": "gtm.js",
    "gtm.uniqueEventId": 1
  },
  {
    "event": "gtm.dom",
    "gtm.uniqueEventId": 2
  },
  {
    "event": "gtm.load",
    "gtm.uniqueEventId": 3
  }
]
```

Whereas it's usually

```JSON
[
  {
    "gtm.blocklist": [
      "customPixels",
      "customScripts",
      "html",
      "nonGoogleScripts"
    ]
  },
  {
    "gtm.start": 1698330296703,
    "event": "gtm.js",
    "gtm.uniqueEventId": 1
  },
  {
    "event": "page_view",
    "page_view": {
      "location": "http://static.dev.gov.uk/templates/gem_layout.html.erb",
      "referrer": "",
      "title": "GOV.UK - The best place to find government services and information",
      "status_code": "200",
      "viewport_size": "829x658",
      "language": "en",
      "history": "false",
      "withdrawn": "false",
      "dynamic": "false"
    },
    "govuk_gem_version": "35.20.0",
    "gtm.uniqueEventId": 2
  },
  {
    "event": "gtm.dom",
    "gtm.uniqueEventId": 3
  },
  {
    "event": "gtm.load",
    "gtm.uniqueEventId": 4
  }
]
```

(Pageview Object exists when the user agent isn't the Smokey one)

andysellick (approved):
Seems like a good solution, well thought through. I wonder if it's worth documenting somehow, maybe just a short comment in the code?

AshGDS:
@andysellick Thanks, sounds good. I'm also wondering if it's worth putting a comment where the user agent is set in smokey.

andysellick:
@AshGDS that sounds like a good idea too 👍

PR #3682, opened 2023-10-24
Add data-ga4-form-no-answer-undefined to header search form

AshGDS:
## What
Adds `data-ga4-form-no-answer-undefined` to the header search form. When this attribute exists, it will send `text: undefined` when the form is empty, instead of `text: "No answer given"`.

## Why

https://trello.com/c/WtvEWiS3/592-placeholder-replace-no-answer-given-with-undefined-on-homepage-search-event


JamesCGDS (approved):
LGTM!

PR #3684, opened 2023-10-25
Add visually hidden text to navbar

patrickpatrickpatrick:
## What

Add visually hidden span to the navbar crown logo link if we are hiding the logo text. Add test to make sure the visually hidden text is being added.

## Why

If users change the styles of the navbar, then there is still fallback text for this link.

[Relevant Trello Card](https://trello.com/c/V1oeVTKL/2206-look-and-feel-accessibility-fix-alternative-text-for-govuk-logo-is-okay-with-aria-label-but-would-be-better-with-visually-hidden)

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
@@ -111,6 +111,11 @@
             label: logo_link_title,
           }
         } do %>
+          <% if hide_logo_text %>
+            <span class="govuk-visually-hidden">
+              <%= logo_text %>
+            </span>
+          <% end %>
           <span class="govuk-header__logotype">
             <!--[if gt IE 8]><!-->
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
```diff
@@ -111,6 +111,11 @@
             label: logo_link_title,
           }
         } do %>
+          <% if hide_logo_text %>
+            <span class="govuk-visually-hidden">
+              <%= logo_link_title %>
```
MartinJJones, line 116: From looking at the issue Anika raised, I think this should use `logo_text` instead, this would change the text from "Go to the GOV.UK homepage" to "GOV.UK".

MartinJJones (commented):
Just one small change to look at and double-check the requirement, if the change is required then the related test would also need to be updated.

MartinJJones (approved):

PR #3689, opened 2023-10-26
Add metadata inverse no padding option

andysellick:
## What
Adds a new option to the metadata component.

- when the inverse option is applied to the metadata component (to turn the text white to make it suitable for a dark background) this also includes extra spacing around the component
- have found two instances recently where this is not required and makes the page layout look odd
- adding the inverse_compress option, which allows the removal of this extra spacing when the inverse option is used

## Why
Needed for a change in `finder-frontend` (when there's an inverse header on search, see https://github.com/alphagov/finder-frontend/pull/3193) and `government-frontend` (about to raise a PR for [this page](https://www.gov.uk/guidance/the-highway-code) where application styles are overriding the appearance of the component).

## Visual Changes

Screenshot 2023-10-26 at 10 10 45

### DIFF
--- a/app/views/govuk_publishing_components/components/_metadata.html.erb
+++ b/app/views/govuk_publishing_components/components/_metadata.html.erb
@@ -9,6 +9,7 @@
 
   other ||= nil
   inverse ||= false
+  inverse_compress ||= false
 
   direction_class = ""
   direction_class = " direction-#{direction}" if local_assigns.include?(:direction)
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_metadata.html.erb
```diff
@@ -9,14 +9,21 @@
...
   direction_class = ""
   direction_class = " direction-#{direction}" if local_assigns.include?(:direction)
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   classes = %w(gem-c-metadata)
   classes << "direction-#{direction}" if local_assigns.include?(:direction)
-  classes << "gem-c-metadata--inverse" if inverse
+
+  if inverse && inverse_compress
+    classes << "gem-c-metadata--inverse" if inverse and inverse_compress
```
AshGDS, line 22: Does `inverse and inverse_compress` need to be checked here again if it already was on the line above?

Same for the elseif below it

AshGDS (approved):
Looks good :+1: added a small question

app/views/govuk_publishing_components/components/_metadata.html.erb
andysellick: Whoops, yes, meant to take that out - that was a hangover from an earlier approach. It's just the `if` statement now.

AshGDS (approved):

PR #3690, opened 2023-10-26
Fix issue with padding on navbar description

patrickpatrickpatrick:
## What

Revert margin on second item description in the navbar to 5px.

## Why

Was accidentally increased in changes to the large navbar.

### Before

Screenshot 2023-10-26 at 15 03 08

### After

Screenshot 2023-10-26 at 15 02 45


hannalaakso (approved):
LGTM 👍

PR #3694, opened 2023-10-30
Fix visual bug with contents list manually numbered headings

AshGDS:
## What

- In `whitehall`, a user can choose to manually number headings in their markdown, e.g. `## 1. Heading`.
- Our contents lists have some code which is meant to take these manually numbered headings, and generate a formatted contents list from them.
- However, when they format a heading like `## 1. Heading`, the space between `##` and `1.` was kept in the heading string. This caused some regex not to fire, which is used to format contents lists with heading numbers. This is because the regex expects the heading to start with a number, and not a space character.
- This caused a visual bug where contents lists were not formatted properly, since the regex was treating them as not being numbered, due to the space character causing the regex not to match against anything.
- I've fixed that in this PR by using a `.strip` version of the heading text for the regex check.
- The regex hasn't been modified to match spaces, as you don't want the space matched as part of the number. Otherwise, the space gets included when the `<span>` for it in the contents list is created.

## Why

- A bug came through on 2nd Line about this. I directed them to write their headings without a trailing space e.g. `##1. Heading` but we should fix this here, as our guidance says to include the space, e.g. `## 1. Heading`

### Before
image

### After
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix visual bug with contents list manually numbered headings ([PR #3694](https://github.com/alphagov/govuk_publishing_components/pull/3694))
+
 ## 35.21.0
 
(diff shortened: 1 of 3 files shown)


kentsanggds (approved):
LGTM

PR #3698, opened 2023-11-02
use asset helper for javascript

patrickpatrickpatrick:
- AssetHelper uses JS
- use asset helper for javascriupt


lib/govuk_publishing_components/app_helpers/asset_helper.rb
```diff
@@ -61,13 +105,43 @@ def all_component_stylesheets_being_used
         @all_component_stylesheets_being_used ||= []
       end
 
+      def all_component_javascripts_being_used
```
patrickpatrickpatrick, line 108: ah yep this wont work, ill need to check the individual lists whoops

patrickpatrickpatrick:
Currently not required for govuk-frontend v5. Might come back to later.

PR #3699, opened 2023-11-03
Make global relative links absolute

AshGDS:
## What

- There is an issue with preview pages on CSVs as they are hosted on a different domain to www.gov.uk - this breaks all the footer and header links as they are relative, so they expect the domain to always be www.gov.uk. This also breaks the search form and feedback form on these pages.
- We can fix this by appending the host name onto relative links to make them absolute. We use `ENV[VIRUTAL_HOST]` for development environments, and `Plek.new.website_root` for integration/staging/production
- I appreciate that this may not be the most elegant solution as I'm transforming data in the view which I know probably isn't the best. I was having issues transforming the data earlier on in the pipeline as the content hashes are frozen and the loops to modify data were getting quite nested.

## Why

- Our assets domain uses the GOVUK Header and Footer, e.g. https://assets.publishing.service.gov.uk/media/6425aa8c60a35e00120cb279/List_of_gov.uk_domains_as_of_30_March_2023.csv/preview
- Our header and footer use relative links
- This causes the header and footer links to be relative, for example our contact link is `https://assets.publishing.service.gov.uk/contact` instead of `https://www.gov.uk/contact`
- I noticed this also breaks `/search` and we noticed it broke the feedback from too, as they use relative `action` values.
- This was raised by someone on 2nd Line

## Visual Changes

None.


app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb
AshGDS, line 1: I had to make a variable called `link_href` in this file, as modifying `item[:href]` wasn't possible as the hash has been frozen.

lib/govuk_publishing_components/presenters/absolute_links_helper.rb
```diff
@@ -0,0 +1,18 @@
+module GovukPublishingComponents
+  module Presenters
+    class AbsoluteLinksHelper
+      def initialize(environment = GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment)
+        @environment = environment
+      end
+
+      def make_url_absolute_on_production(href)
```
Tetrino, line 8: We can remove the check for production here by using `Plek.new.website_root` to return the explicit domain of the environment. Do this instead of hardcoding and this fix is actually quite solid. It may need to be extended to cover the report-a-problem form too.

Tetrino (approved):
Looks great!

andysellick:
@AshGDS do you want to get this into the pending release?

AshGDS:
Going to wait before merging as want to ensure the changes don't break things, since it changes the action for a few forms in the header/footer and all the header/footer links

AshGDS:
Adding [DO NOT MERGE] as i want to test this branch on integration static first

AshGDS:
Worked on static in integration, will merge shortly.

PR #3700, opened 2023-11-06
Use govuk-frontend v5 [DO NOT MERGE]

patrickpatrickpatrick:
[Trello](https://trello.com/c/6LsKqHIO/2231-update-version-of-govuk-frontend-in-govukpublishingcomponents-l)

## What

Make changes in govuk_publishing_components to allow us to update to govuk-frontend v5.

### Asset Changes

- Update references to assets (config, Javascript, CSS, fonts, images) in govuk-frontend
- Remove references to the crown fallback image
- Remove references to 
  - `assets/images/govuk-apple-touch-icon.png`
  - `assets/images/govuk-apple-touch-icon-152x152.png`
  - `assets/images/govuk-apple-touch-icon-167x167.png`
  - `assets/images/govuk-apple-touch-icon-180x180.png`
  - `assets/images/govuk-mask-icon.svg`
- Add references to
  - `assets/manifest.json`
  - `assets/images/govuk-icon-180.png`
  - `assets/images/govuk-icon-mask.svg`
- Change logo in-line svg to be the new svg that includes GOV.UK
  - Rework how the logo is included in `layout_super_navigation_header`

### Javascript Changes

- Add necessary script to append new required class to body for govuk-frontend
- Update initialisation method to be the same as the new govuk-frontend approach for any JS module that uses `GOVUK.Modules`
- Update tests to use new initialisation method for any of the JS modules I had to update
- Remove import of `details.js` from govuk-frontend as the `details` component no longer uses Javascript
- Use [Terser](https://github.com/ahorek/terser-ruby) instead of [Uglifier](https://github.com/lautis/uglifier) to compress Javascript

### Stylesheet Changes
- Remove `$legacy` from the `govuk-colour` mixin (this has been removed from govuk-frontend)
- Remove `$govuk-new-link-styles` feature flag (new link styles now default, feature flag removed)

### Component specific changes

- Components using `tag` component now `titleize` the tag text to adopt the new tag design
   * Update the `header_logo` and `phase_banner` components
   * Update tests of `layout_header` and `phase_banner`
   * Tweak text colour of `gem-c-environment-tag` to ensure contrast (default tag colour is now black).
- `warning_text` component no longer uses the `govuk-warning-text__assistive` class
  * Replaced all instances of `govuk-warning-text__assistive` with `govuk-visually-hidden` as specified in the release notes
  * Update tests of warning_text component

### Config changes
- Add new configuration option `use_es6_components`
  * If true then `layout_for_public` will try to load the ES6 (govuk-frontend) using components from the file `es6-components.js`

### Doc updates
- Add suggested way to include JS in Rails applications in light of the browser target change that has occurred in govuk-frontend v5.

## What I Haven't Changed

This PR is for things that **have** to change. To keep the PR size as manageable as possible (sorry, it is still very long) there are pieces of work that I think should be separate PRs:

- If a module isn't using the `GOVUK.Modules` object (i.e. most of the analytics modules and most of the Ga4 modules) then it haven't been updated to use the new initialisation method introduced by govuk-frontend v5. 
- Opening the component guide in browsers that don't support `type="module"` will result in some ungraceful erroring. This is because in this PR there isn't a way to individually load the JS that contains ES6 code. In frontend rendering applications I have established a pattern to follow to do this and it could be implemented in them gem also (but in a separate PR). 

## Why

There are several breaking changes in the latest version of gouk-frontend which requires us to update govuk_publishing_components. [Please read the full release notes](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0-beta.2). The changes that are specifically being addressed here are:

- Update package file paths
- Remove Internet Explorer 8 stylesheets, settings and mixins
- Remove the fallback GOV.UK crown logo from your HTML
- Removal of Javascript polyfills for Internet Explorer (IE) 8-11
- Update the <script> snippet at the top of your <body> tag
- Check your tags align with design changes to the Tag component
- New link styles are now enabled by default
- Update the HTML for warning text

## Visual Differences

Percy: https://percy.io/55ba5e1a/govuk_publishing_components/builds/33548508/changed/1838427854?browser=firefox&browser_ids=57&subcategories=unreviewed%2Cchanges_requested&viewLayout=overlay&viewMode=new&width=1280&widths=375%2C1280

### Layout Header

The reason for the visual change here is that `.gem-c-header__logotype` was causing the line-height to increase, now this class is removed, the total height of the layout header has decreased from 72.66px to exactly 70px.

#### Before

layout-header-before
layout-header-before-sizes

#### After

layout-header-after
layout-header-after-sizes

### Public Layout

Percy is also showing a visual change in public layout wherever the layout header component is used, this is to be expected following the change mentioned above.

| Before | After |
| --- | --- |
| image |  image |

### Phase Banner

The phase banner will now use the new design included in v5 of the govuk-frontend.

| Before | After |
| --- | --- |
| 68e7957767c852045d2522b12d04913705f749b970d165415e8d84fe60f1726b |  b3460582d31935951cd58afacd27118567d2a311fabd59ba12e8f2a5ab5a430d |

### Form inputs

The visual change show in Percy for the form input component is coming from a fix in version 5.1.0 of govuk-frontend - https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#510-feature-release

| Before | After |
| --- | --- |
| form-input-prefix-before | form-input-prefix-after |


app/assets/javascripts/govuk_publishing_components/components/accordion.js
```diff
@@ -1,10 +1,11 @@
 /* global nodeListForEach */
 //  = require ../vendor/polyfills/common.js
 // This component relies on JavaScript from GOV.UK Frontend
-// = require govuk/components/accordion/accordion.js
+// = require govuk/components/accordion/accordion.bundle.js
+
 window.GOVUK = window.GOVUK || {}
 window.GOVUK.Modules = window.GOVUK.Modules || {}
-window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion;
+window.GOVUK.Modules.GovukAccordion = window.GOVUKFrontend.Accordion.Accordion;
```
domoscargin, line 8: :bug: Hi @patrickpatrickpatrick - you've helped us uncover a bug here with our named exports. We'll hopefully patch this shortly to get rid of this odd duplication (so this import will just be `window.GOVUKFrotnend.Accordion`).

patrickpatrickpatrick: Ah great, that makes sense!

colinrotherham: Thanks again for this

We hadn't realised that Rollup handles naming for us now

Fixed in https://github.com/alphagov/govuk-frontend/pull/4444 and will be in the next beta

patrickpatrickpatrick:
The changes made here will mean that the component guide will no longer use Javascript correctly for certain components when opened in Internet Explorer 11. Currently this will results in errors in the console. Although this isn't ideal, personally I think this could be sorted out in separate branch? I'd assume the component guide is mostly used by developers and those developers won't be using Internet Explorer 11. It should be fixed but I don't think it's a pressing issue.

andysellick:
> The changes made here will mean that the component guide will no longer use Javascript correctly for certain components when opened in Internet Explorer 11. Currently this will results in errors in the console (in IE 11).

Interesting - can you explain more about why this will cause console errors? I thought the introduction of JS modules meant that it would fail silently for IE11. Is there something different about the component guide that causes this?

patrickpatrickpatrick:
> > The changes made here will mean that the component guide will no longer use Javascript correctly for certain components when opened in Internet Explorer 11. Currently this will results in errors in the console (in IE 11).
> 
> Interesting - can you explain more about why this will cause console errors? I thought the introduction of JS modules meant that it would fail silently for IE11. Is there something different about the component guide that causes this?

The issue is the scripts being evaluated at all by IE11. The modules need to be wrapped in a `type="module"` tag for IE11 to 'skip' over them. I have actually started to implement how we would do this in a PRs for [static](https://github.com/alphagov/static/pull/3182) and [collections](https://github.com/alphagov/collections/pull/3436) and it wouldn't be difficult to implement. It's just if it should be included in this initial PR or should be added in a seperate one (to keep this one as manageable as possible).

andysellick:
> It's just if it should be included in this initial PR or should be added in a separate one (to keep this one as manageable as possible as it is already quite long).

Thanks for the explanation, very helpful. Sounds sensible to put that fix in a separate PR, agreed. Shout if you want any help!

colinrotherham:
@patrickpatrickpatrick Probably worth a mention, but in readiness of `type="module"` we had some great feedback from our beta releases in https://github.com/alphagov/govuk-frontend/discussions/4389#discussioncomment-7402623 to look at

Reminders that JavaScript code in `<script type="module">`

* runs in ["strict mode"](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode) but without `"use strict"`
* runs before [DOMContentLoaded](https://developer.mozilla.org/en-US/docs/Web/API/Document/DOMContentLoaded_event) like [`<script defer>`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer)

The script contents itself doesn't need ES module exports, but good to catch any issues early such as:

>Strict mode applies to entire scripts or to individual functions. It doesn't apply to [block statements](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/block) enclosed in {} braces; attempting to apply it to such contexts does nothing. [eval](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/eval) code, [Function](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Function/Function) code, [event handler](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes#event_handler_attributes) attributes, strings passed to [setTimeout()](https://developer.mozilla.org/en-US/docs/Web/API/setTimeout), and related functions are either function bodies or entire scripts, and invoking strict mode in them works as expected.

Similarly the old Rollup UMD factory code had issues with `this` not being `window` by default:

* https://github.com/alphagov/govuk-frontend/issues/3972#issuecomment-1699418524

To avoid any surprises by adding `<script type="module">` later rather than sooner

patrickpatrickpatrick:
As we need to switch to Terser to compress the JS (because of govuk-frontend v5), this PR will mean you can now use ES6 in govuk_publishing_components. I would advise against it currently however as there's no target specified so it might mean you end up including JS features that don't work in all browsers that support `type="module"`.

patrickpatrickpatrick:
This will remain a draft until govuk-frontend v5 is released.

app/assets/javascripts/govuk_publishing_components/components/accordion.js
colinrotherham: @patrickpatrickpatrick Released today in `beta.2`

* https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0-beta.2

patrickpatrickpatrick: Thanks, I have updated the code :)

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -79,11 +79,10 @@
...
     <link rel="shortcut icon" href="<%= asset_path 'favicon.ico' %>" type="image/x-icon" />
-    <link rel="mask-icon" href="<%= asset_path 'govuk-mask-icon.svg' %>" color="#0b0c0c">
-    <link rel="apple-touch-icon" sizes="180x180" href="<%= asset_path "govuk-apple-touch-icon-180x180.png" %>">
-    <link rel="apple-touch-icon" sizes="167x167" href="<%= asset_path "govuk-apple-touch-icon-167x167.png" %>">
-    <link rel="apple-touch-icon" sizes="152x152" href="<%= asset_path "govuk-apple-touch-icon-152x152.png" %>">
-    <link rel="apple-touch-icon" href="<%= asset_path "govuk-apple-touch-icon.png" %>">
+    <link rel="mask-icon" href="<%= asset_path 'govuk-icon-mask.svg' %>" color="#0b0c0c">
+    <link rel="apple-touch-icon" sizes="180x180" href="<%= asset_path "govuk-icon-180.png" %>">
+    <link rel="apple-touch-icon" sizes="167x167" href="<%= asset_path "govuk-icon-192.png" %>">
+    <link rel="apple-touch-icon" sizes="152x152" href="<%= asset_path "govuk-icon-512.png" %>">
```
colinrotherham, line 85: Ideally drop these lines and add `manifest.json` if you can:

```suggestion
```

<br>

The `sizes` attributes don't match up as we've replaced older Apple-specific icon sizes:

* 167×167 @1x — Home screen (iPad Pro)
* 152×152 @1x — Home screen (iPad, iPad mini)

With more [generic "maskable" icons in **manifest.json**](https://web.dev/articles/maskable-icon) with different sizes:

* 180×180
* 192×192
* 512x512

You'll see the web manifest in action when adding to home screens or the Dock

Add to Dock

app/views/govuk_publishing_components/components/layout_for_public/_header_logo.html.erb
```diff
@@ -0,0 +1,46 @@
...
+
+<% unless hide_logo_text %>
+    <svg
+      focusable="false"
+      role="img"
+      class="govuk-header__logotype"
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 152 30"
+      height="30"
+      width="152"
```
colinrotherham, line 17: This missed the cut off for `govuk-frontend@5.0.0-beta.2` but is an [important VoiceOver fix](https://github.com/alphagov/govuk-frontend/pull/4497)

Thanks @patrickpatrickpatrick

```suggestion
      width="152"
      aria-label="GOV.UK"
```

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -1,13 +1,17 @@
...
-      </span>
-    </span>
+    <svg
+      focusable="false"
+      role="img"
+      class="govuk-header__logotype"
+      xmlns="http://www.w3.org/2000/svg"
+      viewBox="0 0 152 30"
+      height="30"
+      width="152"
```
colinrotherham, line 10: Probably the [`aria-label` here too](https://github.com/alphagov/govuk_publishing_components/pull/3700/files/a80b5a65c327cfeded2e6cec43b6266e4bfcde7e..83796778727c9d77b63ff2fe39bb558242fadcdb#r1409591225)

davidtrussler:
I think there's another update to package file paths that is needed, to add the `dist` directory to the `app/assets/config/govuk_publishing_components_sassc-rails_manifest.js` file as below: 
```
//= link_tree ../../../node_modules/govuk-frontend/dist/govuk/assets/images
//= link_tree ../../../node_modules/govuk-frontend/dist/govuk/assets/fonts
```

MartinJJones:
Closing as we have a new PR open to upgrade to 5.1.0 - https://github.com/alphagov/govuk_publishing_components/pull/4041

PR #3705, opened 2023-11-07
Enable GA4 by default on components

andysellick:
## What
Switch the `ga4_tracking` flag to enable tracking on some components for a `disable_ga4` flag, and enable tracking by default on these components.

## Why
Until now our strategy for GA4 tracking on components was to include a `ga4_tracking` flag to enable tracking only on specific component instances. Now that most of the tracking is in place we need to ensure that if components are added to pages in future, tracking is included. 

However we also need to account for some instances where tracking may not be required on these components, so we replace this option with a `disable_ga4` flag.

## Visual Changes
None.

Trello card: https://trello.com/c/Xwp7dgJN/294-tracking-auditing-tracking-by-default

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Enable GA4 by default on components ([PR #3705](https://github.com/alphagov/govuk_publishing_components/pull/3705))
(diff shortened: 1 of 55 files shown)


andysellick:
@AshGDS just asked for your review but realised there's more to do on this PR so hold off for now 👍

andysellick:
@AshGDS okay all good to review now 👍

AshGDS (approved):
Looks good, thanks for doing this :+1:

PR #3706, opened 2023-11-07
Create new political meta tags for GA4

AshGDS:
## What

- We have existing `political-status` and `publishing-government` meta tags.
- These existing meta tags rely on a `government` object, and `political` value being present at the top level of the content item, which is actually now a [deprecated content item JSON format](https://github.com/alphagov/publishing-api/blob/c702d08c9f4e90a4792ac7f9467740b0207bbd01/content_schemas/formats/html_publication.jsonnet#L45).
- On certain pages such as https://www.gov.uk/government/news/new-measures-to-tackle-anti-social-behaviour the `political` value exists, but the `government` object is nested in the content item. Therefore the existing `political-status` and `publishing-government` meta tags do not render.
- For GA4, we want a way to still grab the `political-status` and `publishing-government` if this `government` object is nested in the content item.
- Therefore, I've created two new meta tags, `ga4-political-status` and `ga4-publishing-government`, which account for the nested `government` object.
- These new meta tags are then used by the Pageview tracker.

In order for the meta tag to render:
1. `political` needs to be `true` in the content item
2. the `government` object needs to exist in the content item
3. In the `details` of `government`, the government should not be the current government in power (i.e. the value will be `current: false`). This is because the banner "This was published under [example] government" only appears if that government is no longer in power, so we're only adding the meta tag for pages that refer to a previous government.

## Why

- https://trello.com/c/UPIXwaQF/713-improve-political-status-meta-informations-sent-with-page-view
- It should also resolve this blocked card: https://trello.com/c/JZ8Gehg0/354-page-view-enhancement-add-missing-publishing-government-meta-tag-to-some-content-items-that-should-have-it - I have also addressed the concerns raised in the investigation from that card in the card's comment section.

## Testing
- If you were to run local `static`, with a local `govuk_publishing_components`, and then run `government-frontend` with local `govuk_publishing_components` and local `static`, on [this page]( https://www.gov.uk/government/news/new-measures-to-tackle-anti-social-behaviour ) you would have `political_status` and `publishing_government` in your pageview object:
image

## Visual Changes

None.


lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -48,6 +49,8 @@ def add_core_tags(meta_tags)
         primary_publisher = content_item.dig(:links, :primary_publishing_organisation)
         primary_publisher = primary_publisher.first[:title] unless primary_publisher.blank?
         meta_tags["govuk:primary-publishing-organisation"] = primary_publisher unless primary_publisher.blank?
+
+        # Some browse topics are nested in the content item, we want to grab these for GA4 tracking.
```
AshGDS, line 53: Unrelated to the PR - just added this comment to improve the code for the GA4 meta tags.

andysellick (commented):
Code changes look sensible to me. My hesitation around this work has always been that I've wanted a backend dev to check what we're doing is safe and reasonable - can we get @gclssvglx or @Rosa-Fox to review this as well?

beccapearce (approved):

PR #3716, opened 2023-11-14
Add checking what components use govuk-frontend to components audit

patrickpatrickpatrick:
## What

Adds checking to see what components use govuk-frontend to the components audit.

## Why

Implemented this to double check that I wasn't missing out any component JS when making changes to applications to support govuk-frontend v5. Thought it could be useful if this was a standard feature in the auditing, so I have opened this PR.

### Before

Screenshot 2023-11-14 at 14 50 07

### After

Screenshot 2023-11-14 at 14 49 55

### DIFF
--- a/app/models/govuk_publishing_components/audit_comparer.rb
+++ b/app/models/govuk_publishing_components/audit_comparer.rb
@@ -277,6 +277,7 @@ module GovukPublishingComponents
 
         results << {
           name: component_name,
+          govuk_frontend: component[:uses_govuk_frontend],
           count: locations.length,
           locations:,
         }
--- a/app/models/govuk_publishing_components/audit_components.rb
+++ b/app/models/govuk_publishing_components/audit_components.rb
@@ -118,6 +118,11 @@ module GovukPublishingComponents
         else
           @component_numbers[type.to_sym] += 1
           details["#{type}_exists".to_sym] = true
+
+          if type == "javascript" && uses_govuk_frontend?(file)
+            details[:uses_govuk_frontend] = true
+          end
+
           details["#{type}_lines".to_sym] = count_lines_in(file)
           details["#{type}_link".to_sym] = get_asset_link(type, component)
         end
@@ -128,6 +133,10 @@ module GovukPublishingComponents
 
     def count_lines_in(file)
       File.read(file).each_line.count
+    end
+
+    def uses_govuk_frontend?(file)
+      File.read(file).match(/require govuk\/components/)
     end
 
     def clean_files(files, replace)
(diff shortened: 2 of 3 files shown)


andysellick:
Someone else is adding stuff to the auditing tools 🎉 

Bit confused as to the purpose behind this - why do we need to know which applications use components that use govuk-frontend JS? I thought if we included a component JS in an application that included any govuk-frontend JS as well, or is that going to change?

patrickpatrickpatrick:
> Someone else is adding stuff to the auditing tools 🎉
> 
> Bit confused as to the purpose behind this - why do we need to know which applications use components that use govuk-frontend JS? I thought if we included a component JS in an application that included any govuk-frontend JS as well, or is that going to change?

In the approach I'm doing to ensure govuk-frontend isn't evaluated/run by browsers that don't support it (browsers that don't support `type="module"`, I'm moving the components that use the [ES6 JS from govuk-frontend to a different file](https://github.com/alphagov/collections/pull/3436/files#diff-f9e1c67e1a48706fdcd2b58e50305787c35ac755ec1de188f0c21029ca6ac35b). This file will then be [imported using a `type="module"` tag](https://github.com/alphagov/collections/pull/3436/files#diff-f43fe075643e681b2c01c2f853bb0c4299d135b47fcbd4da96890d521c49e3eb) so that if browser doesn't support the tag then they won't run the modules.

This information just lets me know what components that are using ES6 need to be moved to a separate file and where they are so I can test them. If we are dividing the upgrading of applications between the teams that are responsible for maintaining them, then I thought others might find this information useful.

andysellick:
In that case can I suggest an alternative approach? Rather than add a whole new section, could you modify the existing 'gem components used by applications' to somehow highlight those components using `govuk-frontend`? Something like this maybe?

Screenshot 2023-11-14 at 11 58 47

It also might be useful to have this information included in the main 'component files' section, although there's limited width there already so might not be practical.

patrickpatrickpatrick:
> In that case can I suggest an alternative approach? Rather than add a whole new section, could you modify the existing 'gem components used by applications' to somehow highlight those components using `govuk-frontend`? Something like this maybe?
> 
> Screenshot 2023-11-14 at 11 58 47
> 
> It also might be useful to have this information included in the main 'component files' section, although there's limited width there already so might not be practical.

Ah yep, that makes more sense actually! I'll make that change

andysellick:
@patrickpatrickpatrick would you like a review of this PR?

patrickpatrickpatrick:
Ah yes please @andysellick ! Think this will be useful for anyone upgrading to v5 of govuk-frontend.

andysellick (commented):
Code changes look good, would be good to get a test included for this if possible. There's a load of test files in `/spec/dummy_gem/` that could be adapted, see the other auditing tests.

PR #3717, opened 2023-11-14
Change GA4 video tracking duration handling

andysellick:
## What / why
- for some reason we're getting differing numbers back from the YouTube API when requesting video durations
- solution is to just ask for it once, then store it along with the percentage positions calculated when videos are loading
- in terms of efficiency, it's also therefore sensible to get and store the video title and URL at the start as well, instead of requesting them each time an event fires during playback

## Visual Changes
None.

Trello card: https://trello.com/c/Zj359UYH/727-fix-video-duration-rounding-issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change GA4 video tracking duration handling ([PR #3717](https://github.com/alphagov/govuk_publishing_components/pull/3717))
+
 ## 35.21.4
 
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Looks good :+1: I am guessing this is an experiment into fixing it rather than the definitive fix, as we didn't figure out what the different durations was caused by?

andysellick:
It's a fix that I think is enough of a fix. Given that the returned value for duration only varied by around a second, it seems reasonable to just take the first answer and not worry if it's fractionally wrong. Also the code is a bit more efficient now in terms of load while playing, so I'm happy with it.

PR #3720, opened 2023-11-17
Add alt/option click tracking to GA4 link tracker

AshGDS:
## What

- When the `alt` or `option` key were held down while a link was clicked, it was being tracked as a `primary click`.
- This PR changes that, so that the click is tracked as an `alt/option` click.

## Why

https://trello.com/c/36r3mAUe/734-option-clicks-currently-coming-through-with-method-primary-click-switch-to-option-click

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
@@ -209,6 +209,92 @@ describe('GA4 link tracker', function () {
     })
   })
 
+  describe('different types of clicks', function () {
+    beforeEach(function () {
+      attributes = {
+        event_name: 'navigation',
+        type: 'a link',
+        index_link: 1
+      }
+      expected.event = 'event_data'
+      expected.event_data.type = 'a link'
+      expected.event_data.external = 'false'
+      expected.event_data.index = {
+        index_link: 1,
+        index_section: undefined,
+        index_section_count: undefined
+      }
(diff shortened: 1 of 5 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
```diff
@@ -209,6 +209,92 @@ describe('GA4 link tracker', function () {
...
+      element.textContent = linkText
+      expected.event_data.text = linkText
+      expected.event_data.url = link
+      expected.event_data.link_path_parts['1'] = link
+    })
+
+    it('tracks primary clicks on a single link', function () {
+      expected.event_data.method = 'primary click'
+      initModule(element, true)
+      expect(window.dataLayer[0]).toEqual(expected)
```
andysellick, line 241: Shouldn't this test dispatch a click event? I'm confused - it's clearly passing, though.

andysellick (approved):
Nice. Just one question, otherwise happy to approve.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.spec.js
AshGDS: @andysellick I used the `initModule` function here, which when passed `true` clicks the element that is passed through

andysellick: Ahhh of course. Nice.

PR #3723, opened 2023-11-20
Make data-ga4-set-indexes ignore links with no href

AshGDS:
## What

Make data-ga4-set-indexes ignore links with no href

## Why

- Prevents future issues if invalid links are introduced into the HTML
- https://trello.com/c/UlhiLIok/732-update-index-code-to-count-links-if-they-also-have-href-or-text-values

## Visual Changes

None.


andysellick (approved):
👍

PR #3725, opened 2023-11-21
Consolidate GA4 schema calls

andysellick:
## What

- creates new function in core code 'applySchemaAndSendData', which de-duplicates calls to the schema generation code in trackers
- leaves existing sendData function in place as not always used in this way

## Why
Optimising code.

## Visual Changes
None.

Trello card: https://trello.com/c/sI76EaNX/655-investigate-shortening-calls-to-schema-code

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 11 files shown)


AshGDS (approved):
Looks good :+1:

PR #3729, opened 2023-11-22
Change action link component options

andysellick:
## What
Change some of the options on the action link component, specifically:

- remove `data` in favour of the `data_attributes` option provided by the component wrapped helper

Note that the following PRs must be merged before this one:

- https://github.com/alphagov/frontend/pull/3866
- https://github.com/alphagov/collections/pull/3449

Also remove a load of options on the action link component that aren't used anymore, for example the NHS icon version.

## Why
The component wrapper helper already does this stuff, duplicate code isn't necessary.

There are way too many options on this component and many of them aren't used anymore. For reference the action link is (apparently) only used in two applications now (once the two PRs above have been merged)

Screenshot 2023-11-22 at 11 48 31

## Visual Changes
None.


spec/components/action_link_spec.rb
```diff
@@ -171,15 +108,4 @@ def component_name
     )
     assert_select ".gem-c-action-link--light-text"
   end
-
-  it "has data attributes if data attributes are passed in" do
```
maxgds, line 175: Overall this is a great bit of cleanup, I'm a bit puzzled by the removal of this test, though? Could it not be rewritten and remain valuable? Or do similar enough tests exist elsewhere to invalidate the need for it?

andysellick: Thanks for all the reviews @maxgds, much appreciated. This test should already be covered by tests in the component wrapper helper, which is now applied to this component (replacing the `data` option with `data_attributes`) see - https://github.com/alphagov/govuk_publishing_components/blob/main/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb

maxgds: Ahh, makes sense, thanks!

maxgds (approved):

spec/components/action_link_spec.rb
andysellick: You've got me thinking though - while part of the value of the component wrapper is to reduce what are effectively duplicate tests across components, it still might be good to have something that confirms that the component wrapper has been installed correctly and is working, and continues to work. I've got some rough ideas, I'll put together a PR soon.

PR #3730, opened 2023-11-24
Decommission GA4 link_path_parts and splitting of `taxonomy_all` and `taxonomy_all_ids`

AshGDS:
## What

- Remove `link_path_parts` from our GA4 code.
- Revert `taxonomy_all` and `taxonomy_all_ids` to just grab the meta tag value, instead of splitting it into parts.

## Why

- GA4 now supports values up to 500 characters in length, so we don't need to split long strings anymore.
- https://trello.com/c/EDsICA2E/733-decommission-link-path-parts-solution
- https://trello.com/c/31nPh4KO/736-re-join-taxonomyall-parts
- https://trello.com/c/dpoGrQ29/740-re-join-taxonomyallids-parts

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-schemas.js
```diff
@@ -60,9 +59,9 @@
     schema.event = eventAttribute
     // exceptions should be inserted without checking for sub parameters
     // exceptions should only come from code, not directly from data attributes
-    // e.g. link_path_parts is generated by JS and contains potentially non-unique
+    // e.g. an object generated by JS and contains potentially non-unique
     // sub parameter names i.e. '1', '2'
-    var exceptions = ['link_path_parts']
+    var exceptions = []
```
andysellick, line 64: While this might be useful at some point, unless we're about to need it or can think of a good reason to keep it, we should probably get rid of it entirely. What do you think?

andysellick (commented):
Only one small thing, otherwise looks good 👍 

Sad to see the Mary Poppins references go, but this is nice too:
Screenshot 2023-11-27 at 16 35 46

AshGDS:
@andysellick After rebasing I can't see the `exceptions` code anymore - was it removed as part of one of your changes?

andysellick:
> @andysellick After rebasing I can't see the `exceptions` code anymore - was it removed as part of one of your changes?

Ah yes I took it out in https://github.com/alphagov/govuk_publishing_components/pull/3718

andysellick (approved):

AshGDS:
Thanks @andysellick

PR #3731, opened 2023-11-24
Allow GA4 to be disabled via query string parameters

AshGDS:
## What

- Some previous work was done by myself to try and prevent Smokey tests from sending GA4 data to production
- This didn't work, I assume because the user agent Smokey tried to set wasn't working properly
- Instead, Smokey tests send a `smokey_cachebust` value.
- We can use this to stop data being sent to the GA4 production dashboard.
- Instead, we can send the data to a `fakeDataLayer` array. We can use this to test GA4 is working on production, without actually sending data to GA4.
- Therefore, if the page URL contains the `smokey_cachebust` or `disable_ga4` query string parameters, our GA4 objects will push to `window.fakeDataLayer` instead of `window.dataLayer`
- I added the `disable_ga4` query parameter as it can be used by other testing apps, such as the analytics robot.
- Note: I moved `getSearch()` out of the `PageViewTracker` and put it into `Ga4 Core` since they both make use of this function now.

## Why

- We need to stop Smokey tests sending GA4 data to production, and we can use this fake dataLayer to write Smokey tests that check GA4 is working
- We can use this to stop the analytics_robot, or other tests we do, from sending data to the GA4 production dashboard
- https://trello.com/c/TnMU9ZPm/709-investigate-how-to-filter-smokey-tests#comment-655f61ca6031eb83d2cef1cd

## Visual Changes

None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -52,6 +47,20 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
         }
         console.info(JSON.stringify(data, null, ' '))
       }
+
+      var ga4DisablingQueryStrings = ['smokey_cachebust', 'disable_ga4']
```
andysellick, line 51: I think you might be able to simplify this section with a regex, `[?&](disable_ga4|smokey_cachebust){1}` seems to work.

andysellick (commented):
Looks good, one small optimisation suggestion.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
AshGDS: @andysellick Great spot, I always forget about regex solutions. :+1:

AshGDS:
Thanks @andysellick - should be ready for another review :+1:

andysellick (approved):
Nice.

PR #3734, opened 2023-11-28
[DO NOT MERGE] Add a test for the component wrapper's presence

andysellick:
## What / why

- one of the aims of the component wrapper is to reduce duplicate code by providing a central place where various attributes are added to components
- this should reduce the number of tests we have e.g. once the wrapper is installed, no need to test that data attributes can be passed to that component
- but still need some confidence that the wrapper is installed correctly, so adding this option
- now components can have a single test to check that the wrapper is installed and working as expected (see the documentation change for details)

TODO: finish adding to remaining components that use the wrapper.

## Visual Changes
None.


andysellick:
This has been hanging around for a while and following discussion it might not be the right approach - for now I think we'll have to rely on developer skill to ensure they've implemented this helper correctly.

PR #3741, opened 2023-11-30
 Prevent government_frontend test failure if meta tag key doesn't exist

AshGDS:
## What

- To generate a new meta tag for GA4, we were diving into a nested value within the content_item JSON: `government[:details][:current]`
- `government_frontend` tests generate random but valid schemas to test against. Therefore, the test schema sometimes would contain the `government` object, but the `government` object wouldn't contain `[:details][:current]` and therefore crash.
- The error in `government_frontend` is: 
```Ruby
Minitest::UnexpectedError:         ActionView::Template::Error: undefined method `[]' for nil:NilClass
            app/views/layouts/application.html.erb:63` 

``` 
- and `application.html.erb:63` was the call to render `meta_tags.html.erb.` in `government-frontend`. I used puts statements within `meta_tags.rb` in this repo  to determine which line in our GA4 meta tag code was crashing during the `government-frontend` test.
- This PR should fix the issue. It uses `.dig` in Ruby instead, which will return `nil` instead of crashing if the key doesn't exist.
-  I tested `government_frontend` against this local branch, and forced the test to give me the same erroring content schema each time (instead of giving random ones) and the `government-frontend` test passes now. I also tested it multiple times against random schemas and have not experienced any test failures.
- I also modified the meta tag code slightly as I noticed the tests for it were passing but were false-positives. This was because I was using `.merge` to try and remove values in the existing content item, but this was actually not working reliably. Therefore it was better to use `.delete` when removing values, and to just override keys that needed changing. This surfaced some issues with the meta tag code which are now fixed, and the code is simpler as a result.

## Why

- The current code caused flaky tests in `government-frontend`, because the tests check randomly generated but valid content schemas (that are different to ones on production), so some schemas didn't have the right data we needed and therefore the tests would fail when encountering that version of a schema.
- https://github.com/alphagov/government-frontend/pull/3000
- https://github.com/alphagov/government-frontend/pull/3001

## Visual Changes

None.

### DIFF
--- a/lib/govuk_publishing_components/presenters/meta_tags.rb
+++ b/lib/govuk_publishing_components/presenters/meta_tags.rb
@@ -92,13 +92,12 @@ module GovukPublishingComponents
       def add_ga4_political_tags(meta_tags)
         government = content_item.dig(:links, :government, 0)
 
-        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government[:details]'
+        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government.dig(:details)'
         if government && details[:political]
-          meta_tags["govuk:ga4-political-status"] = government[:details][:current] ? "political" : "historic"
-
-          government_title = government[:title]
-          if government_title && !government[:details][:current]
-            meta_tags["govuk:ga4-publishing-government"] = government_title
+          current_government = government.dig(:details, :current)
+          unless current_government
+            meta_tags["govuk:ga4-political-status"] = "historic"
+            meta_tags["govuk:ga4-publishing-government"] = government[:title] if government[:title]
           end
         end
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Prevent government_frontend test failure if meta tag key doesn't exist ([PR #3741](https://github.com/alphagov/govuk_publishing_components/pull/3741))
+
 ## 36.0.1
 
 * Use component wrapper on option select ([PR #3738](https://github.com/alphagov/govuk_publishing_components/pull/3738))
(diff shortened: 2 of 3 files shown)


lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -92,13 +92,12 @@ def add_political_tags(meta_tags)
...
-        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government[:details]'
+        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government.dig(:details)'
         if government && details[:political]
-          meta_tags["govuk:ga4-political-status"] = government[:details][:current] ? "political" : "historic"
-
-          government_title = government[:title]
-          if government_title && !government[:details][:current]
-            meta_tags["govuk:ga4-publishing-government"] = government_title
+          current_government = government.dig(:details, :current)
+          unless current_government
```
andysellick, line 98: I'm struggling to get my head round this a bit so bear with me. This seems like quite a significant change? It used to always set the `ga4-political-status` meta tag (either to political or historic) but now it seems like it might not set it at all, if there's a `current_government`. And it can only ever be `historic`, not `political`?

AshGDS: @andysellick No worries - It no longer gets set to `political` at all with this change since that's only relevant for pages that are published during the current government. Because the PAs have only requested to have the information of previous governments I have removed it from being set. 

Political would get set if `current` is `true` for the government in the content item.  In the old meta tag, it was setting the meta tag to `political` based on that:

```
if details[:political]
  political_status = details[:government][:current] ? "political" : "historic"
  end
```

lib/govuk_publishing_components/presenters/meta_tags.rb
```diff
@@ -92,13 +92,12 @@ def add_political_tags(meta_tags)
       def add_ga4_political_tags(meta_tags)
         government = content_item.dig(:links, :government, 0)
 
-        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government[:details]'
+        # political: true/false is in a different place to current: true/false, which is why we have 'details' and 'government.dig(:details)'
```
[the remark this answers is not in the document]
AshGDS: @andysellick Populating the meta tag relies on two values existing:

`details[:political]: (true/false)`
`government[:details][:current]: (true/false)`

And the comment was trying to explain why the code is referring to two different `details` objects.

PR #3743, opened 2023-12-01
Change "cross service header" spec filename

jon-kirwan:
## What

Change "cross service header" spec filename.

## Why

To follow the correct conventions and that its discovered and run within the test suite.

## Visual Changes

None.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change "cross service header" spec filename ([PR #3743](https://github.com/alphagov/govuk_publishing_components/pull/3743))
```
andysellick, line 12: I don't think this warrants a changelog entry - it's more of a maintenance change, and it's not relevant to applications since the tests are confined to the gem.

andysellick (approved):

CHANGELOG.md
jon-kirwan: Yep see what you mean. I'll remove the entry. Thanks for the approval.

PR #3744, opened 2023-12-01
Add class to super nav header when menu is open

MartinJJones:
## What
- Add/Remove the `gem-c-layout-super-navigation-header--menu-open` class on the Super Navigation Header depending on the menu state and add related tests for this functionality
- Update the HTML stub in the super navigation header test file to include the required header element
- Add related tests for this functionality

## Why
By adding the new `gem-c-layout-super-navigation-header--menu-open` class to the header element when the super navigation menu is open, we can style sibling elements based on the menu state.

This will be used on the homepage to increase spacing between the menu and GOV.UK heading text when the menu is open - https://github.com/alphagov/frontend/pull/3873
 
[Trello card](https://trello.com/c/FQl2jTYc/2241-further-tweaks-to-homepage-post-launch-s)

### Further info on approach

- The ideal solution would have been to use purely as CSS based approach, but I could not find any good options to achieve the required design that works in all browsers.
- The [module reference is added to the `nav` element](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_super_navigation_header.html.erb#L155) in the layout-super-navigation-header template as data-module="super-navigation-mega-menu", this means it is not possible to query the module for the required header element, I've had to use on `document.querySelector()` instead of `this.$module.querySelector()` for example.
- The header element should always be present as it is required to render the super navigation header correctly and is included in the component view template, if this is not present the menu does not render correctly.
- Tests have been added to check the new feature works as expected and although the header element should always be present, safety checks have been added to the code before trying to modify any attributes of the header element.

## Testing
I've added a "TEMP" commit to allow for deploying and testing the changes following the guidance in - https://github.com/alphagov/govuk_publishing_components/blob/main/docs/test-a-branch-of-the-gem.md

### Testing on Integration

- Deploy the "increase-header-spacing-if-menu-open" branch of the frontend app to Integration - https://github.com/alphagov/frontend/pull/3873
- Deploy the "test-publishing-components-gem-branch" branch of static to Integration - https://github.com/alphagov/static/tree/test-publishing-components-gem-branch

### Testing locally
You can also test locally following the guidance on this page - https://docs.publishing.service.gov.uk/manual/local-frontend-development.html#using-govuk-docker

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
@@ -23,26 +23,32 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   // Wrapper functions to contain all of the mechanisms needed for hiding and
   // toggling the menus.
-  var hide = function ($button, $menu) {
+  var hide = function ($button, $menu, $superNavigationHeader) {
     $button.setAttribute('aria-expanded', false)
     $button.classList.remove('gem-c-layout-super-navigation-header__open-button')
     $menu.setAttribute('hidden', 'hidden')
     setLabel($button, 'show')
+    if ($superNavigationHeader) {
+      $superNavigationHeader.classList.remove('gem-c-layout-super-navigation-header--menu-open')
+    }
   }
-  var show = function ($button, $menu) {
+  var show = function ($button, $menu, $superNavigationHeader) {
     $button.setAttribute('aria-expanded', true)
     $button.classList.add('gem-c-layout-super-navigation-header__open-button')
     $menu.removeAttribute('hidden')
     setLabel($button, 'hide')
+    if ($superNavigationHeader) {
+      $superNavigationHeader.classList.add('gem-c-layout-super-navigation-header--menu-open')
+    }
   }
 
-  var toggle = function ($button, $menu) {
+  var toggle = function ($button, $menu, $superNavigationHeader) {
     var isOpen = $button.getAttribute('aria-expanded') === 'true'
     var trackingLabel = $button.getAttribute('data-tracking-key')
     if (isOpen) {
-      hide($button, $menu)
+      hide($button, $menu, $superNavigationHeader)
     } else {
-      show($button, $menu)
+      show($button, $menu, $superNavigationHeader)
     }
 
     // Fire analytics if analytics are available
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -153,6 +160,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module.querySelector('.gem-c-layout-super-navigation-header__search-item-link')
       .setAttribute('hidden', 'hidden')
 
+    this.$superNavigationHeader = document.querySelector('.gem-c-layout-super-navigation-header')
```
hannalaakso, line 163: I wonder if it would make sense to move this variable assignment to line 100 of the file where we define the other top level variables (I think it's currently being assigned `null` there)? The variable could maybe also be called `$moduleParent` instead of `$superNavigationHeader` to be slightly clearer about what it is we're selecting and the relationship between the module and the super navigation header.

Then if we make `.gem-c-layout-super-navigation-header` a module level variable in this way, we'll be able to call it with `this.$superNavigationHeader` (or `this.$moduleParent`) inside the methods instead of passing it in the method parameters. We're not hiding this element in the DOM, only adding a class to it so it could be a bit confusing that it's being passed to `hide()` and `show()`. (We could even make toggling this class on the header into its own method that is called when show and hide are called, that could be the clearest thing to do in some ways.)

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -98,6 +98,21 @@ $after-button-padding-left: govuk-spacing(4);
   border-top: 1px solid $govuk-brand-colour;
 }
 
+
+.gem-c-layout-super-navigation-header--blue-background.gem-c-layout-super-navigation-header--menu-open {
+  // Add a blue section to the bottom of the open menu
```
hannalaakso, line 103: A small thing but we could say why we're doing this in the comment as that might not be clear.
```
// Add a blue section to the bottom of the open menu.
// This increases the space above the blue banner on the homepage when the menu is open for improved readability.
```

hannalaakso (commented):
This works correctly 👍  I've made some suggestions below about making it a bit clearer what it's doing. But I know this component is generally quite complicated and could do with a rewrite too so it's not the easiest one to work with...

As you mentioned in the PR description, it's unfortunate that `$module` is not the top level `.gem-c-layout-super-navigation-header` as it would make this more straightforward, although the way it is has made sense up until this point as the current `$module` is where the JS gets actioned. We could rewrite the JS so that `gem-c-layout-super-navigation-header` becomes `$module` but the complexity of this component could make it quite a big job and I'm not sure the benefits gained warrant doing that right now.

PR #3746, opened 2023-12-04
Standardise search term formatting across GA4 trackers

AshGDS:
## What
- Search terms were being cleaned up in different ways by different trackers (These were the pageview tracker, the form tracker, and the ecommerce tracker)
- I have fixed this by creating a function which all trackers that utilise the search term now use. This function will cleanup/format the search term instead of the individual trackers.
- I then added extra tests where relevant to the trackers to ensure that they handled search terms in the same way.

## Why

- The PAs require the search term to be consistent across trackers so they can cross reference data/follow user journeys.
- Trello card: ['search_term' page_view parameter is coming through with capitals - should be lowercased](https://trello.com/c/Xuzx1cYM/738-searchterm-pageview-parameter-is-coming-through-with-capitals-should-be-lowercased)
- Trello card: [Ensure clean-up regex is the same across all search_term fields](https://trello.com/c/ntC48CCa/739-ensure-clean-up-regex-is-the-same-across-all-searchterm-fields)

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
@@ -103,35 +103,50 @@ describe('Google Analytics 4 ecommerce tracking', function () {
     })
 
     it('should get the search query', function () {
-      onPageLoadExpected.search_results.term = 'coronavirus'
+      var expected = 'coronavirus'
       searchResultsParentEl.setAttribute('data-ga4-search-query', 'coronavirus')
       GOVUK.analyticsGa4.Ga4EcommerceTracker.init()
 
-      expect(window.dataLayer[1].search_results.term).toBe(onPageLoadExpected.search_results.term)
+      expect(window.dataLayer[1].search_results.term).toBe(expected)
     })
 
     it('should remove PII from search query', function () {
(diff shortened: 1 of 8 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
```diff
@@ -118,6 +118,22 @@ describe('Google Analytics 4 ecommerce tracking', function () {
       expect(window.dataLayer[1].search_results.term).toBe(onPageLoadExpected.search_results.term)
     })
 
+    it('should replace plusses with spaces, and remove extra lines/spaces from search query', function () {
+      onPageLoadExpected.search_results.term = 'i have a lot of spaces'
```
andysellick, line 122: Do you need to set this variable as part of the `onPageLoadExpected` object, or could it just be a `var` by itself?

andysellick (approved):
Nice. One small comment about a test, but otherwise looks good.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
AshGDS: @andysellick Good spot, will change this in multiple tests :+1:

PR #3748, opened 2023-12-05
Prevent crash when link tracker runs on a non-link element

AshGDS:
## What

- If the link tracker exists without `data-ga4-track-links-only`, it will assume every child element inside of it is an `<a>` tag with a href.
- However, sometimes we do not use `data-ga4-track-links-only`, so certain elements (e.g. SVG icons or paragraph text) are being run through our `trackClick` function in the link tracker. Since these elements do not have hrefs, the link tracking crashes when these elements are clicked.
- Examples include the SVG icons on the action links on the homepage, and the "Open Government License" paragraph text in the footer. Clicking either of these produces an error in the console.
- To fix this, we put a guard in that will `return` our `trackClick` function if there is no `href` on the link, or no `url` in the link tracker JSON.

## Why

- Prevents a JS error in the console

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -85,7 +85,13 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       if (!data.text && (element.querySelector('img') || element.querySelector('svg') || element.tagName === 'IMG' || element.closest('svg'))) {
         data.text = 'image'
       }
-      var url = data.url || this.findLink(event.target).getAttribute('href')
+      try {
+        var url = data.url || this.findLink(event.target).getAttribute('href')
+      } catch (e) {
+        // no href found, so abort
+        return
+      }
+
       data.url = trackFunctions.removeCrossDomainParams(url)
       data.url = trackFunctions.applyRedactionIfRequired(this.PIIRemover, element, data.url)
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -85,7 +85,12 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       if (!data.text && (element.querySelector('img') || element.querySelector('svg') || element.tagName === 'IMG' || element.closest('svg'))) {
         data.text = 'image'
       }
-      var url = data.url || this.findLink(event.target).getAttribute('href')
+      var url = data.url || this.getLinkHref(event.target)
```
andysellick, line 88: This is a good solution and well done for spotting it, but it feels slightly too long. It feels like `getLinkHref` and `findLink` should be combined, since they're inherently related and only called here, but I see the problems involved with that. A shorter (but possibly lazier) solution could be to wrap the original code in a try/catch:

```JavaScript
try {
  var url = data.url || this.findLink(event.target).getAttribute('href')
} catch (e) {
  // no href found, so abort
  return
}
```

AshGDS: Thanks @andysellick, I've updated the code to reflect this - should be ready for another review :+1:

andysellick (approved):

PR #3750, opened 2023-12-05
Data attributes for option select button

andysellick:
## What
- modify the option select component to allow it to accept data attributes to be added to the expand/collapse button
- the attributes are rendered on the parent element, then read in by JavaScript to be included in the JavaScript generated button, which replaces the title element of the component
- to be used for GA4 tracking purposes in finder-frontend, in place of an existing solution
- will be a breaking change, as removes the event listener that was previously being used for this
- credit to @floehopper

## Why
We're currently using a system of event and event listeners between `finder-frontend` and this component in order to track this button's expand/collapse action. This approach allows us to remove some JavaScript and move most of the heavy lifting to the backend, which should reduce load on the frontend.

This PR in `finder-frontend` is dependent on this change, and will need to be merged together: https://github.com/alphagov/finder-frontend/pull/3224

## Visual Changes
None.

Trello card: https://trello.com/c/OdYXlYqQ/673-finder-frontend-code-optimisations

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* [BREAKING] Data attributes for option select button ([PR #3750](https://github.com/alphagov/govuk_publishing_components/pull/3750))
 * Remove doNotTrack from GTM component ([PR #3760](https://github.com/alphagov/govuk_publishing_components/pull/3760))
 * Fix rendering of the image card YouTube variation #3757 ([PR #3757](https://github.com/alphagov/govuk_publishing_components/pull/3757))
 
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
Looks good :+1: 

Do you think it's worth having a `try/catch` safeguard on the JavaScript in case the value doesn't come through as a JSON? E.g. passing `button_data_attributes: "hello"` crashes the JS. This is probably an incredibly rare scenario so don't worry about it if you think it's over engineering the code.

andysellick:
@AshGDS good shout, we do that everywhere else so we should do it here. Have pushed another commit if you've got time to take a look, will squash before merging.

AshGDS (approved):
Nice, thanks @andysellick :+1:

AshGDS (approved):

PR #3753, opened 2023-12-06
Fix error in video tracker

andysellick:
## What
- where the analytics code is not run (such as the component guide) the image card video variant was experiencing a JS error
- this shouldn't happen in the wild but it shouldn't happen in our dev environments either
- added a simple check to ensure the code fails silently if analytics is not initialised

## Why
Error was `Uncaught TypeError: Cannot read properties of undefined (reading 'VideoTracker')`

## Visual Changes
None.


MartinJJones (approved):
Thanks for fixing the JS error 👍

PR #3761, opened 2023-12-08
Add GA4 copy event tracker

andysellick:
## What
- tracks text copy events within the browser
- copied text is determined using the window.getSelection() method, assuming that some text has been highlighted to copy, if not does not send an event
- strips PII, line breaks etc. from text

Have tested this in the major browsers, seems to be okay. Just in case I've wrapped the whole thing in a `try/catch` block to catch anything I've not thought of.

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/s6uLhovR/751-add-tracking-browser-copy-events

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-copy-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-copy-tracker.js
@@ -1,0 +1,55 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analyticsModules || {};
+
+(function (analyticsModules) {
+  'use strict'
+
+  var Ga4CopyTracker = {
+    init: function () {
+      window.addEventListener('copy', this.trackCopy.bind(this))
+    },
+
+    trackCopy: function (event) {
+      try {
+        var text = this.getSelectedText()
+        if (!text) {
+          return // do nothing if no text has been selected
+        }
+        var target = event.target.tagName
+        if (target === 'INPUT' || target === 'TEXTAREA') {
+          return // do nothing if text is being copied from an input
+        }
+
+        var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
+        text = window.GOVUK.analyticsGa4.core.trackFunctions.removeLinesAndExtraSpaces(text)
+        text = PIIRemover.stripPIIWithOverride(text, true, true)
+        var length = text.length
+        text = text.substring(0, 30)
+
+        var data = {
+          event_name: 'copy',
+          type: 'copy',
+          action: 'copy',
+          method: 'browser copy',
+          text: text,
+          length: length
+        }
+        window.GOVUK.analyticsGa4.core.applySchemaAndSendData(data, 'event_data')
+      } catch (e) {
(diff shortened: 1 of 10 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-copy-tracker.js
```diff
@@ -0,0 +1,48 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
+window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analyticsModules || {};
+
+(function (analyticsModules) {
+  'use strict'
+
+  var Ga4CopyTracker = {
+    init: function () {
+      var getSelectedText = this.getSelectedText.bind(this)
```
AshGDS, line 10: Haven't tested it, but could you instead bind `this` to the eventListener itself, and then you wouldn't need this variable?

E.g. 

```JS
window.addEventListener('copy', function () {
      var text = this.getSelectedText()
      }.bind(this))
```

andysellick: Good shout, implemented.

AshGDS (commented):
Looks good to me :+1: Will approve once the schema changes are implemented

andysellick:
I've pushed a commit on the assumption that the attribute will be `copy_length` but will change it later if the decision is to use something different.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-copy-tracker.js
```diff
@@ -0,0 +1,55 @@
...
+        var length = text.length
+        text = text.substring(0, 500)
+
+        var data = {
+          event_name: 'copy',
+          type: 'copy',
+          action: 'copy',
+          method: 'browser copy',
+          text: text,
+          copy_length: length
```
AshGDS, line 36: Non-blocking, is it worth setting `text.length` directly here instead of creating the variable?

AshGDS (approved):
LGTM :+1:

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-copy-tracker.js
andysellick: Excellent question. If we set it here, it's already been trimmed to a maximum of 500 characters (the limit for GA4 attribute data) so we have to store it as a variable earlier. Or store the original text as a variable. Either way, a variable is involved - and we'd like to know the original size of the copied text (which could be thousands of characters, if it's the whole page).

AshGDS: Ah, that makes sense - thanks!

AshGDS: @andysellick Interesting, none of this crossed my mind! I think what you've done makes sense :+1:

andysellick:
@AshGDS have updated this PR...

- changing the `video_duration` attribute to `length`
- updating the video tracker and copy tracker code accordingly
- changing the copy tracker to only record 30 chars

Feel free to re-review the whole thing, although it shouldn't be a huge change.

AshGDS (approved):
LGTM 👍

PR #3767, opened 2023-12-12
Adjust option select JavaScript

andysellick:
## What / why
Fixes a possible problem with the button data attributes functionality, where attribute names with more than one underscore might not have been properly converted. Noticed in https://github.com/alphagov/finder-frontend/pull/3224 relating to the expander component, which uses the same code for the same functionality.

## Visual Changes
None.


AshGDS (approved):
LGTM 👍

PR #3769, opened 2023-12-13
Remove hardcoded text from GA4 'Feedback component' tracking, and add disable_ga4 option

AshGDS:
## What

- Removes hardcoded `text` and `section` values from the GA4 Feedback tracking, using the translation variables instead
- Adds the `disable_ga4` variable to the component so that tracking can be disabled

## Why

https://trello.com/c/DbAXO3lU/556-stop-text-that-can-be-translated-using-hard-coded-text-values

## Visual Changes

Adds another example to the component page

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove hardcoded text from GA4 'Feedback component' tracking, and add disable_ga4 option ([PR #3769](https://github.com/alphagov/govuk_publishing_components/pull/3769))
 * Add GA4 copy event tracker ([PR #3761](https://github.com/alphagov/govuk_publishing_components/pull/3761))
 
 ## 37.0.0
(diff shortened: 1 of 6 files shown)


andysellick (commented):
This all looks good, but do we need an option to disable tracking on the feedback component? It's always in the same place in every page, so it shouldn't ever get tangled in any other tracking (apart from the general link tracking, but I don't think they'd conflict). I guess it might be useful if another application was using the component but didn't want to track it, but I can't think of a specific scenario where that would happen. Have you got an example use case for this?

AshGDS:
@andysellick No worries, I can remove it 👍  I was following this guidance: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/analytics.md#code-in-components

andysellick:
@AshGDS I'm quite happy to keep it in (aware you did a fair bit of work on this) if there's a good reason to do so.

AshGDS:
@andysellick Cool, let's leave it in then, I think the code might be a bit more readable/extensible with the change added. 👍

andysellick (approved):
Approved, but looks like you're getting the `logs` error in the tests 😢 we might have to figure out a fix for that..

PR #3775, opened 2023-12-14
Fix failing test

AshGDS:
## What

- Change how a test grabs a console log message
- I think the array of console logs now only has one entry in it with all the console logs in, instead of one entry per console log. Therefore we only need to check the first entry for the inclusion of the 'Accessibility issues' message

## Why

- Fixes a failing test


AshGDS:
Closing as it's caused by Capybara instead of us

PR #3781, opened 2023-12-15
Remove duplicate GA4 tracking

AshGDS:
## What

- Having the tracking on this link triggers a link click and a select_content as it expands an accordion on the same page. So we remove the link click tracking so only one event fires for the interaction. Example of the issue: https://www.gov.uk/government/consultations/primate-licensing-scheme on the 'See all updates' link.

## Why

- https://trello.com/c/jf1V3D02/770-placeholder-duplicate-content-history-events-firing-on-some-pages

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove duplicate GA4 tracking ([PR #3781](https://github.com/alphagov/govuk_publishing_components/pull/3781))
 * Delete unused builds folder ([PR #3778](https://github.com/alphagov/govuk_publishing_components/pull/3778))
 
 ## 37.1.0
--- a/app/views/govuk_publishing_components/components/_metadata.html.erb
+++ b/app/views/govuk_publishing_components/components/_metadata.html.erb
@@ -65,11 +65,7 @@
                     class="gem-c-metadata__definition-link govuk-!-display-none-print js-see-all-updates-link"
                     data-track-category="content-history"
                     data-track-action="see-all-updates-link-clicked"
-                    data-track-label="history"
-                    <% unless disable_ga4 %>
-                      data-module="ga4-link-tracker"
-                      data-ga4-link="<%= ga4_object %>"
-                    <% end%>>
+                    data-track-label="history">
             <%= t("components.metadata.see_all_updates") %>
           </a>
         <% end %>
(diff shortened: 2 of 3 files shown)


andysellick (commented):
Slightly hesitant about this change because it's permanently removing tracking from part of this component, which works in this situation but I don't know what other situations it might impact.

Looking at the page in question, there isn't any other tracking that fires on that component. Again I don't know what variations might occur for these pages, but for this specific page you could achieve the same effect by using `disable_ga4` instead.

(side point - we seem to be only applying tracking to parts of the metadata component, for example in this situation there's no tracking on the `From` link, but links can appear there and elsewhere, [see this example](https://components.publishing.service.gov.uk/component-guide/metadata/extensive_specialist_document))

AshGDS:
@andysellick I think this line of code can be removed because I have a list of all the templates that use this `See all updates` link. So I don't think removing the code would affect anything we weren't aware of, since this tracking was added specifically for these templates and now it's being removed specifically for these same templates. Here is the list of see all updates links in `government-frontend`: https://gist.github.com/AshGDS/0f3bd5aa6c04cefed2839f4d9b6b7337#see-all-updates-link, all of them except the HMRC manual were making use of the code that I'm removing and they all had this bug that will be fixed by removal of this.

But interestingly, it looks like the move to using `disable_ga4` instead of `ga4_tracking: true` is the reason for this PR was raised in the first place. As the previous work to remove the tracking involved removing `ga4_tracking: true`, but of course that wasn't replaced with a `disable_ga4: true` flag as we weren't aware of that coming into use at the time. https://github.com/alphagov/government-frontend/pull/2935 

I can add `disable_ga4` in `government-frontend` instead, but I guess that would mean we potentially leave in tracking code in this component that's never used, and as far as i'm aware this line of tracking was only used in those templates that have the duplicated event issue.

I believe only tracking `See all updates` is intentional, I can't remember specifically why (I vaguely recall something about being told that Pageview events can be used to track the user journey for those internal links) but I guess it can be brought to the team if tracking is needed on those links.

andysellick (approved):
Thanks for the supporting information @AshGDS, that's all very thorough. I think you're right, if this tracking was added specifically for this situation then removing it rather than disabling it seems like the right approach.

Probably need a changelog entry as well.

AshGDS:
Thanks @andysellick 👍

PR #3786, opened 2023-12-19
Details component GA4 tracking

andysellick:
## What
- adds GA4 tracking (by default) to the details component
- includes disable_ga4 option to disable tracking
- uses an OpenStruct to record the index of details components on a page for the `index_section` GA4 attribute
- total number of details components is also required for `index_section_count` attribute, but cannot determine automatically
- instead can pass a hash of GA4 attributes, which are then merged with the defaults
- passing can allow existing attributes to be overridden
- tracking on links not included owing to varying requirements, better to wrap the component in the link tracker in applications, where tracking is required

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/xhASj1rp/765-add-tracking-to-details-component

### DIFF
--- a/app/views/govuk_publishing_components/components/_details.html.erb
+++ b/app/views/govuk_publishing_components/components/_details.html.erb
@@ -2,23 +2,39 @@
   add_gem_component_stylesheet("details")
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
-
   open ||= nil
+  disable_ga4 ||= false
+  @ga4 ||= OpenStruct.new(index_section: 0) unless disable_ga4
+  @ga4[:index_section] += 1 unless disable_ga4
+  ga4_attributes ||= {}
+
   margin_bottom ||= 3
   css_classes = %w(gem-c-details govuk-details)
   css_classes << shared_helper.get_margin_bottom
 
   details_data_attributes = {}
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_details.html.erb
```diff
@@ -2,23 +2,39 @@
...
+    ga4_event.merge!(ga4_attributes)
+    data_attributes[:ga4_event] = ga4_event
+  end
 
   summary_aria_attributes ||= {}
 %>
 <%= tag.details class: css_classes, data: details_data_attributes, open: open do %>
   <%= tag.summary class: "govuk-details__summary", data: data_attributes, aria: summary_aria_attributes do %>
-    <span class="govuk-details__summary-text">
+    <span class="govuk-details__summary-text" data-ga4-expandable>
```
AshGDS, line 37: do you think this attribute should be removed when ga4 is disabled?

AshGDS (approved):
LGTM 👍

app/views/govuk_publishing_components/components/_details.html.erb
andysellick: Agreed. Have amended and altered some tests, thanks for highlighting.

AshGDS (approved):
Thanks @andysellick 👍

PR #3788, opened 2023-12-19
[DO NOT MERGE] Refactor image card tracking approach #1

andysellick:
## What
Rework how GA4 tracking is done on the image card component. This is an experimental approach (see also https://github.com/alphagov/govuk_publishing_components/pull/3789)

Specifically:

- add the GA4 link tracker by default to the component
- document how this works - that `data-ga4-link` data still needs to be passed to the component for tracking to work
- include a `disable_ga4` option (as per our other components with tracking)

Pros:

- decouples the GA4 tracking from the UA tracking (ahead of removing the UA tracking)
- more clearly defines and documents how GA4 tracking should work

Cons:

- initialises the JS module for GA4 link tracking on all instances of the component, whether they are configured for tracking or not

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/IymJPytQ/762-sort-out-image-card-component-tracking

### DIFF
--- a/app/views/govuk_publishing_components/components/_image_card.html.erb
+++ b/app/views/govuk_publishing_components/components/_image_card.html.erb
@@ -3,6 +3,7 @@
 
   sizes ||= false
   brand ||= false
+  disable_ga4 ||= false
 
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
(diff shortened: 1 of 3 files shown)


andysellick:
Closing in favour of https://github.com/alphagov/govuk_publishing_components/pull/3789

PR #3789, opened 2023-12-19
Refactor image card tracking

andysellick:
## What
Rework how GA4 tracking is done on the image card component (prototyped a different approach in https://github.com/alphagov/govuk_publishing_components/pull/3788)

Specifically:

- don't include any tracking at all on the component 😱 
- refactor the component to use the component wrapper helper...
- ...which allows any data attributes to be passed safely to the component without overwriting existing ones
- ...meaning that the link tracker and required data attributes can be passed to the component where required

Pros:

- another component using the component wrapper helper
- no additional code for tracking on the component, all handled by passing options

Cons:

- arguably a breaking change, as the current tracking for the homepage uses of the image card will stop working

This PR removes the test for passing a `lang` attribute to the component, as this functionality is now provided by the component wrapper helper.

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/IymJPytQ/762-sort-out-image-card-component-tracking

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Refactor image card tracking ([PR #3789](https://github.com/alphagov/govuk_publishing_components/pull/3789))
 * Details component GA4 tracking ([PR #3786](https://github.com/alphagov/govuk_publishing_components/pull/3786))
 
 ## 37.1.1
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
Looks good :+1: Just to double check, this would mean that there's no tracking by default on this component anymore, and therefore any new image cards added to the site would need a dev to ensure tracking is added. Is this OK for the purposes of the image card?

andysellick:
@AshGDS that's correct. The problem I faced was that tracking can't be isolated on this component - it needs to be passed data for things like `index_section` (on the homepage at least). I had to decide between a partial implementation of tracking, with some kind of option to pass remaining options, or this. This seemed simpler and cleaner, and affords more flexibility.

I explored the alternative in https://github.com/alphagov/govuk_publishing_components/pull/3788 and it just felt messy to me, but if there's another approach I'd be happy to discuss.

AshGDS:
@andysellick No worries - I'm happy with this approach :+1: Thanks!

PR #3790, opened 2023-12-20
Add timestamp value to GA4 dataLayer pushes

AshGDS:
## What

- Adds a `timestamp` value sent with each dataLayer push. `timestamp` is the user's unix timestamp.
- Most of the tests I used `spyOn` to fake the timestamp as `123456`, probably necessary as it's impossible to predict what the timestamp will be before vs after some tracking code has run. However I've added one test in `ga4-core` that tests the actual timestamp and it works. I think because the test is so simple, modern computers are fast enough that it can grab the timestamp for the expected and actual value, and it's so fast that the timestamp wouldn't have incremented yet.
- Will add a CHANGELOG once approved as I suspect this won't be merged until 2024.

## Why

- So that the PAs can easily sort the events to see a users journey 

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add timestamp value to GA4 dataLayer pushes ([PR #3790](https://github.com/alphagov/govuk_publishing_components/pull/3790))
 * Allow GA4 tracking on dev docs ([PR #3802](https://github.com/alphagov/govuk_publishing_components/pull/3802))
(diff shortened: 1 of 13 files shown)


andysellick (approved):
Looks good to me 👍 🎄

andysellick:
Also will need an update for most of the events and the schema in the implementation record.

andysellick:
@AshGDS probably needs a CHANGELOG entry?

AshGDS:
@andysellick Thanks, have added one. Will merge shortly

PR #3800, opened 2024-01-03
Ensure all GA4 data values are strings

andysellick:
## What
- change the schema code when merging properties to automatically convert all data to strings
- currently passing numbers for various data attributes (e.g. index_section) and all should be consistent types
- updating various tests accordingly

Have looked at various options for converting integers to strings in JS. `2 + ''` seems to be the most widely supported, fastest and least likely to break of the available options, but happy to discuss.

## Why
Part of the GA4 migration.

## Visual Changes
None.

Trello card: https://trello.com/c/nPhYtY0j/761-address-inconsistency-in-index-property-data-types

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Ensure all GA4 data values are strings ([PR #3800](https://github.com/alphagov/govuk_publishing_components/pull/3800))
 * Refactor image card tracking ([PR #3789](https://github.com/alphagov/govuk_publishing_components/pull/3789))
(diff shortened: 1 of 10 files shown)


AshGDS (approved):
Looks good, though I noticed that some eCommerce values in search are still integers - maybe they are nested too far down for this code to convert them?

andysellick:
@AshGDS thanks, well noticed. Yeah I guess so? Not sure what we can do about that from this side of the code. As this is a not particularly important bug fix I'll go with this for now as it at least should solve most of it.

PR #3802, opened 2024-01-04
Allow GA4 tracking on dev docs

andysellick:
## What
Changes necessary to support allowing the [dev docs](https://docs.publishing.service.gov.uk/) to use the GA4 tracking code in the gem. See this [related PR](https://github.com/alphagov/govuk-developer-docs/pull/4376). Specifically:

- adjust the GA4 initialisation code to include variables for the tracking on the dev docs
- include the code from `frontend` for the cookie settings page, as it will now be used in more than just `frontend` (another PR will be raised later to remove this from `frontend`)
- refactor the code in load analytics

## Why
We want to extend GA4 tracking to the whole of the dev docs, not just the implementation record section.

## Visual Changes
None.

Trello card: https://trello.com/c/kA66mxak/755-extend-analytics-to-all-of-dev-docs

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Allow GA4 tracking on dev docs ([PR #3802](https://github.com/alphagov/govuk_publishing_components/pull/3802))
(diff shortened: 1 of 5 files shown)


AshGDS (approved):
LGTM 👍

PR #3807, opened 2024-01-08
Fix Sass deprecation warnings

jon-kirwan:
## What

Fix Sass deprecation warnings related to using `/` for division outside of `calc()`.

## Why

Using `/` for division outside of `calc()` is deprecated and will be removed in Dart Sass 2.0.0.

## Visual Changes

None.


andysellick (approved):
Looks good to me, good work @jon-kirwan 👍

Issue #3369, opened 2023-04-28
Applications do not follow component conventions


andysellick:
I'm doing a bit of a spike into making the component auditing report on components within applications, prompted by recent conversations. Turns out that our naming and file location conventions for components within the gem aren't always being followed by applications. Here's some (very) initial investigations.

For component templates:

| Application | Name | Location |
| ----------- | ------ | --------- |
| Gem |  `_component_name.html.erb` | `app/views/govuk_publishing_components/components/` |
| Collections | `_component-name.html.erb` | `app/views/components/` |
| Finder-frontend | `_component-name.html.erb` | `app/views/components/` |

For component test files:

| Application | Name | Location |
| ----------- | ------ | --------- |
| Gem |  `component_name_spec.rb` | `spec/components/` |
| Collections | `component_name.html.erb_spec.rb` | `spec/views/components/` |
| Finder-frontend | `component_name_spec.rb` | `spec/components/` |

This inconsistency is going to make auditing components in applications difficult, as we'd have to include specific details of all of these things on a per-application basis within the auditing. I'll have a look at whether applications could be restructured slightly instead.

andysellick:
**UPDATE** `government-frontend` has been archived, so this comment is no longer relevant.

-------

There's an interesting inconsistency in `government-frontend` that's worth documenting. The component test files are in `test/components` so I looked into moving them into `spec/components` to match the gem.

We use two different systems for tests - RSpec and Minitest, and which directory the test should go in is related to the system. `government-frontend` uses Minitest, so the tests are in `test/`, and if we wanted to move them we'd have to rewrite them in RSpec first. That's probably not worth the effort for the sake of component auditing.

MartinJJones:
I was having a quick look at this issue and how we can try to avoid using incorrect file names in the future.

When I followed the documentation to [Create a new component in the application](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/generate-a-new-component.md#option-1-create-a-component-in-the-application), I ran the command below to generate a new component:

```Ruby
rails generate govuk_publishing_components:component new_component_name
```

However, the file component file name for the component view template included hyphens instead of underscores.

```Ruby
government-frontend % rails generate govuk_publishing_components:component new_component_name
      create  app/views/components/_new-component-name.html.erb
      create  app/views/components/docs/new-component-name.yml
      create  app/assets/stylesheets/components/_new-component-name.scss
```

[...]

PR #3813, opened 2024-01-09
Test component filenames

andysellick:
## What

- adds a test to check that component template files are named correctly
- already tests for YML and SCSS files, but not for template
- in order to get an unbiased component name, the initial variable has to be modified, this is to catch the two main problems with component names: extension named .erb instead of .html.erb, and `-` used instead of `_`

Relates to issue https://github.com/alphagov/govuk_publishing_components/issues/3369

## Why
Hoping to port this test to other applications to reduce file naming inconsistencies and improve auditing tool impact.

## Visual Changes
None.


AshGDS (approved):
Nice :+1:

PR #3814, opened 2024-01-10
Change social media links advice text

unoduetre:
## What

On pages such as the social media feeds “Sharing will open the page in the new tab” is confusing and not accurate. Suggestion that the copy should be something like “The following links open in a new tab" to be clearer.

## Why

To help users understand what happens on their browse when sharing and it shouldn’t be confusing.

[Trello ticket](https://trello.com/c/UeGITj5l/2276-update-social-media-tab-copy-in-the-gem-s)

## Visual Changes

The screenshots below reflect the old proposed wording. After some discussion, the new wording is "The following links open in a new tab". I have not remade the screenshots.

### Before

before

### After

after

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Remove hyphens and downcase the GA4 publishing government value ([PR #3808](https://github.com/alphagov/govuk_publishing_components/pull/3808))
 * Fix Sass negative mixin / value issue ([PR #3816](https://github.com/alphagov/govuk_publishing_components/pull/3816))
+* Change social media links advice text ([PR #3814](https://github.com/alphagov/govuk_publishing_components/pull/3814))
 
 ## 37.2.1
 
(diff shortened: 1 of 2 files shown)


MartinJJones:
The change looks good to me overall.

There are a couple of other things that we may want to consider further.

- There is also a Welsh translation of the text here - https://github.com/alphagov/govuk_publishing_components/blob/main/config/locales/cy.yml#L185. So with this change we would be out of sync with the Welsh translation.
- From looking at the visual changes in Percy, the change in text may not be ideal when an "Email" link is used, for example:

Screenshot 2024-01-10 at 17 09 02

We could consider using the text "The following links open in a new tab" instead, this is taken from the [Design System guidance page](https://design-system.service.gov.uk/styles/links/).

unoduetre:
Wording changed (and approved by @hannako) as per @MartinJJones's suggestion.

unoduetre:
> Changes look good to me 👍
> 
> non-blocking, but we usually try to keep the name of the PR and the change log entry the same, could either update the PR name or Changelog entry.

Fixed.

PR #3820, opened 2024-01-15
Allow attachment to pass tracking to details

andysellick:
## What
- extend the attachment component (which can contain the details component) to allow GA4 tracking attributes to be passed through to the details component
- details component does most of its own tracking but requires the `index_section_count` attribute to be passed manually

## Why
Part of the GA4 migration.

## Visual Changes
No changes to the component, but have added an extra entry in the attachment doc, so will trigger a diff.

Trello card: https://trello.com/c/xhASj1rp/765-add-tracking-to-details-component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Allow attachment to pass tracking to details ([PR #3820](https://github.com/alphagov/govuk_publishing_components/pull/3820))
 * Allow GA4 link tracker to track to multiple child classes ([PR #3835](https://github.com/alphagov/govuk_publishing_components/pull/3835))
 
 ## 37.2.4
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
Looks good - but two things you might want to be aware of:

- To update the attachments component across our pages, I believe you need to update publishing components in `Whitehall`, and then run a rake task to republish every document that contains attachments
- One render statement handles all the attachments, so you may need to write some logic in Whitehall to trigger the details tracking: https://github.com/alphagov/whitehall/blob/edbcb437179728385f15c8f831958ed82094bbb8/app/helpers/attachments_helper.rb#L69

AshGDS:
Specifically, this rake task will update all the attachments once the latest gem is in publishing components: https://github.com/alphagov/whitehall/blob/edbcb437179728385f15c8f831958ed82094bbb8/lib/tasks/publishing_api.rake#L213

AshGDS:
For more context if you're curious, if you look at the content item JSON of a page with attachments, the HTML for the attachments is grabbed from there and then rendered to the page using `government-frontend`: https://www.gov.uk/api/content/government/publications/new-year-honours-list-2024

So that rake task in Whitehall is what will update that HTML that is in the content item JSON

PR #3827, opened 2024-01-17
Add GOVUK logo to GovernmentOrganization schema.org schemas

AshGDS:
## What

- Adds a reference to the GOVUK logo in our `schema.org` `GovernmentOrganization` schema. 
- This schema is used by organisation index pages such as https://www.gov.uk/government/organisations/cabinet-office to provide metadata about the page to machines i.e. Google Search results.

## Why

- There is an issue on organisation pages in Google search results right now as Google is picking a random image from the organisation pages to feature next to the search results on Google. For example on this search result, it is using the our currently "Featured" image on the cabinet office page as the image next to the search result: https://www.google.com/search?q=cabinet-office
- This is a problem as we do not have control over which image it picks. Google seems to decide what image they think is best. There is a "Related people" section on organisation pages that feature photos of people. Sometimes these photos of people are appearing next to search results for government organisations, which we do not want due to potential interpretations arising from this.
- Initially, I was going to use the meta tag Google provides, which would disable Google from showing image previews on organisation search results entirely. I raised a PR for this in `collections`: https://github.com/alphagov/collections/pull/3504
- However, our Lead PM has suggested we try this first, so that hopefully Google picks the GOVUK logo as the best image, and the GOVUK logo shows up instead of the wrong image or no image at all.

## Testing

Have tested this and it works in `collections`. I have also written it to not crash if `logo_url` happened to no longer exist. `logo_url` is added to `page.local_assigns` for us in `app/views/govuk_publishing_components/components/_machine_readable_metadata.html.erb`.

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add GOVUK logo to GovernmentOrganization schema.org schemas ([PR #3827](https://github.com/alphagov/govuk_publishing_components/pull/3827))
+
 ## 37.2.2
 
 * Remove hyphens and downcase the GA4 publishing government value ([PR #3808](https://github.com/alphagov/govuk_publishing_components/pull/3808))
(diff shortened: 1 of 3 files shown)


sihugh (approved):
👍  Thanks @AshGDS, it's worth a try! (Matches what's in https://schema.org/GovernmentOrganization too)

PR #3829, opened 2024-01-19
[WIP] Add single consent API

andysellick:
## What
Adds code for the [single consent cookie API](https://github.com/alphagov/consent-api) to allow cookie preferences to be shared across GOV.UK.

Potentially quite a few changes here, see individual commits for details. Basic changes when you visit for the first time:

- cookie banner appears and operates as normal
- single consent API is called to check for pre-existing consent preferences (and none found)
- when consent/rejection given, set our cookie as usual, but also send preferences to the consent API

Gets a bit more complicated when consent has been given on another site but not ours, via something like `https://www.gov.uk?gov_singleconsent_uid=<id>`:

- cookie banner appears (for the moment)
- single consent API is called to check for pre-existing consent preferences (using the ID from the URL, finds and sends a response)
- response hits our code, as consent is given we need to now hide the cookie banner, so triggers the `cookie-consent` event
- cookie banner is listening for this event and hides itself
- analytics code is also listening for this event, and initialises
- various components and other bits of code are also listening, and also initialise

Still to do:

- cookie consent page (and more nuanced handling of consent beyond accept/reject all)
- need to test performance impact (particularly on CLS) for scenario described where consent already given on another site
- all code listening for the `cookie-consent` event must also remove the event listener once it fires (consent api can be called more than once, don't want all modules initialising twice on consent)

## Why
Improve user journeys by removing repetitive cookie banners.

## Visual Changes
None.


andysellick:
@AshGDS I think this code is now ready for review. I'd like to add some more tests for when the API isn't available, or times out, but that's dependent on an upstream change so I can't implement right now, but there should be enough here to start figuring out if this is the right approach or not.

Just remembered I also wanted to write some documentation! I'll add that today.

AshGDS (commented):
Nice work, thanks for doing this 👍 

I can't see any glaring issues of the code. I can't get the `uid` to persist across domains for some reason but will look next week, and will look more thoroughly at the code next week.

Is it worth writing a test to check that when a user lands on `www.gov.uk` with `gov_singleconsent_uid` on the URL, it correctly gets removed from the URL and the correct cookie is set? I think it's important we ensure that `gov_singleconsent_uid` is removed from the URL - if this part of the code fails, users could share `www.gov.uk` links that would cause others consent to cookies without their consent. I appreciate if you think this is the consent APIs responsibility to test though.

AshGDS (changes requested):
Code looks OK to me.

I tested this on IE11 and it didn't work properly though. When you accept cookies on our banner you get:

`Object doesn't support property or method 'find'`

However the cookies seem to set correctly:

```JS
"cookies_policy={"essential":true,"usage":true,"campaigns":true,"settings":true}; cookies_preferences_set=true; global_bar_seen={"count":0,"version":8}; _ga=GA1.4.1414675227.1706530172; _gid=GA1.4.1487325447.1706530172; lux_uid=170653017205074608; gov_singleconsent_uid=v2OuKyZeRc-fSsQvywRXeg; _ga=GA1.3.1414675227.1706530172; _gid=GA1.3.1487325447.1706530172"
```

Also, the cookie settings page radio buttons no longer render on IE11. There are these two errors:

- On page load of `/help/cookies`: `Error starting CookieSettings component JS: Object doesn't support property or method 'find' http://frontend.dev.gov.uk/help/cookies`

- When updating cookie preferences: `Single consent setPreferences error: Unable to get property 'setConsents' of undefined or null reference http://frontend.dev.gov.uk/help/cookies`

However the changes seem to take place in the cookie. E.g. I disabled all cookies and `console.log(document.cookie)` gives me: `cookies_policy={"essential":true,"usage":false,"campaigns":false,"settings":false};`

andysellick:
Good testing, thanks @AshGDS. Unfortunately I think the `.find` that IE11 is having a problem with is in the [consent code](https://github.com/alphagov/consent-api/blob/fd8a90305a97c052392783b96bb7e7427d4073fd/client/src/utils.ts#L138). I'll raise.

PR #3830, opened 2024-01-22
Hide our cookie banner when JS is disabled

AshGDS:
## What

- Hides our cookie banner when JS is disabled
- Tested in `static` with Chrome/Firefox/Safari on Mac, and IE11 on Windows
- Has approval from our privacy people. Data protection officer: _"agrees it sounds fine, so long as we have the cookie notice linked at the bottom of the page (no message or banner necessary)"_ 
## Why

- We've had a couple of users who use the page without JavaScript mention to us that the banner is displayed. I assume the main issue being it takes up a lot of screen space for them.

### Static troubleshooting
If you want to test it on `static` but have issues try this. I keep having issues with my `static` updating its assets, so to get `static` to work I had to:

- Change _public_layout.html.erb in publishing components to use `<link rel="stylesheet" href="/assets/static/application.css" media="all">` as the application stylesheet
- Change load-analytics.js to use `<script src="/assets/static/govuk_publishing_components/load-analytics.js"></script>`
- Change `application.js` to use `<script src="/assets/static/application.js" type="text/javascript"></script>`
- In `static`, run `bundle exec rake dartsass:watch`
- Hit save on `_cookie_banner.scss` so it compiles the CSS

### Before (JS disabled)

image

### After (JS disabled)

image

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
@@ -26,13 +26,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
     this.$acceptCookiesButton = this.$module.querySelector('button[data-accept-cookies]')
     if (this.$acceptCookiesButton) {
-      this.$acceptCookiesButton.style.display = 'block'
       this.$acceptCookiesButton.addEventListener('click', this.$module.setCookieConsent)
     }
 
     this.$rejectCookiesButton = this.$module.querySelector('button[data-reject-cookies]')
     if (this.$rejectCookiesButton) {
-      this.$rejectCookiesButton.style.display = 'block'
       this.$rejectCookiesButton.addEventListener('click', this.$module.rejectCookieConsent)
     }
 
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
```diff
@@ -45,6 +45,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       var shouldHaveCookieMessage = (this.$module && window.GOVUK.cookie('cookies_preferences_set') !== 'true')
 
       if (shouldHaveCookieMessage) {
+        this.$module.removeAttribute('hidden')
```
andysellick, line 48: If we're hiding the cookie banner unless JS is enabled, I think there's some additional code that we can change or possibly remove. In the [setupCookieMessage function](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js#L19) there's some code that makes the two buttons appear, because they're hidden by default so they're not shown when JS is disabled.

There's also some stuff in the [showCookieMessage function](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js#L57) where its setting the display to 'none', which also might not be needed anymore? I think it's making the cookie banner hidden on the cookies settings page if you've not made a choice yet, but if it's hidden by default now that might not be needed.

andysellick (commented):
Looks good, some additional things to maybe look at. Also are we allowed to do this?

AshGDS:
Thanks @andysellick - just waiting for someone in our data privacy team to greenlight this change before I work on it again.

app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js
AshGDS: Thanks, should be fixed now 👍

andysellick (approved):

PR #3839, opened 2024-01-25
Move JS spec file for autoscrolltracker

andysellick:
## What / why
- this test file was in the wrong place for some reason, it's part of the UA analytics code, so should be in the analytics test directory

## Visual Changes
None.


AshGDS (approved):
LGTM 👍 

Looks like it was just missed by mistake when the tracking files were moved in the PR that added this tracker

PR #3840, opened 2024-01-29
Add tracking to back link

andysellick:
## What
Adds tracking by default to the back link component.

## Why
Part of the GA4 migration.

## Visual Changes
None (apart from changes to the component guide)

Trello card: https://trello.com/c/Z5FYBU5o/782-back-links

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add tracking to back link ([PR #3840](https://github.com/alphagov/govuk_publishing_components/pull/3840))
 * Allow attachment to pass tracking to details ([PR #3820](https://github.com/alphagov/govuk_publishing_components/pull/3820))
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
LGTM 👍 I guess the component wrapper description removes the need for the `With data attributes` example?

AshGDS:
Also minor non-blocking thing, you might not need to add the `text` value in the JSON? Unless this was done for a reason?

andysellick:
@AshGDS thanks, good points. 

Yes, the component wrapper description removes the need for the data attributes example (I have a vague thought of improving the default component wrapper stuff at some point, but no one's complained it isn't detailed enough so far, so maybe it's fine?)

Also yes, hadn't remembered that about text in the link tracker, have removed the explicit inclusion.

AshGDS (approved):
Great, thanks Andy 👍

PR #3843, opened 2024-01-29
Add GA4 link tracking to govspeak callout links

AshGDS:
## What

- Adds the `ga4-link-tracker` to the highest parent level element of `govspeak` content
- I tried to use the component wrapper, but it did not like the fact that this component uses classes that don't start with `js-`. So I reverted my component wrapper code in one of the commits.
- I had to write some code to make sure nested links with their own `data-ga4-link` don't end up firing tracking twice, as if the govspeak `<div>` that wraps all the content also has a `data-ga4-link` attribute on it then two link trackers would fire on the same link.

## Why

https://trello.com/c/bh96e5IG/776-add-tracking-callout-links

## Visual Changes

None except adding the docs example

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
@@ -52,6 +52,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
         for (var i = 0; i < classes.length; i++) {
           if (target.closest('.' + classes[i].trim())) {
+            // Stops the link tracker firing twice if the link itself has data-ga4-link, and the parent element does as well.
+            if (target.closest('[data-ga4-link]') !== target.closest('[data-ga4-limit-to-element-class]')) {
+              return
+            }
             this.trackClick(event)
           }
         }
--- a/app/views/govuk_publishing_components/components/_govspeak.html.erb
+++ b/app/views/govuk_publishing_components/components/_govspeak.html.erb
@@ -5,13 +5,30 @@
   direction_class = "direction-#{direction}" if local_assigns.include?(:direction)
   disable_youtube_expansions = local_assigns.fetch(:disable_youtube_expansions) if local_assigns.include?(:disable_youtube_expansions)
 
-  classes = []
-  classes << direction_class if direction_class
-  classes << "disable-youtube" if disable_youtube_expansions
-  classes << "gem-c-govspeak--inverse" if inverse
+  data_attributes ||= {}
+  disable_ga4 ||= false
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-govspeak govuk-govspeak")
+  component_helper.add_class(direction_class) if direction_class
+  component_helper.add_class("disable_youtube") if disable_youtube_expansions
+  component_helper.add_class("gem-c-govspeak--inverse") if inverse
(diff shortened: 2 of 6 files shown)


app/views/govuk_publishing_components/components/_govspeak.html.erb
```diff
@@ -9,9 +9,24 @@
...
+      ga4_track_links_only: "",
+      ga4_limit_to_element_class: "call-to-action, info-notice, help-notice, advisory",
+      ga4_link: { "event_name": "navigation", "type": "callout" }.to_json,
+    })
+  end
+
 %>
 
-<div class="gem-c-govspeak govuk-govspeak <%= classes.join(" ") %>" data-module="govspeak">
+<%= tag.div(class: "gem-c-govspeak govuk-govspeak " + classes.join(" "), data: data_attributes) do %>
```
andysellick, line 29: Unrelated - I'm wondering if allowing classes prefixed with `direction-` should be allowed by the component wrapper.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
```diff
@@ -52,6 +52,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
         for (var i = 0; i < classes.length; i++) {
           if (target.closest('.' + classes[i].trim())) {
+            // Stops the link tracker firing twice if the link itself has data-ga4-link, and the parent element does as well.
```
andysellick, line 55: I see you've done quite a bit of work to fix this, but can you explain how this problem arose? Do we have govspeak components that are already wrapped with the link tracker? If that's the case we should ideally remove that tracking, if the component will now track by default. Would that remove the need for this extra code?

app/views/govuk_publishing_components/components/_govspeak.html.erb
AshGDS: @andysellick I think as you created the component wrapper you have the final say 😄.  Though it's probably something that should be done in a separate PR - firstly because `direction_class` is passed through at the govspeak render time, so it might break things if somewhere in our code `direction_class` is passed through with a different prefix. Another issue is there's also the `disable-youtube` class which I guess would have to be refactored. So I guess it's out of the scope of this PR?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
andysellick: Oh I see. Okay, that's fair enough. However it does make me wonder if there would be a cleaner way of doing tracking in govspeak, now we know what the complexities are (outside of the scope of this PR). Maybe something to consider separately.

app/views/govuk_publishing_components/components/_govspeak.html.erb
andysellick: Oh yes definitely - sorry, entirely non blocking comment.

andysellick (approved):

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js
AshGDS: @andysellick sounds good 👍

app/views/govuk_publishing_components/components/_govspeak.html.erb
AshGDS: No worries!

PR #3848, opened 2024-02-01
Allow inputs to be exempt from GA4 form tracker [REDACT] code

AshGDS:
## What
- For the contact form tracking, there are three text inputs currently being collected on form submission
- The analysts are interested in collecting one of these text inputs. Therefore, I needed a way to only include that one for tracking and redact the rest.
- Therefore, a new data attribute called `data-ga4-form-include-input` has been added. If this is placed on the input, it will be included in the `text` list instead of being `[REDACTED]`
- Not using `data-ga4-form-include-text` as I don't want the other three inputs being included in the list. One of them is there user's name for contact form so doing it this way this instead makes sure that is not included in the contact form tracking.

## Why

- Progresses https://trello.com/c/jFRhH4Dh/783-adding-tracking-contact-form

## Visual Changes

None

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
@@ -654,12 +654,17 @@ describe('GA4 core', function () {
         expect(builtEcommerceObject.search_results.ecommerce.items.length).toEqual(200)
       })
     })
+  })
 
-    it('populates the timestamp value correctly', function () {
-      GOVUK.analyticsGa4.core.sendData({})
-      var timestamp = Date.now().toString()
+  it('populates the timestamp value correctly', function () {
+    jasmine.clock().install()
+    jasmine.clock().mockDate(new Date(2050, 1, 1))
 
-      expect(window.dataLayer[0].timestamp).toEqual(timestamp)
-    })
+    GOVUK.analyticsGa4.core.sendData({})
+    var timestamp = Date.now().toString()
+
+    jasmine.clock().uninstall()
+
+    expect(window.dataLayer[0].timestamp).toEqual(timestamp)
   })
 })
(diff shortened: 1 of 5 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
```diff
@@ -658,8 +658,18 @@ describe('GA4 core', function () {
...
+      // There is sometimes a 1 second delay between the timestamp GA4 will generate and the timestamp this test generates.
+      // Therefore directly comparing timestamps makes this test flaky when the time between both lines of code executing is at least 1 second.
+      // Attempt to resolve this by instead checking the numbers are large enough and are within 5 seconds of each other instead.
+      var exampleTimestamp = 1701234567890
+      var areTimestampsLargeNumbers = timestamp > exampleTimestamp && ga4Timestamp > exampleTimestamp
+
+      var timestampDifference = timestamp - ga4Timestamp
+      var areTimestampsCloseEnough = timestampDifference <= 5 && timestampDifference >= -5
+
+      expect(areTimestampsLargeNumbers && areTimestampsCloseEnough).toEqual(true)
```
andysellick, line 672: Good spot - I was sure I've seen this test fail occasionally! Are you aware of the `jasmine.clock()` functionality? Take a look at the option select test for an example - it lets you fast forward through time, so it might allow a more specific test (feel free to split this out into a separate PR if it's easier).

andysellick (commented):
Nice solution to the form thing. Happy to approve but had a suggestion about the flaky test.

AshGDS:
@andysellick Thanks, should be ready for another review 👍

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
andysellick: Yeah, it's pretty cool - a lot like Timecop, which we use in a few of our other applications.

andysellick (approved):

PR #3849, opened 2024-02-01
Update cookie event listeners

andysellick:
## What
We have various bits of code that should only initialise if cookie consent is given. Since this can be given after page load (by the user clicking on the cookie banner) we have event listeners in place for the `cookie-consent` event, which is triggered by user consent.

The JS documentation states that these event listeners should be removed once triggered, but almost none of our code does this. This PR fixes this oversight, as well as updating some tests relating to these scripts. It also includes the same change for the RUM script, which is slightly different (hence a separate commit).

## Why
Following good practice. Also will make working on the single consent API changes easier. These commits were originally part of https://github.com/alphagov/govuk_publishing_components/pull/3829 but separating into their own PR to make that one smaller.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update cookie event listeners ([PR #3849](https://github.com/alphagov/govuk_publishing_components/pull/3849))
 * Remove DI cookie consent code ([PR #3846](https://github.com/alphagov/govuk_publishing_components/pull/3846))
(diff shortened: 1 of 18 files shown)


andysellick:
@AshGDS not sure if this should have a changelog entry or not - any thoughts?

AshGDS (approved):
LGTM 👍 

I think it's worth having a changelog, as if this caused problems in another app the dependabot description would show the devs that this change had been made in the gem

Issue #3776, opened 2023-12-15
Sass deprecation warnings


jon-kirwan:
## What

After migrating to Dart Sass, several deprecation warnings are flagged during compilation. Most of these appear to relate to using `/` for division e.g.

```
Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.
Recommendation: math.div($govuk-gutter, 2) or calc($govuk-gutter / 2)
```

PR #3864, opened 2024-02-07
Fix some Sass deprecation warnings

jon-kirwan:
## What

Fix some Sass deprecation warnings related to using `/` for division outside of `calc()`.

## Why

Using `/` for division outside of `calc()` is deprecated and will be removed in Dart Sass 2.0.0.

## Visual Changes

None.

## Anything else
- previous attempts to fix the warnings led to errors when precompiling assets in applications that bundle `sassc-rails` (LibSass) https://github.com/alphagov/govuk_publishing_components/issues/3776 and https://github.com/alphagov/govuk_publishing_components/pull/3810
- currently, only six deprecation warnings remain
- these warnings are caused by using the `calc()` CSS function as an argument inside of another Sass function or mixin
- the issue seems to have been reported previously: https://github.com/alphagov/govuk-frontend/issues/1350
- perhaps the remaining warnings can be fixed when we remove support for applications that bundle `sassc-rails`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Move the skip link after the cookie banner ([PR #3863](https://github.com/alphagov/govuk_publishing_components/pull/3863))
 * Update border colours on email/print buttons for greater contrast ([PR #3855](https://github.com/alphagov/govuk_publishing_components/pull/3855))
+* Fix some Sass deprecation warnings ([PR #3864](https://github.com/alphagov/govuk_publishing_components/pull/3864))
 
 ## 37.3.0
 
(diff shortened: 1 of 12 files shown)


andysellick (approved):
Looks good to me, thanks for persisting with this. Is there a way we can test it in an application before getting this into a new release? Just remembering what happened last time.

Also personal preference I'd squash these two commits as the changelog entry is for the change.

jon-kirwan:
> Looks good to me, thanks for persisting with this. Is there a way we can test it in an application before getting this into a new release? Just remembering what happened last time.
> 
> Also personal preference I'd squash these two commits as the changelog entry is for the change.

Thanks for the review @andysellick. Yep, I've tested in a few places including:
- Collections (which now bundles Dart Sass)
- Signon (which still uses [`sassc-rails`](https://github.com/alphagov/signon/blob/main/Gemfile#L27).

Both compile the Sass without any problems: Collections reports 6 warnings and then precompiles the assets correctly, whereas Signon doesn't report any warnings or errors and assets are again precompiled with no issues.

I've also squashed the commits together.

andysellick:
Great stuff @jon-kirwan good work 👍

PR #3871, opened 2024-02-12
Add Breadcrumb Handling for Mainstream Browse Pages with no Parent

Tetrino:
## What
This logic rewrite ensures that:

- In cases where specialist (Whitehall) content is tagged to a Mainstream Browse Page **and** its tagged parent is also a MSB, the breadcrumbs are set by ancestry (which was the original behaviour, minus the parent check).
- In cases where specialist (Whitehall) content is tagged to a MSB but there is no parent (or the parent is not an MSB), the breadcrumbs are set from the taxonomy.

## Why
When guidance pages were linked to mainstream browse pages after their related specialist topic was archived, the pages had no parent tag. They were tagged to taxonomy, but this was not reflected in the breadcrumb, only showing "Home."

https://trello.com/c/ls1MlMKC/2369-breadcrumb-disappears-when-whitehall-content-is-tagged-to-msb-m

### Before
Here is an example [MSB tagged guidance page without a parent](https://www.gov.uk/guidance/agricultural-relief-on-inheritance-tax), as it exists on Production right now:
image

### After
And here is it with this change:
image

### DIFF
--- a/lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
+++ b/lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
@@ -45,7 +45,7 @@ module GovukPublishingComponents
             step_by_step: false,
             breadcrumbs: navigation.taxon_breadcrumbs,
           }
-        elsif navigation.content_tagged_to_mainstream_browse_pages?
+        elsif navigation.content_parent_is_mainstream_browse?
           {
             step_by_step: false,
             breadcrumbs: navigation.breadcrumbs,
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   of the commit log.
 
 ## Unreleased
-
+* Ensure whitehall content tagged to mainstream browse and the topic taxonomy has a topic taxonomy breadcrumb ([PR #3871](https://github.com/alphagov/govuk_publishing_components/pull/3871))
 * Add GA4 tracking to the document list component ([PR #3874](https://github.com/alphagov/govuk_publishing_components/pull/3874))
 * Remove GA4 'action' from navigation content history events ([PR #3877](https://github.com/alphagov/govuk_publishing_components/pull/3877))
 
--- a/lib/govuk_publishing_components/presenters/contextual_navigation.rb
+++ b/lib/govuk_publishing_components/presenters/contextual_navigation.rb
@@ -68,6 +68,10 @@ module GovukPublishingComponents
         content_item.dig("links", "mainstream_browse_pages").present?
       end
 
+      def content_parent_is_mainstream_browse?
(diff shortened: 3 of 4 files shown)


lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -45,11 +45,16 @@ def options(navigation)
             step_by_step: false,
             breadcrumbs: navigation.taxon_breadcrumbs,
           }
-        elsif navigation.content_tagged_to_mainstream_browse_pages?
+        elsif navigation.content_tagged_to_mainstream_browse_pages? && navigation.content_parent_is_a_mainstream_browse_page?
```
hannako, line 48: What do you think about just having one method in ContextualNavigation called `content_parent_is_mainstream_browse?`. As that's all the breadcrumb selector needs right? It's not using the fact that it's tagged to mainstream browse.

Tetrino: So changing this check actually breaks the logic somehow and causes a failure where the breadcrumbs are incorrect (text expecting taxons rather than parent). Investigating. There are definitely cases set up in the test environment where the parent is an MSB but there is no content tagged, so I broke Taxon rendering in that case by removing the check that it's tagged. Oops.

Tetrino: I attempted this but it appears doing so breaks cases where a page has a MSB parent but no MSB tags. Flagged by https://github.com/alphagov/govuk_publishing_components/blob/add-mainstream-topics-to-breadcrumbs/spec/components/contextual_breadcrumbs_spec.rb#L96

This flags that there are cases where a we don't want to render the parent if it exists, specifically when there are now MSB related content tagged to the page. I'm hesitant to rework this behaviour at this point as I am unsure on the splash it will have.

I have however renamed the function as your suggestion here is much cleaner.

hannako: > cases where a page has a MSB parent but no MSB tags

I don't think that is a possible scenario. AFAIK it's only Publisher which allows pages to have a MSB added to the parent field (in the UI it's called `Breadcrumb`), and there's a restriction that the MSB page chosen for the parent field has to match the MSB page add to the MSB page field
Screenshot 2024-02-12 at 12 05 49

hannako: So unless I'm missing something, it sounds like the tests contain unrealistic data.

Tetrino: Is it possible that it was once valid but no longer is? Either way I've just removed both the redundant check MSB and the invalid test. Now only checks for parent.

beccapearce (approved):

PR #3872, opened 2024-02-12
Expand GA4 share link tracking to allow for extra values

AshGDS:
## What

- Adds a variable called `ga4_data` to the share links component.
- This variable allows an object to be passed through to the component with extra GA4 values.

## Why

- The PAs would like us to add a `section` value for share links. As the share link component is rendered in a range of different places, the `section` will need to be passed through wherever the component is rendered.
- https://trello.com/c/gAiBgeTx/785-add-tracking-organisation-pages

## Visual Changes

The component example was updated to reflect the changes


JamesCGDS (approved):
LGTM 👍

PR #3874, opened 2024-02-14
Add GA4 tracking to the document list component

AshGDS:
## What
- Adds GA4 tracking to the document list component
- Allows values to be expanded by passing through `ga4_data` object, similar to https://github.com/alphagov/govuk_publishing_components/pull/3872
- Note that we have tracking enabled on components by default now, and use `disable_ga4` to turn it off if needed. We do this instead of using `ga4_tracking: true` like we used to.

## Why

- Required for our organisation pages tracking 
- https://trello.com/c/gAiBgeTx/785-add-tracking-organisation-pages

## Visual Changes

Changed the document list examples to reflect the new tracking


JamesCGDS:
Quick question - would we want to update the [With data attributes on links](https://github.com/alphagov/govuk_publishing_components/blob/53423ed6b1be416c829532572430c7d1a03ee034/app/views/govuk_publishing_components/components/docs/document_list.yml#L75) section in the documentation for this component? Currently it says "[Note that the component does not include built in tracking. If this is required consider using the track click script](https://github.com/alphagov/govuk_publishing_components/blob/53423ed6b1be416c829532572430c7d1a03ee034/app/views/govuk_publishing_components/components/docs/document_list.yml#L76)"

If ga4 tracking is now enabled by default, would it be worth clarifying that the above is referring to UA?

JamesCGDS (changes requested):
Looking good (just got a small change request) 👍

spec/components/document_list_spec.rb
```diff
@@ -533,4 +533,103 @@ def component_name
...
+            rel: "external",
+          },
+        },
+      ],
+    )
+
+    assert_select "li[data-module=ga4-link-tracker]", false
+    assert_select "li[data-ga4-track-links-only]", false
+    assert_select "li[data-ga4-link='{\"event_name\":\"navigation\",\"type\":\"document list\"}']", false
+  end
```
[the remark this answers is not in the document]
AshGDS: Makes sense - added a new test to reflect this, thanks.

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -17,14 +17,28 @@
...
+
 %>
 <% if items.any? %>
   <% unless within_multitype_list %>
     <ul class="<%= classes.join(" ") %>">
   <% end %>
     <% items.each do |item| %>
       <% highlight_class = " gem-c-document-list__item--highlight" if item[:highlight] %>
-      <li class="gem-c-document-list__item<%= within_multitype_list_class %> <%= brand_helper.brand_class %> <%= highlight_class %>">
+      <%= tag.li(class: "gem-c-document-list__item#{within_multitype_list_class} #{brand_helper.brand_class} #{highlight_class}", data: li_data_attributes) do %>
```
andysellick, line 41: Using this approach it looks like the GA4 link tracker module would be initialised individually for each list item. Did you look at adding the module to the list instead?

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -17,14 +17,28 @@
...
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
+
+  li_data_attributes ||= {}
+
+  disable_ga4 ||= false
+  unless disable_ga4
+    ga4_data ||= {}
+    li_data_attributes[:module] = li_data_attributes[:module] ? li_data_attributes[:module] << " ga4-link-tracker" : "ga4-link-tracker"
+    li_data_attributes[:ga4_track_links_only] = ""
```
andysellick, line 27: Just need to check for my own understanding - why is `track-links-only` needed here?

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -17,14 +17,28 @@
 
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
+
+  li_data_attributes ||= {}
+
+  disable_ga4 ||= false
```
andysellick, line 23: I'd reserve the `disable_ga4` option for components where the tracking is entirely handled automatically. For situations like this, where the tracking only works if attributes are passed, would it make more sense to leave out this option but make the developer include the right module in the passed data attributes?

andysellick: (similar to the approach taken for UA tracking)

spec/components/document_list_spec.rb
```diff
@@ -533,4 +533,128 @@ def component_name
...
+          },
+        },
+      ],
+    )
+
+    assert_select "li[data-hello=world]"
+    assert_select "li[data-another=example]"
+  end
+
+  it "can accept data attributes to be added to list elements when GA4 tracking is disabled" do
```
andysellick, line 558: This is a good test, but when would this situation occur? I worry that it's a bit misleading - surely if you can pass any data attributes to the list item even when `disable_ga4` is true, it would still be possible to manually pass a GA4 tracking module and still track despite this?

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: @andysellick I decided not to add it to the `<ul>` due to the existence of that `within_multitype_list` boolean. I was wondering if it was set to `true` then the tracking would not exist. Can change it though if you think it should be on the `ul` instead.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: @andysellick The `<li>` can contain elements other than links, so using `track-links-only` will cancel the link tracking code at an early stage if the element is not a link or within. If this attribute doesn't exist it will try and run some extra lines of code, so I guess it's a minor optimisation. Though I could remove this, and just put the tracking directly on the two link elements within the list.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: @andysellick Sorry for the confusion, here the tracking here can still work when no attributes are passed through to the component. The `ga4_data` is for any extra GA4 values can be added via the data attributes - maybe I should have called it `extra_ga4_data` instead (I can update this name for the share links tracking too if preferred.) I will be using this attribute to pass the `section` value down to the tracking.

Though I'm still happy to remove the tracking from the component and pass it through as data attributes instead, if you prefer that?

spec/components/document_list_spec.rb
AshGDS: @andysellick Good point. I guess I saw it more as a test that `li_data_attributes` code will still works separately from the `ga4_tracking` code, since they are separate features of the component. Though happy to remove if you still think it isn't needed.

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: Looks like we've stumbled into a little rabbit hole here - the `within_multitype_list` option looks like it was added as part of [some other component work](https://github.com/alphagov/govuk_publishing_components/commit/40e9a9d91524c43e5604ae1d18f8b507935ec192) some time ago and was used by another component that wrapped the document list component, which has [since been removed](https://github.com/alphagov/govuk_publishing_components/commit/44e7136052233baf99ef0b4487db83eb84e7e3f5). I suspect this option is no longer in use and should be removed (there's no documentation for it either). That's obviously out of scope for this PR but might be worth raising separately, as another PR or as an issue.

I think if we implement tracking using this approach it would be cleaner to put the module on the parent rather than initialising it multiple times but we should decide the best approach (I'm still thinking about it in light of your responses to my comments).

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: I think it would make more sense to apply the tracking directly to the links instead of the list items but let's check and see how much effort would be involved first. If it's a straight switch then I'd suggest go for it, but otherwise happy to discuss.

Second potential rabbit hole is the `track-links-only` option and why we have it at all instead of just rewriting the code to only track link clicks when the data attributes are on the parent. But again out of scope of this PR and maybe something for me to look at.

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: Gone through the rest of your comments - yes, this approach works, if you can put the module on the parent I think that would be better 👍

spec/components/document_list_spec.rb
andysellick: I think I have clearer understanding of all this now, thanks. In light of that I think the test is necessary, but unless there's a situation where we would disable GA4 but still want to add attributes to this bit of the component (I can't think of one, but happy to consider options) I'd rewrite the code to make the `disable_ga4` option prevent any attributes from being added, regardless of what's passed. Does that seem sensible?

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: Thanks, turns out we can put everything on the `<ul>` so I've done that instead.

I think `track-links-only` came from the UA tracking, I think it was our initial solution to track links rendered within `govspeak` without the tracking firing on non-link elements.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: Thanks, I removed `within_multitype_list` in this PR however as the `unless` was interfering with the `<% end %>` statement I needed when using `<%= tag.ul() do %>` - the `<% end %>` for the `unless` statement was closing the `ul` tag instead of being used for the unless statement.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: Thanks, I've renamed it to `ga4_extra_data`

spec/components/document_list_spec.rb
AshGDS: Thanks - this is no longer needed so I've removed the test.

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: Ah cool in that case `track-links-only` works well here, nice work 👍

app/views/govuk_publishing_components/components/_document_list.html.erb
andysellick: Nice, thanks, I see you've made the corresponding change on the share links component as well 👍

spec/components/document_list_spec.rb
andysellick: Good stuff 👍

app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -17,11 +17,22 @@
...
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
+
+  disable_ga4 ||= false
+  unless disable_ga4
+    ga4_extra_data ||= {}
+    ga4_ul_attributes = {
+      module: "ga4-link-tracker",
+      ga4_track_links_only: "",
+      ga4_link: {
+      "event_name": "navigation",
```
andysellick, line 28: Nitpick: indentation?

andysellick:
@AshGDS the visual diff test is mostly failing because of recently merged changes to the crown logo - if you rebase they should disappear.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: Thanks @andysellick - I've removed the extra references. Happy to look into the component wrapper as a separate PR 👍

AshGDS:
Thanks @andysellick - should be ready to review again. The visual failures seem to be because I added two new examples to the document list docs

andysellick (approved):
👏

PR #3885, opened 2024-02-26
Add GA4 tracking to the notice link component

AshGDS:
## What

- Add GA4 tracking to the notice link component
- This component contains a nested `govspeak` component, so we disable tracking on this component so that our tracking is not duplicated by possibly having two link trackers running.

## Why

https://trello.com/c/f4i7CKvO/791-add-tracking-notice-links

## Visual Changes

Component example updated with a `With ga4 tracking disabled` example

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/notice.yml
+++ b/app/views/govuk_publishing_components/components/docs/notice.yml
@@ -65,3 +65,13 @@ examples:
       description_govspeak: <p>This document is no longer current. We published a new version on 30 September 2015.</p>
       show_banner_title: true
       banner_title: "Information"
+  with_ga4_tracking_disabled:
+    description: |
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/notice.yml
```diff
@@ -65,3 +65,13 @@ examples:
       description_govspeak: <p>This document is no longer current. We published a new version on 30 September 2015.</p>
       show_banner_title: true
       banner_title: "Information"
+  with_ga4_tracking_disabled:
+    description: |
+      Disables GA4 tracking on the component. Tracking is enabled by default. When tracking is enabled, a data module and data attributes is added to the parent element. See the [ga4-link-tracker documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/ga4-link-tracker.md) for more information.
```
andysellick, line 70: Maybe worth mentioning the disabling of tracking on govspeak if passed? That was helpful to know.

andysellick (approved):
One minor comment, but nothing blocking.

PR #3888, opened 2024-02-28
Adjust keyboard functionality of dropdown menu

davidtrussler:
Trello cards: 
- [Adjust keyboard functionality of dropdown menu [L]](https://trello.com/c/ke3jvHve/2400-adjust-keyboard-functionality-of-dropdown-menu-l)
- [Unexpected focus order when activating menu button in header](https://trello.com/c/v6hdKvtj/11-unexpected-focus-order-when-activating-menu-button-in-header)

## What
These changes update the behaviour of some keyboard functionality on the header component: 
- Adds a `handleKeyDown` method to the `layout-super-navigation-header` component to handle 'tab' and 'escape' key events
- Adds tests for the new functionality
- Updates the trigger-event lib component to accept values for the shift key

## Why

There are a some accessibility failures on the header whereby
- a user navigating with the tab key traverses the elements in an unpredictable and confusing order, specifically
  - If the "Menu" button is focussed and the menu is open pressing the tab key moves focus to the "Search" button where expected behaviour is to move to the first link in the menu
  - If the last link of the menu is focused, pressing the tab key moves focus away from the header completely where expected behaviour is to move focus to the "Search" button
  - If the menu is closed, focus remains where it is where expected behaviour is that it should move to the "Search" button
- the escape key does not follow the expected behaviour of closing the navigation

## Visual Changes
|Action|Before|After|
|-|-|-|
|Nav menu open<br/>"Menu"focussed<br/>tab key pressed|Screenshot 2024-02-28 at 15 00 47|Screenshot 2024-02-28 at 15 01 06|
|Nav&nbsp;menu&nbsp;open<br/>last&nbsp;link&nbsp;focussed<br/>tab&nbsp;key&nbsp;pressed|Screenshot 2024-02-28 at 15 08 17|Screenshot 2024-03-04 at 11 57 15|
|Nav&nbsp;menu&nbsp;open<br/>link&nbsp;focussed<br/>escape&nbsp;key&nbsp;pressed|Screenshot 2024-02-28 at 15 09 52|Screenshot 2024-03-04 at 11 55 47|
|Search&nbsp;menu&nbsp;open<br />Last&nbsp;link&nbsp;focussed<br />Tab&nbsp;key&nbsp;pressed|Screenshot 2024-03-04 at 15 29 08|Screenshot 2024-03-04 at 15 29 17|
|Search&nbsp;menu&nbsp;open<br />Link&nbsp;focussed<br />Escape&nbsp;key&nbsp;pressed|Screenshot 2024-03-04 at 15 01 23|Screenshot 2024-03-04 at 15 01 34|

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
@@ -93,6 +93,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.$module = $module
     this.$searchToggle = this.$module.querySelector('#super-search-menu-toggle')
     this.$searchMenu = this.$module.querySelector('#super-search-menu')
+    this.$navToggle = this.$module.querySelector('#super-navigation-menu-toggle')
+    this.$navMenu = this.$module.querySelector('#super-navigation-menu')
 
     // The menu toggler buttons need three attributes for this to work:
     //  - `aria-controls` contains the id of the menu to be toggled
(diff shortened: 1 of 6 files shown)


app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -126,7 +126,56 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+          default:
+            break
+        }
+      }
+    } else if (event.keyCode === KEY_ESC) {
+      if (!$navMenu.hasAttribute('hidden')) {
+        hide($navToggle, $navMenu)
+        this.$searchToggle.focus()
+      }
+    }
```
MartinJJones, line 172: I just want to double-check what behaviour should be here when the open menu is closed using the escape key.

At the moment it does match what is in the Trello card, in that, when you press the Esc key to close the menu, focus jumps to the search menu toggle button, but not sure where that recommendation has come from, I cannot see this anywhere in the comments or from the DAC audit.

Looking at the WCAG documentation, I think the focus state should jump back to the Menu toggle button instead - https://www.w3.org/WAI/WCAG20/Techniques/failures/F85, but would be good to check with an accessible specialist on this.

Lastly, do we also need to add the escape key functionality to the search menu as well?

MartinJJones (commented):
Great work on adding in this functionality!

The menu feels much better to use when testing locally.

I've made a few suggestions and have one question regarding what should happen to the focus order when the menu is closed.

Outside of this, I think we should also add something to the Accessibility acceptance criteria in the docs for this component 
 https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/layout_super_navigation_header.yml#L6-L17

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
davidtrussler: Good point on the expected behaviour on pressing the escape key when the Navigation Menu is open. Wider opinion agrees with you that this should return the user to the means by which they opened it - i.e. the Menu toggle button so this has been updated here.

Although it isn't part of the audit report it would seem to make sense to apply the same functionality to the Search Menu so that has been done as well. 

Additionally the scenario of tabbing from the final link in the navigation has also been added to the Search Menu since this will otherwise leave that menu open as the user tabs through the rest of the page.

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -126,7 +126,56 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     toggle($target, $targetMenu)
   }
 
+  SuperNavigationMegaMenu.prototype.handleKeyDown = function (event) {
+    var KEY_TAB = 9
+    var KEY_ESC = 27
+    var $navToggle = this.$module.querySelector('.gem-c-layout-super-navigation-header__navigation-top-toggle-button')
+    var $navMenu = this.$module.querySelector('.gem-c-layout-super-navigation-header__navigation-dropdown-menu')
```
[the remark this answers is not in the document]
davidtrussler: Ah right, yes that makes sense. Have updated those.

davidtrussler:
> Great work on adding in this functionality!
> 
> The menu feels much better to use when testing locally.
> 
> I've made a few suggestions and have one question regarding what should happen to the focus order when the menu is closed.
> 
> Outside of this, I think we should also add something to the Accessibility acceptance criteria in the docs for this component https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/layout_super_navigation_header.yml#L6-L17

@MartinJJones I've updated the PR based on the points you raised - could you look it over again? I added the escape key stuff to the search menu then realised that should also have the same change as the nav menu to move the focus from the last link on to the next focusable element, so that has been added as well. Cheers 🙂

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
MartinJJones: Nice!, thanks for double-checking and adding the additional functionality to the Search Menu as well, works great!

MartinJJones (approved):
Awesome work on adding this functionality to the super navigation menu @davidtrussler !, I've tested locally in several browsers and it works as expected.

There is just one small non-blocking change to make in the changelog, will likely just need to rebase and create a new "Unreleased" heading to move this entry to.

PR #3893, opened 2024-02-29
Ensure analytics stops firing if a user disables usage cookies on our cookie settings page

AshGDS:
## What

- Currently, if a user has `usage` cookies set to `true`, and then disables them on `/help/cookies`, our tracking still fires on the cookie settings page until they refresh/navigate away. (This doesn't affect other pages on GOV.UK as our JS will reinitialise every time they go to another page, which then checks their consent status.)
- I have fixed this by setting a boolean called `window.GOVUK.stopSendingAnalytics` to `true` when they disable usage cookies. This boolean is then checked in Universal Analytics and GA4 before they send tracking data.
- I've set `window.GOVUK.LUX` to an empty object (`{}`) when they disable the usage cookie as well, but I'm not sure how to test if this actually doing anything.
- I'm assuming the only tracking stuff we have on our site is UA, GA4, and LUX. I'm going by this page: https://www.gov.uk/help/cookie-details

## Why

- Fixes a bug that has existed or years on the cookie settings page.
- I'm not checking the consent cookie directly to stop sending analytics, (i.e. `if usage = false, dont send analytics`) because this caused 200+ of our tests to start failing.
    -  I assume this is because a lot of our tests send tracking data without worrying what our usage cookie is set to, and if I used the usage cookie as a guard at the time we send tracking data, those 200 tests would then stop working. Therefore, using this boolean was a lot simpler.
    -  I also assume using the consent cookie might have impact on the single consent work, as trying to get those 200 tests to work would probably cause a massive merge conflict for the consent PR.
    -  I also thought that since UA will be removed soon, it would be a lot of effort to try and fix all the tests if they'll just be removed eventually.
- I thought about just setting `window.GOVUK.analytics` and `window.GOVUK.analyticsGa4` to empty objects once the usage cookie is false as well, but this also caused issues with the tests - even when I tried to revert the UA/GA4 objects after the tests, the tests were acting strangely. Doing this would have also caused console errors on the cookie settings page, as the existing tracking event listeners would try and run tracking functions that no longer exist.
- Trello card: https://trello.com/c/vAI4kvX5/795-tracking-is-still-active-on-cookie-settings-page-after-rejecting-consent

## How to test
- Run `static` locally, pointing to this branch of the gem
- Run `frontend` locally, pointing to your local `static`, and this branch of the gem
- Go to the cookie settings page, `http://127.0.0.1:3005/help/cookies` with your usage cookie set to `true`.
- If you expand/collapse our header menu, you will see UA events in `Omnibug`, and GA4 events in the `dataLayer`
- If you then set the usage cookie to false via the cookie settings form, you will no longer get any additional `Omnibug` events or `dataLayer` events if you interact with the page, for example expanding the header menu again.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
@@ -39,6 +39,11 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
     },
 
     sendData: function (data) {
+      // Allows us to stop sending tracking at the moment a user sets their usage cookies to "false" on the cookie settings page.
+      if (window.GOVUK.stopSendingAnalytics) {
+        return false
+      }
+
       data.govuk_gem_version = this.getGemVersion()
       data.timestamp = this.getTimestamp()
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
@@ -21,6 +21,11 @@
   Analytics.PIISafe = PIISafe
 
   Analytics.prototype.sendToTrackers = function (method, args) {
+    // Allows us to stop sending tracking at the moment a user sets their usage cookies to "false" on the cookie settings page.
+    if (window.GOVUK.stopSendingAnalytics) {
+      return false
+    }
+
     for (var i = 0, l = this.trackers.length; i < l; i++) {
       var tracker = this.trackers[i]
       var fn = tracker[method]
@@ -49,7 +54,7 @@
   Analytics.prototype.trackPageview = function (path, title, options) {
     arguments[0] = arguments[0] || this.defaultPathForTrackPageview(window.location)
     if (arguments.length === 0) { arguments.length = 1 }
-    this.sendToTrackers('trackPageview', this.pii.stripPII(arguments))
+    return this.sendToTrackers('trackPageview', this.pii.stripPII(arguments))
   }
 
   /*
@@ -59,11 +64,11 @@
     options.nonInteraction – Prevent event from impacting bounce rate
   */
   Analytics.prototype.trackEvent = function (category, action, options) {
-    this.sendToTrackers('trackEvent', this.pii.stripPII(arguments))
+    return this.sendToTrackers('trackEvent', this.pii.stripPII(arguments))
   }
 
   Analytics.prototype.trackShare = function (network, options) {
-    this.sendToTrackers('trackSocial', this.pii.stripPII([network, 'share', global.location.pathname, options]))
+    return this.sendToTrackers('trackSocial', this.pii.stripPII([network, 'share', global.location.pathname, options]))
   }
 
   /*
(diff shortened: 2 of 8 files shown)


app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
```diff
@@ -49,7 +54,7 @@
   Analytics.prototype.trackPageview = function (path, title, options) {
     arguments[0] = arguments[0] || this.defaultPathForTrackPageview(window.location)
     if (arguments.length === 0) { arguments.length = 1 }
-    this.sendToTrackers('trackPageview', this.pii.stripPII(arguments))
+    return this.sendToTrackers('trackPageview', this.pii.stripPII(arguments))
```
andysellick, line 57: Why this change?

andysellick (commented):
Looks good, sensible approach. Just a couple of questions that probably have obvious answers but my brain is struggling right now 😁

app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
AshGDS: @andysellick `sendToTrackers` isn't a public function, so I can't access it in the tests. Therefore I return the value of it in the functions that are publicly accessible `trackPageview / trackEvent / trackShare` so that I can verify in the tests that analytics is disabled.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
```diff
@@ -39,6 +39,11 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {};
     },
 
     sendData: function (data) {
+      // Allows us to stop sending tracking at the moment a user sets their usage cookies to "false" on the cookie settings page.
+      if (window.GOVUK.stopSendingAnalytics) {
+        return 'analytics disabled'
```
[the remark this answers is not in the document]
AshGDS: @andysellick Done, should be ready for re-review. Thanks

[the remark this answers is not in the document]
andysellick: Can you change it in `ga4-core` as well?

app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
andysellick: Just to check, this doesn't impact the outcome? As in, does returning `this.sendToTracker` instead of executing it still call that function?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js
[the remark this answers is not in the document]
andysellick: Sorry, my suggestion has broken some of your tests - they're expecting `analytics disabled` instead of `false` now.

app/assets/javascripts/govuk_publishing_components/analytics/analytics.js
AshGDS: @andysellick It's still being executed, the result of the execution is what is returned. If it didn't have `()` on the function, then the function would be being returned instead of executed.

Example here if you want to play around with it: https://jsfiddle.net/spLfjez5/2/

andysellick: Cool, no worries, just wanted to be sure 👍

andysellick (approved):
Looks like you've addressed my comments, so happy to approve (remember to squash your commits).

PR #3896, opened 2024-03-01
Allow custom text for GA4 scroll tracker

andysellick:
## What
Rework the GA4 scroll tracker slightly so that when tracking headings or markers, the text collected in the event can be set to something else using a data attribute, instead of automatically collecting the text of the element.

Example:

```HTML
<h1>
  This text will be recorded
</h1>
<h2 data-ga4-text="this text will be recorded in the event">
  This text will be ignored
</h2>
```

## Why
Allows us to record the headings consistently in the GA4 data using one language. Currently getting a lot of data for when pages are translated in different languages, which analysis harder.

## Visual Changes
None.

Trello card: https://trello.com/c/VzWOeSJt/798-placeholder-scroll-tracking-heading-marker-question-translations


AshGDS (commented):
Looks good 👍 , though I think you're missing a test that checks `data-ga4-text` works?

andysellick:
@AshGDS 🤦 

After all that effort of changing around the tests and redoing all the logic, you're right, I completely failed to actually test the thing I meant to test. Have rewritten - and made a bit simpler too.

AshGDS (approved):
LGTM 👍

PR #3904, opened 2024-03-05
Update popular links in search header

hannako:
## What
Update popular links in the search in the header to match those on the body of the page, as defined in  https://github.com/alphagov/frontend/blob/main/config/locales/en.yml#L572-L584

## Why
Both parts of the page should contain the same links. These were missed from https://github.com/alphagov/frontend/pull/3524

## Visual Changes

|Before|After|
|-|-|
|Screenshot 2024-03-05 at 20 39 20|Screenshot 2024-03-05 at 21 25 56|

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
   of the commit log.
 
 ## Unreleased
-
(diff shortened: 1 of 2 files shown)


beccapearce (approved):

maxgds (approved):
LGTM - although we really should prioritise making this an update-once thing, I reckon.

PR #3907, opened 2024-03-08
Fix accessibility issue in contents list component

MartinJJones:
## What
Stop dashes before each 'Contents' list item being read by screen readers

## Why
Nearly every screen reader (except NVDA of the major ones) reads out the “em dash“ before each item/link in the “Contents“ list. Screen reader users found that frustrating to read through.

[Trello card](https://trello.com/c/MajjZ6ds/2414-dashes-before-each-contents-list-item-read-by-screen-readers-m-l)

## Accessibility testing

| Screen Reader | Browser | Issue fixed |
| --- | --- | --- |
| NVDA - 2023.3.3 | Firefox - 123 | fixed :white_check_mark: |
| Narrator - 21H2 | Edge 122 | fixed :white_check_mark: |
| JAWS 2022 | Chrome | fixed :white_check_mark: |
| JAWS 2023 | Chrome|  fixed :white_check_mark: |
| VoiceOver 10 | Safari 17.3.1 | fixed :white_check_mark: |
| VoiceOver 10 | Safari 17.3.1 | fixed :white_check_mark: |
| VoiceOver 10 | Firefox 123 | fixed :white_check_mark: |

## VoiceOver Before

https://github.com/alphagov/govuk_publishing_components/assets/28779939/54e964e0-c1bc-4f17-afe3-c924294cedc8

## VoiceOver After

https://github.com/alphagov/govuk_publishing_components/assets/28779939/122f55d3-3607-4c1b-afb9-f158b0be4a09

### Using `list-style-type`

We considered using [list-style-type](https://github.com/alphagov/govuk_publishing_components/pull/3907), however this did present us with 2 issues.

1. The dash style currently used it not available as a list style, we would have to choose something else such as “circle” or “disc” for example.
2. When testing in VoiceOver the list item style is still read out as “You are currently on a AXListMarker”

###  Using `speak`

We also explored other options, such as using the CSS speak property to turn off voice output for screen readers. On top of this, browser support is very limited currently, so this is not an option - ["speak" | Can I use... Support tables for HTML5, CSS3, etc](https://caniuse.com/?search=speak)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix accessibility issue in contents list component ([PR #3907](https://github.com/alphagov/govuk_publishing_components/pull/3907))
+
 ## 37.6.0
 
 * Add single cookie consent API ([PR #3854](https://github.com/alphagov/govuk_publishing_components/pull/3854))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -50,7 +50,7 @@
   padding-left: $contents-spacing;
   padding-right: $contents-spacing;
 
-  &::before {
+  & span::before {
     content: "—";
     position: absolute;
     left: 0;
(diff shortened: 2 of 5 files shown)


davidtrussler (approved):
LGTM 👍 - Nice fix!
Tested locally with Voiceover (using the combination of keys you listed) and worked fine compared to current implementation.

PR #3908, opened 2024-03-11
Include single consent api package

andysellick:
## What
Include the single consent API package in the distributed version of the gem.

## Why
The previous version didn't include this, which results in build time errors.

## Visual Changes
None.


AshGDS (approved):
Nice 👍

PR #3919, opened 2024-03-15
Add GA4 developer guide

andysellick:
## What
Add a developer guide for GA4. Also separate the consent stuff out into its own page (so we can link to it more easily from the implementation record).

## Why
There's some developer confusion around making changes to code containing tracking, this aims to improve understanding.

## Visual Changes
None.

### DIFF
--- a/docs/analytics-ga4/analytics.md
+++ b/docs/analytics-ga4/analytics.md
@@ -4,6 +4,8 @@ This document describes the use of Google Tag Manager (GTM) with Google Analytic
 
 No analytics code is initialised and no data is gathered without user consent.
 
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Nice, LGTM 👍 Non-blocking but I did find the `index_section` and `index_section_count` definitions a bit hard to understand. Maybe it's worth rewording them. It might be worth providing a more visual example? For example `<h2>` elements are usually what determine our sections, so maybe if we had some HTML in the doc that shows a couple sections and then the GA4 attributes they have.

andysellick:
Thanks @AshGDS, good feedback. I'm keen to get this right before merging, so have rewritten that section slightly and added an example. Is it any clearer? Is this a good enough example?

AshGDS (approved):
LGTM 👍

PR #3920, opened 2024-03-15
Add a new GA4 'focus loss' tracker

AshGDS:
## What

- Adds a new tracker called the `ga4-focus-loss-tracker`
- Have tested it with a couple elements in `static` and seems to be working as intended

## Why

- So that we can track what users are searching for on a client side DOM filter, without analytics spam that would occur if we tracked each `keypress`.
- https://trello.com/c/EXuAg9qR/801-add-tracking-filter-search-input
## Visual Changes

None.

### DIFF
--- a/docs/analytics-ga4/ga4-focus-loss-tracker.md
+++ b/docs/analytics-ga4/ga4-focus-loss-tracker.md
@@ -1,0 +1,16 @@
+# Google Analytics 4 focus loss tracker
+
+This script is intended for adding GA4 tracking to any element that needs to be tracked when focus is removed from it.
+
+## Basic use
+
+```html
+<div
+  data-module="ga4-focus-loss-tracker"
+  data-ga4-focus-loss='{ "event_name": "filter", "type": "filter" }'>
+</div>
+```
+
+If the tracker is initialised on an input that has the type `text` or `search`, the tracker will automatically grab the input value and set it as the `text` value in the GA4 JSON. This can be overridden by simply adding your own text value in `data-ga4-focus-loss`.
+
+This module was created to record what users were searching for in a client side DOM search filter. We did not want to track each keypress the user made as that would spam our analytics data. Therefore by tracking when focus is lost on the search input, we can see what keyword was leading to a user navigating off of the search box and onto the result they wanted.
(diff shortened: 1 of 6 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-focus-loss-tracker.spec.js
```diff
@@ -0,0 +1,233 @@
+/* eslint-env jasmine */
+
+describe('GA4 focus loss tracker', function () {
+  var GOVUK = window.GOVUK
+  var element
+  var expected
+
+  function agreeToCookies () {
```
andysellick, line 8: Non-blocking: I think we're using this kind of function in a few places now - worth raising an issue/adding a ticket to see if we can define it once in a helper somewhere instead of repeating it?

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-specialist-link-tracker.spec.js
```diff
@@ -740,7 +740,7 @@ describe('A specialist link tracker', function () {
     })
 
     afterEach(function () {
-      GOVUK.cookie('cookies_policy', null)
+      GOVUK.setCookie('cookies_policy', null)
```
andysellick, line 743: I thought that `.cookie` and `.setCookie` were equivalent, is that not the case? Was this causing an error?

docs/analytics-ga4/ga4-focus-loss-tracker.md
```diff
@@ -0,0 +1,16 @@
+# Google Analytics 4 focus loss tracker
```
andysellick, line 1: Good documentation. Can you include a link to it in `ga4-all-trackers.md`?

andysellick (approved):
Seems good, well done 👍 

Just a couple of minor things.

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-focus-loss-tracker.spec.js
AshGDS: Sounds good 👍

docs/analytics-ga4/ga4-focus-loss-tracker.md
AshGDS: Thanks, done 👍

PR #3923, opened 2024-03-15
Change navigation header GA4 indexes

andysellick:
## What / why
- recent removal of the 'popular links' section means that the `index_section_count` of the other links in the header should be reduced from 4 to 3

## Visual Changes
None.


davidtrussler (approved):
That makes sense to me. 👍

PR #3933, opened 2024-03-20
Use component wrapper in document list component

AshGDS:
## What

- Use component wrapper in document list component

## Why

- I promised I would in the PR that added GA4 tracking to the component

## Visual Changes

Document list yml examples will now have the uses component wrapper text

### DIFF
--- a/app/views/govuk_publishing_components/components/_document_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_document_list.html.erb
@@ -5,11 +5,13 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   items ||= []
 
-  classes = %w[gem-c-document-list]
-  classes << "gem-c-document-list--no-underline" if local_assigns[:remove_underline]
-  classes << "gem-c-document-list--no-top-border" if local_assigns[:remove_top_border]
-  classes << "gem-c-document-list--no-top-border-first-child" if local_assigns[:remove_top_border_from_first_child]
-  classes << shared_helper.get_margin_bottom
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-document-list")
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_document_list.html.erb
```diff
@@ -5,11 +5,14 @@
...
-  classes << "gem-c-document-list--no-top-border" if local_assigns[:remove_top_border]
-  classes << "gem-c-document-list--no-top-border-first-child" if local_assigns[:remove_top_border_from_first_child]
-  classes << shared_helper.get_margin_bottom
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-document-list")
+  component_helper.add_class("gem-c-document-list--no-underline") if local_assigns[:remove_underline]
+  component_helper.add_class("gem-c-document-list--no-top-border") if local_assigns[:remove_top_border]
+  component_helper.add_class("gem-c-document-list--no-top-border-first-child") if local_assigns[:remove_top_border_from_first_child]
+  margin_bottom = shared_helper.get_margin_bottom
+  component_helper.add_class(margin_bottom) if margin_bottom
```
andysellick, line 14: I think this `if margin_bottom` check might be redundant because `shared_helper.get_margin_bottom` always returns a value, so I think these two lines could be written more simply as

```
component_helper.add_class(shared_helper.get_margin_bottom)
```

andysellick (approved):
Nice. One small comment, but otherwise looks good. 

Good job following up on this. Can you update https://docs.google.com/spreadsheets/d/1J-0IPyr2ycPuw1IWOIoX0o9Tk9nF35D0DJdsj_ylLQE/edit?pli=1#gid=1458513074 with this change? Thanks.

app/views/govuk_publishing_components/components/_document_list.html.erb
AshGDS: @andysellick Thanks, will update and then merge 👍

PR #3939, opened 2024-03-20
Restores test and refactors component and test for layout-super-navigation-header

davidtrussler:
## What
These changes restores a test that was temporarily removed and refactors both that and the component it relates to. 

(Trello ticket: [Refactor and improve recent change to layout-super-navigation-header component](https://trello.com/c/zhgxoNEK/2485-refactor-and-improve-recent-change-to-layout-super-navigation-header-component))

## Why
When we made the changes in [PR#3918](https://github.com/alphagov/govuk_publishing_components/pull/3918) (Remove "Popular" links from super navigation header) we lost a recently added accessibility improvement, specifically to close the search menu when a keyboard user tabs out of that element. This was because this addition relied on the presence of links that were removed in the previous PR. We decided at the time to go ahead with that change and return to this to restore the lost behaviour. 

This restores that by refactoring the original method dealing with that so that it doesn't rely on the presence of links but broadens the selector to include other tab-able elements. It also makes a small change to the original code for the search menu to make both more consistent with each other.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
@@ -134,8 +134,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var $navMenuLinks = this.$navMenu.querySelectorAll('li a')
     var $firstNavLink = $navMenuLinks[0]
     var $lastNavLink = $navMenuLinks[$navMenuLinks.length - 1]
-    var $searchMenuLinks = this.$searchMenu.querySelectorAll('li a')
-    var $lastSearchLink = $searchMenuLinks[$searchMenuLinks.length - 1]
+    var $searchMenuTabbable = this.$searchMenu.querySelectorAll('li a, input, button')
+    var $lastSearchMenuTabbable = $searchMenuTabbable[$searchMenuTabbable.length - 1]
 
     if (event.keyCode === KEY_TAB) {
       if (!this.$navMenu.hasAttribute('hidden')) {
@@ -169,7 +169,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
             break
         }
       } else if (!this.$searchMenu.hasAttribute('hidden')) {
-        if (document.activeElement === $lastSearchLink) {
+        if (document.activeElement === $lastSearchMenuTabbable) {
           if (!event.shiftKey) {
             hide(this.$searchToggle, this.$searchMenu)
           }
--- a/spec/javascripts/components/layout-super-navigation-header-spec.js
+++ b/spec/javascripts/components/layout-super-navigation-header-spec.js
@@ -328,15 +328,21 @@ describe('The super header navigation', function () {
     var $navLinks
     var $firstNavLink
     var $lastNavLink
+    var $searchMenu
+    var $searchMenuTabbable
+    var $lastSearchMenuTabbable
 
     beforeEach(function () {
       thisModule.init()
       $navMenuButton = document.querySelector('#super-navigation-menu-toggle')
       $searchMenuButton = document.querySelector('#super-search-menu-toggle')
       $navMenu = document.querySelector('#super-navigation-menu')
-      $navLinks = $navMenu.querySelectorAll('li')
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -134,8 +134,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     var $navMenuLinks = this.$navMenu.querySelectorAll('li a')
     var $firstNavLink = $navMenuLinks[0]
     var $lastNavLink = $navMenuLinks[$navMenuLinks.length - 1]
-    var $searchMenuLinks = this.$searchMenu.querySelectorAll('li a')
-    var $lastSearchLink = $searchMenuLinks[$searchMenuLinks.length - 1]
+    var $searchMenuTabable = this.$searchMenu.querySelectorAll('li a, input, button')
```
MartinJJones, line 137: Small nitpick, should "Tabable" be "Tabbable" used in the variable names?

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
```diff
@@ -134,8 +134,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+    var $lastSearchTabable = $searchMenuTabable[$searchMenuTabable.length - 1]
```
MartinJJones, line 138: It makes the variable name a bit longer, but wonder if we should keep this consistent with the variable above, for example `lastSearchMenuTabbable`

spec/javascripts/components/layout-super-navigation-header-spec.js
```diff
@@ -328,23 +328,29 @@ describe('The super header navigation', function () {
...
       $searchMenuButton = document.querySelector('#super-search-menu-toggle')
       $navMenu = document.querySelector('#super-navigation-menu')
-      $navLinks = $navMenu.querySelectorAll('li')
-      $firstNavLink = $navLinks[0].querySelector('a')
-      $lastNavLink = $navLinks[$navLinks.length - 1].querySelector('a')
+      $navLinks = $navMenu.querySelectorAll('li a')
+      $firstNavLink = $navLinks[0]
+      $lastNavLink = $navLinks[$navLinks.length - 1]
+      $searchMenu = document.querySelector('#super-search-menu')
+      $searchTabable = $searchMenu.querySelectorAll('li a, input, button')
```
MartinJJones, line 344: Small nitpick, can we change the variable name here to be the same as the main JS file, so it would change from `searchTabbable` to `searchMenuTabbable`

spec/javascripts/components/layout-super-navigation-header-spec.js
```diff
@@ -374,6 +380,17 @@ describe('The super header navigation', function () {
 
       expect(document.activeElement).toEqual($lastNavLink)
     })
+
+    it('when the last search menu link is focussed and the tab key is pressed the search menu is closed', function () {
```
MartinJJones, line 384: I think we can make this a bit more descriptive now it is checking for more than the last link element, for example:

"when the last tabbable element in the search menu is focussed and the tab key is pressed the search menu is closed"

MartinJJones (commented):
Nice work on adding this functionality back in to the super navigation menu.

The code itself looks good to me, just made some small suggestions around the naming of things.

app/assets/javascripts/govuk_publishing_components/components/layout-super-navigation-header.js
davidtrussler: Yeah, fair nitpick! I hovered between the two a bit. Happy to go with "bb"!

davidtrussler:
> Nice work on adding this functionality back in to the super navigation menu.
> 
> The code itself looks good to me, just made some small suggestions around the naming of things.

Thanks @MartinJJones - I've made those changes and updated the CHANGELOG for this PR. (I'll squash the commits if it's all OK). Could you have another look over it? Cheers.

MartinJJones (approved):
Nice work, changes look good to me 👍

PR #3944, opened 2024-03-22
Update breadcrumb tests

hannako:
## What
Minor update to contextual breadcrumb tests to better represent live behaviour

## Why
To improve our confidence when we retire specialist topics in future work.


unoduetre (approved):
Approved.

PR #3946, opened 2024-03-22
Remove GA4 callout tracking from the govspeak component

AshGDS:
## What

- Remove the `callout` tracking from the `govspeak` component (again)

## Why

- We have deemed this too fiddly to keep in
- https://trello.com/c/P20AYeRP/810-remove-callout-link-tracking

## Visual Changes

None.


andysellick (approved):
Sorry - feels like a lot of back and forth on this bit of work, but well done for getting it this far.

PR #3951, opened 2024-03-25
Add discovery_engine_attribution_token to GA4 pageview

andysellick:
## What / why
Adds a new parameter to the GA4 page view, `discovery_engine_attribution_token`. This is a unique token generated by each search result, used for debugging purposes in search. Additional work will be done in `finder-frontend` to add this token as a meta tag so this code can read it - on all non-search pages this attribute will be undefined in the page view.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@
   of the commit log.
 
 ## Unreleased
+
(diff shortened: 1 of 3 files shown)


leenagupte (approved):
"searchyMcSearch" LOL!

LGTM! 👍 🙇‍♀️

PR #3952, opened 2024-03-25
docs: Fix test instructions and generally spruce up the README.

sengi:
The 'Run the test suite' instructions didn't work without a couple of build steps which weren't mentioned anywhere in the README. Add them to the example so that it works for new contributors.

Also remove some confusing information about GOV.UK Docker and generally tidy up the README to bring it in line with GDS tech writing style.

No changes to the gem itself apart from the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,28 @@
 [![JavaScript Style Guide](https://img.shields.io/badge/code_style-standard-brightgreen.svg)](https://standardjs.com)
 [![Stylelint Style Guide](https://img.shields.io/badge/code_style-stylelint-brightgreen.svg)](https://github.com/alphagov/stylelint-config-gds/)
 
-A Ruby gem to document and distribute components for GOV.UK applications. It provides:
+govuk_publishing_components is a Ruby gem to document and distribute components for GOV.UK applications. It provides:
 
 - a shared library of components
 - helpers to generate component payloads
 - an application to preview components and provide guidance
 - [GOV.UK Design System](https://design-system.service.gov.uk/) styles and components (via [GOV.UK Frontend](https://github.com/alphagov/govuk-frontend))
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -4,63 +4,75 @@
...
 - [Generate a new component](docs/generate-a-new-component.md)
 - [Component conventions](docs/component_conventions.md)
 - [Component principles](docs/component_principles.md)
 
 ### Further documentation
 
-- [Testing a component](docs/testing-components.md)
-- [Component auditing](docs/auditing.md)
+- [Test a component](docs/testing-components.md)
+- [Audit a component](docs/auditing.md)
```
andysellick, line 71: Nitpick: `Audit a component` to me implies that this link will give information about how to audit a single component, or that there's a process for doing an audit of a component. But the component auditing gives information about where and how the components are used, which I think is slightly different.

Now that I write this I'm wondering if 'audit' is even the right word for this thing, but that's probably out of scope here! Otherwise all these changes look good 👍

sengi: Ah cool, I'll leave 'Component auditing' as it was for now then. Thanks!

sengi:
Ready for another look :)

README.md
```diff
@@ -4,63 +4,75 @@
...
 
-To run JavaScript tests (only):
+### Run the JavaScript tests alone
+
+In browser:
 
 ```sh
-# run JS tests in browser
 yarn run jasmine:browser
+```
```
andysellick, line 42: Just realised this doesn't state that the tests can be seen in a browser at http://localhost:8888 - do you think it's worth adding? (non-blocking)

andysellick: Also worth noting that the test server needs a restart to pick up any changes to the JavaScript - again potentially out of scope, might be better to link to another page to go more in depth. Happy to raise a separate PR if you think it's useful.

andysellick (approved):

README.md
sengi: Ah cool, thanks! I'll add.

sengi: Done!

PR #3956, opened 2024-03-26
Fix anchoring in ID validation regex.

sengi:
Ruby regexes are — bizarrely — multiline by default, so `ComponentWrapperHelper::check_id_is_valid` would have erroneously accepted an ID string that contained newlines.

Fixes https://github.com/alphagov/govuk_publishing_components/security/code-scanning/8.

No visual changes. Should also be no functional change for normal usage.

### DIFF
--- a/lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
+++ b/lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
@@ -60,7 +60,7 @@ module GovukPublishingComponents
       def check_id_is_valid(id)
         return if id.blank?
 
-        raise(ArgumentError, "Id (#{id}) cannot start with a number or contain whitespace and can only contain letters, digits, `_` and `-`") unless /^[a-zA-Z][\w:-]*$/.match?(id)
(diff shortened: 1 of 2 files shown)


lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
```diff
@@ -60,7 +60,7 @@ def set_lang(lang)
       def check_id_is_valid(id)
         return if id.blank?
 
-        raise(ArgumentError, "Id (#{id}) cannot start with a number or contain whitespace and can only contain letters, digits, `_` and `-`") unless /^[a-zA-Z][\w:-]*$/.match?(id)
+        raise(ArgumentError, "Id (#{id}) cannot start with a number or contain whitespace and can only contain letters, digits, `_` and `-`") unless /\A[a-zA-Z][\w:-]*\z/.match?(id)
```
andysellick, line 63: I didn't realise this about newlines, thanks! Is it worth including a test in the component wrapper spec to check that a passed ID containing newlines is rejected?

sengi: Good point — I agree, seems a good candidate for a regression test.

PR #3973, opened 2024-04-08
Remove show password component

andysellick:
## What
Remove the show password component.

## Why
This component has now been [migrated upstream to the Design System](https://design-system.service.gov.uk/components/password-input/). We're not currently using it, and haven't been for some time. If we need to use it again in the future we should use that version.

## Visual Changes
None.


maxgds (approved):
Seems sensible to remove as it's currently redundant and the overhead of re-implementing can be spent within the project it's needed for in future.

PR #3976, opened 2024-04-09
Explicitly require ostruct

andysellick:
## What / why
- needed for the details component but wasn't being explicitly required, therefore causing errors when installed in applications

## Visual Changes
None.


davidgisbey (approved):
lgtm!

PR #3984, opened 2024-04-18
Add tool_name to GA4 feedback component tracking

AshGDS:
## What

- Adds `tool_name` to the GA4 feedback component tracking
- Adds tests to the component - for context this tracking was added early on in the GA4 project, so I don't think we had our spec test process set up at that time.

## Why

- https://trello.com/c/fLwaDGY3/793-feedback-form-tool-name

## Visual Changes

None.


andysellick (approved):
Nice.

PR #3988, opened 2024-04-19
Fix rendering of unnumbered papers

leenagupte:
## What
Fix rendering of unnumbered papers so that "Unnumbered..." is displayed

## Why
The "false" check in the unnumbered command and act papers was failing because the component is passed an empty string rather than nil for the number, which does not evaluate to false:

e.g.

```bash
a = false
!a
=> true
b = nil
!b
=> true
c = ""
!c
=> false
```

Therefore the check was returning `nil` rather than the "Unnumbered..." text.

### Before
Screenshot 2024-04-19 at 16 38 17

### After
Screenshot 2024-04-19 at 16 38 04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Fix rendering of unnumbered papers ([PR #3988](https://github.com/alphagov/govuk_publishing_components/pull/3988))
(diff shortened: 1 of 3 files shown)


leenagupte:
@andysellick Thanks for reviewing. Will squash the commits.

> Also it looks like the 'before' screenshot was taken with the old icon
Yeah, the new one hasn't been released yet.

PR #3991, opened 2024-04-23
Fix misaligned button and input on search component

matthillco:
## What
Make the search input field and button the same height after applying type scale changes in design system v5.

## Why
These two elements became misaligned when the input field font-size was increased as part of the design system typography improvements. We no longer need adjusted padding for the tablet+ view as the font size is now the same 19px at all screen sizes.

## Visual Changes

1) Original site with 16px fonts
2) Updated typescale with 19px font and broken search field
3) Updated typescale with 19px font and fixed search field (this PR)

image


AshGDS:
We incorporated this change into the `govuk-frontend` upgrade PR https://github.com/alphagov/govuk_publishing_components/pull/4160 and put you as the commit author, so you still get the credit. I'll close this PR now - thanks

PR #3992, opened 2024-04-24
Update `module.js` ready for v5 of govuk-frontend

MartinJJones:
## What

- Update the start function in `module.js` to support how JS modules are initialisted in govuk-frontend v5 as initialisation now happens automatically
  - see "Remove calls to component init methods from your JavaScript" in the release notes of [GOV.UK Frontend v5.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0)
- Add tests to `modules.spec.js` to check modules are initialised as expected when the `init` method is not on the modules prototype

## Why

[Trello card](https://trello.com/c/Vm74rv2e/938-prep-for-51-update-modulejs-in-govuk-publishing-components-to-support-multiple-initialisation-approaches-m)

These changes are required to support v5 of govuk-frontend, by making these changes in advance of the release of v5 of govuk-frontend we can help avoid any potentially breaking changes when migrating to the new version of the gem that upgrades govuk-frontend to v5.

The main challenge here is when using static and frontend rendering applications. Modules are found and started in [dependencies.js](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/dependencies.js) by including ./modules.js, this is called by [static](https://github.com/alphagov/static/blob/main/app/assets/javascripts/application.js#L1) once for all applications on GOV.UK. If we were to include the changes made to `modules.js` as part of the upgrade to v5, this could cause some breaking changes in certain scenarios, further examples below.

### Accordion Component

**Scenario**
- `collections` is running v5 of govuk-frontend
- `static` is running an old version of the gem that does not support both initialisation approaches in `modules.js`

**Outcome**
As the `init` method is removed from the accordion component in v5 of govuk-frontend, the accordion component from govuk-frontend will not be initialised in `modules.js`

**Impact**
The content will not be accessible as shown in the screenshots below. This is a known issue and further details are included in https://github.com/alphagov/govuk-frontend/issues/999

| initialised | not initialised |
| --- | --- |
| accordion-working | accordion-component-not-working |

### Tabs Component

**Scenario**
- `frontend` is running v5 of govuk-frontend
- `static` is running an old version of the gem that does not support both initialisation approaches in `modules.js`
- the init method is removed from the tabs component in v5 of govuk-frontend

**Outcome**
As the `init` method is removed from the tabs component in v5 of govuk-frontend, the tabs component from govuk-frontend will not be initialised in `modules.js`

**Impact**
- tabs are still styled as expected
- tab content is accessible
- all tab content will appear on the page
- clicking a tab will scroll a user to the related section of the page

| initialised | not initialised |
| --- | --- |
| tabs-component-working | tabs-component-not-initialised |

## Further information

- GOV.UK Design System Architecture decision record - [Remove init() method from JavaScript components](https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/010-remove-init-method.md)
- [Remove component init() methods and initialise in constructor](https://github.com/alphagov/govuk-frontend/pull/4011)
- [govuk publishing components - JavaScript Modules](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/javascript-modules.md)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/modules.js
+++ b/app/assets/javascripts/govuk_publishing_components/modules.js
@@ -36,15 +36,20 @@
           var moduleName = camelCaseAndCapitalise(moduleNames[j])
           var started = element.getAttribute('data-' + moduleNames[j] + '-module-started')
           if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
-            // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend Modules
-            if (GOVUK.Modules[moduleName].prototype.init) {
-              try {
+            try {
+              if (GOVUK.Modules[moduleName].prototype.init) {
+                // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend V4 Modules
                 new GOVUK.Modules[moduleName](element).init()
-                element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
-              } catch (e) {
-                // if there's a problem with the module, catch the error to allow other modules to start
-                console.error('Error starting ' + moduleName + ' component JS: ', e, window.location)
+              } else {
+                // GOV.UK Frontend V5 Modules - removed component init() methods and initialise in constructor
+                // https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/010-remove-init-method.md
+                /* eslint-disable no-new */
+                new GOVUK.Modules[moduleName](element)
               }
+              element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
+            } catch (e) {
+              // if there's a problem with the module, catch the error to allow other modules to start
+              console.error('Error starting ' + moduleName + ' component JS: ', e, window.location)
             }
           }
         }
--- a/spec/javascripts/govuk_publishing_components/modules.spec.js
+++ b/spec/javascripts/govuk_publishing_components/modules.spec.js
@@ -74,10 +74,12 @@ describe('GOVUK Modules', function () {
   describe('when modules exist', function () {
     var container
     var callbackFrontendModule
+    var callbackFrontendModuleVersionFive
     var callbackGemFrontendModule
 
     beforeEach(function () {
       callbackFrontendModule = jasmine.createSpy()
+      callbackFrontendModuleVersionFive = jasmine.createSpy()
       callbackGemFrontendModule = jasmine.createSpy()
 
       // GOV.UK Frontend Modules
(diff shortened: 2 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/modules.js
```diff
@@ -36,7 +36,7 @@
           var moduleName = camelCaseAndCapitalise(moduleNames[j])
           var started = element.getAttribute('data-' + moduleNames[j] + '-module-started')
           if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
```
andysellick, line 38: Suggestion: seems like there's some repetition here, I think it could be simplified slightly.

```Javascript
if (typeof GOVUK.Modules[moduleName] === 'function' && !started) {
  try {    
      // Vanilla JavaScript GOV.UK Modules and GOV.UK Frontend V4 Modules
      new GOVUK.Modules[moduleName](element).init()
    } else {
      // GOV.UK Frontend V5 Modules - removed component init() methods and initialise in constructor
      // https://github.com/alphagov/govuk-design-system-architecture/blob/main/decision-records/010-remove-init-method.md
      /* eslint-disable no-new */
      new GOVUK.Modules[moduleName](element)
    }
    element.setAttribute('data-' + moduleNames[j] + '-module-started', true)
  } catch (e) {
    console.error('Error starting ' + moduleName + ' component JS: ', e, window.location)
  }
}
```

spec/javascripts/govuk_publishing_components/modules.spec.js
```diff
@@ -154,12 +167,19 @@ describe('GOVUK Modules', function () {
     })
 
     it('does not start modules that are already started', function () {
-      var module = $('<div data-module="test-alert-frontend-module"></div>')
-      container.append(module)
+      var modules = $(
+        '<div data-module="test-alert-frontend-module"></div>' +
+        '<div data-module="test-alert-frontend-module-version-five"></div>' +
+        '<div data-module="gem-test-alert-frontend-module"></div>'
```
andysellick, line 173: This test looks to have changed from one test case to three. Just interested to know why this third one is in here - does the test need all three?

MartinJJones: I don't think the test needs all 3 to be honest, this could be reduced to testing 2 modules, those that have the `init` method and those that do not.

The reason for 3 different modules at the moment is to test, govuk-frontend frontend modules V4 and V5, and then modules from the gem.

Happy to change this and reduce down to 2 if that makes more sense?

app/assets/javascripts/govuk_publishing_components/modules.js
MartinJJones: Thanks this looks much cleaner, I've made this change 👍

spec/javascripts/govuk_publishing_components/modules.spec.js
andysellick: That explanation makes sense, thanks. Happy for you to leave as-is.

andysellick (approved):
Nice work. Personally I'd merge the two commits together as they relate directly to each other.

PR #3993, opened 2024-04-24
Update layout_for_admin component to take custom asset filenames 

davidgisbey:
## What

Allow custom JS and CSS filenames to be passed to the layout_for_admin component.

## Why

We're adding an admin interface to our application and want to use retain application.js & application.scss for the public interface.

This will allow us to pass admin.js and admin.scss in as args so we can retain only loading the required components for each so the page load time is fast.

## Visual Changes

N/A

## Testing

It was a bit awkward to test this as passing it a filename that doesn't exist causes exceptions. Using any_instance_of is a bit of a smell, but it's probs okay(ish) here to avoid additional complexity.

## Trello card

https://trello.com/c/r6uafy2D/1354-basic-admin-area

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+- Update layout_for_admin component to take custom asset filenames ([PR #3993](https://github.com/alphagov/govuk_publishing_components/pull/3993))
+
 ## 38.1.0
 
 * Fix rendering of unnumbered papers ([PR #3988](https://github.com/alphagov/govuk_publishing_components/pull/3988))
--- a/app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
+++ b/app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
@@ -23,3 +23,21 @@ examples:
       browser_title: 'A page title'
       block: |
         <!-- You probably want to use the header, main & footer components here -->
+  with_custom_js_filename:
+    description: An alternative JS filename can be used in place of the default `application.js` if required (note that this cannot easily be demonstrated here).
+    data:
+      environment: production
+      product_name: Publishing
+      browser_title: 'A page title'
+      js_filename: "application"
+      block: |
+        <!-- You probably want to use the header, main & footer components here -->
+  with_custom_css_filename:
+    description: An alternative JS filename can be used in place of the default `application.scss` if required (note that this cannot easily be demonstrated here).
+    data:
+      environment: production
+      product_name: Publishing
+      browser_title: 'A page title'
+      css_filename: "application"
+      block: |
+        <!-- You probably want to use the header, main & footer components here -->
--- a/spec/components/layout_for_admin_spec.rb
+++ b/spec/components/layout_for_admin_spec.rb
@@ -16,4 +16,25 @@ describe "Layout for admin", type: :view do
 
     assert_select 'meta[name="robots"][content="noindex,nofollow,noimageindex"]', visible: false
   end
+
+  it "can receive a custom js filename" do
+    allow_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("govuk_publishing_components/vendor/modernizr")
+    expect_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("admin").once
+
+    render_component(
+      browser_title: "Hello, admin page",
+      environment: "production",
+      js_filename: "admin",
+    )
+  end
+
+  it "can receive a custom css filename" do
+    expect_any_instance_of(ActionView::Base).to receive(:stylesheet_link_tag).with("admin", media: "all")
+
+    render_component(
+      browser_title: "Hello, admin page",
+      environment: "production",
+      css_filename: "admin",
+    )
+  end
 end
--- a/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
@@ -1,5 +1,7 @@
 <%
   add_gem_component_stylesheet("layout-for-admin")
+  js_filename ||= "application"
+  css_filename ||= "application"
 
   product_name ||= "Publishing"
 %>
@@ -13,7 +15,7 @@
     <meta name="viewport" content="width=device-width, initial-scale=1">
     <%= csrf_meta_tags %>
     <%= favicon_link_tag "govuk_publishing_components/favicon-#{environment}.png" %>
-    <%= stylesheet_link_tag "application", media: "all" %>
+    <%= stylesheet_link_tag css_filename, media: "all" %>
     <%= javascript_include_tag "govuk_publishing_components/vendor/modernizr" %>
     <%= yield :head %>
   </head>
@@ -22,6 +24,6 @@
       document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' govuk-frontend-supported' : '');
     <% end -%>
     <%= yield %>
-    <%= javascript_include_tag "application" %>
+    <%= javascript_include_tag js_filename %>
   </body>
 </html>


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
```
davidgisbey, line 10: I've added this to unreleased for now, but we might want to just bump the version to unblock us. Wdyt @kevindew?

andysellick: For preference we'd follow the normal release process: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/publishing-to-rubygems.md

kevindew: Yup agree on normal release process, can be part of next batch release

spec/components/layout_for_admin_spec.rb
```diff
@@ -16,4 +16,25 @@ def component_name
 
     assert_select 'meta[name="robots"][content="noindex,nofollow,noimageindex"]', visible: false
   end
+
+  it "can receive a custom js filename" do
+    expect_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("govuk_publishing_components/vendor/modernizr").once
```
kevindew, line 21: Do we need to do some mocking? i.e. does it error without it?

If we do need it, I'd suggest we use an allow here, as this isn't actually an assertion we're making about js_filename, if not I'd just delete it.

(Ideally it won't be too long until modernizr can be removed - I think it's only for IE < 11)

kevindew (commented):
Great stuff David 👍

spec/components/layout_for_admin_spec.rb
davidgisbey: Yes unfortunately it blows up without t

kevindew: Sure, can we change it to allow to indicate it's not one of our expectations?

davidgisbey: oh right sorry i get what you mean 👍 sorry i thought you meant both

kevindew (approved):
Great, all the changes look good to me. Only thing I remembered to suggest is that we should add some information in the commit message about how we ended up committing the sin of using `(allow|expect)_any_instance_of` and how we ended up stubbing ActiveView::Base - as there is quite opaque.

It'd be good to get a nod from @andysellick or one of the frontend devs on our team before merging though

davidgisbey:
Yea great shout. Sorry i was using CI to run my tests as i was having issues locally and forgot to update the commit message. I'll sort that now 👍

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
...
+
+- Update layout_for_admin component to take custom asset filenames ([PR #3993](https://github.com/alphagov/govuk_publishing_components/pull/3993))
```
andysellick, line 12: Nitpick: can this match the PR name please?

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
```diff
@@ -23,3 +23,19 @@ examples:
       browser_title: 'A page title'
       block: |
         <!-- You probably want to use the header, main & footer components here -->
+  with_custom_js_filename:
+    data:
+      environment: production
+      product_name: Publishing
+      browser_title: 'A page title'
+      js_filename: "application"
```
andysellick, line 31: These examples pass the default value, would it be clearer to show this functionality in action?

spec/components/layout_for_admin_spec.rb
```diff
@@ -16,4 +16,25 @@ def component_name
...
+
+    render_component(
+      browser_title: "Hello, admin page",
+      environment: "production",
+      js_filename: "admin",
+    )
+  end
+
+  it "can receive a custom css filename" do
+    expect_any_instance_of(ActionView::Base).to receive(:stylesheet_link_tag).with("admin", media: "all")
```
andysellick, line 32: Interested to know how these tests work - I was expecting an `assert` or similar, how do they work?

spec/components/layout_for_admin_spec.rb
```diff
@@ -16,4 +16,25 @@ def component_name
 
     assert_select 'meta[name="robots"][content="noindex,nofollow,noimageindex"]', visible: false
   end
+
+  it "can receive a custom js filename" do
+    allow_any_instance_of(ActionView::Base).to receive(:javascript_include_tag)
```
andysellick, line 21: Again just interested to know how this works - why does the JS test have `allow_any_instance_of` and `expect_any_instance_of`, but the CSS test only has `expect_any_instance_of`?

andysellick (commented):
Change looks good, just a few comments/questions 👍

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
davidgisbey: The problem with that is that i'd have to add a blank `admin` stylesheet, because it will actually try to compile the file. So it if update it to use "admin" it'll look for "admin.js" and blow up if it can't find it.

I figured this was the lesser of 2 evils, but let me know if you'd prefer another approach 👍

andysellick: Oh I see that makes sense. Tricky! In that case maybe add a description to these examples to clarify? Something like:

```
description: An alternative JS filename can be used in place of the default `application.js` if required (note that this cannot easily be demonstrated here).
```

spec/components/layout_for_admin_spec.rb
davidgisbey: Similar to above, the reason we're unable to make an assertion on the page is that it will blow up while trying to render the component if it's unable to find the file based on filename passed into the component. 

Due to that we have to stub out the method that handles the compilation of the assets. It feels like in this situation the thing we really care about is that the correct file is being included.

How allow and expect is that they stub out the method for the class passed in and return "" (by default).
So essentially what happens here is that the layout makes the call to compile `admin.js` and we intercept that call and return an empty string. Blocking it from actually trying to compile and blowing up.

The reason that we need `allow_any_instance_of` in the JS test is due to there being a second method call [here](https://github.com/alphagov/govuk_publishing_components/pull/3993/files#diff-1ebac9542b105f130b13ae04925c85d875d4dc231391387ecb4206b359fc0c07R19) 

So two calls are made:

```
 <%= javascript_include_tag "govuk_publishing_components/vendor/modernizr" %>
```
then
```
 <%= javascript_include_tag "application" %>
```
Without the allow this line

```
 expect_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("admin").once
```

blows up because the `javascript_include_tag` method has been stubbed, but only once with expectation that the argument will be "admin".

The output of the test is something similar to

```
expected  javascript_include_tag with the argument of "admin", but it was actually "govuk_publishing_components/vendor/modernizr"
```

The `allow` allows the method to be called with any argument any amount of times which ensures we don't see the error above.  While the expect is checking that the call is made with "admin" once which is the desired out come

Hopefully that makes sense.

Maybe it's clearer if i update the allow to be more explicit and do 

```
allow_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("govuk_publishing_components/vendor/modernizr")
 expect_any_instance_of(ActionView::Base).to receive(:javascript_include_tag).with("admin").once
```

So it's clear that the allow is due to the method being called twice, once with a file we're not worried about with our change to the implementation.

davidgisbey: There are other approaches we could use like dynamically creating an empty `admin.js` file before action before running the test, testing what's in the file, then deleting it after running the test, but that comes with it's own setbacks. Still complex, far slower etc.

But happy to chat it through if you'd like!

I'll go back and update the commit message with some of the above.

spec/components/layout_for_admin_spec.rb
davidgisbey: I've wrote a lengthy (sorry about that) comment [here](https://github.com/alphagov/govuk_publishing_components/pull/3993#discussion_r1579178525) which hopefully addresses this.

Let me know if any of it doesn't make sense

CHANGELOG.md
davidgisbey: Have updated the PR  since i think the changelog was a little more concise

app/views/govuk_publishing_components/components/docs/layout_for_admin.yml
davidgisbey: great shout. Thanks Andy will sort that now

davidgisbey:
@andysellick i've made the requested changes. Hopefully the above explanation makes sense. Let me know if you not and/or you want any other changes 👍

spec/components/layout_for_admin_spec.rb
andysellick: Really interesting stuff, thanks for taking the time to explain 👍

davidgisbey:
Have tested it's working as expected on the component guide so i'll merge if there's no objections

PR #3994, opened 2024-04-24
Limit GA4 view_item_list arrays to 15,000 UTF-16 code units

AshGDS:
## What
- On the following pages, GA4 is throwing a 413 HTTP error, as we're sending too much data:
    - https://www.gov.uk/world/organisations
    - https://www.gov.uk/government/groups
    - https://www.gov.uk/government/organisations/hm-revenue-customs/contact
- This is because the ecommerce arrays we were sending were too large (20,000 - 40,000+ bytes)
- I've done some investigation and it looks like GTM errors if a request exceeds 16,000 bytes (16KB)
- I've also discovered that each push to the dataLayer also seems to push the Pageview data with it as well. Therefore, when sending the ecommerce array we must leave some room for the pageview data as well in the request.
- Therefore I've limited the eCommerce arrays to 15,000 UTF-16 code units, which should give about 1384 bytes of room for the pageview data.
- We can't really solve this by limiting the array to a certain amount of items, because an array with 50 items could be 100,000 bytes depending on what it's storing, and an array with 1000 items in it could be 10,000 bytes. Therefore we need to limit it by it's size in some way.
- The code to limit the ecommerce array works by:
    - Turning the array into a string using `JSON.stringify` 
    - Grabbing the `.length` of this string
    - The length of this string gives us a good-enough estimation of the array size in UTF-16 code units.
    - We can then use this byte number to stop the array from exceeding 15,000 UTF-16 code units. Every time a new item wants to be added to the array, we check if adding that item will cause the array to exceed 15,000 UTF-16 code units. If adding that item will cause the array to become too large, we stop adding items to the array.

## Why

https://trello.com/c/i3XBl4Hh/815-fix-viewitemlist-events-on-contact-hm-revenue-customs-page

## Testing
- I have tested this is working by taking the new `dataLayer` object that is output locally on the offending pages, and pushing that object to the `dataLayer` via our homepage on `integration`. There is no error from GTM about the payload being too large.

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
@@ -629,36 +629,31 @@ describe('GA4 core', function () {
         expect(builtEcommerceObject).toEqual(expectedEcommerceObject)
       })
 
-      it('the ecommerce items array is limited to 200 items', function () {
+      it('the ecommerce items array is limited to a maximum of 15,000 UTF-16 code units', function () {
         var innerHTML = ''
         var ecommerceItems = []
 
-        // Adds 500 search results to the DOM, but only adds 200 to our expected ecommerce object array
-        for (var i = 0; i < 500; i++) {
+        for (var i = 0; i < 4000; i++) {
           innerHTML = innerHTML + '<a data-ga4-ecommerce-path="https://www.gov.uk/the-warm-home-discount-scheme" href="https://www.gov.uk/the-warm-home-discount-scheme" data-ga4-ecommerce-index="' + (i + 1) + '">Check if you’re eligible for the Warm Home Discount scheme</a>'
 
-          if (i < 200) {
-            ecommerceItems.push({
-              item_id: 'https://www.gov.uk/the-warm-home-discount-scheme',
-              item_content_id: undefined,
-              item_list_name: 'Smart answer results',
-              index: i + 1
-            })
-          }
+          ecommerceItems.push({
+            a: 'b'
+          })
         }
 
-        resultsCount.innerHTML = '500 results'
         results.innerHTML = innerHTML
-        expectedEcommerceObject.search_results.results = 500
-        expectedEcommerceObject.search_results.ecommerce.items = ecommerceItems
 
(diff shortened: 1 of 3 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
```diff
@@ -629,36 +629,34 @@ describe('GA4 core', function () {
...
-            })
-          }
+          ecommerceItems.push({
+            a: 'b'
+          })
         }
 
-        resultsCount.innerHTML = '500 results'
+        resultsCount.innerHTML = '4000 results'
         results.innerHTML = innerHTML
```
andysellick, line 645: Is `results` used anywhere? Wondering if line 637's HTML chunk is needed now?

andysellick (approved):
Nice solution. Just one comment about some data in the test that might not be needed 👍

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
AshGDS: @andysellick I've deleted line 644, but I can't remove the HTML chunk as the eCommerce tracking needs the HTML to exist, as that's what it uses to populate its GA4 data

AshGDS:
@andysellick For correctness I've done away with the term 'bytes' in the code/comments and replaced it with 'UTF-16 code units' as that's more correct than saying bytes, as UTF-16 characters are variable length byte wise (i.e. emojis would take up more bytes than an alphanumeric character) - will merge soon unless any issues 👍

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.spec.js
andysellick: Ah okay no worries. Just couldn't see where it was used and wanted to check.

AshGDS:
@andysellick I've also removed the regex just now as I think it was redundant (since GA4 would send the array as a JSON anyway so the array would be stringified, so I don't think there's a need to remove the stringified characters from the size calculation)

PR #3995, opened 2024-04-26
Add favicon for test environment

davidgisbey:
## What

Duplicated the development favicon and named it favicon-test.png to unblock our system specs.

## Why

When we run our system specs that use the layout_for_admin component they fail due to this line of code

```
<%= favicon_link_tag "govuk_publishing_components/favicon-#{environment}.png" %>
```

This is because the `favicon-test.png` file does not exist. There are favicons for:

- prod
- staging
- integration
- example (component guide)
- development

I've duplicated the development favicon and named it favicon-test.png to unblock our system specs.

## Visual Changes

image

## Unsure why it's working for other apps

I'm trying to figure out why tests aren't failing for other apps that use the component (Whitehall, Content Publisher etc.)

Perhaps i can update some config locally and avoid adding this 🤔


davidgisbey:
Aha i've tracked it down. They use `GovukPublishingComponents::AppHelpers::Environment.current_acceptance_environment` in the layout which passes development through as the env for tests 👍

PR #3996, opened 2024-04-26
WIP drop JS support for IE

andysellick:
## What
Work in progress spike switching over to using `type="module"` for scripts, so that IE11 and below don't run any of our JavaScript. Needs some accompanying changes in both `static` and the application being rendered. Currently getting weird JS errors like `application.js:367 Uncaught TypeError: Cannot read properties of undefined (reading 'GOVUKFrontend')`, which looks more like a bundling error than an actual problem.

## Why
https://github.com/alphagov/govuk-rfcs/pull/171

## Visual Changes

@AshGDS


AshGDS:
@andysellick I think the issues you were having were due to using an old version of govuk-frontend, and I hackily got the JS to work now by upgrading to V5, but at the expense of the styles breaking. Therefore it might be worth basing the spike off the work already done in https://github.com/alphagov/govuk_publishing_components/pull/3700

AshGDS:
Superseeded by https://github.com/alphagov/govuk_publishing_components/pull/4000

PR #4000, opened 2024-04-29
WIP drop JS support for Grade X browsers

AshGDS:
## What
- switch to type="module
- remove polyfills
- remove references to polyfills
- use ES6 for loop in accordion component
- based off of https://github.com/alphagov/govuk_publishing_components/pull/3700
- Seems to be working fine, the JS does have a console error in IE11 but our pages seem to be falling back to HTML/HTTP nicely

## Why

- Spike to see how easy it is to drop IE11
- https://trello.com/c/u4MwaHgI/75-spike-into-dropping-javascript-support-for-internet-explorer

PR #4003, opened 2024-04-30
Make YoutubeLinkEnhancement URL parsing more robust.

sengi:
[URL API] `searchParams` works [pretty much everywhere](https://caniuse.com/?search=searchparams) since around 2017, which includes [all the user agents we support](https://www.gov.uk/service-manual/technology/designing-for-different-browsers-and-devices). It also considerably simplifies the code and makes the intent much clearer.

(We have to avoid static `canParse()` though, because Samsung Internet is weird.)

Resolves https://github.com/alphagov/govuk_publishing_components/security/code-scanning/1.

[URL API]: https://developer.mozilla.org/en-US/docs/Web/API/URL

No visual or functional change intended.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -32,7 +32,7 @@
         link: $link
       }
 
-      if (href.indexOf('/live_stream') >= 0) {
+      if (href.includes('/live_stream')) {
         var channelId = YoutubeLinkEnhancement.parseLivestream(href)
 
         if (!this.hasDisabledEmbed($link) && channelId) {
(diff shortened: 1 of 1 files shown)


andysellick (changes requested):
@sengi looks like a sensible change except that `includes` and (haven't confirmed this yet but fairly sure) `searchParams` aren't supported by older browsers like IE. While we don't support those browsers we do want to ensure they have a reasonable experience of GOV.UK and this change would probably cause all of the JS to break suddenly.

We're currently working on an RFC to remove JS support for IE11, can this PR wait until then? https://github.com/alphagov/govuk-rfcs/pull/171

sengi:
Ah ok, TIL we still de facto support IE11 😅

If you're sure we need to continue to support embedded YT videos in IE11 for the time being then I'm happy to stick this on the back burner until you're ready to drop IE.

I'll make it draft for now anyway.

sengi:
Actually wouldn't be just dropping support for adding/editing YT embeds in the content management tools in IE? Is this library even used in the public-facing frontends?

sengi:
Even so, I'm not sure this would affect anything public-facing because it's in the govspeak library and surely we're not doing the govspeak->html transform on the client? 🤔

andysellick:
> Even so, I'm not sure this would affect anything public-facing because it's in the govspeak library and surely we're not doing the govspeak->html transform on the client? 🤔

Correct, we're not doing the govspeak->html transform on the client, but we are running this JS on that output in the client to convert links to videos into embedded videos. See [this page](https://www.gov.uk/guidance/help-and-support-for-self-assessment) for an example - if JS is disabled the embedded videos show simply as links to Youtube.

sengi:
Aha thanks! Super helpful example ❤️

sengi:
I'm guessing this depends on #4000 — does that sound right?

andysellick (approved):

PR #4007, opened 2024-05-02
Update LUX to 314

jon-kirwan:
## What / why
Update to LUX 314 from version 313.

See https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md for details.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
+++ b/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
@@ -46,6 +46,7 @@
       auto: autoMode,
       beaconUrl: getProperty(obj, "beaconUrl", luxOrigin + "/lux/"),
       conversions: getProperty(obj, "conversions"),
+      cookieDomain: getProperty(obj, "cookieDomain"),
       customerid: getProperty(obj, "customerid"),
       errorBeaconUrl: getProperty(obj, "errorBeaconUrl", luxOrigin + "/error/"),
       jspagelabel: getProperty(obj, "jspagelabel"),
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
```diff
@@ -1492,13 +1522,36 @@
...
+      return [
+        "&INP=" + details.duration,
+        details.selector ? "&INPs=" + encodeURIComponent(details.selector) : "",
+        "&INPt=" + clamp(floor(details.startTime)),
+        "&INPi=" + clamp(floor(details.processingStart - details.startTime)),
+        "&INPp=" + clamp(floor(details.processingEnd - details.processingStart)),
+        "&INPd=" + clamp(floor(details.startTime + details.duration - details.processingEnd)),
+      ].join("");
     }
     function getCustomerId() {
```
andysellick, line 1551: I thought this function needs to be as follows, but I can't reproduce the error that it solves. Take a look at this related PR: https://github.com/alphagov/govuk_publishing_components/pull/3592

It's possible the error only happens in production, but I think this code change makes sense regardless.

```JS
function getCustomerId() {
  return LUX.customerid || "";
}
```

I probably need to clarify the documentation on this, apologies.

jon-kirwan: Thanks Andy. Good spot, and I've now reverted the `getCustomerId` function change.

andysellick (approved):
Nice work 👏

PR #4009, opened 2024-05-07
Add blue option

JamesCGDS:
## What
This PR extends the [button component](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_button.html.erb) to accept a `blue` option so it has a blue background.

## Why
The landing page for GOV.UK Chat uses a blue button and I thought the cleanest solution would be to extend the gem's button component to also accept a blue background colour as opposed to creating a duplicate application component with the only difference being the background colour - [trello card](https://trello.com/c/cf8aTngw/1407-build-landing-page).

## Visual Changes

When the `blue` option is not passed/set to `false`:

image

When the `blue` option is passed and set to `true`:

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_button.scss
@@ -8,6 +8,14 @@
   margin-bottom: 0;
 }
 
+.gem-c-button--blue {
+  background-color: govuk-colour("blue");
+
+  &:hover {
+    background-color: govuk-shade(govuk-colour("blue"), 20%);
+  }
+}
+
 // this will be moved and extended into a model for general component spacing
 // once this has been decided upon and other work completed, see:
 // https://trello.com/c/KEkNsxG3/142-3-implement-customisable-spacing-for-components
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -52,6 +52,10 @@ examples:
       href: '#'
       start: true
       rel: external
+  blue_button:
+    data:
+      text: Blue button
+      blue: true
   secondary_button:
     data:
       text: Secondary button
--- a/lib/govuk_publishing_components/presenters/button_helper.rb
+++ b/lib/govuk_publishing_components/presenters/button_helper.rb
@@ -16,6 +16,7 @@ module GovukPublishingComponents
                   :target,
                   :type,
                   :start,
+                  :blue,
                   :secondary,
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_button.scss
```diff
@@ -8,6 +8,14 @@
   margin-bottom: 0;
 }
 
+.gem-c-button--blue {
+  background-color: govuk-colour("blue");
```
andysellick, line 12: This code looks fine, but I'm concerned that this is a deviation from the Design System - this blue is already in use for other elements. What's the need for blue buttons and how this has been tested with users?

JamesCGDS: Thanks for taking a look at this Andy.

We haven’t conducted user testing for the blue buttons specifically, it’s been more testing the user journey as a whole. User feedback has been positive overall but I understand your concern - maybe creating a custom component in the application would be more appropriate (although I understand that would still mean deviance from the design system, just in a different place).

I’ve raised these questions with the design folks on the team and can update you when I get a response.

JamesCGDS: Hi Andy, this is the response I've received from the design folks on the team:

> The rationale for using blue button:
> - Buttons in Chat are with the user through the whole journey (e.g. ‘Start chatting’ ‘I understand’ ‘Send’) they do not take the user to another page/initiate a process which is the typical behaviour for the green button
> - Blue aligns with the blue search button
> - Blue also works to align the Chat brand with the [GOV.UK](http://gov.uk/) brand

Happy to discuss further

andysellick: Sorry, going to dig my heels in on this one until someone from the Design System comments.

The first rationale sounds like 'the button is blue because all the buttons we want to use are blue'. And all of those button text examples clearly either take the user to another page or initiate a process.

The blue search button is a special case in the header (which also deviates from the Design System) and isn't used in any page content.

I'm not sure how having blue buttons aligns the Chat brand with the GOV.UK brand - blue isn't generally used for buttons.

I'm not against new design ideas but I know that the Design System is based on years of research and testing and I think if we're going to start doing something different from them we should have a very clear idea of why.

JamesCGDS: No need to apologise, thanks for your input Andy. I think I'll close this PR following the additional feedback in the community meeting earlier.

PR #4011, opened 2024-05-09
Update Twitter logo

andysellick:
## What
Update the old Twitter logo in the share links component to match the new Twitter logo, which is an X.

## Why
The Twitter logo recently changed from a bird to an X. In order to keep up with Twitter, we should update the Twitter logo. Twitter logo sourced from Twitter's own brand pages.

Note that this change will update all of the Twitter logos on GOV.UK where this component is being used to to link to Twitter pages or allowing users to follow organisations on Twitter.

## Visual Changes
This is an example of the new Twitter logo in the share links component.

Screenshot 2024-05-09 at 15 21 51

Trello card: https://trello.com/c/Qb9h2prz/109-update-twitter-icon


andysellick:
@maxgds the technical work for this is pretty small, as you can see. There's a wider question about how we refer to Twitter (Twitter/X/the organisation formerly known as Twitter?) but that's text that's passed to the component rather than hard coded inside it, so I suggest we get this change out and look at the text separately. Thoughts?

maxgds:
Seems like a sensible approach, but please get a design sign off before putting it out. Suggest @JeremyHHY is probably best placed for that.

JeremyHHY:
@andysellick hey up, this looks good to me. And the approach regarding what to call it makes sense as a separate thing for now. 

I would suggest that we should put all social media icons into a circle so that they become consistent visually as currently could get a little visually misaligned when stacked in a vertical list – though we can review this after this work is done and see how it turns out.

Consider this design signed off. :)

AshGDS (approved):

PR #4019, opened 2024-05-13
WIP - Switch from standardx to eslint with standard config

richardTowers:
[standardx](https://github.com/standard/standardx) appears to be a bit... unmaintained. It's pinning us to a pretty old version of eslint, and a few other dependencies.

We switched to using eslint in Whitehall in https://github.com/alphagov/whitehall/pull/8178

We should consider rolling this out everywhere.

TODO - more details on what dependencies we're stuck on, and the horrible story of the eslint config file format :see_no_evil:

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -1,0 +1,23 @@
+module.exports = {
+  extends: 'standard',
+  rules: {
+    'no-var': 0,
+    'object-shorthand': 0
+  },
+  ignorePatterns: [
+    'app/assets/javascripts/component_guide/vendor/**/*.js',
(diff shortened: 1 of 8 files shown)


kevindew:
It might be worth a chat with the design system about this idea as we've traditionally tried to align CSS and JS linting across GDS with Design System tending to be the leaders in the decision. We've got it documented in the GDS way to use standard and only switched to standardx for GOV.UK's legacy approach to JS.

I do agree that standard seems to not really be a moving project anymore. It'd be cool if we could replace it with something that played nicely with prettier too (I think eslint does) given the CSS linter (stylelint) expects you to have that

PR #4020, opened 2024-05-13
Add a linting rule to enforce 'use strict'

richardTowers:
This is a draft PR just to show where we're not using strict mode


andysellick:
@richardTowers is this PR still needed? I think as we've now deployed the switch to `type="module"` this work is done?

richardTowers:
Yep, no need for this now - it was just to identify which functions weren't already using strict mode, and they all will be now we're on type="module" 🎉

PR #4022, opened 2024-05-14
Prevent error if input field is missing

lauraghiorghisor-tw:
## What
Prevent errors if input field is missing from the feedback form, by adding a null check. 
This is currently causing smokey to [fail](https://argo.eks.integration.govuk.digital/applications/cluster-services/smokey?node=%2FPod%2Fapps%2Fsmokey-28595090-rh8qb&tab=logs&orphaned=false&resource=). There will be work in this area around cleaning up the feedback, but this is just to quickly fix smokey.

## Why
Due to recent changes that removed the email address from the survey form, there is no input element available to focus on when the feedback form loads.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/feedback.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/feedback.js
@@ -161,8 +161,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.prompt.hidden ? this.prompt.hidden = false : this.prompt.hidden = true
 
     if (!this.activeForm.hidden) {
-      this.activeForm.querySelectorAll('.gem-c-textarea .govuk-textarea, .gem-c-input.govuk-input')[0]
-        .focus()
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/components/feedback.js
```diff
@@ -161,8 +161,10 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.prompt.hidden ? this.prompt.hidden = false : this.prompt.hidden = true
 
     if (!this.activeForm.hidden) {
-      this.activeForm.querySelectorAll('.gem-c-textarea .govuk-textarea, .gem-c-input.govuk-input')[0]
-        .focus()
+      var input = this.activeForm.querySelectorAll('.gem-c-textarea .govuk-textarea, .gem-c-input.govuk-input')
+      if (input) {
```
richardTowers, line 165: Ah, you'll need to check the length of the result, because `querySelectorAll(.whargleblargle)` returns `[]`, and `if([]) console.log('truthy')` logs "truthy".

```suggestion
      if (input.length > 0) {
```

richardTowers (approved):

PR #4041, opened 2024-05-31
Upgrade to govuk frontend 5.1

MartinJJones:
## What

Make changes in govuk_publishing_components to allow us to update to govuk-frontend v5.1.0.

[Trello](https://trello.com/c/6LsKqHIO/2231-update-version-of-govuk-frontend-in-govukpublishingcomponents-l)

### Asset Changes

- Update references to assets (config, Javascript, CSS, fonts, images) in govuk-frontend
- Remove references to the crown fallback image
- Remove references to 
  - `assets/images/govuk-apple-touch-icon.png`
  - `assets/images/govuk-apple-touch-icon-152x152.png`
  - `assets/images/govuk-apple-touch-icon-167x167.png`
  - `assets/images/govuk-apple-touch-icon-180x180.png`
  - `assets/images/govuk-mask-icon.svg`
- Add references to
  - `assets/images/govuk-icon-180.png`
  - `assets/images/govuk-icon-mask.svg`
- Change logo in-line SVG to be the new SVG that includes "GOV.UK"
  - Rework how the logo is included in `layout_super_navigation_header`

### Javascript Changes

- Remove import of `details.js` from govuk-frontend as the `details` component no longer uses Javascript
- Updated `option-select.js` to ensure checkboxes render as intended when focused

### Stylesheet Changes

- Remove `$legacy` from the `govuk-colour` mixin (this has been removed from govuk-frontend)
- Remove `$govuk-new-link-styles` feature flag (new link styles now default, feature flag removed)

### Component specific changes

- Components using `tag` component now `titleize` the tag text to adopt the new tag design
   * Update the `header_logo` and `phase_banner` components
   * Update tests of `layout_header` and `phase_banner`
   * Tweak text colour of `gem-c-environment-tag` to ensure contrast (default tag colour is now black).
- `warning_text` component no longer uses the `govuk-warning-text__assistive` class
  * Replaced all instances of `govuk-warning-text__assistive` with `govuk-visually-hidden` as specified in the release notes
  * Ensure font-size is always 19px for warning-text even when used as a heading
  * Update tests of warning_text component
- Remove aria-labelledby from accordion sections
  * This is a recommend change in v5.1.0 of govuk-frontend - https://github.com/alphagov/govuk-frontend/releases/tag/v5.1.0

### Config changes

- Add new configuration option `use_es6_components`
  * If true then `layout_for_public` will try to load the ES6 (govuk-frontend) using components from the file `es6-components.js`

### Update the "Install and use this gem" docs to include further JS guidance
- Added a new section for "[Create a separate JS bundle for govuk-frontend components](https://github.com/alphagov/govuk_publishing_components/pull/4041/commits/ec39705c4d310e02c48e42447c085e5080b3d0af#diff-c686b22631f5561b405713d3b98226ef60313bfd6a73dc390f95cd8e3eae4298)" that provides guidance on how to include JS in Rails applications in light of the browser target change that has occurred in govuk-frontend v5.

### Polyfills copied from GOV.UK Frontend v4
In GOV.UK Frontend v5 polyfills have been removed, see “Verify your code does not rely on polyfills we have now removed” in the [release notes for 5.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0).

The polyfills originally provided by GOV.UK Frontend v4 have been [copied to the publishing-components gem](https://github.com/alphagov/govuk_publishing_components/pull/4041/commits/db1ebfc299a44cde196149a23b2b16418cc99631). This will allow you to still use the existing polyfills where required.

## What I Haven't Changed

This PR is for things that **have** to change. To keep the PR size as manageable as possible (sorry, it is still very long) there are pieces of work that I think should be separate PRs:

### Does not use the new manifest.json file provided in v5 of govuk-frontend

This PR does not implement the [new manifest.json file provided in v5 of govuk-frontend](https://github.com/alphagov/govuk-frontend/pull/4445). The main reason for this is that the icon names included in the manifest.json are hardcoded and do not have the correct file name once [fingerprinted](https://guides.rubyonrails.org/asset_pipeline.html#what-is-fingerprinting-and-why-should-i-care-questionmark).

Taking `favicon.svg` as an example, the filename for this in the assets folder would be `favicon-[SHA].svg`, but the reference in manifest.json would still be `favicon.svg`

This would result in the error below:

> Error while trying to use the following icon from the Manifest: /favicon.svg

There is a GitHub issue logged for this in govuk-frontend - https://github.com/alphagov/govuk-frontend/issues/5024

We would likely need a custom solution to be able to create our own manifest.json file in the future.

In this PR we mostly keep things the same as before for our icons, and implement the changes below:

- Add the new `favicon.svg`icon
- Update filenames where required
- Does not implement manifest.json
- Does not implement the 2 new icons new to v5, `icongovuk-icon-192.png` and `icongovuk-icon-512.png`, further details below:

Name | Type | Dimensions | Notes
-- | -- | -- | --
Android icongovuk-icon-192.png | PNG | 192×192 | Primarily used on the Android homescreen. This is new for v5.
Large icongovuk-icon-512.png | PNG | 512×512 | Fallback for any applications that request a larger size but cannot use SVG. May also be used by Safari 17's "Add to dock" feature on macOS. This is new for v5.

### Component guide does not support Grade X browsers
Opening the component guide in browsers that don't support `type="module"` will result in some ungraceful erroring. This is because in this PR there isn't a way to individually load the JS that contains ES6 code. In frontend rendering applications I have established a pattern to follow to do this and it could be implemented in them gem also (but in a separate PR)

## Why

There are several breaking changes in the latest version of gouk-frontend which requires us to update govuk_publishing_components. Please refer to the full release notes for [v5.0.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0) and [5.1.0](https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0). The changes that are specifically being addressed here are:

- Update package file paths
- Remove Internet Explorer 8 stylesheets, settings and mixins
- Remove the fallback GOV.UK crown logo from your HTML
- Check your tags align with design changes to the Tag component
- New link styles are now enabled by default
- Update the HTML for warning text

## Visual Differences

Percy: https://percy.io/55ba5e1a/govuk_publishing_components/builds/34630046?utm_campaign=55ba5e1a&utm_content=govuk_publishing_components&utm_source=github_status_public

### Layout Header

The reason for the visual change here is that `.gem-c-header__logotype` was causing the line-height to increase, now this class is removed, the total height of the layout header has decreased from 72.66px to exactly 70px.

#### Before

layout-header-before
layout-header-before-sizes

#### After

layout-header-after
layout-header-after-sizes

### Public Layout

Percy is also showing a visual change in public layout wherever the layout header component is used, this is to be expected following the change mentioned above.

| Before | After |
| --- | --- |
| image |  image |

### Phase Banner

The phase banner will now use the new design included in v5 of the govuk-frontend.

| Before | After |
| --- | --- |
| 68e7957767c852045d2522b12d04913705f749b970d165415e8d84fe60f1726b |  b3460582d31935951cd58afacd27118567d2a311fabd59ba12e8f2a5ab5a430d |

### Form inputs

The visual changes shown in Percy for forms inputs are expected and come from changes made in v5.1.0 of govuk-frontend:

- https://github.com/alphagov/govuk-frontend/pull/4157
- https://github.com/alphagov/govuk-frontend/pull/4093

There is a slight alignment bug with the left border when conditional radio or checkboxes are used, this was logged as an issue and fixed in govuk-frontend - https://github.com/alphagov/govuk-frontend/issues/5069


app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
```diff
@@ -25,5 +25,8 @@
     <% end -%>
     <%= yield %>
     <%= javascript_include_tag js_filename %>
+    <% if GovukPublishingComponents::Config.use_es6_components %>
+      <%= javascript_include_tag 'es6-components', type: "module" %>
+    <% end %>
```
kevindew, line 30: This has a hardcoded filename compared the configurable ones. 

Would we be able to have a js_module_filename variable? 

Also, should this actually be in the head element given it's loading is deferred?

docs/install-and-use.md
```diff
@@ -93,6 +93,26 @@ If your application does use Slimmer/Static:
 //= require govuk_publishing_components/components/button
 ```
 
+As of v5.0.0, the target for browser support for govuk-frontend has changed to browsers that support using the `type="module"` attribute on `script` tags. This means that that govuk_publishing_components Javascript files that include files govuk-frontend will now error ungracefully and potentially function in an unpredictable way if a browser that doesn't support `type="module"` tries to evaluate the Javascript in a script tag without the `type="module"` attribute.
```
andysellick: Duplicated word: `that that`

andysellick: `This means that that govuk_publishing_components Javascript files that include files govuk-frontend`

Should this be `that include files from govuk-frontend`?

docs/install-and-use.md
```diff
@@ -93,6 +93,26 @@ If your application does use Slimmer/Static:
...
+
+To ensure that browsers that are below the target don't evaluate the Javascript that they don't support, consider using a separate file called `es6-components.js` for these components:
+
+```ruby
+# es6-components.js
+//= require govuk_publishing_components/lib
+//= require govuk_publishing_components/components/accordion
```
andysellick, line 105: Is it worth including a list of components for which this is the case, or pointing to a list somewhere? (it's in the auditing tools, for example, where it'll be automatically updated).

app/views/govuk_publishing_components/components/layout_for_public/_header_logo.html.erb
```diff
@@ -0,0 +1,41 @@
+<%
+  hide_logo_text ||= false
+  large_navbar ||= false
+  logo_text ||= t("components.layout_super_navigation_header.logo_text")
+  logotype_classes = %w(govuk-header__logotype gem-c-layout-super-navigation-header__logotype-crown)
+  logotype_classes << "gem-c-layout-super-navigation-header__logotype-crown--large-navbar" if large_navbar  
+%>
+
+<% unless hide_logo_text %>
+    <svg
```
andysellick, line 10: Nitpick: is something wrong with the indenting here? Looks like it's been double indented?

app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
```diff
@@ -10,7 +14,7 @@
   margin-left: 0;
 }
 
-.gem-c-warning-text__text--large {
+.gem-c-warning-text .gem-c-warning-text__text--large {
```
andysellick, line 17: This looks like the specificity of the selector has been increased - is this for a CSS reason or just to match the rest of the file? Just checking if this is needed since these are `gem-c` classes.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.js
```diff
@@ -1,4 +1,4 @@
-//= require govuk/vendor/polyfills/Element/prototype/closest.js
+//= require ../vendor/polyfills-govuk-frontend-v4/Element/prototype/closest.js
```
andysellick, line 1: This probably doesn't matter since we'll be deleting the polyfills in the future soon anyway, but in the commit that adds polyfills there are loads of polyfills (plus some `map` files, do we need those?) but in this commit it looks like the components only use a small handful of them. What are the extra polyfills for?

docs/install-and-use.md
MartinJJones: Thanks Andy, this is now fixed

MartinJJones: Good spot, this is now fixed 👍

app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
MartinJJones: Yes, the main reason for this is that the font-size is now set for `.gem-c-warning-text .govuk-warning-text__text` - https://github.com/alphagov/govuk_publishing_components/pull/4041/files#diff-4d9503e344b5466e77c03df22484e4e39d15eb4144003d9a8ed763eb6561d49bR5-R8, so I've increased the specificity here to ensure is has a font-size of 24px instead of 19px.

andysellick: Great, thanks for the explanation 👍

app/views/govuk_publishing_components/components/layout_for_public/_header_logo.html.erb
MartinJJones: Good spot, I've fixed this 👍

docs/install-and-use.md
MartinJJones: Yes good point, I've updated the docs to reference the auditing tools:

> **Note:** You can get an up-to-date list of components that use govuk-frontend JS using the [component auditing tools](./auditing.md), components that use JS from govuk-frontend will be marked as "Yes" under the "GF JS" column.

andysellick (approved):
Have done a second review and happy to approve. Good work.

I note there's an outstanding comment from @kevindew that needs a followup before merging.

app/views/govuk_publishing_components/components/_layout_for_admin.html.erb
MartinJJones: Thanks for the feedback Kevin, I've made the changes as suggested.

I've moved the script to the the head section and added the `js_module_filename` variable in the admin layout template.

PR #4050, opened 2024-06-05
Allow Date in YAML.load_file

andysellick:
## What
Modify the call to `YAML.load_file` in the component guide code to include `Date` in the list of permitted classes.

## Why
Turns out that the component guide in `frontend` uses `Date`, and without this change completely fails to load (`Tried to load unspecified class: Date`). Looks like this is the only application where `Date` is needed in a component (I think it's related to the calendar component in `frontend`).

Have tested this change locally in a locally running `frontend` and it fixes the problem.

Thanks to @KludgeKML for the fix!

## Visual Changes
None.

Trello card: https://trello.com/c/N3eQJZlU/168-fix-application-component-tests

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


KludgeKML (approved):
LGTM

PR #4058, opened 2024-06-12
Add govuk-frontend checking to the component auditing

andysellick:
## What
Expand the component auditing to check components for the presence of `govuk-frontend` JS and CSS imports i.e. to show which components directly use assets from `govuk-frontend`.

## Why
This was a feature previously raised by @patrickpatrickpatrick in https://github.com/alphagov/govuk_publishing_components/pull/3716 as something that would be useful as we move to using `govuk-frontend` v5 and the ES6 code additions that this includes. I've changed where the information is shown and expanded to include both JS and CSS.

## Visual Changes
Now includes the two 'GF' columns.

Screenshot 2024-06-12 at 13 44 47


AshGDS (approved):
Nice, LGTM 👍

MartinJJones:
Looks good to me, think the version number may need to be updated in the change log though

PR #4061, opened 2024-06-13
Align checkboxes component more toward Design System

andysellick:
## What

- switch from lists to using divs for checkbox wrappers (to align with Design System component)
- remove custom styling for 'or' divider and use Design System style
- remove unneeded style override for hint margin
- make checkbox hint a div, not a span
- remove other custom styles to set list styles as no longer using lists

## Why
Part of some prep work towards fixing a bug in the v5 upgrade. Also we should align more with the Design System if possible, as it saves code.

## Visual Changes
Shouldn't be any, although there is a minor spacing difference that occurs underneath one instance of the component, but this is acceptable.

Trello card: https://trello.com/c/DzMiicUF/74-checkboxes-unnecessarily-in-a-list

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Align checkboxes component more toward Design System ([PR #4061](https://github.com/alphagov/govuk_publishing_components/pull/4061))
 * Add govuk-frontend checking to the component auditing ([PR #4058](https://github.com/alphagov/govuk_publishing_components/pull/4058))
 
 ## 39.0.0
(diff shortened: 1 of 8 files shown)


MartinJJones (commented):
Nice work! I only have 1 small comment which related to the hint text, this is [currently in a span element](https://github.com/alphagov/govuk_publishing_components/blob/checkboxes-ds-alignment/lib/govuk_publishing_components/presenters/checkboxes_helper.rb#L100), but the [design system use a div element](https://design-system.service.gov.uk/components/checkboxes/), I'm wondering if we should make the change here as well? I don't think it will impact anything in terms of the styling in the V5 upgrade, so perhaps not essential and could always be handled separately.

Other than that, the changes look good to me, I tested them out locally as well using finder-frontend facets and email sign up.

MartinJJones (approved):

PR #4066, opened 2024-06-18
Prevent screen readers from announcing document list child/parts items dashes

andysellick:
## What
Prevent screen readers from announcing the dash before document list child/part items, by changing the CSS pseudo element to use a border to show the dash, instead of a dash character.

## Why
Screen readers were announcing the dash character ("em-dash") which is unnecessary and distracting.

## Visual Changes
The new border is slightly crisper than the dash character and fractionally narrower and higher than the dash character, but I don't believe the differences are a problem.

Before | After
------ | ------
Screenshot 2024-06-18 at 09 52 56 | Screenshot 2024-06-18 at 09 53 02

I've used a specific width for the element relative to the padding left of the list element, so that when the text size increases the dash remains the same size (otherwise it could overlap the text of the item, as the padding is fixed). Here's a screenshot of it with an increased text size.

Before | After
------ | ------
Screenshot 2024-06-18 at 09 57 20 | Screenshot 2024-06-18 at 09 56 14

Trello card: https://trello.com/c/yhLOWoMp/876-dashes-before-each-document-list-item-in-mobile-tablet-view-read-by-screen-readers-m


MartinJJones (approved):
Nice work on the fix and updating the accessibility criteria, tested in VoiceOver in finder-frontend and the dash is no longer announced 👍

PR #4069, opened 2024-06-20
Add guard to accordion component init method

MartinJJones:
## What

- Add guard to accordion component init method
- Add new test and update fixture

## Why

When the accordion component from govuk-frontend is initialised, the `govuk-accordion__show-all` class is added to the button element.

It is possible that the accordion component from govuk-frontend does not get initialised, but we have no checks in place to guard against this when initialising the accordion component from the gem.

Example error when using the accordion component with V5 govuk-frontend in Safari 10.1
Screenshot 2024-06-20 at 13 57 38

This change checks if the accordion component from govuk-frontend has initialised before trying to initialise the accordion component from the publishing components gem.

A new test has been added to check it does not initialise, if the accordion from govuk-frontend has not initialised. `gem-c-accordion__show-all` was removed from the fixture as this is added by the JS from the publishing-components accordion.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add guard to accordion component init method ([PR #4069](https://github.com/alphagov/govuk_publishing_components/pull/4069))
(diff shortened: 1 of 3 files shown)


andysellick (approved):
👍

PR #4073, opened 2024-06-21
Update and/or add component print styles

matthillco:
## What
This PR updates the print styles for any public-facing components. The changes broadly fall into these categories:

- hiding/removing components that should never be printed
- improving components that already have print styles 
- creating print styles for components that don't already have them

## Why

## Visual Changes
There are too many differences to include all before/after shots here. Changes will be reviewed with a designer before deployment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix some layout_header search visual bugs ([PR #4156](https://github.com/alphagov/govuk_publishing_components/pull/4156))
 * Fix issue with using special characters in content-list ([PR #4157](https://github.com/alphagov/govuk_publishing_components/pull/4157))
 * Add 'Chat entry' component ([PR #4151](https://github.com/alphagov/govuk_publishing_components/pull/4151))
+* Update and/or add component print styles ([PR #4073](https://github.com/alphagov/govuk_publishing_components/pull/4073))
 
 ## 42.0.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment-link.scss
@@ -1,4 +1,19 @@
+@import "govuk_publishing_components/individual_component_support";
+
 .gem-c-attachment-link__abbr {
   text-decoration: none;
   cursor: help;
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-attachment-link .govuk-link {
+    color: $govuk-print-text-colour !important;
+
+    &::after {
+      font-size: inherit;
+      color: inherit;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -84,3 +84,23 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey");
   text-decoration: none;
   cursor: help;
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-attachment__details {
+    padding: 0;
+  }
+
+  .gem-c-attachment__metadata,
+  .gem-c-attachment__metadata .govuk-link::after {
+    font-size: 12pt;
+  }
+
+  .gem-c-attachment__link.govuk-link {
+    word-break: break-word;
+
+    &::after {
+      font-size: 12pt;
+      line-height: 1 !important; // stylelint-disable-line declaration-no-important
+    }
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
@@ -54,3 +54,23 @@
     }
   }
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-breadcrumbs {
+    font-size: 12pt;
+  }
+
+  .gem-c-breadcrumbs,
+  .govuk-breadcrumbs {
+    .govuk-breadcrumbs__list-item {
+      .govuk-breadcrumbs__link:link,
+      .govuk-breadcrumbs__link:visited {
+        color: $govuk-print-text-colour;
+      }
+
+      &::before {
+        border-color: $govuk-print-text-colour;
+      }
+    }
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
@@ -143,21 +143,42 @@
   }
 }
 
+// stylelint-disable declaration-no-important
 @include govuk-media-query($media-type: print) {
   .gem-c-cards__list {
     display: block;
   }
 
   .gem-c-cards__list-item {
-    padding-bottom: 0;
+    padding: 0;
   }
 
-  .gem-c-cards__sub-heading {
-    margin-top: govuk-spacing(4);
-    margin-bottom: govuk-spacing(1);
+  .gem-c-cards__list-item-wrapper {
+    padding: 5mm 0;
   }
 
-  .gem-c-cards__link::before {
-    display: none;
+  .gem-c-cards__description {
+    margin: 0;
+  }
+
+  .gem-c-cards__link {
+    &,
+    &:link,
+    &:link:visited {
+      color: $govuk-print-text-colour !important;
+    }
+
+    &::before {
+      display: none;
+    }
+
+    &::after {
+      position: static;
+      display: block;
+      margin: 1mm auto;
+      font-size: 12pt !important;
+      color: $govuk-print-text-colour;
+    }
   }
 }
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -84,14 +84,14 @@
     margin-left: 0;
   }
 
-  // Put indentation back where we use list style types
-  .gem-c-contents-list__list-item--dashed {
-    margin-left: govuk-spacing(3);
-    list-style-type: disc;
-  }
-
   .gem-c-contents-list__list-item--numbered,
   .gem-c-contents-list__list-item--parent {
     list-style-type: none;
+  }
+
+  .gem-c-contents-list__link {
+    &.govuk-link {
+      color: $govuk-print-text-colour;
+    }
   }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cross-service-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cross-service-header.scss
@@ -436,3 +436,32 @@ $govuk-header-link-underline-thickness: 3px;
   }
 }
 // end service navigation styles
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-cross-service-header {
+    margin-bottom: 5mm;
+  }
+
+  .gem-c-one-login-header {
+    background-color: initial;
+    border-bottom: 2pt solid $govuk-print-text-colour;
+
+    * {
+      color: $govuk-print-text-colour;
+    }
+
+    .govuk-width-container {
+      margin: 0;
+    }
+  }
+
+  .gem-c-service-header {
+    .govuk-width-container {
+      margin: 0;
+    }
+
+    .gem-c-service-header__container {
+      padding: 0;
+    }
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_devolved-nations.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_devolved-nations.scss
@@ -10,3 +10,26 @@
     padding: govuk-spacing(4) govuk-spacing(6);
   }
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-devolved-nations {
+    border: 2pt solid $govuk-print-text-colour;
+    margin: 0 0 5mm;
+
+    * {
+      color: $govuk-print-text-colour !important;
+    }
+
+    .govuk-list li {
+      margin-bottom: 2mm;
+    }
+
+    .govuk-link::after {
+      display: block;
+      margin: 1mm auto;
+      font-size: 12pt;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
@@ -45,3 +45,19 @@
 .gem-c-heading--inverse {
   color: govuk-colour("white");
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  // Override margins and colors
+  .gem-c-heading {
+    margin: 0 !important;
+    color: $govuk-print-text-colour !important;
+    border-color: $govuk-print-text-colour !important;
+    padding: 0;
+  }
+
+  .gem-c-heading--padding[class*="--border-top"] {
+    padding-top: govuk-spacing(3);
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
@@ -21,3 +21,32 @@
   height: 1em;
   margin-bottom: -2px;
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-intervention {
+    break-inside: avoid;
+    background-color: transparent;
+    border-color: $govuk-print-text-colour;
+
+    .govuk-body {
+      margin: 0;
+    }
+  }
+
+  .gem-c-intervention .govuk-link {
+    &,
+    &:link,
+    &:link:visited {
+      color: $govuk-print-text-colour !important;
+    }
+
+    &::after {
+      display: block;
+      margin: 1mm auto;
+      font-size: 12pt !important;
+      color: $govuk-print-text-colour;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
@@ -12,3 +12,9 @@
 .gem-c-lead-paragraph--inverse {
   color: govuk-colour("white");
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-lead-paragraph {
+    color: $govuk-print-text-colour;
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
@@ -94,3 +94,19 @@
 .govuk-frontend-supported .gem-c-metadata__toggle-items.js-hidden {
   display: none;
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-metadata {
+    break-inside: avoid;
+    color: $govuk-print-text-colour !important;
+    background-color: unset;
+    margin-bottom: 5mm;
+
+    a,
+    a:visited {
+      color: $govuk-print-text-colour !important;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
@@ -26,3 +26,24 @@
     }
   }
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-phase-banner {
+    &,
+    & * {
+      background-color: unset !important;
+      color: $govuk-print-text-colour !important;
+    }
+
+    .govuk-tag {
+      border: 1pt solid $govuk-print-text-colour;
+    }
+  }
+
+  .govuk-phase-banner__content__app-name {
+    margin-bottom: 1mm;
+    font-size: 12pt;
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
@@ -1,2 +1,10 @@
 @import "govuk_publishing_components/individual_component_support";
 @import "govuk/components/pagination/pagination";
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .govuk-pagination * {
+    color: $govuk-print-text-colour !important;
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_success-alert.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_success-alert.scss
@@ -4,3 +4,25 @@
 .gem-c-success-alert__message {
   @include govuk-font(19, $weight: bold);
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-success-alert {
+    break-inside: avoid;
+    background-color: initial;
+    border-color: $govuk-print-text-colour;
+
+    * {
+      color: $govuk-print-text-colour !important;
+    }
+
+    .govuk-notification-banner__header {
+      border-bottom: 1pt solid $govuk-print-text-colour;
+    }
+
+    .govuk-notification-banner__title {
+      padding: 2mm 0;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_title.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_title.scss
@@ -32,11 +32,19 @@
 }
 
 @include govuk-media-query($media-type: print) {
+  .gem-c-title {
+    margin: 5mm 0 10mm !important; // stylelint-disable-line declaration-no-important
+  }
+
   .gem-c-title__context {
     margin: 0;
   }
 
   .gem-c-title__text {
     margin-top: 0;
+  }
+
+  .gem-c-title--inverse {
+    color: $govuk-print-text-colour;
   }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
@@ -21,3 +21,14 @@
 .gem-c-warning-text__text--highlight {
   color: $govuk-error-colour;
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-warning-text * {
+    color: $govuk-print-text-colour;
+  }
+
+  .govuk-warning-text__icon {
+    background-color: transparent;
+    color: $govuk-print-text-colour;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
@@ -10,9 +10,11 @@
   <% if breadcrumb_selector.step_by_step %>
     <%= render 'govuk_publishing_components/components/step_by_step_nav_header', breadcrumb_selector.breadcrumbs %>
   <% elsif breadcrumb_selector.breadcrumbs %>
-    <%= render 'govuk_publishing_components/components/breadcrumbs',
-               breadcrumbs: breadcrumb_selector.breadcrumbs,
-               inverse: inverse,
-               collapse_on_mobile: collapse_on_mobile %>
+    <div class="govuk-!-display-none-print">
+      <%= render 'govuk_publishing_components/components/breadcrumbs',
+                 breadcrumbs: breadcrumb_selector.breadcrumbs,
+                 inverse: inverse,
+                 collapse_on_mobile: collapse_on_mobile %>
+    </div>
   <% end %>
 </div>
--- a/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_accordion.scss
@@ -11,10 +11,15 @@
 //
 // TODO: Replace with the print styles that will come from GOV.UK Frontend.
 
+// stylelint-disable declaration-no-important
 @include govuk-media-query($media-type: print) {
   // Open all of the accordion sections.
   .govuk-accordion__section-content {
-    display: block !important; // stylelint-disable-line declaration-no-important
+    display: block !important;
+
+    > :last-child {
+      margin-bottom: 20px;
+    }
   }
 
   .govuk-frontend-supported .govuk-accordion__section-content[hidden] {
@@ -25,7 +30,9 @@
 
   // Change the colour from the blue link colour to black.
   .govuk-accordion__section-button {
-    color: govuk-colour("black") !important; // stylelint-disable-line declaration-no-important
+    color: govuk-colour("black") !important;
+    padding: 20px 0 0 !important;
+    border-bottom: 0 !important;
   }
 
   // Change the summary subheading size.
@@ -37,6 +44,14 @@
   // Hide the unusable "Show all" and "Show" sections.
   .govuk-accordion__show-all,
   .govuk-accordion__section-toggle {
-    display: none !important; // stylelint-disable-line declaration-no-important
+    display: none !important;
+  }
+
+  // Remove extra space for expanded accordions, not needed on a printout
+  .govuk-accordion__section--expanded {
+    .govuk-accordion__section-content {
+      padding: 0 !important;
+    }
   }
 }
+// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -183,3 +183,22 @@
     display: none;
   }
 }
+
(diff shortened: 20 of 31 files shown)


AshGDS (commented):
Looks good, nice work 👍  I've added some comments but overall it looks good.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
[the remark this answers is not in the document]
matthillco: This is an interesting one. You're seeing black-on-blue here due to the emulated print preview in Chrome dev tools. When previewing this through the printer dialogue, it's displayed as black on white for clearer printing, unless the "Background graphics" option is ticked, in which case it will print as black on blue on a colour printer, or black on grey on a B&W printer, neither of which are ideal. We want to print these as black on white.

The way to fix this is to ensure that the background is removed for these component previews when in print mode (emulated or real). I've resolved this in [`_application.scss lines 140-145`](https://github.com/alphagov/govuk_publishing_components/pull/4073/files#diff-f5da6136011213e1fc9a94945071686dca467cd7602e3d6535c652f121dc263c).

app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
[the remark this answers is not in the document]
matthillco: This is resolved by the [fix above](https://github.com/alphagov/govuk_publishing_components/pull/4073#discussion_r1696711090).

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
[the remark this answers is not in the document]
matthillco: This is resolved by the [fix above](https://github.com/alphagov/govuk_publishing_components/pull/4073#discussion_r1696711090).

app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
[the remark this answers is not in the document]
matthillco: These are set to not print at all, so nothing more required here.

app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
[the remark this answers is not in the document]
matthillco: This is resolved by the [fix above](https://github.com/alphagov/govuk_publishing_components/pull/4073#discussion_r1696711090).

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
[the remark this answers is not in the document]
matthillco: It matches the spacing on other variants that have two lines so I'll keep it as-is.

app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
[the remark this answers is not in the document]
matthillco: I've made a note of this as something to look at in future, I think it's an extreme edge-case and out of scope for this PR.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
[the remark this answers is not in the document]
matthillco: I've made a note of this as something to look at in future: I think it's an edge-case and out of scope for this PR which is focussed on how things look when printed.

app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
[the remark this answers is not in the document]
matthillco: This is now fixed at lines [32-37](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/59b1c93ad41e6dba437782691eb4e63b51957ea2#diff-cf0c63d138f6df25bcb21fa41da642f7b3cc37a93822b02e78ee1258387f3a16R32-R37).

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -84,3 +84,45 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey");
   text-decoration: none;
   cursor: help;
 }
+
+@include govuk-media-query($media-type: print) {
```
andysellick, line 88: The result of this looks very neat and tidy but I think it might be overly complicated - personally I'd just hide the icon, it doesn't really do much for print. Then the CSS would be much shorter:

```CSS
  .gem-c-attachment__thumbnail {
    display: none;
  }

  .gem-c-attachment__details {
    padding: 0;
  }

  .gem-c-attachment__metadata,
  .gem-c-attachment__metadata .govuk-link::after {
    font-size: 12pt;
  }

  .gem-c-attachment__link.govuk-link {
    word-break: break-word;

    &::after {
      font-size: 12pt;
    }
  }
}
```

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -84,3 +84,45 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey");
...
+    word-break: break-word;
+
+    &::after {
+      font-size: 12pt;
+      line-height: 1 !important; // stylelint-disable-line declaration-no-important
+    }
+  }
+
+  .gem-c-attachment__metadata,
+  .gem-c-attachment__metadata .govuk-link::after, {
```
andysellick, line 125: Nitpick: there's a stray comma at the end of this line.

app/assets/stylesheets/govuk_publishing_components/components/_attachment-link.scss
```diff
@@ -1,4 +1,12 @@
+@import "govuk_publishing_components/individual_component_support";
```
andysellick, line 1: Nitpick: would you be able to add a little more detail about this change in the commit message? For example, in this commit I'd write something like:

```
Improve attachment link component print styles

- add individual_component_support to allow `govuk-media-query` to work
- set font size to inherit because...
```

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
```diff
@@ -58,3 +58,27 @@
...
   vertical-align: middle;
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-big-number__value {
+    font-size: 56pt;
+  }
+
+  .gem-c-big-number__suffix {
+    vertical-align: unset;
```
andysellick, line 68: Nitpick: is this needed? Oddly in screen when this is removed the plus moves up, but in print it moves down - but either way I don't think it looks bad.

andysellick (commented):
Have only reviewed the first few commits so far - up to and including the big number changes.

app/views/govuk_publishing_components/components/_copy_to_clipboard.html.erb
```diff
@@ -2,7 +2,7 @@
   button_data_attributes ||= nil
   input_data_attributes ||= nil
 %>
-<div class="gem-c-copy-to-clipboard" data-module="copy-to-clipboard">
+<div class="gem-c-copy-to-clipboard govuk-!-display-none-print" data-module="copy-to-clipboard">
```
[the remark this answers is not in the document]
matthillco: We chose to hide this for print as it's only useful in its interactive state on screen. I don't see the harm including it even if it's not used publicly: the change covers every scenario where it might be printed.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
matthillco: The icons are a quick way to determine the filetype, so I favour keeping them for both screen and print. Better comprehension of printed documents is what we're aiming for, and visuals help with that. I'd favour keeping them for this reason, but perhaps we need another design opinion here?

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
matthillco: This component actually needs more attention (possible refactor) because the screen styles for the suffix are broken: they don't work with different types of suffix, for example % or letters. However I didn't want to refactor the screen styles as part of this work which is print focussed. (I've recorded this as something to look at in the future).

The `vertical-align: unset` fixes the incorrect vertical centring of other characters, which you can see if you use dev tools to change the + in the example to %, £, M etc. Everything now lines up much better. 

That said, perhaps it's strange to fix this for print and not screen, so I'd be happy to remove it here and defer the whole thing to a different ticket to fix the entire component if that makes more sense?

app/assets/stylesheets/govuk_publishing_components/components/_attachment-link.scss
matthillco: Commit description updated as requested. Note that I made a couple of changes to this commit compared to what you previously reviewed.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
matthillco: Fixed.

app/views/govuk_publishing_components/components/_contextual_guidance.html.erb
```diff
@@ -9,7 +9,7 @@
   guidance_id ||= nil
 %>
 
-<%= tag.div class: "gem-c-contextual-guidance", id: id, data: data_attributes do %>
+<%= tag.div class: "gem-c-contextual-guidance govuk-!-display-none-print", id: id, data: data_attributes do %>
```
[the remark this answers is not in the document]
matthillco: OK, I've removed this commit.

app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb
```diff
@@ -6,7 +6,7 @@
   collapse_on_mobile ||= true unless local_assigns[:collapse_on_mobile].eql?(false)
 %>
 
-<div class="gem-c-contextual-breadcrumbs">
+<div class="gem-c-contextual-breadcrumbs govuk-!-display-none-print">
```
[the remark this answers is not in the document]
matthillco: Yes, we weren't sure about this one. I've done as suggested and only applied the 'no print' class to the breadcrumbs only.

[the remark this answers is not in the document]
matthillco: Yes, we were unsure about this one. I've now done as suggested, the breadcrumbs are wrapped with a 'no print' div.

matthillco:
Due to the size of this PR, I'm going to remove the following 10 commits to their own PRs today (7/Aug/2024). FYI @andysellick. I'll make another comment when this has been completed.

- [`inverse_header` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/e204a01811c010f8793c9896a4f85da0bbbb3fdf)
- [`step_by_step_nav_header` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/dd2d02e7c69b28d9974067f73b5c97ab3bfa7224)
- [`step_by_step_nav_related` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/5cca5a5e4f167fc2aaad4286fea970449280cb12)
- [govspeak `attachment` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/59821058f6c510b2f23deeb4f8e7196d6ed11ee8)
- [govspeak `button` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/f38c3c4b34fd2543b928bdf6e4d1c36971fd205a)
- [govspeak `highlight_answer` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/7f8b65b4ca4e55aa749640fbf74c70c8b4ce3eb3)
- [govspeak `media_player` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/25f0eb4fb61bdacb2778ef32ce18e8c0cce2b5b2)
- [`table` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/6bc362c20fa92a6d437ce72eedd12da2de7165f7)
- [`tabs` component print styles](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/b7a80237fa924a9f9bc9e7c5aab1fa57960f4b1b)
- [Remove dark backgrounds from inverted headings](https://github.com/alphagov/govuk_publishing_components/pull/4073/commits/12a35b2238cea35636eac657fe44a28339bbe580)

matthillco:
The 12 commits listed above have now been moved over to their own PRs. @andysellick

app/views/govuk_publishing_components/components/_copy_to_clipboard.html.erb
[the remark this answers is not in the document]
andysellick: Only problem is this component is a form field and we decided not to print form fields, and in this case it would be the only one in the form that won't appear in print, which might be a bit strange.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
andysellick: The file type is written underneath the title, and I think it's less useful to know information about a link on a printed page. Fine to ask design though.

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
andysellick: Yes, I think that makes more sense. Can you raise an issue in the repo detailing this? Thanks.

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
```diff
@@ -131,21 +131,42 @@
...
+    &:link,
+    &:link:visited {
+      color: $govuk-print-text-colour !important;
+    }
+
+    &::before {
+      display: none;
+    }
+
+    &::after {
```
andysellick, line 163: What do these styles do? It doesn't look like the after element contains anything, looks like it's just there as an overlay on the link so that the clickable area is expanded. I think you can set this to just `display: none` and it'll work fine?

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -84,12 +84,6 @@
     margin-left: 0;
   }
 
-  // Put indentation back where we use list style types
-  .gem-c-contents-list__list-item--dashed {
```
andysellick, line 88: Good improvement. Is it also worth setting the link colour on the contents list to the print/black?

On that, I've been thinking about a general approach and wondering if that's something we can have in more general print styles. Not for this PR, but we could have a general class to apply to links where we want them to print not in the link colour. Separating it out like this would help clarify the intent behind it, and limit the number of times we have to repeat this across the components.

```CSS
.print-link-style {
  color: $govuk-print-text-colour !important;
}
```

```HTML
<a href="#" class="govuk-link">Don't care what this looks like when printed</a>
<a href="#" class="govuk-link print-link-style">Do care what this looks like when printed</a>
```

app/assets/stylesheets/govuk_publishing_components/components/_devolved-nations.scss
```diff
@@ -10,3 +10,25 @@
...
+    * {
+      color: $govuk-print-text-colour !important;
+    }
+
+    .govuk-list li {
+      margin-bottom: 2mm;
+    }
+
+    .govuk-link::after {
+      display: block;
```
andysellick, line 29: It looks like the `after` text containing the link href is smaller than the actual link text (in line with your other changes, which I agree with) but it looks the same here without an explicit font size declaration. Have I remembered this wrong? Is the font size of the `after` automatically smaller anyway and we can remove those other lines, or does it need adding here as well?

andysellick: Same comment for document list, and possibly others.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -45,3 +45,25 @@
...
+  // Override margins and colors
+  .gem-c-heading {
+    margin: 0 !important;
+    color: $govuk-print-text-colour !important;
+    border-color: $govuk-print-text-colour !important;
+  }
+
+  // Remove extra padding to reduce whitespace, except for headings with borders
+  .gem-c-heading--padding:not([class*="--border-top"]) {
+    padding: 0 !important;
```
andysellick, line 61: Not sure `important` is needed here?

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -45,3 +45,25 @@
...
+  }
+
+  // Remove lower padding for headings with borders
+  .gem-c-heading--padding[class*="--border-top"] {
+    padding-bottom: 0 !important;
```
andysellick, line 66: Not sure `important` is needed here?

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -45,3 +45,25 @@
 .gem-c-heading--inverse {
   color: govuk-colour("white");
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+
```
andysellick, line 51: Nitpick: can remove empty line here.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
```diff
@@ -45,3 +45,25 @@
...
+    border-color: $govuk-print-text-colour !important;
+  }
+
+  // Remove extra padding to reduce whitespace, except for headings with borders
+  .gem-c-heading--padding:not([class*="--border-top"]) {
+    padding: 0 !important;
+  }
+
+  // Remove lower padding for headings with borders
+  .gem-c-heading--padding[class*="--border-top"] {
```
andysellick, line 65: Appreciate that this is slightly complex to restyle given the various colliding classes, but I think this whole section can be written slightly simpler as:

```CSS
// stylelint-disable declaration-no-important
  // Override margins and colors
  .gem-c-heading {
    margin: 0 !important;
    color: $govuk-print-text-colour !important;
    border-color: $govuk-print-text-colour !important;
    padding: 0;
  }

  .gem-c-heading--padding[class*="--border-top"] {
    padding-top: govuk-spacing(3);
  }
}
// stylelint-enable declaration-no-important

```

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
andysellick, line 1: Probably worth adding the govuk don't print class to the `govuk-link js-dismiss-link` element (the 'hide this suggestion' link) - it doesn't add any value to the print experience.

andysellick: Might also be worth setting the border-left colour on `gem-c-intervention` to the print colour?

app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
```diff
@@ -12,3 +12,13 @@
 .gem-c-lead-paragraph--inverse {
   color: govuk-colour("white");
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-lead-paragraph {
+    break-inside: avoid;
```
andysellick, line 19: Nitpick: I wonder if this is necessary - the lead paragraph generally occurs at the start of content, so it's unlikely to be at the boundary of a page, plus even if it does overlap the bottom of a page, forcing it to start on a new page might actually end up using more paper than necessary - thoughts?

andysellick (commented):
Another partial review - this time got up to and including lead paragraph.

app/assets/stylesheets/govuk_publishing_components/components/_success-alert.scss
andysellick, line 1: Do we need to print the success alert component at all?

app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
andysellick, line 1: Do we need to print the previous and next component?

andysellick (commented):
Finished going through all the commits.

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
matthillco: In the print view, the `::after` element changes from the overlay to the printable URL link. These styles are making the printable link a bit smaller and on their own line, to be consistent with the same treatment in other components.

app/assets/stylesheets/govuk_publishing_components/components/_devolved-nations.scss
matthillco: I missed setting the explicit 12pt font-size here, so it's still using the 90% font-size as determined by the design system. However both 90% and 12pt look very closely matched in size for default text. 

The reason I've set 12pt sizes for printed links is to ensure printed consistency, just in case the 90% value might calculate to something other than 12pt in unexpected scenarios.

I've set this to 12pt in this case (and will do the same for other components that have the same problem).

app/assets/stylesheets/govuk_publishing_components/components/_success-alert.scss
matthillco: Nikin and I discussed this and we felt it was potentially useful if a user printed the page, as it confirms the successful completion of a task.

app/assets/stylesheets/govuk_publishing_components/components/_previous-and-next-navigation.scss
matthillco: Nikin and I discussed this and we felt this was useful when printed because:
- it adds context to the currently printed step, ie, where it fits in a sequence
- it could help people organise the stack of printed steps in the correct sequence.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
matthillco: Superseded by your code block below.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
matthillco: Superseded by your code block below.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
matthillco: Superseded by your code block below.

app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
matthillco: Aha, yeah that's cleaner, thanks for taking the time to improve it.

app/assets/stylesheets/govuk_publishing_components/components/_cross-service-header.scss
```diff
@@ -436,3 +436,40 @@ $govuk-header-link-underline-thickness: 3px;
...
+
+    .gem-c-service-header__container {
+      padding: 0;
+    }
+  }
+
+  .gem-c-one-login-header__nav,
+  .gem-c-service-header nav,
+  .gem-c-service-header button {
+    display: none !important;
```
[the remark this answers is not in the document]
matthillco: I've removed this block and added classes in the views as suggested.

app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
```diff
@@ -21,3 +21,18 @@
 .gem-c-warning-text__text--highlight {
   color: $govuk-error-colour;
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-warning-text,
+  .gem-c-warning-text__text,
+  .gem-c-warning-text__text--highlight {
+    color: $govuk-print-text-colour !important;
```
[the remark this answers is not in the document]
matthillco: I've removed `!important` and also reduced the named classes by using the universal selector.

app/views/govuk_publishing_components/components/_copy_to_clipboard.html.erb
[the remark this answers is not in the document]
matthillco: I understand that point, but this component really is redundant in print so I'd prefer to remove it.

app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
matthillco: Good point, I've removed it and also the `!important` and the stylelint exclusion for important.

app/assets/stylesheets/govuk_publishing_components/components/_warning-text.scss
```diff
@@ -21,3 +21,18 @@
...
+  }
+
+  .govuk-warning-text__icon {
+    background-color: unset;
+    color: $govuk-print-text-colour !important;
```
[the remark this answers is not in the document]
matthillco: Removed.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
matthillco: I checked with Nikin and he's happy to remove the thumbnail, so this is now updated to match your suggestion. I've also reduced stylelint block down to a single line for the single `!important`.

app/assets/stylesheets/govuk_publishing_components/components/_big-number.scss
matthillco: I've deleted this commit and raised an issue here: [4159](https://github.com/alphagov/govuk_publishing_components/issues/4159)

app/assets/stylesheets/govuk_publishing_components/components/_phase-banner.scss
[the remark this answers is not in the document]
matthillco: Nikin and I did discuss this and we felt this is a useful reminder that the page was printed from a service in alpha/beta.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: I've added in a colour rule to force the content list links to print in black. For your other comment I've copied this out as a note into our backlog ticket as something to look at.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
matthillco: Both these fixes have been applied. I've also added a rule to remove the bottom margin of the last paragraph that's followed by the dismiss link (otherwise there's extra spacing when the dismiss link is hidden for print).

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -94,3 +94,17 @@
...
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-metadata--inverse {
+    color: $govuk-print-text-colour !important;
+    background-color: unset;
+
+    a,
+    a:visited {
+      color: $govuk-print-text-colour !important;
```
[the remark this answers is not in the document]
matthillco: I've removed the `--inverse` from the selector so the nested rules now apply to all variants.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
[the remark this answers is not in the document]
matthillco: This seems to be an issue with the component being printed across a page break. I've added `break-inside: avoid` to prevent that.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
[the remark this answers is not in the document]
matthillco: The border you've shown here seems to be around the component preview element, which has extra spacing and margins included. Taking that into account, I've reduced the title element margins to 5mm above and 10mm below. This will probably need refinement in the page types print styles work to follow.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
```diff
@@ -39,4 +40,10 @@
   .gem-c-title__text {
     margin-top: 0;
   }
+
+  .gem-c-title--inverse {
+    color: $govuk-print-text-colour !important;
```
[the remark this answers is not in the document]
matthillco: OK, removed.

app/views/govuk_publishing_components/components/_copy_to_clipboard.html.erb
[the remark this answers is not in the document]
andysellick: I did mean that, sorry for the confusion. Thanks for updating.

CHANGELOG.md
```diff
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 42.0.0
```
andysellick, line 10: I think something's gone wrong with this commit - it shouldn't be in here, it's already in `main`. Try a rebase first, and if that doesn't remove it probably just delete this commit from this PR.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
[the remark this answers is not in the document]
andysellick: No I just meant there's a lot of space below the 'Last updated' line and the bottom edge of the component. I didn't mean to add the break-inside - I think it's a case of removing or reducing some margin/padding.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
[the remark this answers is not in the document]
andysellick: Yes, I put the border round there deliberately because if it was on the component it wouldn't show the margins. Happy with your change - although the last example (`Using design system template`) still seems to have a lot of spacing?

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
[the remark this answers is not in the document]
matthillco: I think in your screenshot it's a combination of a lower margin that's too big and proximity of the component to the page end. If the browser can't quite fit the wrapper preview on the printed page, it seems to increase its size so that it'll be printed over two pages. Adding the `break-inside:avoid` does fix this, but I've also reduced the bottom margin on `gem-c-metadata` to 5mm rather than the 40px it currently uses. We might need to tweak this during the page styles work.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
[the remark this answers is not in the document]
matthillco: Investigating this, that instance is constructed differently. It has an extra `main` element with class `govuk-main-wrapper` wrapping the component. That class has extra 40px top and bottom padding. 

I can remove that extra padding for print, but perhaps there's a good reason to keep it that we're not currently aware of. Maybe this is one to leave for now and investigate as part of the page types print styles work?

CHANGELOG.md
matthillco: Strange! I'm not sure how this got in here, perhaps I mistakenly included it by going too far back in the history during an interactive rebase. I've rebased against main again and it's disappeared now.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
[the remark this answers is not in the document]
andysellick: Fair enough.

app/assets/stylesheets/govuk_publishing_components/components/_intervention.scss
```diff
@@ -21,3 +21,32 @@
...
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-c-intervention {
+    break-inside: avoid;
+    background-color: transparent;
+    border-color: $govuk-print-text-colour;
+
+    .govuk-body:has(+ .govuk-body .js-dismiss-link) {
```
andysellick, line 32: `.govuk-body:has(+ .govuk-body .js-dismiss-link)` seems quite complicated. Now that we're hiding the dismiss link, I think there'll only be one `.govuk-body` element shown in print, so could this be simplified? e.g.

```CSS
    .govuk-body {
      margin: 0;
    }
```

matthillco: I did consider that, but if there's a possibility of more than one paragraph with `.govuk-body` then `margin: 0` will remove all of the spacing between paragraphs which we wouldn't want.

andysellick: There isn't a possibility of more than one paragraph, see https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_intervention.html.erb

While I was looking at that file, I noticed something else - the `govuk-!-display-none-print` class is on the `govuk-link` element, but it should be on the wrapping `govuk-body` element (at the moment it'll not show the link in print, but will show the 'empty' wrapping paragraph, which could include margin).

matthillco: OK, thanks, yes your solution is better, that's now updated.

app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
```diff
@@ -12,3 +12,10 @@
 .gem-c-lead-paragraph--inverse {
   color: govuk-colour("white");
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-lead-paragraph {
+    color: $govuk-print-text-colour;
+    background-color: unset;
```
andysellick, line 19: I don't think we need to set a background colour here - inverse components don't set a background colour, they sit on elements that already specify the colour.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
```diff
@@ -32,11 +32,20 @@
...
     margin: 0;
   }
 
   .gem-c-title__text {
     margin-top: 0;
   }
+
+  .gem-c-title--inverse {
+    color: $govuk-print-text-colour;
+    background-color: unset;
```
andysellick, line 49: I don't think we need to unset the background colour here - the component doesn't set one, it's handled by whatever element it is placed into.

andysellick (commented):
Doing a last pass review, just noticed a handful of minor things.

andysellick:
@matthillco also can you add a CHANGELOG entry for this PR? Thanks.

app/assets/stylesheets/govuk_publishing_components/components/_title.scss
matthillco: OK, removed.

app/assets/stylesheets/govuk_publishing_components/components/_lead-paragraph.scss
matthillco: OK, removed.

andysellick (approved):
Great stuff

PR #4068, opened 2024-06-19
Remove Universal Analytics

andysellick:
## What
Remove the Universal Analytics code. Supersedes https://github.com/alphagov/govuk_publishing_components/pull/4067

BREAKING CHANGES:

- details component no longer has a JS file
- TrackSelectChange (`track-select-change`) module has been removed, referenced in `finder-frontend` at least
- organisation logo no longer accepts data attributes passed to it
- intervention component no longer accepts data attributes passed to it
- image card component no longer accepts data attributes for child links

Things to potentially investigate further:

- does step nav (and related components) need the `tracking_id` option anymore? Looks like that value is recorded in the GA4 page view and nowhere else. ANSWER: yes it is needed, it gets used in session storage to record which bits of a step nav are open, have emphasised this in the comments for clarity.

## Why
Universal Analytics is being switched off at the end of June and has been replaced with Google Analytics 4.

## Visual Changes
Several visual changes, but all because of removed entries from YML files. This reflects that some components no longer have specific options for UA tracking, and some have had their ability to accept data attributes in some circumstances removed because this was only used for UA tracking.

Trello card: https://trello.com/c/rNPqwDIr/188-remove-ua-code-from-components-gem

PR #4074, opened 2024-06-24
Add open attribute to component wrapper

andysellick:
## What
- adds the ability to pass an open attribute to the component wrapper helper
- needed for the details component, where open can be passed as true to have the details component expanded by default

## Why
Originally done as part of https://github.com/alphagov/govuk_publishing_components/pull/4068 but that PR is huge enough without this additional change, so I'm separating them out.

## Visual Changes
None.

Trello card: https://trello.com/c/rNPqwDIr/188-remove-ua-code-from-components-gem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add open attribute to component wrapper ([PR #4074](https://github.com/alphagov/govuk_publishing_components/pull/4074))
 * Add guard to accordion component init method ([PR #4069](https://github.com/alphagov/govuk_publishing_components/pull/4069))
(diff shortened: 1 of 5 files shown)


AshGDS (commented):
Looks good 👍 

According to MDN, setting open to `false` will open the details component:


> Note: You have to remove this attribute entirely to make the details hidden. open="false" makes the details visible because this attribute is Boolean.


Unless I'm misreading the code, does this need to be reflected in this PR? I can see that in your `"can set an open attribute, overriding a passed value"` test the open attribute is `nil` after you set the value to `false`, so maybe you've already accounted for this?

andysellick:
Good point @AshGDS thanks. Have added an explicit test to check this, do you think that covers this?

AshGDS (approved):
Thanks, looks good 👍

andysellick:
Thanks @AshGDS

PR #4084, opened 2024-06-28
Fix missing number formatting in contents list component

andysellick:
## What
- the contents list component does clever stuff when you pass the `format_numbers` option, where it identifies numbers at the start of the link text, splits the string into two, and wraps the number and text in separate spans so that the spacing between them can be controlled with CSS (otherwise 1. is misaligned with 10.)
- in some circumstances the content arriving at the component has a non-breaking space instead of an actual space, i.e. `1.&nbsp;Text of link`, which the component doesn't recognise and fails to split and wrap with spans for nice formatting, e.g. https://www.gov.uk/government/publications/financial-sanctions-guidance-for-russia/financial-sanctions-guidance-for-russia
- this change allows for this situation and converts the non-breaking space into a regular space prior to formatted number checking
- it also changes how the wrapped spans are returned (by now simply extracting the number and remaining text, wrapping them in spans and returning them) instead of a slightly more complex string replacing approach designed to preserve anything outside of the string
- this includes refactoring the code and tests to simplify the code and limit the method to assume that the string passed to it should only be text, not markup, by stripping any HTML from the input
- the helper spec in particular implied that markup (including a wrapping link) should be accepted by the method, but this is unrealistic - the output from the method is always wrapped in a parent (either a link or a span) and should not contain markup, as this is the text of the link, not a free area to inject HTML
- this was being done in one place (to include an abbreviation tag inside the link) but has been corrected, see https://github.com/alphagov/government-frontend/pull/3241

## Why
Formatting was incorrect in some circumstances.

## Visual Changes
Before | After
------ | ------
Screenshot 2024-06-28 at 14 37 57 | Screenshot 2024-06-28 at 14 38 10

Trello card: https://trello.com/c/S5QIYLRx/9-mis-aligned-sub-headings

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix missing number formatting in contents list component ([PR #4084](https://github.com/alphagov/govuk_publishing_components/pull/4084))
+
 ## 39.2.0
 
 * Disable Universal Analytics ([PR #4083](https://github.com/alphagov/govuk_publishing_components/pull/4083))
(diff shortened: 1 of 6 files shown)


KludgeKML (approved):
LGTM!

PR #4089, opened 2024-07-02
Update to LUX 4.0.20

andysellick:
## What
Update to the newest version of LUX. See https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#how-to-update-to-a-new-version for full details.

This seems to be quite a significant change since the last version (3.14) so there are a lot of changes in the diff. I've tried to reformat them as cleanly as possible to minimise this, but it is a large change nonetheless. I've tested it as per the procedure and it appears to be working correctly.

## Why
If we don't update LUX will stop working.

## Visual Changes
None.

Trello card: https://trello.com/c/vf2DrJR8/210-update-lux-script

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Update to LUX 4.0.20 ([PR #4089](https://github.com/alphagov/govuk_publishing_components/pull/4089))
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Code changes look good to me, can see `customerid` and `samplerate` are set as expected 👍

PR #4091, opened 2024-07-03
Add more applications to component audit

andysellick:
## What
Adds some more applications that use `govuk_publishing_components` to the component auditing tools.

## Why
The app consistency sheet was recently updated and there were a number of applications using the gem on there that weren't in here.

## Visual Changes
Only that I put the number of applications in at the top.

Screenshot 2024-07-03 at 14 29 38


maxgds (approved):
... even though I hate that Maslow is still in there and wonder if search-v2-evaluator is current or should be retired

andysellick:
@maxgds thanks, and agreed, but I think there's value in auditing them for now at least.

PR #4097, opened 2024-07-09
Only set gtm_cookies_win=x for preview in GTM script

AgaDufrat:
## What
Only set gtm_cookies_win=x for preview in GTM script

## Why
The GTA doesn’t work as expected for DGU.

It appears that gtm_cookies_win=x allows GTM Preview to change the environment of a page, which means that they win over gtm_auth and gtm_preview parameters.

Setting gtm_cookies_win=x  let the GTM editor decide which environment they want to use, regardless of what was in the tracking snippet. 

This makes it identically to the behaviour in ga4-core.js. The commit message suggest that it can slow down GTM:
https://github.com/alphagov/govuk_publishing_components/blob/eb66a18f86/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-core.js#L25

https://govuk.zendesk.com/agent/tickets/5626599

## Visual Changes
n/a

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Only set `gtm_cookies_win=x` for preview in Google Tag Manager script ([PR #4097](https://github.com/alphagov/govuk_publishing_components/pull/4097))
+
 ## 39.2.2
 
 * Adjust govspeak chart label positioning ([PR #4094](https://github.com/alphagov/govuk_publishing_components/pull/4094))
--- a/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
+++ b/app/views/govuk_publishing_components/components/_google_tag_manager_script.html.erb
@@ -2,11 +2,12 @@
   gtm_auth ||= nil
   gtm_preview ||= nil
 
-  gtm_attributes = %w(gtm_cookies_win=x)
+  gtm_attributes = []
     gtm_attributes << "gtm_auth=" + gtm_auth if gtm_auth
-    gtm_attributes << "gtm_preview=" + gtm_preview if gtm_preview
+    gtm_attributes << "gtm_preview=#{gtm_preview}&gtm_cookies_win=x" if gtm_preview
   gtm_attributes = gtm_attributes.join('&')
 %>
+
 <%= javascript_tag nonce: true do -%>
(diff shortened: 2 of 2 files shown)


CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## 39.2.3
```
andysellick, line 10: New additions to the changelog need to be added under the heading `Unreleased`, see https://github.com/alphagov/govuk_publishing_components/blob/main/docs/publishing-to-rubygems.md

Gemfile.lock
```diff
@@ -1,7 +1,7 @@
 PATH
   remote: .
   specs:
-    govuk_publishing_components (39.2.2)
+    govuk_publishing_components (39.2.3)
```
andysellick, line 4: As per previous comment, please remove this.

lib/govuk_publishing_components/version.rb
```diff
@@ -1,3 +1,3 @@
 module GovukPublishingComponents
-  VERSION = "39.2.2".freeze
+  VERSION = "39.2.3".freeze
```
andysellick, line 2: As per previous comment, please remove this.

CHANGELOG.md
andysellick: Thanks @AgaDufrat

andysellick (approved):

CHANGELOG.md
AgaDufrat: Updated as requested but I’m keen to understand the reasons for that. It seems rather unproductive to have to create a new PR just to publish the gem. 

Platform Security and Reliability is trying to improve our gem release process by introducing automation and encouraging engineers to release the changes in the same PR. This would be very valuable insight for this piece of work.

andysellick: Yeah, I can see that. Sorry. The reasoning behind our release process hasn't been documented, but I'd imagine:

- it helps us to minimise the number of releases that happen, by bundling changes together (we're also encouraged to ask others if they have PRs to merge first when doing a release)
- it prevents two PRs from containing the same version number when merging (you'd hope a merge conflict would happen to prevent this, but sometimes it doesn't)
- it gives a clearer history of the releases

It is something that adds a bit of friction to our workflow, so if you have any corresponding insight, I'm sure the frontend team would be up for a discussion.

AgaDufrat: Thank you! Very useful to know.

PR #4100, opened 2024-07-11
Add brand colour for MHCLG

GDSNewt:
The new MHCLG has a new brand colour which will need adding into Whitehall. It's a fairly urgent change hence why we are adding it to publishing components, and not just waiting for the update to govuk-frontend


MartinJJones (approved):
Changes look good to me, colour contrast passed and the CSS class matches what will be set in Whitehall 👍

PR #4102, opened 2024-07-12
LUX version 4.0.23

andysellick:
## What
Updates to the latest version of LUX, as documented here: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md

Have tested it using SpeedCurve's online parser, which weirdly suggests that its detected `v4.0.20`, which is incorrect, so I'm not quite sure what's going on there, but there are clearly some changes in this new version, so I'm certain this code is needed.

## Why
New version.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
LGTM 👍

PR #4106, opened 2024-07-16
Drop libsass support

andysellick:
**BREAKING CHANGE**

## What
Drops support for Sass compilation using libsass. Applications wishing to upgrade must upgrade to dart-sass instead ([example PR of dart-sass upgrade](https://github.com/alphagov/frontend/pull/3908)). 

The only apps where this is still a problem are:

- `content-tagger` (has now been upgraded)
- `collections-publisher` (has now been upgraded)
- `publisher` (being handled by another team)
- `transition` (has now been upgraded see https://github.com/alphagov/transition/pull/1661)
- `govspeak-preview` (has now been upgraded see https://github.com/alphagov/govspeak-preview/pull/509)
- `govuk-developer-docs` (has not been upgraded as sass compilation is handled by middleman, but the changed sass files aren't imported by the dev docs, so it's not currently a problem, see https://github.com/alphagov/govuk-developer-docs/pull/4766)

This PR also fixes several Sass compilation warnings. It should be noted that attempting to compile these changes using libsass will cause problems, either at compilation time or worse, by silently failing to compile properly, leaving parts of the UI incorrectly styled.

## Why
We're having to follow the Design System on this one, but for good reason. Libsass has been deprecated for a while now, but support is likely to be removed completely in v6 (we're on v5). These changes will make upgrading to v6 much easier, although the initial reason for this work was to remove the compilation warnings, which are so verbose that they've caused real problems when trying to debug unrelated problems.

Related PRs to remove Sass compilation warnings:

- https://github.com/alphagov/collections/pull/3701
- https://github.com/alphagov/government-frontend/pull/3268

## Visual Changes
None.

Trello card: https://trello.com/c/gW2NW1sB/142-fix-sass-compilation-warnings


MartinJJones (commented):
Updating `govuk-font-size` to `govuk-typography-responsive` looks good to me.

The only one I'm not 100% on is the change to use `calc` as this previously needed to be reverted, I think this was an issue in application using sassc-rails - https://github.com/alphagov/govuk_publishing_components/pull/3810

jon-kirwan:
> Updating `govuk-font-size` to `govuk-typography-responsive` looks good to me.
> 
> The only one I'm not 100% on is the change to use `calc` as this previously needed to be reverted, I think this was an issue in application using sassc-rails - #3810

Yep, that's right. Most alphagov repos now seem to be using `dartsass-rails` but there are a couple still on `sassc-rails` (such as https://github.com/alphagov/content-tagger). Also, there's other [repositories](https://github.com/alphagov/govuk_publishing_components/network/dependents) that depend on govuk_publishing_components that use `sassc-rails` but I don't know if we need to worry about them or not?

andysellick:
Thanks for the input both. I'm upgrading `content-tagger` to use `dart-sass` as suggested: https://github.com/alphagov/content-tagger/pull/1763

As for the other repositories I don't think we need to worry about them unless they're part of GOV.UK, but I'm happy to look at other applications if there are any.

MartinJJones:
> As for the other repositories I don't think we need to worry about them unless they're part of GOV.UK, but I'm happy to look at other applications if there are any.

Nice work 👍 

There are a few other GOV.UK apps still using sass/sassc-rails which could also be impacted:

**Collections publisher**
https://github.com/alphagov/collections-publisher/blob/main/Gemfile#L11)
[Trello card](https://trello.com/c/FQzURnD0/86-migrate-collections-publisher-to-dart-sass)

**Publisher**
https://github.com/alphagov/publisher/blob/main/Gemfile#L36

**Transition**
https://github.com/alphagov/transition/blob/main/Gemfile#L31

**Govspeak Preview**
https://github.com/alphagov/govspeak-preview/blob/main/Gemfile#L11
[Trello card](https://trello.com/c/PkIk3K1Y/87-migrate-govkspeak-preview-to-dart-sass)

**Developer Docs**
[developer docs Trello card](https://trello.com/c/0dmhie85/41-developer-docs-uses-middleman-with-deprecated-libsass)

Some of these have Trello cards to migrate to DartSass, some a bit more complicated than others as they rely on other gems that use libsass, happy to help out further on this where I can 👍

andysellick:
@MartinJJones all of the other apps on the list have now been updated to `dart-sass` apart from `Publisher`, which is being handled by another team (see PR description for details). Are you happy for this to be merged now? I'm wondering if we should merge it, but mark it as a breaking change?

config/initializers/assets.rb
```diff
@@ -9,11 +9,7 @@
 # from version 4: https://github.com/rails/sprockets/blob/58cca17aa447fcee17703e4ab4dbfaab630e7ed4/UPGRADING.md
 # Rails.application.config.assets.precompile += %w[govuk_publishing_components_manifest.js]
```
MartinJJones, line 10: Small nitpick: non-blocking, we should be able to remove this line to remove a bit of duplication, essentially putting this block of code back to how it was before supporting both `dartsass-rails` and `sassc-rails` - https://github.com/alphagov/govuk_publishing_components/pull/3726/commits/0898868e3aa2921d1b6cb6b73633592eb2ad6e64#diff-52a4487c125f78c29e9b0cec0fbae1a2330507596d5f7cdd32c3d52886c84d4aL10

MartinJJones (approved):
Nice work! I have one small non-blocking comment, other than that the changes look good to me 👍

config/initializers/assets.rb
andysellick: Have changed, does that look correct?

MartinJJones: Looks good to me, thanks 👍

PR #4111, opened 2024-07-23
Drop support for legacy browsers

AshGDS:
## What / Why
- Adds `type=module` to our JavaScript to drop support for legacy browsers
- Moves the `js-enabled` class so it only works in ES6 browsers
- Deletes polyfill files that are no longer needed
- We will combine `es6-components.js` again as a separate task unless you disagree with this
- Trello card: https://trello.com/c/Hoa45SMD/141-turn-off-javascript-in-legacy-browsers
- `type=module` shouldn't have an affect on how our JS is executed, because we use `DOMContentLoaded`, and according to MDN:

> The DOMContentLoaded event fires when the HTML document has been completely parsed, and all deferred scripts ([<script defer src="…">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#defer) and [<script type="module">](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#module)) have downloaded and executed.

- In legacy browsers, this will stop the `static` `layout_for_public` JS from executing. The application specific JS will stop running once the other frontend app PRs are merged. This means for a period of time between `static` being deployed, and the frontend apps being deployed, the legacy browsers may have their application JS being executed, but the JS disabled styles from this gem. I'm not sure the full impact of this, and I'm not sure it will be that big of an issue. I did test this scenario in `finder-frontend`, and the `finder-frontend` search had the non-JS styles from this gem, but it still had the client-side realtime search  from it's own application.js. The page functionality still worked fine in that example, so I don't think there will be an issue. And I guess we can try and deploy the frontend apps quickly to avoid any issues once this change is deployed.

- I've put this as a breaking change in the changelog, since it stops the JS working in old browsers.

## Visual Changes

Browsers that do not support ES6 will have the non `js-enabled` styles.


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
MartinJJones, line 1: Looks good to me, I think we will also need to add a similar update to the layout_for_admin.html.erb template - https://github.com/alphagov/govuk_publishing_components/blob/update-cards-component-to-one-column/app/views/govuk_publishing_components/components/_layout_for_admin.html.erb#L31

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -97,7 +97,7 @@
   </head>
   <%= tag.body class: body_css_classes do %>
     <%= javascript_tag nonce: true do -%>
-      document.body.className += ' js-enabled' + ('noModule' in HTMLScriptElement.prototype ? ' govuk-frontend-supported' : '');
+      document.body.className += ('noModule' in HTMLScriptElement.prototype ? ' js-enabled govuk-frontend-supported' : '');
```
MartinJJones, line 100: I think we should avoid moving the `js-enabled` class here to help keep the page template consistent with the design system - https://design-system.service.gov.uk/styles/page-template/, this would also help show that JS is enabled in the template, but modules are not supported (Grade X browsers)

Although it will likely be more work, we would likely want to update the CSS instead to use the `govuk-frontend-supported` class.

This was the approach used for the upgrade of govuk-frontend from v4 to v5:
- [Tabs govuk-frontend v4 using js-enabled](https://github.com/alphagov/govuk-frontend/blob/support/4.x/package/govuk/components/tabs/_index.scss#L44)
- [Tabs govuk-frontend v5 using govuk-frontend-supported](https://github.com/alphagov/govuk-frontend/blob/953f15903426a8dc508592ed7dad90a253b058d4/packages/govuk-frontend/src/govuk/components/tabs/_index.scss#L48)

Places where this may need to be updated - https://github.com/search?q=repo%3Aalphagov%2Fgovuk_publishing_components%20js-enabled&type=code

MartinJJones (commented):
Changes look good to me overall, just a couple of things to look at, happy to discuss further 👍

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
AshGDS: Thanks - can add, but does this mean I'll also need to sort out adding `type=module` on publishing apps?

MartinJJones: Thanks, good question, I don't think you would need to add type="module" in the publishing apps where `layout_for_admin` template is used.

It does make me think that we if we should make using `type=module` on `application.js` optional in the admin template, defaulting to `false`. This may help avoid situations where application.js is including custom code / polyfills for older browsers and by adding default type="module" we remove support, collections publisher could be an example of this - https://github.com/alphagov/collections-publisher/blob/main/app/assets/javascripts/application.js#L19

AshGDS: Thanks @MartinJJones - we can look into making it conditional for this template 👍

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
AshGDS: Thanks @MartinJJones - I will look into this - `.js-enabled` is also used in all our other frontend apps, so I guess all of those will need to be changed to use `.govuk-frontend-supported` as well 👍  

https://github.com/search?q=org%3Aalphagov+js-enabled+repo%3Aalphagov%2Fgovuk_publishing_components+OR+repo%3Aalphagov%2Fstatic+OR+repo%3Aalphagov%2Fcollections+OR+repo%3Aalphagov%2Ffrontend+OR+repo%3Aalphagov%2Ffinder-frontend+OR+repo%3Aalphagov%2Femail-alert-frontend+OR+repo%3Aalphagov%2Fsmart-answers+OR+repo%3Aalphagov%2Fgovernment-frontend+OR+repo%3Aalphagov%2Fsmokey&type=code&p=1

MartinJJones: Thanks for adding this change in 👍

MartinJJones (commented):
Nice work Ash, just a few more questions and suggestions.

**strict mode**

This change was originally included in the [tech spike](https://github.com/alphagov/govuk_publishing_components/pull/4000) as modules will always run in strict mode, but the change is not included here, just want to check if this is something we need to explore further?

Thinking about the scope of the PR to Drop support for legacy browsers, there could be a couple more changes that are worth including, non-blocking for this PR, just want to highlight them as it seems a good opportunity.

**Remove [ie.js](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/ie.js#L2-L3)**

This file is served to Internet Explorer 8 and below in [layout_for_public](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L82)

This file bundles together `html5shiv-printshiv.js` and `json2` which can also be removed, [json2.js example from the readme](https://github.com/douglascrockford/JSON-js) 😅 

> There is no reason to use this file unless fate compels you to support IE8, which is something that no one should ever have to do again.

**Remove X-UA-Compatible meta tag**
[Removing the X-UA-Compatible meta tag](https://github.com/alphagov/govuk-frontend/pull/4434) is a recommendation in the release notes of v5.0.0 of govuk-frontend, I did not remove this when completing the upgrade work to V5 as we still provided some level of support for legacy browsers.

The change would involve [removing the tag and related comment in layout_for_public](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L90-L91), interestingly it does not appear to be used in the admin template.

Not essential, and can always come back to this as part of another cleanup PR when looking at removing/updating CSS/JS where we support legacy browsers, [IE8 code comments example](https://github.com/search?q=repo%3Aalphagov%2Fgovuk_publishing_components%20IE8&type=code)

AshGDS:
Thanks @MartinJJones 👍 

1. Regarding use strict, because the goal of this is to just drop support for grade X browsers, the extra work of adding `use strict` is not needed as `type=module` achieves what we want in this case.

2. I've removed `ie.js` in the latest commits, good spot, thanks for flagging this.

3. Thanks, I've removed this meta tag in the latest commits.

MartinJJones (approved):
Great work on this Ash! 👍

PR #4112, opened 2024-07-24
Allow disabling `search` input spelling corrections

csutter:
## What
- Add `disable_corrections` parameter to `search` component that can be set to true to disable both `autocorrect` and `autocapitalize` attributes on the search input field [^1]

## Why
As part of our work on site search, we've discovered a pain point where mobile browsers' autocorrection features mess up user queries, especially when it comes to domain specific terminology. For example, searches for "HMRC" are frequently autocorrected to "Hercules", or searches for "SORN" to "sworn".

As our new site search engine used by Search API v2 is much better at implicitly handling misspelled user queries than the previous search engine, we want to allow disabling all mobile browser attempts at second guessing what the user has typed as the end result tends to be worse with two separate spellchecks turning into a game of query telephone.

### Before
image

### After
image
(note that the autocorrect suggestions are still displayed, but not automatically applied)

[^1]: Note that there is little by way of standardisation for this behaviour, and indeed MDN refers to the `autocorrect` attribute as Safari-only, but in practice Chrome for Android honours it as well, see https://issues.chromium.org/issues/40335663

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,8 +7,12 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Allow disabling `search` component input spelling correction ([PR #4112](https://github.com/alphagov/govuk_publishing_components/pull/4112))
+
 ## 40.0.1
-* Add Brakeman to CI jobs ([PR #4108](https://github.com/alphagov/govuk_publishing_components/pull/4108)) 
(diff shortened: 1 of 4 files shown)


maxgds (approved):
I find the logic a bit muddled - set a thing to true to make sure some things get turned off . I can see why it's been done that way, though, so happy to let it through.

PR #4118, opened 2024-07-26
Update cards component design and use one column by default

MartinJJones:
## What
- Update the cards component to use a one column by default
  - The one column layout variation was the winner of the AB test when used on the browse pages, so we will be using the variation on both `collections` and `frontend` in the future. The only place that would still use the three column layout is the account settings page in `signon`
  - Add a new `columns` option which is used to set the number for columns to use
  - Removed the `two_column_layout` option
- Update tests and docs for the new `columns` option
  - Use new `columns` option
  - Add test for three column variation of the card component
- Update spacing for the card component to match the new design
  - Use grid `column-gap` instead of margin
- Align the chevron icon vertically, the card component content now has margin-right set to 30px to ensure it does not overlap with the chevron icon

## Why
Several variations of the card component exist, updating the card component in the gem will allow us to:
- Remove custom styling for the homepage from `frontend`
- Update the browse templates used in `collections` to display a single column layout instead of a grid as this was the most performant variant in our AB test

[Trello](https://trello.com/c/U1DIkmLc/2721-update-card-component-to-support-single-column-layout-and-vertically-aligned-chevron-m)

### One column layout
| Before | After |
| --- | --- |
| card-before | card-after |

### Two/Three column layout
| Before | After |
| --- | --- |
| card-2-3-before | cards-2-3-after |

The new card component design and variations can be previewed on https://components-gem-pr-4118.herokuapp.com/component-guide/cards/preview

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Resolve DartSass mixed declaration deprecations ([PR #4125](https://github.com/alphagov/govuk_publishing_components/pull/4125))
 * Drop support for Ruby 3.1 ([PR #4124](https://github.com/alphagov/govuk_publishing_components/pull/4124))
+* Update cards component design and use one column by default ([PR #4118](https://github.com/alphagov/govuk_publishing_components/pull/4118))
 
 ## 41.0.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
@@ -9,23 +9,12 @@
 .gem-c-cards__list {
   list-style: none;
   padding: 0;
-  // Remove the outermost left and right margin from the internal margin of the list items.
-  margin: 0 govuk-spacing(-3);
-
+  margin: 0;
   display: grid;
-  grid-auto-flow: row;   // Use CSS grid to calculate the number of rows
-  grid-auto-rows: fractions(1); // Set all rows to same fractional height of the complete grid
-  grid-template-columns: fractions(1);
-
-  @include govuk-media-query($from: "tablet") {
-    grid-template-columns: fractions(2);
-  }
+  grid-column-gap: govuk-spacing(6); // Use deprecated grid-column-gap for Grade C browser support
+  column-gap: govuk-spacing(6);
 
   @include govuk-media-query($from: "desktop") {
-    // Note that browsers that don't support CSS grid display the component as one column on all
-    // breakpoints
-    grid-template-columns: fractions(3);
-
     // For browsers that don't support CSS grid, constrain the width to 50% to improve usability
     // especially for screen magnifier users
     width: 50%;
@@ -38,16 +27,21 @@
 }
 
 .gem-c-cards__list--two-column-desktop {
+  @include govuk-media-query($from: "tablet") {
+    grid-auto-rows: fractions(1); // Set all rows to same fractional height of the complete grid
+    grid-template-columns: fractions(2); // Note that browsers that don't support CSS grid display the component as one column on all breakpoints
+  }
+}
+
+.gem-c-cards__list--three-column-desktop {
   @include govuk-media-query($from: "desktop") {
-    grid-template-columns: fractions(2);
+    grid-auto-rows: fractions(1); // Set all rows to same fractional height of the complete grid
+    grid-template-columns: fractions(3); // Note that browsers that don't support CSS grid display the component as one column on all breakpoints
   }
 }
 
 .gem-c-cards__list-item {
   border-top: 1px solid $govuk-border-colour;
-  // We use grid to split the container into column widths, so manage the horizontal spacing with
-  // internal margins.
-  margin: 0 govuk-spacing(3);
   padding: govuk-spacing(1) 0 govuk-spacing(4) 0;
 }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
@@ -49,12 +49,17 @@
   // this wrapper ensures that the clickable area of the card only
   // covers the area of the card containing text so in a grid of cards
   // there is space above and below each link
-  padding: govuk-spacing(2) govuk-spacing(6) govuk-spacing(2) 0;
+  padding: 19px 0 4px;
   position: relative;
 }
 
 .gem-c-cards__sub-heading {
-  margin-bottom: govuk-spacing(2);
+  margin: 0 govuk-spacing(6) govuk-spacing(2) 0;
+
+  // Ensure card content width is constrained to two thirds on desktop
+  @include govuk-media-query($from: "desktop") {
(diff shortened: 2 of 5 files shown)


AshGDS (approved):
LGTM, nice work 👍

MartinJJones:
@AshGDS Sorry, I've fixed one small layout issue relating to the alignment of the chevron icon when using the 2/3 column layout, I've fixed it as part of this commit - https://github.com/alphagov/govuk_publishing_components/pull/4118/commits/41c11cd757eb6ef3c3484237baf6a88a89fcf126

Please can you review the latest change when you get chance

AshGDS (approved):

MartinJJones:
Thanks for the feedback @maxgds 

> * why did you choose to make it a breaking change? Is it because the existing layout will be less common as more things move to a list view?

Yes, this is the main reason, the one column layout variation was the winner of the AB test when used on the browse pages, so we will be using the variation on both `collections` and `frontend` in the future. The only place that would still use the three column layout is the account settings page in Signon. I will include this info in the PR description as well.

> * would it not make more sense to set a "columns" param? eg columns: 3 rather than three_column_layout: true ? You can still default to 1 if you really want it to be the default. It feels to me like a more standard approach.

Good point, I think this would also make it more flexible, I was mainly following the existing convention used in the component (not to say this is right), for example `two_column_layout`. I'll take a look a further look at this.

> You've not updated the description of the component which says it displays a grid. If that won't be the case it really needs an edit.
> 
> I think we can remove the experimental labelling here - you've done research to back up your choices, the older choices were live for some time.

Good spot 👍 , I've updated the description and removed any references to the card component being experimental

MartinJJones:
> would it not make more sense to set a "columns" param? eg columns: 3 rather than three_column_layout: true ? You can still default to 1 if you really want it to be the default. It feels to me like a more standard approach.

@maxgds I've updated this PR to make use of a new `columns` param instead, `three_column_layout` and `two_column_layhout` have been removed, just let me know if any other changes are required, thanks 👍

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
```diff
@@ -9,23 +9,12 @@
...
+  column-gap: govuk-spacing(6);
 
   @include govuk-media-query($from: "desktop") {
-    // Note that browsers that don't support CSS grid display the component as one column on all
-    // breakpoints
-    grid-template-columns: fractions(3);
-
     // For browsers that don't support CSS grid, constrain the width to 50% to improve usability
     // especially for screen magnifier users
     width: 50%;
```
andysellick, line 20: Question - if non-grid browsers restrict the single column default component width to 50% for accessibility, shouldn't that be the default? If someone's using a screen magnifier on a modern browser, they'd get the full width, which is implied by the comment to not be as accessible?

This might be a bigger question than this PR, so maybe not worth agonising over for the initial release.

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
```diff
@@ -9,23 +9,12 @@
...
   display: grid;
-  grid-auto-flow: row;   // Use CSS grid to calculate the number of rows
-  grid-auto-rows: fractions(1); // Set all rows to same fractional height of the complete grid
-  grid-template-columns: fractions(1);
-
-  @include govuk-media-query($from: "tablet") {
-    grid-template-columns: fractions(2);
-  }
+  grid-column-gap: govuk-spacing(6); // Use deprecated grid-column-gap for Grade C browser support
+  column-gap: govuk-spacing(6);
```
andysellick, line 15: Non-blocking: it's probably more efficient in terms of lines of CSS the way it is, but it occurs to me that the grid attributes aren't needed by default, and seem to belong more logically only in the two and three column class variants.

CHANGELOG.md
```diff
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Drop support for Ruby 3.1 ([PR #4124](https://github.com/alphagov/govuk_publishing_components/pull/4124))
+* **BREAKING:** Update cards component design and use one column by default ([PR #4118](https://github.com/alphagov/govuk_publishing_components/pull/4118))
```
andysellick, line 13: Is this actually a breaking change? What's the worst that could happen if it was rolled out without changes in an application?

MartinJJones: Yes I know what you mean, it does feel quite a lot to mark this as a breaking change, but this does seem the correct approach following [semver](https://semver.org/): 

> MAJOR version when you make incompatible API changes

The worst that would happen in this case is that if you are using the card component with a default 3 column layout, this will now change to be a single column layout and you will need to follow the guidance to use a 3 column layout again if required.

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
MartinJJones: Good point, it would definitely be possible to remove the `display: grid` from the `.gem-c-cards__list` and either set this on the `two / three` column classes, or even a separate modifier/utility class specifically for using CSS grid layouts.

Would need to explore this further, can think of lots of pros and cons to each approach.

app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
MartinJJones: Good point, definitely worth thinking about further and how we provide support for these browsers in the future.

When using the one column layout variation on desktop in browsers that do support CSS grid, the content of the card will be restricted to two-thirds (66.667%) for the [heading](https://github.com/alphagov/govuk_publishing_components/pull/4118/files#diff-0f64daead1e205c757c525d6c2800f6716edee2c7046bc590883efc2cf42ef92R59-R61) and [text](https://github.com/alphagov/govuk_publishing_components/pull/4118/files#diff-0f64daead1e205c757c525d6c2800f6716edee2c7046bc590883efc2cf42ef92R109-R112). Although not exactly the same implementation, we are quite close to the 50% width referenced here.

CHANGELOG.md
andysellick: I'd agree with semver, because it's sensible, although in this case it doesn't feel _that_ bad as a thing that would happen? But I'm not going to argue this.

app/views/govuk_publishing_components/components/_cards.html.erb
```diff
@@ -3,9 +3,12 @@
   heading ||= nil
   items ||= nil
   sub_heading_level ||= 3
-  two_column_layout ||= false
+  columns ||= 1
```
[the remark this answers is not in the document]
MartinJJones: I've made this change 👍

spec/components/cards_spec.rb
```diff
@@ -44,7 +44,7 @@ def component_name
 
   it "renders two column list variant" do
     test_data = {
-      two_column_layout: true,
+      columns: 2,
```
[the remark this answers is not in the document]
MartinJJones: Yes!, I've updated the tests to check one column layout is the default, even when using an invalid parameter

PR #4129, opened 2024-08-05
Upgrade to LUX v4.0.25

andysellick:
## What
Upgrades LUX/Real User Monitoring to the latest version. See https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#how-to-update-to-a-new-version for more information and the procedure that was followed.

## Why
This latest version came out today (?) and we were already one version behind, if we leave it too long to upgrade they stop supporting older versions and this won't work anymore.

## Visual Changes
None.


MartinJJones (approved):
Changes look good to me 👍 

I reviewed by installing the lux.js repo and comparing the changes from version 4.0.23 to 4.0.25 and reading the docs.

I couldn't see lux.js available in the repo, I'm happy to take a look at the docs and update them if required.

PR #4135, opened 2024-08-07
Fix the print styles for inverse components

matthillco:
## What
A number of our components can be displayed inverted: white text on a dark background. These components should be set back to non-inverted for print: black text on a white background. 

For components that imply layout separation with their background colours, those aspects should be communicated with borders or other solutions that work better in print.

[Trello card](https://trello.com/c/tYpEnJfI/257-check-print-styles-for-inverse-components)

## Why

- Inverted text can be harder to read on the printed page
- Inverted components waste ink if the browser is set to print background colours
- Inverted components use styles that print with low contrast

### DIFF
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -136,6 +136,12 @@ $gem-guide-border-width: 1px;
   &.component-output {
     padding: 0;
   }
+
+  @include govuk-media-query($media-type: print) {
+    &[class*="-background"] {
+      background-color: unset;
+    }
+  }
 }
 
 .component-guide-preview--simple {
--- a/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
@@ -27,3 +27,15 @@
 .gem-c-inverse-header--padding-top {
   padding-top: govuk-spacing(3);
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-inverse-header {
+    background-color: unset;
+    border-bottom: 2pt solid $govuk-print-text-colour;
+
+    &,
+    & * {
+      color: $govuk-print-text-colour !important; // stylelint-disable-line declaration-no-important
+    }
+  }
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -913,12 +913,6 @@ $after-button-padding-left: govuk-spacing(4);
   font-size: 24px;
 }
 
-@include govuk-media-query($media-type: print) {
-  .gem-c-layout-super-navigation-header__content {
-    display: none;
-  }
-}
-
 // Ensure the total space to the left of the logo for mobile screen sizes is 24px (margin is 15px)
 @include govuk-media-query($until: tablet) {
   .gem-c-layout-super-navigation-header__header-logo--large-navbar {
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/component_guide/application.scss
```diff
@@ -136,6 +136,13 @@ $gem-guide-border-width: 1px;
   &.component-output {
     padding: 0;
   }
+
+  @include govuk-media-query($media-type: print) {
+    &[class*="-background"] {
+      color: $govuk-print-text-colour;
```
andysellick, line 142: Good change, makes sense, but why also set the `color` - `.component-guide-preview` elements shouldn't ever have any child elements or text other than components. I wouldn't want this rule to accidentally impact print styles for components in the guide so it looks like they print correctly but when used in applications actually print differently.

app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
```diff
@@ -27,3 +27,15 @@
 .gem-c-inverse-header--padding-top {
   padding-top: govuk-spacing(3);
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-inverse-header {
+    background-color: unset;
+    border-bottom: 2pt solid $govuk-print-text-colour;
```
andysellick, line 34: What's this new border for?

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
andysellick, line 1: Forgive me if I'm wrong, but this seems like quite a lot of extra CSS for not much difference. Should we have a chat and you can show me the reasoning behind this change?

Before | After
------ | -------
Screenshot 2024-08-13 at 09 16 36 | Screenshot 2024-08-13 at 09 16 41

app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
matthillco: @andysellick This was missing the first time around. It's to implement the separation of components as noted in the PR description: "For components that imply layout and separation with their background colours, those aspects should be communicated with borders or other solutions that work better in print."

Without this border, big inverted headers don't appear to be separated from other page content. The border is now doing the job in print that the background colour does on screen.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
matthillco: @andysellick It would seem these screen shots are wrong, my apologies. The comments in the scss describe why these changes were made, it's to standardise the size and spacing of the crown logo so that it prints the same on every page, including the homepage. The way that element is constructed differs across the layout variants (which does beg the question _Why_?, it's overlay complex for what it is and I'd suggest a refactor) -- these changes make it look the same when printed. I'll update the screenshot with a correct before/after.

matthillco: UPDATE: I've noticed that there's a difference between the size of the crown logo in the emulated print view, vs. the actual print view. I started this work in the emulated view, hence this extra CSS, but you're right it does seem that the print view already made this consistent.

app/assets/stylesheets/component_guide/application.scss
matthillco: I thought this was a good catch all, but you're right, it has the potential to cause problems. I've removed it. Other components that have inverse variants should already be handled in #4073.

app/assets/stylesheets/govuk_publishing_components/components/_inverse-header.scss
andysellick: Okay, thanks.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
matthillco: @andysellick This is now updated as you suggested.

andysellick (approved):

PR #4136, opened 2024-08-07
New print styles for govspeak components

matthillco:
## What
This PR adds/improves print styles for the following govspeak components. Follow the links for examples:

[`attachment`](https://components.publishing.service.gov.uk/component-guide/govspeak#attachment)
[`button`](https://components.publishing.service.gov.uk/component-guide/govspeak#buttons)
[`highlight_answer`](https://components.publishing.service.gov.uk/component-guide/govspeak#highlight_answer)
[`media_player`](https://components.publishing.service.gov.uk/component-guide/govspeak#with_youtube_embed)

## Why

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -31,6 +31,10 @@
       padding-top: 0;
     }
 
+    @include govuk-media-query($media-type: print) {
+      padding-left: 0;
+    }
+
     .attachment-thumb {
       position: relative;
       float: left;
@@ -51,6 +55,10 @@
       svg {
         fill: govuk-colour("mid-grey");
         stroke: govuk-colour("mid-grey");
+      }
+
+      @include govuk-media-query($media-type: print) {
+        margin-left: 0;
       }
     }
 
@@ -109,6 +117,23 @@
       .js-hidden {
         display: none;
       }
+
+      // stylelint-disable declaration-no-important
+      @include govuk-media-query($media-type: print) {
+        a,
+        a:link,
+        a:link:visited {
+          color: $govuk-print-text-colour !important;
+
+          &::after {
+            display: block;
+            margin: 1mm auto;
+            font-weight: normal;
+            font-size: 12pt !important;
+          }
+        }
+      }
+      // stylelint-enable declaration-no-important
     }
   }
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
@@ -24,4 +24,21 @@
   .govuk-button--start {
     @extend .govuk-button--start;
   }
+
+  @include govuk-media-query($media-type: print) {
+    .gem-c-button {
+      box-shadow: none;
+      border: 2pt solid $govuk-print-text-colour;
+
+      &,
+      & * {
+        background-color: unset;
+        color: $govuk-print-text-colour !important; // stylelint-disable-line declaration-no-important
+      }
+
+      &::after {
+        display: none;
+      }
+    }
+  }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
@@ -43,5 +43,16 @@ $highlight-answer-color: govuk-colour("white");
         }
       }
     }
+
+    // stylelint-disable declaration-no-important
+    @include govuk-media-query($media-type: print) {
+      background-color: unset;
+
+      &,
+      & * {
+        color: $govuk-print-text-colour !important;
+      }
+    }
+    // stylelint-enable declaration-no-important
   }
 }
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * New print styles for the tabs component ([PR #4140](https://github.com/alphagov/govuk_publishing_components/pull/4140))
 * New print styles for layout components ([PR #4137](https://github.com/alphagov/govuk_publishing_components/pull/4137))
 * New print styles for step-by-step components ([PR #4138](https://github.com/alphagov/govuk_publishing_components/pull/4138))
+* New print styles for govspeak components ([PR #4136](https://github.com/alphagov/govuk_publishing_components/pull/4136))
 
 ## 41.1.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_media-player.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_media-player.scss
@@ -25,3 +25,9 @@
   height: 100%;
   border: 0;
 }
+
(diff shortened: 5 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -109,6 +109,23 @@
       .js-hidden {
         display: none;
       }
+
+      // stylelint-disable declaration-no-important
```
andysellick, line 113: You can add `// stylelint-disable-line declaration-no-important` at the end of specific lines, e.g.

```CSS
display: none !important; // stylelint-disable-line declaration-no-important
```

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -109,6 +109,23 @@
...
+      @include govuk-media-query($media-type: print) {
```
andysellick, line 114: The thumbnail image (`.attachment-thumb`) doesn't appear in print view, but there's still the gap on the left where it would be. Would it be possible to remove this? (I can't see where this is already hidden).

Screenshot 2024-08-08 at 08 40 58

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
```diff
@@ -24,4 +24,33 @@
   .govuk-button--start {
     @extend .govuk-button--start;
   }
+
+  // stylelint-disable declaration-no-important
+  @include govuk-media-query($media-type: print) {
+    .gem-c-button {
+      display: inline-block;
```
andysellick, line 31: Buttons appear to already have a display of `inline-flex`, what's the benefit of changing it to `inline-block`?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
```diff
@@ -24,4 +24,33 @@
...
+      border: 2pt solid $govuk-print-text-colour;
+
+      &,
+      & * {
+        background-color: unset;
+        color: $govuk-print-text-colour !important;
+      }
+
+      &::before {
+        display: none;
```
andysellick, line 42: What's the before element and why does it need to be hidden?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
```diff
@@ -24,4 +24,33 @@
...
+      }
+
+      &::after {
+        display: none;
```
andysellick, line 46: If the after element is `display: none` why set any other attributes, like position?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
```diff
@@ -43,5 +43,16 @@ $highlight-answer-color: govuk-colour("white");
         }
       }
     }
+
+    // stylelint-disable declaration-no-important
+    @include govuk-media-query($media-type: print) {
+      background-color: unset;
```
andysellick, line 49: Just interested - the background colour is already disabled, I'd guess by the browser but I'm not certain. Is there a benefit to explicitly unsetting it as well?

andysellick (commented):
Changes look good overall, just have a few questions.

It wasn't immediately obvious which bits of govspeak some of these changes applied to, perhaps some links to the specific examples in the component guide, or some before/after screenshots might have helped.

andysellick:
Also I think worth a CHANGELOG entry, although I appreciate there's a few of these PRs now so might be easier to add them as you merge rather than having to keep rebasing.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
matthillco: The occurrence of `// stylelint-disable declaration-no-important` that appears in this file wraps a block of rules that include two instances of `!important`. In this case, I think it's fine.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
matthillco: It seems like some kind of overlay for capturing full width clicks. It's definitely not needed for print, so I've deleted it. Good spot, thanks.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
matthillco: This isn't even needed as there's no `::after` element in the markup. I think this was mistakenly copied from another component. I've deleted it.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
matthillco: If "Print background graphics" is ticked in the printer dialogue settings, the green background will continue to print (looking dark grey on a b&w printer, with black text over it). By explicitly unsetting the background colour here, we override that printer setting and ensure the output will be black on white. 

We could use `none` or `white` or `transparent` or even `initial` values here, but I find the `unset` keyword better communicates the intent (it behaves the same as `initial` in this context).

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
matthillco: Yep this seems to be an error. I've reverted it.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
matthillco: Fixed in latest update.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
andysellick: It looks like it's now set to `inline-flex` in the print styles, but that's already set by default, is that line needed?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
andysellick: Thanks. But it doesn't seem like the before element has any visual styles, so what's the need to hide it on print?

matthillco: Every little helps: The browser still has to compute to appearance of the element before composing the printout, so by hiding it we save it some time.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
matthillco: An oversight sorry, it's now deleted.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_button.scss
andysellick: I think that would apply if it was for everyone's benefit, but since everyone downloads and only a few people print I'd rather save everyone the few bytes of download than some people the rendering benefit when printing.

andysellick (approved):

PR #4137, opened 2024-08-07
New print styles for layout components

matthillco:
## What
New and updated print styles for the `layout-header` and `layout-footer` components.

For the header, we hide some of the elements that would be redundant in print, and force text and border colours to black. The background is also unset to ensure it doesn't print, otherwise it would print with black text on black.

## Why

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -167,3 +167,25 @@
 .gem-c-header__nav {
   clear: both;
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-layout-header .govuk-header__container {
+    border-color: $govuk-print-text-colour;
+
+    .govuk-grid-column-two-thirds {
+      width: auto;
+      float: none;
+    }
+
+    .govuk-header__logo {
+      margin: 0;
+      padding: 0 2mm 2mm 0;
+    }
+
+    .gem-c-header__product-name,
+    .gem-c-environment-tag {
+      color: $govuk-print-text-colour;
+      background-color: unset;
+    }
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * New print styles for the table component ([PR #4139](https://github.com/alphagov/govuk_publishing_components/pull/4139))
 * New print styles for the tabs component ([PR #4140](https://github.com/alphagov/govuk_publishing_components/pull/4140))
+* New print styles for layout components ([PR #4137](https://github.com/alphagov/govuk_publishing_components/pull/4137))
 
 ## 41.1.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
@@ -20,3 +20,11 @@
   // the right column.
   margin-bottom: 20px;
 }
+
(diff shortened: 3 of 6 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
andysellick, line 1: Visually these changes are good but I wanted more explanation in the commit message so I could understand the changes, which seemed overly complicated. I think almost exactly the same change for the layout header component can be done with fewer lines of CSS.

```CSS
// stylelint-disable declaration-no-important
  .gem-c-layout-header .govuk-header__container {
    .gem-c-layout-header__search,
    .gem-c-header__content,
    .govuk-header__navigation-list {
      display: none;
    }

    .govuk-grid-column-two-thirds {
      width: auto;
      float: none;
    }

    .govuk-header__logo {
      margin: 0;
      padding: 0 2mm 2mm 0;
    }

    .gem-c-header__product-name,
    .gem-c-environment-tag {
      color: $govuk-print-text-colour !important;
    }
  }
}
// stylelint-enable declaration-no-important
```

matthillco: OK, I think I understand why you made most of these changes. What's the reason for the removal the `background-color: initial` from the last rule block? That's necessary otherwise the colours will cause conflicts if the browser is set to print background colours. (It could actually be set to `none` or `unset` or `transparent` too, they all would work here.)

Also I've noticed that there should be a rule for forcing the coloured border to print in black, as is our standard. I'll add that in too.

andysellick: I'd removed that because I couldn't see any difference, but now you've explained that, makes sense, fine to keep it in.

It also occurred to me that this bit can be removed and replaced with the `govuk-frontend` don't print class on the relevant elements:

```CSS
    .gem-c-layout-header__search,
    .gem-c-header__content,
    .govuk-header__navigation-list {
      display: none;
    }
```

matthillco: @andysellick I've now implemented your suggested changes, and also forced the border colour to black. There's now no `!important` rules either, so the style-lint overrides were removed.

andysellick: Great, thanks. Just one small oddity - one of the variants has a bit of extra spacing beneath it, can you check?

Screenshot 2024-08-13 at 12 01 59

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -167,3 +167,26 @@
 .gem-c-header__nav {
   clear: both;
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-layout-header .govuk-header__container {
+
```
andysellick, line 173: Nitpick: no need for empty line here.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
matthillco: I've applied some extra `govuk-!-display-none-print` classes to search and navigation items which removes the extra space.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
matthillco: Line removed.

andysellick (approved):

PR #4139, opened 2024-08-07
New print styles for the `table` component

matthillco:
## What
This PR adds/improves print styles for tables:

- sortable tables now print the same as other tables
- the filter element is hidden for print

## Why

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_table.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_table.scss
@@ -125,3 +125,35 @@ $table-row-even-background-colour: govuk-colour("light-grey");
     border: 0;
   }
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .govuk-table--sortable {
+    outline: none;
+
+    * {
+      color: $govuk-print-text-colour !important;
+      background-color: initial !important;
+    }
+
+    .govuk-table__header,
+    .govuk-table__cell {
+      border-bottom: 1px solid $govuk-print-text-colour !important;
+    }
+
+    .govuk-table__header {
+      font-weight: 700;
+      border-right: 0;
+    }
+
+    .app-table__sort-link {
+      padding: 0 !important;
+
+      &::before,
+      &::after {
+        display: none !important;
+      }
+    }
+  }
+}
+// stylelint-enable declaration-no-important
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -125,3 +125,39 @@ $table-row-even-background-colour: govuk-colour("light-grey");
...
+
+      &::before,
+      &::after {
+        display: none !important;
+      }
+    }
+  }
+
+  .js-gem-c-table__filter {
+    display: none !important;
```
andysellick, line 160: Doesn't seem like `important` is needed? Doesn't look like this class has any other declarations set anywhere.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
```diff
@@ -125,3 +125,39 @@ $table-row-even-background-colour: govuk-colour("light-grey");
...
 }
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .govuk-table--sortable {
+    outline: none;
+
+    * {
+      color: $govuk-print-text-colour !important;
+      background-color: initial !important;
```
andysellick, line 136: Is the `important` needed here? Just tested it without and it seemed okay.

andysellick (commented):
Looks good, worth a CHANGELOG entry before merging.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
matthillco: I'm not seeing the same thing: if I remove `!important` then the old blue/inverted headings reappear and the zebra striping on the rows re-appears. (This is when the browser is set to print background graphics, did you test it that way?)

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
matthillco: I've removed this declaration in favour of the `govuk-!-display-none-print` class on the filter element in the view.

app/assets/stylesheets/govuk_publishing_components/components/_table.scss
andysellick: Ah yes, I did not have the option to print background graphics enabled, fair enough.

andysellick (approved):
Suggest merging the two commits together.

PR #4140, opened 2024-08-07
New print styles for the `tabs` component

matthillco:
## What
This PR adds/improves print styles for tabs:

- display all tabs when printed, don't replicate the shown/hidden screen states
- linearise the tabs into a single column

## Why

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
@@ -31,3 +31,35 @@ li.govuk-tabs__list-item { // stylelint-disable-line selector-no-qualifying-type
     }
   }
 }
+
+@include govuk-media-query($media-type: print) {
+  .govuk-tabs__title {
+    display: block !important; // stylelint-disable-line declaration-no-important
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
```diff
@@ -31,3 +31,37 @@ li.govuk-tabs__list-item { // stylelint-disable-line selector-no-qualifying-type
     }
   }
 }
+
+// stylelint-disable declaration-no-important
```
andysellick, line 35: Can this stylelint rule be applied specifically to the line where `important` is used, rather than the entire block?

andysellick (commented):
Looks good. Probably worth a CHANGELOG entry.

app/assets/stylesheets/govuk_publishing_components/components/_tabs.scss
matthillco: OK, done. New changelog entry added too.

andysellick (approved):
Suggest merging the commits.

PR #4141, opened 2024-08-07
Update card component styles

MartinJJones:
## What
When using a single column layout, the top border will now only be visible when a heading is used.

Updated `gem-c-cards__heading` to use a responsive margin-bottom.

## Why
To match the intended design of the card component when using a single column layout with a heading.

[Trello card](https://trello.com/c/F3N53ubq/1189-update-border-top-for-card-component)

### Card top border

| Before | After |
| --- | --- |
| cards-desktop-before | cards-desktop-after |

### Card heading margin now responsive

| Before | After |
| --- | --- |
| card-mobile-heading-before | card-mobile-heading-after |

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/cards.yml
+++ b/app/views/govuk_publishing_components/components/docs/cards.yml
@@ -40,34 +40,30 @@ examples:
             text: Citizenship and living in the&nbsp;UK
             path: http://www.gov.uk
           description: Voting, community participation, life in the UK, international projects
+  one_column_layout_without_heading:
+    description: No border at the top of the list of cards is present when using a one column layout without a heading. The first card item will still include a top border when using a 2 or 3 column layout as shown in the examples below.
+    data:
+      items:
         - link:
-            text: Crime, justice and the&nbsp;law
+            text: Benefits
             path: http://www.gov.uk
-          description: Legal processes, courts and the police
+          description: Includes eligibility, appeals, tax credits and Universal Credit
         - link:
-            text: Disabled people
+            text: Births, deaths, marriages and&nbsp;care
             path: http://www.gov.uk
-          description: Includes carers, your rights, benefits and the Equality Act
+          description: Parenting, civil partnerships, divorce and Lasting Power of Attorney
         - link:
-            text: Driving and transport
+            text: Business and self-employed
             path: http://www.gov.uk
-          description: Includes vehicle tax, MOT and driving licences
+          description: Tools and guidance for businesses
         - link:
-            text: Education and learning
+            text: Childcare and parenting
             path: http://www.gov.uk
-          description: Includes student loans, admissions and apprenticeships
+          description: Includes giving birth, fostering, adopting, benefits for children, childcare and schools
         - link:
-            text: Employing people
+            text: Citizenship and living in the&nbsp;UK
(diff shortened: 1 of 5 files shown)


andysellick (commented):
This works as expected but I it feels a little over-complicated - there's some slightly complex CSS here and a situation where a border is set, unset and then has to be reset.

I've had a quick play to see if there's a clearer way to do this using additional classes added at the template level, but feel free to point out where I've gone wrong: https://github.com/alphagov/govuk_publishing_components/pull/4143

I also think there should be an explanation of this change in the documentation, to be really clear about what happens here with all the variants of heading/no heading/columns (probably only needs one example with a description).

MartinJJones:
> This works as expected but I it feels a little over-complicated - there's some slightly complex CSS here and a situation where a border is set, unset and then has to be reset.
>
> I've had a quick play to see if there's a clearer way to do this using additional classes added at the template level, but feel free to point out where I've gone wrong: https://github.com/alphagov/govuk_publishing_components/pull/4143

Yes agreed, I did try to do something similar by adding a class to the template, by checking the 1st item in the card item in the list, adding it to the heading itself seems much cleaner, thanks!

> I also think there should be an explanation of this change in the documentation, to be really clear about what happens here with all the variants of heading/no heading/columns (probably only needs one example with a description).

Good point, I've added an example in the documentation

app/views/govuk_publishing_components/components/docs/cards.yml
```diff
@@ -68,6 +68,58 @@ examples:
             text: Housing and local services
             path: http://www.gov.uk
           description: Owning or renting and council services
+  one_column_layout_without_heading:
+    description: No border at the top of the list of cards is present when using a one column layout without a heading.
```
andysellick, line 72: I think it's worth noting that a two column layout with a heading does have borders at the top of the list items.

app/views/govuk_publishing_components/components/docs/cards.yml
```diff
@@ -68,6 +68,58 @@ examples:
...
+            path: http://www.gov.uk
+          description: Includes pay, contracts, hiring and redundancies
+        - link:
+            text: Environment and countryside
+            path: http://www.gov.uk
+          description: Includes flooding, recycling and wildlife
+        - link:
+            text: Housing and local services
+            path: http://www.gov.uk
+          description: Owning or renting and council services
```
andysellick, line 122: Would it be reasonable to trim the data in these examples down a bit? There's so many rows here the example doesn't fit onto my screen, making it harder to navigate this page.

andysellick (commented):
Much clearer, thanks. Just a couple of comments, nothing major.

app/views/govuk_publishing_components/components/docs/cards.yml
MartinJJones: Yeah makes sense to me, I've made this change, looks much better in the component guide.

app/views/govuk_publishing_components/components/docs/cards.yml
MartinJJones: Done 👍

andysellick (approved):
Changes look good, thanks for putting up with my nitpicking 👍

PR #4142, opened 2024-08-08
Add dynamic search autocomplete component (experimental) 

alexbowen:
Taking this out of draft status and its becoming reviewable although work continuing:

TODO
- more javascript tests
- ~add more config options (e.g relative/absolute result display, submit on enter)~
- ~add aria live region to rendered component~
- add analytics tracking
- add visual tests

## What
The autocomplete component lets users select from a dynamic list of suggestions to populate the value of a search input element `govuk_publishing_components/components/search`. If suggestions come from a static source then enhancing a server rendered HTML select element is a better option [govuk accessible autocomplete](https://github.com/alphagov/accessible-autocomplete)

Inspiration for this came from:

[govuk accessible autocomplete](https://github.com/alphagov/accessible-autocomplete)
[Adam Silver Blog](https://adamsilver.io/blog/building-an-accessible-autocomplete-control/)

The configuration options are deliberatley minimal to serve immediate requirements but additional options should be easy to add. For that would be nice if argument spread syntax (ES6) could be used and would make adding options easier. As currently the JS is written in ES5 style.

## Why
A component is needed that displays dynamic suggestions from a http endpoint based on what the user types. The current use case is suggestions from Google's Vertex solution that is being integrated on homepage and search results page(s)

## Visual Changes

Screenshot 2024-08-16 at 08 40 31

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
@@ -552,8 +552,6 @@ window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
 
     var renderResultHtml = function (result, props, inputVal) {
       var index = result.toLowerCase().indexOf(inputVal.toLowerCase())
-
-      console.log('props', props)
       return `<li ${props}>${suggestionIcon ? searchIcon : ''}<span class='govuk-!-font-weight-bold'>${result.substring(0, index)}</span>${result.substring(index, index + inputVal.length)}<span class='govuk-!-font-weight-bold'>${result.substring(index + inputVal.length, result.length)}<span><div class='gem-c-autocomplete__result--border'></div></li>`
     }
 
--- a/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
@@ -68,7 +68,7 @@ window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
   // string in the format: `key1='value1' key2='value2'` for easy use in an HTML string.
   function Props (index, selectedIndex, baseClass) {
     this.id = `${baseClass}-result-${index}`
-    this.class = `${baseClass}-result`
+    this.class = `${baseClass}__result`
     this['data-result-index'] = index
     this.role = 'option'
     if (index === selectedIndex) {
@@ -330,7 +330,7 @@ window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
     this.resetPosition = true
     this.positionResults = positionResults
     this.root = typeof root === 'string' ? document.querySelector(root) : root
-    this.root.insertAdjacentHTML('beforeend', '<ul class="gem-c-autocomplete-result-list govuk-body" role="listbox"></ul><div aria-atomic="true" aria-live="polite" role="status" class="govuk-visually-hidden">No results.</div>')
+    this.root.insertAdjacentHTML('beforeend', '<ul class="gem-c-autocomplete__result-list govuk-body" role="listbox"></ul><div aria-atomic="true" aria-live="polite" role="status" class="govuk-visually-hidden">No results.</div>')
     this.input = this.root.querySelector('input')
     this.assistiveHint = this.root.querySelector('.assistive-hint')
     this.resultList = this.root.querySelector('ul')
@@ -472,7 +472,7 @@ window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
 
     this.input.setAttribute(
       'aria-activedescendant',
-      selectedIndex > -1 ? `${this.baseClass}-result-${selectedIndex}` : ''
+      selectedIndex > -1 ? `${this.baseClass}__result-${selectedIndex}` : ''
     )
 
     if (this.resetPosition) {
@@ -548,11 +548,13 @@ window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
     var positionResults = this.positionResults
     var submitOnSelect = this.submitOnSelect
 
-    var searchIcon = '<span class="search-icon"><svg width="20" height="20" viewBox="0 0 27 27" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false"><circle cx="12.0161" cy="11.0161" r="8.51613" stroke="currentColor" stroke-width="3" /><line x1="17.8668" y1="17.3587" x2="26.4475" y2="25.9393" stroke="currentColor" stroke-width="3" /></svg></span>'
+    var searchIcon = '<span class="gem-c-autocomplete__result--search-icon"><svg width="20" height="20" viewBox="0 0 27 27" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false"><circle cx="12.0161" cy="11.0161" r="8.51613" stroke="currentColor" stroke-width="3" /><line x1="17.8668" y1="17.3587" x2="26.4475" y2="25.9393" stroke="currentColor" stroke-width="3" /></svg></span>'
 
     var renderResultHtml = function (result, props, inputVal) {
       var index = result.toLowerCase().indexOf(inputVal.toLowerCase())
-      return `<li ${props}>${suggestionIcon ? searchIcon : ''}<span class='govuk-!-font-weight-bold'>${result.substring(0, index)}</span>${result.substring(index, index + inputVal.length)}<span class='govuk-!-font-weight-bold'>${result.substring(index + inputVal.length, result.length)}<span><div class='border'></div></li>`
+
+      console.log('props', props)
+      return `<li ${props}>${suggestionIcon ? searchIcon : ''}<span class='govuk-!-font-weight-bold'>${result.substring(0, index)}</span>${result.substring(index, index + inputVal.length)}<span class='govuk-!-font-weight-bold'>${result.substring(index + inputVal.length, result.length)}<span><div class='gem-c-autocomplete__result--border'></div></li>`
     }
 
     new Autocomplete(this.$module, // eslint-disable-line no-new
--- a/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
@@ -1,0 +1,569 @@
+//= require ../vendor/polyfills/closest.js
+/* global fetch */
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
+
+(function (Modules) {
+  'use strict'
+
+  var util = {
+    getRelativePosition: function (element1, element2) {
+      var position1 = element1.getBoundingClientRect()
+      var position2 = element2.getBoundingClientRect()
+
+      var positionAbove =
+        /* 1 */ position1.bottom + position2.height > window.innerHeight &&
+        /* 2 */ window.innerHeight - position1.bottom < position1.top &&
+        /* 3 */ window.pageYOffset + position1.top - position2.height > 0
+
+      return positionAbove ? 'above' : 'below'
+    },
+    debounce: function (func, wait, immediate) {
+      let timeout
+
+      return function executedFunction () {
+        var context = this
+        var args = arguments
+
+        var later = function () {
+          timeout = null
+          if (!immediate) func.apply(context, args)
+        }
+
+        var callNow = immediate && !timeout
+        clearTimeout(timeout)
+        timeout = setTimeout(later, wait)
+
+        if (callNow) func.apply(context, args)
+      }
+    },
+    getAriaLabel: function (labelStr) {
+      if (labelStr?.length) {
+        var isLabelId = labelStr.startsWith('#')
+
+        return {
+          attribute: isLabelId ? 'aria-labelledby' : 'aria-label',
+          content: isLabelId ? labelStr.substring(1) : labelStr
+        }
+      }
+    },
+    isPromise: function (value) {
+      return Boolean(value && typeof value.then === 'function')
+    },
+    isFunction: function (value) {
+      return Boolean(typeof value === 'function')
+    },
+    isArray: function (string) {
+      try {
+        return Array.isArray(JSON.parse(string))
+      } catch (e) {
+        return false
+      }
+    }
+  }
+
+  // Creates a props object with overridden toString function. toString returns an attributes
+  // string in the format: `key1='value1' key2='value2'` for easy use in an HTML string.
+  function Props (index, selectedIndex, baseClass) {
+    this.id = `${baseClass}-result-${index}`
+    this.class = `${baseClass}-result`
+    this['data-result-index'] = index
+    this.role = 'option'
+    if (index === selectedIndex) {
+      this['aria-selected'] = 'true'
+    }
+  }
+
+  Props.prototype.toString = function () {
+    return Object.keys(this).reduce(
+      (str, key) => `${str} ${key}='${this[key]}'`,
+      ''
+    )
+  }
+
+  function AutocompleteCore (
+    search,
+    autoSelect,
+    setValue,
+    setAttribute,
+    onUpdate,
+    onSubmit,
+    onShow,
+    onHide,
+    onLoading,
+    onLoaded,
+    submitOnSelect
+  ) {
+    this.value = ''
+    this.searchCounter = 0
+    this.results = []
+    this.selectedIndex = -1
+    this.selectedResult = null
+
+    switch (true) {
+      case util.isFunction(search):
+        this.search = (value) => Promise.resolve(search(value))
+        break
+      case util.isArray(search):
+        this.search = () => Promise.resolve(JSON.parse(search))
+        break
+      default:
+        this.search = async (value) => {
+          var url = new URL(search)
+          url.searchParams.append('q', value)
+
+          var response = await fetch(url)
+          var completes = await response.json()
+          return completes
+        }
+        break
+    }
+
+    this.autoSelect = autoSelect || false
+    this.setValue = setValue || function () { }
+    this.setAttribute = setAttribute || function () { }
+    this.onUpdate = onUpdate || function () { }
+    this.onSubmit = onSubmit || function () { }
+    this.onShow = onShow
+    this.onHide = onHide || function () { }
+    this.onLoading = onLoading || function () { }
+    this.onLoaded = onLoaded || function () { }
+    this.submitOnSelect = submitOnSelect || false
+  }
+
+  AutocompleteCore.prototype.destroy = function () {
+    this.search = null
+    this.setValue = null
+    this.setAttribute = null
+    this.onUpdate = null
+    this.onSubmit = null
+    this.onShow = null
+    this.onHide = null
+    this.onLoading = null
+    this.onLoaded = null
+  }
+
+  AutocompleteCore.prototype.handleInput = function (event) {
+    var value = event.target.value
+    this.updateResults(value)
+    this.value = value
+  }
+
+  AutocompleteCore.prototype.handleKeyUp = function (event) {
+    var activeEl = event.target.getAttribute('aria-activedescendant')
+    if (activeEl) {
+      event.target.value = document.getElementById(activeEl).innerText
+    }
+  }
+
+  AutocompleteCore.prototype.handleKeyDown = function (event) {
+    const keyCodes = {
+      9: 'tab',
+      13: 'enter',
+      27: 'escape',
+      32: 'space',
+      38: 'up',
+      40: 'down'
+    }
+
+    switch (keyCodes[event.keyCode]) {
+      case 'up':
+      case 'down': {
+        const selectedIndex =
+        event.keyCode === 38 ? this.selectedIndex - 1 : this.selectedIndex + 1
+        event.preventDefault()
+        this.handleArrows(selectedIndex)
+        break
+      }
+      case 'tab': {
+        this.selectResult()
+        break
+      }
+      case 'enter':
+      case 'space': {
+        if (event.keyCode === 32) {
+          event.preventDefault()
+        }
+        this.selectedResult =
+          this.results[this.selectedIndex] || this.selectedResult
+        this.selectResult()
+
+        if (this.submitOnSelect && event.target.closest('form')) {
+          event.target.closest('form').submit()
+        }
+        break
+      }
+      case 'escape': {
+        this.hideResults()
+        this.setValue()
+        break
+      }
+    }
+  }
+
+  AutocompleteCore.prototype.handleFocus = function (event) {
+    var value = event.target.value
+    this.updateResults(value)
+    this.value = value
+  }
+
+  AutocompleteCore.prototype.handleBlur = function () {
+    this.hideResults()
+  }
+
+  AutocompleteCore.prototype.handleResultMouseDown = function (event) {
+    event.preventDefault()
+  }
+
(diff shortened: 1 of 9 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_autocomplete.scss
```diff
@@ -0,0 +1,49 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-autocomplete {
+  position: relative;
+
+  > div {
+    margin-bottom: 0;
+  }
+  
+  .gem-c-autocomplete-result-list {
```
andysellick, line 10: We use BEM for our classes, as you've noticed. This looks like an element class, so I think it should be `.gem-c-autocomplete__result-list`?

There's a few other classes as well - I'm guessing `.gem-c-autocomplete-result` should be `.gem-c-autocomplete__result` and `.border` should be `.gem-c-autocomplete__border` - or is that one a modifier class and should be `.gem-c-autocomplete--border`?

app/assets/stylesheets/govuk_publishing_components/components/_autocomplete.scss
```diff
@@ -0,0 +1,49 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-autocomplete {
+  position: relative;
+
+  > div {
```
andysellick, line 6: We try to avoid styling elements - is there a class name you could create for this?

app/assets/stylesheets/govuk_publishing_components/components/_autocomplete.scss
```diff
@@ -0,0 +1,49 @@
...
+    padding: 0;
+    box-sizing: border-box;
+    overflow-y: auto;
+    -webkit-overflow-scrolling: touch;
+    background: govuk-colour("white");
+    list-style: none;
+    position: relative;
+  }
+  
+  .gem-c-autocomplete-result {
```
andysellick, line 22: Using BEM means these classes shouldn't need nesting (something we try to avoid for inheritance reasons) so I think you should be able to structure this file more like:

```CSS
.gem-c-autocomplete {
  ...
}

.gem-c-autocomplete__result {
  ...
}
```

app/assets/stylesheets/govuk_publishing_components/components/_autocomplete.scss
andysellick: Also `.search-icon`

andysellick (commented):
Just doing an initial first pass review, focussing only on the Sass file, nothing else.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+name: Autocomplete (experimental)
+description: The autocomplete component lets users select from a dynamic list of suggestions to populate the value of a text input element.
+govuk_frontend_components:
+  - autocomplete
+body: |
+  If suggestions come from a static source then enchancing a server rendered HTML select element is a better option.
+
+  This component is currently experimental because more research is needed to validate it pending autocomplete implementation documented in the [GDS roadmap](https://design-system.service.gov.uk/community/upcoming-components-patterns/).  If using this component, [please feed back any research findings to the Design System team](https://design-system.service.gov.uk/get-in-touch/).
+accessibility_criteria: |
+  * Be focusable with a keyboard
```
andysellick, line 11: I think this list should have a sentence at the top to make it more readable, something like:

```
The component must:

* Be focusable with a keyboard
* Indicate when it has keyboard focus
```

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+  * Inform the user if a match is pre-selected
+  * Enable the user to confirm the selected match
+  * Inform the user when a match is confirmed
+  * Return focus to the editable field when a selected match is confirmed
+uses_component_wrapper_helper: true
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    description: url to get suggestion data from or an array of suggestions (required). Query parameter of ?q={inputValue} will be appended The array option exists mainly for the purposes of demonstration and testing where an HTTP endpoint is not ideal. If you have a static array of suggestions consider another approach that enhances an HTML select element.
```
andysellick, line 31: A quirk of our component guide is that the `description` for the `default` example isn't displayed. If this information is useful I'd suggest moving it into the `body` field (see example: https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/accordion.yml#L5)

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+    data:
+      source: ['prime minister','deputy prime minister','contact prime minister','email prime minister','last prime minister']
+  custom_input_element:
```
andysellick, line 34: Should this be an option on the component? It seems like it would be possible to pass anything here, even something like a simple paragraph - that wouldn't be then rendered as an autocomplete. I suspect you've had to do extra work to make this option possible. 

I'd recommend not doing that (it feels inherently brittle, as presumably your JS has to assume certain things about what might be in this block) but instead give the component options to display whatever is required here - the example talks about a custom input element, what are the requirements? e.g. if it's to be able to set the label text, or the input id, or having the search button at the end - make those options that get passed to the component instead.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+            inline_label: false,
+            label_text: "Search"
+          } %>
+  number_of_suggestions_to_display:
+    description: The number of suggestions to be displayed to the user. If not specified the default is 5 suggestions. Number of results if they are truly dynamic should ideally limited at the backend with a limit url param.
+    data:
+      source: ['prime minister','deputy prime minister','contact prime minister','email prime minister','last prime minister']
+      display_number_suggestions: 3
+  position_results:
+    description: Wether results are positioned relative or absolute to surrounding content. Defaults to relative
```
andysellick, line 50: Nitpick: `Wether` and missing a full stop at the end of a sentence.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+          <%= render "govuk_publishing_components/components/search", {
+            inline_label: false,
+            label_text: "Search"
+          } %>
+  number_of_suggestions_to_display:
+    description: The number of suggestions to be displayed to the user. If not specified the default is 5 suggestions. Number of results if they are truly dynamic should ideally limited at the backend with a limit url param.
+    data:
+      source: ['prime minister','deputy prime minister','contact prime minister','email prime minister','last prime minister']
+      display_number_suggestions: 3
+  position_results:
```
andysellick, line 49: Suggestion: `relative` and `absolute` in this context make sense to me as a frontend developer, but this language could be clearer, e.g.

`description: Controls whether autocomplete suggestions are positioned relative to the input (so that content below is pushed down when they appear) or absolute (content below is overlapped when they appear).`

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,70 @@
...
+  suggestion_icon:
+    description: Display each suggestion with left aligned search icon
+    data:
+      source: ['prime minister','deputy prime minister','contact prime minister','email prime minister','last prime minister']
+      suggestion_icon: true
+  other_configuration_options:
+    description: |
+      Options that cannot be demoed in the context of the example below.
+
+      * **submit_form_on_select** Submits the surrounding form when a suggestion is selected. Defaults to false. Note that this isnt neccessarily the best behaviour in accessibilty terms.
```
andysellick, line 63: Question: if this isn't good accessibility practice why include it?

app/views/govuk_publishing_components/components/docs/autocomplete.yml
andysellick, line 1: This PR is still in progress so this may be intentional, but I've noticed that in the component guide all options in the autocomplete are shown regardless of whether they match what I type or not - just to check that you're aware of this.

andysellick (commented):
Just doing an initial first pass review, focussing only on the YML file, nothing else.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
alexbowen: contentious point...

app/views/govuk_publishing_components/components/docs/autocomplete.yml
alexbowen: > This PR is still in progress so this may be intentional, but I've noticed that in the component guide all options in the autocomplete are shown regardless of whether they match what I type or not - just to check that you're aware of this.

It is intentional. I have added to the .yml file `The assumption is that the backend that the query string is supplied to does the matching and returns only results that are appropriate based on that query string`. For the purposes of the demo pages a static array is used to avoid needing an http endpoint, that would only supply static data/suggestions anyway.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
andysellick: I've considered this and I don't think allowing a block to be passed to this component makes any sense. If there's a requirement to have different input types I think that should be controlled through options, not through passing a block - there's too much risk involved.

app/views/govuk_publishing_components/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
+add_gem_component_stylesheet("autocomplete")
+display_number_suggestions ||= 5
+suggestion_icon ||= ''
+request_debounce_time ||= ''
+source ||= nil
+position_results ||= nil
+submit_form_on_select ||= false
+idPostfix ||= "#{SecureRandom.hex(4)}"
```
andysellick, line 9: What is `idPostfix` used for? If this is an option that can be passed to the component, it should have an example and explanation in the YML file.

app/views/govuk_publishing_components/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,42 @@
...
+component_helper.add_data_attribute({ source: source })
+component_helper.add_data_attribute({ id_postfix: idPostfix })
+
+component_helper.add_data_attribute({ display_number_suggestions: display_number_suggestions })
+component_helper.add_data_attribute({ suggestion_icon: suggestion_icon })
+component_helper.add_data_attribute({ submit_on_select: submit_form_on_select })
+
+if position_results.present?
+  component_helper.add_data_attribute({ position_results: 'absolute' })
+end
```
andysellick, line 23: I think these three lines could be written as

```Ruby
component_helper.add_data_attribute({ position_results: 'absolute' }) if position_results.present?
```

(same for `request_debounce_time`)

app/views/govuk_publishing_components/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,42 @@
...
+component_helper.add_data_attribute({ module: "gem-autocomplete" })
+component_helper.add_data_attribute({ source: source })
+component_helper.add_data_attribute({ id_postfix: idPostfix })
+
+component_helper.add_data_attribute({ display_number_suggestions: display_number_suggestions })
+component_helper.add_data_attribute({ suggestion_icon: suggestion_icon })
+component_helper.add_data_attribute({ submit_on_select: submit_form_on_select })
+
+if position_results.present?
+  component_helper.add_data_attribute({ position_results: 'absolute' })
```
andysellick, line 22: I haven't looked at the JS in detail yet but it looks like this data attribute is read by the JS then used to configure the autocomplete. Have you considered instead adding this as a modifier class and controlling the position of the results list entirely using CSS?

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,78 @@
+
+name: Autocomplete (experimental)
+description: The autocomplete component lets users select from a dynamic list of suggestions to populate the value of a search input element.
+body: |
+  This component does not perform any matching of user input to refine suggestions. The assumption is that the backend that the query string is supplied to does the matching and returns only results that are appropriate.
```
andysellick, line 5: Nitpick: sentence structure here is a bit confusing... is it trying to say `The assumption is that the backend returns only results that match the query string.`?

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  A static list of suggestions are used for these examples (see source option below) but if suggestions come from a static source then enhancing a server rendered HTML select element is a better option.
```
andysellick, line 7: I understand this but I think it might be confusing for developers looking to use this component in future as it implies that this option is available here, when it's actually saying "you should do something entirely different that doesn't exist yet". It might be easiest to remove this line - this file documents this component, it doesn't need to go into other autocomplete possibilities.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,78 @@
...
+
+  This component is currently experimental because more research is needed to validate it pending autocomplete implementation documented in the [GDS roadmap](https://design-system.service.gov.uk/community/upcoming-components-patterns/).  If using this component, [please feed back any research findings to the Design System team](https://design-system.service.gov.uk/get-in-touch/).
```
andysellick, line 9: Nitpick: this links to the Design System roadmap, not the GDS roadmap. Also "this component" seems to refer to the component on this page, which the Design System team haven't been involved with, so we shouldn't be encouraging people to feedback research findings in this way.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
andysellick: This `description` field is still present.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,78 @@
...
+uses_component_wrapper_helper: true
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    description: (Required) Url to get suggestion data from or an array of suggestions. Query parameter of ?q={inputValue} will be appended The array option exists mainly for the purposes of demonstration and testing where an HTTP endpoint is not ideal. If you have a static array of suggestions consider another approach that enhances an HTML select element.
+    data:
+      source: ['prime minister','deputy prime minister','contact prime minister','email prime minister','last prime minister']
+  source:
+    description: (Required) Url to get suggestion data from or an array of suggestions. Query parameter of ?q={inputValue} will be appended The array option exists mainly for the purposes of demonstration and testing where an HTTP endpoint is not ideal. If you have a static array of suggestions consider another approach that enhances an HTML select element.
```
andysellick, line 37: I'm not sure `source` is worth explaining as a separate example, when it's a required option? This should be already explained in the `body`.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
```diff
@@ -0,0 +1,78 @@
...
+      block: |
+        <!-- example content -->
+        <%= render "govuk_publishing_components/components/input", {
+          label: {
+            text: "Enter some text"
+          },
+          name: "query"
+        } %>
+  number_of_suggestions_to_display:
+    description: The number of suggestions to be displayed to the user. If not specified the default is 5 suggestions. Number of results if they are truly dynamic should ideally limited at the backend with a limit url param.
```
andysellick, line 53: `Number of results if they are truly dynamic should ideally limited at the backend with a limit url param.` - not quite sure what the purpose of this sentence is - is it advising how the backend API should be built? This doesn't seem relevant to this option on the component.

andysellick (commented):
Some more thoughts on the Sass, template and YML file only.

app/views/govuk_publishing_components/components/docs/autocomplete.yml
alexbowen: ok - its no bother to remove the block option. if this component gets taken forward in the future and it becomes a requirement, someone can implement it in a better thought out way than maybe i have :D

app/views/govuk_publishing_components/components/_autocomplete.html.erb
alexbowen: no it cant be passed to component (so i will remove the ||). it is used to append to ids for elements within autocomplete so they are unique (in case another autocomplete is on page)

app/views/govuk_publishing_components/components/_autocomplete.html.erb
alexbowen: good point, sorted i think

app/views/govuk_publishing_components/components/_autocomplete.html.erb
```diff
@@ -0,0 +1,42 @@
+<%
+add_gem_component_stylesheet("autocomplete")
+display_number_suggestions ||= 5
+suggestion_icon ||= ''
```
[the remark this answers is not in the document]
alexbowen: i think similar to the position thing - this should be handled as a modifier css class. it is currently inlined in the JS atm i cant remember why i initially did it this way, i was probably still getting to grips with the nitty gritty of publishing components, but seems a misuse of JS

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
andysellick, line 1: General comment about this file - I'm finding it hard to follow the order of the code. Would it be possible to reorder the functions to some degree based on the order in which they're called? e.g. code initialises, autocomplete initialises, user interaction functions.

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,561 @@
...
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
+
+(function (Modules) {
+  'use strict'
+
+  var util = {
+    getRelativePosition: function (element1, element2) {
```
andysellick, line 12: This function looks like it might be used to determine if the autocomplete result list should be appended above or below the text input, is that right? I'm not seeing that behaviour in the live example, is this doing something else?

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,561 @@
...
+    onLoaded,
+    submitOnSelect
+  ) {
+    this.value = ''
+    this.searchCounter = 0
+    this.results = []
+    this.selectedIndex = -1
+    this.selectedResult = null
+
+    switch (true) {
```
andysellick, line 105: Haven't seen a switch statement based on `true` before - would this be cleaner as an if statement? e.g.

```JS
if (util.isFunction(search)) {
  //
} else if (util.isArray(search)) {
  //
} else {
  //
}
```

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,561 @@
...
+        }
+        break
+    }
+
+    this.autoSelect = autoSelect || false
+    this.setValue = setValue || function () { }
+    this.setAttribute = setAttribute || function () { }
+    this.onUpdate = onUpdate || function () { }
+    this.onSubmit = onSubmit || function () { }
+    this.onShow = onShow
```
andysellick, line 129: Why do all the variables in this section have fallback values apart from `onShow`?

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,561 @@
...
+    this.onHide = onHide || function () { }
+    this.onLoading = onLoading || function () { }
+    this.onLoaded = onLoaded || function () { }
+    this.submitOnSelect = submitOnSelect || false
+  }
+
+  AutocompleteCore.prototype.destroy = function () {
```
andysellick, line 136: Why are there two `destroy` functions - is it not possible to combine them into one? And when is this called - as far as I can see the autocomplete doesn't get destroyed during normal user operation.

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
andysellick, line 1: Think I've found a bug... if you type anything into the autocomplete, say `prime` then press spacebar a few times quickly (doesn't seem to happen if you do it slowly) a `.` character is inserted into the input. Don't know why.

Screenshot 2024-08-22 at 09 26 28

andysellick (commented):
This is a very high level, quick pass of the JavaScript file - I haven't gone in depth yet, but hopefully these initial comments will be helpful.

This is a lot of code to review. Something that I think this PR should justify is why this approach has been taken - and why we aren't importing the JS from the existing Design System Accessible Autocomplete.

app/views/govuk_publishing_components/components/_autocomplete.html.erb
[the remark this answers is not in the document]
alexbowen: i have left the search icon stuff as is. im not sure what changing it to a pure css icon acheives. especially when the icon wont be exactly the same as the svg that already lives in this repo. can use css svgs as it violates CRS stuff and errors

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,461 @@
...
+/* global fetch */
+
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {}
+window.GOVUK.Modules.GovukAutocomplete = window.GOVUKFrontend.Autocomplete;
+
+(function (Modules) {
+  'use strict'
+
+  function GemAutocomplete ($module) {
```
kevindew, line 11: I notice that across the code in here we've got some parts embracing es6 (asnyc, const, arrow functions) and other bits that are es5 (function.prototype classes, var) it'd be good to just consistently write it in es6 code.

kevindew (commented):
Nice job - this looks super cool 😎 

I think the JS is in quite severe need of some refactoring - which I'd like to hope would make testing it easier. It's currently very hard to understand. My other concern is it looks like this is somewhat caught between trying to handle lots of things while also being, in all practical terms, limited to the desired Search API functionality - I think it'd be most pragmatic to get this as simple as it can be - future iterations could deal with other use cases.

app/assets/javascripts/govuk_publishing_components/components/autocomplete.js
```diff
@@ -0,0 +1,461 @@
...
+    var suggestionIcon = this.suggestionIcon
+
+    var searchIcon = '<span class="gem-c-autocomplete__result--search-icon"><svg width="20" height="20" viewBox="0 0 27 27" fill="none" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false"><circle cx="12.0161" cy="11.0161" r="8.51613" stroke="currentColor" stroke-width="3" /><line x1="17.8668" y1="17.3587" x2="26.4475" y2="25.9393" stroke="currentColor" stroke-width="3" /></svg></span>'
+
+    var renderResultHtml = function (result, props, inputVal) {
+      var index = result.toLowerCase().indexOf(inputVal.toLowerCase())
+      return `<li ${props.toString()}>${suggestionIcon ? searchIcon : ''}<span class='govuk-!-font-weight-bold'>${result.substring(0, index)}</span>${result.substring(index, index + inputVal.length)}<span class='govuk-!-font-weight-bold'>${result.substring(index + inputVal.length, result.length)}<span><div class='gem-c-autocomplete__result--border'></div></li>`
+    }
+
+    new Autocomplete(this.$module, // eslint-disable-line no-new
```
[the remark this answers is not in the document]
alexbowen: 100% agree. i have borrowed bits of code from gov accesible autocomplete, scotland gov autocomplete, blog posts and its become a bit jumbled

PR #4150, opened 2024-08-14
Add icon for external in attachment component

andysellick:
## What / why
- adds a specific icon for an external attachment, rather than the generic 'blank' icon
- external attachments are basically links to another site, which isn't really an attachment anyway, so already a bit confusing for users
- see example https://www.gov.uk/government/publications/form-pa1p-apply-for-probate-the-deceased-had-a-will

## Visual Changes
From the component guide:

Before | After
------ | ------
Screenshot 2024-08-14 at 09 56 21 | Screenshot 2024-08-14 at 09 56 28

From the example page:

Before | After
------ | ------
Screenshot 2024-08-14 at 09 43 30 | Screenshot 2024-08-14 at 09 43 38

Trello card: https://trello.com/c/C5G3fHke/146-redesign-of-publication-pages-follow-up


lib/govuk_publishing_components/presenters/attachment_helper.rb
```diff
@@ -1,7 +1,7 @@
 module GovukPublishingComponents
   module Presenters
     class AttachmentHelper
-      delegate :opendocument?, :document?, :spreadsheet?, :pdf?, to: :content_type
+      delegate :opendocument?, :document?, :spreadsheet?, :pdf?, :external?, to: :content_type
```
KludgeKML, line 4: I'm a bit confused here because we delegate external? to the content_type, but then provide an implementation of it.

andysellick: Sorry, that was left over from some experimenting I was doing to get this to work. Have removed.

KludgeKML (approved):
LGTM!

PR #4151, opened 2024-08-14
Add 'Chat entry' component

jon-kirwan:
## What

Add 'Chat entry' component

**Review URL**: https://components-gem-pr-4151.herokuapp.com/component-guide/chat_entry

## Why

A new component is required that will need to be integrated across multiple apps as the entry point to GOV.UK Chat.

### > Tablet

0 0 0 0_3212_component-guide_chat_entry(Surface Pro 7) (2)

### Mobile

0 0 0 0_3212_component-guide_chat_entry(iPhone SE)

### With a border top

0 0 0 0_3212_component-guide_chat_entry(Surface Pro 7) (3)

### With a border top and a border bottom

0 0 0 0_3212_component-guide_chat_entry(Surface Pro 7) (6)

## Anything else

The description text "Find answers to your business questions with GOV.UK's experimental new AI tool" is currently displayed using the old typescale, as shown in the screenshots above and in the component guide. This aligns with the current text display across GOV.UK, since the new typescale is still behind a feature flag. Once the new typescale is enabled, either by default or manually, the text will display at 19px across all viewport sizes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Ensure that organisation logos always print ([PR #4162](https://github.com/alphagov/govuk_publishing_components/pull/4162))
 * Fix some layout_header search visual bugs ([PR #4156](https://github.com/alphagov/govuk_publishing_components/pull/4156))
 * Fix issue with using special characters in content-list ([PR #4157](https://github.com/alphagov/govuk_publishing_components/pull/4157))
+* Add 'Chat entry' component ([PR #4151](https://github.com/alphagov/govuk_publishing_components/pull/4151))
 
 ## 42.0.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
@@ -1,0 +1,36 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-chat-entry {
+  display: flex;
+  align-items: flex-start;
+}
+
+.gem-c-chat-entry--border-top {
+  border-top: 2px solid govuk-colour("blue");
+  padding-top: govuk-spacing(5);
+}
+
+.gem-c-chat-entry--border-bottom {
+  border-bottom: 2px solid govuk-colour("blue");
+  padding-bottom: govuk-spacing(5);
+}
+
+.gem-c-chat-entry__image {
+  flex-shrink: 0;
+  height: 104px;
+  margin-right: govuk-spacing(3);
+  width: 104px;
+
+  @include govuk-media-query($from: tablet) {
+    height: 80px;
+    width: 80px;
+  }
+}
+
+.gem-c-chat-entry__heading {
+  margin: 0 0 govuk-spacing(2);
+  font-size: 19px;
+  font-size: govuk-px-to-rem(19);
+  line-height: 1.25;
+  @include govuk-typography-common;
+}
--- a/app/views/govuk_publishing_components/components/_chat_entry.html.erb
+++ b/app/views/govuk_publishing_components/components/_chat_entry.html.erb
@@ -1,0 +1,46 @@
+<%
+  add_gem_component_stylesheet("chat-entry")
+  local_assigns[:margin_bottom] ||= 6
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  href ||= "/chat"
+  heading_text ||= t("components.chat_entry.heading")
+  description_text ||= t("components.chat_entry.description")
+  border_top ||= false
+  border_bottom ||= false
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-chat-entry")
+  component_helper.add_class("gem-c-chat-entry--border-top") if border_top
+  component_helper.add_class("gem-c-chat-entry--border-bottom") if border_bottom
+  component_helper.add_class(shared_helper.get_margin_bottom)
+%>
+
+<%= tag.div(**component_helper.all_attributes) do %>
+  <svg class="gem-c-chat-entry__image" viewBox="0 0 104 104" fill="none" xmlns="http://www.w3.org/2000/svg" focusable="false" aria-hidden="true">
+    <rect width="100%" height="100%" fill="#D2E2F1" />
+    <path d="M37.5733 65.7452L51.6131 77.9999H27.9707L37.5733 65.7452Z" fill="#1D70B8" />
+    <path d="M52 78C66.3594 78 78 66.3594 78 52C78 37.6406 66.3594 26 52 26C37.6406 26 26 37.6406 26 52C26 66.3594 37.6406 78 52 78Z" fill="#1D70B8" />
+    <mask id="mask0_48346_4836" style="mask-type: alpha" maskUnits="userSpaceOnUse" x="40" y="58" width="24" height="10">
+      <rect x="41.2962" y="59.1165" width="21.4029" height="6.98153" fill="#D9D9D9" stroke="#F499BE" stroke-width="1.89313" />
+    </mask>
+    <g mask="url(#mask0_48346_4836)">
+      <path
+        d="M46.5668 60.156C47.8844 63.1549 51.3835 64.518 54.3824 63.2004C57.3813 61.8829 58.7443 58.3837 57.4268 55.3848C56.1093 52.3859 52.6101 51.0229 49.6112 52.3404C46.6123 53.658 45.2493 57.1571 46.5668 60.156Z"
+        stroke="#F499BE"
+        stroke-width="4.43733"
+      />
+    </g>
+    <path d="M43.3401 39.3198L50.2778 51.515H36.4023L43.3401 39.3198Z" fill="#FFDD00" />
+    <circle cx="60.2488" cy="45.4078" r="6.10621" fill="#F47738" />
+  </svg>
+
+  <div class="gem-c-chat-entry__description">
+    <%= content_tag(shared_helper.get_heading_level, class: "gem-c-chat-entry__heading") do %>
+      <%= link_to heading_text, href, class: "govuk-link" %>
+    <% end %>
+
+    <% if description_text %>
+      <p class="govuk-body govuk-!-margin-bottom-0"><%= description_text %></p>
+    <% end %>
+  </div>
+<% end %>
--- a/app/views/govuk_publishing_components/components/docs/chat_entry.yml
+++ b/app/views/govuk_publishing_components/components/docs/chat_entry.yml
@@ -1,0 +1,36 @@
+name: Chat entry
+description: Entry point for GOV.UK Chat
+accessibility_criteria: |
+  The heading must:
+
+  - be part of a correct heading structure for a page
+  - be semantically represented as a heading
+  - convey the heading level
+uses_component_wrapper_helper: true
+shared_accessibility_criteria:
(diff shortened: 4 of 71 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
```diff
@@ -0,0 +1,27 @@
...
+  .gem-c-chat-entry__image {
+    flex-shrink: 0;
+    margin-right: govuk-spacing(3);
+  }
+
+  .gem-c-chat-entry__heading {
+    @include govuk-typography-common;
+    @include govuk-font-size(24);
+    margin: 0 0 govuk-spacing(2);
+  }
```
dnkrj, line 26: From my understanding of the styleguide, these should not be nested and be positioned at the top level with `.gem-c-chat-entry`.

app/views/govuk_publishing_components/components/_chat_entry.html.erb
```diff
@@ -0,0 +1,47 @@
+<%
+  add_gem_component_stylesheet("chat-entry")
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  href ||= nil
+  heading_text ||= nil
```
dnkrj, line 5: Do we want the `href`, `heading_text` and `description_text` to be parameters?
My understanding is that they will be same everywhere so it would be much simpler to bake them into the component and avoid the need to update them on every page if they change.
Alternatively we could use the current text as the default and allow it to be overridden.

dnkrj (changes requested):
Code looks good 🎉
Main question is about what should be a parameter for the component and what should be baked in

app/views/govuk_publishing_components/components/_chat_entry.html.erb
jon-kirwan: Good thinking. I thought that component options might add a bit more flexibility but perhaps that's not really needed. Default text might be a good idea. @JamesCGDS what do you think?

app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
jon-kirwan: @dnkrj could you please share a link to the style notes? I noticed some comments about [avoiding deep nesting](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md#bem), but I see a variety of approaches to nesting in other component styles (which might be incorrect).

andysellick: I think that's the guidance you've found there. Generally the use of BEM means we shouldn't have to nest element or modifier classes like these (although of course there are exceptions).

app/views/govuk_publishing_components/components/_chat_entry.html.erb
```diff
@@ -0,0 +1,47 @@
...
+  description_text ||= nil
+  border_top ||= false
+  border_bottom ||= false
+
+  classes = %w[gem-c-chat-entry]
+  classes << "gem-c-chat-entry--border-top" if border_top
+  classes << "gem-c-chat-entry--border-bottom" if border_bottom
+%>
+
+<%= content_tag("div", class: classes ) do %>
```
andysellick, line 15: Non-blocking: have you considered using the [component wrapper helper](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component-wrapper-helper.md) for this component?

spec/components/chat_entry_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+  it "renders the chat entry component with a border top" do
+    render_component({
+      href: "/get-started",
+      heading_text: "Try GOV.UK Chat",
+      description_text: "Find answers to your business questions",
+      border_top: true,
+    })
+
+    assert_select ".gem-c-chat-entry.gem-c-chat-entry--border-top"
+  end
```
JamesCGDS, line 40: Quick question - do we need a companion test for this to check the component renders with a bottom border when passed `border_bottom: true`?

app/assets/images/chat-entry/icon-chat-entry.svg
kevindew, line 1: Doesn't look like anything makes use of this, given it's been inlined so I don't think we gain anything adding it to the repo

jon-kirwan: Thanks. I had planned to use the `inlinesvg` gem, but after deciding not to, I forgot to remove the asset. It’s now been removed.

spec/components/chat_entry_spec.rb
jon-kirwan: Yep, I think that's a good idea. It's now been added.

app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
jon-kirwan: Thanks both. I've removed the nesting now

spec/components/chat_entry_spec.rb
JamesCGDS: Awesome, thanks Jon. Just one more quick thing (sorry I just noticed this) - do we need to add a corresponding section in the component guide doc for the `border_bottom` option?

app/views/govuk_publishing_components/components/_chat_entry.html.erb
jon-kirwan: I did consider it, but initially thought I might not need it since I was only adding class names. However, I’ve changed my mind and decided to use it after all, as it will be useful if I need to add other attributes, which is quite likely

spec/components/chat_entry_spec.rb
jon-kirwan: Yep sounds good to me. I've added a separate test now.

app/views/govuk_publishing_components/components/_chat_entry.html.erb
jon-kirwan: Thanks both. I've made the changes in a separate commit. I've added default text for href, heading, and description. The description text has been updated to the latest version. I've also revised the tests and removed the href condition, as I think we'll always need the href. The only thing I'm still unsure about is what the final href will be, so I'll need to confirm that.

JamesCGDS: Cool, that all sounds good to me 👍

dnkrj (approved):

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * **BREAKING:** Drop libsass support ([PR #4106](https://github.com/alphagov/govuk_publishing_components/pull/4106))
 * **BREAKING:** Drop support for legacy browsers ([PR #4111](https://github.com/alphagov/govuk_publishing_components/pull/4111))
 
+## Unreleased
+
+* Add 'Chat entry' component ([PR #4151](https://github.com/alphagov/govuk_publishing_components/pull/4151))
```
andysellick, line 17: Something weird with the CHANGELOG here - this needs to be above `42.0.0`.

app/assets/stylesheets/govuk_publishing_components/components/_chat-entry.scss
```diff
@@ -0,0 +1,27 @@
...
+}
+
+.gem-c-chat-entry__image {
+  flex-shrink: 0;
+  margin-right: govuk-spacing(3);
+}
+
+.gem-c-chat-entry__heading {
+  @include govuk-typography-common;
+  @include govuk-font-size(24);
```
andysellick, line 25: Nitpick: I think all includes like this need to now follow regular lines of CSS, rather than precede them, owing to a problem with a dependency that @kevindew fixed.

app/views/govuk_publishing_components/components/docs/chat_entry.yml
```diff
@@ -0,0 +1,32 @@
+name: Chat entry
+description: Entry point for GOV.UK Chat
+accessibility_criteria: |
+  The heading must:
+
+  - be part of a correct heading structure for a page
+  - be semantically represented as a heading
+  - convey the heading level
+shared_accessibility_criteria:
+  - link
```
andysellick, line 10: Nitpick: you need to add `uses_component_wrapper_helper: true` for the component wrapper information to be displayed in the component guide for this component.

app/views/govuk_publishing_components/components/_chat_entry.html.erb
```diff
@@ -0,0 +1,44 @@
+<%
+  add_gem_component_stylesheet("chat-entry")
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
```
andysellick, line 3: Nitpick: the component by default appears to have no bottom margin, which means it might collide with anything below it (we typically apply bottom margin to elements in order to avoid top/bottom margin overlap issues). Since you've got the shared helper in here, you can do something like this to set a default bottom margin:

```
local_assigns[:margin_bottom] ||= 6
```

app/views/govuk_publishing_components/components/docs/chat_entry.yml
```diff
@@ -0,0 +1,32 @@
...
+examples:
+  default:
+    data:
+  with_custom_href,_heading_and_or_description:
+    data:
+      href: /chat
```
andysellick, line 16: Nitpick: would it be clearer to use something other than the default here?

app/views/govuk_publishing_components/components/docs/chat_entry.yml
jon-kirwan: Thanks, it's been added now.

CHANGELOG.md
jon-kirwan: Thanks, fixed now.

app/views/govuk_publishing_components/components/_chat_entry.html.erb
jon-kirwan: Added 👍

app/views/govuk_publishing_components/components/docs/chat_entry.yml
jon-kirwan: The default will likely be the most common way that the component will be used. Perhaps I should add a description for it for clarity?

andysellick: Sure, the default is likely to be unchanged, but to demonstrate that a different URL can be passed I think it's worth demonstrating that.

Related - the test that checks that a custom URL can be passed also uses the default, so it doesn't actually test that a different URL can be passed.

jon-kirwan: Cool thanks, I've updated the default to include the `heading_text` option and also modified the test to use a different `href` instead of the default one.

app/views/govuk_publishing_components/components/docs/chat_entry.yml
```diff
@@ -0,0 +1,37 @@
...
+  - be part of a correct heading structure for a page
+  - be semantically represented as a heading
+  - convey the heading level
+uses_component_wrapper_helper: true
+shared_accessibility_criteria:
+  - link
+examples:
+  default:
+    data:
+      heading_text: Go to GOV.UK Chat
```
leenagupte, line 15: The default is to not pass it anything, so perhaps that should be shown here?  Especially as there is an example with a custom heading?

jon-kirwan: Thanks @leenagupte. I see now that components without any required options, like the cookie banner or skip link components, are created in a similar way. I think I'll follow that approach here too.

PR #4156, opened 2024-08-16
Fix some layout_header search visual bugs

AshGDS:
## What / Why
 - Makes the search box in `layout_header` by using the `one-half` column sizes instead to fix a resizing issue
 - Adds `cursor: text` to the placeholder text label, as the cursor looked incorrect when you hovered over the label without this
 - Modifying the label text on the service standard page itself may be complicated, so I'll explore that as a separate task - essentially the 'Search on GOV.UK' text is the default from the component, and I believe the header component is rendered via `static`, so modifying it for one page template in `government_frontend` might be a bit complex
 
### Before

image

image

### After

image

image

image


andysellick (approved):
Looks good 👍

PR #4162, opened 2024-08-19
Ensure that organisation logos always print

matthillco:
## What
When printing a page containing organisation logos, we want to ensure that any images are always printed. Without the images, each logo is printed as text only with an empty space for the logo and looks quite strange. 

In addition, browsers don't print background images by default, and will only print them if a hidden option is selected from the printer dialogue. This PR forces the browser to always print the logo images regardless of settings.

This is part of the work to improve our print styles more generally. [Trello](https://trello.com/c/K58GNTxY/268-fix-organisational-logo-components-to-print-correctly)

## Why
Printing the logo images helps users to establish recognition for each organisation. 

### Before

image

### After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


andysellick (approved):
Looks good, probably worth a CHANGELOG entry.

Issue #4154, opened 2024-08-15
Layout header does not toggle search component on mobile or tablet


MartinJJones:
## What

Clicking the search button does not toggle visibility of the search menu on screen widths <= 640px

A live example of this issue can seen in the service manual - https://www.gov.uk/service-manual/service-standard

AshGDS:
I unintentionally fixed this before realising it was a separate issue card, so to fix it, it should just be a case of:

- Switching the `js-header-toggle` to `govuk-js-header-toggle` on the button.
- Adding `aria-controls='search'` to the button.
- Removing In `_layout-header.scss`, delete the `display: none` for `.gem-c-layout-header__search-form {` (line 151)

MartinJJones:
That's great, thanks @AshGDS

PR #4163, opened 2024-08-20
Fix the search toggle in layout header

MartinJJones:
## What
Fix the search toggle in layout header

## Why
On mobile and tablet screen sizes clicking on the search icon in the layout header did not reveal the hidden search form element, `gem-c-layout-header__search-form`.

The reason for this is that as of [V3.0.0 of govuk-frontend](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#update-the-html-for-navigation-in-the-header), classes have a `govuk-` prefix. The HTML for the search template has been updated to reflect this, as well as using the `aria-controls` attribute to target the element to be toggled.

The toggle functionality, makes use of the `hidden` attribute to show/hide the target element, `display: none` has been removed from `.gem-c-layout-header__search-form` to ensure that the search form visibility can be toggled and can be accessed when JS is disabled.

[Trello card]()
Fixes: https://github.com/alphagov/govuk_publishing_components/issues/4154

### JS Disabled - Mobile/Tablet

| Before | After |
| --- | --- |
| js-disabled-before | js-enabled-after |

### JS Enabled - Search toggle button clicked - Mobile/Tablet

| Before | After |
| --- | --- |
| js-enabled-click-before |   js-disabled-click-after |

**Note:** Percy is showing some visual regression changes on mobile, this is likely because the search-form is now displayed by default, then later hidden by JavaScript.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix the search toggle in layout header ([PR #4163](https://github.com/alphagov/govuk_publishing_components/pull/4163))
+
 ## 43.0.1
 
 * Fix card component padding based on heading ([PR #4169](https://github.com/alphagov/govuk_publishing_components/pull/4169))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -148,8 +148,6 @@
 }
 
 .gem-c-layout-header__search-form {
-  display: none;
-
   &.js-visible {
     display: block;
   }
--- a/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
+++ b/app/views/govuk_publishing_components/components/layout_header/_navigation_items.html.erb
@@ -1,9 +1,12 @@
-<% navigation_aria_label ||= "Top level" %>
+<%
+  navigation_aria_label ||= "Top level"
+  navigation_id = "navigation-#{SecureRandom.hex(4)}"
+%>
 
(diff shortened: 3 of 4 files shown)


AshGDS (approved):
Nice work 👍 

I think the Percy visual change might be something that could potentially be fixed. I think what's happening is that because there's multiple versions of the `layout_header` on the component guide, the JS that selects the DOM element is always grabbing the first instance of the `layout_header`, when it should be grabbing the one that it's initialised on. I think this is the case because if you view the `layout_header` with search bar example in isolation, the JavaScript works OK whereas it doesn't when it's in the list with all the examples. 

I'm not sure we're able to fix this, but if you have any ideas feel free to push them to this branch and I can re-review 👍

MartinJJones:
Thanks for the review @AshGDS, good spot on the issue in the component guide as well.

I've had a look at this and think I have found a good fix for it, let me know what you think, thanks 😄

AshGDS (approved):
Nice solution 👍 

There is this slimmer header that references `#search`, but I don't think it's used anymore as it's also looking for `#global-header` which I don't think exists in our HTML anywhere. https://github.com/alphagov/slimmer/blob/4f8f5ad6ce6c91cdddaa4bfa14e16a0777065f4e/lib/slimmer/processors/search_remover.rb#L9

MartinJJones:
Thanks @AshGDS for the review and highlighting the slimmer header as well. Had a further look at this and agree that it does not appear to be used anymore, the search component is still removed from the [service manual homepage](https://www.gov.uk/service-manual) in the layout header as it references the [`.gem-c-layout-header__search`](https://github.com/alphagov/slimmer/blob/4f8f5ad6ce6c91cdddaa4bfa14e16a0777065f4e/lib/slimmer/processors/search_remover.rb#L15) class

PR #4165, opened 2024-08-20
Fix print issues on layout-super-navigation-header

matthillco:
## What

The print styles on the layout-super-navigation-header are incorrect and need to be updated:

- Background should always print transparent
- Foreground should always print black
- No top border
- Bottom border should match new changes in cross service header
- Left/right margins are not required for print due to background removal

[Trello](https://trello.com/c/4Lk0dUxa/292-fix-print-issues-on-layout-super-navigation-header)

## Why
- The top bar has an extra top border and redundant left/right spacing when printed, it also doesn’t match the changes made to the cross service header (coming in #4073)
- The background color bug was recently introduced in the previous print styles work and needs fixing: it prevents the inverted background from printing correctly.

### Before
image

### After
image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -986,9 +986,23 @@ $after-button-padding-left: govuk-spacing(4);
 
 @include govuk-media-query($media-type: print) {
   .gem-c-layout-super-navigation-header {
-    border-top: 1pt solid $govuk-print-text-colour;
-    margin: 0;
-    background-color: none;
+    border-top: 0;
+    border-bottom: 2pt solid $govuk-print-text-colour;
+    margin: 0 0 5mm;
+    background-color: transparent;
+
+    &:has(.gem-c-layout-super-navigation-header__header-logo .govuk-visually-hidden) {
+      border: 0;
+      margin-left: 24px;
+    }
+
+    * {
+      color: $govuk-print-text-colour;
+    }
+
+    .govuk-width-container {
+      margin: 0;
+    }
   }
 
   .gem-c-layout-super-navigation-header__header-logo {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+Unreleased
+
+* Fix print issues on layout-super-navigation-header ([PR #4165](https://github.com/alphagov/govuk_publishing_components/pull/4165))
+
 ## 43.0.0
 
 * **BREAKING:** Upgrade to govuk-frontend v5.5.0 ([PR #4160](https://github.com/alphagov/govuk_publishing_components/pull/4160))


matthillco:
> Just wondering why the homepage doesn't have text?

I did originally write some new print styles to ensure that all the headers looked the same with the same "GOV.UK" text ext to the crown logo. However during review @andysellick suggested it wasn't necessary to do that since the homepage is an edge case where the "GOV.UK" text is part of the title content, and the crown is left as a single ident. 

The problem now is that by having the border underneath the crown/logo, it looks good for most pages but might look odd if the homepage is printed.

matthillco:
> Just wondering is there a screenshot available of a GOV.UK page with the updated print styles?
> 
> Be good to see it within the context an end-user experiences it.

I'll sort out a preview app of `frontend` for you to look at this, I'll let you know when it's available.

andysellick (dismissed):
Looks good to me 👍 

Have checked in `frontend` as well, all looks okay.

matthillco:
@nnagewad Here's before/after screenshots of the revised top header bar. (Note there's some other print changes on these pages, but we're not looking at those in this PR.)

## Standard Page
image

## Homepage
image

nnagewad:
> @nnagewad Here's before/after screenshots of the revised top header bar. (Note there's some other print changes on these pages, but we're not looking at those in this PR.)
> 
> ## Standard Page
> image
> 
> ## Homepage
> image

This looks really good. Just wondering, is it worth having the horizontal rule on the homepage?

From your experience, how difficult will it be to remove it?

matthillco:
@nnagewad This is updated now so that variants for the homepage will not show the border, and the logo is indented to align with the intro text below: 
image

nnagewad (approved):

AshGDS (approved):

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -986,9 +986,23 @@ $after-button-padding-left: govuk-spacing(4);
...
   .gem-c-layout-super-navigation-header {
-    border-top: 1pt solid $govuk-print-text-colour;
-    margin: 0;
-    background-color: none;
+    border-top: 0;
+    border-bottom: 2pt solid $govuk-print-text-colour;
+    margin: 0 0 5mm;
+    background-color: transparent;
+
+    &:has(.gem-c-layout-super-navigation-header__header-logo .govuk-visually-hidden) {
```
andysellick, line 994: Appreciate I'm a bit late to this, but this feels like a brittle solution - I'd rather tie this to specific classes than assume a particular markup. I think it's also a little hard to read - it took me a bit of digging in the code to understand why this rule worked.

I think a better alternative would be to add a class somewhere to indicate when the header logo text is hidden, and also target the homepage class that already exists, something like:

```CSS
&.gem-c-layout-super-navigation-header--blue-background,
.gem-c-when-the-text-of-the-logo-is-hidden {
  border: 0;
  margin-left: 24px;
}
```

matthillco: I looked at this again because I didn't really see a problem with my original fix, but I've now noticed a different issue. We don't need the `visually-hidden` GOV.UK text at all and should remove it, then use your suggested fix instead.

**Longer explanation:**

There are two different logos that render depending on the value of `hide_logo_text` [link](https://github.com/alphagov/govuk_publishing_components/commit/6053c6d32fa2a78d7c8b33ed433c3f42a70f0e23).
-`hide_logo_text: false` Render a crown logo that inlcudes GOV.UK text
-`hide_logo_text: true` Render **only** a crown logo **without** GOV.UK text

Further logic renders a `visually-hidden` span containing GOV.UK text for the crown-only logo, but this seems redundant given that:

- Screen-readers read GOV.UK from the title of the SVG for pages with the crown logo + text.
- The Homepage displays "GOV.UK" text further down the page, clarifying the textless crown logo.

Also, there are some improvements to be made around how the GOV.UK text could be inserted in the SVG the same way as the `aria_label_text` is inserted in [_govuk_logo.html.erb](app/views/govuk_publishing_components/components/govuk_logo/_govuk_logo.html.erb)

This all needs further discussion, and given that my original fix has already been merged, it would be better to open a new issue for this within our team to try and address these issues collectively.

andysellick: Agreed. If you can condense all of the above into a draft card on the backlog, and we can look at it as a follow up piece of work in a later sprint, thanks.

PR #4167, opened 2024-08-21
[DO NOT MERGE] Preview gem print styles

matthillco:
## What
A temporary PR for previewing WIP gem component print styles, required for [#4205 in `frontend`](https://github.com/alphagov/frontend/pull/4205)

## Why
We need to see print style components in context of various pages to be sure they print correctly.


matthillco:
No longer needed.

PR #4176, opened 2024-08-27
Add the password input component

AshGDS:
## What / Why
- Adds the password input component from the GOVUK Design System - [see their docs here](https://design-system.service.gov.uk/components/password-input/)
- There is a bug which I wasn't sure what to do about. Our component docs say that text in components should be translatable - so I tried to allow the `Show/Hide` button text to be overwritten. However, the button text gets changed by the GOVUK Frontend JavaScript on click. Therefore even if we change the text in the HTML, it gets overwritten as soon as you click the button. Therefore I've added the ability to change the button text in the template file, but I haven't made it known in the `yml` docs that it's a feature.
- Hope everything's up to standard - first time adding a component!
- Trello card: https://trello.com/c/Gj5yIsxH/298-implement-password-input-component

## Visual Changes

- The addition of the component has added visual changes to review.

### DIFF
--- a/app/views/govuk_publishing_components/components/_password_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_password_input.html.erb
@@ -1,0 +1,78 @@
+<%
+  add_gem_component_stylesheet("password-input")
+
+  label_text ||= t("components.password_input.label")
+
+  error_text ||= nil
+  error_text_prefix ||= t("components.password_input.error_text_prefix")
+
+  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_data_attribute({ module: "govuk-password-input" })
+  component_helper.add_class('govuk-form-group')
+  component_helper.add_class('govuk-form-group--error') if error_text
+  component_helper.add_class('govuk-password-input')
+  component_helper.add_class('gem-c-password-input')
+  component_helper.add_class(shared_helper.get_margin_bottom)
+
+  uid = SecureRandom.hex(4)
+
+  label_for = uid + '-password-input'
+
+  input_name = 'password'
+  input_id = uid + '-password-input'
+
+  aria_controls = uid + '-password-input'
+
+  input_classes = %w(govuk-input govuk-password-input__input govuk-js-password-input-input)
+
+  if error_text
+    label_for << '-with-error-message'
+    input_id << '-with-error-message'
+    input_name << '-input-with-error-message'
+    input_classes << 'govuk-input--error'
+    aria_controls << '-with-error-message'
+    paragraph_id = uid + '-password-input-with-error-message-error'
+  end
+
+%>
+
+<%= tag.div(**component_helper.all_attributes) do %>
+  <%= render "govuk_publishing_components/components/label", {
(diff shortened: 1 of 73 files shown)


app/views/govuk_publishing_components/components/docs/password_input.yml
```diff
@@ -0,0 +1,42 @@
...
+uses_component_wrapper_helper: true
+
+examples:
+  default:
+    data:
+  with_error:
+    description: If there is a validation error, passing error text will style the password input component with error styles, and semantically state that there was a validation error.
+    data:
+      error_text: "Enter a password"
+  with_custom_text:
```
andysellick, line 37: Could this be `with_custom_label_and_text`? I was initially confused reading it.

app/views/govuk_publishing_components/components/_password_input.html.erb
```diff
@@ -0,0 +1,80 @@
...
+  label_text ||= t("components.password_input.label")
+
+  # This overrides the 'Show/Hide' button text, but this button text is getting overwritten by GOVUK Frontend JS currently when you interact with the button. Therefore this functionality is not being exposed in the docs.
+  button_text ||= t("components.password_input.button")
+  button_aria_label_text ||= t("components.password_input.button_aria_label")
+
+  error_text ||= nil
+  error_text_prefix ||= t("components.password_input.error_text_prefix")
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
```
andysellick, line 13: Nice. Also is it worth adding the shared helper for margin? It's usually the first thing we need.

Side note: I'm planning on moving the margin functionality into the component wrapper helper at some point, but it's a non-trivial task.

andysellick (commented):
Looks good! Just a few minor things to look at.

AshGDS:
Thanks for reviewing @andysellick 👍 I've made every change requested, so this should be ready for another review.

app/views/govuk_publishing_components/components/_password_input.html.erb
```diff
@@ -0,0 +1,80 @@
+<%
+  add_gem_component_stylesheet("password-input")
+
+  label_text ||= t("components.password_input.label")
+
+  # This overrides the 'Show/Hide' button text, but this button text is getting overwritten by GOVUK Frontend JS currently when you interact with the button. Therefore this functionality is not being exposed in the docs.
```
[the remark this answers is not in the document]
AshGDS: @andysellick It looks possible, so I'll implement it tomorrow 👍 https://gds.slack.com/archives/CAF8JA25U/p1725376708644069?thread_ts=1725374973.976699&cid=CAF8JA25U

spec/components/password_input_spec.rb
```diff
@@ -5,24 +5,30 @@ def component_name
     "password_input"
   end
 
+  before { allow(SecureRandom).to receive(:hex).and_return("1234") }
+
   it "renders a password input component correctly by default" do
     render_component({})
+    puts rendered
```
andysellick, line 12: Debug?

andysellick (approved):
Nice work. One bit of debug to remove, and obviously squash the commits, but happy to approve ahead of that 👍

spec/components/password_input_spec.rb
AshGDS: Whoops, thanks!

PR #4178, opened 2024-08-28
Improve print styles for layout-footer component

matthillco:
## What
This improves the print styles for the `layout-footer` component as part of the work to improve print styles for page layouts. [Trello](https://trello.com/c/tcXZsXup/193-review-and-fix-page-level-print-styles). 
- Extra borders and padding are removed
- internal elements now run flush to their left/right edges

Note that for consistency [all variants of the layout-footer](https://components.publishing.service.gov.uk/component-guide/layout_footer) now render identically when printed .

## Why
Although we made improvements to the `layout-footer` as part of #4073, we did not look at how the component might interact with other components on the page. This PR further improves the component in context of page types. 

### Before and After: Component on a typical page
image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
@@ -21,10 +21,39 @@
   margin-bottom: 20px;
 }
 
-// stylelint-disable declaration-no-important
 @include govuk-media-query($media-type: print) {
+  .govuk-footer {
+    background: none;
+    margin: 25mm 0 0;
+    padding: 5mm 0 0;
+    border-top: 2pt solid $govuk-print-text-colour;
+
+    .govuk-width-container {
+      margin: 0;
+    }
+  }
+
+  .govuk-footer__meta,
+  .govuk-footer__meta-item {
+    border: 0;
+    padding: 0;
+    margin: 0;
+  }
+
+  .govuk-footer__licence-description {
+    padding-right: 10mm;
+  }
+
+  .govuk-footer__copyright-logo {
+    min-width: auto;
+    padding-top: 13mm;
+    background-size: 15mm;
+    background-position: center top;
+    print-color-adjust: exact;
+    -webkit-print-color-adjust: exact;
+  }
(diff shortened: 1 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
```diff
@@ -21,10 +21,43 @@
...
+    padding: 5mm 0 0;
+    border-top: 2pt solid $govuk-print-text-colour;
+
+    .govuk-width-container {
+      margin: 0;
+    }
+  }
+
+  .govuk-footer__section-break {
+    display: none;
```
AshGDS, line 37: Is there a reason that this is preferred over adding `govuk-!-display-none-print` to the HTML itself?

app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
```diff
@@ -21,10 +21,43 @@
   margin-bottom: 20px;
 }
 
-// stylelint-disable declaration-no-important
 @include govuk-media-query($media-type: print) {
+  .govuk-footer {
+    background: none;
+    margin: 25mm 0 0;
```
AshGDS, line 27: Was there a specific reason for this much margin? I'm mindful that the more margin there is, the more likely the footer will spill onto another page, which would waste paper.

AshGDS (approved):
LGTM so will approve, just had a couple questions which may result in changes.

app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
matthillco: Yes, this was discussed with Nikin and we agreed that more space was needed between the last element of the page and the footer, otherwise it looks too cramped.

app/assets/stylesheets/govuk_publishing_components/components/_layout-footer.scss
matthillco: Good spot thanks, I've changed it as you suggested.

PR #4179, opened 2024-08-28
Improve print styles for inverse-header component

matthillco:
## What
This improves the print styles for the `inverse-header` component as part of the work to improve print styles for page layouts:

- Padding is removed for better alignment with printed page content

[Trello](https://trello.com/c/tcXZsXup/193-review-and-fix-page-level-print-styles)

## Why
Although we made improvements to this component as part of #4073, we did not look at how the component might interact with other components on the page. This PR further improves the component in context of page types. 

### Before and After: Component Preview

imageimage

### Before and After: Component on a HTML Publication page
imageimage


AshGDS (commented):
Looks good - but similar to the other review, does it need this much top and bottom padding?

matthillco:
> Looks good - but similar to the other review, does it need this much top and bottom padding?

It's been reviewed by Nikin and we've kept some of that internal top and bottom padding as it allows breathing space between elements of the page. If the padding is reduced too much, the design suffers and becomes harder to scan.

PR #4182, opened 2024-08-30
Add Summary Card component to Gem

davidtrussler:
## What
Adds a new component - Summary Card - to the Gem. 
These changes 
- copy the component code from the Whitehall codebase ([here](https://github.com/alphagov/whitehall/blob/main/app/views/components/_summary_card.html.erb), [here](https://github.com/alphagov/whitehall/blob/main/app/assets/stylesheets/components/_summary-card.scss) and [here](https://github.com/alphagov/whitehall/blob/main/app/views/components/docs/summary_card.yml)) to this one
- update some class names in line with the Publishing Components standards
- refactor the ERB file to use Rails tags
- adds a test for the new component

## Why
This component already exists on Whitehall ([Summary Card](http://whitehall-admin.dev.gov.uk/component-guide/summary_card)) but is required on other apps. To facilitate this we need to move the component from Whitehall to the Publishing Components gem for wider use. 

## Visual Changes
This should not introduce any visual changes. For reference the component in its default state is shown below. 

Screenshot 2024-08-30 at 09 33 54

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_summary-card.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_summary-card.scss
@@ -1,0 +1,16 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk/components/summary-list/summary-list";
+
+.gem-c-summary-card {
+  .govuk-summary-list__key {
+    vertical-align: top;
+  }
+
+  .govuk-summary-list__actions {
+    width: auto;
+  }
+
+  img {
+    display: block;
+  }
+}
--- a/app/views/govuk_publishing_components/components/_summary_card.html.erb
+++ b/app/views/govuk_publishing_components/components/_summary_card.html.erb
@@ -1,0 +1,62 @@
+<%
+  add_gem_component_stylesheet("summary-card")
+
+  id ||= nil
+  title ||= nil
+  data_attributes ||= {}
+  summary_card_actions ||= []
+  rows ||=[]
+%>
+<% if title || rows.any? %>
+  <%= tag.div class: "gem-c-summary-card", id: id, data: data_attributes do %>
+    <%= tag.div class: "govuk-summary-card" do %>
+      <%= tag.div class: "govuk-summary-card__title-wrapper" do %>
+        <%= tag.h2 class: "govuk-summary-card__title" do %>
+          <%= title %>
+        <% end %>
+
+        <%= tag.ul class: "govuk-summary-card__actions" do %>
+          <% summary_card_actions.each do |action| %>
+            <%= tag.li class: "govuk-summary-card__action" do %>
+              <%= link_to sanitize(action[:label] + tag.span(" #{title}", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state #{"gem-link--destructive govuk-!-font-weight-bold" if action[:destructive]}".strip %>
+            <% end %>
+          <% end %>
+        <% end %>
+      <% end %>
+
+      <% if rows.present? %>
+        <%= tag.div class: "govuk-summary-card__content" do %>
+          <%= tag.dl class: "govuk-summary-list" do %>
+            <% rows.each do |row| %>
+              <%= tag.div class: "govuk-summary-list__row" do %>
+                <%= tag.dt class: "govuk-summary-list__key" do %>
+                  <%= row[:key] %>
+                <% end %>
+
+                <%= tag.dt class: "govuk-summary-list__value" do %>
+                  <%= row[:value] %>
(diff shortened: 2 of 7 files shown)


app/views/govuk_publishing_components/components/_summary_card.html.erb
```diff
@@ -0,0 +1,58 @@
...
+                <% end %>
+
+                <% if row[:actions].present? %>
+                  <%= tag.dd class: "govuk-summary-list__actions" do %>
+                    <% row[:actions].each do |action| %>
+                      <% if action[:opens_in_new_tab] %>
+                        <%= link_to sanitize(action[:label] + tag.span(" #{row[:key]} (opens in new tab)", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state govuk-!-margin-left-2", rel: "noreferrer noopener", target: "_blank" %>
+                      <% else %>
+                        <%= link_to sanitize(action[:label] + tag.span(" #{row[:key]}", class: "govuk-visually-hidden")), action[:href], class: "govuk-link govuk-link--no-visited-state govuk-!-margin-left-2 #{"gem-link--destructive" if action[:destructive]}".strip %>
+                      <% end %>
```
jon-kirwan, line 47: I'm just wondering if these actions should be pipe separated?

components-gem-pr-4182 herokuapp com_component-guide_summary_card

app/assets/stylesheets/govuk_publishing_components/components/_summary-card.scss
```diff
@@ -0,0 +1,16 @@
...
+.gem-c-summary-card {
+  .govuk-summary-list__key {
+    vertical-align: top;
+  }
+
+  img {
+    display: block;
+  }
+
+  .govuk-summary-card__action:not(:first-child):not(:last-child) {
```
jon-kirwan, line 13: A small suggestion - I wonder if nesting the pseudo selectors might be a little more readable?
```
  .govuk-summary-card__action {
    &:not(:first-child):not(:last-child) {
      padding-left: govuk-spacing(2);
    }
  }
```

jon-kirwan (commented):
Thanks David. It looks good to me. I was considering whether the summary card could be an extension of the summary list instead of a standalone component, but I see the code is quite different, and making that change would likely involve breaking changes to the summary list.

I also thought about the accessibility of the action links, like ‘view,’ but I noticed that’s how they are in the Design System. I assume they get their context from the summary list key.

davidtrussler:
> Thanks David. It looks good to me. I was considering whether the summary card could be an extension of the summary list instead of a standalone component, but I see the code is quite different, and making that change would likely involve breaking changes to the summary list.
>
Thanks Jon, yes I was thinking about that. In the end I decided to keep it as much the same as possible as the component on Whitehall. Also, as you have said, it felt like it could cause problems that way.
 
> I also thought about the accessibility of the action links, like ‘view,’ but I noticed that’s how they are in the Design System. I assume they get their context from the summary list key.

Yes, I think that should be OK with the visually hidden text so screen readers get "View key one" etc.

app/views/govuk_publishing_components/components/_summary_card.html.erb
davidtrussler: Yes, they display like that when used in the title section so replicating that here makes sense. I've just added the `govuk-summary-card__action` class to the links that adds that.

jon-kirwan: Great thanks. Although, I think you can actually now remove these styles:
```
  .govuk-summary-card__action {
    &:not(:first-child):not(:last-child) {
      padding-left: govuk-spacing(2);
    }
  }
```
The `govuk-summary-card__action` seems to add the correct spacing between items (otherwise, there's a bit too much spacing on the middle item)

jon-kirwan: Great, that looks good to me. Just one last thing: I noticed that on Desktop, when all three actions are displayed, the third one wraps to the next line. This happens because the Design System code sets a fixed width of 20% on the actions column. You could set the cell width to `auto` to keep them all in line:
```
.govuk-summary-list__actions {
  width: auto;
}
```

jon-kirwan (commented):
Thanks for the changes, David. Everything looks good to me. Just one last suggestion: you can probably squash a few commits together, but otherwise, nice work.

app/views/govuk_publishing_components/components/_summary_card.html.erb
davidtrussler: Ah yes, nice one. Have done that. 👍

jon-kirwan (approved):

PR #4184, opened 2024-09-02
Replace 'unset' property in printed background colours

matthillco:
## What
In improving print styles in gem components (#4073), we used `background-color: unset` to revert backgrounds to transparent for print. This would ensure that inverse components would always print non-inverted, saving ink.

After discussion we decided to change this to use `background: none` instead, because:
- the use of `unset` is uncommon and not well understood, and its behaviour can be unpredictable
- `background: none` is a more common accepted approach and easier to understand. We've taken into account that resetting all background properties can be considered a [CSS anti-pattern](https://csswizardry.com/2016/12/css-shorthand-syntax-considered-an-anti-pattern/), however we're confident in this case that it won't have any adverse effects because we're not using any other background properties in the affected components.

This PR also makes consistent other treatments of background, such as `background-color: transparent` or `background: initial`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix the search toggle in layout header ([PR #4163](https://github.com/alphagov/govuk_publishing_components/pull/4163))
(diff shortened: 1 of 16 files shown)


AshGDS (approved):
LGTM 👍

PR #4196, opened 2024-09-04
Add ga4 tracking to chat entry component

JamesCGDS:
## What
This PR adds GA4 tracking to the chat entry component.

## Why
Requested by the PA on the AI team.

### DIFF
--- a/app/views/govuk_publishing_components/components/_chat_entry.html.erb
+++ b/app/views/govuk_publishing_components/components/_chat_entry.html.erb
@@ -7,12 +7,24 @@
   description_text ||= t("components.chat_entry.description")
   border_top ||= false
   border_bottom ||= false
+  disable_ga4 ||= false
 
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-chat-entry")
   component_helper.add_class("gem-c-chat-entry--border-top") if border_top
   component_helper.add_class("gem-c-chat-entry--border-bottom") if border_bottom
   component_helper.add_class(shared_helper.get_margin_bottom)
+  component_helper.add_data_attribute({ module: "ga4-link-tracker" }) unless disable_ga4
+
+  unless disable_ga4
+    ga4_link_data = {
+      ga4_link: {
+        event_name: "navigation",
+        type: "callout",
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/chat_entry.yml
```diff
@@ -34,3 +34,6 @@ examples:
   with_margin_bottom:
     data:
       margin_bottom: 3
+  without_ga4_tracking:
+    data:
+      disable_ga4: true
```
AshGDS, line 39: I think you should add the description to this like so https://github.com/alphagov/govuk_publishing_components/blob/7372183409497ce80237f0ed950bde9b697b3db2/app/views/govuk_publishing_components/components/docs/back_link.yml#L20

app/views/govuk_publishing_components/components/_chat_entry.html.erb
```diff
@@ -7,12 +7,23 @@
...
+  component_helper.add_data_attribute({ module: "ga4-link-tracker" }) unless disable_ga4
+
+  unless disable_ga4
+    ga4_link_data = {}
+    ga4_link_data[:ga4_link] = {
+      event_name: "navigation",
+      type: "callout",
+      section: "GOV.UK Chat entry",
+  }.to_json
+  end
```
AshGDS, line 26: Can you put the `ga4_link` straight into the object? E.g.

```ruby
ga4_link_data = {
  ga4_link: {
    # attributes in here
  }
}.to_json
```

AshGDS (changes requested):
Looks good 👍  Have some minor suggestions, and this will also need a changelog entry.

app/views/govuk_publishing_components/components/docs/chat_entry.yml
JamesCGDS: Thanks for spotting that! Have updated

app/views/govuk_publishing_components/components/_chat_entry.html.erb
JamesCGDS: Cool, have updated

JamesCGDS:
Thanks for the review @AshGDS - have made those changes now 👍

AshGDS (approved):
LGTM 👍

Issue #4202, opened 2024-09-05
Search title is cut off at certain screen sizes


ch-ncolley:
On certain screen sizes the .UK in GOV.UK is cut off from the Search heading/label.

I noticed the example in the component guide doesnt use this nested heading approach so maybe this is why it was missed.

<img width="657" alt="GOV.UK Search screenshot with the .UK from GOV.UK is cut off on small" src="https://github.com/user-attachments/assets/ebabd171-cc4e-42cf-92e4-e9abde463117">

## Steps to reproduce
1. Go to https://www.gov.uk/search
2. Resize browser window to 650px

andysellick:
Thanks for raising @ch-ncolley. This was a regression caused by a recent fix to a related problem. Will hopefully be able to deploy soon 👍

PR #4204, opened 2024-09-05
Fix search label width issue

andysellick:
## What
Change the label in the search component to put the `max-width: 68%` onto it only when it appears 'inline' as fake placeholder text.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/4202

## Visual Changes
Before | After
----- | -----
Screenshot 2024-09-05 at 10 55 10 | Screenshot 2024-09-05 at 10 55 23


AshGDS (approved):
Thanks for fixing, LGTM 👍

ch-ncolley:
Nice work mate 👍🏻

PR #4209, opened 2024-09-09
Hide printed URLs when printing tables

matthillco:
## What
When printing tables, hide the URLs generated by the global print styles in govuk-frontend. [Trello](https://trello.com/c/zuT5SiYJ/324-hide-printed-urls-when-printing-tables)

## Why
When links are printed, the print styles in govuk-frontend append the URL of the link so that they are more useful to the reader. But these links can break the layout of tables and prevent all columns from printing.

### BEFORE
image

### AFTER
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Personally I'd squash the two commits together, as there's only one change, but otherwise good 👍

PR #4210, opened 2024-09-09
Make links to parts in `document_list` component respect `remove_underline` setting

csutter:
The links to parts/"subdocuments" on the document list component are currently always rendered without underlines, regardless of whether `remove_underline` is actually set on the component itself. This is probably because it's only used in Finder Frontend's search results page, which always uses `remove_underline`.

## What
- Make links to parts render with underlines unless `remove_underline` is set on the component
- Remove custom CSS for removing underlines when `remove_underline` is set, and just use `govuk-link--no-underline` instead

## Why
We are starting to implement a UI overhaul of the search results page which has underlines on all links in the document list component, including those for parts.

### Before
image

### After
image
image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Hide printed URLs when printing tables ([PR #4209](https://github.com/alphagov/govuk_publishing_components/pull/4209))
 * Fix action link component print styles ([PR #4213](https://github.com/alphagov/govuk_publishing_components/pull/4213))
 * Improve the password input component custom text feature ([PR #4211](https://github.com/alphagov/govuk_publishing_components/pull/4211))
+* Make links to parts in `document_list` component respect `remove_underline` setting ([PR #4210](https://github.com/alphagov/govuk_publishing_components/pull/4176))
 
 ## 43.1.0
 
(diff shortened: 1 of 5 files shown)


csutter:
> This is probably because it's only used in Finder Frontend's search results page, which always uses `remove_underline`.

See [Github Code Search](https://github.com/search?q=org%3Aalphagov+remove_underline+NOT+repo%3Aalphagov%2Fgovuk_publishing_components&type=code) for uses of this component – all in Finder Frontend or search-related internal apps (some of which are decommissioned anyway)

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Make links to parts in `document_list` component respect `remove_underline` setting ([PR #4210](https://github.com/alphagov/govuk_publishing_components/pull/4176))
```
andysellick, line 12: Nitpick: can this match the PR name? Feel free to change either of them, whichever is easier.

andysellick (approved):
Good fix for a long standing issue, and good solution - I didn't know about the `govuk-link--no-underline` class 👏

PR #4213, opened 2024-09-09
Fix action link component print styles

matthillco:
## What
Re-enable the printing of the action-link component and apply consistent improvements to the printed URLs (reducing size, placing on their own line, forcing all colours to black) [Trello](https://trello.com/c/cVgisHQw/312-fix-action-link-component-print-styles)

## Why
As part of the improvements to print styles, we prevented the `action-link` component from printing at all as generally speaking printing links isn’t very helpful as they’re mainly interactive elements. However we noticed this caused problems printing certain pages, particularly the homepage Popular section which now displays nothing when printed as the entire block is composed of `action-link` components.

## Visual Changes

Before | After
------ | ------
image | image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
@@ -178,3 +178,22 @@
     }
   }
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-action-link {
+    * {
+      color: $govuk-print-text-colour !important; // stylelint-disable-line declaration-no-important
+    }
+
+    &::before {
+      display: none;
+    }
+
+    .gem-c-action-link__link {
+      &::after {
+        display: block;
+        font-size: 12pt;
+      }
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


andysellick:
@matthillco unrelated to the change here, did you know you can use tables in markdown to put your before/after screenshots side by side? e.g.

```
Before | After
------ | ------
[img1] | [img2]
```

andysellick (commented):
Code seems fine but a couple of the component variations still have quite light text when printed, can you check?

Screenshot 2024-09-10 at 09 37 01

matthillco:
> Code seems fine but a couple of the component variations still have quite light text when printed, can you check?
> 
@andysellick  Oops, an oversight. I've sorted that now and also changed the stylelint important rule to be a single line rather than a block.

app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss
```diff
@@ -178,3 +178,23 @@
     }
   }
 }
+
+@include govuk-media-query($media-type: print) {
+  .gem-c-action-link {
+    &,
```
andysellick, line 184: I'm not sure this line is needed, since the line after seems to cover all the bases - there isn't any text within `.gem-c-action-link` that isn't already wrapped in a sub element (i.e. the `*`).

matthillco: I hadn't noticed that, thanks. Now fixed.

andysellick (approved):

PR #4214, opened 2024-09-09
Use component wrapper in hint component

kevindew:
## What

This adds the component wrapper to the hint component in order to provide additional options to a hint.

## Why
The primary motivation of this was that we want to dynamically change the value of a hint in GOV.UK Chat and thus want to set a .js-class to act as a selector [1] (private repo) and instead are using a custom id to work around it.

Looking at other components it seemed that using the component wrapper was a more conventional approach than just utilising the shared helper to add only classes.

[1]: https://github.com/alphagov/govuk-chat/pull/660#discussion_r1750067107

## Visual Changes

None

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/hint.yml
+++ b/app/views/govuk_publishing_components/components/docs/hint.yml
@@ -2,6 +2,7 @@ name: Form hint text
 description: Use hints for any form fields.
 govuk_frontend_components:
   - hint
+uses_component_wrapper_helper: true
 accessibility_criteria: |
   All text must have a contrast ratio higher than 4.5:1 against the background colour to meet [WCAG AA](https://www.w3.org/TR/WCAG20/#visual-audio-contrast-contrast)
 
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/hint.yml
```diff
@@ -25,3 +26,9 @@ examples:
     data:
       text: "العربيَّة"
       right_to_left: true
+  with_js_class:
```
andysellick, line 29: You're very welcome to add an explicit documentation of this feature, but just to let you know that the options for the component wrapper are automatically added to the top of this guide page when `uses_component_wrapper_helper` is included.

andysellick (approved):

app/views/govuk_publishing_components/components/docs/hint.yml
kevindew: That's a good shout. I'll remove it. I added a test for it already so there's something to stop a regression.

PR #4217, opened 2024-09-11
Set aria-label text in govuk_logo.html to "GOV.UK"

MartinJJones:
## What
Update the govuk_logo.html.erb template to set the arial-label attribute value to "GOV.UK" and remove support for custom values to be passed in.

## Why
Help make future updates to the logo a little easier, ideally in the future we would simply be able to copy and past the required SVG code into the `govuk_logo.html.erb` template with no further changes required.

[Trello card](https://trello.com/c/N7N2TaNP/2807-refine-crownlogocrownonly-accessibility-approach-m)


hannalaakso (approved):
Thanks Martin, this is a nice tidy up 👍 

For reference, [the GOV.UK Design System also hardcodes the value of this aria-label](https://github.com/alphagov/govuk-frontend/blob/88dc4af4888baaaacddffb96f901cacc337bb363/packages/govuk-frontend/src/govuk/components/header/template.njk#L14) so it makes sense for us to do the same.

PR #4219, opened 2024-09-11
Fix cookie expiration date potentially relying on user's timezone

AshGDS:
## What /  Why

- We were using `Date.now()` as our way of instantly expiring cookies, but based on testing in the browser, the cookie would expire based on the UTC timezone rather than the user's timezone.  I believe `Date.now()` was grabbing the user's timezone which may have been causing issues if their timezone was ahead of UTC.
- It looks like the cookie would persist for an period of time depending on how far their timezone is ahead of UTC. An assumption of the worst case scenario is that their timezone could be 14 hours ahead of UTC, so certain cookies may have persisted for that long.
- I believe this is what is causing some weird cookie behaviour on [this GA4 card](https://trello.com/c/5TNoJ3rt/842-issue-with-absent-client-ids-related-to-postal-voteservicegovuk-redirect-links):  - I think on load our code is trying to delete this cookie to be extra safe, but it's doing that by making it expire at the `Date.now()` time. Currently we're an hour ahead of UTC, so I think this was setting an empty GA cookie to expire an hour in the future instead of deleting instantly.

As the screenshot shows, the two GA4 cookies should not exist when consent hasn't been given, but because their expiry date was set to `Date.now()` they were set to expire an hour in the future (as we're in BST at the moment, so 1 hour ahead of UTC - this screenshot was taken at 14:30 BST and UTC would've been 13:30.)

image

- Therefore, I've set the cookies to expire 1st January 1970, just as that's the simplest value to pass through to `new Date()`.
- I separated the expiration code into a function as it was difficult to test when it was nested within `deleteCookie`.  `deleteCookie` caused false positives as it seems to have two passes of cookie deletion, so the other deletion pass was causing my expiration code tests to falsely pass, as the expiration code wasn't actually the thing that was deleting the cookie, it was the other line in that function.

## Visual Changes

None

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
@@ -202,12 +202,14 @@
 
   window.GOVUK.deleteCookie = function (cookie) {
     window.GOVUK.cookie(cookie, null)
+    window.GOVUK.expireCookie(cookie)
+  }
 
-    if (window.GOVUK.cookie(cookie)) {
-      // We need to handle deleting cookies on the domain and the .domain
-      document.cookie = cookie + '=;expires=' + new Date() + ';'
-      document.cookie = cookie + '=;expires=' + new Date() + ';domain=' + window.location.hostname + ';path=/'
-    }
+  window.GOVUK.expireCookie = function (cookie, value = '') {
+    // We need to handle deleting cookies on the domain and the .domain
+    var thePast = new Date(0) // 0 = 0 seconds since UTC started (1970/01/01)
+    document.cookie = cookie + '=' + value + ';expires=' + thePast + ';'
+    document.cookie = cookie + '=' + value + ';expires=' + thePast + ';domain=' + window.location.hostname + ';path=/'
   }
 
   window.GOVUK.deleteUnconsentedCookies = function () {
--- a/app/assets/javascripts/govuk_publishing_components/lib/cookie-settings.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/cookie-settings.js
@@ -107,7 +107,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     try {
       documentReferrer = document.referrer || new URL(document.referrer).pathname
     } catch (e) {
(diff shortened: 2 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/lib/cookie-settings.js
```diff
@@ -107,7 +107,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     try {
       documentReferrer = document.referrer || new URL(document.referrer).pathname
     } catch (e) {
-      console.error('Error grabbing referrer for cookie settings', window.location, e)
+      console.warn('Error grabbing referrer for cookie settings', window.location, e)
```
AshGDS, line 110: I've changed this to warn, otherwise it makes the cookie page look like the JS is crashing when you visit the page without a referrer.

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
```diff
@@ -201,13 +201,18 @@
   }
 
   window.GOVUK.deleteCookie = function (cookie) {
```
andysellick, line 203: This all looks fine but I'm trying to get my head around the logic of it, bear with me.

Originally, to delete a cookie, we would call `window.GOVUK.cookie` with the cookie name and null, which I think clears the content of the cookie. Then an `if` statement would check if that cookie existed (which presumably always was the case) and then within it set the expiry date for that cookie. Now, we have the `if` first, followed by the call to `window.GOVUK.cookie`.

I guess I have two questions - firstly, is the `if` statement really needed, and secondly what difference does changing the order of these two things make? (if the answer to the first is 'no' then I think the second question is redundant).

andysellick (commented):
Change looks sensible, I just have a question about the logic.

app/assets/javascripts/govuk_publishing_components/lib/cookie-functions.js
AshGDS: @andysellick Good point - I don't think the `if statement` is needed, so have removed and put the order back.

andysellick (approved):

PR #4222, opened 2024-09-13
Update metatags for GA4

andysellick:
## What
Updates GOV.UK page metatags used by GA4 to make them more consistent. See individual commits for details.

Must be deployed with https://github.com/alphagov/slimmer/pull/341

## Why
This is some final tidying up post GA4 launch, to make things more consistent and clear.

## Visual Changes
None.

Trello card: https://trello.com/c/lIVmnFpY/336-page-view-enhancement-some-meta-tags-start-with-analytics

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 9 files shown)


andysellick:
Good point @AshGDS, thanks, have updated as suggested.

PR #4223, opened 2024-09-13
Stop redacting dates in GA4 tracking on GOVUK search pages

AshGDS:
## What / Why
- Removes date redaction from the `standardiseSearchTerm` function, - this function ensures the search term tracks the same across multiple parameters, so dates should no longer be redacted wherever we grab a search term.
- Also removes date redaction from the `location`, `query_string`, and `search_term` values in the PageView tracker.
- The PAs have requested this, as some of the filters use dates, and I believe there was a discussion at some point about dates being useful for search tracking. https://trello.com/c/RYSxnFZ1/841-remove-date-redaction-from-search-terms.

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -11,6 +11,7 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
 
     init: function (referrer) {
       if (window.dataLayer) {
+        this.stripDates = !this.getSearchTerm()
         var data = {
           event: 'page_view',
           page_view: {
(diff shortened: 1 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -78,7 +78,9 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
     },
 
     getLocation: function () {
-      return this.PIIRemover.stripPIIWithOverride(this.stripGaParam(document.location.href), true, true)
+      // We don't want to remove dates on search pages.
+      var stripDates = !this.getSearchTerm()
```
andysellick, line 82: This line gets repeated later - would it be possible to store this value on initialisation?

andysellick (approved):
Looks good, just one non-blocking small suggestion.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
AshGDS: Thanks @andysellick - good point, i've changed it so that it's only called once. Are you able to re-review?

andysellick (approved):

PR #4224, opened 2024-09-13
New class to collapse columns for print

matthillco:
## What
Provide a new helper class to "collapse" row/column layouts when printing. This class forces all floated columns in the associated row to print at full page width. [Trello](https://trello.com/c/MTJDEqDS/325-investigate-generic-print-styles-for-govuk-frontend-layout)

## Why
When printing various page types, the 2/3rds + 1/3rds column layout causes issues with cropped content and a generally poor printing experience because the full page isn't used. By selectively collapsing floated columns, we can avoid most of these issues and improve the printing experience by using the full width of the page.

### Before/After page type: GUIDE
| Before    | After |
| -------- | ------- |
| image  | image |

### Before/After page type: DETAILED_GUIDE
| Before    | After |
| -------- | ------- |
| image| image|

### Before/After page type: HTML_PUBLICATION (Portrait, no change}
| Before    | After |
| -------- | ------- |
| image| image|

### Before/After page type: HTML_PUBLICATION (Landscape}
| Before    | After |
| -------- | ------- |
| image| image|

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/component_support.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/component_support.scss
@@ -12,3 +12,5 @@
 @import "govuk_publishing_components/components/mixins/media-down";
 @import "govuk_publishing_components/components/mixins/margins";
 @import "govuk_publishing_components/components/mixins/css3";
+
+@import "govuk_publishing_components/lib/print_support";
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,6 +16,7 @@
 * Remove 100 character limit on search results ([PR #4230](https://github.com/alphagov/govuk_publishing_components/pull/4230))
 * Add GA4 redaction to GWF and GB EORI numbers ([PR #4227](https://github.com/alphagov/govuk_publishing_components/pull/4227))
 * Add files for secondary navigation: ([PR #4229](https://github.com/alphagov/govuk_publishing_components/pull/4229))
+* New class to collapse columns for print ([PR #4224](https://github.com/alphagov/govuk_publishing_components/pull/4224))
 
 ## 43.1.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
@@ -1,0 +1,27 @@
+// A helper class for removing column layouts when printing.
+// This should be applied to row/column layouts, by adding the
+// class to elements with a `govuk-grid-row` class. The columns
+// will then print at the full width of the page.
+// Note that only the immediate child columns will be affected.
+// If a child column also contains a row/column layout of
+// further elements, these will be unaffected and will retain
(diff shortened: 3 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/lib/print_support.scss
```diff
@@ -0,0 +1,29 @@
...
+    > .govuk-width-container {
+      margin: 0;
+      max-width: none;
+    }
+
+    > [class*="govuk-grid-column"] {
+      position: static !important;
+      float: none !important;
+      clear: both !important;
+      width: 100% !important;
```
andysellick, line 25: Just want to check, is `!important` needed for all of these?

andysellick (commented):
Looks good, only one question, also probably worth a CHANGELOG entry.

app/assets/stylesheets/govuk_publishing_components/lib/print_support.scss
matthillco: Without checking every possible layout/view I can't be sure, but the `position:static` needs to trump the `position:sticky` used in html publication on the contents list which currently has higher specificity. 

Given that, it seemed likely there might be other places where rules like this might exist in other layouts, so I felt using `!important` on these properties would ensure we're not caught out. I don't see the issue with using `!important` in these print styles, since they are only used for print, they're neatly encapsulated and they are at the bottom of our CSS specificity tree.

andysellick: Okay, but for reference I've tested this and none of the `!important`s are needed. Position static doesn't need to override position sticky, because sticky doesn't apply to print.

I think I'd rather do what we know is needed, then test this on other layouts and iterate it as we go if necessary.

app/assets/stylesheets/govuk_publishing_components/component_support.scss
```diff
@@ -12,3 +12,5 @@
 @import "govuk_publishing_components/components/mixins/media-down";
 @import "govuk_publishing_components/components/mixins/margins";
 @import "govuk_publishing_components/components/mixins/css3";
+
+@import "govuk_publishing_components/lib/print_support";
```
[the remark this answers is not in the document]
matthillco: How am I only learning this now! Thank you 👍🏻

app/assets/stylesheets/govuk_publishing_components/lib/print_support.scss
andysellick: Personally I wouldn't rely on the emulated print view in dev tools. I appreciate it's easier to play around with style changes but it's usually way off what the actual print appearance is.

matthillco: That was my point :) This is more for others who may be using emulated views and wondering why the contents-list is overlaying the rest of the page. It's a useful fix, so why can't we include it?

andysellick: That's fine 👍

PR #4225, opened 2024-09-16
Fix inconsistent search label appearance

andysellick:
## What
Fix the label styles in the search component. The label was appearing broken in some pages in the component guide, although still appearing correctly in the actual homepage.

- label stylesheet was only being included if wrap_label_in_a_heading was true
- custom gem label class for white on dark colour wasn't overriding standard govuk-frontend label class colour

## Why
Appearance of the component varied considerably within the component guide. I think this is because on some pages the label styles were already being included by other instances of the component on the page.

## Visual Changes
Fixes the following inconsistencies, so all pages now appear correctly (first one was correct already).

Component guide page | Appearance
----------------------- | ----------
Main page `/search` | Screenshot 2024-09-16 at 11 38 07
Preview all `/search/preview/` | Screenshot 2024-09-16 at 11 39 10
Detail page `/search/homepage` | Screenshot 2024-09-16 at 11 39 48
Detail preview page `/search/homepage/preview` | Screenshot 2024-09-16 at 11 40 19

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_search.scss
@@ -229,7 +229,7 @@ $large-input-size: 50px;
   }
 }
 
-.gem-c-search__label--white {
+.govuk-label.gem-c-search__label--white {
   color: govuk-colour("white");
 }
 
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):

csutter (approved):
🎉

PR #4229, opened 2024-09-16
Add files for secondary navigation:

davidtrussler:
## What
Add secondary navigation component. 

Files added
- ERB
- YML
- SCSS
- Rspec test

## Why
This component currently exists only on Whitehall (see [here](https://github.com/alphagov/whitehall/blob/main/app/views/components/_secondary_navigation.html.erb)). Since it is now needed on other apps (e.g. Mainstream Publisher) this is to make it more widely available. 

## Visual Changes
This is a new component - screenshot below. 

Screenshot 2024-09-16 at 15 57 14


spec/components/secondary_navigation_spec.rb
```diff
@@ -0,0 +1,41 @@
...
+          label: "Nav item 3",
+          href: "#",
+        },
+      ],
+    })
+
+    assert_select ".gem-c-secondary-navigation", count: 1
+    assert_select ".gem-c-secondary-navigation__list-item", count: 3
+    assert_select ".gem-c-secondary-navigation__list-item--current", count: 1
+    assert_select ".gem-c-secondary-navigation__list-item--current", text: "Nav item 1"
```
AshGDS, line 39: Possibly worth adding an `aria-current` check in here as well

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,20 @@
+<%
+  add_gem_component_stylesheet("secondary-navigation")
+
+  id ||= nil
+  items ||= []
```
AshGDS, line 5: Is it worth using the component wrapper here? I think the `shared_helper` would be needed on this component as well?

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,20 @@
+<%
+  add_gem_component_stylesheet("secondary-navigation")
+
+  id ||= nil
```
AshGDS, line 4: This `id` is not exposed in the `.yml` example, and is also untested.

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
```diff
@@ -0,0 +1,20 @@
...
+        item_classes = %w( gem-c-secondary-navigation__list-item )
+        item_classes << "gem-c-secondary-navigation__list-item--current" if item[:current]
+        item_aria_attributes = { current: "page" } if item[:current]
+      %>
+      <%= tag.li class: item_classes do %>
+        <%= link_to item[:label], item[:href], class: "govuk-link govuk-link--no-visited-state gem-c-secondary-navigation__list-item-link", data: item[:data_attributes], aria: item_aria_attributes %>
+      <% end %>
+    <% end %>
+  <% end %>
+<% end %>
```
AshGDS, line 20: Is it worth wrapping the render in an `if items` statement, so that there isn't a scenario where an empty nav/ul is rendered? (I appreciate this may be a very rare edge case)

AshGDS (commented):
Looks good, just had a few questions/comments 👍

jon-kirwan (commented):
Looks good to me. Just a small comment to update the changelog.

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
davidtrussler: Hi @AshGDS - not sure what the component wrapper is that you are referring to here. And also what the `shared_helper` does in this case.

AshGDS: @davidtrussler The component wrapper is something built by Andy to standardise the way data attributes / classes / aria tags are added to the parent HTML tag of the component. The shared helper AFAIK allows margin bottom to be applied to the component for styling purposes. 

An example of both helpers exist in the password input component I added - see lines 9 to 16, as well as line 40: https://github.com/alphagov/govuk_publishing_components/blob/eacb1572137f5678ffe4dcbe376d75a3162bf8b4/app/views/govuk_publishing_components/components/_password_input.html.erb#L9

You would then need to reflect the existence of these helpers in your `.yml` file - you would add `uses_component_wrapper_helper: true` to the file, and give an example of your component with margin bottom:  https://github.com/alphagov/govuk_publishing_components/blob/eacb1572137f5678ffe4dcbe376d75a3162bf8b4/app/views/govuk_publishing_components/components/docs/password_input.yml#L41

A test of the margin functionality in your spec file would be needed as well.

Component wrapper docs: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component-wrapper-helper.md

Shared helper docs: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md#shared-helper

Though not sure these are necessary for all components - may be worth asking Andy.

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
davidtrussler: Good point, added an id in the `yml` and added a test for it. 
At the same time we noticed the two tests that were there could really be reduced to one since there was some duplication going on.

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
davidtrussler: Thanks for that, Ash. Have now added the wrapper as suggested.

spec/components/secondary_navigation_spec.rb
davidtrussler: Good spot: have added that.

app/views/govuk_publishing_components/components/_secondary_navigation.html.erb
davidtrussler: Yes, probably rare but worth it - have now add that.

davidtrussler:
> Looks good to me. Just a small comment to update the changelog.

Thanks Jon, have done that.

AshGDS (approved):
Nice, thanks 👍

Issue #4208, opened 2024-09-06
Super navigation header visual bug on mobile when text scale is increased


AshGDS:
On smartphones, it's possible to increase the font size of the entire operating system in the settings. Some stats suggest it's quite common to increase the font size e.g. https://dev.to/sigute/accessible-text-scaling-for-android-1ham . If you increase your font size in Android, Firefox automatically increases the font scale of pages. On Chrome however, you need to go to Settings > Accessibility in the app and increase the font scale for web pages.

When the font scale is increased, this creates a visual bug with our menu button as it moves onto the grey background.

You can replicate this on Firefox Desktop by enabling "Zoom Text Only" in the browser settings, then making your browser width a mobile size and zooming in.

![Screenshot_20240903_140351_Firefox](https://github.com/user-attachments/assets/99874370-18ff-474e-be08-e94d244bc209)

PR #4232, opened 2024-09-17
Fix homepage super navigation buttons when text scale is increased

AshGDS:
## What / Why

- Add a blue background to the Menu / Search buttons, so they don't blend into the grey background for users that increase their text scale / have a screen resolution that is >= 300px in width (mobile users)
- Fixes https://github.com/alphagov/govuk_publishing_components/issues/4208

### Before
image
image

### After
image
image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -290,6 +290,8 @@ $after-button-padding-left: govuk-spacing(4);
 // after it has been toggled closed if the blue background has
 // been enabled
 .gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background {
+  background-color: $govuk-brand-colour;
+
   &:focus:not(:focus-visible) {
     &::after {
       background: none;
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -606,6 +608,10 @@ $after-button-padding-left: govuk-spacing(4);
 // Styles for search toggle button.
 .gem-c-layout-super-navigation-header__search-toggle-button {
   background: none;
+
+  &--blue-background {
```
jon-kirwan, line 612: Just a small suggestion - I don't think we have any style guidelines on this, but these types of selectors can be a bit tricky to search for. We could change it to `gem-c-layout-super-navigation-header__search-toggle-button--blue-background` though I realise that might be a bit verbose so no worries if not.

jon-kirwan (approved):
It looks good to me and have approved (I just added one small suggestion around the ampersand selector).

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
AshGDS: Thanks @jon-kirwan - good suggestion, will do this and then merge.

PR #4234, opened 2024-09-18
removed unused cookies in cookie category

sairamya93:
## What
 
To remove 'govuk_not_first_visit' and 'govuk_browser_upgrade_dismissed' 

## Why
These are no longer being used and we don’t actually present a banner of this nature any more. We should remove references to it from the govuk_publishing_components gem
https://trello.com/c/lR7DOoLn/341-we-refer-to-cookies-that-are-no-longer-in-use

## Visual Changes
Ensure that references to it are removed from [Details about cookies on GOV.UK](https://www.gov.uk/help/cookie-details)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


andysellick (approved):
Looks good. Can you also add an entry to the CHANGELOG? This is a small change, but it think it warrants an entry.

PR #4236, opened 2024-09-18
Make deploying meta tags changes easier

andysellick:
## What / why
- recently updated the names of some of the metatags that GA4 reads for page views
- but it's possible that the changes to those metatags won't be appearing for a while on a lot of pages, as they probably need republishing first
- so for now let's read either the new or the old metatag (which will contain the same information) and avoid breaking a part of the GA4 tracking
- haven't done the same for the strip dates/postcodes/etc. metatags as I think they're all handled internal to the gem
- follow up to changes made in https://github.com/alphagov/govuk_publishing_components/pull/4222

## Visual Changes
None.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -49,8 +49,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
             publishing_government: this.removeHyphensAndDowncase(this.getMetaContent('publishing-government') || this.getMetaContent('ga4-publishing-government')),
             political_status: this.getMetaContent('political-status') || this.getMetaContent('ga4-political-status'),
             primary_publishing_organisation: this.getMetaContent('primary-publishing-organisation'),
-            organisations: this.getMetaContent('organisations'),
-            world_locations: this.getMetaContent('world-locations'),
+            organisations: this.getMetaContent('organisations') || this.getMetaContent('analytics:organisations'),
+            world_locations: this.getMetaContent('world-locations') || this.getMetaContent('world-locations'),
```
AshGDS, line 53: Duplication of the same meta tag?

AshGDS (changes requested):
Looks good, needs a small fix

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
andysellick: D'oh! 🤦 

Fixed.

AshGDS (approved):

PR #4238, opened 2024-09-18
Allow custom name attribute on password input

yndajas:
## What

Allow custom name attribute on password input

## Why

On Signon, we use Devise for authentication. We customise the login screen in order to use the GOV.UK Design System, but in order for everything to work seemlessly with Devise (without undue hacking), we use the default Devise input names of `user[email]` and `user[password]`. Other apps may well have a similar need given that Devise is a pretty standard library


andysellick (approved):
👍

PR #4241, opened 2024-09-19
Remove 'inverse' background style from phase banner

jon-kirwan:
## What
Remove 'inverse' background style from phase banner

## Why

Components with an inverse option should not set a background colour as the inverse option is designed to work against a dark background, and we can't always know exactly what colour that background will be (it's usually GOV.UK dark blue, but it might be a dark green or something else in the future).

## Anything else

I've checked for instances where the banner is used with the inverse option [here](https://github.com/search?q=org%3Aalphagov+components%2Fcomponents%2Fphase_banner&type=Code). The only occurrence I found is in [Finder Frontend](https://github.com/alphagov/finder-frontend/blob/main/app/views/finders/_before_content.html.erb), which already applies a parent background style `app-before-content--inverse`, see [guidance and regulation](https://www.gov.uk/search/guidance-and-regulation?topic=c58fdadd-7743-46d6-9629-90bb3ccc4ef0&parent=foreign-commonwealth-office), so it won't cause any breaking changes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove left search variant of layout header ([PR #4239](https://github.com/alphagov/govuk_publishing_components/pull/4239))
+* Remove 'inverse' background style from phase banner ([PR #4241](https://github.com/alphagov/govuk_publishing_components/pull/4241))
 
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Nice work 👍

PR #4242, opened 2024-09-19
Update accessibility criteria in component docs

MartinJJones:
## What

Update the accessibility criteria for the breadcrumbs and step by step navigation header components, stating they should be placed before the `<main>` element to avoid failing WCAG 2.2 success criterion 2.4.1 Bypass Blocks (Level A). This is similar to the guidance provided in the "How it works" section of the [breadcrumbs component in the Design System](https://design-system.service.gov.uk/components/breadcrumbs/#how-it-works).

[Trello card](https://trello.com/c/6aPsnjO2/2860-update-gem-docs-to-say-that-breadcrumbs-should-be-placed-outside-main-container-s?filter=label:Frontend%20devs)

## Why

We found a number of places where the breadcrumbs were incorrectly placed inside the main container, which failed [WCAG 2.2 success criterion 2.4.1 Bypass Blocks (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks). Updating the dev docs should help developers when looking at how best to use the component.

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
+++ b/app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
@@ -6,9 +6,11 @@ body: |
   We recommend that if using the breadcrumbs for navigation purposes, you set `collapse_on_mobile` to `true` to make things more readable for mobile users. However, you can specify `collapse_on_mobile:``false` or remove the flag completely to stop this behaviour.
 shared_accessibility_criteria:
   - link
-accessibility_criteria:
+accessibility_criteria: |
   The breadcrumb links must have a text contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
   (this especially applies when [using the inverse flag](/component-guide/breadcrumbs/inverse)).
+
+  Always place breadcrumbs at the top of a page, before the `<main>` element. Placing them here means that the "Skip to main content" link allows the user to skip all navigation links, including breadcrumbs.
 accessibility_excluded_rules:
   - skip-link # This component is creating a reference to #content which is part of the layout
 display_html: true
--- a/app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
+++ b/app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
@@ -8,6 +8,8 @@ accessibility_criteria: |
   An earlier version of the component did not use a heading element &ndash; this failed WCAG 2.1 Success Criterion 1.3.1 ("Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.")
 
   An early version of the component contained a hidden skip link for keyboard and screen reader users, that jumped to the step by step navigation component in the sidebar (similar to the 'skip to content' link at the top of all GOV.UK pages). User testing suggested that rather than helping users it confused them, so this has been removed.
+
+  Always place the step by step navigation header at the top of a page, before the `<main>` element. Placing the component here means that the "Skip to main content" link allows the user to skip all navigation links, including the step by step navigation header.
 shared_accessibility_criteria:
   - link
 examples:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,10 +7,15 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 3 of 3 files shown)


app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
```diff
@@ -6,9 +6,11 @@ body: |
   We recommend that if using the breadcrumbs for navigation purposes, you set `collapse_on_mobile` to `true` to make things more readable for mobile users. However, you can specify `collapse_on_mobile:``false` or remove the flag completely to stop this behaviour.
 shared_accessibility_criteria:
   - link
-accessibility_criteria:
+accessibility_criteria: |
   The breadcrumb links must have a text contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
   (this especially applies when [using the inverse flag](/component-guide/breadcrumbs/inverse)).
+
+  Always place breadcrumbs at the top of a page, before the `<main>` element. Placing them here means that the "Skip to main content" link allows the user to skip all navigation links, including breadcrumbs. This relates to WCAG 2.2 success criterion [2.4.1 Bypass Blocks (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks).
```
unoduetre, line 13: Isn't this line too long?

unoduetre: Also, there are cases of breadcrumbs being used twice on the page, so maybe "always" is too strong? Maybe something like: "Always place the top-level breadcrumbs at the top of the page, before the \`\<main\>\` element... This does not apply to secondary breadcrumbs which are e.g. below the title."

app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
```diff
@@ -8,6 +8,8 @@ accessibility_criteria: |
   An earlier version of the component did not use a heading element &ndash; this failed WCAG 2.1 Success Criterion 1.3.1 ("Information, structure, and relationships conveyed through presentation can be programmatically determined or are available in text.")
 
   An early version of the component contained a hidden skip link for keyboard and screen reader users, that jumped to the step by step navigation component in the sidebar (similar to the 'skip to content' link at the top of all GOV.UK pages). User testing suggested that rather than helping users it confused them, so this has been removed.
+
+  Always place the step by step navigation header at the top of a page, before the `<main>` element. Placing the component here means that the "Skip to main content" link allows the user to skip all navigation links, including the step by step navigation header. This relates to WCAG 2.2 success criterion [2.4.1 Bypass Blocks (Level A)](https://www.w3.org/WAI/WCAG22/Understanding/bypass-blocks).
```
unoduetre, line 12: Isn't this line too long?

app/views/govuk_publishing_components/components/docs/breadcrumbs.yml
hannalaakso: > there are cases of breadcrumbs being used twice on the page

I don't _think_ we know if two sets of breadcrumbs on a page is a pattern we want to encourage, or at least we don't know enough about it to document it here yet. We do have the one known occurrence two sets of breadcrumbs on the HMRC internal manuals page but as Martin as pointed out separately, there might be an accessibility issue with that implementation which we'd like to investigate but that is out of scope here. The Design System doesn't document two sets of breadcrumbs on a page as a design pattern. 

I think the best thing to do here would be to document the standard use of breadcrumbs as Martin has proposed. If we find out later that in fact the use of two sets of breadcrumbs is something we'd like to encourage and we understand the accessibility implications of doing so, then we could come back to this piece of documentation to update it. Reaching out to the accessibility community sounds like a good idea, but maybe doing it as part of that separate card two investigate two sets of breadcrumbs on a page.

app/views/govuk_publishing_components/components/docs/step_by_step_nav_header.yml
MartinJJones: I've made this a little shorter by removing the links to the success criterion.

unoduetre (approved):
As discussed with @MartinJJones, the YAML files can be improved by using the operators like `>-` etc. as the lines are too long for effective editing and git tracking of the changes. As this PR is not directly related to that, this would need to be addressed separately.

Besides that separate issue, it looks good.

PR #4244, opened 2024-09-20
Remove breadcrumbs example from inverse_header docs

MartinJJones:
## What
Remove breadcrumbs example from inverse_header docs

[Trello card](https://trello.com/c/YFgXTbi2/2859-remove-the-breadcrumbs-option-from-the-inverse-header-docs-in-the-gem-s)

## Why
It's not possible to include the breadcrumbs inside the header component without ending up with a WCAG fail because the variant places it inside the <main> container since the header itself is inside the main.


unoduetre (approved):
Looks good to me, but needs a rebase as there is a conflict.

app/views/govuk_publishing_components/components/docs/inverse_header.yml
```diff
@@ -45,30 +41,6 @@ examples:
...
-              }
-            ],
-            inverse: true
-          } %>
-          <%= render "govuk_publishing_components/components/lead_paragraph", {
-            text: "Schools and academies, further and higher education, apprenticeships and other skills training, student funding, early years.",
-            inverse: true
-          } %>
-          <!-- end of example content -->
   with_link:
```
hannalaakso, line 44: This is actually an oversight on my part when I wrote the card for this, I think I could have suggested that we rename this link variant to:
> With link and paragraph

since the variant we're removing mentions using a paragraph. This one with the link doesn't mention it, even though it also contains a paragraph so it might be good to document that.

MartinJJones: Thanks Hanna, good suggestion, I've updated the text for the variation example to be "With paragraph and link"

hannalaakso (approved):

PR #4255, opened 2024-09-24
Add handling of for_static parameter

KludgeKML:
- original behaviour was to always include wrapper class unless show_account_layout or custom_layout was present (custom layout is as of yet not used). Alter behaviour so that default is not to have the wrapper, and it's only included if for_static is true.

## Why
layout_for_public does not behave like a standard layout - it's default is to include a wrapper element which is cut out by slimmer and replaced with the wrapper from the client app. If we try to use layout_for_public directly, this results in a double wrapper (or a single wrapper, but without the ability to add classes to it). Adding a parameter which maintains this behaviour (which can be added in Static: https://github.com/alphagov/static/pull/3440 ), but defaulting to not including the wrapper element allows other apps to use the layout component directly.

https://trello.com/c/COQYfYgn/347-allow-layoutforpublic-to-work-like-a-normal-layout

## Visual Changes

None expected

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add component wrapper helper to radio component ([PR #4366](https://github.com/alphagov/govuk_publishing_components/pull/4366))
 * Add component wrapper to feedback component ([PR #4351](https://github.com/alphagov/govuk_publishing_components/pull/4351))
 * Add component wrapper helper to error message component ([PR #4359](https://github.com/alphagov/govuk_publishing_components/pull/4359))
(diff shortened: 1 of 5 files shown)


andysellick (approved):
Probably need a CHANGELOG entry, but otherwise looks good 👍

KludgeKML:
Have approved the Percy visual regression - it's one added example (with the for_static parameter), then the other examples getting shrunk slightly because they don't have the for_static: true flag set, so they are missing the wrappers and main elements (the point of this PR).

PR #4256, opened 2024-09-25
Center text and icons vertically in the option select component

matthillco:
## What

Vertically centre align the toggle icons with the label in Option Selects. [Trello](https://trello.com/c/65zYsdtH/275-type-scale-finder-frontend-improve-spacing-and-alignment-in-search-filters-on-mobile)

## Why

The font size of this element increased with the typography changes deployed in a recent update to `govuk-frontend`. This means that the up/down icons no longer align vertically. This PR vertically centres the icons with the text for better vertical balance. A new container for the icon and button is also required so that the selected counter element can be styled independently.

This also supports the same change made in `finder-frontend` where search filter titles had the same tweak to icon positions. See the PR at https://github.com/alphagov/finder-frontend/pull/3334.

## Visual Changes

| Before    | After |
| -------- | ------- |
|image|image|

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_option-select.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_option-select.scss
@@ -76,11 +76,10 @@
 
 .gem-c-option-select__icon {
   display: none;
-  position: absolute;
-  top: 0;
-  left: 9px;
+  flex-shrink: 0;
   width: 30px;
   height: 40px;
+  margin-inline: 10px 4px;
   fill: govuk-colour("black");
 }
 
@@ -126,8 +125,14 @@
 .govuk-frontend-supported {
   .gem-c-option-select__heading {
     position: relative;
-    padding: 10px 8px 5px 43px;
-    margin: 0;
+    margin: 0 0 govuk-spacing(1);
+  }
+
+  .gem-c-option-select__toggle {
+    display: flex;
+    flex-direction: row-reverse;
+    justify-content: start;
+    align-items: center;
   }
 
   [aria-expanded="true"] ~ .gem-c-option-select__icon--up {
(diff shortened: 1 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_option-select.scss
```diff
@@ -76,11 +76,9 @@
 
 .gem-c-option-select__icon {
```
MartinJJones, line 77: One thing that could happen when switching to use flex for the layout is that the icon will shrink depending on the length of the text, for example:
Screenshot 2024-10-03 at 09 27 12

One possible option to stop the icon from shrinking is to add `flex: 0 0 30px`. I imagine the same CSS rule would then need to be added to the expander component in finder-frontend.

MartinJJones (commented):
Changes look good to me overall, and the approach seems consistent with those made to the expander component in finder-frontend.

Just one thing which is likely worth looking at further in relation to the icon size.

app/assets/stylesheets/govuk_publishing_components/components/_option-select.scss
matthillco: Ah good spot, thanks! I've added a `flex-shrink: 0` to solve this rather than your suggestion, as it avoids the need to update the number if the size of the icon changes. @MartinJJones

MartinJJones (commented):
Thanks for making the CSS change, this looks good to me 👍 

Sorry to be a pain, I just spotted one other small thing, it looks like the HTML used in `option-select-spec.js` will also need to be updated so they are in sync with the latest changes, would need to make the addition in the 2 sections below:

- https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/components/option-select-spec.js#L9-L11
- https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/components/option-select-spec.js#L31-L35

Everything else looks good to me, happy to approve after this change, thanks 👍

matthillco:
> Sorry to be a pain, I just spotted one other small thing, it looks like the HTML used in `option-select-spec.js` will also need to be updated so they are in sync with the latest changes [...]

This issue was discussed offline and we've agreed to leave the spec alone for now as we noticed other problems with it: addressing these in this PR is out of scope, so we'll note this in the backlog and work on that separately in the future.

MartinJJones (approved):

PR #4270, opened 2024-09-26
Use breadcrumbs based on organisation, if no parent linked

unoduetre:
## What

The Content Types Discovery team's audit uncovered that corporate_information pages lack a breadcrumb that connects back to the organisation.

Currently corporate information pages without the linked parent use taxon based breadcrumbs. An example would be the following page: https://www.gov.uk/government/organisations/government-actuarys-department/about

This PR fixes that, and if there is no parent linked, it uses the first linked organisation to generate breadcrumbs.

The logic is the following:
1. if parent is linked to, use it to generate breadcrumbs;
2. otherwise, if organisations are linked to, use the first one to generate breadcrumbs;
3. otherwise, if taxons are linked to, use them to generate breadcrumbs;
4. otherwise use the default breadcrumbs linking to homepage.

## Why

So that users can navigate back to the organisation to which this corporate information relates.

[Trello ticket](https://trello.com/c/lm0JyNVJ/2832-fix-corporate-information-pages-should-have-a-parent-of-an-org-page-l)

## Testing

https://government-frontend-pr-3347.herokuapp.com/government/organisations/government-actuarys-department/about
https://government-frontend-pr-3347.herokuapp.com/government/organisations/department-for-business-energy-and-industrial-strategy/about/personal-information-charter
https://government-frontend-pr-3347.herokuapp.com/government/organisations/standards-and-testing-agency/about
https://government-frontend-pr-3347.herokuapp.com/government/organisations/hm-courts-and-tribunals-service/about

### Before

Screenshot 2024-09-26 at 15-48-50 About us - Government Actuary's Department - GOV UK

### After

Screenshot 2024-09-26 at 15-49-17 About us - Government Actuary's Department - GOV UK


MartinJJones (commented):
I've had a quick look and have one thought on the updated breadcrumbs, using the page below as an example:

www.gov.uk/government/organisations/government-actuarys-department/about

The breadcrumbs are updated from `Home > Government` to `Home > Government Actuary's Department` which definitely feels like a nice a improvement.

It does appear that this page also has a grandparent, "Departments, agencies and public bodies", I think we should check with a content designer if this should also be included in the breadcrumbs as well, for example:

`Home > Departments, agencies and public bodies > Government Actuary's Department`

Unfortunately, I can't find the information about the higher levels in the content item, so getting this information would require either adding things to the content item (with the publishing team's help) or doing a separate call to the api to retrieve that, which seems like an overkill. If we want/can modify the content item in the first place, we can add a parent link instead. This would resolve the issue. So maybe we can ask the publishing team to add the parent link and in the meantime we can use this solution?

unoduetre:
> I've had a quick look and have one thought on the updated breadcrumbs, using the page below as an example:
> 
> [www.gov.uk/government/organisations/government-actuarys-department/about](http://www.gov.uk/government/organisations/government-actuarys-department/about)
> 
> The breadcrumbs are updated from `Home > Government` to `Home > Government Actuary's Department` which definitely feels like a nice a improvement.
> 
> It does appear that this page also has a grandparent, "Departments, agencies and public bodies", I think we should check with a content designer if this should also be included in the breadcrumbs as well, for example:
> 
> `Home > Departments, agencies and public bodies > Government Actuary's Department`

Unfortunately, I can't find the information about the higher levels in the content item, so getting this information would require either adding things to the content item (with the publishing team's help) or doing a separate call to the api to retrieve that, which seems like an overkill. If we want/can modify the content item in the first place, we can add a parent link instead. This would resolve the issue. So maybe we can ask the publishing team to add the parent link and in the meantime we can use this solution?

unoduetre:
Or maybe we could use something like Home > Organisations >  Government Actuary's Department
as in the Government Actuary's Department's page the breadcrumbs are Home > Organisations?

But would that apply to all corporate information pages, or would it require some additional logic?

unoduetre:
> Or maybe we could use something like `Home > Organisations > Government Actuary's Department` as in the Government Actuary's Department's page the breadcrumbs are `Home > Organisations`?

This change has now been applied.

unoduetre:
This is still worth considering: https://www.gov.uk/government/organisations/hm-courts-and-tribunals-service/about

In this case taxon-based breadcrumbs might be considered slightly better: https://www.gov.uk/government/organisations/hm-courts-and-tribunals-service/about

MartinJJones (commented):
Looks good to me from a frontend perspective

@hannako please can you help review the backend changes

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
```diff
@@ -55,6 +55,11 @@ def options(navigation)
             step_by_step: false,
             breadcrumbs: navigation.breadcrumbs,
           }
+        elsif navigation.content_is_a_corporate_information_page? && navigation.content_has_related_organisations?
```
hannako, line 58: It might be more readable if you use a similar approach to on line 63 where we have
 `elsif navigation use_taxon_breadcrumbs?` ie a single method `use_organisation_breadcrumbs?`

spec/components/contextual_breadcrumbs_spec.rb
```diff
@@ -97,6 +97,22 @@ def set_live_taxons(content_item)
...
+  end
+
+  it "renders organisation breadcrumbs if the content item is a corporate information page with no parent but with a linked organisation" do
+    content_item = example_document_for("corporate_information_page", "best-practice-about-page")
+    render_component(content_item:)
+    assert_select "a", text: "Home"
+    assert_select "a", text: "Organisations"
+    assert_select "a", text: "Intellectual Property Office"
+  end
+
```
hannako, line 115: Ah I hadn't clocked this. 
- Is it possible for a corporate information page to have no parent, but still have a link to an organisation? 
- Is it possible for a corporate information page to have a parent that is different to its linked organisation
- Is it possible for a corporate information page to be linked to more than one organisation?

spec/lib/govuk_publishing_components/presenters/breadcrumb_selector_spec.rb
```diff
@@ -0,0 +1,89 @@
...
+    it "does not raise exception" do
+      expect { subject }.not_to raise_error
+    end
+  end
+
+  describe "#breadcrumbs" do
+    context "when content item is corporate information page with parent" do
+      let(:example) { %w[corporate_information_page corporate_information_page_complaints] }
+
+      it "returns breadcrumbs based on ancestors" do
```
hannako, line 34: You're saying based on ancestors here, but refer to them as parent breadcrumbs in the other test. Could they both be the same pls

spec/lib/govuk_publishing_components/presenters/breadcrumb_selector_spec.rb
```diff
@@ -0,0 +1,89 @@
+require "spec_helper"
+
+RSpec.describe GovukPublishingComponents::Presenters::BreadcrumbSelector do
```
hannako, line 3: Thank you for adding a test! I can't believe there wasn't already a file!

spec/lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_organisations_spec.rb
```diff
@@ -0,0 +1,25 @@
+require "spec_helper"
+
+RSpec.describe GovukPublishingComponents::Presenters::ContentBreadcrumbsBasedOnOrganisations do
+  subject { described_class.new(content_item) }
+  let(:content_item) { GovukSchemas::Example.find("document_collection", example_name: "document_collection") }
```
hannako, line 5: Given that currently only one document type supports organisation breadcrumbs, I think we should use that schema example here, ie corporate_information_page.

hannako (commented):
Thanks for adding all these tests Matt 🏅 

I've requested a few tiny refactors. But my main question is on what valid corporate information pages look like

lib/govuk_publishing_components/presenters/contextual_navigation.rb
```diff
@@ -72,6 +76,10 @@ def content_has_curated_related_items?
         content_item.dig("links", "ordered_related_items").present? && content_item.dig("links", "parent").present?
       end
 
+      def content_has_related_organisations?
+        ContentItem.new(content_item).related_organisations.present?
```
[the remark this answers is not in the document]
unoduetre: This must match line 26 of lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_organisations.rb
So if this is to be changed, the other line needs to be changed as well.
It wouldn't be good, if we duplicate the code, as this wouldn't ensure the code is always matching.
This solution calls the same function `#related_organisations`. (Also this function contains additional logic which ensures the organisation we are interested in is always of the document type "organisation".
I know currently all organisations in links -> organisations are of "organisation" document type, so this is redundant right now.)

If we still wanted to change that, we'd  need to create a method somewhere which will be called by both `lib/govuk_publishing_components/presenters/contextual_navigation.rb and lib/govuk_publishing_components/presenters/content_breadcrumbs_based_on_organisations.rb which will ensure exactly the same list of methods is used.

lib/govuk_publishing_components/presenters/breadcrumb_selector.rb
unoduetre: Fixed.

spec/components/contextual_breadcrumbs_spec.rb
unoduetre: > Is it possible for a corporate information page to have no parent, but still have a link to an organisation?
Yes, that's the main problem this ticket resolves. E.g. https://www.gov.uk/api/content/government/organisations/government-actuarys-department/about

> Is it possible for a corporate information page to have a parent that is different to its linked organisation

TODO

> Is it possible for a corporate information page to be linked to more than one organisation?
Depending on what you mean by "possible".

I checked the schema and [this line](https://github.com/alphagov/publishing-api/blob/056cd207fad7f17aabb86a5f8ca0782258a55f96/content_schemas/dist/formats/corporate_information_page/frontend/schema.json#L131) just specified that it's frontend_links_with_base_path and [the definition of it](https://github.com/alphagov/publishing-api/blob/056cd207fad7f17aabb86a5f8ca0782258a55f96/content_schemas/dist/formats/corporate_information_page/frontend/schema.json#L492) is a kind of array that does not have any restriction on the number of entries. So in this sense, it's allowed by the schema.

On the other hand, I did a query in content store and it seems there are currently no records with more than one organisation linked. So in this sense, the answer is "no".

spec/lib/govuk_publishing_components/presenters/breadcrumb_selector_spec.rb
unoduetre: Fixed.

spec/components/contextual_breadcrumbs_spec.rb
hannako: Thanks for looking into this Mat - would you mind checking the code in Whitehall that creates corporate information pages - is it a model that has a relationship to an org? That would give us the confidence we need I think... everything else looks great.

unoduetre: Yes, here's the line: https://github.com/alphagov/whitehall/blob/dedeaee2f486fd6c52cbfdc4a1077174ba4b01f8/app/models/corporate_information_page.rb#L11

hannako (approved):
Thanks for addressing the notes, and for adding all the tests ⭐

PR #4273, opened 2024-10-01
Remove Modernizr

jon-kirwan:
## What

Remove Modernizr

## Why

Although Modernizr is included in the admin layout, which is utilised by various publishing applications, none of these applications seem to use the [included browser features](https://modernizr.com/download/?flexbox-flexboxtweener-history-printshiv-setclasses) (`flexbox`, `flexbox-tweener`, history API), except for the printshiv which enables HTML5 elements to be styled when printed in Internet Explorer versions 6-8.

## Anything else

Marked as a breaking change due to the potential effect on printing in older versions of IE. I recommend reviewing any printed pages that use the admin layout.


MartinJJones (approved):
Nice work, thanks for the link to the Modernizr build tool as well 👍

When it comes to the release, do you think this needs to be a breaking change because of the removal of the printshiv?

jon-kirwan:
> Nice work, thanks for the link to the Modernizr build tool as well 👍
> 
> When it comes to the release, do you think this needs to be a breaking change because of the removal of the printshiv?

Thanks again for the review 😎 Good point, I'll add it as a breaking change

PR #4275, opened 2024-10-02
Global print link class

matthillco:
## What
Two new helper classes for applying consistent formatting to printed links. These classes replace the same repeated CSS previously created as part of the print improvement work. [Trello](https://trello.com/c/7opRx4sU/336-add-a-global-link-print-style)

## Why
Most links inherit styles from `govuk-frontend`, including printing the contents of the `href`. This makes printed links more useful, but they can take up a lot of space on the page in the default font size and will print in blue by default. These helper classes sets the printed colour to black, reduce the font size, and improve the layout of the printed href.

## Visual Changes
Not applicable as the visual output is the same as before.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
@@ -162,21 +162,16 @@
   }
 }
 
-// stylelint-disable declaration-no-important
 @include govuk-media-query($media-type: print) {
   .gem-c-document-list__item {
     break-inside: avoid;
+
+    * {
+      color: $govuk-print-text-colour;
+    }
   }
 
-  .gem-c-document-list__item-title {
-    .govuk-link {
-      color: $govuk-print-text-colour !important;
-    }
-
-    .govuk-link::after {
-      display: block;
-      margin: 1mm auto;
-    }
+  .gem-c-document-list__item-context {
+    display: block;
   }
 }
-// stylelint-enable declaration-no-important
--- a/app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
@@ -25,3 +25,48 @@
     }
   }
 }
+
+// Two helper classes to ensure that all printed links
+// receive consistent formatting. Most links inherit
+// styles from govuk-frontend, including the printing
+// of the href that a link contains. This makes printed
+// links more useful, but they can take up a lot of space
+// on the page in the default font size and will print
+// in blue by default. These helpers set the printed colour
+// to black, reduce the font size, and improve the layout
+// of the printed href.
+//
+// .gem-print-link
+// ----------------------------------------------------------
+// Typically this class will be applied to existing
+// elements that have the `govuk-link` class but it can
+// also be used on other link elements.
+//
+// .gem-print-links-within
+// ----------------------------------------------------------
+// A variation of the previous print style, to be used on
+// parent elements that contain links with the `govuk-link`
+// class. This is typically only required for views where
+// the `govuk-link` element is not exposed in the markup
+// of the view as it's coming from a self-contained helper.
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-print-link,
+  .gem-print-links-within .govuk-link {
+    &,
+    &:link,
+    &:visited {
+      color: $govuk-print-text-colour !important;
+    }
+
+    &::after {
+      display: block;
+      margin: 1mm auto;
+      font-size: 12pt;
+      font-weight: normal;
+      color: $govuk-print-text-colour !important;
+    }
+  }
+}
+// stylelint-enable declaration-no-important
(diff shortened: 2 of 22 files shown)


app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
```diff
@@ -25,3 +25,48 @@
...
+// elements that have the `govuk-link` class but it can
+// also be used on other link elements.
+//
+// .gem-print-links-within
+// ----------------------------------------------------------
+// A variation of the previous print style, to be used on
+// parent elements that contain links with the `govuk-link`
+// class. This is typically only required for views where
+// the `govuk-link` element is not exposed in the markup
+// of the view as it's coming from a self-contained helper.
```
andysellick, line 51: Have you got some examples of where this is needed?

app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
```diff
@@ -25,3 +25,48 @@
...
+
+// stylelint-disable declaration-no-important
+@include govuk-media-query($media-type: print) {
+  .gem-print-link,
+  .gem-print-links-within .govuk-link {
+    &,
```
andysellick, line 57: What does this line do?

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
```diff
@@ -96,8 +96,6 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey");
   }
 
   .gem-c-attachment__link.govuk-link {
-    word-break: break-word;
```
andysellick, line 99: I'm not sure how removing this line relates to this change, can you clarify?

app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
```diff
@@ -162,21 +162,16 @@
...
+    * {
+      color: $govuk-print-text-colour;
     }
+  }
 
-    .govuk-link::after {
-      display: block;
-      margin: 1mm auto;
-    }
+  .gem-c-document-list__item-context {
```
andysellick, line 174: Can you explain this change?

app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
matthillco: Yep, it's required in `inset-text` component which doesn't expose any links directly in the view, I added it here: https://github.com/alphagov/govuk_publishing_components/pull/4275/files#diff-666dbf87f00cea8384ab451ab812f206f7c93e02915ab4edd1bcfcf664040de9R13

The whole view looks like this:
```<%
  add_gem_component_stylesheet("inset-text")

  id ||= "inset-text-#{SecureRandom.hex(4)}"
  margin_top ||= 6
  margin_bottom ||= 6

  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new({
    margin_top: margin_top,
    margin_bottom: margin_bottom
  })

  classes = %w[gem-c-inset-text govuk-inset-text]
  classes << shared_helper.get_margin_top
  classes << shared_helper.get_margin_bottom
%>
<%= tag.div id: id, class: classes do %>
  <% if defined? text %>
    <%= text %>
  <% elsif block_given? %>
    <%= yield %>
  <% end %>
<% end %>```

app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
matthillco: Do you mean the & on line 57? This is a 'failsafe' to ensure consistency between directly styled links such as `a {rules}` and links styled only with pseudo selectors `a:link {rules}`, because they don't behave the same unless you set them all.

app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
matthillco: I'm not sure either, I think that was deleted by mistake (although I didn't notice any new problems). I'll revert it.

app/assets/stylesheets/govuk_publishing_components/components/_document-list.scss
matthillco: For a document list with context, the context can wrap in a strange way depending on the length of the printed URL: 
image

Setting the context explicitly to `display: block` forces it to clear the space and start on a new line:
image

matthillco: Lines 169-171 force all the contents of this component to black. It's a new adjustment that I discovered was necessary because some elements of some variants render in a lighter grey otherwise, for example `.gem-c-document-list__attribute` is set to print in `#505a5f` unlesss we override it. I used the `*` selector to ensure that everything inside this component will be printed in black.

app/assets/stylesheets/govuk_publishing_components/lib/_print_support.scss
matthillco: Also, I've changed line 59 from `&:link:visited` to just `&:visited` as the simpler form seems fine in all modern browsers. For a long time I used the doubled-up form to ensure that old IE applied link styles consistently and it's an ingrained habit that I must stop doing by default...!

andysellick: Thanks.

andysellick (approved):
Just need a CHANGELOG entry. Also ideally this PR would also document these two new classes - I know that's a separate card, but this could feed into that.

PR #4282, opened 2024-10-04
Remove instances of ga4_tracking

AshGDS:
## What / Why

- Looks we accidentally left in `ga4_tracking: true` references after moving to `disable_ga4`.

## Visual Changes

None.


andysellick (approved):
Nice spot 👏

PR #4284, opened 2024-10-07
Add inverse option for organisation logo

andysellick:
## What
Adds an inverse option for the organisation logo component.

This is currently WIP and limited to one organisation. Since organisations have different logos, rolling out a universal inverse option would require inverse versions of every logo.

## Why
Part of some requirements.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
@@ -82,6 +82,21 @@
   }
 }
 
+.gem-c-organisation-logo--inverse {
+  .gem-c-organisation-logo__container {
+    border-color: govuk-colour("white");
+    @include govuk-link-style-inverse;
+  }
+
+  .gem-c-organisation-logo__crest--eo {
+    filter: brightness(0) invert(1);
+
+    &:focus {
+      filter: none;
+    }
+  }
+}
+
 @mixin crest($crest, $xpos:0, $ypos:0) {
   background: url("govuk_publishing_components/crests/#{$crest}_18px_x2.png") no-repeat $xpos $ypos;
   background-size: auto 32px;
--- a/app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
@@ -88,7 +88,7 @@
     @include govuk-link-style-inverse;
   }
 
-  .gem-c-organisation-logo__crest--eo {
+  .gem-c-organisation-logo__crest {
     filter: brightness(0) invert(1);
 
     &:focus {
--- a/app/views/govuk_publishing_components/components/docs/organisation_logo.yml
+++ b/app/views/govuk_publishing_components/components/docs/organisation_logo.yml
@@ -75,7 +75,7 @@ examples:
         name: "Prime Minister's Office,<br>10 Downing Street"
         url: '/government/organisations/prime-ministers-office-10-downing-street'
         brand: 'prime-ministers-office-10-downing-street'
-        crest: 'no10'
+        crest: 'eo'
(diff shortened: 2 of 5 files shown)


matthillco:
I just noticed this PR and thought I'd comment with a possible quick solution for the inverted icons. We might consider using either `filter: invert(1)` (or `background-blend-mode: difference`) for the icon. When a black icon is placed on a black background, these rules will make the icon appear white. Both of these are available for our target browsers. However I appreciate this might be considered a bit of a hack.

JamesCGDS:
Thanks for the suggestion @matthillco - I ended up using a variation of this (`filter: filter: brightness(0) invert(1);`) because `filter: invert(1)` was also inverting the white link text, making it black. Adding `brightness(0)` seemed to resolve things though.

app/views/govuk_publishing_components/components/docs/organisation_logo.yml
```diff
@@ -182,3 +182,13 @@ examples:
         brand: cabinet-office
         crest: 'single-identity'
       inline: true
+  on_a_dark_background:
+    data:
+      organisation:
+        name: "Prime Minister's Office,<br>10 Downing Street"
+        url: '/government/organisations/prime-ministers-office-10-downing-street'
+        brand: 'prime-ministers-office-10-downing-street'
+        crest: 'no10'
```
AshGDS, line 191: I assume you want `eo` instead of `no10` here?

spec/components/organisation_logo_spec.rb
```diff
@@ -71,4 +71,9 @@ def component_name
     render_component(organisation: { name: "Name", url: "/some-link" }, inline: true)
     assert_select "a.gem-c-organisation-logo__container--inline"
   end
+
+  it "renders on a dark background" do
+    render_component(organisation: { name: "Name", url: "/some-link" }, inverse: true)
+    assert_select ".gem-c-organisation-logo.gem-c-organisation-logo--inverse"
+  end
```
AshGDS, line 78: Nitpick: I also like to add another test that checks that the class doesn't exist when `inverse` is `false`, just so we're ensured classes are definitely tied to their boolean being `true`, and I think technically is better code coverage. (I appreciate the other examples don't have this type of test, so feel free to ignore this)

app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
```diff
@@ -82,6 +82,21 @@
...
+  }
+
+  .gem-c-organisation-logo__crest--eo {
+    filter: brightness(0) invert(1);
+
+    &:focus {
+      filter: none;
+    }
+  }
+}
```
AshGDS, line 98: Is there a reason this class is on `gem-c-organisation-logo__crest--eo` instead of just `gem-c-organisation-logo__crest`? Currently, this would mean the variant only works with the `eo` crest.

AshGDS (commented):
Looks good, had a few questions, and would need a changelog entry 👍

app/assets/stylesheets/govuk_publishing_components/components/_organisation-logo.scss
JamesCGDS: Oops good spot - I'll change that

spec/components/organisation_logo_spec.rb
JamesCGDS: No problem, can add that in

app/views/govuk_publishing_components/components/docs/organisation_logo.yml
JamesCGDS: I think we might want to retain `no10` here - have messaged you on Slack.

JamesCGDS:
Thanks for the review @AshGDS 👍 Have addressed the comments and added a changelog entry - ready for a re-review when you get a mo

AshGDS (approved):
LGTM 👍

PR #4286, opened 2024-10-07
LUX 4.0.27

andysellick:
## What
Upgrade our LUX Real User Monitoring script to the latest version, as per https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#updating

Currently can't do the last step locally (checking it against the debug tool) owing to problems with my local setup, so that still needs checking.

## Why
I'm actually a bit behind on this one, it's been out for more than a week. We need to keep up to date or it'll stop working.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
+++ b/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
@@ -336,7 +336,7 @@
     return str;
   }
 
-  var VERSION = "4.0.26";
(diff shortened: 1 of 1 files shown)


MartinJJones (approved):
Changes look good to me 👍 

I also tested the changes locally following the [documentation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#testing), there were no errors with the output when run through the SpeedCurve RUM Debug Parser tool

PR #4287, opened 2024-10-07
Add component wrapper helper to error alert component

AshGDS:
## What / Why

- Adds the component wrapper to the error alert component
- The component contains a `tabindex` attribute, but the component wrapper didn't support this, so I've added support for `tabindex` in the component wrapper.
- https://trello.com/c/qH4NyWJw/364-add-component-wrapper-to-more-components

## Visual Changes

None.

### DIFF
--- a/lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
+++ b/lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
@@ -12,6 +12,7 @@ module GovukPublishingComponents
         check_lang_is_valid(@options[:lang]) if @options.include?(:lang)
         check_open_is_valid(@options[:open]) if @options.include?(:open)
         check_hidden_is_valid(@options[:hidden]) if @options.include?(:hidden)
+        check_tabindex_is_valid(@options[:tabindex]) if @options.include?(:tabindex)
       end
 
       def all_attributes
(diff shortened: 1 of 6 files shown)


lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
```diff
@@ -146,6 +153,16 @@ def check_hidden_is_valid(hidden_attribute)
         end
       end
 
+      def check_tabindex_is_valid(tabindex_attribute)
+        return if tabindex_attribute.blank?
+
+        tabindex_attribute = tabindex_attribute.to_s
+
+        unless /^-?[0-9]+$/.match?(tabindex_attribute)
```
andysellick, line 161: I thought tabindex values should only be positive integers or `-1`, but this seems to be allowing other values, including numbers with underscores? Can you clarify?

andysellick (commented):
Looks good, just one confusion.

lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
AshGDS: @andysellick MDN says you can pass any negative value:

> A negative value (the exact negative value doesn't actually matter, usually tabindex="-1") means that the element is not reachable via sequential keyboard navigation.

The underscored numbers are a result of our Rubocop linting, as that's how it wants to represent large numbers. I have checked how this renders in our frontend and it appears as a regular number.

However checking MDN, the maximum positive number it says is possible is for `tabindex` is `32767`. Do you want me to incorporate this check into the code?

andysellick: No, that's fine 👍

PR #4288, opened 2024-10-08
Standardise inverse options for components

andysellick:
## What
Update components that have an option for displaying them on a dark background to use `inverse` as the option name.

## Why
We're doing some unrelated work with the action link component and the fact that action link uses something other than `inverse` causes extra code to be required.

For reference, here are all the components I could find prior to starting this PR with an `inverse` option and what they called it:

- action link - **light_text**
- breadcrumbs - inverse
- contextual breadcrumbs - inverse
- govspeak - inverse
- heading - inverse
- lead paragraph - inverse
- metadata - inverse
- phase banner - inverse
- search with autocomplete - **on_govuk_blue**
- search - **on_govuk_blue**
- title - inverse
- translation nav - inverse

## Visual Changes
Should be no actual visual changes apart from some of the names of the examples in the component guide.


richardTowers (approved):
Nice! Thanks for cleaning up this little inconsistency.

AshGDS (approved):
LGTM 👍

PR #4291, opened 2024-10-08
Reskin tabs for the landing page

andysellick:
## What
Adds a visual variant for the tabs component. This overrides a lot of the existing CSS on the tabs.

For mobile we're going to just hide the tabs themselves, and make all the tab content visible and stack it.

Note that the content of the tabs is going to be provided by the application, hence why there's no padding in the content.

## Why
Needed for the landing page.

## Visual Changes
New option looks like this:

Screenshot 2024-10-08 at 15 00 20

Trello card: https://trello.com/c/YCUbimh9/50-build-featured


andysellick:
Closing as the designs have changed.

PR #4292, opened 2024-10-08
Add variants required for square blue icon share links

AshGDS:
## What / Why

- Add variants required for square blue icon share links
- Reasoning: https://trello.com/c/LNOqGMgZ/49-build-social-media-links
- The existing component styles use `position: absolute` for the icons, which doesn't work nicely when we need them square. This is because the absolutely positioned icons don't affect the width/height of the DOM, so adding square background styles just ended up in overlapping, fiddly overflows, and made it difficult to align the text in the centre of the icons.
- Therefore, I've used `display: flex` along with `inline-block` to get the styles we want.
- ~~I added a `min-width` at a certain breakpoint to make the reflow look nicer, but this is specific to the page we're going to put this variation on. (i.e. because we know the social icons / share link text that will be on the page, we can make the shorter sized social links look nicer when they stack next to longer social links by adding a `min-width`.) Of course that CSS rule is not very helpful if this variation was used on other pages that are using an arbitrary number of social links.~~

## Visual Changes

Multiple examples added, so will need a Percy review.

image

image

image

image

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
@@ -83,3 +83,35 @@ $column-width: 9.5em;
     display: block;
   }
 }
+
+.gem-c-share-links--square-icons {
+  .gem-c-share-links__list-item {
+    padding-left: 60px;
+    padding-top: 12px;
+    margin-bottom: 30px;
+  }
+
+  .gem-c-share-links__link-icon {
+    background-color: govuk-colour("dark-blue");
+    color: govuk-colour("white");
+    padding: govuk-spacing(2);
+    margin-right: govuk-spacing(2);
+  }
+
+  .gem-c-share-links__link:hover {
+    .gem-c-share-links__link-icon {
+      background-color: govuk-colour("black");
+    }
+  }
+
+  .gem-c-share-links__link:focus {
+    .gem-c-share-links__link-icon {
+      background-color: govuk-colour("black");
+    }
+  }
+
+  .gem-c-share-links__label {
+    color: govuk-colour("black");
+    @include govuk-font(19, $weight: bold);
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add variants required for square blue icon share links ([PR #4292](https://github.com/alphagov/govuk_publishing_components/pull/4292))
+
 ## 44.2.0
 
 * Upgrade to version 5.7.0 of govuk-frontend ([PR #4298](https://github.com/alphagov/govuk_publishing_components/pull/4298))
--- a/app/views/govuk_publishing_components/components/_share_links.html.erb
+++ b/app/views/govuk_publishing_components/components/_share_links.html.erb
@@ -8,6 +8,7 @@
   ga4_extra_data ||= {}
   stacked ||= false
   columns ||= false
+  square_icons ||= false
 
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
@@ -15,6 +16,8 @@
   classes = %w(gem-c-share-links govuk-!-display-none-print)
   classes << "gem-c-share-links--stacked" if stacked
   classes << "gem-c-share-links--columns" if columns
+  classes << "gem-c-share-links--square-icons" if square_icons
+
   classes << brand_helper.brand_class
 
   data_attributes ||= {}
@@ -110,7 +113,7 @@
                   <path fill="currentColor" d="M36.92,0A36.92,36.92,0,1,0,73.84,36.92,36.92,36.92,0,0,0,36.92,0ZM56.3,48.27a1.42,1.42,0,0,1-1.42,1.42h-19v9.18l-9.18-9.18H19a1.43,1.43,0,0,1-1.43-1.43V20.52A1.43,1.43,0,0,1,19,19.09H54.88a1.43,1.43,0,0,1,1.42,1.42Z"/>
                 </svg>
               <% end %>
-            </span><%= link_text %><% end %>
+            </span><%= tag.span(class: "gem-c-share-links__label") do %><%= link_text %><% end %><% end %>
         </li>
       <% end %>
     </ul>
--- a/app/views/govuk_publishing_components/components/docs/share_links.yml
+++ b/app/views/govuk_publishing_components/components/docs/share_links.yml
@@ -256,3 +256,39 @@ examples:
           icon: 'youtube'
         }
       ]
+  with_square_icons:
+    data:
+      square_icons: true
+      columns: true
+      links: [
+        {
+          href: '/twitter-share-link',
+          text: 'Twitter',
+          icon: 'twitter'
+        },
+        {
+          href: '/instagram-share-link',
+          text: 'Instagram',
+          icon: 'instagram'
+        },
+                {
+          href: '/flickr-share-link',
+          text: 'Flickr',
+          icon: 'flickr'
+        },
+        {
+          href: '/facebook-share-link',
+          text: 'Facebook',
+          icon: 'facebook'
+        },
+        {
+          href: '/youtube-share-link',
+          text: 'YouTube',
+          icon: 'youtube'
+        },
+        {
+          href: '/other-share-link',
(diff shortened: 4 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
andysellick, line 1: I think this new option `square_blue_icons ` is trying to do two things at once (change the icons and change the layout) and I think it would be better to separate layout from appearance in this component.

Did you try using the `stacked` option? It seems to do almost exactly what's needed here in terms of layout - and then you could apply the blue icons as a separate option (allowing more flexibility by separating out the layout change from the icon change).

AshGDS: Thanks for reviewing @andysellick - I'm not sure exactly what you mean here, as `stacked` would make the links vertical no matter what, but they need to be along one row for desktop. If you meant `columns: true`, I believe as that's based on an assumed width for the links it doesn't work as well here as this variation is a lot bigger - unless you want me to make a modification of that variation to try and work here?

Also, looking into it, `display: flex` wasn't actually needed - so I have removed that from the code.

I can remove the `min-width` rule if needed too, if that's what's causing concern, as it was a visual enhancement rather than something necessary. It would then look like this when stacked: 

## With min-width
image

## Without
image

If you're referring to the icons themselves, do you mean to have a `relative_square_icons: true` variation, and then a colour option that would have the colours `dark blue / white` passed to it?

andysellick: Sorry, I meant the `columns` option 🤦

AshGDS: @andysellick No worries - I think using `columns` makes the spacing of the share links a bit less even, as it's using `display: grid` with a `min-width` to achieve this, so it makes some of the links wider than needed, which is noticeable on this variation. Also the `column` variation's `min-width` gets a bit tight for our share links when you resize the page (the icons nearly butt right up against each other before falling onto a new row,) so the CSS would probably need to be adjusted. I think this variation would work if the links/icons were smaller, but as they're larger it isn't as compatible. I think the squareness of the icons also shows the uneven spacing a bit clearer if we used this, than if we were using our standard icons. Therefore do you think it's still worth going with this sort of technique?

See the Flickr link on the examples for the spacing differences:

## With columns option

image

## Without

image

andysellick: The design is probably a bit open to interpretation but yeah they do seem to be variable width. I still think we should separate the layout from the presentation though. Is it worth doing one option for the new icons, and another one for the layout - maybe try some new layout with flexbox? Seems like a good candidate over grid.

Screenshot 2024-10-09 at 15 45 32

matthillco: Note that the icons look broken in Firefox with 'Zoom Only Text' option and the font size increased: 
image

It's easily fixed if you change the `display: inline-block` to `display: inline-flex` in `.gem-c-share-links__link-icon--square-blue`. The flexbox defaults will then auto-center the icon:

image

matthillco (commented):
I don't have anything to add to the discussion about the layout with Andy, the overall solution seems robust and rational to me. I did play with it a bit to see if we could remove the 180px min-width which is a bit of a magic number but it seems required: no combination of flex alignments/adjustments would solve it.

I just have one comment about the icons breaking in Firefox zoom text only mode. https://github.com/alphagov/govuk_publishing_components/pull/4292#discussion_r1794297771

AshGDS:
Thanks for the suggestions @andysellick - It should be ready for another review. I've refactored the implementation completely to try and separate layout and presentation. Hopefully I haven't overdone it, as I've ended up with five different variants. However I couldn't think of an other way of representing all the changes needed without creating the five variants. All of the new variants combine together to give us the icon styles we need for our other project. I've added them as separate commits to try and make it easier to review.

@matthillco Thanks for looking at the Firefox zoom as I didn't check that, I've incorporated your suggested fix into the work 👍 .

andysellick:
For the flexbox layout, I was wondering about this kind of thing...

```CSS
.gem-c-share-links--flexbox {
  .gem-c-share-links__list {
    display: flex;
    flex-wrap: wrap;

      flex-direction: column;
    }
  }

  .gem-c-share-links__list-item {
    flex: 1 1 0px;
  }
}
```

Which gives the below - expands to fill the available width on a wide screen, but allows for overlap with extreme content, then collapsing to a single column on mobile.

Mobile | Tablet | Desktop 
------ | ------ | -----
Screenshot 2024-10-10 at 15 55 15 | Screenshot 2024-10-10 at 15 55 08 | Screenshot 2024-10-10 at 15 54 57

andysellick:
@AshGDS I think there might be too many options here now. I'd do one for the layout, and one for the new icon style.

AshGDS:
Thanks @andysellick - I have moved back to two variants and pushed a new commit to review.

Thanks for writing that code - on the square icons it behaves a bit differently because my icons were `inline-block` which meant they influenced the width of the links, and therefore led to scenarios like this: 
image

Therefore I've reverted back to `position: absolute` to try and better accommodate using your flex options. Using `position: absolute` which was simpler than I initially thought, apologies. This works well, but when two icons are on a row on their own it will look like the below screenshot - are we OK with this?:

image

As I'm back using `position: absolute` for the icons, we can also go back to just using the existing `columns: true` functionality, which makes it collapse like the non-square icons would. The minor flaw with `columns: true` is that the text will wrap even if there is still space for the text to the right.  

image

I've left it with the absolute positioned icons, and using the existing `columns: true` functionality for now as this solution requires the least amount of new code, but let me know if you think I should look at using `flex`.

Apologies if I'm over complicating this...maybe I'm worrying too much about accommodating the `Anything else that might be included that could have quite a long name` example, when it will likely never show up in production.

AshGDS:
Thanks @matthillco 👍  - it's a good idea, I did a small spike however and `space-evenly` lead to scenarios like this: 
image

I appreciate your idea was just a starting point though. My original implementation was a more complex iteration of `display: flex`, but because I had used a `min-width` on the list elements to make the wrapping of items look nicer, and I wasn't using the existing `columns: true` functionality, it was the cause of some of the back and forth in the PR.

I think the problem is trying to make elements of varying widths act and wrap perfectly, and therefore any solution we use is going to have a trade off with either alignment or text wrapping.

I agree that position absolute is fiddly, though as it's the existing functionality of the share links I'm more inclined to keep using it just to make this PR as small as possible. I'm also leaning towards just go with the existing functionality of `columns` for this reason.

matthillco:
> Thanks @matthillco 👍 - it's a good idea, I did a small spike however and `space-evenly` lead to scenarios like this: image
> 
> I appreciate your idea was just a starting point though. My original implementation was a more complex iteration of `display: flex`, but because I had used a `min-width` on the list elements to make the wrapping of items look nicer, and I wasn't using the existing `columns: true` functionality, it was the cause of some of the back and forth in the PR.
> 
> I think the problem is trying to make elements of varying widths act and wrap perfectly, and therefore any solution we use is going to have a trade off with either alignment or text wrapping.
> 
> I agree that position absolute is fiddly, though as it's the existing functionality of the share links I'm more inclined to keep using it just to make this PR as small as possible. I'm also leaning towards just go with the existing functionality of `columns` for this reason.

app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
```diff
@@ -83,3 +83,37 @@ $column-width: 9.5em;
     display: block;
   }
 }
+
+.gem-c-share-links--extra-spacing {
```
andysellick, line 87: I think given that this extra spacing is really only needed for the new icon set I'd suggest combining the two into a single option - I don't think we need them to be separate.

I'd also suggest this because the extra spacing option on the normal icons makes the text align to the bottom of the icons, which I'm not sure looks right.

andysellick (commented):
I think I'm basically happy with this now, apart from one suggestion.

app/assets/stylesheets/govuk_publishing_components/components/_share-links.scss
AshGDS: Thanks @andysellick - I've merged the two down to one now. Have pushed it as a separate commit but will squash eventually.

andysellick (approved):
Well done - sorry this got so complicated 👏

PR #4293, opened 2024-10-08
Document list: Add option for equal item spacing

csutter:
## What
The items in a document list currently have slightly more padding above the separator than below. This adds an option to have equal spacing on both sides.

## Why
For the new design for GOV.UK Search, we want the spacing above and below the separator to be equal.

### Before
image

### After
image


MartinJJones (approved):
Change looks good to me, nice work 👍

PR #4301, opened 2024-10-11
Add chart component

andysellick:
## What
Adds a chart component.

Based on a component from `content-data-admin` originally added here: https://github.com/alphagov/content-data-admin/pull/9

TODO:

- check the ONS guidance on accessible graphs to see what improvements we can make
- fix the console error when the graph is hovered

## Visual Changes
New component:

Screenshot 2024-10-17 at 10 55 07

Trello card: https://trello.com/c/gvLYQCUp/57-build-statistics-charts

### DIFF
--- a/app/views/govuk_publishing_components/components/_chart.html.erb
+++ b/app/views/govuk_publishing_components/components/_chart.html.erb
@@ -1,0 +1,153 @@
+<%
+  add_gem_component_stylesheet("chart")
+  add_gem_component_stylesheet("table")
+  add_gem_component_stylesheet("details")
+  require "chartkick"
+
+  chart_id ||= false
+  chart_label ||= "data"
+  chart_width ||= "90%"
+  table_id ||= false
+  table_direction ||= "horizontal"
+  from ||= false
+  to ||= false
+  percent_metric ||= false
+  rows ||= []
+  keys ||= []
+
+  # http://strftime.org/
+  Date::DATE_FORMATS[:table_format] = '%-d %b %Y'
+
+  # https://developers.google.com/chart/interactive/docs/reference#dateformatter
+  chart_date_format = 'd MMM Y'
+
+  if percent_metric
+    maximum_y = 100
+  elsif rows.any?
+    row_maximums = rows.map { |row| row[:values].compact.max }
+    overall_maximum = row_maximums.max
+
+    if overall_maximum == 0
+      # set max value to 3 if data is all 0 avoid negative numbers on Y axis
+      maximum_y = 3
+    elsif overall_maximum < 10
+      # for low numbers let line hit top to avoid a fractional max value
+      maximum_y = overall_maximum
+    else
+      # for higher numbers give breathing space to line
+      maximum_y = overall_maximum * 1.3
+
+      if overall_maximum > 99999
+        chart_width = "80%"
+      end
+    end
+  end
+
+  Chartkick.options[:html] = '<div id="%{id}"><noscript><p>Our charts are built using javascript but all the data is also available in the table.</p></noscript></div>'
+  # config options are here: https://developers.google.com/chart/interactive/docs/gallery/linechart
+  chart_library_options = {
+    legend: 'none',
+    chartArea: { width: chart_width, height: '85%' },
+    curveType: 'none',
+    tooltip: { showColorCode: true, isHtml: true, trigger: 'both'},
+    crosshair: { orientation: 'vertical', trigger: 'both', color: '#ccc' },
+    hAxis: {
+      viewWindow: { min: from, max: to },
+      textStyle: { color: '#000', fontName: 'nta', fontSize: '12' },
+      format: chart_date_format
+    },
+    vAxis: {
+      viewWindow: { min: 0, max: maximum_y },
+      format: '#,###,###',
+      textStyle: { color: '#000', fontName: 'nta', fontSize: '12' }
+    },
+    pointSize: 0,
+    series: {
+      0 => { lineWidth: 2, color: '#2b8cc4' },
+      1 => { lineWidth: 2, lineDashStyle: [10, 2], color: '#454a4c' }
+    }
+  }
+  if rows.length > 0 && keys.length > 0
+    chart_format_data = rows.map do |row|
+      {
+        name: row[:label],
+        linewidth: 10,
+        data: keys.zip(row[:values])
+      }
+    end
+  end
+%>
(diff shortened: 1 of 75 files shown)


AshGDS (commented):
Code wise it looks good, but I just had a few questions about the mobile styles, as the Axis Labels seem to disappear on mobile, and when you click on a data point, it covers other data points so you can't click on them. Are we OK with this on mobile?

AshGDS:
Also, I appreciate that this was a difficult component to take on, so thanks for doing this 👍

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,138 @@
...
+      <%= render "govuk_publishing_components/components/heading", {
+        text: chart_heading,
+        heading_level: chart_heading_level,
+        margin_bottom: 2,
+      } %>
+    <% end %>
+
+    <div class="gem-c-chart__chart" id="<%= chart_id %>">
+      <div class="gem-c-chart__accessibility-message">
+        This chart is a visual representation of the data available in the <a href="#<%= table_id %>">table</a>.
```
jon-kirwan, line 72: Should this be added to the locale(s)?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,174 @@
+name: "Chart"
+description: "The chart component presents a data chart and a tabular version of the same data"
+part_of_admin_layout: true
+body: |
+  This component takes a set of data and presents it as a line graph and a table with one or more rows and lines.
+  The chart relies upon chartkick and renders using JavaScript, so the table is provided as a fallback for a lack of JavaScript,
+  an accessible view on the data for screenreaders, and a simple view of the raw data for all users.
+
+  The `overview` option can be used to provide a explanation for screen reader users of what the graph shows.
```
jon-kirwan, line 9: Sorry really minor one! I think 'provide a explanation' should be 'provide an explanation'.

jon-kirwan (commented):
It looks good to me. I've tested across various screens / devices, with VoiceOver, with JS disabled and seems to work well for me. I just added two small comments / questions.

andysellick:
> Code wise it looks good, but I just had a few questions about the mobile styles, as the Axis Labels seem to disappear on mobile, and when you click on a data point, it covers other data points so you can't click on them. Are we OK with this on mobile?

Yeah, the axis labels disappearing as the graph gets narrower is a bit unhelpful, although this seems to be the behaviour of the graph so I think we're stuck with it. I think the problem might be that (certainly with some data) it wouldn't be possible to fit the labels (certainly for the Y axis) in with the graph on mobile - imagine if the Y scale was `0` -> `1,000,000,000`. Hopefully the presence of the tooltips will help. I'm also anticipating that once we start using this component we might find some alternative ways of configuring it to better suit what it gets used for.

On the point about the tooltip, I've just pushed a commit to switch to using HTML tooltips (instead of SVG). It seems now that it's much easier to dismiss the tooltip by tapping elsewhere on the page, so hopefully this has improved things.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Yes, thanks good spot. Have added.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Good spot, have fixed 👍

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
...
+  rows ||= []
+  keys ||= []
+  chart_overview ||= nil
+  hide_legend ||= false
+
+  chart_id = "chart-id-#{SecureRandom.hex(4)}"
+  table_id = "table-id-#{SecureRandom.hex(4)}"
+
+  require "chartkick"
+  Chartkick.options[:html] = '<div id="%{id}"><noscript><p class="govuk-body">Our charts are built using JavaScript but all the data is also available in the table below.</p></noscript></div>'
```
AshGDS, line 21: Also, does this text need editorial approval?

andysellick:
> It seems that you can't right click the element - is that intentional?

It's not intentional from us, but appears to be intentional in Google charts, see this stack overflow for info: https://stackoverflow.com/questions/20266832/why-google-charts-are-without-right-click-menu

> The charts seem to have an `aria-label` that is `A chart.` - is the lack of customisation for this intentional? The table has an aria label as well.

This is generated automatically by Google charts. It doesn't seem to be possible to directly customise this text - there are some complicated JS solutions online, which I'm hesitant to implement at this point as I think there's not much more helpful information we can provide at this level - the chart should already have a title, and an accessible description.

Interestingly while digging into this I realised the component has two copies of the table - the one that we create inside the details component, and another one that Google charts automatically creates and appends after the chart. It's visually hidden, so still readable by screenreaders. I'll make a note to see if I can hide it somehow.

> The HTML seems to have `col"` attributes in it, which isn't a valid attribute - is this a chartkick thing we can ignore, or a bug?

I'm not sure, where are these exactly?

> The ONS charts fallback to an image when JS is disabled, so that may be worth noting for the future improvements to this component.

Noted, thanks.

AshGDS:
@andysellick No worries - thanks for looking into those.

I don't think the `col"` thing is actually an issue - instead I think it is a bug with the W3C HTML validator I used on the chart, my bad.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Possibly. If it's alright with you I think we'll look at that later when we're reviewing the pages as a whole.

spec/components/chart_spec.rb
```diff
@@ -0,0 +1,91 @@
...
+  end
+
+  it "renders the correct table data horizontally" do
+    data[:table_direction] = "horizontal"
+    render_component(data)
+    assert_select ".govuk-table__body .govuk-table__header:nth-child(1)", text: "2017"
+    assert_select ".govuk-table__cell--numeric", 26
+    assert_select ".govuk-table__header", 14
+    assert_select "td:first", text: "5"
+    assert_select "td:last", text: "121"
```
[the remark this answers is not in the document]
andysellick: Sure, have added.

AshGDS (approved):
Nice work 👍

PR #4305, opened 2024-10-15
New options for contents-list component

matthillco:
## What

Two new options for the `contents-list` component:
- `alternative_line_style`
- `title`

These can be used to format the list as a navigation block on landing pages. [Trello](https://trello.com/c/l9yaSdmV/53-build-side-navigation)

## Why

Landing pages have a navigation element in a sidebar. This navigation element can use the `contents-list` component. However, there are additional requirements which require new options:
- the active state is shown with a vertical border and bold blue text.
- increased vertical spacing between lines
- the Contents title can be replaced by an alternative title.

## Visuals

| title    | alternative_line_style |
| -------- | ------- |
| image  | image    |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -50,7 +50,7 @@
   padding-left: $contents-spacing;
   padding-right: $contents-spacing;
 
-  & span::before {
+  .gem-c-contents-list__list-item-dash::before {
     content: "—";
     position: absolute;
     left: 0;
--- a/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
@@ -27,11 +27,11 @@
 }
 
 .gem-c-contents-list__link {
-  .gem-c-contents-list__list-item--parent > & {
-    font-weight: bold;
-  }
-
   @include govuk-template-link-focus-override;
+}
+
+.gem-c-contents-list__list-item--parent > .gem-c-contents-list__link {
+  font-weight: $govuk-font-weight-bold;
 }
 
 .gem-c-contents-list__list-item {
@@ -73,6 +73,29 @@
   .lte-ie8 & .gem-c-contents-list__link {
     display: inline-block;
     vertical-align: top;
+  }
+}
+
+.gem-c-contents-list--alternative-line-style {
+  .gem-c-contents-list__title {
+    margin-bottom: govuk-spacing(3);
+  }
+
+  .gem-c-contents-list__list-item {
+    padding: govuk-spacing(2) 0;
+    margin: 0;
+  }
+
+  .gem-c-contents-list__list-item-dash::before {
+    display: none;
+  }
+
+  .gem-c-contents-list__list-item--active {
+    padding: govuk-spacing(1) govuk-spacing(2);
+    margin: govuk-spacing(1) 0;
+    border-left: 4px solid govuk-colour("blue");
+    color: govuk-colour("blue");
+    font-weight: $govuk-font-weight-bold;
   }
 }
 
--- a/app/views/govuk_publishing_components/components/_contents_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_contents_list.html.erb
@@ -4,6 +4,7 @@
   cl_helper = GovukPublishingComponents::Presenters::ContentsListHelper.new(local_assigns)
   underline_links ||= false
   format_numbers ||= false
+  alternative_line_style ||= false
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
   title_fallback = t("components.contents_list.contents", locale: I18n.locale, fallback: false, default: "en")
@@ -22,6 +23,7 @@
   local_assigns[:aria] ||= {}
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-contents-list #{brand_helper.brand_class}")
+  component_helper.add_class("gem-c-contents-list--alternative-line-style") if alternative_line_style
   component_helper.add_data_attribute({ module: "ga4-link-tracker" }) unless disable_ga4
   component_helper.add_aria_attribute({ label: t("components.contents_list.contents") }) unless local_assigns[:aria][:label]
   component_helper.add_role("navigation")
--- a/app/views/govuk_publishing_components/components/_contents_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_contents_list.html.erb
@@ -40,7 +40,7 @@
       <% index_link = 1 unless disable_ga4 %>
       <% contents.each do |contents_item| %>
         <li class="<%= cl_helper.list_item_classes(contents_item, false) %>" <%= "aria-current=true" if contents_item[:active] %>>
-          <span aria-hidden="true"></span>
+          <span class="gem-c-contents-list__list-item-dash" aria-hidden="true"></span>
           <% link_text = format_numbers ? cl_helper.wrap_numbers_with_spans(contents_item[:text]) : cl_helper.clean_string(contents_item[:text])
             unless disable_ga4
               ga4_data[:event_name] = cl_helper.get_ga4_event_name(contents_item[:href]) if contents_item[:href]
@@ -58,7 +58,7 @@
             <ol class="gem-c-contents-list__nested-list">
               <% contents_item[:items].each do |nested_contents_item| %>
                 <li class="<%= cl_helper.list_item_classes(nested_contents_item, true) %>" <%= "aria-current=true" if nested_contents_item[:active] %>>
-                  <span aria-hidden="true"></span>
+                  <span class="gem-c-contents-list__list-item-dash" aria-hidden="true"></span>
                   <%
                     unless disable_ga4
                       ga4_data[:event_name] = cl_helper.get_ga4_event_name(nested_contents_item[:href]) if nested_contents_item[:href]
--- a/app/views/govuk_publishing_components/components/_contents_list.html.erb
+++ b/app/views/govuk_publishing_components/components/_contents_list.html.erb
@@ -5,6 +5,7 @@
   underline_links ||= false
   format_numbers ||= false
   alternative_line_style ||= false
+  title ||= nil
   brand ||= false
   brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
   title_fallback = t("components.contents_list.contents", locale: I18n.locale, fallback: false, default: "en")
@@ -24,6 +25,7 @@
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-contents-list #{brand_helper.brand_class}")
   component_helper.add_class("gem-c-contents-list--alternative-line-style") if alternative_line_style
+  component_helper.add_class("gem-c-contents-list--custom-title") if title
   component_helper.add_data_attribute({ module: "ga4-link-tracker" }) unless disable_ga4
   component_helper.add_aria_attribute({ label: t("components.contents_list.contents") }) unless local_assigns[:aria][:label]
   component_helper.add_role("navigation")
@@ -35,7 +37,7 @@
         title_fallback == "en" ? {:lang => "en",} : {}
       )
     ) do %>
-      <%= t("components.contents_list.contents") %>
+      <%= title || t("components.contents_list.contents") %>
     <% end %>
 
     <ol class="gem-c-contents-list__list">
--- a/app/views/govuk_publishing_components/components/docs/contents_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/contents_list.yml
@@ -209,6 +209,22 @@ examples:
           text: Fourth page title link
         - href: "#fifth-thing"
           text: Fifth page title link
+  with_custom_title:
+    description: With this option, the 'Contents' title is replaced with the supplied alternate title. This should only be used when using this component as a navigation element on landing pages. Typically used with `alternative_line_style`.
+    data:
+      title: "An alternate title"
+      contents:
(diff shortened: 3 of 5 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -76,6 +76,33 @@
...
+      display: none;
+    }
+  }
+
+  .gem-c-contents-list__list-item--active {
+    padding: govuk-spacing(1) govuk-spacing(2);
+    margin: govuk-spacing(1) 0;
+    border-left-style: solid;
+    border-left-width: 4px;
+    border-color: govuk-colour("blue");
```
jon-kirwan, line 94: Minor comment but perhaps these could be shortened with `border-left: 4px solid #1d70b8;`

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -76,6 +76,27 @@
   }
 }
 
+.gem-c-contents-list--alternate-active {
+  .gem-c-contents-list__list-item {
+    padding: govuk-spacing(2) 0;
+    margin: 0;
+
+    & span::before {
```
andysellick, line 84: I know you didn't add this span, but can we take the opportunity to give it a proper class? Something like `.gem-c-contents-list__dash` or something.

app/views/govuk_publishing_components/components/docs/contents_list.yml
```diff
@@ -193,6 +193,22 @@ examples:
             text: 2. Numbers not parsed
           - href: "#third-thing"
             text: 3. Numbers are just text
+  with_alternate_active:
```
andysellick, line 196: It's hard naming things but maybe something more descriptive like `without_dashes` would be clearer?

app/views/govuk_publishing_components/components/docs/contents_list.yml
```diff
@@ -209,6 +209,22 @@ examples:
           text: Fourth page title link
         - href: "#fifth-thing"
           text: Fifth page title link
+  with_alternate_title:
```
andysellick, line 212: It's hard naming things but maybe `with_custom_title` would be clearer?

app/views/govuk_publishing_components/components/docs/contents_list.yml
```diff
@@ -209,6 +209,22 @@ examples:
...
+    description: With this option, the 'Contents' title is replaced with the supplied alternate title. This should only be used when using this component as a navigation element on landing pages. Typically used in conjunction with the `alternate_active` option.
+    data:
+      alternate_title: "An alternate title"
```
andysellick, line 215: Likewise I think this could just be `title`.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -97,6 +97,12 @@
   }
 }
 
+.gem-c-contents-list--alternate-title {
+  .gem-c-contents-list__title {
+    margin-bottom: govuk-spacing(3);
```
andysellick, line 102: What's this spacing for? Is it required for the non-dashed style? I don't think it belongs with this option - I wouldn't expect passing a custom title to also alter the appearance of the component.

app/views/govuk_publishing_components/components/_contents_list.html.erb
```diff
@@ -35,7 +37,11 @@
         title_fallback == "en" ? {:lang => "en",} : {}
       )
     ) do %>
-      <%= t("components.contents_list.contents") %>
+      <% if alternate_title.empty? %>
```
andysellick, line 40: This section can be written more simply as below, if you amend the original variable declaration...

```Ruby
alternate_title ||= nil

# ...

<%= alternate_title || t("components.contents_list.contents") %>
```

andysellick: And I think with this change you can also shorten line 28 to 

```Ruby
component_helper.add_class("gem-c-contents-list--alternate-title") if alternate_title
```

andysellick (commented):
Nice. A few suggestions for you. Also remember a CHANGELOG entry.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: Hmm, yeah this is a bit of a weird one because the extra space is required for the landing page design: it's to match the spacing of the `alternate-active` option which has larger line spacing than other contents-list variants:

image

So the only time we need the extra space is if both a custom title is set AND the alternate-active option is used. Is it OK then to have a combined selector to handle this, eg: 

`.gem-c-contents-list--alternate-active.gem-c-contents-list--alternate-title`

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: Since that applies to many other variants, should that be in this PR?

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: It's not just without dashes, it gets different vertical line-spacing between each option and the left hand blue vertical border for the active state. I also wanted to avoid using a name that included 'landing page'. I think this name is OK given the nature of what it does, but maybe removing the `with-` part is enough to simplify it?

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: As above, I could change this to just `custom-title`? I used the `with-` prefix to match other existing options, but there isn't much consistency in how they're named.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -76,6 +76,33 @@
...
+  }
+
+  .gem-c-contents-list__list-item--active {
+    padding: govuk-spacing(1) govuk-spacing(2);
+    margin: govuk-spacing(1) 0;
+    border-left-style: solid;
+    border-left-width: 4px;
+    border-color: govuk-colour("blue");
+    color: govuk-colour("blue");
+    font-weight: bold;
```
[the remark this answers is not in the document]
matthillco: Good spot thanks, now fixed.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: I've no idea why I did those so verbosely, I usually use shorthand! Now fixed, thanks.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
andysellick: Let's not make it overly complicated. If this extra spacing is needed for the landing page design, then let's include it with that variant.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
andysellick: I think that's fine and it should be a small change - just add the class in the markup and adjust the instances of `span` in the Sass to the new classname.

app/views/govuk_publishing_components/components/docs/contents_list.yml
andysellick: Sure.

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: I wonder now if this means we need a different option: expanded line spacing? This would allow the expanded spacing to be used by itself without also adopting the alternate active state... It does seem a little confusing that the alternate active state is also applying a different line height.

andysellick: `alternate_active` in my head implies that every other element is active. How about `with_no_dashes_and_wider_spacing`?

andysellick: > I wonder now if this means we need an extra option: expanded line spacing? This would allow the expanded spacing to be used by itself without also adopting the alternate active state. It does seem a little confusing that the alternate active state is also applying a different line height.

I don't think we want to start adding multiple variations when we only need one - we can always split them apart later if it turns out we need one without the other.

matthillco: Hmm, but it also gets a completely different visual treatment of the active state which isn't communicated by that name. Getting complex if we add that in though!

What about simply `alternative-line-style`? This isn't explicit, but it does cover all the changes if you think about it :) The associated description then explains all the changes.

andysellick: Sure, go with that.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: OK this is now done.

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: Rebasing went a bit awry, but after a bit of multi-branch cherry-picking, this is now changed as discussed.

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: Rebasing went a bit awry, but after a bit of multi-branch cherry-picking, this is now changed to 'with-custom-title`.

app/views/govuk_publishing_components/components/docs/contents_list.yml
matthillco: Done.

app/views/govuk_publishing_components/components/_contents_list.html.erb
matthillco: Nice, thanks for sharing that. I'm still learning these shortcuts. I've updated it as you described, along with the changed variable names from the other comments.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -50,7 +50,7 @@
   padding-left: $contents-spacing;
   padding-right: $contents-spacing;
 
-  & span::before {
+  & .gem-c-contents-list__list-item-dash::before {
```
andysellick, line 53: With the introduction of this class, I think this can now be further simplified. This line can change to:

```Sass
.gem-c-contents-list__list-item-dash::before {
```

...and the new section you added on line 84 can be moved out of the 2nd nested level and done as a single child of the parent, i.e.

```Sass
.gem-c-contents-list--alternative-line-style {
  .gem-c-contents-list__list-item {
    padding: govuk-spacing(2) 0;
    margin: 0;
  }

  .gem-c-contents-list__list-item-dash::before {
    display: none;
  }
```

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
andysellick: Looks like this is still part of the custom title variant, can it be part of the alternative style variant instead?

matthillco: Deleted from the title variant.

matthillco:
Hi @andysellick I've fixed all your requests, hopefully this can be merged now.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
andysellick: Thanks. I think you can also remove the `&` from the start of this line now as well.

app/views/govuk_publishing_components/components/_contents_list.html.erb
```diff
@@ -33,14 +37,14 @@
...
-      <%= t("components.contents_list.contents") %>
+      <%= title || t("components.contents_list.contents") %>
     <% end %>
 
     <ol class="gem-c-contents-list__list">
       <% index_link = 1 unless disable_ga4 %>
       <% contents.each do |contents_item| %>
         <li class="<%= cl_helper.list_item_classes(contents_item, false) %>" <%= "aria-current=true" if contents_item[:active] %>>
-          <span aria-hidden="true"></span>
+          <span class="gem-c-contents-list__list-item-dash" aria-hidden="true"></span>
```
andysellick, line 47: This class also needs adding on line 65, otherwise nested contents lists don't have dashes.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
```diff
@@ -28,7 +28,7 @@
 
 .gem-c-contents-list__link {
   .gem-c-contents-list__list-item--parent > & {
-    font-weight: bold;
+    font-weight: $govuk-font-weight-bold;
```
andysellick, line 31: I appreciate that you didn't write this whole structure, but can we take the opportunity to make it more readable? I think lines 29 to 35 can be replaced with this:

```
.gem-c-contents-list__link {
  @include govuk-template-link-focus-override;
}

.gem-c-contents-list__list-item--parent > .gem-c-contents-list__link {
  font-weight: $govuk-font-weight-bold;
}
```

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: Darn, I missed that during my rebase, fixed now. @andysellick

app/views/govuk_publishing_components/components/_contents_list.html.erb
matthillco: Ooops sorry I missed that, now fixed.

app/assets/stylesheets/govuk_publishing_components/components/_contents-list.scss
matthillco: Fixed as suggested.

andysellick (approved):
👏

PR #4316, opened 2024-10-18
Use flexbox for share links with square icons

AshGDS:
## What / Why

- Stop using `columns: true` for our square icons as it doesn't give us the visual balance we want where we're using it.
- Therefore, use flexbox with a `min-width` for this variant instead. 
- The `min-width` is needed because alignment/stacking visual inconsistencies occur when the elements have varying widths. Therefore, if we ensure there is a minimum width for the elements, the alignment and flow when resizing is more visually pleasing. If you think of a component like our image cards, they stack and align nicely as they are all the same width, so we need some of that guarantee here, hence the `min-width`. It's also worth noting that `columns: true` was utilising a `min-width` for its balancing of links. I don't think we should be putting any max width type limitations though like `columns` does. (Mainly mentioning this as in the previous PR I originally used `min-width`, which was what led to back and forth conversation and alternatives being suggested.)
- Trello: https://trello.com/c/mK8f3WMQ/93-adjust-share-links

### Before
image

image

### After

image

image

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/share_links.yml
+++ b/app/views/govuk_publishing_components/components/docs/share_links.yml
@@ -256,10 +256,45 @@ examples:
           icon: 'youtube'
         }
       ]
-  with_square_icons:
+  with_flexbox:
+    data:
+      flexbox: true
+      links: [
+        {
+          href: '/twitter-share-link',
+          text: 'Twitter',
+          icon: 'twitter'
+        },
+        {
+          href: '/instagram-share-link',
+          text: 'Instagram',
+          icon: 'instagram'
+        },
+                {
+          href: '/flickr-share-link',
+          text: 'Flickr',
+          icon: 'flickr'
+        },
+        {
+          href: '/facebook-share-link',
+          text: 'Facebook',
+          icon: 'facebook'
+        },
+        {
+          href: '/youtube-share-link',
+          text: 'YouTube',
+          icon: 'youtube'
+        },
+        {
+          href: '/other-share-link',
+          text: 'Anything else that might be included that could have quite a long name',
+          icon: 'other'
+        },
+      ]
+  with_square_icons_and_flexbox:
     data:
       square_icons: true
-      columns: true
+      flexbox: true
       links: [
         {
           href: '/twitter-share-link',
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/share_links.yml
```diff
@@ -256,10 +256,45 @@ examples:
...
+          text: 'YouTube',
+          icon: 'youtube'
+        },
+        {
+          href: '/other-share-link',
+          text: 'Anything else that might be included that could have quite a long name',
+          icon: 'other'
+        },
+      ]
   with_square_icons:
```
JamesCGDS, line 294: I know this isn't strictly part of this PR, but would it be possible to change this to something like `with_flexbox_and_square_icons`? Only because `with_square_icons` might suggest that square icons can be used independently of any other layout options but I think it currently only displays correctly when used in conjunction with `flexbox: true`.

JamesCGDS (commented):
Nice work, looks good! Got a couple of very minor comments. Also I tried a whole bunch of options trying to get this to work as per the designs (e.g. `justify-content`, playing with auto margins, playing with a `grid` layout) but I just couldn't get it to work nicely across all the different scenarios - I think this might be as close as we can get for the moment.

spec/components/share_links_spec.rb
```diff
@@ -125,6 +125,16 @@ def links
     assert_select ".gem-c-share-links .gem-c-share-links__link[href=\"/twitter\"] .govuk-visually-hidden", text: "Tweet to"
   end
 
+  it "adds the correct classes when flexbox is true" do
+    render_component(links:, flexbox: true)
+    assert_select ".gem-c-share-links--flexbox"
+  end
+
+  it "does not add extra classes when flexbox is false" do
+    render_component(links:, square_icons: false)
```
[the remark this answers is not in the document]
AshGDS: Whoops, it should have been `flexbox: false` in the first place, thanks

app/views/govuk_publishing_components/components/docs/share_links.yml
AshGDS: Thanks, done 👍

JamesCGDS (approved):
Looks good to me 👍

PR #4317, opened 2024-10-18
GA4: Fix ecommerce tracking of searches without query

csutter:
## What
The GA4 ecommerce tracking for search results doesn't currently work if the user is searching for an empty query, as the check for whether the `ga4-search-query` data attribute is present uses loose Javascript boolean logic (i.e. returning false for an empty string).

This amends it to strictly check for `null` instead (data attribute not present at all), and adds some tests for the search ecommerce data being properly set with the query present and absent.

## Why
So we can actually track empty searches like all other searches.


AshGDS (approved):

AshGDS:
Needs a CHANGELOG entry probably 👍

PR #4318, opened 2024-10-18
Add new chart component options

andysellick:
## What

Adds some options to the chart component:

- minimal, for when the graph is in a small space and should display minimal information and link to more information
- height, to set the height in situations like the above

I've started going down a bit of a rabbit hole with this one, so I've raised https://github.com/alphagov/govuk_publishing_components/issues/4322 as an issue to record my concerns rather than try to address them here and spiral the scope out of control.

## Why
Needed for a page where the graphs are going to be small and link to another page.

## Visual Changes
Minimal version:

Screenshot 2024-10-21 at 09 19 18

Minimal version with a smaller height:

Screenshot 2024-10-21 at 09 19 26

Trello card: https://trello.com/c/DFT05NWN/97-implement-small-graphs-on-main-landing-page

### DIFF
--- a/app/views/govuk_publishing_components/components/_chart.html.erb
+++ b/app/views/govuk_publishing_components/components/_chart.html.erb
@@ -42,8 +42,9 @@
   enableInteractivity = false if minimal
   textPosition = nil
   textPosition = 'none' if minimal
-  valid_minimal = true
-  valid_minimal = false if minimal && !minimal_link # minimal version must include a link
+  if minimal && !minimal_link
+    raise ArgumentError, "Minimal version must include a link"
+  end
 
   chart_library_options = {
     chartArea: { width: '80%', height: '60%' },
@@ -80,7 +81,7 @@
     end
   end
 %>
-<% if rows.length > 0 && keys.length > 0 && valid_minimal %>
+<% if rows.length > 0 && keys.length > 0 %>
   <%= javascript_include_tag "https://www.gstatic.com/charts/loader.js" if @external_script[:loaded] == 1 %>
   <%= tag.div(**component_helper.all_attributes) do %>
     <% if chart_heading && !minimal %>
--- a/app/views/govuk_publishing_components/components/_chart.html.erb
+++ b/app/views/govuk_publishing_components/components/_chart.html.erb
@@ -20,6 +20,8 @@
 
   chart_id = "chart-id-#{SecureRandom.hex(4)}"
   table_id = "table-id-#{SecureRandom.hex(4)}"
+  @external_script ||= OpenStruct.new(loaded: 0)
+  @external_script[:loaded] += 1
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -75,77 +89,85 @@
     <% end %>
 
     <div class="gem-c-chart__chart" id="<%= chart_id %>">
```
AshGDS, line 94: If the purpose of the minimal version is just for presentation, and not for conveying information, should it be hidden from the accessibility tree? E.g. give it a `tabindex` so that it can't be focused on, and use `aria-hidden="true"`?

If not, I think the focus style for the minimal version should be improved - for example the yellow is not visible around the whole chart container. Otherwise, I think it will be flagged as an accessibility issue. 

Also, how will users know that this is a link? Will it be next to some content that implies so?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -286,3 +286,43 @@ examples:
...
+            - 79
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
```
AshGDS, line 328: Nitpick: no empty line at the end of the file

AshGDS (commented):
Looks good, just had some questions first 👍

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -30,26 +33,37 @@
   font_19 = { color: '#000', fontName: 'GDS Transport', fontSize: '19', italic: false }
   legend = 'none'
   legend = { position: 'top', textStyle: font_16 } unless hide_legend
+  pointSize = 10
+  pointSize = 0 if minimal
+  enableInteractivity = true
+  enableInteractivity = false if minimal
+  textPosition = nil
+  textPosition = 'none' if minimal
+  valid_minimal = true
```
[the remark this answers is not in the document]
andysellick: Good shout, thanks, have updated.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Fixed.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Good shout on the first point - have added `aria-hidden` as suggested so now the only thing a screenreader should encounter is the link. Have also improved the focus style for the link to make it look better.

How will users know this is a link? That's an excellent question and I'm not sure of the answer. In the designs we have they're directly underneath some headings, which themselves are also links, so this being a link as well is kind of a backup (although that itself introduces a duplicate link, which isn't great).

I'm left wondering if minimal should be a link at all, but the link is needed to allow users to go to the full graph and table of data - it might as well just be an image if it isn't a link, but we decided to generate a chart otherwise we have to keep updating images every time the chart they link to changes. A bit of a rabbit hole in general.

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -92,14 +92,15 @@
       } %>
     <% end %>
 
-    <div class="gem-c-chart__chart" id="<%= chart_id %>">
+    <% aria_attributes = { hidden: true } if minimal %>
```
AshGDS, line 95: I think if we have `aria-hidden` i don't think it can be keyboard focusable as well, as the component guide is flagging this as an issue. Happy for you to pick between either having focus or having `aria-hidden`.

AshGDS (approved):
Looks good, minor comment but happy for you to sort that on your own, so will approve 👍

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Argh, good spot, thanks. Think I've fixed this now - since the chart is just a link, I've included some logic to simply not render the visually hidden text (it's the thing that contained a link).

andysellick:
@AshGDS are you happy to approve the visual changes?

PR #4321, opened 2024-10-21
Refactor organisation styles and upgrade to govuk-frontend v5.7.1

AshGDS:
**Breaking change, as it stops certain colours working on components that are using the `brand` option.`**

## What / Why

- Removes generic `.brand__color` class, which was used to give organisations custom link colours from `govuk-frontend`. However we keep a namespaced `.brand__color` style for `prime-ministers-office-10-downing-street`, so that links will remain black on the no 10 page only. 
- Also removes the `brand` option from the `share_links` component, and replaces it with `black_icons` and `black_links`. These are needed because the default brand styles are now going to be a black icon with our standard blue links, but the No 10 page needs black links as well, hence needing both options. Only `collections` used the `brand` variant of this component, so this shouldn't break any other pages. Using these new options in `collections` is simple: https://github.com/alphagov/collections/pull/3802
- The two above points will introduce minor tech debt in the sense that `collections` will be passing `brand:` to the share links component when that is no longer a variant, and organisation components will have a `brand__colour` class on them which isn't doing anything now that they all use the default blue colour. These two things may be a backendy thing to resolve, as the `brand` options in `collections` seem quite tightly coupled to the backend/models of the org pages.
- This also updates the DSIT class override we have, instead of removing it. We need this because the class was renamed in `govuk-frontend` from `department-for-science-innovation-and-technology` to `department-for-science-innovation-technology` ( - the `and` was dropped.) Our `whitehall` model still uses the old class name, and therefore renders the page using the old class name, so we need to keep this temporary class in for now. Once we've updated `whitehall` model to use the new class name, it will grab the styles directly from `govuk-frontend` allowing us to remove the class in our gem.

## Testing

- To test this, you'll need to run `static` with this gem. Then run `collections` with this branch: https://github.com/alphagov/collections/pull/3802 while pointing at your local `static` and this version of the gem.

### Before
image

image

### After

image

image

### DIFF
--- a/app/views/govuk_publishing_components/components/_share_links.html.erb
+++ b/app/views/govuk_publishing_components/components/_share_links.html.erb
@@ -11,16 +11,16 @@
   flexbox ||= false
   square_icons ||= false
 
-  brand ||= false
-  brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
+  black_icons ||= false
+  black_links ||= false
 
   classes = %w(gem-c-share-links govuk-!-display-none-print)
   classes << "gem-c-share-links--stacked" if stacked
   classes << "gem-c-share-links--columns" if columns
   classes << "gem-c-share-links--flexbox" if flexbox
   classes << "gem-c-share-links--square-icons" if square_icons
-
-  classes << brand_helper.brand_class
+  classes << "gem-c-share-links--black-icons" if black_icons
+  classes << "gem-c-share-links--black-links" if black_links
 
   data_attributes ||= {}
   ((data_attributes[:module] ||= "") << " " << "ga4-link-tracker").strip! if track_as_sharing || track_as_follow
@@ -76,7 +76,7 @@
             target: "_blank",
             rel: "noopener noreferrer external",
             data: data_attributes,
-            class: "govuk-link govuk-link--no-underline gem-c-share-links__link #{brand_helper.color_class}" do %>
+            class: "govuk-link govuk-link--no-underline gem-c-share-links__link" do %>
             <span class="gem-c-share-links__link-icon">
               <% if link[:icon] == 'facebook' %>
                 <svg viewBox="0 0 24 24" xmlns="http://www.w3.org/2000/svg" aria-hidden="true" focusable="false" width="0" height="0" class="gem-c-share-links__svg">
--- a/app/views/govuk_publishing_components/components/docs/contents_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/contents_list.yml
@@ -190,10 +190,10 @@ examples:
               text: "تقديم الطلب"
     context:
       right_to_left: true
-  with_branding:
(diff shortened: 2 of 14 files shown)


app/views/govuk_publishing_components/components/_share_links.html.erb
```diff
@@ -10,15 +10,15 @@
   columns ||= false
   square_icons ||= false
 
-  brand ||= false
-  brand_helper = GovukPublishingComponents::AppHelpers::BrandHelper.new(brand)
+  black_icons ||= false
+  black_links ||= false
```
andysellick, line 14: Do we need `black_icons` and `black_links` to be separate options? The use case we're working towards would have both of these options enabled together, is that right?

app/views/govuk_publishing_components/components/docs/contents_list.yml
```diff
@@ -181,9 +181,9 @@ examples:
     context:
       right_to_left: true
   with_branding:
```
andysellick, line 183: Is it worth renaming these options to `with_no_10_branding`? Since that's the only option for branding now available?

docs/component_branding.md
```diff
@@ -22,12 +26,43 @@ To add colours to a component, modify the component to follow the example below.
   </div>
 
   <a href="#" class="<%= brand_helper.color_class %>">
-    Example element that requires coloured text
+    Obscure use case where a link requires coloured text
```
andysellick, line 29: Maybe soften the language here? `Rare use case...`?

docs/component_branding.md
```diff
@@ -22,12 +26,43 @@ To add colours to a component, modify the component to follow the example below.
...
   </a>
 </div>
 ```
 
-Note that the helper must be called for each element that needs a border or link colour applying. This allows for flexibility if one is required but not the other.
+The above will style borders with colours from the design system automatically, as we have a SASS mixin for this in `_brand_colours.scss` that will map the class to its colour in the design system.
+
+If your use case fits one of these scenarios:
```
andysellick, line 36: I think for the list below this line to format correctly there needs to be an empty line between the two?

andysellick (commented):
Looks pretty comprehensive, just a few comments.

app/views/govuk_publishing_components/components/_share_links.html.erb
AshGDS: @andysellick `black_icons` will be used on all organisations, they will have a black icon and blue links. Whereas the no 10 page will have black icons and black links, which is why both are separate

docs/component_branding.md
andysellick: If it looks fine then that's fine 👍 Wasn't able to check myself.

app/views/govuk_publishing_components/components/_share_links.html.erb
andysellick: That makes sense, thanks 👍

maxgds:
In terms of the two things outlined in your PR description:
Could you raise a ticket on the tech debt board for Web to get the collections update done so we stop passing around redundant brand names?
Similarly, could you ask the Whitehall team about their plans to update the DSIT name and ask them to inform us when it's done so we can remove the temporary fix.

AshGDS:
Thanks both 👍 @maxgds will do. Last time we did the Whitehall change ourselves, should we do it ourselves again just to reduce dependency on other teams? https://github.com/alphagov/whitehall/pull/9379

PR #4326, opened 2024-10-21
Make it easier to find application level components in component guides

hannalaakso:
## What

Reorganise the list of components in the application component guides so that:
- App level components are presented first
- Instead of listing All gem component in an app component guide, list gem components NOT used by the app, to potentially help users identify gem components that could be utilised by the app

## Why
Application level components are always the fewest in number in each application. They are not easily visible when looking through the list of components, even though we are most likely interested in checking the documentation for the application level components. Listing “All gem components” in a component app guide can make it difficult to identify gem components available to this application that have not already been used, so instead list "Gem components not used by this app".

### Before in collections component guide

Screenshot 2024-10-29 at 14 53 30

### After in collections component guide

Screenshot 2024-10-29 at 14 52 29

### Before in gem component guide 
Screenshot 2024-10-29 at 14 54 30

### After in gem component guide (No change)
Screenshot 2024-10-29 at 14 54 55

## To view examples

- [Gem component guide](https://components-gem-pr-4326.herokuapp.com/component-guide)
- To view an app component guide such as the collections component guide, pull down this gem branch and link to it from collections Gemfile (`gem "govuk_publishing_components", path: "../govuk_publishing_components"`). Navigate to`/component-guide/` in the frontend app eg. collections.

https://trello.com/c/twxnT667/2950-update-component-guide-in-applications

### DIFF
--- a/app/controllers/govuk_publishing_components/component_guide_controller.rb
+++ b/app/controllers/govuk_publishing_components/component_guide_controller.rb
@@ -9,7 +9,8 @@ module GovukPublishingComponents
       @component_gem_path = Gem.loaded_specs["govuk_publishing_components"].full_gem_path
       @component_docs = component_docs.all
       @gem_component_docs = gem_component_docs.all
-      @components_in_use_docs = components_in_use_docs.used_in_this_app
+      @used_components = used_components_names.get_component_docs
+      @unused_components = unused_components_names.get_component_docs
       @components_in_use_sass = components_in_use_sass
       @components_in_use_js = components_in_use_js
     end
@@ -47,7 +48,7 @@ module GovukPublishingComponents
       additional_files = "@import 'govuk_publishing_components/govuk_frontend_support';\n"
       additional_files << "@import 'govuk_publishing_components/component_support';\n"
 
-      components = find_all_partials_in(components_in_use)
+      components = find_all_partials_in(get_used_component_names)
 
       components.map { |component|
         "@import 'govuk_publishing_components/components/#{component.gsub('_', '-')}';" if component_has_sass_file(component.gsub("_", "-"))
@@ -57,7 +58,7 @@ module GovukPublishingComponents
     def components_in_use_js
       additional_files = "//= require govuk_publishing_components/lib\n"
 
-      components = find_all_partials_in(components_in_use)
+      components = find_all_partials_in(get_used_component_names)
 
       components.map { |component|
         "//= require govuk_publishing_components/components/#{component.gsub('_', '-')}" if component_has_js_file(component.gsub("_", "-"))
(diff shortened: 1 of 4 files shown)


app/controllers/govuk_publishing_components/component_guide_controller.rb
```diff
@@ -78,6 +79,10 @@ def components_in_use_docs
       @components_in_use_docs ||= ComponentDocs.new(gem_components: true, limit_to: components_in_use)
     end
 
+    def components_not_in_use_docs
+      @components_not_in_use_docs ||= ComponentDocs.new(gem_components: true, limit_to: components_in_use)
+    end
+
```
beccapearce, line 85:
```suggestion
    def components_not_in_use_docs
      @components_not_in_use_docs ||= ComponentDocs.new(gem_components: true, limit_to: components_not_in_use)
    end
```

I think this should be `components_not_in_use` for this method

app/models/govuk_publishing_components/component_docs.rb
```diff
@@ -19,6 +19,10 @@ def used_in_this_app
       fetch_component_docs.map { |component| build(component) if component_in_use(component[:id]) }.compact.sort_by(&:name)
     end
 
+    def not_used_in_this_app
+      fetch_component_docs.map { |component| build(component) if component_not_in_use(component[:id]) }.compact.sort_by(&:name)
+    end
+
```
beccapearce, line 25: I wonder if we could reduce a bit of duplication here?

maybe something like this:
```
    def used_in_this_app
      fetch_component_docs.map { |component| build(component) if component_in_use(component[:id]) }.compact.sort_by(&:name)
    end

    def not_used_in_this_app
      fetch_component_docs.map { |component| build(component) if component_not_in_use(component[:id]) }.compact.sort_by(&:name)
    end
    
    private
        def filter_by_usage(in_use: true)
          fetch_component_docs.map { |component|
        build(component) if component_in_use?(component[:id], in_use:)
      }.compact.sort_by(&:name)
      end
```

What do you think? is it over complicated?

app/models/govuk_publishing_components/component_docs.rb
```diff
@@ -34,6 +38,10 @@ def component_in_use(component)
       true if @limit_to.include?(component)
     end
 
+    def component_not_in_use(component)
+      true if @limit_to.exclude?(component)
+    end
+
```
beccapearce, line 44: Again similar to above maybe we could combine `component_in_use` and `component__not_in_use`? like this:

```
    def component_in_use?(component, in_use:)
      in_use ? @limit_to.include?(component) : @limit_to.exclude?(component)
    end
```

Again no problem if it's too confusing!

beccapearce (commented):
I think there was one issue with `component_in_use` being used instead of `component_not_in_use` in the controller but otherwise looks good! I made a couple of suggestions, but not sure if it's an improvement or just confusing, let me know what you think!

We could also maybe add some tests in to get some more confidence in the code. I pushed some up to a [separate branch](https://github.com/alphagov/govuk_publishing_components/blob/55f628c7a7dd1c27ba779767eedfeb033c167065/spec/component_guide/component_docs_spec.rb) and we can add them here if you're happy with them!

PR #4333, opened 2024-10-24
Contents list add margin_bottom option

andysellick:
## What
Adds the shared helper for margin bottom to the contents list component.

## Why
We want to use it somewhere and it needs a different bottom margin.

## Visual Changes
None.

Trello card: https://trello.com/c/t0DlqDkW

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/contents_list.yml
+++ b/app/views/govuk_publishing_components/components/docs/contents_list.yml
@@ -1,12 +1,11 @@
 name: Contents list
 description: Provides a list of links with options for dashes or numbering.
 body: |
-  Commonly used to lists a page’s contents with links pointing to headings within the document, but can also be used for a list of links to other pages.
+  Commonly used to list the contents of a page with links pointing to headings within the document, but can also be used for a list of links to other pages.
 
   Pass a list of contents each with an `href` and `text`. The `href` can point at the ID of a heading within the page.
 
-  Supports nesting contents one level deep, currently only used by specialist documents. When nesting the top level list items
-  display in bold.
+  Supports nesting contents one level deep, currently only used by specialist documents. When nesting the top level list items display in bold.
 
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/contents_list.yml
```diff
@@ -36,6 +35,17 @@ examples:
           text: Second thing
         - href: "#third-thing"
           text: Third thing
+  with_bottom_margin:
+    description: The component accepts a number for margin bottom from `0` to `9` (`0px` to `60px`) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). The default big number margin bottom is `20px` (`govuk-spacing(4)`).
```
maxgds, line 39: Big number doesn't seem relevant here - copy paste error?

spec/components/contents_list_spec.rb
```diff
@@ -77,6 +77,19 @@ def contents_list_with_markup
     assert_select ".gem-c-contents-list__link.govuk-link--no-underline[href='#two']", text: "2. Two"
   end
 
+  it "applies default margin to the component" do
+    render_component(contents: contents_list)
+    assert_select '.gem-c-contents-list.govuk-\!-margin-bottom-4'
+  end
+
+  it "applies margin to the component" do
```
maxgds, line 85: maybe make the description more explicit - "applies requested margin to the component" as it reads more like the default option as is.

app/views/govuk_publishing_components/components/docs/contents_list.yml
andysellick: 🤦 thanks, fixed.

spec/components/contents_list_spec.rb
andysellick: No problem, updated.

maxgds (approved):

PR #4337, opened 2024-10-24
Fix incorrect underline styles on share links

AshGDS:
## What / Why

- There's a hover underline visual issue when using the `flexbox` variant. 
- Looks like it was due to main label being a span (meaning it had `display: inline`).
- Switching to `inline-block` seems to have fixed it. This removed the default underline styles, so I reintroduced them via the CSS, which looks better anyway as there's now spacing between the line and the word.
- Not sure what's going on with the Percy visual change.

### Before
image

image

### After
image

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


andysellick (approved):
👍

PR #4342, opened 2024-10-25
Adjust chart component options

andysellick:
## What
Change the chart component. Specifically:

- remove the requirement for a link from the minimal mode
- adds an option to pad the text and elements around the graph

## Why
For a minimal chart we thought that it should link to a page with more data, but this has proved complicated in an actual page, and now I think it's better handled in an application rather than dictated by the component.

Padding is needed for when the chart is inside an element with a border, as these elements touch the outer edge.

## Visual Changes
No change to the minimal version, except that it no longer has a link.

Padded version shown below with the normal version for comparison (it's a bit subtle, look at the left edge).

Screenshot 2024-10-25 at 15 12 51

Trello card: https://trello.com/c/VKrLuGvg

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/chart.yml
+++ b/app/views/govuk_publishing_components/components/docs/chart.yml
@@ -243,10 +243,11 @@ examples:
             - 80
   minimal_version:
     description: |
-      The minimal version presents a simplified version of the chart and should only be used where the user is then directed to a more detailed view of the data. Specifically, minimal mode:
+      The minimal version presents a simplified version of the chart. This should only be used where there is not enough space to display a full chart, and the user is then given an option to see more information about the chart, for example by including a link next to the chart component. This must be part of the page, as this is not provided by the component itself.
 
-      - turns the chart into a link that should point to a page with a full version of the chart with all data
-      - hides the chart heading, but uses the text as the text for the link
+      Specifically, minimal mode:
+
+      - hides the chart heading
       - removes the legend and X and Y axis values
       - removes the data table and link to the data (if supplied) beneath the chart
       - removes the part of the visually hidden accessibility message that links to the table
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -243,9 +243,10 @@ examples:
             - 80
   minimal_version:
     description: |
-      The minimal version presents a simplified version of the chart and should only be used where the user is then directed to a more detailed view of the data. Specifically, minimal mode:
+      The minimal version presents a simplified version of the chart. This should only be used where there is not enough space to display a full chart, and the user is then given an option to see more information about the chart, for example by including a link next to the chart component. This must be part of the page, as this is not provided by the component itself.
+
+      Specifically, minimal mode:
 
-      - turns the chart into a link that should point to a page with a full version of the chart with all data
       - hides the chart heading, but uses the text as the text for the link
```
AshGDS, line 286: Does the reference to a link need to be removed on this line as well?

AshGDS (approved):
Looks good, does it need a CHANGELOG?

I've also added a small comment about one of the lines in the variant description.

Also, if you rebase with main the Percy diff should be more accurate.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Good spot, have removed, and added a changelog entry. Thanks! 👍

andysellick: @AshGDS are you happy to approve the visual diff?

AshGDS (approved):
Thanks 👍

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Yep, done 👍

PR #4344, opened 2024-10-28
Chart component accessibility improvements

AshGDS:
## What / Why

- In the chart component, there is a "Skip to table" link that was visually hidden, but still focusable. This caused an accessibility issue as the link was focusable but had no focus styles. Therefore, I have used the [existing skip link component](https://components.publishing.service.gov.uk/component-guide/skip_link) in order to ensure the link has styles when focused on.
    - I had to add a couple custom styles to the parent element of the component, so that the skip link didn't appear when JavaScript was disabled, as it isn't needed when JS is disabled.
    - I also moved the link out of the `<div>` it was originally in, as that `<div>` uses `govuk-visually-hidden`, which clashes with how the skip link component works.
- Additionally, the data tables also did not have a caption, which was a separate accessibility issue. Therefore I have added a visually hidden caption to the data tables.
- Both of these accessibility improvements rely on a `chart_heading` existing. Therefore I have added some validation to ensure `chart_heading` text is passed through, unless the chart is in `minimal` mode.
- Fixes these two trello cards: [Chart focus not visible](https://trello.com/c/Nz6SSyIc) and [Chart tables without caption or heading](https://trello.com/c/ybFOZLr7)
- I've had the accessibility fixes checked by our accessibility specialist.

### This skip link now appears when focused

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
@@ -47,3 +47,18 @@
     padding: 0 govuk-spacing(4);
   }
 }
+
+.gem-c-chart .gem-c-chart__a11y-note-link {
+  display: none;
+  margin-bottom: govuk-spacing(2);
+}
+
+.js-enabled {
+  .gem-c-chart .gem-c-chart__a11y-note-link {
+    display: inline;
+  }
+
+  .gem-c-chart .gem-c-chart__a11y-note-link:focus-within {
+    display: inline-block;
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Chart component accessibility improvements ([PR #4344](https://github.com/alphagov/govuk_publishing_components/pull/4344))
+
 ## 44.9.1
 
 * Layout super nav header: Correct name for search input ([PR #4347](https://github.com/alphagov/govuk_publishing_components/pull/4347))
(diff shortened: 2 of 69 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
```diff
@@ -47,3 +47,14 @@
...
 }
+
+.gem-c-chart .gem-c-chart__a11y-note-link {
+  display: none;
+}
+
+.js-enabled {
+  .gem-c-chart .gem-c-chart__a11y-note-link {
+    display: inline-block;
+    margin-bottom: govuk-spacing(2);
```
matthillco, line 58: A really tiny point, but this margin might be better on the rule above, rather than introducing it as part of a state change?

matthillco (commented):
Looks good, nice work, just one very tiny comment:

app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
AshGDS: Thanks @matthillco - good point, I have now made that change 👍

AshGDS:
Thanks @matthillco - should be ready for another review 👍

matthillco (approved):

AshGDS:
Thanks @matthillco - are you able to approve the Percy diff as well? It looks a bit odd, but I think this is because my addition to the accessibility acceptance criteria list shifted all the examples down a few pixels.

AshGDS:
@matthillco actually, looking at it, the margin may have changed between the chart and the heading. Will investigate.

AshGDS:
Sorry @matthillco - could you check my latest commit and then approve again if what I've done is OK? When the skip link had `inline-block` it affected the DOM spacing, so I've given it `display: inline` until it's focused on.

matthillco (approved):
Latest changes look good to go.

PR #4331, opened 2024-10-23
Add data attributes to summary cards and lists

pezholio:
## What

This allows data attributes to be added to summary card and summary list rows

## Why

This will allow developers to add Javascript interactivity to individual rows via module initialisation.

PR #4346, opened 2024-10-29
Release version 44.9.0

pezholio:
* Layout super nav header: Correct URL for search form ([PR #4341](https://github.com/alphagov/govuk_publishing_components/pull/4341))
* Add support for data attributes in Summary List and Summary Card rows ([PR #4331](https://github.com/alphagov/govuk_publishing_components/pull/4331))

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
-## Unreleased
+## 44.9.0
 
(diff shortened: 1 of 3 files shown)


AshGDS (changes requested):
Nitpick: I think this would actually be `44.9.0` as your data attribute PR adds a new feature to a component, rather than a patch? (No need to create a new branch to change this - happy with the version numbers just being updated in the file changes)

Also, I'd recommend just updating the `govuk_publishing_components` version manually in Gemfile.lock, instead of updating all the dependencies, as we usually let dependabot handle the dependency upgrades.

Thanks!

AshGDS (approved):
LGTM 👍

PR #4349, opened 2024-10-30
Remove govspeak advisory component

catalinailie:
## Why
Trello card: https://trello.com/c/Z26fvy04/2957-work-out-how-to-retire-advisory-from-govspeak-gem


beccapearce (approved):

beccapearce:
The [whitehall rake task](https://github.com/alphagov/whitehall/blob/main/lib/tasks/remove_advisory.rake) was sucessfully run and there are no instances of advisory that are currently published. So this PR is all good to merge.

PR #4354, opened 2024-10-31
Add new custom analytics tracker for search

csutter:
## What
This adds a new `Ga4SearchTracker` analytics tracker and replaces the use of the generic `Ga4FormTracker` in the layout super navigation header with it.

In the near future, we will also use it to replace:
- the usage of the generic `Ga4FormTracker` on the homepage in Frontend
- the custom search event code on the all content finder ("site search") in Finder Frontend

For now, this fires exactly the same event with the same data as the existing use of the form tracker/custom search event code.

## Why
This new module allows us to share consistent logic in `search` event data between all the different places we offer a search box across GOV.UK. This will be particularly important once we add the ability to track usage of the upcoming autocomplete functionality as well.


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
```diff
@@ -0,0 +1,72 @@
...
+      this.initialKeywords = this.$searchInput.value
+
+      if (window.GOVUK.getConsentCookie() && window.GOVUK.getConsentCookie().usage) {
+        this.start()
+      } else {
+        window.addEventListener('cookie-consent', () => this.start())
+      }
+    }
+
+    start () {
```
kevindew, line 34: Might be good to call this startTracking ?

I found myself wondering how this was different from init, which itself is confusing how it's different from the constructor 😆.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
```diff
@@ -0,0 +1,72 @@
...
+      // present, but has not changed from its initial value
+      return this.searchTerm() !== '' && this.searchTerm() === this.initialKeywords
+    }
+
+    searchTerm () {
+      // `standardiseSearchTerm` returns undefined for empty strings, whereas we actively want an
+      // empty string as part of search events (undefined would not overwrite the current value in
+      // the analytics state)
+      const value = this.$searchInput.value
+      return window.GOVUK.analyticsGa4.core.trackFunctions.standardiseSearchTerm(value) || ''
```
kevindew, line 67: This is a bit of a beast might be good to do:

```suggestion
      const { standardiseSearchTerm } = window.GOVUK.analyticsGa4.core.trackFunctions
      return standardiseSearchTerm(value) || ''
```

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.spec.js
```diff
@@ -0,0 +1,136 @@
...
+
+      form.removeChild(input)
+      ga4SearchTracker = new GOVUK.Modules.Ga4SearchTracker(form)
+      ga4SearchTracker.init()
+    })
+
+    it('does not track search events', () => {
+      GOVUK.triggerEvent(form, 'submit')
+
+      expect(GOVUK.analyticsGa4.core.applySchemaAndSendData).not.toHaveBeenCalled()
```
kevindew, line 62: Is this able to be more succinct at all? as it's quite a mouthful.

Not sure if we can do: 

```
const applySchemaAndSendDataSpy = spyOn(GOVUK.analyticsGa4.core, 'applySchemaAndSendData')
expect(applySchemaAndSendDataSpy).not.toHaveBeenCalled()
```

kevindew (commented):
This looks great. I've not quite given it a thorough review as I've had to jump on a call but don't expect I'll have other comments.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
csutter: Ah yes - actually `startModule` is the [canonical name](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/javascript-modules.md#modules-and-cookie-consent) for modules that need cookie consent, so will update 👍

kevindew (approved):
Looks great 👍

PR #4359, opened 2024-10-31
Add component wrapper helper to error message component

AshGDS:
## What
- Adds functionality to the component wrapper helper to accept/validate the `dir` attribute.
- Adds the component wrapper helper to the `Form character count` component.

## Why

## Visual changes

None.


lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
```diff
@@ -76,6 +78,11 @@ def set_tabindex(tabindex_attribute)
         @options[:tabindex] = tabindex_attribute
       end
 
+      def set_dir(dir_attribute)
```
andysellick, line 81: I'm starting to wonder if there's a Ruby way of writing a generic method here to save repeating essentially the same method. Not for this PR, but something to ponder, maybe something like:

```Ruby
def set_thing(method_name, attribute)
  # call function "check_#{method_name}"_is_valid(attribute)
  @options[method_name.to_key] = attribute
end
```

spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
```diff
@@ -215,6 +215,21 @@
...
     end
 
+    it "does not accept an invalid dir value" do
+      error = "dir attribute (false) is not recognised"
+      expect {
+        GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(dir: "false")
+      }.to raise_error(ArgumentError, error)
+    end
+
+    it "accepts valid dir attribute value" do
```
andysellick, line 225: Trying to remember how this works, so I could be wrong, but does this also need a test for `I've set an initial value for dir and I'm trying to override it`?

andysellick (approved):
Looks good, just a couple of non-blocking thoughts.

lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
AshGDS: Good point, I noticed this as well so will have a think 👍

spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
AshGDS: Good spot, will add a test - thanks

PR #4362, opened 2024-11-01
Extend chart component to allow headings to be removed from the DOM

AshGDS:
## What / Why
- In [a previous PR](https://github.com/alphagov/govuk_publishing_components/pull/4344), I added validation so that `chart_heading` is a required field.
- This was done for accessibility purposes, otherwise skip links and data tables didn't have proper labels. They used `chart_heading` to tell the user what chart the labels were related to.
- This worked in the gem, because all the examples in the gem had a `chart_heading`.
- However, in practice we veered away from the component examples, and were rendering charts without any `chart_heading` value. (The tasks page in our project.)
- Therefore, that tasks page is crashing on the latest version of the gem. Passing a `chart_heading` to this page would change the design. Therefore to fix this I have added an `hide_heading` variable that can be passed to the chart as well, and that allows the validation to work without rendering the heading in the DOM. Thanks to Max H for the idea.

## Visual Changes

A new component example.

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/chart.yml
+++ b/app/views/govuk_publishing_components/components/docs/chart.yml
@@ -362,4 +362,40 @@ examples:
             - 118
             - 85
             - 80
+  with_hidden_heading:
+    description: |
+      The heading element is optional, allowing a heading to be set outside the component. Hiding the heading doesn't remove the need to populate `chart_heading` with text, as `chart_heading` is needed for accessibility.
+    data:
+      chart_heading: Page views
+      hide_heading: true
+      h_axis_title: Day
+      v_axis_title: Views
+      chart_overview: This is a graph of views per day
+      keys:
+        - 1st
+        - 2nd
+        - 3rd
+        - 4th
+        - 5th
+        - 6th
+        - 7th
+      rows:
+        - label: January 2015
+          values:
+            - 5
+            - 119
+            - 74
+            - 117
+            - 33
+            - 89
+            - 79
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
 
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -362,4 +362,40 @@ examples:
             - 118
             - 85
             - 80
+  with_hidden_heading:
+    description: |
+      The heading element can be removed entirely from the DOM. This is needed if the heading will come from somewhere outside the component. Hiding the heading doesn't remove the need to populate `chart_heading` with text, as `chart_heading` is needed for accessibility.
```
maxgds, line 367: Could we reword this - "removed entirely from the DOM" sounds like a js interaction after load - it's actually never in the DOM to be removed. Maybe just:
The heading element is optional, allowing a heading to be set outside the component. This doesn't remove the need to populate...

spec/components/chart_spec.rb
```diff
@@ -85,6 +85,14 @@ def component_name
     assert_select "h4.gem-c-heading", text: "hello"
   end
 
+  it "can exclude the heading from the DOM" do
```
maxgds, line 88: likewise - it "can be rendered without a visible heading"

maxgds (changes requested):
Seems nice and clean. A couple of nitpicky change requests, though.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Good point, I've changed the wording.

spec/components/chart_spec.rb
AshGDS: Good point, I've changed the wording here too 👍

AshGDS:
Thanks @maxgds - should be ready for another review 👍

maxgds (approved):
Looks good!

PR #4364, opened 2024-11-04
Add component wrapper helper to label component

AshGDS:
## What
- Adds the component wrapper helper to the `form label` component.

## Why

## Visual changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_label.html.erb
+++ b/app/views/govuk_publishing_components/components/_label.html.erb
@@ -4,25 +4,26 @@
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
   hint_text ||= ""
-  id ||= nil
   is_radio_label ||= false
   bold ||= false
   heading_size = false unless shared_helper.valid_heading_size?(heading_size)
   is_page_heading ||= false
   right_to_left ||= false
 
-  css_classes = %w[gem-c-label govuk-label]
-  css_classes << "govuk-label--s" if bold
-  css_classes << "govuk-radios__label" if is_radio_label
-  css_classes << "govuk-label--#{heading_size}" if heading_size
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-label govuk-label")
+  component_helper.add_class("govuk-label--s") if bold
+  component_helper.add_class("govuk-radios__label") if is_radio_label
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_label.html.erb
```diff
@@ -11,18 +11,20 @@
...
 %>
 
 <% if is_page_heading %>
   <%= tag.h1 text, class: "govuk-label-wrapper" do %>
-    <%= tag.label text, id: id, for: html_for, class: css_classes, dir: right_to_left ? "rtl" : nil %>
+    <%= tag.label text, for: html_for, **component_helper.all_attributes %>
   <% end %>
 <% else %>
-  <%= tag.label text, id: id, for: html_for, class: css_classes, dir: right_to_left ? "rtl" : nil %>
+    <%= tag.label text, for: html_for, **component_helper.all_attributes %>
```
andysellick, line 27: Nitpick: extra indentation here not needed?

app/views/govuk_publishing_components/components/_label.html.erb
```diff
@@ -11,18 +11,20 @@
...
+  component_helper.add_class("gem-c-label govuk-label")
+  component_helper.add_class("govuk-label--s") if bold
+  component_helper.add_class("govuk-radios__label") if is_radio_label
+  component_helper.add_class("govuk-label--#{heading_size}") if heading_size
+  component_helper.set_dir("rtl") if right_to_left
 %>
 
 <% if is_page_heading %>
   <%= tag.h1 text, class: "govuk-label-wrapper" do %>
-    <%= tag.label text, id: id, for: html_for, class: css_classes, dir: right_to_left ? "rtl" : nil %>
```
andysellick, line 22: This is worth some testing, but I think because the component wrapper helper looks out for a value called `id` in `local_assigns`, we don't need to explicitly declare it at the top of the template. Unless there's some logic elsewhere that relies on it (e.g. `<% if id %>` or we need to set a default value other than `nil`.

I remembered this last week when I was working on some related code. It might be worth looking back to see if there's some lines could be removed from other components like this (unless I'm completely wrong, as I said, worth testing).

andysellick (approved):
Non-blocking comments, happy to approve 👍

app/views/govuk_publishing_components/components/_label.html.erb
AshGDS: Thanks, good spot - mistakenly was aligning it with the other label, which is nested one tab further in.

PR #4365, opened 2024-11-04
Improve metadata component to support RTL print styles

matthillco:
## What
Improvements to the metadata component to improve RTL print styles.

This is part of the work to improve print styles more generally. [Trello](https://trello.com/c/6E5UoTiK/335-rtl-print-styles-have-been-considered-and-implemented-where-required)

_Note: This is a dependency of https://github.com/alphagov/government-frontend/pull/3404_

## Why
Some pages are broken for foreign languages that read right to left, with styles not set correctly for the changed direction.

## Visual Changes

| Before    | After |
| -------- | ------- |
| image  |image  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Use component wrapper on contextual sidebar ([PR #4561](https://github.com/alphagov/govuk_publishing_components/pull/4561))
 * Correctly translate 'Published' word to Arabic ([PR #4563](https://github.com/alphagov/govuk_publishing_components/pull/4563))
 * Style breadcrumbs in RTL (right-to-left) writing mode ([PR #4559](https://github.com/alphagov/govuk_publishing_components/pull/4559))
+* Improve metadata component to support RTL print styles ([PR #4365](https://github.com/alphagov/govuk_publishing_components/pull/4365))
 
 ## 49.0.0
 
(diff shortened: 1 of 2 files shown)


AshGDS (changes requested):
Could you please add a right to left example to the component using the `context` parameter to `metadata.yml` ? E.g.

```
  right_to_left:
    data:
      other:
        Release date: 23 August 2018 9:30am
        Reason for change: Date of release brought forward from 23 August 2018 to 21 August 2018 as data in its final form is now available to be published on this new date.
      title: The release date has been changed
    context:
      right_to_left: true
```

But instead of the content above, content that better reflects what's on production?

Thanks

AshGDS:
Sorry, just realised there's already a RTL example in the component - so up to you if you think it's worth doing.

matthillco:
> Sorry, just realised there's already a RTL example in the component - so up to you if you think it's worth doing.

@AshGDS I think it's OK as it stands because this fix was just about ensuring the terms are positioned correctly for RTL. For some reason, there are different possibilities for where the `.direction-rtl` class gets applied for child elements to hook into, and this PR is only to fix one of the possibilities that was being missed. 

In the future perhaps there will be an opportunity to revisit RTL support as the way its handled is very inconsistent across views and components.

AshGDS (approved):

PR #4368, opened 2024-11-06
Append no-actions class to rows without actions

pezholio:
See https://design-system.service.gov.uk/components/summary-list/#showing-rows-with-and-without-actions

This ensures that the bottom border is drawn correctly in some browsers, as well as ensures that items without actions span the full width of the component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,9 @@
   of the commit log.
 
 ## Unreleased
+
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
LGTM 👍 

Non-blocking, but for readability maybe it's worth putting the classes into their own variable now? E.g.:

```erb
row_classes = "govuk-summary-list__row"
row_classes << " govuk-summary-list__row--no-actions" if row[:actions].blank?
```

pezholio:
Good plan 👍

PR #4371, opened 2024-11-06
Add GA4 tracking for search autocomplete

csutter:
## What
This PR:
* adds tracking attributes to the `search_with_autocomplete` component
* amends the `Ga4SearchTracker` to pick them up as part of the `search` event
* extends the GA4 schema to cover the new `search` event fields

## Why
We will soon be going live with autocomplete for site search on GOV.UK and need to track user interactions to see if it works as intended.

See [measurement approach document (internal)](https://docs.google.com/document/d/1X6vMjJpr8SiLegBGjnjL4-HSEScG-SWTgk2uy7bnWiM) for details on why we're measuring what we're measuring, and the approach taken.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
@@ -87,13 +87,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       url.searchParams.set('q', query)
       fetch(url, { headers: { Accept: 'application/json' } })
         .then(response => response.json())
-        .then((data) => { populateResults(data[this.sourceKey]) })
+        .then((data) => {
+          const results = data[this.sourceKey]
+
+          this.setTrackingAttributes(query, results)
+          populateResults(results)
+        })
         .catch(() => { populateResults([]) })
+    }
+
+    // Set tracking attributes on the input field. These can be used by the containing form's
+    // analytics module to track the user's interaction with the autocomplete component.
+    setTrackingAttributes (query, results) {
+      const formattedResults = results.slice(0, 5).join('|')
+
+      this.$autocompleteInput.dataset.autocompleteTriggerInput = query
+      this.$autocompleteInput.dataset.autocompleteSuggestions = formattedResults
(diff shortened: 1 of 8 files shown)


app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
```diff
@@ -87,13 +87,31 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
 
+    // Set tracking attributes on the input field. These can be used by the containing form's
+    // analytics module to track the user's interaction with the autocomplete component.
+    setTrackingAttributes (query, results) {
+      const formattedResults = results.slice(0, 5).join('|')
+
+      this.$autocompleteInput.setAttribute('data-autocomplete-trigger-input', query)
+      this.$autocompleteInput.setAttribute('data-autocomplete-suggestions', formattedResults)
+      this.$autocompleteInput.setAttribute('data-autocomplete-suggestions-count', results.length)
+      this.$autocompleteInput.setAttribute('data-autocomplete-accepted', false)
```
kevindew, line 107: Can we use dataset to set these?

spec/javascripts/components/search-with-autocomplete-spec.js
```diff
@@ -215,4 +215,69 @@ describe('Search with autocomplete component', () => {
...
+    const input = fixture.querySelector('input')
+
+    stubSuccessfulFetch([
+      'my favourite album is red',
+      'my favourite album is midnights',
+      'my favourite album is lover',
+      'my favourite album is folklore',
+      'my favourite album is reputation',
+      'my favourite album is 1989'
+    ])
```
kevindew, line 252: Had not realised you were a swiftie 😁

kevindew (commented):
Looks good to me. Only thing I'm a bit confused by is using different methods to set/get data attributes - it's unclear to me if that's on purpose or not.

spec/javascripts/components/search-with-autocomplete-spec.js
csutter: Oh you'll enjoy the Finder Frontend search tests 😂 I've almost crowded out @sihugh's Harry Potter references!

kevindew (approved):
Looks good 👍

PR #4372, opened 2024-11-07
Search with autocomplete: Work around Enter edge case

csutter:
## What
The `accessible-autocomplete` component has an edge case where when the menu is visible, [it `preventDefault()`s on the Enter key event](https://github.com/alphagov/accessible-autocomplete/blob/b8369d037fd5d26d30a831ed73baa3dd566293ef/src/autocomplete.js#L361), even if the user hasn't yet put keyboard focus on a suggestion. This results in a scenario where the user types something, does _not_ interact with the autocomplete menu at all, and then hits Enter to try to submit the form - but it isn't submitted.

This implements a workaround that adds our own event listener and ensures `requestSubmit` is run for the form if it hasn't been already.

- Add event listener for `keydown` event and submit the containing form ourselves if the key is `Enter`
- Split out code meant to run on accepting a suggestion (`onConfirm`) from form submission logic (`submitContainingForm`) so we can reuse the latter
- Add a check to `submitContainingForm` that it hasn't already been called for the component anyway

## Why
This is not something we can fix upstream as it is usually desirable behaviour for the library.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Add GA4 tracking for search autocomplete ([PR #4371](https://github.com/alphagov/govuk_publishing_components/pull/4371))
 * Append no-actions class to rows without actions in Summary Cards ([PR #4368](https://github.com/alphagov/govuk_publishing_components/pull/4368))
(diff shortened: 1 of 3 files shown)


kevindew (approved):
Looks good to me, seems a nasty issue. Real shame that the autocomplete doesn't have more event hooks.

PR #4375, opened 2024-11-07
Rename gem-print-link and gem-print-links-within

AshGDS:
## What / Why
- [The component wrapper helper has some validation that restrict what classes can be added to components.](https://github.com/alphagov/govuk_publishing_components/blob/a2285efb2dde7d156178a8e8e0eb6c7bd903d6a7/lib/govuk_publishing_components/presenters/component_wrapper_helper.rb#L100)
- This validation doesn't like `gem-print-link` and `gem-print-links-within` as they don't conform to the validation.
- Therefore, I discussed with Andy and we've renamed them to use `gem-c-` so that they pass the validation.
- This will unblock me adding the component wrapper to more components.

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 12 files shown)


matthillco (approved):
Looks good to me.

PR #4376, opened 2024-11-07
Add a content block component

pezholio:
Currently, the logic for rendering content blocks logic is locked up in the Publishing API. The Content Modelling team has need to use this logic in multiple places outside the Publishing API, so rather than have to create an API endpoint to render a content block, we decided we should have one source of truth.

We initially looked at incorporating this into Govspeak, but the initial rendering of content blocks when a content block changes has to take place in the Publishing API, and currently, there are some publishing apps (e.g. Whitehall) that render Govspeak _before_ sending content to the Publishing API, and this is too difficult to unpick without a lot of work. 

Additionally, this will give us a degree of futureproofing for the longer term, when we start to move away from using the Content Store towards rendering content directly in frontend apps via the (currently in development) GraphQL API.

After some discussing with our team and @richardTowers, we decided that this gem was the best place for this logic to live. More context is in a draft ADR here - https://github.com/alphagov/publishing-api/pull/2974

Once this is in, we’ll call this component when rendering content blocks in Publishing API, so the Publishing Components gem becomes the source of truth.

Currently some of the blocks look a little bare bones, but we will be expanding on this in the near future, especially when it comes to the contact block.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Add component wrapper to feedback component ([PR #4351](https://github.com/alphagov/govuk_publishing_components/pull/4351))
 * Add component wrapper helper to error message component ([PR #4359](https://github.com/alphagov/govuk_publishing_components/pull/4359))
 * Require layout_for_public will not render wrapper unless `for_static: true` is set. ([PR #4255](https://github.com/alphagov/govuk_publishing_components/pull/4255))
+* Add a Content Block component ([PR #4376](https://github.com/alphagov/govuk_publishing_components/pull/4376))
 
 ## 45.1.0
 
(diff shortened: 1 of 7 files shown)


andysellick:
@pezholio can you put a bit more detail into the description to explain what this is and why it's needed? Appreciate that you may still be working on this so it might not be complete, but currently I'm wondering what this does over say, a govspeak component?

pezholio:
@andysellick Have updated 👍

pezholio:
I've also added a link to an ADR to outline our thinking in more detail

maxgds:
Are you ok to close this following our recent discussion?

pezholio:
Yep, sorry! Will close now 👍

PR #4377, opened 2024-11-08
Expand event tracker documentation to include advice for multiple elements

dnkrj:
## What
- Expand event tracker documentation to include advice for multiple elements
- Make clear that the `ga4-event-tracker` should not be placed at the top of the page

## Why
- To avoid other teams experiencing the double tracking issue the was present on Whitehall
  - https://trello.com/c/Mt7YXtIc/3040-bugs-double-tracking-in-some-events


ryanb-gds (approved):
This helps to clarify things, thanks Dan

PR #4381, opened 2024-11-08
Track autocomplete usage in session storage

csutter:
## What
This _temporarily_ persists acceptance of an autocomplete suggestion in session storage as part of the `Ga4SearchTracker`, so we can enhance the `page_view` event on the next page view (the results page) with the `tool_name` attribute.

## Why
This will allow us to correlate users' interactions on the results page with whether they have accepted an autocomplete suggestion to get there, for example to answer the question: "Is a user accepting an autocomplete suggestion more or less likely to interact with a search result?"


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
```diff
@@ -60,6 +60,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         // having seen the autocomplete and not used it, and not having seen it at all)
         if (this.$searchInput.dataset.autocompleteAccepted === 'true') {
           data.tool_name = 'autocomplete'
+
+          // Keep track of the fact that the autocomplete was accepted, so that we can use this as
+          // part of the `page_view` event on the subsequent page load
+          window.sessionStorage &&
+            window.sessionStorage.setItem('searchAutocompleteAccepted', 'true')
```
andysellick, line 67: I'm sure you're aware, but local storage comes under the umbrella of cookie consent, so we need to be careful with this stuff. I think this one is probably fine, but I wanted to mention it.

csutter: Absolutely 💯 – both the reading and writing happen in module functions that won't run if the module hasn't been started, which it only is if consent has been given!

andysellick: Ah of course, sorry. Thanks for clarifying 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -109,6 +110,17 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
       return searchTerm
     },
 
+    getToolName: function () {
+      // This is set by the Ga4SearchTracker module if the user has accepted an autocomplete
+      // suggestion on the search field.
+      if (window.sessionStorage && window.sessionStorage.getItem('searchAutocompleteAccepted')) {
```
kevindew, line 116: Is this object detection in case a browser doesn't support window.sessionStorage? it seems such a supported API that this seems a bit superfluous given our browser support (apologies if I've misunderstood).

kevindew (approved):
Just a minor quibble on object detection

PR #4382, opened 2024-11-08
Search with autocomplete: Set minimum query length

csutter:
## What
This stops the `search_with_autocomplete` component from wasting requests trying to fetch suggestions when our backend won't provide any for query lengths < 3.

## Why
Fewer pointless requests is better!

Note that I've not bothered making this configurable - it's unlikely for the time being that this component will be used with a different source/backend, so there isn't much point in adding a configuration option. I've also not added a changelog option as that felt over the top 😄


kevindew (approved):
Looks good.

Issue #3482, opened 2023-07-06
Review `metadata` component usage


MartinJJones:
## What
We have a metadata component available in the publishing components gem that can be used in rendering applications.

However, several applications appear to also have their own metadata components, applications could be using their own variation of the component for several reasons, either because the `metadata` component was not available at the time, or it does not provide all the requirements.

We should look to either directly replace the `metadata` components in the rendering app with the one from the publishing components gem, or update the `metadata` component in the gem where appropriate, so the app component can be replaced.

## Why
- Help keep the UI consistent across applications
- Helps to keep the codebase DRY, we only need to make changes in one place
- Make it easier to know which component to choose from
- We may improve the existing component by providing additional features identified in rendering applications

The screenshots below show examples of the varying implementation of the `metadata` component, as well as the different HTML used, visually they will have different font-size, color and spacing.

### Content Data Admin

| Publishing component | Content data admin |
| --- | --- |
| <img width="669" alt="metadata-gem" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/cf9ef530-9bb1-449a-80ac-0051f6714d99"> | <img width="352" alt="content-data-admin" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/6fb27672-6888-4b08-8217-5fac5571e614"> |

### Frontend

| Publishing component | Frontend |
| --- | --- |
| <img width="239" alt="metadata-last-updated" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/2a376950-ae27-4d28-8cad-738198358712"> | <img width="290" alt="metadata-government-frontend" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/c41744a2-001d-4320-b4df-d116a7a9bb96"> |

### Content Publisher

| Publishing component | Content publisher |
| --- | --- |
| <img width="316" alt="metadata-gem-c-2-lines" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/c1067e34-318b-48cf-80e6-d126b072de8b"> | <img width="438" alt="metadata-content-publisher" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/30167367-b2ac-4a69-84a0-083b97b5dd94"> |

### Government Frontend

| Publishing component - inversed | Government Frontend - metadata |
| --- | --- |
[...]

PR #4383, opened 2024-11-11
Add heading option to metadata component

jon-kirwan:
## What
Add heading option to metadata component.

## Why

The important metadata block component, used in Government Frontend e.g. https://www.gov.uk/government/statistics/announcements/core-statistics-book-august-2018, can be replaced with the metadata block component. However, some block instances include a title so we should provide a heading option and we can then replace the app variant with this component.

## Visual Changes

govuk-publishing-components dev gov uk_component-guide_metadata_with_title_preview

govuk-publishing-components dev gov uk_component-guide_metadata_on_a_dark_background_with_title_preview

## Anything else

- https://github.com/alphagov/govuk_publishing_components/issues/3482
- https://github.com/alphagov/government-frontend/pull/3414


MartinJJones (approved):

PR #4392, opened 2024-11-12
Initial implementation of "add another" component

ryanb-gds:
__VERY EARLY DRAFT__

Just looking for really fast feedback on whether we think this is likely to make a suitable component and agree an implementation approach. I've left some comments in the code for things I would like feedback on. At the moment the code is more in the style of Whitehall than Publishing components so I can do some refactoring before I get a final review.

I'm also catching up with the Mainstream team next week to see what their needs are. It may be useful to add some more options to the component.

## What

This is more or less a lift and shift of [the existing "Add another" component from Whitehall Publisher](https://github.com/alphagov/whitehall/blob/f67e53356728907f911d8fa51066afb04268bc8a/app/assets/javascripts/admin/modules/add-another.js). The Whitehall component in turn was inspired by [the MoJ's "Add Another" component](https://design-patterns.service.justice.gov.uk/components/add-another/).

## Why

We want to share this component across Whitehall and Specialist Publisher. The Mainstream team are also potentially considering using it, so we believe there is a strong need here.

The component takes a fragment based approach, which means that the consumer is responsible for rendering the buttons for removing a set of fields from their form. This decision was made so that the buttons can be present as part of the HTML markup, enabling the component to function without Javascript.

The MoJ component and the original Whitehall component create the remove button within Javascript, which means they requires JS to be enabled to function. Making the user responsible for rendering the removal button means that they can use it as a form submit button in cases where JS is not enabled, and re-render the page without the specified field set. When JS is enabled, the module overrides the types of the buttons to "button" so that form submissions do not take place.

## Visual Changes

Screenshot 2024-11-18 at 15 14 32

Trello: https://trello.com/c/BIfsF9pP

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/add-another.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/add-another.js
@@ -8,32 +8,27 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
   AddAnother.prototype.init = function () {
     this.createButtons()
-    this.initialiseClickEvent()
   }
 
   AddAnother.prototype.createButtons = function () {
     function createButton(textContent, additionalClass = "") {
       var button = document.createElement('button')
-      button.className = "js-add-another__add-button gem-c-button govuk-button " + additionalClass
+      button.className = "gem-c-button govuk-button " + additionalClass
       button.type = "button"
       button.textContent = textContent
       return button
     }
 
-    addButton = createButton(this.module.dataset.addButtonText)
+    addButton = createButton(this.module.dataset.addButtonText, "js-add-another__add-button")
     addButton.addEventListener('click', this.addFields.bind(this))
     this.module.appendChild(addButton)
-  }
 
-  AddAnother.prototype.initialiseClickEvent = function () {
-    this.module.addEventListener(
-      'click',
-      function (e) {
-        if (e.target.classList.contains('js-add-another__remove-button')) {
-          this.removeFields(e.target)
-        }
-      }.bind(this)
-    )
+    var repeatedFieldsets = document.querySelectorAll('.js-add-another__repeated-fields')
+    repeatedFieldsets.forEach(function (fieldset) {
+      var removeButton = createButton("Delete", "js-add-another__remove-button")
+      removeButton.addEventListener('click', this.removeFields.bind(this))
+      fieldset.appendChild(removeButton)
+    }.bind(this))
   }
 
   AddAnother.prototype.addFields = function (event) {
@@ -71,7 +66,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       .focus()
   }
 
-  AddAnother.prototype.removeFields = function (button) {
+  AddAnother.prototype.removeFields = function (event) {
+    var button = event.target
     var set = button.parentNode
     var input = set.querySelectorAll(
       'input:not([type="hidden"]), select, textarea'
--- a/app/assets/javascripts/govuk_publishing_components/components/add-another.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/add-another.js
@@ -7,8 +7,22 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   AddAnother.prototype.init = function () {
+    this.createButtons()
     this.initialiseClickEvent()
-    this.preventFormSubmissions()
+  }
+
+  AddAnother.prototype.createButtons = function () {
+    function createButton(textContent, additionalClass = "") {
+      var button = document.createElement('button')
+      button.className = "js-add-another__add-button gem-c-button govuk-button " + additionalClass
+      button.type = "button"
+      button.textContent = textContent
+      return button
+    }
+
+    addButton = createButton(this.module.dataset.addButtonText)
+    addButton.addEventListener('click', this.addFields.bind(this))
+    this.module.appendChild(addButton)
   }
 
   AddAnother.prototype.initialiseClickEvent = function () {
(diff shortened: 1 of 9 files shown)


app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  function AddAnother (module) {
```
ryanb-gds, line 5: Is "Add Another" the best name for this component? It's consistent with the MoJ naming which has a certain amount of value, but I feel that it isn't hugely descriptive.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
...
+    var removeButtons = this.module.querySelectorAll('.js-add-another__remove-button')
+    addButton.setAttribute('type', 'button')
+    removeButtons.forEach(function (removeButton) {
+      removeButton.setAttribute('type', 'button')
+    })
+  }
+
+  AddAnother.prototype.addFields = function (button) {
+    var allFields = button.parentNode.querySelectorAll(
+      '.js-add-another__repeated-fields'
```
ryanb-gds, line 38: Would appreciate any ideas for better names on this class

app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
...
+      .querySelectorAll('input, textarea, select')
+      .forEach(function (element) {
+        element.value = ''
+      })
+
+    // Increment values for id, for and name of cloned fields
+    this.setValues(newFields, null)
+
+    // Add cloned fields to the DOM
+    button.before(newFields)
```
ryanb-gds, line 62: It strikes me that there should probably be an announcement for screen reader users that content has been added to the DOM. Would appreciate some guidance on this from somebody more knowledgeable in the area.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
...
+    set.remove()
+
+    var sets = this.module.querySelectorAll('.js-add-another__repeated-fields')
+
+    // Add hidden field for removed set
+    var hiddenField = document.createElement('input')
+    hiddenField.type = 'hidden'
+    hiddenField.classList.add('js-hidden-destroy')
+    hiddenField.id = baseId.replace(/_[a-zA-Z]+$/, '__destroy')
+    hiddenField.name = baseName.replace(/\[[_a-zA-Z]+\]$/, '[_destroy]')
```
ryanb-gds, line 87: I have a slight concern that this is obviously very tightly coupled to Rails, but I think that's acceptable in the Publishing Components gem. It does impose certain expectations on the form submission route but I think that's in line with Rails conventions and is unlikely to vary between apps.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
...
+    hiddenField.setAttribute('value', 'true')
+    this.module.append(hiddenField)
+
+    // Hide "Remove" button if only first set displayed
+    if (sets.length === 1) {
+      sets[0].querySelector('.js-add-another__remove-button').style.display =
+                'none'
+    }
```
ryanb-gds, line 95: Not sure about hiding the button for the first set - presumably there might be situations where it would be appropriate to have no input values? However I wasn't certain about this so have left it in place for now.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -0,0 +1,159 @@
...
+
+    // Move focus to first visible field
+    sets[0]
+      .querySelectorAll('input:not([type="hidden"]), select, textarea')[0]
+      .focus()
+  }
+
+  // Set values for index, for and name of supplied fields
+  AddAnother.prototype.setValues = function (set, index) {
```
ryanb-gds, line 104: This function feels like it could be expressed more elegantly, I will try to adapt the MoJ version if possible: https://github.com/ministryofjustice/moj-frontend/blob/1684952f8c443c2910264676a5c1c68970c01988/src/moj/components/add-another/add-another.js#L43

app/views/govuk_publishing_components/components/_add_another.html.erb
```diff
@@ -0,0 +1,10 @@
+<%
+  add_button_text ||= "Add another"
+%>
```
ryanb-gds, line 3: Would it be appropriate to apply the [wrapper helper](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component-wrapper-helper.md) to this component? I thought possibly not as this wrapper element isn't really particularly significant

app/views/govuk_publishing_components/components/_add_another.html.erb
```diff
@@ -0,0 +1,10 @@
...
+
+<div data-module="add-another" class="gem-c-add-another">
+  <div class="gem-c-add-another__repeated-fields js-add-another__repeated-fields">
```
ryanb-gds, line 6: I would have liked to have used the fieldset component here, but it doesn't implement the component wrapper so I can't add the JS hook class to it. Do you think I should add the wrapper to the fieldset class?

app/views/govuk_publishing_components/components/add_another/_remove_button.html.erb
```diff
@@ -0,0 +1,11 @@
+<%
+  margin_bottom ||= 0
+%>
+
+<%= render "govuk_publishing_components/components/button", {
+  text: "Delete",
```
ryanb-gds, line 6: Could set up an option for this if it's considered worthwhile

ryanb-gds (commented):
I have quite a few questions for readers

dnkrj (commented):
I'm not sure that having the delete button visible to non-js users is the right direction to take this.
The handling of that post request seems like it would require quite a lot of custom complexity.
My inclination is that non-JS users simply clear the values of the fields that they want to delete and have them cleaned up on save.

Happy to schedule some time to talk though/pair on this.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
dnkrj: I'm pro sticking with MoJ unless we can come up with something significantly better

PR #4400, opened 2024-11-13
Search with autocomplete: track non-empty suggestions

csutter:
## What
This amends the `search_with_autocomplete` component's tracking attributes to track the last non-empty set of suggestions the user has seen, rather than the last set of suggestions regardless of if there were any.

## Why
This allows us to track and evaluate why a user doesn't accept any of the suggestions they were offered, choosing instead to type in their own query.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
@@ -114,10 +114,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     // Set tracking attributes on the input field. These can be used by the containing form's
     // analytics module to track the user's interaction with the autocomplete component.
     setTrackingAttributes (query, results) {
-      const formattedResults = results.slice(0, 5).join('|')
+      // Only set the suggestions attribute when results actually come back (so this attribute
+      // tracks the last seen non-empty suggestions, even if the user then amends their query to one
+      // that doesn't generate any)
+      if (results.length > 0) {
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/components/search-with-autocomplete.js
```diff
@@ -114,10 +114,15 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     // Set tracking attributes on the input field. These can be used by the containing form's
     // analytics module to track the user's interaction with the autocomplete component.
     setTrackingAttributes (query, results) {
-      const formattedResults = results.slice(0, 5).join('|')
+      // Only set the suggestions attribute when results actually come back (so this attribute
+      // tracks the last seen non-empty suggestions, even if the user then amends their query to one
+      // that doesn't generate any)
+      if (results.length > 0) {
+        const formattedResults = results.slice(0, 5).join('|')
+        this.$autocompleteInput.dataset.autocompleteSuggestions = formattedResults
```
kevindew, line 122: Should this be renamed autocompleteLastSuggestions to better reflect it's intent? Or is that a quagmire?

csutter: Sadly falls under quagmire - tl;dr is it's name to be consistent with the (meh) naming we've already agreed for the analytics event. Hence the unusual amount of commentary preceding it 😁

kevindew: Not to worry!

kevindew (approved):

PR #4405, opened 2024-11-14
Adjust chart options and settings

andysellick:
## What
- include an option to allow the Y axis to not start at zero but based on the data
- remove unnecessary formatting options
- add a point_size option to allow control over the chart appearance
- add examples of these options and behaviours

## Why
Required for chart usage.

## Visual Changes
With a different point size:

Screenshot 2024-11-14 at 18 08 35

With a break in the data and not starting at zero on the Y axis:

Screenshot 2024-11-14 at 18 08 40

Trello card: https://trello.com/c/aKI3YGL4

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/chart.yml
+++ b/app/views/govuk_publishing_components/components/docs/chart.yml
@@ -6,7 +6,7 @@ body: |
   The chart relies upon chartkick and renders using JavaScript, so the table is provided as a fallback for a lack of JavaScript,
   an accessible view on the data for screenreaders, and a simple view of the raw data for all users.
 
-  The `chart_overview` option can be used to provide an explanation for screen reader users of what the graph shows.
+  The `chart_overview` option can be used to provide an explanation for screen reader users of what the chart shows.
 accessibility_criteria: |
   Charts must:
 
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -137,6 +137,77 @@ examples:
...
+            - 740
+            - 820
+            - 570
+            - 950
+            - 810
+            - 610
+            - 770
+            - 530
+  scale_y_axis_based_on_data:
+    description: By default the Y-axis will always start at zero, but this may not always be required. Note also that if passed gaps in the data, the chart will simply not render points and discontinue the line.
```
AshGDS, line 173: Nitpick: I found `Note also that if passed gaps in the data` a bit confusing to read the first couple times, maybe it could be reworded slightly?

AshGDS (commented):
Looks good to me, just had a small comment.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: That was a bit of a mess, wasn't it? I've reworded 👍

AshGDS: Thanks @andysellick 👍

AshGDS (approved):

PR #4409, opened 2024-11-15
Change icon and link colours on square option for share links

maxgds:
## What
Updates the colour of the share links in square option to meet design updates

Trello: https://trello.com/c/37SyRmUF/150-rework-share-links

### Before
image

### After

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change colours on square option for share links ([PR #4409](https://github.com/alphagov/govuk_publishing_components/pull/4409))
+
 ## 45.4.1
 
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
LGTM 👍 

Non-blocking nitpick thing, which I probably should've done in the initial implementation of this:

Do you think it's worth modifying the focus style `box-shadow` so that we don't have that 1px yellow border on top? E.g.

```CSS
.gem-c-share-links__link:focus
    box-shadow: 0 0px #fd0, 0 4px #0b0c0c;
}
```
(The colour values would be govuk vars instead of course)

### Before
image

### After
image

maxgds:
> Do you think it's worth modifying the focus style `box-shadow` so that we don't have that 1px yellow border on top? E.g.

Hmm, I'm not sure, maybe one for design. I think maybe we should actually be adding more yellow to the left to indicate that the icon is bound within the clickable space. I'll leave it as is for now.

PR #4417, opened 2024-11-19
Add margin_top_until_tablet option to chat entry component

JamesCGDS:
## What
This PR adds a `margin_top_until_tablet` option to the chat entry component, which adds `margin-top: 35px` on smaller screen sizes (i.e. <= tablet). I've opted for applying this via a `.gem-c-chat-entry--margin-top-until-tablet` rather than using the `margin_top` helper as 1). it should only be applied at certain screen sizes and 2). the specific amount of margin required isn't supported by the helper.

## Why
Design requirements - [trello card](https://trello.com/c/iOuUw0vs/2157-entry-banner-tweaks).

## Visual Changes
N/A

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix yellow focus colour overspill ([PR #4418](https://github.com/alphagov/govuk_publishing_components/pull/4418))
+* Add margin_top_until_tablet option to chat entry component ([PR #4417](https://github.com/alphagov/govuk_publishing_components/pull/4417))
 
 ## 45.5.0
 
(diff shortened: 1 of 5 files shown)


jon-kirwan (commented):
Looks good to me. I just have one comment/suggestion: should we include a component guide example?

I also wondered about the name `margin_top` and whether it might cause any confusion because of the approach used elsewhere for setting margins using the [shared helper](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_print_link.html.erb#L8-L9) and passing it a number. That said, I think it’s fine, as otherwise would probably require a very specific name like `margin_top_until_tablet`.

JamesCGDS:
Thanks @jon-kirwan, great shouts - have made both those changes 👍

jon-kirwan (approved):
That's great! Thanks for the changes @JamesCGDS

PR #4418, opened 2024-11-20
Fix yellow focus colour overspill

AshGDS:
## What / Why

- On the share links with square icons, the focus styles overspilled by a few pixels over the top of the element
- I showed this to our designer, he agreed that we should get rid of this few pixel overspill
- See screenshots

### Before

image

### After

image


maxgds (approved):
Maybe I should have listened to you in the first place :D

PR #4420, opened 2024-11-20
Introduce "add another" component

dnkrj:
## What
- Add the component wrapper helper to fieldsets
- Introduce the add another component based on the [MoJ design system](https://design-patterns.service.justice.gov.uk/components/add-another/) and [existing component in Whitehall](https://github.com/alphagov/whitehall/blob/main/app/assets/javascripts/admin/modules/add-another.js). 
- An empty field and destroy checkboxes for the existing fields are required and displayed to the user when javascript is disabled in keeping with rails conventions.
- When Javascript is enabled, an "add another" button is introduced to allow users to add copies of the empty field and the checkboxes replaced with "Delete" buttons which hide the fields and checks the appropriate checkbox.

## Why
 - This component is to be used when users need to add similar information a couple of times, such as several featured links for an organisation.
 - The current implementation in Whitehall requires JavaScript to work correctly
 - We want to use the pattern in Specialist Publisher and Mainstream.
 - https://trello.com/c/BIfsF9pP/3139-move-add-another-pattern-to-govukpublishingcomponents

## Screenshots
To see further examples of usage, the [migration PR for Whitehall](https://github.com/alphagov/whitehall/pull/9644) can be looked at.
> The screenshots are taken with `input` and `checkbox` styles added to the page.
> This are not present in the code in the PR.
### With Javascript
Screenshot 2024-11-25 at 17 45 43
Screenshot 2024-11-25 at 17 45 55

### Without Javascript
Screenshot 2024-11-25 at 17 46 56

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_add-another.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_add-another.scss
@@ -1,0 +1,12 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk/components/button/button";
+@import "govuk/components/fieldset/fieldset";
+
+.gem-c-add-another__remove-button {
+  margin-top: govuk-spacing(6);
+  margin-bottom: 0;
+}
+
+.js-add-another__remove-button--hidden {
+  display: none;
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Add 'toggle' module to each expandable region in metadata component ([PR #4397](https://github.com/alphagov/govuk_publishing_components/pull/4397))
 * Add a description for the metadata block 'see_updates_link' example ([PR #4447](https://github.com/alphagov/govuk_publishing_components/pull/4447))
+* Add the component wrapper to the fieldset component ([PR #4420](https://github.com/alphagov/govuk_publishing_components/pull/4434))
 
 ## 45.9.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
 * Add 'toggle' module to each expandable region in metadata component ([PR #4397](https://github.com/alphagov/govuk_publishing_components/pull/4397))
 * Add a description for the metadata block 'see_updates_link' example ([PR #4447](https://github.com/alphagov/govuk_publishing_components/pull/4447))
 * Add the component wrapper to the fieldset component ([PR #4420](https://github.com/alphagov/govuk_publishing_components/pull/4434))
+* Introduce "add another" component ([PR #4420](https://github.com/alphagov/govuk_publishing_components/pull/4434))
 
 ## 45.9.0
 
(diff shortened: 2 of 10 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_add-another.scss
```diff
@@ -0,0 +1,6 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk/components/button/button";
```
jon-kirwan, line 2: I think some additional styles need to be imported for:
```
@import "govuk/components/checkboxes/checkboxes";
@import "govuk/components/input/input";
@import "govuk/components/label/label";
```

jon-kirwan (commented):
It's looking good 👍 I've just added a couple of smallish comments.

Also, I think the `add-another` styles need to be added to the `_all_components.scss` stylesheet.

jon-kirwan (commented):
Are there any accessibility issues with having multiple labels and delete buttons that all share the same name? For example, if a screen reader user navigates the form by buttons using the rotor, could this cause confusion? Would it be worth considering adding a visually hidden element next to each label and delete button to make them uniquely identifiable e.g.
```
<label>Full name <span class="visually-hidden">1</span></label>
```
and
```
<button>Delete <span class="visually-hidden">full name 1</span></button>
```

app/assets/stylesheets/govuk_publishing_components/components/_add-another.scss
dnkrj: The checkboxes and inputs are provided by the caller of the component, so it's up to whoever calls the component to ensure that they are covered (the component could also contain other input types such as radio buttons or something entirely custom)

dnkrj:
Thanks for the review @jon-kirwan!
You're right that the repeated button labels aren't good for accessibility.
It's a little bit difficult to update them programatically since we don't know exactly what the caller will put in so I've updated the code to wrap all of the fields in a numbered fieldset (and updated the screenshots to match).
This follows one of examples for ["Providing Accessible Names and Descriptions"](https://www.w3.org/WAI/ARIA/apg/practices/names-and-descriptions/#naming_with_legends) in the ARIA Authoring Practices Guide (APG). Let me know what you think!

app/assets/stylesheets/govuk_publishing_components/components/_add-another.scss
dnkrj: In this case it's similar to the the other inputs.

Whilst we expect the user to pass in a checkbox - it doesn't necessarily mean that they're using the govuk_publishing_components checkbox, they may want to style checkboxes in a particular way.

In the case that they are using the govuk_publising_components checkbox they'll be calling the render so should also be responsible for ensuring that the checkbox style is included - rather than am implicit dependancy through this component.

Happy to chat about it more if you want or if you still think we need the dependancy here, I can add it, just wanted to explain why I think it's better not to have it.

jon-kirwan (approved):
Great work and thanks for answering all my questions @dnkrj!

davidtrussler:
Hello @dnkrj Just been looking at this - looks really good. The only thing that came to mind was to maybe explain a bit more in the documentation about the empty item: I don't think it's immediately obvious what that does and how it should look.

PR #4421, opened 2024-11-20
Fix `search_with_autocomplete` double submit scenario

csutter:
## What
Our workaround for the `accessible-autocomplete` library hogging the Enter key when the menu is open was too greedy, and also submitted the form when the menu in fact wasn't open.

This adds a check to the workaround to ensure the menu is actually shown, as the behaviour doesn't actually happen otherwise.

## Why
Double submits are unnecessary, and while it wouldn't cause any user-visible impact, it would cause some problems with analytics.


kevindew (approved):
Looks good to me 👍

PR #4427, opened 2024-11-22
Fix chart minimal option

andysellick:
## What
Fixes the `minimal` option in the chart component, because it wasn't hiding the legend, removing interactivity, or reducing point size automatically, as it should have been.

## Why
Managed to break this as part of https://github.com/alphagov/govuk_publishing_components/pull/4424

## Visual Changes
None (except to fix minimal mode back to what it should be).


AshGDS (approved):
LGTM 👍 

I don't remember any Percy changes needing to be reviewed in the last PR, so it's interesting that this slipped through the cracks!

andysellick:
I think it was probably that there were so many visual changes it was hard to spot.

PR #4429, opened 2024-11-25
search_with_autocomplete: Standardise trigger input

csutter:
## What
We want to apply the search input standardisation for the search event not just on the search term itself, but also on the (partial) user input that led to the autocomplete suggestions being triggered.

- Refactor `Ga4SearchTracker.searchTerm()` into a more generic `.standardiseInput()` that we can use for both search term and trigger input value
- Standardise search term for both using this new function
- Add missing changelog entry for a previous PR (sorry!)

## Why
This will make our search events more consistently sanitised for analytics purposes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


davidgisbey (approved):
I've gone through the module for context and it looks good to me 👍

PR #4435, opened 2024-11-26
Limit chart heading width

andysellick:
## What
Remove the padded option for the chart and limit the heading width to two thirds page width.

## Why
Padded option no longer needed, headings need to be a limited width when charts are full page width.

## Visual Changes
Chart with a long heading now wraps at two thirds width.

Screenshot 2024-11-26 at 11 10 50


AshGDS (commented):
Looks good to me, happy to approve, but had a couple questions first:

1. Could you add long heading text to one of the examples, so the two-thirds visual styles are surfaced in the component guide?
2. The WebAIM WAVE extension complains about an accessibility issue on each chart. It seems that an empty table header `<th>` element is being rendered on each chart. Is that fixable in this PR or do you want to look at that separately?
image

Thanks

andysellick:
Thanks @AshGDS. I actually have another feature to add so will look into that issue as well.

andysellick:
@AshGDS actually I think that might be a bigger problem and I don't want it to block these changes. Are you okay to review without? (I've done the heading example)

AshGDS (approved):
👍

PR #4438, opened 2024-11-26
Allow licence details in footer to be hidden

pezholio:
The layout footer component is used quite a lot by internal services, where it doesn’t make sense to show licence details. This adds a `show_licence` option to allow users to hide the licence if they want/need to. This will always default to `true`, unless the option is specifically set to `false`.

### DIFF
--- a/app/views/govuk_publishing_components/components/_layout_footer.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_footer.html.erb
@@ -4,6 +4,7 @@
   meta ||= []
   navigation ||= []
   with_border ||= false
+  hide_licence ||= false
   layout_footer_helper = GovukPublishingComponents::Presenters::LayoutFooterHelper.new(navigation, meta)
   absolute_links_helper = GovukPublishingComponents::Presenters::AbsoluteLinksHelper.new()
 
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_layout_footer.html.erb
```diff
@@ -4,6 +4,7 @@
   meta ||= []
   navigation ||= []
   with_border ||= false
+  show_licence = show_licence.to_s.strip == 'false' ? false : true
```
andysellick, line 7: Might be easier to approach this using `hide_licence ||= false` - I think it might make the logic easier to follow. Then your change on line 99 would be `<% unless hide_licence %>` etc.

pezholio: That makes more sense! Have sorted 👍

andysellick (approved):
Looks good. Does the PR description need updating now the approach has changed?

andysellick:
@pezholio just to check - it looks like this new option hides the licence information, but also all of the links? Is that correct?

andysellick:
Scratch that last comment - it's just that the example doesn't include those links. Ignore me!

PR #4440, opened 2024-11-26
Allow the devolved nations component to work properly on Welsh pages

catalinailie:
## What
Allow the devolved administration component to work properly on Welsh pages.

## Why
Trello card: https://trello.com/c/EFcpWfY2/3071-allow-devolved-administration-component-to-support-welsh-m

### Before
Screenshot 2024-11-27 at 17 06 03

### After
Screenshot 2024-12-02 at 16 09 45

### DIFF
--- a/config/locales/cy.yml
+++ b/config/locales/cy.yml
@@ -70,9 +70,9 @@ cy:
     devolved_nations:
       applies_to: Yn berthnasol i
       connectors:
-        last_word:
-        two_words:
-      england:
+        last_word: " a "
+        two_words: " a "
+      england: Loegr
       northern_ireland:
       scotland:
       type:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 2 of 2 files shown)


config/locales/cy.yml
```diff
@@ -70,9 +70,9 @@ cy:
     devolved_nations:
       applies_to: Yn berthnasol i
       connectors:
-        last_word:
-        two_words:
-      england:
```
hannako, line 75: In the trello card, it says:


> Translation confirmed:
> 
> Standalone, 'England and Wales' is 'Lloegr a Gymru' but if it's got 'i' before it as it does in this context it would be 'Loegr a Gymru'. Also there is a letter missing from the beginning - it should be 'yn' not 'n'. So translation is: 'Yn berthnasol i Loegr a Gymru'.


So should this be `Loegr` rather than `Lloegr`?

catalinailie: Oh that's right. I read it the other way around 🤦‍♀️

hannako: Would you mind updating the screenshot? 

Also , it's England and wales in English, but Wales and England in Welsh. Is that how it works?

hannako (approved):

PR #4446, opened 2024-11-28
Use component wrapper on notice component

AshGDS:
## What
- Adds the component wrapper helper to the `notice` component.
- **Note:** This component uses an `aria_live` boolean to set `aria-live: polite`. I assume it would be best for this to be changed so that you pass `aria: { live: 'polite' }` - but was not sure if I should do this, as that would make this a breaking change which increases deployment complexity. I guess I'm not sure what our stance is on whether we want things perfectly using the wrapper or not.

## Why

## Visual changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_notice.html.erb
+++ b/app/views/govuk_publishing_components/components/_notice.html.erb
@@ -16,13 +16,16 @@
   heading_level = show_banner_title ? "h3" : "h2"
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("govuk-notification-banner gem-c-notice")
+  component_helper.add_class(shared_helper.get_margin_bottom)
 
-  css_classes = %w[govuk-notification-banner gem-c-notice]
-  css_classes << shared_helper.get_margin_bottom
+  component_helper.add_aria_attribute({ label: "Notice" })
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_notice.html.erb
```diff
@@ -16,13 +16,16 @@
...
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("govuk-notification-banner gem-c-notice")
+  component_helper.add_class(shared_helper.get_margin_bottom)
 
-  css_classes = %w[govuk-notification-banner gem-c-notice]
-  css_classes << shared_helper.get_margin_bottom
+  component_helper.add_aria_attribute({ label: "Notice" })
+  component_helper.add_aria_attribute({ live: "polite"}) if aria_live
```
andysellick, line 24: Nitpick (surprised the linter didn't complain) needs a space after `"polite"`?

andysellick, line 24: Yeah good thought about passing the aria live as an attribute rather than a parameter. Maybe just log an issue about this and we'll circle back to it later? There's probably not too many instances of this to fix.

andysellick (approved):

app/views/govuk_publishing_components/components/_notice.html.erb
AshGDS: Thanks - I don't think we have a linter for `.html.erb`!

PR #4451, opened 2024-11-28
[DO NOT MERGE] Layout super nav header: Add search autocomplete

csutter:
## What
This changes the layout super navigation header to use the new `search_with_autocomplete` component instead of the vanilla `search` one.

## Why
So users can enjoy the same autocompleting experience as on the homepage and search page.

## Visual Changes
image

### DIFF
--- a/spec/component_guide/component_index_spec.rb
+++ b/spec/component_guide/component_index_spec.rb
@@ -71,6 +71,7 @@ describe "Component guide index", :capybara do
 @import 'govuk_publishing_components/components/print-link';
 @import 'govuk_publishing_components/components/related-navigation';
 @import 'govuk_publishing_components/components/search';
+@import 'govuk_publishing_components/components/search-with-autocomplete';
 @import 'govuk_publishing_components/components/skip-link';
 @import 'govuk_publishing_components/components/step-by-step-nav';
 @import 'govuk_publishing_components/components/step-by-step-nav-header';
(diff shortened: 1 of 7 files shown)


csutter:
⚠️ This should be merged when a decision is taken to go fully live with autocomplete.

leenagupte (approved):

spec/component_guide/component_index_spec.rb
```diff
@@ -1,6 +1,10 @@
 require "rails_helper"
 
 describe "Component guide index" do
+  around do |example|
+    ClimateControl.modify(GOVUK_ENABLE_SEARCH_AUTOCOMPLETE: "true") { example.run }
+  end
```
kevindew, line 6: So the reason adding this in didn't make the tests pass is because the suggested js / css is looking for strings in the code that match the [full component name](https://github.com/alphagov/govuk_publishing_components/blob/219710480ba337d04dee0408f9df8e52ff26e83a/app/controllers/govuk_publishing_components/component_guide_controller.rb#L154). So before a variable was used in the component name it would match `govuk_publishing_components/components/search_with_autocomplete` but it can't match on a variable of `govuk_publishing_components/components/#{search_component}` as the analysis is done on the source code not the product of the source code.

I imagine the way to resolve it most in the spirit of the component gem suggestion functionality would be to do the conditional and have separate component calls. 

E.g:

```
<% search_options = {
  name: "keywords",
  inline_label: false,
  label_size: "m",
  label_text: search_text,
  label_custom_class: "gem-c-layout-super-navigation-header__search-label--large-navbar",
  size: "large",
  margin_bottom: 0,
  disable_corrections: true,
} %>
<% if ENV["GOVUK_ENABLE_SEARCH_AUTOCOMPLETE"] == "true" %>
  <%= render "govuk_publishing_components/components/search_with_autcomplete", search_options.merge({
    source_url: [Plek.new.website_root, "/api/search/autocomplete.json"].join,
    source_key: "suggestions"
  }) %>
<% else %>
  <%= render "govuk_publishing_components/components/search", search_options %>
<% endif %>
<% end %>
```

Though another way of course would be to just remove it from the suggested components list - but that'd be accepting this component couldn't be suggested which feels against the spirit of this functionality.

leenagupte (commented):
The second commit looks ok to me 👍

PR #4452, opened 2024-11-28
Add a `glance_metric` component

pezholio:
This is based on the `glance_metric` component used in [content-data-admin](https://github.com/alphagov/content-data-admin) with some unused options removed. I’ve also tweaked the display of the explicit label, as `aria-label` is not supported on non-semantic elements,

## Screenshot

image

Trello card: https://trello.com/c/eJTfFnFN/700-create-rollup-for-view-page-and-review-page

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_glance-metric.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_glance-metric.scss
@@ -1,0 +1,75 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-glance-metric {
+  border-top: 1px solid $govuk-border-colour;
+  border-bottom: 0;
+  @include govuk-responsive-margin(6, "bottom");
+
+  &__heading {
+    min-height: 2.6em;
+    @include govuk-font($size: 24, $weight: bold);
+
+    @include govuk-responsive-margin(3, "top");
+    @include govuk-responsive-margin(1, "bottom");
+    @include govuk-responsive-padding(3, "top");
+
+    @include govuk-media-query($until: tablet) {
+      min-height: 4em;
+    }
+
+    @include govuk-media-query($until: mobile) {
+      min-height: 0;
+    }
+  }
+
+  &__figure {
+    display: flex;
+    @include govuk-font($size: 36, $weight: bold);
+    @include govuk-responsive-margin(1, "bottom");
+
+    @include govuk-media-query($until: mobile) {
+      @include govuk-font($size: 48, $weight: bold);
+    }
+  }
+
+  &__explicit-label {
+    @include govuk-visually-hidden;
+  }
+
+  &__context {
+    min-height: 5em;
+    border-bottom: 1px solid $govuk-border-colour;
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/_glance_metric.html.erb
```diff
@@ -0,0 +1,29 @@
...
+  add_gem_component_stylesheet("glance-metric")
+
+  name ||= false
+  figure ||= nil
+  context ||= ""
+  measurement_display_label ||= ""
+  measurement_explicit_label ||= ""
+%>
+<% if name %>
+  <div class="gem-c-glance-metric govuk-body">
```
AshGDS, line 11: For future proofing, could you add the component wrapper helper to this component? Otherwise it creates more work for our team.

For example see this commit: https://github.com/alphagov/govuk_publishing_components/pull/4287/commits/e2a09b40bff246066330225226d781a364fd71af (However you'd only need to use `add_class` for yours, as your parent `<div>` only uses CSS classes.)

app/views/govuk_publishing_components/components/docs/glance_metric.yml
```diff
@@ -0,0 +1,23 @@
+name: "Glance metric"
+description: "The glance metric component is an at-a-glance view of data for a content item. The minimum requirements for it to display are a name and figure. "
+part_of_admin_layout: true
+body: |
+
+accessibility_criteria: |
+  - Where an abbreviated label for a metric is used it MUST be accompanied with the explicit label
```
AshGDS, line 7: Can you add in the accessibility acceptance criteria that the element must be under a suitable `<h2>`, as they are using `<h3>` elements, and if the heading structure of the page breaks it becomes an accessibility issue. 

It may be worth tweaking the current criteria you've written to add what the big number component mentions too: https://components.publishing.service.gov.uk/component-guide/big_number#accessibility-acceptance-criteria

app/assets/stylesheets/govuk_publishing_components/components/_glance-metric.scss
```diff
@@ -0,0 +1,74 @@
...
+    @include govuk-responsive-margin(3, "top");
+    @include govuk-responsive-margin(1, "bottom");
+    @include govuk-responsive-padding(3, "top");
+    min-height: 2.6em;
+
+    @include media-down(tablet) {
+      min-height: 4em;
+    }
+
+    @include media-down(mobile) {
```
AshGDS, line 19: I think `media-down` might be deprecated, as it seems to be related to Internet Explorer. Are you able to change these to use our more modern media queries, e.g.

```sass
    // stuff here
  }
    // stuff here
  }
```

AshGDS (commented):
Looks good, added some comments.

app/views/govuk_publishing_components/components/docs/glance_metric.yml
pezholio: Good call, I've made a tweak so the `heading_level` can be customised if necessary

PR #4453, opened 2024-11-28
Apply helper to `fieldset` instead of parent `div`

dnkrj:
## What
- Apply the Component Wrapper Helper to the `fieldset` instead of parent `div`

## Why
- Its introduction in #4420 was causing `id` to be set on both the parent and fieldset.

### DIFF
--- a/app/views/govuk_publishing_components/components/_fieldset.html.erb
+++ b/app/views/govuk_publishing_components/components/_fieldset.html.erb
@@ -5,12 +5,10 @@
 
   text = text || yield
   describedby ||= nil
-  role ||= nil
   heading_level ||= nil
   heading_size = false unless shared_helper.valid_heading_size?(heading_size)
   error_message ||= nil
   error_id ||= nil
-  id ||= nil
 
   if error_message
     error_id = "error-#{SecureRandom.hex(4)}" unless error_id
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/_fieldset.html.erb
```diff
@@ -5,29 +5,28 @@
...
+  css_classes = %w(gem-c-fieldset govuk-form-group)
+  css_classes << "govuk-form-group--error" if error_message
 
   legend_classes = %w(govuk-fieldset__legend)
   legend_classes << "govuk-fieldset__legend--#{heading_size}" if heading_size
 %>
-<%= tag.div(**component_helper.all_attributes) do %>
-  <%= tag.fieldset class: fieldset_classes, aria: { describedby: describedby }, role: role, id: id do %>
+<%= tag.div class: css_classes do %>
+  <%= tag.fieldset(**component_helper.all_attributes) do %>
```
andysellick, line 29: I'm slightly concerned that using the wrapper helper on an inner element opens the door to inconsistency and the risk that things that the wrapper does could be overridden or counteracted by styles on the parent. I think I need to understand this more: why does the fieldset need an id, and if it does need one, could we instead add a new param for `id_fieldset ||= nil` instead? (admittedly a breaking change, but seems like a cleaner solution?)

AshGDS:
Would it be possible to remove the `<div>` entirely, and move the classes it has on to the `<fieldset>`? That would keep the component wrapper on the parent as the `<fieldset>` would then be the parent.

app/views/govuk_publishing_components/components/_fieldset.html.erb
dnkrj: I've opened a PR with the breaking change here:
https://github.com/alphagov/govuk_publishing_components/pull/4454

dnkrj:
> Would it be possible to remove the `<div>` entirely, and move the classes it has on to the `<fieldset>`? That would keep the component wrapper on the parent as the `<fieldset>` would then be the parent.

The parent div was introduced to [help with the error state](https://github.com/alphagov/govuk_publishing_components/pull/1143). I assume it was needed, but I'll see if they can be combined.

dnkrj:
> > Would it be possible to remove the `<div>` entirely, and move the classes it has on to the `<fieldset>`? That would keep the component wrapper on the parent as the `<fieldset>` would then be the parent.
> 
> The parent div was introduced to [help with the error state](https://github.com/alphagov/govuk_publishing_components/pull/1143). I assume it was needed, but I'll see if they can be combined.

I've explored this and there are some awkward interactions between the `govuk-form-group` class and the `govuk-fieldset` class. We would likely need to abandon the Design System styles, so I'm closing this in favour of the [breaking change](https://github.com/alphagov/govuk_publishing_components/pull/4454).

PR #4461, opened 2024-11-29
Add large_font_size option to govspeak

richardTowers:
## What

We want big govspeak for our hero blocks, and although we can do this with CSS, it breaks component isolation, so we're adding a font_size option to the govspeak component instead.

## Why

In rare cases, we want to use govspeak to render rich text, but the design calls for a larger font size. We want to support this.

[Trello card](https://trello.com/c/vxoiwFba/191-make-headings-and-text-in-main-hero-blocks-bigger)

## How

If the large_font_size option is set to true we'll add a gem-c-govspeak--font-size-l modifier class.

This is used to increase the default font sizes one step on the type scale, so 16 -> 19, 19 -> 24 and 27 -> 36.

## Visual Changes

image

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
@@ -15,6 +15,10 @@
 .gem-c-govspeak {
   @include markdown-typography;
 
+  &.gem-c-govspeak--font-size-l {
+    @include markdown-typography($base_font_size: 19, $font_size: 24, $heading_font_size: 36);
+  }
+
   &.gem-c-govspeak--direction-rtl ol,
   &.gem-c-govspeak--direction-rtl ul {
     margin-left: 0;
--- a/app/views/govuk_publishing_components/components/_govspeak.html.erb
+++ b/app/views/govuk_publishing_components/components/_govspeak.html.erb
@@ -5,7 +5,7 @@
   local_assigns[:margin_bottom] ||= 0
   direction_class = "gem-c-govspeak--direction-#{direction}" if local_assigns.include?(:direction)
   disable_youtube_expansions = local_assigns.fetch(:disable_youtube_expansions) if local_assigns.include?(:disable_youtube_expansions)
-  font_size_class = local_assigns.fetch(:font_size, "m").then { |font_size| "gem-c-govspeak--font-size-#{font_size}" }
+  large_font_size = local_assigns.fetch(:large_font_size, false)
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
@@ -14,7 +14,7 @@
   component_helper.add_class("js-disable-youtube") if disable_youtube_expansions
   component_helper.add_class("gem-c-govspeak--inverse") if inverse
   component_helper.add_class(shared_helper.get_margin_bottom)
-  component_helper.add_class(font_size_class)
+  component_helper.add_class("gem-c-govspeak--font-size-l") if large_font_size
   component_helper.add_data_attribute({ module: "govspeak" })
 %>
 <%= tag.div(**component_helper.all_attributes) do %>
--- a/app/views/govuk_publishing_components/components/_govspeak.html.erb
+++ b/app/views/govuk_publishing_components/components/_govspeak.html.erb
@@ -5,6 +5,7 @@
   local_assigns[:margin_bottom] ||= 0
   direction_class = "gem-c-govspeak--direction-#{direction}" if local_assigns.include?(:direction)
   disable_youtube_expansions = local_assigns.fetch(:disable_youtube_expansions) if local_assigns.include?(:disable_youtube_expansions)
+  font_size_class = local_assigns.fetch(:font_size, "m").then { |font_size| "gem-c-govspeak--font-size-#{font_size}" }
 
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
@@ -13,6 +14,7 @@
   component_helper.add_class("js-disable-youtube") if disable_youtube_expansions
   component_helper.add_class("gem-c-govspeak--inverse") if inverse
   component_helper.add_class(shared_helper.get_margin_bottom)
+  component_helper.add_class(font_size_class)
   component_helper.add_data_attribute({ module: "govspeak" })
 %>
 <%= tag.div(**component_helper.all_attributes) do %>
--- a/app/views/govuk_publishing_components/components/docs/govspeak.yml
+++ b/app/views/govuk_publishing_components/components/docs/govspeak.yml
@@ -19,7 +19,7 @@ shared_accessibility_criteria:
   - link
 accessibility_excluded_rules:
   - landmark-complementary-is-top-level # Statistic headlines are generating an aside element which can not be a top level in the examples
-  - different_font_sizes # Example includes headings, which don't conform with the document structure
+  - large_font_size # Example includes headings, which don't conform with the document structure
 uses_component_wrapper_helper: true
 examples:
   basic_content:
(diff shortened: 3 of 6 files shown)


app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -35,6 +35,23 @@ examples:
           <p>This is some body text with <a href="https://example.com">a link</a>.</p>
     context:
       dark_background: true
+  different_font_sizes:
+    description: |
+      Set a different font size for the govspeak. The only valid option is `l`.
+    data:
+      block: |
+        <p>Look how big this govspeak is!?</p>
```
andysellick, line 43: Let's add the headings into this example to see what they look like.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
```diff
@@ -15,6 +15,10 @@
 .gem-c-govspeak {
   @include markdown-typography;
 
+  &.gem-c-govspeak--font-size-l {
+    @include markdown-typography($base_font_size: 19, $font_size: 24, $heading_font_size: 36);
+  }
+
```
richardTowers, line 21: Might be better to default the class to `.gem-c-govspeak--font-size-m` and then

```suggestion
  &.gem-c-govspeak--font-size-m {
    @include markdown-typography;
  }

  &.gem-c-govspeak--font-size-l {
    @include markdown-typography($base_font_size: 19, $font_size: 24, $heading_font_size: 36);
  }
```

... so I'm not including the rules twice in the large-text case

app/views/govuk_publishing_components/components/docs/govspeak.yml
richardTowers: Screenshot in the PR description added ☝🏻

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
richardTowers: Changed my mind about this - I forgot the CSS gets emmitted twice anyway with @include, so it's not like I would be reducing the file size at all.

The current approach relies on specificity, but it's a simple case so I don't think it should cause much confusion.

I'll leave this comment thread open in case there's any discussion though.

app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -911,3 +912,28 @@ examples:
             <p>Deforested area. Credit: Blue Ventures-Garth Cripps</p>
           </figcaption>
         </figure>
+  different_font_sizes:
```
andysellick, line 915: Suggestion: `large_font_sizes`

app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -911,3 +912,28 @@ examples:
...
+        <p>It is:</p>
+        <ul>
+          <li>Bigger than normal govspeak</li>
+          <li>Intended to be used rarely</li>
+          <li>A way to include large rich text when the design calls for it</li>
+        </ul>
+        <div class="call-to-action">
+          <p>It should support all the usual govspeak features, such as calls to action.</p>
+        </div>
+        <h1>Heading 1</h1>
```
andysellick, line 933: I think you can remove `h1` as I don't think we do H1s in page content.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
```diff
@@ -15,6 +15,10 @@
 .gem-c-govspeak {
   @include markdown-typography;
 
+  &.gem-c-govspeak--font-size-l {
```
andysellick, line 18: I would structure this slightly differently (assuming there's no specificity problems) as

```Sass
.gem-c-govspeak {
  @include markdown-typography;
  // etc.
}

&.gem-c-govspeak--font-size-l {
  // etc.
}
```

This means that if we decided to add another size in future we don't have to keep overriding it - `gem-c-govspeak--font-size-l` is a modifier class.

app/views/govuk_publishing_components/components/_govspeak.html.erb
```diff
@@ -5,6 +5,7 @@
   local_assigns[:margin_bottom] ||= 0
   direction_class = "gem-c-govspeak--direction-#{direction}" if local_assigns.include?(:direction)
   disable_youtube_expansions = local_assigns.fetch(:disable_youtube_expansions) if local_assigns.include?(:disable_youtube_expansions)
+  font_size_class = local_assigns.fetch(:font_size, "m").then { |font_size| "gem-c-govspeak--font-size-#{font_size}" }
```
andysellick, line 8: Could this be simplified we modified line 17 to something like

```Ruby
component_helper.add_class("gem-c-govspeak--font-size-l") if use_large_font
```
?

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
andysellick: Yeah I think the behaviour should be that by default the govspeak component is as before, but if passed the right option, it will have the additional `gem-c-govspeak--font-size-l` class added. See my other comment about the class structure.

app/views/govuk_publishing_components/components/_govspeak.html.erb
richardTowers: While we only support `"l"` as an option, yes, but if we decided to support `"s"` or `"xl"` in the future, this wouldn't need updating.

app/views/govuk_publishing_components/components/docs/govspeak.yml
richardTowers: I guess assuming we'll never want "small" govspeak?

If we know we're only ever going to support one other size, maybe a boolean parameter would be better than a string.

i.e.

```
large_font_size: true
# instead of
font_size: l
```

?

andysellick: Exactly, yes.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
andysellick: I'm just suggesting a slightly cleaner class structure. Our approach is to try to avoid nesting where possible, because that makes the tree more complicated and harder to change later (you have to keep overriding stuff).

It would be:

```Sass
.gem-c-govspeak {
  @include markdown-typography;
  // etc.
}

&.gem-c-govspeak--font-size-l {
  @include markdown-typography($base_font_size: 19, $font_size: 24, $heading_font_size: 36);
}
```

That would produce two separate classes, but we treat the second class as a modifier, so that when it's added to an element with the first class, it changes the styles. It's not a huge difference, it just avoids potential difficulties further down the road.

richardTowers: Ah okay, gotcha.

I think we'd be relying on source order to break the specificity tie though - both rules just have single class selectors, so they're exactly as specific as each other. Is that correct? Is using source order something we're okay with?

If we specify both classes in the override, then it's definitely more specific.

richardTowers: Just to be super clear, the markup in both cases is the same:

```
<div class="gem-c-govspeak gem-c-govspeak--font-size-l">...</div>
```

app/views/govuk_publishing_components/components/docs/govspeak.yml
richardTowers: Addressed in [91389d3](https://github.com/alphagov/govuk_publishing_components/pull/4461/commits/91389d3508e8f3a404ecc9fc8ee13178c1daa17e)

app/views/govuk_publishing_components/components/_govspeak.html.erb
richardTowers: Addressed in [b38cc23](https://github.com/alphagov/govuk_publishing_components/pull/4461/commits/b38cc23c274b22c7300f0d0b7f95ae92fdaddf23)

richardTowers:
We've decided against doing this for now, because calling the markdown-typography mixin twice causes quite a bit of extra CSS to be emitted, and it feels ugly.

If we really want big govspeak for the PfC project, we can do it with some specific code in frontend.

PR #4469, opened 2024-12-03
Remove heading helper from search component

andysellick:
## What / why
Removes the heading helper from the search component.

- not used
- also remove it from the auditing tools, because it's only used by one other component (the heading component) and we only need to audit helpers that are used by multiple components, lots of components already have their own individual helpers

## Visual Changes
None.


AshGDS (approved):
LGTM 👍 I'm guessing it's small enough to not require a changelog update?

andysellick:
@AshGDS yeah I reckon it's too small for that, thanks 👍

PR #4470, opened 2024-12-03
[WIP] Move margin_bottom option into component wrapper helper

andysellick:
## What
Moves the code to handle margin bottom on components from the shared helper to the component wrapper helper, and updates components that depend on it.

This PR may not entirely remove the margin bottom code from the shared helper, as this would probably be a breaking change. We'll see how it goes.

## Why
The component wrapper helper seems like the more logical place for the margin bottom option to live, as it (should be) applied exclusively to the wrapping element of a component.

This work will also smooth the path for a forthcoming look at standardising component margins.

## Visual Changes
None.

Trello card: https://trello.com/c/Y0pDWbHw/242-move-some-shared-helper-options-into-component-wrapper


andysellick:
Closing in favour of https://github.com/alphagov/govuk_publishing_components/pull/4494

PR #4480, opened 2024-12-06
Use correct button type for copy to clipboard component

ryanb-gds:
## What
Use the "button" button type for the copy to clipboard button

## Why
The copy to clipboard component is not used as a form submit button, and therefore should have the "button" button type instead of the "submit" button type.

This affects the tracking implementation in Whitehall ([related Trello](https://trello.com/c/mVJGdWsZ))


jon-kirwan (approved):
Nice one. Looks good to me 👍

PR #4484, opened 2024-12-10
Fix broken spec assertions

kevindew:
This PR was prompted after discovering in https://github.com/alphagov/govuk_publishing_components/pull/4451 that the commonly used `assert_no_selector` test assertion in view specs didn't actually assert anything and would pass even if given `assert_no_selector("*")`

The reason this non-assertion was available was because it is using a Capybara assertion (from the [Capybara::DSL](https://github.com/alphagov/govuk_publishing_components/blob/52b5260d4809c31015aee8806c21833e99237f4a/spec/rails_helper.rb#L13)) - however these assertions only have an impact if you have run `visit(url)` and have a `page` variable.

To remedy this I have taken 2 steps: 

1. I have replaced all usages of `assert_no_selector(selector)` to be `assert_select(selector, false)` to confirm when a selector isn't available for a view spec (unfortunately `assert_no_select` isn't available [yet](https://github.com/rails/rails-dom-testing/blob/da0d96bc872e7e9b2656bc17ef3467bdf4bd9d48/CHANGELOG.md#next--unreleased))
2. I have added a `capybara` tag so that only specs that are tagged with capybara will have access to the Capybara DSL - to prevent using assertions that have no impact. Ideally we'd only have these scoped to a type of test (i.e. feature) but the ship has sailed on that for this project.

No changelog as this is only an internal change.

### DIFF
--- a/spec/component_guide/component_audit_spec.rb
+++ b/spec/component_guide/component_audit_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-describe "Component audit page" do
+describe "Component audit page", :capybara do
   it "renders the audit page" do
     visit "/component-guide/audit"
     expect(page).to have_title "Component audit - Component Guide"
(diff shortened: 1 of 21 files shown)


sihugh (approved):
Good spot, thanks @kevindew 🫡

PR #4491, opened 2024-12-11
Add /media/ path to GA4 download link tracking

AshGDS:
## What / Why
- Ensure links starting with `https://www.gov.uk/media/` are tracked as download links
- We either missed tracking these, or these links were recently reintroduced after an issue with attachments. Therefore the 'View Online' links were not being tracked if they used this old type of link.
- Example link: https://www.gov.uk/government/statistical-data-sets/management-information-ofsteds-initial-teacher-education-ite-inspections-outcomes
- https://trello.com/c/SlmI0kKL/847-attachment-navigation-link-clicks-not-firing-on-certain-links-view-online-links
## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


andysellick (approved):
👍

PR #4494, opened 2024-12-11
Add margin_bottom option to component wrapper helper

andysellick:
## What
Adds a `margin_bottom` option to the component wrapper helper, effectively copying it out of the shared_helper.

I wanted to write a new set of classes here for margins to stop using the govuk-frontend margin override classes, because they should only be used where an override is necessary (they use `!important`). However right out the gate it turns out we do need the override classes - just to override the default margin applied to some of the govuk-frontend components (details, for example, where we want a 15px margin and the component comes with 30px by default).

Supercedes https://github.com/alphagov/govuk_publishing_components/pull/4470

## Why
The CW helper seems like the more logical home for this option, rather than the shared helper. Additionally, cleaning up and standardising this will help us make component spacing more consistent later.

I'd remove the shared_helper margin option entirely but other parts of our code are using it, so it seems better to roll out this addition, switch, then remove it from the shared_helper later.

## Visual Changes
Hopefully none.

Trello card: https://trello.com/c/Y0pDWbHw/242-move-some-shared-helper-options-into-component-wrapper

### DIFF
--- a/docs/component-wrapper-helper.md
+++ b/docs/component-wrapper-helper.md
@@ -29,6 +29,7 @@ These options can be passed to any component that uses the component wrapper.
 - `data_attributes` - accepts a hash of data attributes
 - `aria` - accepts a hash of aria attributes
 - `classes` - accepts a space separated string of classes, these should not be used for styling and must be prefixed with `js-`
+- `margin_bottom` - accepts a number from `0` to `9` (`0px` to `60px`) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale) (defaults to no margin)
 - `role` - accepts a space separated string of roles
 - `lang` - accepts a language attribute value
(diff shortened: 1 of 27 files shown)


docs/component-wrapper-helper.md
```diff
@@ -65,6 +66,9 @@ The component wrapper includes several methods to make managing options easier,
   aria_attributes ||= {}
   role ||= nil
 
+  # margin_bottom will be applied by default, use this line to set a different default
```
AshGDS, line 69: Maybe it's worth rewording this a bit, until I read the ruby code I thought this meant that every component using the wrapper will get a margin bottom by default

AshGDS (approved):
LGTM, nice work 👍  Just needs a changelog entry, and I left a non-blocking comment

docs/component-wrapper-helper.md
andysellick: Thanks, have removed. That was a hangover from an earlier attempt at this.

PR #4500, opened 2024-12-13
Add canonical_url value to GA4 page view tracking

AshGDS:
## What / Why

- Adds a `canonical_url` attribute to the GA4 page view tracker.
- https://trello.com/b/rm8pjAAk/ga4-gtm-implementation-ga4-team-user-experience-govuk

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
@@ -66,7 +66,8 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
             search_term: this.getSearchTerm(),
             tool_name: this.getToolName(),
             spelling_suggestion: this.getMetaContent('spelling-suggestion'),
-            discovery_engine_attribution_token: this.getMetaContent('discovery-engine-attribution-token')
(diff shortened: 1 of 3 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-page-views.js
```diff
@@ -79,6 +80,16 @@ window.GOVUK.analyticsGa4.analyticsModules = window.GOVUK.analyticsGa4.analytics
...
     },
 
+    getCanonicalHref: function () {
+      var link = document.querySelector('link[rel=canonical]')
+
+      if (link) {
+        return link.href
+      }
+
+      return undefined
```
andysellick, line 90: I think a JS function will return `undefined` by default, so this line might not be needed? Might be worth having a test to check this behaviour as well.

AshGDS: Thanks - I've replaced it with `this.nullValue` as that's what other functions are using, but can remove that line entirely if you think I should, as `this.nullValue` still just resolves to `undefined`.

andysellick: I think remove it if it's safe to remove it - no need having it if it's not doing something.

AshGDS: Will do - thanks! 👍

PR #4504, opened 2024-12-16
[WIP] Recreate global bar component from static app

matthillco:
## What
Copy the `global-bar` component from the static app as the initial step in removing it from static.
[Trello](https://trello.com/c/WtPkMQFM/386-move-global-bar-component-from-static-into-publishing-components-gem)

## Why
The static app will be retired, but the `global-bar` component will still be needed.


KludgeKML:
Replaced by: https://github.com/alphagov/govuk_publishing_components/pull/4538

PR #4510, opened 2024-12-20
Add context option to heading component

andysellick:
## What
Adds a context option to the heading component, to mimic the behaviour of context on the page title component.

This also removes the 'with mobile top margin' option from the component, as we're trying to move away from margin top on components. Instead, this spacing will be applied to previous elements on a case by case basis, in further pull requests.

- instances of `mobile_top_margin` are [limited to government-frontend](https://github.com/search?q=org%3Aalphagov%20mobile_top_margin&type=code)
- PR to remove those instances: https://github.com/alphagov/government-frontend/pull/3489

## Why
We're removing the margin_top option from the title component (https://github.com/alphagov/govuk_publishing_components/pull/4508) and there are some places where this is still needed. However if the heading component has context, those instances can be replaced with the heading component.

Eventually we're also working towards merging the title component into the heading component, and this is also a step in that direction.

## Visual Changes
None, apart from the new options on the heading component.

Screenshot 2025-01-06 at 13 10 02

Trello card: https://trello.com/c/l4VyD7Nm/395-retire-page-title-component

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Use govuk-spacing for highlight answer govspeak component ([PR #4515](https://github.com/alphagov/govuk_publishing_components/pull/4515))
+* Add context option to heading component ([PR #4510](https://github.com/alphagov/govuk_publishing_components/pull/4510))
 
 ## 46.4.0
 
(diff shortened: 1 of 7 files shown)


AshGDS (approved):
LGTM 👍

PR #4518, opened 2025-01-06
Remove chart component

andysellick:
## What
Removes the chart component.

Code removal based on changes added in https://github.com/alphagov/govuk_publishing_components/pull/4301 and https://github.com/alphagov/govuk_publishing_components/pull/4312

Slightly depends on https://github.com/alphagov/frontend/pull/4562 (which is the only place this component is used) but since none of the instances of this are actually used should still be safe to merge and deploy.

## Why
The component is no longer required. It was experimental and not fully tested with users, so if we have a requirement for charts again in the future it would be better to build a new component based on further research.

## Visual Changes
No more charts...

Screenshot 2025-01-06 at 15 58 31


andysellick:
@maxgds potentially controversial PR, any thoughts?

maxgds:
> @maxgds potentially controversial PR, any thoughts?

Well we need to either retire this chart component or retire the one in Content Data, and one of them is actually in use so... That said, it would be good to shift the improvements to the in-use version, though probably not all the extra options. And it's questionable whether that should be the web group's job.

andysellick:
> Well we need to either retire this chart component or retire the one in Content Data, and one of them is actually in use so... That said, it would be good to shift the improvements to the in-use version, though probably not all the extra options. And it's questionable whether that should be the web group's job.

Agreed. I've posted a message in slack for the platform engineering team about this change and will leave it to them to decide if they want to migrate any of the changes we've made, which will still be visible to them after this PR is done with.

andysellick:
Update - the team aren't doing any improvement work in `content-data-admin`, so don't need any of the changes we've made.

AshGDS (approved):
Code removal looks ok to me 👍

PR #4521, opened 2025-01-06
Use component wrapper on step nav header component

AshGDS:
## What
- Adds the component wrapper helper to the `step nav header` component.

## Why

## Visual changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
+++ b/app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
@@ -10,10 +10,11 @@
   ]
   breadcrumb_presenter = GovukPublishingComponents::Presenters::Breadcrumbs.new(breadcrumbs)
 
-  data = {}
-  data[:module] = "ga4-link-tracker" unless disable_ga4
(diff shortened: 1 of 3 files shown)


app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
```diff
@@ -10,28 +10,25 @@
...
+    component_helper.add_data_attribute({ module: "ga4-link-tracker" })
     ga4_data = {
       event_name: "navigation",
       type: "super breadcrumb",
       index_link: "1",
       index_total: "1",
     }.to_json
   end
-
-  shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
```
andysellick, line 25: This looks fine but I was a bit confused by it being in two separate commits. Maybe worth either merging them or adding something to the commit message to clarify that the shared helper is no longer needed as the margin bottom is now provided by the component wrapper helper?

andysellick (approved):
One minor comment, non-blocking.

app/views/govuk_publishing_components/components/_step_by_step_nav_header.html.erb
AshGDS: Good point - will do, thanks!

PR #4536, opened 2025-01-09
Add type option to component wrapper helper

andysellick:
## What
Adds a 'type' option to the component wrapper helper, so that the HTML attribute `type` can be set on a component.

The button component only needs `submit` and `button` as options, but I've included what I think is most of the possible options for type in the error checking.

## Why
This is needed to support moving the button component to the component wrapper helper, as it requires a type of 'submit' when the component is a button.

Note that this PR https://github.com/alphagov/govuk_publishing_components/pull/4535 must be merged first as 'type' conflicts with an option on the devolved nations component.

## Visual Changes
None.


lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
```diff
@@ -192,6 +199,15 @@ def check_dir_is_valid(dir_attribute)
         end
       end
 
+      def check_type_is_valid(type_attribute)
+        return if type_attribute.nil?
+
+        options = %w[button checkbox file hidden radio reset submit text]
```
AshGDS, line 205: I think more types need to be included - unless you're only including ones we currently use? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#input_types

AshGDS (commented):
Looks good, but just want something clarified before approving.

lib/govuk_publishing_components/presenters/component_wrapper_helper.rb
andysellick: Initially I only included the ones for the button component, since we don't need any others currently, but I added some more after I went through the list based on what we might probably need in the near future.

I think since the component wrapper helper is normally used on the outside element of a component it's not likely to be used for actual inputs, so we probably don't need radio/checkbox/etc. I'm not sure. Shall I just include the ones needed for this change and add any others later? So far the button component is the only component that needs `type`.

AshGDS: That makes sense, I guess my concern was that `aria`, `lang` etc. seem to have all the valid values included - for example we have the latin language code `la` but I suspect we won't have any Latin components on GOVUK! So would things go a bit out of sync if all the others contain the valid ?

andysellick: That's reasonable. I'll add the others.

AshGDS (approved):

PR #4537, opened 2025-01-09
Report cache status to RUM

richardTowers:
## What

Refactor the custom RUM measurement code, and add some new code to report whether or not a navigation was a cache hit.

## Why

It would be great if we could segment our performance statistics into cache hits and cache misses. This would help us quantify how much benefit the CDN is giving us, and also how much of an impact cache miss performance has on users.

If this shows that (a) quite a lot of real user requests are cache misses and (b) cache misses are much slower than cache hits then we'd be more strongly incentivised to improve cache hit rate and origin performance. On the other hand, if overall web performance isn't significantly degraded by cache misses, we might be more willing to change the architecture of origin in ways that might make it's performance less good. It would also be good to measure the impact of any such moves, which this should help us do.

## How

I've split this into two commits - the first refactors the custom RUM data so it's in its own module, which simplifies a lot of the checks it was doing. The second adds the code I actually want. This uses data from the Server-Timing header, which is populated by Fastly using some VCL code I added in https://github.com/alphagov/govuk-fastly/pull/130 (links with more context in that PR).

## Visual Changes

None


sihugh (commented):
Looks OK to me 👍

AshGDS (approved):
LGTM 👍 

Nitpick: Can the last two commits be merged together?

PR #4538, opened 2025-01-10
Add global bar component from static

andysellick:
## What
Copies the code for the global banner from `static`, with modifications. See commit history for details, but modifications include:

- renaming the component
- consolidating the JS together into the module JS and removing the 'init' JS
- removing click events on the banner
- removing the block list URL code (will be handled by the banners gem)
- making version be an option passed to the component, instead of hard coded in the JS
- significant code refactor
- using the component wrapper helper on the component

Note that this will be a breaking change as it changes the name of the component from `global-bar` to `global-banner`. This has been changed everywhere in this repo including in the GA4 code, except for the bit where we pass data to GA4, where we keep it as global-bar because that's what GA4 expects to receive.

Corresponding PR to remove this from static: https://github.com/alphagov/static/pull/3561

## Why
We want to use the global banner in the new banners gem and we're working towards retiring `static`, so this is the logical place for the component to live.

## Visual Changes

Trello card: https://trello.com/c/WtPkMQFM/386-move-global-bar-component-from-static-into-publishing-components-gem


spec/javascripts/components/global-banner-spec.js
```diff
@@ -0,0 +1,200 @@
...
+    it('makes banner visible if view count is less than 3', function () {
+      GOVUK.setCookie('global_banner_seen', JSON.stringify({ count: 1, version: 5 }))
+      globalBanner = new GOVUK.Modules.GlobalBanner(element[0])
+      globalBanner.init()
+
+      expectGlobalBannerToShow()
+      expectGa4AttributeToExist()
+    })
+  })
+})
```
AshGDS, line 200: Unless I can't see it, should there be a test to ensure the count isn't going above 3 and one that ensures the banner stays hidden if the count is equal to or more than 3?

app/assets/javascripts/govuk_publishing_components/components/global-banner.js
```diff
@@ -0,0 +1,68 @@
...
+    this.$module = $module
+    this.GLOBAL_BANNER_SEEN_COOKIE = 'global_banner_seen'
+    this.alwaysOn = this.$module.getAttribute('data-global-banner-permanent') === 'true'
+    this.bannerVersion = parseInt(this.$module.getAttribute('data-banner-version'))
+  }
+
+  GlobalBanner.prototype.init = function () {
+    // if no cookie consent just show the banner
+    // if there is cookie consent...
+    //   if there's no cookie or the banner should always be shown, set the cookie and show the banner
```
AshGDS, line 22: It may be worth directly mentioning the `global_banner_seen` cookie here instead of just `no cookie` / `a cookie`, especially to differentiate this sentence from the previous one about cookie consent. I initially thought `a cookie` was referring to the consent cookie.

AshGDS (commented):
Looks good 👍  Just have a few questions before approving.

app/assets/stylesheets/govuk_publishing_components/components/_global-banner.scss
```diff
@@ -0,0 +1,46 @@
...
+
+    &:visited {
+      color: govuk-colour("black");
+    }
+
+    &:focus {
+      color: govuk-colour("black");
+    }
+  }
+}
```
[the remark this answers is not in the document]
andysellick: Not sure I knew that, thanks, have added 👍

spec/javascripts/components/global-banner-spec.js
```diff
@@ -0,0 +1,200 @@
...
+  }
+
+  beforeEach(function () {
+    window.GOVUK.setConsentCookie({ settings: true })
+    document.cookie = 'global_banner_seen=; Path=/; Expires=Thu, 01 Jan 1970 00:00:01 GMT;'
+  })
+
+  afterEach(function () {
+    window.GOVUK.setConsentCookie({ settings: null })
+    $('#global-banner').remove()
```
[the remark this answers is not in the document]
andysellick: We can use it in tests. Ideally we'd remove it but there's loads of it in other tests so I figured it was okay. I think if I was writing this from scratch I'd avoid using it, but all this is copied from static.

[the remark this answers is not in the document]
AshGDS: Ah ok, no worries 👍

spec/javascripts/components/global-banner-spec.js
andysellick: Thanks, have added some new tests.

app/assets/javascripts/govuk_publishing_components/components/global-banner.js
andysellick: Good point, have reworded, thanks.

spec/components/global_banner_spec.rb
```diff
@@ -0,0 +1,37 @@
...
+  end
+
+  it "renders with a link on the title" do
+    render_component(title: "Look here you", banner_version: 1, title_href: "https://www.gov.uk")
+    assert_select ".gem-c-global-banner a.gem-c-global-banner__title[href='https://www.gov.uk']", text: "Look here you"
+  end
+
+  it "includes GA4 attributes when there is a link" do
+    render_component(title: "Look here you", banner_version: 1, title_href: "https://www.gov.uk")
+    assert_select 'a.gem-c-global-banner__title[data-ga4-link=\'{"event_name":"navigation","type":"global bar","section":"Look here you"}\']'
```
AshGDS, line 24: Nitpick: should there be a test that checks for `data-module="ga4-link-tracker"` as well?

AshGDS (approved):
LGTM 👍 Added one non-blocking comment.

KludgeKML (approved):
This all LGTM, and it's safe to merge because if we need to deploy a banner on static as is we'll have to do some coding work in Static, which could include updating the variable name call in public layout.

spec/components/global_banner_spec.rb
andysellick: Thanks, have added.

PR #4540, opened 2025-01-10
BREAKING Add component wrapper helper to the search component

AshGDS:
## What
- Adds the component wrapper helper to the `search` component.
- Breaking as it:
  - Renames the `id` option to `label_id`
  - Renames `data_attributes` to `button_data_attributes` - though this is not used anywhere on GOV.UK (should it removed? I think it was needed for when Universal Analytics was on GOV.UK.)
- Should only affect three templates/two repos across GOV.UK: 
  - https://github.com/alphagov/signon/blob/2cef467f8c0fc43ac633c05b9b10124a27d0ad86/app/views/users/_users_filter.html.erb#L4
  - https://github.com/alphagov/finder-frontend/blob/cb655bed56409ea6cbe4f196a58665b3b6e0a6fa/app/views/finders/_show_header.html.erb#L13
  - https://github.com/alphagov/finder-frontend/blob/cb655bed56409ea6cbe4f196a58665b3b6e0a6fa/app/views/finders/_facet_collection.html.erb#L9

## Why
As the [trello card](https://trello.com/c/OJ0nOkQr/420-add-component-wrapper-helper-to-layout-super-navigation-component) states:

## Visual changes

A new example has been added.

### DIFF
--- a/app/views/govuk_publishing_components/components/_search.html.erb
+++ b/app/views/govuk_publishing_components/components/_search.html.erb
@@ -7,7 +7,7 @@
   aria_controls ||= nil
   button_text ||= t("components.search_box.search_button")
   correction_value = "off" if local_assigns[:disable_corrections]
-  id ||= "search-main-" + SecureRandom.hex(4)
+  label_id ||= "search-main-" + SecureRandom.hex(4)
   wrap_label_in_a_heading ||= false
   label_margin_bottom ||= nil
   label_size ||= nil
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -19,21 +19,23 @@
...
-  classes = %w[gem-c-search govuk-!-display-none-print]
-  classes << shared_helper.get_margin_top
-  classes << shared_helper.get_margin_bottom if local_assigns[:margin_bottom]
-  classes << "gem-c-search--large" if size == "large"
-  classes << "gem-c-search--large-on-mobile" if size == "large-mobile"
-  classes << "gem-c-search--homepage" if homepage
-  classes << "gem-c-search--no-border" if no_border
   if local_assigns[:on_govuk_blue].eql?(true)
-    classes << "gem-c-search--on-govuk-blue"
+    component_helper.add_class("gem-c-search--on-govuk-blue")
```
andysellick, line 34: Non-blocking: you could probably rewrite this bit as follows, but it might be clearer as-is. I'll leave it to you.

```
component_helper.add_class("gem-c-search--on-govuk-blue") if local_assigns[:on_govuk_blue].eql?(true)
component_helper.add_class("gem-c-search--on-white") unless local_assigns[:on_govuk_blue].eql?(true)
```

app/views/govuk_publishing_components/components/_search.html.erb
```diff
@@ -19,21 +19,23 @@
...
   else
-    classes << "gem-c-search--on-white"
+    component_helper.add_class("gem-c-search--on-white")
   end
-  classes << "gem-c-search--separate-label" if local_assigns.include?(:inline_label) or local_assigns.include?(:label_size)
+  component_helper.add_class("gem-c-search--separate-label") if local_assigns.include?(:inline_label) or local_assigns.include?(:label_size)
```
andysellick, line 38: Nitpick: I'd move this line up to join the rest of the add_class lines, above.

spec/components/search_spec.rb
```diff
@@ -100,7 +100,7 @@ def component_name
   it "applies data attributes when provided" do
     render_component(
       button_text: "Some test text",
-      data_attributes: {
+      button_data_attributes: {
```
andysellick, line 103: If you're certain that these button attributes aren't set anywhere, I think it'd be a good opportunity to remove this option. I'd double check it's not being picked up in any JS anywhere.

andysellick (approved):

PR #4544, opened 2025-01-14
Add 'draggable' attribute to component wrapper helper

andysellick:
## What
Adds the [draggable attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/draggable) to the component wrapper helper.

Also includes a clean up/restructure of the tests, and adding some missing things to the documentation.

## Why
Needed for the button component.

## Visual Changes
None.

### DIFF
--- a/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
+++ b/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
@@ -14,7 +14,9 @@ RSpec.describe GovukPublishingComponents::Presenters::ComponentWrapperHelper do
         open: true,
         hidden: "",
         tabindex: "0",
+        dir: "rtl",
         type: "submit",
+        draggable: "true",
       }
       component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(args)
       expected = {
@@ -30,7 +32,9 @@ RSpec.describe GovukPublishingComponents::Presenters::ComponentWrapperHelper do
         open: true,
         hidden: "",
         tabindex: "0",
+        dir: "rtl",
         type: "submit",
+        draggable: "true",
       }
       expect(component_helper.all_attributes).to eql(expected)
     end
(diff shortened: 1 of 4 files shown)


spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
```diff
@@ -156,10 +170,16 @@
...
-          helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(aria: { label: "something" })
+          helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(aria: { label: "something_valid" })
           helper.add_aria_attribute(invalid_aria)
         }.to raise_error(ArgumentError, error)
       end
+
+      it "can add aria attributes to already passed aria attributes" do
+        helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(aria: { label: "original-label", describedby: "other" })
+        helper.add_aria_attribute({ label: "extra-label", controls: "something" })
+        expect(helper.all_attributes[:aria]).to eql({ label: "original-label extra-label", describedby: "other", controls: "something" })
```
AshGDS, line 181: Possibly a nitpick/complicating the wrapper, but should there be a `set_aria_attribute` function, or a boolean passed to `add_aria_attribute` that allows you to override the existing aria attributes instead of appending to the existing one? Currently this test appends more text to the label when overriding a label entirely seems like the more common use case. For example:

```erb
helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(aria: { label: "Expand menu", })
helper.add_aria_attribute({ label: "Collapse Menu" }) if the_menu_is_open_on_pageload

# In our current implementation, the label becomes "Expand menu Collapse Menu"
```

AshGDS (approved):
LGTM - needs a changelog entry, and I've added a non-blocking comment 👍

spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
andysellick: That's a good question. If it's all right I think I'll look at it in a separate PR, as there might be more options where having both an `add` and a `set` method might be useful.

PR #4551, opened 2025-01-14
Add rel attribute to component wrapper

andysellick:
## What
Adds the [rel](https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel) attribute to the component wrapper helper.

## Why
Needed for the migration of the [button component](https://components.publishing.service.gov.uk/component-guide/button) onto the component wrapper.

## Visual Changes
None.

### DIFF
--- a/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
+++ b/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
@@ -388,6 +388,32 @@ RSpec.describe GovukPublishingComponents::Presenters::ComponentWrapperHelper do
       end
     end
 
+    describe "rel" do
+      it "does not accept an invalid rel value" do
+        error = "rel attribute (hubris) is not recognised"
+        expect {
+          GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(rel: "hubris")
+        }.to raise_error(ArgumentError, error)
+      end
+
+      it "accepts valid rel value" do
+        component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(rel: "nofollow")
+        expect(component_helper.all_attributes[:rel]).to eql("nofollow")
+      end
+
+      it "can set a rel, overriding a passed value" do
+        helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(rel: "nofollow")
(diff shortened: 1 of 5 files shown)


spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
```diff
@@ -388,6 +388,32 @@
...
+      it "accepts valid rel value" do
+        component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(rel: "nofollow")
+        expect(component_helper.all_attributes[:rel]).to eql("nofollow")
+      end
+
+      it "can set a rel, overriding a passed value" do
+        helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(rel: "nofollow")
+        helper.set_rel("noreferrer")
+        expect(helper.all_attributes[:rel]).to eql("noreferrer")
+      end
```
AshGDS, line 408: Non-blocking comment - Just wondering about the semantics of `set_rel`/`add_rel`. I assume best practice would be to ensure `set_rel` is only used when paired with a boolean flag (e.g. `helper.set_rel() if something == true`? Otherwise, if we freely used `set_rel` on a component it would always override what the developer has passed through as `rel` in a render statement. And do you think there would ever be a situation where the developer needs to distinguish whether they want a `rel` value they've passed through to be an `add_rel` or a `set_rel`? How would we handle that?

AshGDS (approved):
LGTM - thanks for adding a `set` function as well. Approved with a non-blocking comment 👍

spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
andysellick: I can see two use cases for the `set_(thing)` method (and you're right, I'll look at expanding that for the other attributes where appropriate) - one is the one you said, when paired with a boolean flag, but the other is for a component where we don't want something to be overridden, so it should always be set to what is in the template no matter what is passed.

I'll have a think about the best way of documenting this, might be something to add to this PR.

AshGDS: Thanks 👍

AshGDS (approved):
LGTM 👍

PR #4552, opened 2025-01-15
Fix component wrapper rel option

andysellick:
## What
Fixes a bug in how the component wrapper handles the `rel` option.

- code previously assumed that only one rel option would be passed, but you can have more than one in the attribute
- passing more than one previously caused an error even if all of the rels were valid, this change corrects that

## Why
Because I didn't think of this when I was writing it yesterday 🤦 

## Visual Changes
None.


andysellick:
@AshGDS pause on review of this - just realised I can improve the test coverage.

andysellick:
@AshGDS okay for review now whenever you're free 👍

AshGDS (approved):
LGTM - Good spot 👍

PR #4568, opened 2025-01-17
[DO NOT MERGE] Add name and value attributes to component wrapper helper

andysellick:
## What
Adds `name` and `value` attributes to the component wrapper helper.

## Why
Needed for adding the wrapper to the button component.

## Visual Changes
None.


AshGDS (commented):
Haven't looked at the code properly yet, but just wanted to flag that some components such as `date_input`, `glance_metric`, password use `name` for child components. Therefore the name value is going to be rendered on the component wrapper element as well as the child element.

Similarly, components such as `file_upload`, `input` search and texture use "value" for child elements so the component wrapper element could end up having things like `value=email@example.com`.

I believe the majority of these are form components, so declaring this a breaking change and trying to fix those might be an enormous task.

Instead, maybe there needs to be something that can be passed to components so that the component wrapper ignores them? For example

```erb
      label: {
        text: "Title",
        bold: true,
      },
      name: "title_filter",
      id: "title_filter",
      value: @presenter.title,
      component_wrapper_ignore: { value: true }
    } %>
```

However this would still mean you'd need to modify component renders wherever `value` and `name` are being used on child elements.

andysellick:
That's a good point @AshGDS, thanks for flagging. I'll go back to the original PR and have a rethink.

PR #4570, opened 2025-01-20
GA4 search tracker: Normalise initial value

csutter:
## What
In the search tracker, we do not want to fire the search event if the keyword(s) are unchanged.

The way we do this is by comparing the current value of the search field to its initial value, but there was a subtle bug: we normalised the current value on every check, but not the initial value when it was being set.

This changes the setting of the initial value from using the input's raw value to using the (existing) `searchTerm()` method which includes normalisation.

## Why
Making sure our analytics are more accurate 🙇


csutter:
Converting to draft while I make some amendments

csutter:
Closing in favour of new PR with different approach.

PR #4574, opened 2025-01-20
Add component wrapper helper to textarea

andysellick:
## What
Apply the component wrapper helper to the textarea component.

- breaking change, renames `id` option to `textarea_id` as this is applied to a child element and `id` is applied to the parent by the component wrapper
- removes the shared helper, instead relying on the component wrapper helper to provide the same functionality for `margin_bottom`
- removes a bit of code that sets the margin back to normal if an invalid value is passed, but this outcome is now handled by the component wrapper

To deal with this breaking change, other PRs will need to be raised in applications as follows ahead of deployment.

- collections-publisher https://github.com/alphagov/collections-publisher/pull/2373
- content-publisher https://github.com/alphagov/content-publisher/pull/3295
- govspeak-preview https://github.com/alphagov/govspeak-preview/pull/604
- manuals-publisher https://github.com/alphagov/manuals-publisher/pull/2609
- places-manager uses textarea but doesn't pass `id`
- publisher https://github.com/alphagov/publisher/pull/2504
- release uses textarea but doesn't pass `id`
- search-admin https://github.com/alphagov/search-admin/pull/1342
- transition https://github.com/alphagov/transition/pull/1821
- travel-advice-publisher https://github.com/alphagov/travel-advice-publisher/pull/2121
- whitehall https://github.com/alphagov/whitehall/pull/9833

This change also impacts the character count component, which wraps the textarea component. The character count component accepts an `id` option, which is then passed to the textarea to be the id of the textarea element (not the component id). I've updated the component to explicitly pass this when it exists, because there was a corner case where passing an id as part of the textarea object overwrote the value of id passed directly to character count. Hopefully the tests clarify this behaviour (without causing a breaking change to the character count component).

## Why
The code for `margin_bottom` is moving from the shared helper to the component wrapper helper, so we need to remove any instance of this before the work can be completed. Additionally, having the component wrapper on a component helps reduce code and maintenance effort.

## Visual Changes
None.

Trello card: https://trello.com/c/GQ1p2oSC/438-not-doing-add-component-wrapper-helper-to-form-textarea-component

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/textarea.yml
+++ b/app/views/govuk_publishing_components/components/docs/textarea.yml
@@ -14,6 +14,7 @@ accessibility_criteria: |
   - have correctly associated labels
 
   Labels use the [label component](/component-guide/label).
+uses_component_wrapper_helper: true
 examples:
   default:
     data:
@@ -26,7 +27,7 @@ examples:
       label:
         text: "What is the nature of your medical emergency?"
       name: "emergency-name"
-      id: "emergency-id"
+      textarea_id: "emergency-id"
   with_margin_bottom:
     description: The component accepts a number for margin bottom from `0` to `9` (`0px` to `60px`) using the [GOV.UK Frontend spacing scale](https://design-system.service.gov.uk/styles/spacing/#the-responsive-spacing-scale). It defaults to a margin bottom of 6 (30px).
     data:
@@ -102,7 +103,7 @@ examples:
         <%= component %>
       <% end %>
     data:
-      id: "contextual-guidance-id"
+      textarea_id: "contextual-guidance-id"
       label:
         text: "Title"
         bold: true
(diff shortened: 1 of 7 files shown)


app/views/govuk_publishing_components/components/docs/textarea.yml
```diff
@@ -102,7 +102,7 @@ examples:
         <%= component %>
       <% end %>
     data:
-      id: "contextual-guidance-id"
+      textarea_id: "contextual-guidance-id"
       label:
         text: "Title"
         bold: true
```
AshGDS, line 108: This file needs the `uses_component_wrapper_helper: true` line

AshGDS (commented):
Looks ok to me, but the Percy diff implies something is up. Can you confirm if the Percy diff is what you expect?

andysellick:
@AshGDS finally fixed the visual diff - there was a surprisingly complex thing happening with the textarea and the elements that followed it, but I think I've finally got it 👍

AshGDS (approved):
LGTM 👍

Issue #4564, opened 2025-01-17
Configure how the "Add another" component displays default form fields and "Delete" buttons


davidtrussler:
The current default behaviour of the "Add another" component when it loads is to:
- display empty form fields where no content is provided
- not display a “Delete” button where a single instance of content is provided
  
There is a need to customise this behaviour for these scenarios when it is implemented on the Mainstream publisher app so that:
- no empty fields are displayed where no content is provided
- a "Delete" button is displayed where a single instance of content is provided

PR #4575, opened 2025-01-20
Add configuration to "Add another"

davidtrussler:
## What
[Publishing Components Issue #4564](https://github.com/alphagov/govuk_publishing_components/issues/4564)

This adds some configuration to the ["Add another" component](https://components.publishing.service.gov.uk/component-guide/add_another) so that authors can choose how the component displays by default when it is rendered on a page and no content is provided to it.

## Why
The current default behaviour of the "Add another" component when it loads is to:
- display empty form fields where no content is provided
- not display a “Delete” button where a single instance of content is provided

There is a need to customise this behaviour for these scenarios when it is implemented on the Mainstream publisher app so that:
- no empty fields are displayed where no content is provided
- a "Delete" button is displayed where a single instance of content is provided

## Visual Changes
The screenshots below show the component implemented on Mainstream Publisher as it currently displays and how it is displayed with these changes. 

||Before|After|
|-|-|-|
|On load|Screenshot 2025-01-24 at 09 59 50|Screenshot 2025-01-24 at 09 51 18|
|"Add another link" clicked||Screenshot 2025-01-24 at 09 51 09|

### DIFF
--- a/app/views/govuk_publishing_components/components/_add_another.html.erb
+++ b/app/views/govuk_publishing_components/components/_add_another.html.erb
@@ -4,19 +4,27 @@
   empty ||= ""
   fieldset_legend ||= ""
   add_button_text ||= "Add another"
+  empty_fields ||= false
 %>
 
-<div data-module="add-another" class="gem-c-add-another" data-add-button-text="<%= add_button_text %>" data-fieldset-legend="<%= fieldset_legend %>">
-  <% items.each_with_index do |item, index| %>
-    <%= render "govuk_publishing_components/components/fieldset", {
-      classes: "js-add-another__fieldset",
-      legend_text: "#{fieldset_legend} #{index + 1}",
-      heading_size: "m"
-    } do %>
-      <div class="js-add-another__destroy-checkbox">
-        <%= item[:destroy_checkbox] %>
-      </div>
-      <%= item[:fields] %>
+<%= tag.div class: "gem-c-add-another", data: {
+  module: "add-another",
+  "add-button-text": add_button_text,
+  "fieldset-legend": fieldset_legend,
+  "empty-fields": empty_fields
+} do %>
+  <% unless empty_fields && items.count == 0 %>
+    <% items.each_with_index do |item, index| %>
+      <%= render "govuk_publishing_components/components/fieldset", {
+        classes: "js-add-another__fieldset",
+        legend_text: "#{fieldset_legend} #{index + 1}",
+        heading_size: "m"
+      } do %>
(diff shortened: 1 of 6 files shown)


app/views/govuk_publishing_components/components/_add_another.html.erb
```diff
@@ -4,19 +4,22 @@
...
-    <%= render "govuk_publishing_components/components/fieldset", {
-      classes: "js-add-another__fieldset",
-      legend_text: "#{fieldset_legend} #{index + 1}",
-      heading_size: "m"
-    } do %>
-      <div class="js-add-another__destroy-checkbox">
-        <%= item[:destroy_checkbox] %>
-      </div>
-      <%= item[:fields] %>
+<div data-module="add-another" class="gem-c-add-another" data-add-button-text="<%= add_button_text %>" data-fieldset-legend="<%= fieldset_legend %>" data-empty-fields="<%= empty_fields %>">
```
AshGDS, line 10: I think at this point it may be better to use the built in Rails `tag.div` syntax for adding data attributes e.g. `<%= tag.div class: "classes", data: { fieldset_legend: "etc" } do %>`

spec/components/add_another_spec.rb
```diff
@@ -45,4 +45,16 @@ def default_items
...
+    render_component({ items: default_items })
+
+    assert_select "div[data-module='add-another'][data-empty-fields='false']"
+  end
+
+  it "adds a data-empty-fields attribute of 'true' when `empty_fields` is set to 'true'" do
+    render_component({ items: default_items, empty_fields: true })
+
+    assert_select "div[data-module='add-another'][data-empty-fields='true']"
+  end
```
AshGDS, line 59: Does the behaviour need to be tested e.g. it renders the empty state correctly?

AshGDS (approved):
LGTM, just had a couple questions 👍

spec/components/add_another_spec.rb
davidtrussler: Good point: I've updated these tests a bit to provide duplicate and change them for the "Default" and "Start empty" options, which is probably the clearest way to present what it being tested.

davidtrussler:
> LGTM, just had a couple questions 👍

Thanks @AshGDS - I've made a couple of updates in resonse to your queries if you'd be able to hae another quick look: 
- use the `tag.div` syntax
- update the RSpec test to split it into "Default" and "Start empty" scenarios. That seemed to be the clearest way to express the difference in what is being tested in terms of the expected output.

AshGDS (approved):

PR #4579, opened 2025-01-21
Make "Add another" component styles more specific

davidtrussler:
## What
This change wraps the styles for the "Add another" component in their parent class.

## Why
The change here raises the specificity of the component-specific styles so that they are not over-written by more generic styles on the page on which the component is rendered. This is demonstrated below on an implementation of the component on Mainstream Publisher:
- before this change the `margin-top` value of the "Delete" button is over-written by the value in the "Button" component
- after the change the value for `margin-top` is applied as expected.

## Visual Changes
|Before|After|
|-|-|
|External-links-actual|External-links-expected|


jon-kirwan (approved):
All looks fine to me 👍

PR #4582, opened 2025-01-22
Improve GA4 search tracking

csutter:
## What
This change updates the behaviour of the `Ga4SearchTracker` module as follows:
- update the tracking to _always_ fire a `search` event when the form is submitted
- make the action of the `search` event dynamic
- update the logic to establish whether the user is searching vs filtering by checking which fields the user has interacted with (instead of checking if the keyword has changed), and set the event's action property accordingly

## Why
Now that we have collected a decent amount of tracking data on search across GOV.UK, we've come across a couple of trip hazards with the existing tracking. In particular:
- suppressing the event if the keyword hasn't changed is a bit brittle (and subtly buggy as we've forgotten to normalise the initial keyword in the JS component state)
- Relying on separate tracking in Finder Frontend for filter application has led to duplication and unreliable behaviour (for example, as we only track clicks of the filter button, we do not get events when a user submits the form through other means such as the Enter key or the main search field's submit button)

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
@@ -14,6 +14,9 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       this.section = this.$module.dataset.ga4SearchSection
       this.indexSection = this.$module.dataset.ga4SearchIndexSection
       this.indexSectionCount = this.$module.dataset.ga4SearchIndexSectionCount
+
+      // At the beginning, the user has not yet interacted with any form fields
+      this.triggeredAction = 'unchanged'
     }
 
     init () {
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
```diff
@@ -7,13 +7,16 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   class Ga4SearchTracker {
     constructor ($module) {
       this.$module = $module
-      this.$searchInput = this.$module.querySelector('input[type="search"]')
+      this.$searchInput = $module.querySelector('input[type="search"]')
```
AshGDS, line 10: Should this be `this.$module` for consistency?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
```diff
@@ -35,16 +42,24 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
 
+    setTriggeredAction (event) {
+      if (event.target.type === 'search') {
+        this.triggeredAction = 'search'
+      } else if (this.triggeredAction !== 'search') {
+        // The 'search' action always takes precedence over the 'filter' action, so only set the
+        // action to 'filter' if it is not already 'search'.
+        this.triggeredAction = 'filter'
+      }
+    }
```
AshGDS, line 53: Does there need to be some code to set the `triggeredAction` back to `filter` if the user did modify the search input, but then undoes the change, so in the end they didn't actually change the keyword?

AshGDS (commented):
Looks good to me, just had a couple questions first.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-search-tracker.js
csutter: Good question!

We ended up deciding for the sake of simplicity to treat any change of search or filters as a done deal, even if the user then changes the fields back to their original value (so we're measuring _intent_ rather than what they end up doing). We may decide to be more specific about it in the future though.

PR #4584, opened 2025-01-23
Change id validation on component wrapper

andysellick:
## What
Reduce the severity of the validation checking for `id` attributes passed to the component wrapper helper.

- previous validation was too strict, HTML5 valid characters in id attributes are almost anything
- validation now only complains if an id contains spaces, newlines or `.` characters
- `.` is technically allowed, but can lead to problems in testing as an id of `#id.stillid` could be mistakenly matched to an element with an id of `#id` and a class of `stillid`

## Why
The validation was based on HTML4, but HTML5 is a lot more relaxed: https://html.spec.whatwg.org/multipage/dom.html#global-attributes:the-id-attribute-2

## Visual Changes
None.

Trello card: https://trello.com/c/GQ1p2oSC/438-add-component-wrapper-helper-to-form-textarea-component

### DIFF
--- a/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
+++ b/spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
@@ -118,7 +118,7 @@ RSpec.describe GovukPublishingComponents::Presenters::ComponentWrapperHelper do
 
     describe "setting an id" do
       it "does not accept invalid ids" do
-        ["1dstartingwithnumber", "id with spaces", "idwith.dot", "id\nwithnewline"].each do |id|
+        ["id with spaces", "idwith.dot", "id\nwithnewline"].each do |id|
           expect {
             GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(id:)
           }.to raise_error(ArgumentError, / contain/)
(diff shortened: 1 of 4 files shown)


spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
```diff
@@ -118,16 +118,16 @@
 
     describe "setting an id" do
       it "does not accept invalid ids" do
-        ["1dstartingwithnumber", "id with spaces", "idwith.dot", "id\nwithnewline"].each do |id|
+        ["id with spaces", "idwith.dot", "id\nwithnewline"].each do |id|
           expect {
             GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(id:)
           }.to raise_error(ArgumentError, / contain/)
```
AshGDS, line 124: Just curious, is there a reason this doesn't match the exact error string whereas other tests do?

AshGDS (approved):
LGTM 👍

spec/lib/govuk_publishing_components/components/component_wrapper_helper_spec.rb
andysellick: I think because the error message changes each time, as it outputs the id in the error message.

AshGDS: Ah OK - though it seems the other tests account for that, e.g. the other test is`"Classes (#{classes}) must be prefixed with `js-`"` - but this a non blocking nit pick thing anyway so feel free to ignore it.

andysellick: Oddly I don't recognise this code and it doesn't appear elsewhere in the tests, but always interesting to see different ways of doing the same thing. Thanks.

PR #4586, opened 2025-01-23
Add another: fix problem in `createRemoveButtons` method:

davidtrussler:
## What
This change updates a `querySelectorAll` call to be made on the module rather than the document. 

## Why
As it is this creates multiple "Delete" buttons if there are more than one instance of the component on the page (i.e. one per instance).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add custom padding to inverse header ([PR #4590](https://github.com/alphagov/govuk_publishing_components/pull/4590))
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
LGTM 👍 The component example throws an unrelated JavaScript error, does that need to be investigated?

davidtrussler:
> LGTM 👍 The component example throws an unrelated JavaScript error, does that need to be investigated?

Hello @AshGDS Thanks for the review and yes, good spot, there is an error that should probably be looked at. I'll raise an issue separately.

PR #4588, opened 2025-01-23
Add a `form` attribute to the button component

pezholio:
This allows us to have a button outside a containing form element (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form) This is useful if we want to have two buttons in a form group next to one another inside a form group (for example if one form action submits the form and the other button submits to a form to delete the item that is being edited)

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/button.yml
+++ b/app/views/govuk_publishing_components/components/docs/button.yml
@@ -147,3 +147,12 @@ examples:
     data:
       text: Button
       aria_describedby: with_aria_describedby
+  with_form_attribute:
+    description: |
+      Buttons will usually be contained within their containing form, but sometimes it is desirable to have the button 
+      outside the form with the `form` attribute pointing to the ID of the form. See 
+      [The Mozilla dev docs on a button's `form` attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form)
+      for more information.
+    data:
+      text: With `form` argument
+      form: some_form_id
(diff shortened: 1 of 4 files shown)


app/views/govuk_publishing_components/components/docs/button.yml
```diff
@@ -147,3 +147,8 @@ examples:
     data:
       text: Button
       aria_describedby: with_aria_describedby
+  with_form_id:
```
andysellick, line 150: Suggestion `with_form_attribute` ?

app/views/govuk_publishing_components/components/docs/button.yml
```diff
@@ -147,3 +147,8 @@ examples:
...
+    description: Buttons will usually be contained within their containing form, but sometimes it is desirable to have the button outside the form with the `form` attribute pointing to the ID of the form
```
andysellick, line 151: Nitpick: full stop missing at end of sentence.

andysellick, line 151: Suggestion: I'd not used this attribute before and had to dig a bit to find information, worth adding a link maybe? https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button#form

app/views/govuk_publishing_components/components/docs/button.yml
andysellick: Also there's a bug in the component guide that trims `_id` off the end of these entries, so in the guide this currently reads as `With form`, which is a little unclear.

andysellick (approved):
Some very tiny suggestions, but happy to approve and let you deal with them. Also please remember a changelog entry 👍

PR #4596, opened 2025-01-29
Fix RTL support in metadata component

unoduetre:
## What

It was flagged that the metadata Arabic version requires structural changes. See the details in the document attached to the ticket. This change only relates to structural changes. The translations have already been applied by other tickets.

## Why
[Trello ticket](https://trello.com/c/yWKQ6Jyr/3214-fix-how-arabic-translation-is-structured-on-metadata)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
@@ -59,17 +59,16 @@
     box-sizing: border-box;
     float: left;
     clear: left;
+    padding-left: 0;
     padding-right: govuk-spacing(1);
     margin-top: 0;
   }
 }
 
-.direction-rtl .gem-c-metadata .gem-c-metadata__term,
-.gem-c-metadata.direction-rtl .gem-c-metadata__term {
-  float: inline-start;
-  clear: inline-start;
-
+.direction-rtl .gem-c-metadata__term {
   @include govuk-media-query($from: tablet) {
+    float: right;
+    clear: right;
     padding-left: govuk-spacing(1);
     padding-right: 0;
   }
@@ -83,11 +82,6 @@
       margin-bottom: govuk-spacing(1);
     }
   }
-}
-
-.direction-rtl .gem-c-metadata .gem-c-metadata__definition,
-.gem-c-metadata.direction-rtl .gem-c-metadata__definition {
-  float: inline-start;
 }
 
 .gem-c-metadata__toggle-wrap {


unoduetre:
@matthillco I know you worked on a similar task previously. Please, let me know if you have any insights, e.g. if I need to watch out for something.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -59,17 +59,16 @@
...
 
-.direction-rtl .gem-c-metadata .gem-c-metadata__term,
-.gem-c-metadata.direction-rtl .gem-c-metadata__term {
-  float: inline-start;
-  clear: inline-start;
-
+.direction-rtl .gem-c-metadata__term {
   @include govuk-media-query($from: tablet) {
+    float: right;
+    clear: right;
```
jon-kirwan, line 71: Thanks, @unoduetre. The changes look great and work fine. I have one suggestion. I noticed that [inline-start](https://developer.mozilla.org/en-US/docs/Web/CSS/float#inline-start) should automatically adjust positioning based on the content direction which would mean there may be no need to manually change the float position.

```
@media (min-width: 40.0625em) {
dt.gem-c-metadata__term {
    float: inline-start;
}

dd.gem-c-metadata__definition {
    /* float: inline-start; */
}
}
```
With this setup, the metadata component content appears to display correctly for both `rtl` and `ltr`.

unoduetre: Yes, but float: inline-start started to be supported in e.g. Chrome 118. This version allows us to support grade C browsers as well. Do you think it makes more sense to not support grade C browsers?

jon-kirwan: Okay, I thought we could continue to use it, since it's already implemented in a few places, perhaps as a progressive enhancement. However, it'd be great to hear @matthillco's thoughts?

matthillco: The RTL work that I undertook did not consider grade C browsers.

Logical properties have been available in evergreen browsers since 2021 and there are diminishing returns in continuing to support older browsers with fallback code. I believe we should be bold in bringing our codebases up to date. 

However, there are some dangers with this approach. RTL styles implemented with logical properties have the potential to cause a page to be unreadable for some languages in unsupported browsers. So if in doubt, we can provide the fallback physical properties as well. Of course, if we're providing the fallback then there's an argument not to use the newer logical properties at all as it's just adding unnecessary code.

This is my personal view and may not align with the majority view in light of our browser support policies. If in doubt, you might want to refer this to one of the lead front-end devs for a different perspective.

maxgds: The crux of it lies in this point "RTL styles implemented with logical properties have the potential to cause a page to be unreadable for some languages in unsupported browsers" - we need to ensure that we build for everyone. Our unsupported browsers may not give a good experience, but they shouldn't give an unreadable or unusable one. Without looking into the full detail of this PR I'd advise doing some testing to see the extent of the impact and making a call based on that. It may be necessary to avoid the newest methods for now.

unoduetre: @maxgds Below I've added screenshots for comparison between a grade C browser (older version of Chrome) and a modern browser (newer version of Chrome). I think the older version is acceptable, but let me know if you don't consider it as such.

### Older

Screenshot 2025-01-31 at 09-28-07 Dashboard

### Newer
 
Screenshot 2025-01-31 at 09-27-16 Dashboard

maxgds: Thanks that makes it clear. I think the newer option looks acceptable. The words are all there and in the right order and properly aligned to the same side, it's just slightly undesirable wrapping which I think is fine in this instance as there's no real chance of introducing confusion between differences like
`publication date: 5th May 2022`
and
```
publication date:
5th May 2022
```

unoduetre: @maxgds Older/newer mean the browser version, but based on what you said, I understand the way it looks like in older browsers is acceptable.

@jon-kirwan I've updated the code to use `inline-start`.

app/assets/stylesheets/govuk_publishing_components/components/_metadata.scss
```diff
@@ -57,24 +57,14 @@
 
   @include govuk-media-query($from: tablet) {
     box-sizing: border-box;
-    float: left;
-    clear: left;
-    padding-right: govuk-spacing(1);
+    float: inline-start;
+    clear: inline-start;
+    padding-inline-start: 0;
```
jon-kirwan, line 62: Thanks @unoduetre. Just one final comment - are `clear: inline-start;` and `padding-inline-start: 0;` needed? I tested in a few browsers, and disabling them didn’t seem to have much effect. It might be we can remove both.

unoduetre: `clear` has been removed.

jon-kirwan (approved):
Thanks for the changes. All looks good to me

PR #4597, opened 2025-01-30
Update to LUX 4.0.29

andysellick:
## What
Update to LUX 4.0.29, following the instructions here: https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#how-to-update-to-a-new-version

I've tested this using the LUX debugging tool but strangely it reports a version number of 4.0.28 (the previous version) so not sure what's going on there.

## Why
We need to update to the latest version or this code will stop working.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


pezholio (approved):
Happy to approve this as it seems to be just a dependency update

AshGDS (approved):

PR #4601, opened 2025-01-31
[DO NOT MERGE] Pass attachment name to GA4 details component tracking

AshGDS:
## What/Why

- On pages with attachments, all of the `Request an accessible format` details elements have the same GA4 `title`/`section` - they all pass  `Request an accessible format` as their `title`/`section` when clicked. 
- This means PAs can't tell which details component have been clicked, as they're all passing the same data.
- For example, see the attachments on https://www.gov.uk/guidance/prove-your-english-language-abilities-with-a-secure-english-language-test-selt 
- Therefore this PR makes the `section` values the name of the attachment they are sitting under.

image

## Visual Changes

None.

### DIFF
--- a/app/views/govuk_publishing_components/components/_attachment.html.erb
+++ b/app/views/govuk_publishing_components/components/_attachment.html.erb
@@ -8,6 +8,7 @@
   attributes = []
   url_data_attributes ||= {}
   details_ga4_attributes ||= {}
+  details_ga4_attributes[:section] = attachment.title unless details_ga4_attributes[:section]
   shared_helper = GovukPublishingComponents::Presenters::SharedHelper.new(local_assigns)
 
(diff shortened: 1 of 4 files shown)


AshGDS:
@andysellick Yeah, `govspeak` renders the attachment component, though the attachment component is what will be passing the title to the details component. The details component is rendered inside of the attachment component, so I think this should be OK as we don't need to pass anything ourselves, the attachment component does it for us. 

It's difficult to test locally because in order to update the HTML on all these pages, we have to republish all our documents with attachments, so that govspeak regenerates the attachment component HTML.

See https://github.com/alphagov/govspeak/blob/df4d00a7e8d073a31677acf4e88093e157819b4d/lib/govspeak/post_processor.rb#L69

However, I've been asked to hold off on this change for now, so there's less need to review this.

AshGDS:
Closing as this branch is quite old and is likely not needed anymore.

PR #4606, opened 2025-02-04
Remove `data-govuk-chckboxes-init` attribute

davidtrussler:
## What

This removes some errant code from the "Add another" component

## Why

The inclusion of this code was causing the "Checkbox" component to be initialised twice when the "Add another" component was loaded. While this didn't cause any problems with the component's functioning it was displaying an error in the browser console. Removing this allows the "Checkbox" component to initialise as intended without any errors. 

|Description|Screenshot|
|-|-|
|Error thrown by the component on load|Screenshot 2025-02-04 at 13 27 51|

## Visual Changes

There are no visual changes to the component as a result of this change.


MartinJJones (commented):
Nice work, the changes look good to me overall 👍 

The changelog entry looks like it has a merge conflict which a rebase would resolve, also there is a small typo in `-chckboxes-`.

In this case though, I think we can simply remove the changelog entry since it is a small fix to the developer docs we usually don't include these in the changelog.

davidtrussler:
> The changelog entry looks like it has a merge conflict which a rebase would resolve, also there is a small typo in `-chckboxes-`.
> 
> In this case though, I think we can simply remove the changelog entry since it is a small fix to the developer docs we usually don't include these in the changelog.

Ah right, thanks Martin, I've removed that from the CHANGELOG. (The typo was already there BTW!)
And thanks for suggesting the change to fix the problem as well. 👍

MartinJJones (approved):
Nice work, looks good to me 👍

PR #4608, opened 2025-02-04
Remove margin bottom functionality from the shared helper

AshGDS:
## What / Why
- Removes `shared_helper.get_margin_bottom` from the gem
- As the `title` component still used`shared_helper.get_margin_bottom`, I added some specific code for `title` so that the change isn't breaking - the `title` component is being retired soon too.
- Updates the shared helper docs too. I was going to mark the shared helper as deprecated in the docs, but wasn't sure if that was true.
- Trello card https://trello.com/c/hnfXA4lu/471-move-marginbottom-option-into-component-wrapper

## Visual Changes

Should be none.


andysellick:
@AshGDS we'll keep the shared helper as it still contains some needed stuff, it shouldn't be deprecated.

PR #4610, opened 2025-02-05
Make padding around component previews consistent

AshGDS:
## What / Why

- Make the padding around component previews `30px`
- Previously the padding was `30px` for the sides/bottom, and `45px` for the top
- This made testing spacing of components more confusing as the top always had more spacing around it.
- It looks like the previous padding was added 7.5 years ago to make it the same as GOVUK Elements, but we want the padding to be the same around the whole component to make the spacing of components easier to test. See trello card for sources: https://trello.com/c/pvbZKJJX/473-investigate-component-guide-example-vertical-spacing
- See https://components-gem-pr-4610.herokuapp.com/ for visual testing

## Visual Changes
None, surprisingly - I thought Percy would complain!


andysellick (approved):
Nice. Good PR description 👍

Issue #4072, opened 2024-06-21
Govspeak missing spacing between heading and attachment component


andysellick:
On [this corporate information page](https://www.gov.uk/government/organisations/government-property-agency/about/about-our-services#our-service-catalogue) there's a chunk of govspeak containing a heading followed by an attachment component near the bottom. There's no margin bottom on the heading or margin top on the component, so they're uncomfortably close.

![Screenshot 2024-06-21 at 12 18 12](https://github.com/alphagov/govuk_publishing_components/assets/861310/307320fc-2d8f-40f2-a2ee-60586fb230c9)

I thought this might be a messy one but it looks like changing the `.govspeak h2, .gem-c-govspeak h2` class to have a `margin-bottom` of something will fix this issue without impacting the normal spacing between headings and paragraphs that follow, but I've only had a cursory glance so some more in-depth investigation is probably needed.

Thanks to @selfthinker for reporting 👍

andysellick:
Unfortunately the page linked to above seems to have changed, in that it doesn't contain an attachment anymore. I can reproduce this issue locally so have implemented a fix, but not sure where else this might be occurring.

PR #4618, opened 2025-02-06
Fix spacing between govspeak headings and attachment

andysellick:
## What / why
Add a margin bottom to govspeak `H2` elements.

- headings don't have margin bottom in govspeak, so rely on elements following them to have top margin for spacing
- when followed by an attachment, this is a problem because an attachment is a gem component and therefore doesn't have top margin
- solution is to add bottom margin to headings that does not exceed the top margin of other elements, so if both are present they will overlap without a visual difference
- but for attachments the space will be provided by this new bottom margin
- only doing this for H2 for now as this is the only example of this problem in the wild and I'm hesitant to make sweeping changes to govspeak

## Visual Changes
Before | After
------ | ------
Screenshot 2025-02-06 at 13 13 29 | Screenshot 2025-02-06 at 13 13 45

Fixes https://github.com/alphagov/govuk_publishing_components/issues/4072

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix spacing between govspeak headings and attachment ([PR #4618](https://github.com/alphagov/govuk_publishing_components/pull/4618))
+
 ## 51.2.0
 
 * Remove classes functionality from shared helper ([PR #4605](https://github.com/alphagov/govuk_publishing_components/pull/4605))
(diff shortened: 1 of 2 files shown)


AshGDS (approved):
LGTM - I noticed on the Issue that you mentioned you couldn't find an example page anymore, so I did a weird Google Search and managed to find one:

https://www.gov.uk/guidance/disposal-system-specifications

(The Google search was `site:www.gov.uk "PDF" "Contents" "Corporate information"` to try and find a page with similar content to the old page you had)

andysellick:
@AshGDS thanks, great find.

Are you happy with the visual diff? I'm not sure what's causing it, but it seems minor.

AshGDS:
@andysellick Looks OK to me 👍

Issue #4613, opened 2025-02-05
Govspeak CSS is a bit unknown


andysellick:
Govspeak's CSS seems to be quite a large and complex thing that I personally don't have a lot of knowledge or experience of. I have a few concerns about this:

- we have documentation about [govspeak](https://github.com/alphagov/govspeak) itself and the component guide pages for the component but nothing much beyond that
- the Sass files here for the govspeak styles are sometimes long and complicated
- the Sass is heavily nested in places (we try to use BEM where we can for clarity)
- the CSS doesn't appear to follow many of our conventions e.g. paragraphs and headings often use margin top and bottom, in contrast to styles from `govuk-frontend` where headings and `govuk-body` for paragraphs only use margin bottom
- there's some styling around elements based on their position e.g. `last-child`, `p + p` but so many elements and therefore combinations of elements that it's hard to know what impact changes will have

Making changes to govspeak styles therefore feels quite brittle and scary and I worry that we don't make changes because of some of the above. Here are some things we could look at (just ideas off the top of my head, we'd probably want to refine this list, I'm starting the conversation).

- `markdown-typography` is used mainly in govspeak but is in the `helpers/` directory and therefore not obviously part of govspeak. I think it's separated out because it's also used in the component guide's stylesheet, to style chunks of text such as the accessibility criteria. We might be able to do away with this by simply wrapping those blocks of text in a govspeak component.
- Look at ways of simplifying/de-nesting the Sass.
- Explore moving away from margin-top for common typography elements.
- Instead of having CSS for elements already styled in `govuk-frontend` like headings, could we use e.g. `@extend %govuk-heading-m;`?

andysellick:
Noting that some of the issues above have now been (at least partially) addressed:

- https://github.com/alphagov/govuk_publishing_components/pull/4623
- https://github.com/alphagov/govuk_publishing_components/pull/4628
- https://github.com/alphagov/govuk_publishing_components/pull/4659

andysellick:
Worth noting that there is this RFC, proposing that govspeak is rendered in applications, rather than being pre-rendered in whitehall: https://github.com/alphagov/govuk-rfcs/pull/194

If this happened it would allow us much more control over govspeak and its styles, which would address many of the issues above.

PR #4619, opened 2025-02-06
Change how markdown typography stylesheet is used

andysellick:
## What
`_markdown-typography.scss` is a stylesheet used primarily by govspeak. It isn't inside the govspeak stylesheets directory, probably because it is also imported into the component guide styles to provide styling on formatted text blocks (such as the component descriptions and accessibility criteria).

This PR replaces formatted text block styling by wrapping that content in a govspeak component instead of importing `markdown-typography` into the component guide stylesheet. It also moves the `markdown-typography` file into the govspeak directory, to make its purpose clear.

## Why
Trying to simplify/clarify how govspeak fits together, relating to https://github.com/alphagov/govuk_publishing_components/issues/4613

## Visual Changes
Shouldn't be any visual changes anywhere, although will need a manual check as none of the bits of the component guide affected by this change are included in the Percy visual diff check.


kevindew:
I think I might have been involved in the adding of the markdown-typography file so I can probably add some context that may be useful if you go further with this and some potential areas of confusion you could fall into following the path you're on (not that I have strong opinions as to exactly what is best).

An interesting quirk we have is that because the govspeak gem depends on govuk_publishing_components, govuk_publishing_components cannot depend on govspeak, as that would produce a cyclic dependency. 

So the use of the `.component-markdown` and reduced styling was intended as a distinction between what is full on govspeak and what is plain markdown.

I think at the time of setting bits up there were discussions that in future it'd be good to have a component, similar to the govspeak one that was just for rendering plain old markdown - particularly as we were using that a fair bit in publishing apps and didn't want the extra JS features, however that seems to have never transpired. 

So I think bringing them back together does risk some confusion that what we're dealing with is not govspeak but markdown, however can understand wanting the component. I do feel a more general markdown component and keeping the bits separate is probably still the best route (be also cool if it could accept either html or markdown input).

kevindew:
> @kevindew thanks that's really helpful. So if the gem uses the govspeak component to render content in the component guide, that's the cyclic dependency? Even though the guide is using the component, which is itself part of the gem?

It can use the govspeak component - as that doesn't actually render the govspeak, more wraps the resulting HTML. 

However the places where this gem is generating HTML from markdown, it is not using govspeak but markdown directly: https://github.com/alphagov/govuk_publishing_components/blob/bf823a8a5cf6a85bef811d748e343f4e00640383/app/models/govuk_publishing_components/component_doc.rb#L100-L102

So the changes you've made should work without an issue, but if there's any expectation of using govspeak specific functionality that won't work.

andysellick:
@kevindew I think I still don't understand this well enough to make a sensible decision on it so I'm going to close this PR for now. Thanks for your input, much appreciated 👍

kevindew:
@andysellick no problem, let me know anytime if you'd like to chat about it.

Issue #1995, opened 2021-03-23
Deprecate `gem-c-heading--font-size-27` in heading component


injms:
The gem now mirror GOV.UK Frontend's t-shirt sizing for the heading sizes - `s`, `m`, `l`, and `xl` - there is a stray rule left in as a remnant of the gem's previous naming convention for sizes. This could be removed in the next planned breaking change release.

The Sass that could be removed:

https://github.com/alphagov/govuk_publishing_components/blob/909bf77ca8323432876b99c0c87cf541ec287eeb/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss#L5-L8

And the `HeadingHelper` would also need to be updated to set a new default heading size:
https://github.com/alphagov/govuk_publishing_components/blob/909bf77ca8323432876b99c0c87cf541ec287eeb/lib/govuk_publishing_components/presenters/heading_helper.rb#L19-L35

PR #4620, opened 2025-02-06
Make heading component default size match govuk-frontend

andysellick:
## What
Change the default font size of the heading component.

- the heading component has options to control the font size (independently of the heading level) and all the options directly match the govuk-frontend heading sizes: xl, l, m and s
- however if no size option is passed to the component it defaults to a font size of 27px, which is between l (36px) and m (24px)
- this could lead to a confusion in the visual hierarchy of headings on a page
- removing this custom code and allowing the component to default to 24px (to match govuk-heading-m)

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/1995 

This issue is marked as a breaking change, but I don't think it will actually cause any errors in our code, only that a fair amount of headings will suddenly appear slightly smaller.

## Visual Changes
Before | After
----- | -----
Screenshot 2025-02-06 at 17 44 42 | Screenshot 2025-02-06 at 17 44 58

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_heading.scss
@@ -5,11 +5,6 @@
   margin: 0;
 }
 
-.gem-c-heading--font-size-27 {
-  @include govuk-text-colour;
-  @include govuk-font(27, $weight: bold);
-}
-
 .gem-c-heading--padding {
   padding: govuk-spacing(3) 0;
 }
--- a/lib/govuk_publishing_components/presenters/heading_helper.rb
+++ b/lib/govuk_publishing_components/presenters/heading_helper.rb
@@ -17,7 +17,6 @@ module GovukPublishingComponents
     private
 
       def heading_size(option)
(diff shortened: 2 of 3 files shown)


andysellick:
I've discovered a bit more about this. Yes, we do deviate from the Design System, but apart from (the slightly odd case) where a heading component can be told to be 24px instead of 27px, we seem to be internally consistent across www.gov.uk.

One possible reason why our H2s are a bit bigger is because we have a [lead paragraph component](https://components.publishing.service.gov.uk/component-guide/lead_paragraph) that often appears with a H2 that is also 24px in size. It's likely that the increased size of our H2 is to help distinguish the two.

Here's an example: https://www.gov.uk/find-licences/tv-licence

So really the choice here is whether we should force ourselves to use the slightly smaller heading size and potentially confuse users with the lead paragraph, or simply update the heading component so that if you pass it the M option it uses 27px instead of 24px (thus making it internally consistent).

AshGDS (approved):
LGTM - maybe check with a designer on the lead paragraph thing? 

And this likely needs a changelog entry.

PR #4621, opened 2025-02-07
Update LUX to v4.0.30

AshGDS:
## What /  Why

- Updates LUX to `4.0.30`
- Also updated the docs, as I think they were out of date (I can't find `dist/lux.js` within their repo so I added build instructions instead.)
-  In the docs I've also added a way to find out what lines need to be updated in the script using a diff check, which might make things a bit simpler. I added this as I initially copied and pasted the whole `lux.js` file over, when in reality only a few lines needed changing and I spent a lot of time trying to fix the formatting when I copied the whole updated file over. 
    - However if you think what I've added is too brittle of a way of updating LUX I can remove it. Though if that's the case I think there may need to be some more guidance on how to fix the formatting, as I didn't know of an automatic way to format the updated LUX file to it matches how it is currently, and that ended up being a bit of a faf for me!
- I have confirmed `LUX.getDebug()` is working on my local static.
- Trello card: https://trello.com/c/KEsQoGTf/470-update-lux-to-version-4030

## Visual Changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Fix spacing between govspeak headings and attachment ([PR #4618](https://github.com/alphagov/govuk_publishing_components/pull/4618))
+* Update LUX to v4.0.30 ([PR #4621](https://github.com/alphagov/govuk_publishing_components/pull/4621))
 
 ## 51.2.0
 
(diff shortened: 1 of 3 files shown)


andysellick (approved):
Nice update, good documentation improvements 👏

Issue #4350, opened 2024-10-30
IE11 specific code in our Feedback component


AshGDS:
Our feedback component contains JavaScript specific to IE11, which isn't supported anymore. Therefore the code check isn't needed.

See the code here: https://github.com/alphagov/govuk_publishing_components/blob/2c1702a2ebbafcd5a82ef3c049c27ec49cc5d595/app/assets/javascripts/govuk_publishing_components/components/feedback.js#L72-L107

Work would involve remove this code and associated tests, as well as removing the IE11 references in the `feedback.yml` description.

PR #4622, opened 2025-02-07
Remove IE11 specific code from feedback.js

AshGDS:
## What / Why
- Removes some if statements that were for IE11 from our JS code
- Closes https://github.com/alphagov/govuk_publishing_components/issues/4350
- When viewing the diff, checking Hide Whitespace in the settings will make it easier to view the changes.

## Visual Changes

None.


jon-kirwan (approved):
Thanks @AshGDS. All looks good to me and thanks for mentioning the "hide whitespace changes" suggestion as well: very handy!

PR #4623, opened 2025-02-07
Change govspeak spacing model

andysellick:
## What
Change the govspeak CSS model from a margin top and bottom approach to a margin bottom only approach.

Trying to keep visual changes to a minimum but some small changes are likely, particularly as this change will include switching to `govuk-spacing()` for many margins instead of using `em` values.

To do:

- check appearance of many page types with changes
- check contents of main component govspeak sass, and govspeak html publication sass
- check with govspeak preview, and see how dependency works
- is govspeak previewed in any publishing interfaces?

For reference, I've checked these changes in the following page types. While there are minor differences I don't think there are any problems.

- news article (e.g. https://www.gov.uk/government/news/10-things-to-love-about-us)
- publication (e.g. https://www.gov.uk/government/publications/hm-land-registry-disclosure-log-april-to-june-2018)
- specialist document (e.g. https://www.gov.uk/residential-property-tribunal-decisions/fearn-island-mills-leeds-ls9-8ee-man-slash-00da-slash-ldc-slash-2021-slash-0064)
- detailed guide (e.g. https://www.gov.uk/guidance/remote-gaming-duty-appointing-a-representative-in-the-uk)
- hmrc manuals section (e.g. https://www.gov.uk/hmrc-internal-manuals/appeals-reviews-and-tribunals-guidance/artg6380)
- case study (e.g. https://www.gov.uk/government/case-studies/driving-economic-growth-and-recovery-in-the-legal-sector)
- about (e.g. https://www.gov.uk/government/organisations/disclosure-and-barring-service/about)
- access and opening (e.g. https://www.gov.uk/government/organisations/companies-house/about/access-and-opening)
- answer (e.g. https://www.gov.uk/personal-tax-account)
- call for evidence (e.g. https://www.gov.uk/government/calls-for-evidence/non-road-mobile-machinery-decarbonisation-options)
- html publications (e.g. https://www.gov.uk/government/publications/car-show-me-tell-me-vehicle-safety-questions/car-show-me-tell-me-vehicle-safety-questions)

## Why
Margin top and bottom relies upon overlapping margins to provide spacing. This is wasteful (why declare two things when you can declare one?), involves lots of overrides for first/last children, and does not follow the Design System approach, or the approach we try to use throughout the components gem.

## Visual Changes
Detailed below. Haven't included screenshots for everything as the preview app is better in many cases.

**Address** - margin top is removed, but otherwise largely the same.

Before | After
------ | ------
Screenshot 2025-02-14 at 15 16 59 | Screenshot 2025-02-14 at 15 17 20

**Attachment link** - no visual changes.

**Attachment** - no visual change, but due to the change in spacing around paragraphs, the gap between attachments and following paragraphs is reduced. Was going to change this but the attachment is using styles from the attachment component (gem-c-attachment) so the change would need to be made there.

Before | After
------ | ------
Screenshot 2025-02-14 at 15 20 41 | Screenshot 2025-02-14 at 15 20 59

**Attachment inline** - no visual changes.

**Whitehall block attachments** - removed padding top, so the spacing between these elements is smaller. I thought about increasing the margin bottom to compensate, but decided to stick with 30px as that seems to be the standard.

Before | After
------ | ------
Screenshot 2025-02-14 at 15 26 03 | Screenshot 2025-02-14 at 15 26 22

**Blockquote** - no visual changes, although I'm not entirely comfortable over how the margin bottom is handled by child elements rather than the blockquote itself. Although it does save having extra CSS for last child, and blockquotes seem like they'd always contain a `p`.

**Buttons** - no visual changes.

**Call to action** - slight visual change as we switch from using `em` to `govuk-spacing()` for the element padding. The other change is using a smaller padding on the bottom, assuming that child paragraphs will provide part of the spacing. Again, a bit of an assumption, but I think a reasonable one.

Before | After
------ | ------
Screenshot 2025-02-14 at 15 37 13 | Screenshot 2025-02-14 at 15 37 26

This change also improves the spacing in some call to action boxes, as seen on https://www.gov.uk/government/news/transport-minister-visits-brighton-to-see-uks-most-accessible-buses-for-international-day-of-people-with-disabilities

Before | After
------ | ------
Screenshot 2025-02-25 at 09 42 35 | Screenshot 2025-02-25 at 09 42 46

**Information callout** - margin top has been removed, plus switching from `em` to `govuk-spacing()` results in a slight visual change.

Before | After
------ | ------
Screenshot 2025-02-14 at 15 58 47 | Screenshot 2025-02-14 at 15 59 16

**Warning callout** - this one was a bit tricky, because it's an element that requires a bit of padding top and also contains elements that themselves could have margins. Since so many possible elements can be inside it I'm letting them keep their own margin, which is actually overlapped by a larger margin on the bottom of the warning callout itself (screenshots don't adequately convey this).

**Contact** - broadly the same, but loses margin top.

**Charts** - no visual changes.

**Example** - no margin top now, plus slight left/right spacing changes due to switching away from `em`.

Before | After
------ | ------
Screenshot 2025-02-14 at 16 05 45 | Screenshot 2025-02-14 at 16 05 59

**Footnotes** - some reduction in spacing, because this is partly made up of paragraphs and the spacing around paragraphs has changed. Again, could have added some more spacing to reduce the visual difference but it's good to have a consistent spacing approach.

Before | After
------ | ------
Screenshot 2025-02-14 at 16 08 09 | Screenshot 2025-02-14 at 16 08 25

**Form download** - moves up slightly as padding top is removed. Padding bottom is also reduced, but margin bottom is added to compensate.

**Highlight answer** - increased margin bottom to be more consistent with other elements (was a bit small).

**Headings** - broadly the same, despite all the changes. Removed margin top but also added a rule so that whenever a heading follows anything else, padding top is added to increase the space. Otherwise the gap between paragraphs and headings is the same as the gap between paragraphs, which I think would make the headings harder to spot (see the common example for clarity). H4s downwards are deliberately unstyled (to discourage their use).

**Image** - padding top is replaced with margin bottom, otherwise unchanged.

**Lists** - removes margin top. The rest is a bit tricky, because it turns out some lists can contain paragraphs, so there needs to be spacing between them, but less than between normal paragraphs. As a compromise the spacing between list items is increased to match the reduced spacing between paragraphs in lists, so although overall list spacing is increased, normal lists and lists containing paragraphs have consistent spacing.

Before | After
------ | ------
Screenshot 2025-02-14 at 16 26 42 | Screenshot 2025-02-14 at 16 26 59

**Govspeak HTML publication**

Changes are minimal and there should be no visual difference.

Trello card: https://trello.com/c/IGbGxOO6/481-standardise-govspeak-element-spacing

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -9,22 +9,13 @@
 
 .govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
 .gem-c-govspeak {
-  // This block is duplicated from Whitehall as a transitional step, see the
-  // commit message for 2d893c10ee3f2cab27162b9aba38b12379a71d07 before making
-  // changes, original version:
-  // https://github.com/alphagov/whitehall/blob/main/app/assets/stylesheets/frontend/helpers/_attachment.scss
   $thumbnail-width: 99px;
 
   .attachment {
     position: relative;
-    margin: govuk-spacing(6) 0;
-    padding: govuk-spacing(3) 0 0 ($thumbnail-width + govuk-spacing(6));
+    margin: 0 0 govuk-spacing(6) 0;
+    padding: 0 0 0 ($thumbnail-width + govuk-spacing(6));
     @include govuk-clearfix;
-
-    &:first-child {
-      margin-top: 0;
-      padding-top: 0;
-    }
 
     @include govuk-media-query($media-type: print) {
       padding-left: 0;
@@ -59,12 +50,12 @@
 
     .attachment-details {
       h2 {
-        margin: 0;
+        margin: 0 0 govuk-spacing(2) 0;
         @include govuk-font($size: 27);
       }
 
       p {
-        margin: govuk-spacing(2) 0;
+        margin: 0 0 govuk-spacing(2) 0;
       }
 
       .metadata {
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
@@ -10,7 +10,7 @@ $highlight-answer-color: govuk-colour("white");
     color: $highlight-answer-color;
     text-align: center;
     padding: govuk-spacing(5) govuk-spacing(2) govuk-spacing(4);
-    margin: 0 0 govuk-spacing(2);
+    margin: 0 0 govuk-spacing(4);
 
     p {
       color: $highlight-answer-color;
@@ -29,7 +29,6 @@ $highlight-answer-color: govuk-colour("white");
     }
 
     @include govuk-media-query($until: tablet) {
-      margin: 0 0 govuk-spacing(2);
       @include govuk-font($size: 48);
 
       p {
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_stat-headline.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_stat-headline.scss
@@ -8,13 +8,7 @@
 .govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
 .gem-c-govspeak {
   .stat-headline {
-    margin-bottom: govuk-spacing(3);
-    margin-top: govuk-spacing(3);
-
-    @include govuk-media-query($from: tablet) {
-      margin-bottom: govuk-spacing(4);
-      margin-top: govuk-spacing(4);
-    }
+    margin: 0 0 govuk-spacing(4) 0;
 
     p {
       margin: 0;
@@ -23,7 +17,6 @@
 
     em {
       display: block;
-      margin: 3px 0 -5px;
       @include govuk-font($size: 80, $weight: bold);
     }
   }
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_warning-callout.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_warning-callout.scss
@@ -14,8 +14,8 @@
     $icon-size: 35px;
     $icon-spacing: 10px;
 
-    margin: 2em 0;
-    padding-left: calc($icon-size + $icon-spacing);
+    padding: govuk-spacing(3) 0 0 calc($icon-size + $icon-spacing);
+    margin: 0 0 govuk-spacing(6) 0;
     position: relative;
 
     &::before {
@@ -26,13 +26,6 @@
 
     p {
       @include govuk-font($size: 19, $weight: bold);
-    }
-  }
-
-  // Prevent excessive spacing when placed in a call-to-action block
-  .call-to-action {
-    .help-notice:first-child {
-      margin-top: 0;
     }
   }
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
@@ -1,22 +1,23 @@
 @mixin markdown-typography {
   @include govuk-text-colour;
-
   @include govuk-font($size: 16);
 
   ol,
   ul,
   p {
-    margin-top: govuk-spacing(3);
-    margin-bottom: govuk-spacing(3);
+    margin: 0 0 govuk-spacing(4) 0;
     @include govuk-font($size: 19);
-
-    @include govuk-media-query($from: tablet) {
-      margin-top: govuk-spacing(4);
-      margin-bottom: govuk-spacing(4);
-    }
   }
 
   // Headings
+
+  // put some spacing between any element followed by a heading
+  // has to be padding on the heading otherwise margins would overlap
+
+  * + h2,
+  * + h3 {
+    padding-top: govuk-spacing(4);
+  }
 
   // Markdown is expected to be styled within a document with a H1 title
   // thus H1's are not expected and are discouraged by bare styling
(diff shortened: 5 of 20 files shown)


nnagewad (commented):
This looks great. 👍🏽

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
AshGDS, line 1: Should attachment spacing should be looked at prior to the merge? On mobile especially it's a bit too tight https://components-gem-pr-4623.herokuapp.com/component-guide/govspeak/attachment 
image

Separately - I'm wondering if we should raise an issue for fixing the attachment thumbnails, I think they look too big on mobile or should become `inline-block` elements.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
AshGDS, line 1: Non-blocking: This component drops to 10px of margin bottom on mobile/tablet, is this OK? https://components-gem-pr-4623.herokuapp.com/component-guide/govspeak/highlight_answer

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_stat-headline.scss
```diff
@@ -8,22 +8,10 @@
...
-    margin-top: govuk-spacing(3);
-
-    @include govuk-media-query($from: tablet) {
-      margin-bottom: govuk-spacing(4);
-      margin-top: govuk-spacing(4);
-    }
-
-    p {
-      margin: 0;
-      @include govuk-font($size: 19, $weight: bold);
```
AshGDS, line 21: Text seems to have lost bold weight on mobile 
image
image

app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
AshGDS, line 1: `<blockquote>` doesn't seem to have any margin-bottom and is relying on the `p:last-child`- just wanted to check this is OK? https://components-gem-pr-4623.herokuapp.com/component-guide/govspeak/blockquote

AshGDS (changes requested):
Looks good to me, thanks for providing such as thorough PR description. Added some minor comments to address before approving.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_warning-callout.scss
```diff
@@ -14,8 +14,9 @@
     $icon-size: 35px;
     $icon-spacing: 10px;
 
-    margin: 2em 0;
+    padding: govuk-spacing(3) 0 0 0;
```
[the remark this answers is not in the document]
andysellick: Good shout, have merged the two lines.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
andysellick: That is too close but I think it's been fixed by the merging of https://github.com/alphagov/govuk_publishing_components/pull/4659. However the spacing isn't as wide as it could be, but it's down to the attachment component styles. I'll look into whether the default bottom margin for that component could be increased separately, which should provide more space here.

Agreed on the mobile thumbnails, could be half the size and fine. If I look into the above, do you want to raise an issue for this?

AshGDS: Nice, yeah it looks better now 👍 

Happy to raise an issue for the thumbnail stuff!

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_highlight-answer.scss
andysellick: You're right, it's a bit small. I initially looked at increasing it to 15px, but I've since decided to be bold and just remove this specific line for mobile/tablet, so it defaults to the desktop spacing. This way it has the same space below it as other things (paragraphs etc.) so feels a bit more consistent, and spaces things out a bit more, which I think looks better. Happy to hear your thoughts.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_stat-headline.scss
andysellick: Whoops, think I got a bit over-zealous there (had disappeared on desktop too). Have fixed.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
AshGDS: Done: https://github.com/alphagov/govuk_publishing_components/issues/4662

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_stat-headline.scss
AshGDS: No worries, thanks!

app/assets/stylesheets/govuk_publishing_components/components/helpers/_markdown-typography.scss
andysellick: I might have overlooked this one, will check.

andysellick: Blockquotes are in the typography file, which I don't remember looking at previously. Possibly an oversight, but looking at it now I don't think much needs to change, although I have removed a redundant line (see new commit).

Regarding the blockquotes, I think it's okay in this case? We could put a margin on the blockquote and remove the margin from the last child para (I've seen a few examples of them containing multiple paras) but that would mean more CSS at this stage.

AshGDS: No worries, happy to leave it as is 👍

AshGDS (approved):
LGTM, just needs a changelog entry 👍

PR #4634, opened 2025-02-14
Swap inset text top margin with padding

AshGDS:
## What / Why

- Swap the margin top on this component for padding as we're trying to standardise the top/bottom spacing on components. 
- Trello card: https://trello.com/c/tpFgf7uM/498-investigate-margin-top-on-inset-text-component
- This required nesting the component in a `<div>`, as we can't change the padding directly on the existing component, as the current `padding` here is used for the `border-left` effect where the border is taller than the text content. Therefore, padding had to be added to a new parent component so that the new padding doesn't affect how the component itself looks and ensures it's just used for spacing.
 - You can compare the spacing with the live component: https://components.publishing.service.gov.uk/component-guide/inset_text/preview vs https://components-gem-pr-4634.herokuapp.com/component-guide/inset_text/preview

## Visual Changes

None, hopefully

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_inset-text.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_inset-text.scss
@@ -6,3 +6,15 @@
     break-inside: avoid;
   }
 }
+
+.gem-c-inset-text {
+  padding-top: govuk-spacing(4);
+
+  .govuk-inset-text {
+    margin-top: 0;
+  }
+
+  @include govuk-media-query($from: tablet) {
+    padding-top: govuk-spacing(6);
+  }
+}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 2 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_inset-text.scss
```diff
@@ -6,3 +6,14 @@
     break-inside: avoid;
   }
 }
+
+.govuk-inset-text {
```
JamesCGDS, line 10: Would there be any mileage in nesting this class inside `.gem-c-inset-text`? Just to make sure everything's water tight and we're limiting the scope of this override within a gem specific class.

JamesCGDS (commented):
Looks good 👍 Just have one quick comment/question

app/assets/stylesheets/govuk_publishing_components/components/_inset-text.scss
AshGDS: Good point, have done this now, thanks @JamesCGDS 👍

AshGDS:
Thanks @JamesCGDS - should be ready for another review now.

JamesCGDS (approved):
LGTM 👍

PR #4638, opened 2025-02-18
Allow margin bottom to be configured on the table component

AshGDS:
## What / Why

- Adds a `margin_bottom` option to the `table` component
- This is needed so that we can set the `margin_bottom` to 0 where this component is nested within another component (as parent components should control all of the `margin_bottom`, not child components that render inside them)
- Usually we would add the component wrapper helper to achieve this, as that helper includes `margin_bottom` functionality, but here that isn't easy due to how the component is architected. Therefore, I've added setting margin bottom as it's own option to the component.
- This is a pre requisite to completing this trello card: https://trello.com/c/Q4quM0zo/500-fix-calendar-component-margin
- Test link: https://components-gem-pr-4638.herokuapp.com/component-guide/table/with_custom_margin_bottom

## Visual Changes

None - though I've added a new example in the docs, so there is a visual change in Percy.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Swap inset text top margin with padding ([PR #4634](https://github.com/alphagov/govuk_publishing_components/pull/4634))
+* Allow margin bottom to be configured on the table component ([PR #4638](https://github.com/alphagov/govuk_publishing_components/pull/4638))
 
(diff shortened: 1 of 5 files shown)


JamesCGDS (approved):
LGTM, nice work 👍 FYI I added a commit to the end because I spotted some unrelated typos in the table component docs

PR #4643, opened 2025-02-19
BREAKING: Remove layout_header component from layout_for_public

MartinJJones:
## What

- Remove the layout_header component from the layout_for_public view template
- Update layout_for_public tests:
  - Removed test for `show_explore_header` as this option is no longer available in layout_for_public
  - Removed tests for `show_search` and `product_name` options, these were only required when `layout_header` was included in layout_for_public. The same tests are included in [layout_header.spec.rb](https://github.com/alphagov/govuk_publishing_components/blob/main/spec/components/layout_header_spec.rb)

[Trello card](https://trello.com/c/qE79TFYj/3303-remove-legacy-header-variants-from-layoutforpublic)

## Why

To help keep provide a consistent experience when navigating GOV.UK.

### Service Manual/Toolkit

The layout_header component is only used by [/service-manual](https://www.gov.uk/service-manual) and [/service-toolkit](https://www.gov.uk/service-toolkit) pages on GOV.UK, these pages will now use the layout_super_navigation_header to help provide a more consistent navigation experience when on GOV.UK, for example, the super navigation menu is shown when navigating to /guidance/government-design-principles from the /service-toolkit page.

### Email Alert Frontend

The `show_explore_header` option is set to `false` in the [gem_layout_account_manager.html.erb](https://github.com/alphagov/static/blob/fbc921281d7161efc2adbfe8005c4069271021c7/app/views/root/gem_layout_account_manager.html.erb#L14) template in static, this template is [used by email-alert-frontend](https://github.com/alphagov/email-alert-frontend/blob/bc3ffdcd72f09708635558350493d5ef9549a018/app/controllers/subscriptions_management_controller.rb#L101) when a user is logged in.

This option is not required here as email-alert-frontend sets [show_cross_service_header](https://github.com/alphagov/static/blob/fbc921281d7161efc2adbfe8005c4069271021c7/app/views/root/gem_layout_account_manager.html.erb#L13 ) to true to use the [One Login header](https://components.publishing.service.gov.uk/component-guide/cross_service_header) instead.

### Help simplify the logic in the code

The default value for `show_explore_header` is false in the layout_for_public view template, however on most of GOV.UK the super navigation menu is used, as the [value for show_explore_header is set to true in static](https://github.com/alphagov/static/blob/ffa7fba8bed9156f36b29e20657fa773dff907e5/app/views/root/_gem_base.html.erb#L12C3-L12C22).

This means that you have to set the `show_explore_header` value back to false in static if you do not want the super navigation menu, this is currently [only used by the service manual/toolkit](https://github.com/alphagov/static/blob/ffa7fba8bed9156f36b29e20657fa773dff907e5/app/views/root/gem_layout_no_footer_navigation.html.erb#L4).

## Visual Changes

The visual changes in Percy are expected following the updates to the docs.

### Service Manual - Desktop

| Before | After |
| --- | --- |
| sm-desktop-before | sm-desktop-after |

| Before | After |
| --- | --- |
| sm1-desktop-before | sm1-desktop-after |

### Service Manual - Mobile

| Before | After |
| --- | --- |
| smh-mobile-before | smh-mobile-after |

| Before | After |
| --- | --- |
| sm-mobile-before | sm-mobile-after |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Move govspeak attachment sass includes into view ([PR #4659](https://github.com/alphagov/govuk_publishing_components/pull/4659))
+* **BREAKING:** Remove layout_header component from layout_for_public ([PR #4643](https://github.com/alphagov/govuk_publishing_components/pull/4643))
 
 ## 53.0.0
 
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -19,7 +19,6 @@
   omit_footer_border ||= false
   omit_header ||= false
   product_name ||= nil
-  show_explore_header ||= false
   show_cross_service_header ||= false
   draft_watermark ||= false
   show_search = local_assigns.include?(:show_search) ? local_assigns[:show_search] : true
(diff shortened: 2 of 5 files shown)


jon-kirwan (commented):
Looks great. I have one suggestion. I think now the `remove_bottom_border: true` option / setting is removed from the if / else condition that displays the service manual header, we no longer need the setting from the layout public header and it can be removed.

jon-kirwan (commented):
Likewise, is `show_search` still needed? It doesn't seem to be passed through to any of the layout header options now?

MartinJJones:
@jon-kirwan thanks for the review and suggestions

As you say, following the changes in this PR, it should now be safe to remove the `remove_bottom_border` and `show_search` options from the layout-header component as they are no longer used.

There could also be more options that can be removed, for example, the [search partial](https://github.com/alphagov/govuk_publishing_components/blob/4a5e21b751076086cb72f0e4a6c70fdf38c6ecf8/app/views/govuk_publishing_components/components/layout_header/_search.html.erb#L26) used by the layout-header component also passes `no_border: true` to the search component, this appears to be the [only place it is used](https://github.com/search?q=org%3Aalphagov+no_border+AND+search&type=code).

jon-kirwan:
Thanks @MartinJJones I agree - it makes sense to handle all the surplus options in a separate PR. I'll approve 👍

PR #4656, opened 2025-02-26
Add custom css exclude list to Asset Helper

KludgeKML:
## What
Add an option to supply a custom css exclude list - this is like the current list that excludes anything always supplied by static from the individual component support, but for non-static based apps, where they might as a stopgap want to provide their own application.css with a hardcoded list of styles to include, but that might not be the same list as static's list.

Also mild refactoring of the asset helper to improve test coverage.

## Why
To support a bundled set of stylesheets for email-alert-frontend, as here: https://github.com/alphagov/email-alert-frontend/pull/1890.

https://trello.com/c/O6JhukLu/487-remove-slimmer-email-alert-frontend

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Add custom css exclude list to Asset Helper ([PR #4656](https://github.com/alphagov/govuk_publishing_components/pull/4656))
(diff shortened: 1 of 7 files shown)


MartinJJones (approved):
Very nice work on this! I like the addition of the `custom_css_exclude_list` feature, the refactored code and updated tests look good to me as well 👍

PR #4659, opened 2025-02-27
Move govspeak attachment sass includes into view

andysellick:
## What
- govspeak attachment sass included all necessary attachment component sass files in order to render any attachment components within govspeak
- this was done before our per-page asset model, which meant that these styles were always compiled into the govspeak sass regardless of whether they were already included on the page
- moving them out of the sass and into the view for the govspeak component means they are now included in the per-page asset handling and can be de-duplicated
- this only impacts pages where attachment components are present alongside govspeak containing attachments, which probably isn't that many, but it makes the sass handling more transparent and included in our general approach
- some extra sass is now required in the attachment component to prevent overrides, which admittedly isn't ideal
- however it will mean that if there is more than one govspeak section on a page, these assets will be de-duplicated

Worth noting that a similar thing is happening in govspeak with button styles, but they're included specifically there in order to give them higher precedence and override the govspeak link styles, which otherwise would break button styles.

## Why
We want to optimise our stylesheet loading and ensure that every part of our system is using a unified approach.

## Visual Changes
None.

Trello card: https://trello.com/c/IGbGxOO6/481-standardise-govspeak-element-spacing (not really related but discovered as part of this work)

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -9,11 +9,6 @@
 
 .govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
 .gem-c-govspeak {
-  // Scope attachment and attachment-link component styles to gem-c-govspeak
-  @import "../attachment-link";
-  @import "../attachment";
-  @import "../details";
-
   // This block is duplicated from Whitehall as a transitional step, see the
   // commit message for 2d893c10ee3f2cab27162b9aba38b12379a71d07 before making
   // changes, original version:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Move govspeak attachment sass includes into view ([PR #4659](https://github.com/alphagov/govuk_publishing_components/pull/4659))
+
 ## 53.0.0
 
 * **BREAKING:** Remove title component ([PR #4653](https://github.com/alphagov/govuk_publishing_components/pull/4653))
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -76,6 +76,18 @@ $thumbnail-icon-border-colour: govuk-colour("mid-grey");
   }
 }
 
+.govspeak,
+.gem-c-govspeak {
+  .gem-c-attachment__title {
+    @include govuk-font($size: 27, $weight: regular);
+    margin: 0 0 govuk-spacing(3) 0;
+  }
+
+  .gem-c-attachment__metadata {
+    margin: 0 0 govuk-spacing(3) 0;
+  }
+}
+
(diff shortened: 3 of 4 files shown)


andysellick:
@MartinJJones would you be able to review this PR please? Feels like you'll be able to tell me if this is sensible or not.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -9,11 +9,6 @@
 
 .govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
 .gem-c-govspeak {
-  // Scope attachment and attachment-link component styles to gem-c-govspeak
-  @import "../attachment-link";
-  @import "../attachment";
```
MartinJJones, line 14: It looks like the main reason for importing the attachment stylesheets here is to ensure the components are rendered correctly when nested in either a `govspeak` or `gem-c-govspeak` class.

For example, Percy is now showing a [visual change for the attachment component](https://percy.io/55ba5e1a/web/govuk_publishing_components/builds/38950974/changed/2104954930?browser=firefox&browser_ids=65&group_snapshots_by=similar_diff&subcategories=unreviewed%2Cchanges_requested&utm_source=github_status_public&viewLayout=overlay&viewMode=original&width=1280&widths=375%2C1280), essentially reverting the change made in  https://github.com/alphagov/govuk_publishing_components/pull/856

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -9,11 +9,6 @@
...
-  @import "../details";
```
MartinJJones, line 15: Moving the details component stylesheet to the govspeak view template does not appear to cause any rendering issues when tested on https://www.gov.uk/government/groups/secretary-of-state-for-transports-honorary-medical-advisory-panel-on-driving-and-psychiatric-disorders

MartinJJones (commented):
It looks like we could move the `details` component to the `govspeak.html` view template without causing any rendering issues, but moving the attachment stylesheets may cause rendering issues when used in govspeak. I've included further details in the comments, happy to review/discuss further 👍

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
andysellick: Thanks for spotting this @MartinJJones. I think I can mitigate this by adding these extra styles into the attachment component sass file. This isn't ideal but I think it might still be better than duplicating all this Sass. There's still a slight visual diff as the last paragraph now has an override bottom margin that isn't then overridden by govspeak (it would normally take the bottom margin off the last paragraph) - but I think that's about to change anyway in https://github.com/alphagov/govuk_publishing_components/pull/4623

MartinJJones: This approach makes sense to me, the fix looks good. As you say https://github.com/alphagov/govuk_publishing_components/pull/4623 seems to be a quite a big update to spacing, agree it would make sense to get this change in first and then look at it again in the spacing updates PR.

MartinJJones (approved):
Nice work on this, the changes look good to me and the commit message explains why the override is needed 👍 

I think it just needs a changelog entry

PR #4660, opened 2025-02-27
Rubocop rspec

KludgeKML:
## What
Turn on rails and rspec rubocops, and make current rspec code compliant with the new cops.

## Why
Currently we're only using the default config file for govuk-rubocop, but we have an rspec test suite and this is technically a rails app since it's an engine, so we should be using the extended configurations for rails and rspec.

## Visual Changes
No visual changes.


spec/visual_regression_tests/all_components_spec.rb
AshGDS, line 1: I think this is configured to run on PRs only https://github.com/alphagov/govuk_publishing_components/blob/370053e8c55b86540a7d6523f4b3a52bf25f0699/.github/workflows/visual-regression-tests.yml#L53

Here's the PR that introduced it if helpful: https://github.com/alphagov/govuk_publishing_components/pull/1013

AshGDS (approved):
LGTM 👍 Could you make people aware of this on #govuk-ruby-frontenders as the renames etc. might cause merge conflicts.

AshGDS:
May also be worth adding this as a patch in the changelog.

spec/visual_regression_tests/all_components_spec.rb
KludgeKML: Ah, that makes sense.

KludgeKML: Comment updated.

KludgeKML:
> LGTM 👍 Could you make people aware of this on #govuk-ruby-frontenders as the renames etc. might cause merge conflicts.

Ah yes, good call. Done!

PR #4661, opened 2025-02-28
Fix failing Magna Charta test in Safari

### LINKED ISSUE #2229: Magna charta test failing in Safari 14.1.1
It looks as if Safari rounds differently to other browsers, which causes a test for Magna Charta to fail.

Failing test: https://github.com/alphagov/govuk_publishing_components/blob/8a45baa70e7bbd2b7eb2aa94103b69395f54e059/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js#L269-L277

The test fails with the error for line 272:

> Expected '27.083333333333336%' to equal '27.0833%'.

 and for line 275:

> Expected '10.833333333333334%' to equal '10.8333%'.

which makes me think that the `cW` helper function is the culprit:

https://github.com/alphagov/govuk_publishing_components/blob/8a45baa70e7bbd2b7eb2aa94103b69395f54e059/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js#L15-L25

MartinJJones:
## What

Fix failing Magna Charta test in Safari:
- Updated the `cW` function to avoid truncating the calculated width, this matches the default behaviour in Magna Charta which does not restrict the number of decimal places
- Use the [toBeCloseTo](https://jasmine.github.io/api/5.6/matchers#toBeCloseTo) matcher and set the number of decimal points to check to 4 when checking the calculated width

## Why

Fixes: https://github.com/alphagov/govuk_publishing_components/issues/2229

As mentioned in the [GitHub issue ](https://github.com/alphagov/govuk_publishing_components/issues/2229) issue there is a difference in the number of decimal places a browser will go to, using a [chart from the component guide](https://components.publishing.service.gov.uk/component-guide/govspeak/charts/preview) as an example:

### Chrome

```HTML
<div
  class="mc-td mc-bar-cell mc-bar-1 mc-bar-outdented"
  style="width: 2.70833%;" <- 5 decimal places
>
  <span style="margin-left: 100%; display: inline-block;">2</span>
</div>
```

### Safari

```HTML
<div
  class="mc-td mc-bar-cell mc-bar-1 mc-bar-outdented"
  style="width: 2.708333%;" <- 6 decimal places
>
  <span style="margin-left: 100%; display: inline-block;">2</span>
</div>
```

## Further info

The failing test in Safari highlights a couple of other things we may want to consider in the future, for example:
- We could update how widths are calculated in Magna Charta and restrict the number of decimal places so it is consistent across all browsers, unlikely to cause a noticeable difference visually when converted to pixels
- There are lots of other failings tests in Safari, I've not investigated them further but I will create a new issue for this as it would be good to understand what is causing them and if they need to be fixed in Safari
- Lastly, if the failing Safari tests in Jasmine highlight some differences/bugs in different browsers, we could consider running the tests in different browsers to help make them more visible

### DIFF
--- a/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js
+++ b/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js
@@ -14,14 +14,8 @@ describe('Magna charta', function () {
   // widths are 65/max * val (65 by default)
   var cW = function (max, val, padding) {
     padding = padding || 0
-    var result = ((65 / max) * val + padding).toString()
-    // truncate the result to only 4 digits after the decimal place
-    // e.g. 27.0833333333 becomes 27.0833, 27.1 and 27 remain the same
-    var split = result.split('.')
-    if (split.length > 1) {
-      result = split[0] + '.' + split[1].substring(0, 4)
-    }
-    return result + '%'
+    var width = ((65 / max) * val + padding)
+    return `${width}%`
   }
 
   var single =
@@ -282,12 +276,15 @@ describe('Magna charta', function () {
 
     it('the bar cells are given the right widths', function () {
       var cells = graph.find('.mc-bar-cell')
-      expect(cells.get(0).style.width).toEqual(cW(12, 5))
-      expect(cells.get(1).style.width).toEqual(cW(12, 6))
-      expect(cells.get(2).style.width).toEqual(cW(12, 6))
-      expect(cells.get(3).style.width).toEqual(cW(12, 2))
-      expect(cells.get(4).style.width).toEqual(cW(12, 3))
-      expect(cells.get(5).style.width).toEqual(cW(12, 9))
+
+      for (var cell of cells) {
+        var cellText = cell.textContent
+        var cellWidth = cell.style.width
+        var calculatedWidth = parseFloat(cW(12, cellText))
+
+        expect(cellWidth).toContain('%')
+        expect(parseFloat(cellWidth)).toBeCloseTo(calculatedWidth, 4)
+      }
     })
 
     it('the bar cells are given classes denoting their index', function () {
--- a/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js
+++ b/spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js
@@ -15,7 +15,7 @@ describe('Magna charta', function () {
   var cW = function (max, val, padding) {
     padding = padding || 0
     var width = ((65 / max) * val + padding)
-    return `${width}%`
+    return `${parseFloat(width.toFixed(4))}%`
   }
 
   var single =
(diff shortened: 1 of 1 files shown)


spec/javascripts/govuk_publishing_components/lib/govspeak/magna-charta-spec.js
```diff
@@ -282,12 +276,15 @@ describe('Magna charta', function () {
...
-      expect(cells.get(4).style.width).toEqual(cW(12, 3))
-      expect(cells.get(5).style.width).toEqual(cW(12, 9))
+
+      for (var cell of cells) {
+        var cellText = cell.textContent
+        var cellWidth = cell.style.width
+        var calculatedWidth = parseFloat(cW(12, cellText))
+
+        expect(cellWidth).toContain('%')
+        expect(parseFloat(cellWidth)).toBeCloseTo(calculatedWidth, 4)
```
jon-kirwan, line 286: Just a small note - do we need two expectations here? Perhaps we should merge them into a single test that checks the correct width as a percentage string?

MartinJJones: Agreed, I think this would be the ideal approach, I cannot see a good way to do this however while supporting both Safari and Chrome. For example, Magna Charta does not currently set the number of decimal places, using `2.7083333333` as an example width, this will get set as below in each browser:

Chrome = `2.70833%`
Safari = `2.708333%`

We have a couple of options to get around this:

**Update the widths before doing the comparison**

I've added an example of this in the latest commit - https://github.com/alphagov/govuk_publishing_components/pull/4661/commits/2f6b5f7ef14b6b398580d9aab6b94d0d6f53cea9.

I'm not 100% sure about this approach though, especially as it updates the width set in the browser, using the `toBeCloseTo` matcher feels better here I think.

**Update Magna Charta**

Another option could be to update Magna Charta so it does restrict the number of decimal places, ensuring a consistent width in each browser, overall this feels like it would be done to satisfy the tests and seems unlikely to provide any benefits, could be worth exploring further.

**Does the issue need fixing?**

Lastly, I think even if we do update the tests to support Safari, we are unlikely to see the errors unless we test this locally as we use [headlessChrome](https://github.com/alphagov/govuk_publishing_components/blob/e55960fb09385faa0a2af76052f727aa17731e06/spec/support/jasmine-browser.json#L28).

I would be reluctant to dismiss it completely until we can be certain that the other failing tests are not genuine issues in Safari, but does raise the question of how we could make failures in Safari more visible in the future if required.

jon-kirwan: Hi @MartinJJones, sorry to take a while to respond and thanks for adding the extra commit. I see what you mean and agree with you regarding the above options.

I think then your first fix is probably the best approach:
```js
expect(cellWidth).toContain('%')
expect(parseFloat(cellWidth)).toBeCloseTo(calculatedWidth, 4)
``` 
Two final things: perhaps it's worth raising it again at the community meeting to see if we should fix Jasmine Safari issues? Or whether it's better to only acknowledge the issue.

One final idea we could try is to [match a Regex](https://jasmine.github.io/api/2.8/matchers.html#toMatch) which _should_ allow for one expectation to check the correct value and appended percentage sign e.g.
```js
var cellStyleWidth = cell.style.width
var calculatedWidth = parseFloat(cW(12, cell.textContent))
var regex = new RegExp(`^${calculatedWidth}(\\d{0,2})?%$`);

expect(cellStyleWidth).toMatch(regex);
```

PR #4679, opened 2025-03-05
Fix govspeak legislative lists styles

andysellick:
## What
Fix the appearance of legislative lists on pages like https://www.gov.uk/guidance/immigration-rules/immigration-rules-introduction (see the 'interpretation' section).

Have also updated the component guide to include a more detailed example of a legislative list, hopefully so any future changes will be caught by the visual diff.

## Why
This was accidentally broken in https://github.com/alphagov/govuk_publishing_components/pull/4623

## Visual Changes

Before | After
------ | ------
Screenshot 2025-03-05 at 14 43 40 | Screenshot 2025-03-05 at 14 43 56

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


KludgeKML (approved):
LGTM

PR #4686, opened 2025-03-06
Add 'contents list with body' component from government-frontend

AshGDS:


andysellick:
The other advantage of my suggestion is it might be possible to make the functionality work as intended in the component guide - doesn't seem to currently.

andysellick (approved):

AshGDS:
Thanks @andysellick  - I think it's worth raising as a separate issue.

The [JS in the gem preview](https://components-gem-pr-4686.herokuapp.com/component-guide/contents_list_with_body/preview) seems to act in the same way to how the live site is working for me https://www.gov.uk/cma-cases/refresco-cott-merger-inquiry - I think it just might not be visible if your screen is too tall / the example content is too short. I can make the gem example content longer if that would be helpful?

andysellick:
@AshGDS ah yes I see it working now, thanks 👍

PR #4687, opened 2025-03-06
Remove search from the layout header component

jon-kirwan:
## What

**Breaking**: Remove search from the layout header component. These options seem to be used only by the service manual which was recently updated to incorporate the layout super navigation header used across all public-facing GOV.UK pages

## Why

https://trello.com/c/JAzqXWVY/1848-remove-the-search-functionality-from-the-layout-header-component

## Anything else

I wasn't sure if any JavaScript needed to be removed with these changes, but I see that the toggle button (`govuk-js-header-toggle`) behaviour comes from GOV.UK Frontend. I also note that this behaviour is set to be removed in the next breaking release.
> GOV.UK header with navigation
We’ve deprecated the [GOV.UK header with navigation](https://design-system.service.gov.uk/components/header/with-service-name-and-navigation/) and will remove this option in the next breaking release of GOV.UK Frontend.

https://design-system.service.gov.uk/components/header/#govuk-header-with-navigation


app/assets/stylesheets/govuk_publishing_components/components/_search.scss
```diff
@@ -271,50 +271,3 @@ $large-input-size: 50px;
     padding-left: 0;
   }
 }
-
-.search-toggle {
-  display: none;
-  background-color: govuk-colour("blue");
-  background-image: url("govuk_publishing_components/search-button.png");
```
MartinJJones, line 278: Can the `search-button.png` image file also be removed?

MartinJJones (commented):
Nice work!, the changes look good to me, just a couple of other questions/suggestions:

- Does this need to go out as a breaking change?
- Can we also remove `search-button.png`? I don't think this is used anywhere now, if it is used in other applications and this does go out as a breaking change, we could make the recommendation to use the [SVG search icon](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/search/_search_icon.html.erb) used by the main search component? 🤔

app/assets/stylesheets/govuk_publishing_components/components/_search.scss
jon-kirwan: Yep, I agree. That doesn't seem to be used anywhere. Removed now.

jon-kirwan:
> Nice work!, the changes look good to me, just a couple of other questions/suggestions:
> 
> * Does this need to go out as a breaking change?
> * Can we also remove `search-button.png`? I don't think this is used anywhere now, if it is used in other applications and this does go out as a breaking change, we could make the recommendation to use the [SVG search icon](https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/search/_search_icon.html.erb) used by the main search component? 🤔

Thanks @MartinJJones. Good point. I've updated it to be a breaking change. I'm not sure if we need to provide a workaround for the removed image, though. I scanned the alphagov repos, and it looks like any instances of `search-button` are local images rather than being referenced from the gem.

MartinJJones (approved):

PR #4690, opened 2025-03-10
Remove IE8 support from action-link component

MartinJJones:
## What

- Remove PNG images from the [action-link](https://components.publishing.service.gov.uk/component-guide/action_link) component
- Remove IE8 fix that ensured PNG instead of SVG images are requested

## Why

The PNG images are no longer required, they were originally added to support IE8, however the icon does not load in IE8 as we use [double-colon](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/_action-link.scss#L6) on the before pseudo-element in the action-link component.

> IE8 only supports the single-colon CSS 2.1 syntax (i.e. :pseudo-class). It does not support the double-colon CSS3 syntax (i.e. ::pseudo-element) - https://caniuse.com/css-gencontent

This change means we only need to create 1 SVG image in the future, this could also make it easier to apply different colour variations to the arrow icon using CSS instead of creating a new asset each time.

[Trello card](https://trello.com/c/isZSvjKH/3352-remove-ie8-support-from-action-link-component)


jon-kirwan (approved):
Looks good to me!

PR #4691, opened 2025-03-10
Add published dates component

AshGDS:
## What / Why

- Migrates the published dates component from [government-frontend](https://govuk-government-frontend.herokuapp.com/component-guide/published_dates/) into [this gem]()
- Trello card: https://trello.com/c/eF1B4BDf/513-migrate-government-frontend-components-to-the-gem

## Visual Changes

There should be none, but you can compare here: [government-frontend](https://govuk-government-frontend.herokuapp.com/component-guide/published_dates/preview) and [this gem]()


app/views/govuk_publishing_components/components/_published_dates.html.erb
```diff
@@ -0,0 +1,50 @@
...
+  component_helper.set_lang("en")
+%>
+<% if published || last_updated %>
+  <%= tag.div(**component_helper.all_attributes) do %>
+    <h2 class="govuk-visually-hidden"><%= t('components.published_dates.hidden_heading') %></h2>
+    <% if published %>
+      <%= t('components.published_dates.published', date: published) %>
+    <% end %>
+    <% if last_updated %>
+      <% if published %><br /><% end %><%= t('components.published_dates.last_updated', date: last_updated) %>
```
andysellick, line 24: Not for this PR, but we should maybe circle back and look at why this `<br/>` is being inserted and whether it could be resolved with CSS instead.

andysellick: Looking through the rest of this file there's some similar things that perhaps could be done better, like the non breaking space on line 36 and other HTML symbols e.g. line 32.

andysellick (approved):

app/views/govuk_publishing_components/components/_published_dates.html.erb
AshGDS: Thanks - will raise an issue for this 👍

PR #4702, opened 2025-03-12
Delete unused CSS from layout super navigation component

jon-kirwan:
## What

Delete unused CSS from layout super navigation component

## Why

https://trello.com/c/mKyTK3U3/3358-remove-unused-css-in-super-navigation-layout-header

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Shift footnotes above 100/1000 further to the left ([PR #4705](https://github.com/alphagov/govuk_publishing_components/pull/4705))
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Nice work on this, good spot on some of the other unused styles as well 👍 

As you mentioned, looks like the commits could be squashed, and I think a changelog entry.

jon-kirwan:
> As you mentioned, looks like the commits could be squashed, and I think a changelog entry.

Thanks @MartinJJones! I decided to squash the first few commits together but kept separate ones for the chevron removal and `govuk-link` mixin replacements just in case there's any issues and it's easier to track down what changed.

MartinJJones:
> I decided to squash the first few commits together but kept separate ones for the chevron removal and govuk-link mixin replacements just in case there's any issues and it's easier to track down what changed.

That makes sense to me, looks good 👍

PR #4718, opened 2025-03-18
Make minor improvements to the figure and published date components 

AshGDS:
## What / Why

- Adds some improvements to the components that were migrated over from `government-frontend`.
- More specifically:
    - For the figure component, I fixed an issue where the `<figure>` element would still render if the image src was an empty string. Currently this is an issue on some pages on production, as content items can state that they have an image to render but then the `src` is just `""`. This is being fixed by backend devs in `frontend`, but I thought it was worth guarding against this at the component level as well. I didn't use validation for this in case it caused a bunch of production pages to start crashing, so instead it just renders nothing to the DOM silently.
    - I cleaned up the published date component. I also decided to hide the symbols (the hyphen, plus and em dash) from the accessibility API as I think they would make the screen reader experience more confusing.
- It's probably worth reading the changes commit by commit for them to make sense.
- Trello card: https://trello.com/c/Zcsu8Xth/532-audit-app-components-migrated-in-from-government-frontend

## Component links
- Figure: https://components-gem-pr-4718.herokuapp.com/component-guide/figure
- Published dates: https://components-gem-pr-4718.herokuapp.com/component-guide/published_dates

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Make minor improvements to the `figure` and `published_dates` components ([PR #4718](https://github.com/alphagov/govuk_publishing_components/pull/4718))
+
 ## 55.1.0
 
 * Limit organisation logos to 320px/240px width ([PR #4708](https://github.com/alphagov/govuk_publishing_components/pull/4708))


app/views/govuk_publishing_components/components/_figure.html.erb
```diff
@@ -3,26 +3,27 @@
...
   caption ||= ""
   credit ||= ""
+  src ||= nil
   local_assigns[:lang] ||= "en"
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-figure")
 %>
-<%= tag.figure(**component_helper.all_attributes) do %>
-  <% if src.present? %>
+<% if src.present? && !src.blank? %>
```
JamesCGDS, line 11: I think we can just use `src.present?` here as I believe it checks against `nil`, empty strings, and whitespace-only strings.

JamesCGDS (commented):
Nice work, just have one minor comment

app/views/govuk_publishing_components/components/_figure.html.erb
AshGDS: Good spot, I've made this change now, thanks @JamesCGDS

PR #4724, opened 2025-03-21
Fix tests

andysellick:
## What / why
Fixes a few problems with the JavaScript tests.

Firstly:

- if you're running the JavaScript tests in a browser using `yarn run jasmine:browser` and click anything like the seed link, this test would fail
- because it expected a pageview object with nothing in the query string part of the URL e.g. ?search=
- this updates these tests to allow for this possibility, meaning they no longer fail when clicking a specific seed during testing

Secondly:

- when all tests are run with seed 54426 an error to do with handling an undefined variable occurred
- tracked it down to this line in `cookie-settings.js` - `delete currentConsentCookieJSON.essential`
- seems like in some circumstances currentConsentCookieJSON doesn't exist, which explains why an attribute can't be deleted from it
- wrapping the code in a check fixes the failing test

## Visual Changes
None.


andysellick:
@AshGDS have added the extra bit so this is ready for review, although I'm still uncertain about what the actual cause of the second issue is.

AshGDS (approved):

PR #4732, opened 2025-03-25
Add NI numbers to GA4 PII redaction

AshGDS:
## What / Why

- Adds national insurance redaction to the GA4 PII remover so that it doesn't have to be redacted further down the line
- The regex I wrote is based partly on this guidance https://www.gov.uk/hmrc-internal-manuals/national-insurance-manual/nim39110 and https://github.com/alphagov/govuk-chat-private-beta/blob/2add98c736e68f630d825cbbf00e1db11a42e68d/app/validators/pii_validator.rb#L27
- For the tests I wrote some code to randomly generate numbers, just so we aren't actually storing any in the code.
- Trello card: https://trello.com/c/ndiVNNuG

## Visual Changes

None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
@@ -32,6 +32,10 @@
   var VISA_PATTERN_GWF = /GWF\d{9}/g
   var VISA_PATTERN_GB = /GB\d{12}/g
 
+  // 2 valid letters - optional +/space - 6 digits with optional +/spaces every 2 digits (to account for '12 34 56') - optional +/space - 1 valid letter. Case insensitive.
+  // e.g. 'AB123456A', 'AB 12 34 56 A', 'ab 123456 a', 'ab 12 34 56 a', 'AB+12+34+56+A'
+  var NATIONAL_INSURANCE_NUMBER = /[A-CEGHJ-OPR-TW-Z]{2}(\s+|\++)?(\d{2}(\s+|\++)?){3}[A-D]/gi
+
   function shouldStripDates () {
     var metas = document.querySelectorAll('meta[name="govuk:ga4-strip-dates"]')
     return metas.length > 0
@@ -104,6 +108,7 @@
     stripped = stripped.replace(STATE_PATTERN, 'state=[state]')
     stripped = stripped.replace(VISA_PATTERN_GWF, '[gwf number]')
     stripped = stripped.replace(VISA_PATTERN_GB, '[gb eori number]')
+    stripped = stripped.replace(NATIONAL_INSURANCE_NUMBER, '[ni number]')
     stripped = this.stripQueryStringParameters(stripped)
 
     if (this.stripDatePII === true) {
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
@@ -275,6 +275,97 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
     }
   })
 
+  describe('national insurance numbers', function () {
+    it('are stripped if they are valid NI numbers', function () {
+      // Generate 50 random NI numbers so we aren't storing any actual ones in the code.
+      for (var i = 0; i < 50; i++) {
+        var testNumber = generateNINumber(true, false)
+        var testNumberWithSpaces = generateNINumber(true, true)
+
+        var redacted1 = pii.stripPIIWithOverride(testNumber, false, false)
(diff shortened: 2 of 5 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -275,6 +275,94 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
...
+
+  function getRandomArrayValue (array) {
+    return array[getRandomNumber(array.length)]
+  }
+
+  function generateNINumber (validNumber, withSpaces) {
+    var firstLetters, lastLetters
+
+    if (validNumber) {
+      firstLetters = ['A', 'C', 'E', 'G', 'H', 'J', 'O', 'P', 'R', 'T', 'W', 'Z']
```
andysellick, line 335: Non blocking: as a readability improvement here I think you could do something like

```JavaScript
firstLetters = 'A C E G H J O P R T W Z'.split(" ")
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -32,6 +32,10 @@
   var VISA_PATTERN_GWF = /GWF\d{9}/g
   var VISA_PATTERN_GB = /GB\d{12}/g
 
+  // 2 valid letters - optional space - 6 digits with optional spaces every 2 digits (to account for '12 34 56') - optional space - 1 valid letter
+  // e.g. AB123456A or AB 12 34 56 A
+  var NATIONAL_INSURANCE_NUMBER = /[A-CEGHJ-OPR-TW-Z]{2}\s?(\d{2}\s?){3}\s?[A-D]/gi
```
andysellick, line 37: Will this be used to redact NI numbers from URLs? If you type a national insurance number into the GOV.UK search the URL comes out with `+` characters instead of spaces e.g. `https://www.gov.uk/search/all?keywords=BR+12+34+56+` - will this still be redacted?

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -275,6 +275,94 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
...
+    var niNumber = getRandomArrayValue(firstLetters) + getRandomArrayValue(firstLetters)
+
+    // Add optional space
+    if (withSpaces) {
+      niNumber += ' '
+    }
+
+    // Add 6 random digits between 0 and 9
+    for (var j = 0; j < 6; j++) {
+      niNumber += getRandomNumber(10)
```
andysellick, line 352: Where does this insert a space character every two numbers? I'm possibly overlooking it.

AshGDS: Good spot, I've now added some code to handle that.

AshGDS:
Thanks @andysellick - I've pushed the improvements as separate commits so it should be ready for another review 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
```diff
@@ -34,7 +34,7 @@
 
   // 2 valid letters - optional space - 6 digits with optional spaces every 2 digits (to account for '12 34 56') - optional space - 1 valid letter
   // e.g. AB123456A or AB 12 34 56 A
-  var NATIONAL_INSURANCE_NUMBER = /[A-CEGHJ-OPR-TW-Z]{2}\s?(\d{2}\s?){3}\s?[A-D]/gi
+  var NATIONAL_INSURANCE_NUMBER = /[A-CEGHJ-OPR-TW-Z]{2}(\s+|\++)?(\d{2}(\s+|\++)?){3}[A-D]/gi
```
andysellick, line 37: This better but sorry, should have thought of this sooner - it only matches uppercase letters. We should probably match lowercase as well.

AshGDS: Ah, sorry maybe it's not clear enough but with the `/gi`, at the end, the `i` means case insensitive and I've written a test for lowercase too. Should I update the comment to mention it's case insensitive? I probably should add the `+` functionality to the comment anyway.

andysellick: Ah of course it does, sorry hadn't copied and pasted the full regex for testing. Yeah, might be worth updating the comment.

andysellick (approved):

app/assets/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.js
AshGDS: No worries, will do - thanks 👍

Issue #3503, opened 2023-07-18
Review `attachment` component with external links


MartinJJones:
## What

There appears to be a couple of issues with the [attachment component with external links](https://components.publishing.service.gov.uk/component-guide/attachment/external_attachment/preview) that are worth looking into further:

- In IE11, long URL strings are not contained and may overlap with other elements on the page as shown in the screenshot below
- The metadata text size is not consistent with other metadata component styles, here it is 19px, should it be 16px?
- The URL's visual don't look great on the page
- How well do they work with screenreaders?

### Screenshot showing the issue in IE11

<img width="1141" alt="before-text-no-hyphen" src="https://github.com/alphagov/govuk_publishing_components/assets/28779939/d987c160-c98f-4f04-9534-8c191eda7bdf">

## Further info

- https://github.com/alphagov/govuk_publishing_components/pull/3501
- https://github.com/alphagov/govuk_publishing_components/issues/3482
- https://css-tricks.com/snippets/css/prevent-long-urls-from-breaking-out-of-container/

nnagewad:
@MartinJJones is this still an issue?

MartinJJones:
> @MartinJJones is this still an issue?

This does still appear to be an issue in IE11, I was able to reproduce this in BrowserStack by testing in IE11 and modifying an attachment url to include a long string with no hyphens to break the text.

jon-kirwan:
Try the GOV.UK Frontend break word class

PR #4737, opened 2025-03-27
Use govuk-frontend mixin on attachment URL text

MartinJJones:
## What

Update the [attachment component](https://components.publishing.service.gov.uk/component-guide/attachment) to:
- Use the `govuk-text-break-word` mixin from the design system to forcibly break URLs across multiple lines when they wouldn't otherwise fit, moved to `.gem-c-attachment__metadata` to ensure the fix works as expected in IE11
- Update the external attachment example to reference the word break helper used from the design system, the hyphens have been removed from the url text to capture any visual regression issues in the future

## Why

By aligning our approach with the design system, we will benefit from any future updates to the [word break helper](https://github.com/alphagov/govuk-frontend/blob/main/CHANGELOG.md#stop-long-words-breaking-out-of-components-with-govuk--text-break-word). The updated external URL example in the documentation should also help capture any issues with the approach in the future, at the moment the URL example includes hyphens, so the text would automatically break.

### Fixed issue in IE11

Related issue: https://github.com/alphagov/govuk_publishing_components/issues/3503

| Before | After |
| --- | --- |
| ie11-before | ie11-after |

### Percy
- The visual change in shown in Percy is expected as the hyphens have been removed from the url text to capture any visual regression issues in the future

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@
 * Fix govspeak print styles ([PR #4730](https://github.com/alphagov/govuk_publishing_components/pull/4730))
 * Update LUX to 4.0.32 ([PR #4725](https://github.com/alphagov/govuk_publishing_components/pull/4725))
 * Fix print styles ([PR #4738](https://github.com/alphagov/govuk_publishing_components/pull/4738))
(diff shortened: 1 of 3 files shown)


jon-kirwan (approved):
Nice! Glad to see the mixin working for IE11. I’ve also linked the pull request to the issue, so it will close automatically when merged

PR #4739, opened 2025-03-27
Fix govspeak blockquotes

andysellick:
## What
Fix the appearance of govspeak blockquotes, primarily to remove the negative left margin that means blockquotes escape from their parent element as shown below (red outline added for clarity)

Screenshot 2025-03-27 at 10 46 02

Example in the wild on this page, note how the `cite` element is sticking out to the left: https://www.gov.uk/government/history/10-downing-street

Screenshot 2025-03-27 at 14 25 31

I've also fixed a few other things (see commits for details) and generally tried to clean up and simplify these styles.

## Why
As often happens, I was working on something else and noticed that this was a bit broken.

## Visual Changes

Before | After
------ | ------
Screenshot 2025-03-27 at 14 20 37 | Screenshot 2025-03-27 at 14 20 45

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
@@ -33,29 +33,20 @@
   // Block quotes
 
   blockquote {
-    padding: 0 0 0 govuk-spacing(4);
-    margin: 0;
+    padding: 0 0 0 govuk-spacing(6);
+    margin: 0 0 govuk-spacing(4) 0;
     border: 0;
 
-    @include govuk-media-query($from: desktop) {
-      margin: 0 0 0 (- govuk-spacing(6));
-    }
-
     p {
-      padding-left: govuk-spacing(3);
+      position: relative;
 
-      @include govuk-media-query($from: tablet) {
-        padding-left: govuk-spacing(6);
-      }
-
-      &::before {
+      &:first-of-type::before {
         content: "\201C";
-        float: left;
-        clear: both;
-        margin-left: (- govuk-spacing(3));
+        position: absolute;
+        right: 100%;
       }
 
-      &:last-child::after {
+      &:last-of-type::after {
         content: "\201D";
       }
     }
--- a/app/views/govuk_publishing_components/components/docs/govspeak.yml
+++ b/app/views/govuk_publishing_components/components/docs/govspeak.yml
@@ -50,7 +50,12 @@ examples:
         </ul>
         <p>A paragraph of text following the list. I'm not going to talk about the words anymore.</p>
         <blockquote>
-          <p>This text is inside a blockquote.</p>
+          <p>This text is inside a blockquote. It is being made deliberately long so as to wrap onto a second line, which hopefully has happened by now.</p>
+        </blockquote>
+        <blockquote>
+          <p>This text is inside a blockquote with a citation.</p>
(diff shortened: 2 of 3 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
```diff
@@ -33,32 +33,29 @@
...
 
-      &:last-child::after {
+      &:last-of-type::after {
         content: "\201D";
       }
     }
+
+    cite {
+      display: block;
+      margin-bottom: govuk-spacing(4);
```
JamesCGDS, line 57: Would there by any mileage in amending the margin ruleset for the blockquote as a whole (currently on line 37) to something like `margin: 0 0 govuk-spacing(4) 0`? Only because I recall a conversation we had a while back about margins and, if I remember correctly, one of the things we mused about was moving all (or the majority) of spacing to the parent component to prevent child elements from causing confusion when looking at spacing.

This all looked good though so totally happy to leave things as they are if that's preferable.

app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -50,7 +50,12 @@ examples:
         </ul>
         <p>A paragraph of text following the list. I'm not going to talk about the words anymore.</p>
         <blockquote>
-          <p>This text is inside a blockquote.</p>
+          <p>This text is inside a blockquote. It is being made deliberately long so as to wrap onto a second line, which hopefully has happened by now.</p>
+        </blockquote>
+        <blockquote>
+          <p>This text is inside a blockquote with a citation.</p>
+          <p>It contains more than one paragraph, and also has extra words in order to make it wrap onto a second line so we can check that everything's okay.</p>
+          <cite>John Blockquote, inventor of the blockquote</cite>
```
JamesCGDS, line 58: Would there be any mileage in adding this example to the [individual blockquote example](https://github.com/alphagov/govuk_publishing_components/blob/41b98c097a0142e096dfd77ce2ff079b180fc509/app/views/govuk_publishing_components/components/docs/govspeak.yml#L464) too? Only because when I was reviewing and refreshing the page I had to search a bit to find the blockquote amongst the other common elements.

This might just be me though so totally happy to leave things are they are if it's not an issue.

JamesCGDS (commented):
Looks good! Just got a couple of minor comments/questions

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
```diff
@@ -33,32 +33,29 @@
...
+      position: relative;
 
-      &::before {
+      &:first-of-type::before {
         content: "\201C";
-        float: left;
-        clear: both;
-        margin-left: (- govuk-spacing(3));
+        position: absolute;
+        top: 0;
```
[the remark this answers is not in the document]
andysellick: Good shout, have removed 👍

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_typography.scss
andysellick: That's a good idea. I'm not going to add additional rules for the paragraphs to remove their margin, because their margin is the same as the blockquote they just overlap. But this also means we don't need styles for the `cite` anymore, so this feels a bit cleaner and more efficient.

app/views/govuk_publishing_components/components/docs/govspeak.yml
andysellick: Good shout, have copied and pasted it 👍

PR #4742, opened 2025-03-31
Add a more descriptive accessible label to summary banner

hannalaakso:
## What
Use the title text of the summary banner as the accessible label for the summary banner which is used as the label for the landmark region, by adding an aria-labelledby and link it to the summary banner title ID. Since we have no unique ID to use, use an approach similar to the notice and password input components to generate a unique ID for the title.

[Trello](https://trello.com/c/165G7d6U/3360-investigate-and-fix-accessibility-issues-on-the-gem-summary-banner-component)

## Why
 We currently use the same aria-label value ("notice") on the summary banner as is already used by the notice component. This doesn't comply with the ARIA authoring practise for each landmark to have a unique label, since the notice and the summary banner often appear on the same page. The current value "notice" also doesn't describe what the summary banner is or does very well. The change in this PR should make the label unique in most cases since we shouldn't have more than one summary banner on a page and if we do, they shouldn't have the same heading text to help the user differentiate them. 

## Screen reader testing results

Correctly announces the title text of the summary banner as the label for the landmark region:

✅ NVDA 2024 with Chrome
✅ Jaws 2025 with Chrome

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Fix HTML validation bug in cookie banner ([PR #4743](https://github.com/alphagov/govuk_publishing_components/pull/4743))
 * Add service navigation component ([PR #4731](https://github.com/alphagov/govuk_publishing_components/pull/4731))
+* Add a more descriptive accessible label to summary banner ([PR #4742](https://github.com/alphagov/govuk_publishing_components/pull/4742))
 
 ## 56.1.0
 
--- a/app/views/govuk_publishing_components/components/_summary_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_summary_banner.html.erb
@@ -2,13 +2,14 @@
   add_gem_component_stylesheet("summary-banner")
 
   title ||= false
+  title_id = "summary-banner-title-#{SecureRandom.hex(4)}"
   text ||= false
   secondary_text ||= false
   local_assigns[:margin_bottom] ||= 7
 
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-summary-banner")
-  component_helper.add_aria_attribute({ label: "Notice" })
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -11,6 +11,8 @@
 
 * Fix HTML validation bug in cookie banner ([PR #4743](https://github.com/alphagov/govuk_publishing_components/pull/4743))
 * Add service navigation component ([PR #4731](https://github.com/alphagov/govuk_publishing_components/pull/4731))
+* Add a unique accessible label to summary banner ([PR #4742](https://github.com/alphagov/govuk_publishing_components/pull/4742))
+
```
jon-kirwan, line 15: Small thing, but there's an extra line here you could remove.

jon-kirwan (approved):
Thanks @hannalaakso. All looks great to me! I tested in JAWS, NVDA, and VoiceOver, and everything works as expected. I've gone ahead and approved - just two small things:
- there's an extra line in the changelog that can be removed.
- you could also disable the `landmark-unique` accessibility warnings in the Component Guide like so: https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/service_navigation.yml#L7-L9.

hannalaakso:
> you could also disable the landmark-unique accessibility warnings in the Component Guide like so: https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/docs/service_navigation.yml#L7-L9.

Thanks for the suggestion @jon-kirwan 👍  I realised that my PR name was misleading since it's not the label that is _guaranteed_ to be unique, but the heading ID that links to it! I've amended it now. I've also amended the PR description to explain this a bit better, in that the label should in most cases be unique too but it's not guaranteed programmatically:
> The change in this PR should make the label unique in most cases since we shouldn't have more than one summary banner on a page and if we do, they shouldn't have the same heading text to help the user differentiate them.

So there is an edge case where the label could still be the same. So I think it's useful to retain the unique labels accessibility warning.

Issue #3559, opened 2023-08-18
HTML validation bug in cookie banner


maxgds:
## What:
Noted by Anika, https://gds.slack.com/archives/CARGH33JS/p1692285458097259 we have ```<p>``` tags inside a ```<span>``` tag in the cookie banner - the fix is probably to simply switch the span to a div. 

## Why:

No visible issues, but invalid HTML is not a good look. Affects all pages as the banner is presented to whatever page is the first landing page.

## Prioritisation score: `3`

Reach: `3`, Severity: `1`, Effort: `1`

KludgeKML:
Is the span even doing anything useful? Like, could those P tags just be straight inside the enclosing div?

maxgds:
> Is the span even doing anything useful? Like, could those P tags just be straight inside the enclosing div?

It's the parent container that gets hidden upon choosing a preference, so it does serve a purpose.

PR #4743, opened 2025-04-01
Fix HTML validation bug in cookie banner

jon-kirwan:
## What

Fix HTML validation bug in cookie banner

See [https://validator.w3.org HTML checker results](https://validator.w3.org/nu/?doc=https%3A%2F%2Fcomponents-gem-pr-4743.herokuapp.com%2Fcomponent-guide%2Fcookie_banner%2Fdefault%2Fpreview)

## Why

https://github.com/alphagov/govuk_publishing_components/issues/3559.

Additionally, the change aligns with the Design System implementation; see https://design-system.service.gov.uk/components/cookie-banner/default/.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


MartinJJones (approved):
Nice work, validates and matches the design system approach, the change looks good to me 👍

PR #4745, opened 2025-04-02
remove unused labels for tracking-key in super navigation header

KludgeKML:
## What

Remove tracking key values from super navigation header, and variables that feed into it.

## Why

These are UA variables, not needed now we're in the GA4 world.


AshGDS (commented):
LGTM 👍 

There are some `data-tracking-key` references in the JS spec, so it would probably be good to remove those as well:

https://github.com/alphagov/govuk_publishing_components/blob/main/spec/javascripts/components/layout-super-navigation-header-spec.js#L26

AshGDS (approved):
LGTM 👍 Probably worth logging it in the CHANGELOG too

PR #4751, opened 2025-04-08
Add an aria-label to the devolved nations component

hannalaakso:
## What
Add an aria-label to the devolved nations component. The attribute value was agreed with our content designer.

## Why
The devolved nations component currently uses the section tag but doesn't have an aria-label. This might have been an oversight since it means the component doesn't get announced as a landmark region by screen readers. Both the notice and summary banner components are currently set as landmark regions since they are deemed to contain important information that it might be useful for the user to know about when they inspect the outline of the page. To be consistent, it makes sense to make devolved nations into a landmark region as well as it often appears on the same page with the components.

## Background
There has been some conversation though about whether some of these components could be consolidated to reduce the number of landmark regions and to generally help to to improve the information hierarchy on document pages. Our team have tickets in our backlog to look at wider related improvements for this.

[Trello](https://trello.com/c/piJlXt43/3356-fix-issue-on-consultations-and-statistics-announcements-where-notice-and-banner-components-and-metadata-and-notice-components-do?filter=member:hannalaakso4)

### DIFF
--- a/app/views/govuk_publishing_components/components/_devolved_nations.html.erb
+++ b/app/views/govuk_publishing_components/components/_devolved_nations.html.erb
@@ -10,6 +10,7 @@
 
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-devolved-nations")
+  component_helper.add_aria_attribute({ label: t("components.devolved_nations.aria_label") })
 
   unless disable_ga4
     component_helper.add_data_attribute({
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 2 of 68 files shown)


MartinJJones (approved):
Nice work! the changes look good to me

app/views/govuk_publishing_components/components/_devolved_nations.html.erb
```diff
@@ -10,6 +10,7 @@
 
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class("gem-c-devolved-nations")
+  component_helper.add_aria_attribute({ label: "Nation" })
```
jon-kirwan, line 13: One minor suggestion - should we consider adding the label text to the locale file?

hannalaakso: Good shout, I've just done this 👍

jon-kirwan (approved):
Thanks for making the change! Looks good to me.

PR #4760, opened 2025-04-15
Table: add ability to specify column width

csutter:
## What
This enhances the `table` component with the ability to give columns a specific width by assigning the header cell a width override class from the width scale as suggested by the [upstream table component].

## Why
This is useful when:
* the width of a column is always static and less than the automatically inferred width (for example, a set of action links or buttons), or
* a certain column often contains long and/or irregular content, throwing the visual balance of the table out of whack

[upstream table component]: https://design-system.service.gov.uk/components/table/

### Before
image

### After
image

[upstream table component]: https://design-system.service.gov.uk/components/table/

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/table.yml
+++ b/app/views/govuk_publishing_components/components/docs/table.yml
@@ -200,3 +200,26 @@ examples:
           format: numeric
         - text: £125
           format: numeric
+  with_custom_width:
+    description: >
+      You can use `width` on a header cell to set the width of a table column if you do not want the
+      width to be inferred by the browser based on the content of its cells.
+    data:
+      head:
+        - text: Phrase
+        - text: Rating
+          width: one-half
+      rows:
+      -
+        - text: Lorem ipsum
+        - text: good
+      -
+        - text: dolor sit amet, consectetur adipiscing elit
+        - text: okay
+      -
+        - text: >
+            sed do eiusmod tempor incididunt ut labore et dolore magna aliqua duis aute irure dolor
+            in reprehenderit in voluptate velit esse cillum dolore eu fugiat nulla pariatur
+            molestiae non numquam eius modi tempora incidunt ut labore et dolore magnam aliquam
+            quaerat voluptatem!!!
+        - text: bad
--- a/spec/components/table_spec.rb
+++ b/spec/components/table_spec.rb
@@ -153,4 +153,25 @@ describe "Table", type: :view do
     assert_select ".js-gem-c-table__filter"
     assert_select ".js-gem-c-table__message"
   end
+
+  it "accepts a width value for head cells" do
+    render_component(
+      head: [
+        {
+          text: "Month you apply",
+          width: "one-third",
+        },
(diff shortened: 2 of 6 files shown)


app/views/govuk_publishing_components/components/docs/table.yml
```diff
@@ -200,3 +200,28 @@ examples:
...
+    description: >
+      You can use `extra_classes` to add additional classes to a header or body cell. This lets you
+      apply width override classes for tables where the data is inconsistent in length and would
+      otherwise cause the table to display poorly.
+    data:
+      head:
+        - text: Phrase
+        - text: Rating
+          extra_classes:
+            - 'govuk-!-width-one-third'
```
andysellick, line 213: As a general rule we don't allow arbitrary classes to be passed to components, as this would allow a component to be styled by an application, which breaks our [principle of isolation](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_principles.md#a-component-is-isolated-when).

However, from this example it seems that you might only want to pass existing govuk-frontend column width classes. Would it be possible to rework this PR so that the component adds these styles itself when passed specific options? e.g.

```
head:
- text: Phrase
- text: Rating
  width: 1
```
...and then have the component interpret `1` as `govuk-!-width-one-third` (or similar) and apply the class itself internally?

andysellick: Yes, having a `width` option that gets turned into `one-third` or similar somehow sounds good. Maybe something like this?

```Ruby
# for each column
width ||= nil
sizes = %w[one-third one-quarter]
this_column_class = 'govuk-!-width-#{width}' if [1,2].contains? width
```

csutter: I've played around with it briefly and I think allowing to specify the size as an (allowlisted) string like `width: "three-quarters"` instead of a number feels like a better option.

That way a user of the component doesn't have to try and figure out what `width: 1` or `width: 2` means, but we still guard against use of invalid sizes.

Have updated the PR, let me know if you think this works!

spec/components/table_spec.rb
```diff
@@ -153,4 +153,25 @@ def component_name
...
+        [
+          {
+            text: "January",
+          },
+        ],
+      ],
+    )
+
+    # This needs [class=...] attribute syntax as Nokogiri doesn't like the `!` in the class name
+    assert_select ".govuk-table th[class='govuk-table__header govuk-!-width-one-third']", "Month you apply"
```
andysellick, line 175: I think you can also do the following to match classes with a `!` in them, for information. No need to change this line though.

```
assert_select '.govuk-\!-margin-bottom-6'
```

andysellick (approved):
Looks good 👍

spec/components/table_spec.rb
csutter: Oh cool – I think I _did_ try that but it but it still complained, maybe because it got tripped up by the sibling class 👍

PR #4771, opened 2025-04-25
Update super navigation menu button toggle states

MartinJJones:
## What

- Updated the styles in the super navigation menu to ensure the "Menu" and "Search" toggle buttons use a consistent approach
- Fix an issue where the hover styles would stick when toggling a menu button in the super navigation menu on touchscreen devices
  - https://trello.com/c/6YteIa3f/687-top-nav-hover-state-doesnt-disappear-on-touch-screens

### Search toggle button

The styling of the "Search" button toggle states have been updated to align with the approach used in the "Menu" toggle button: 

- Now uses the `pseudo-underline` mixin instead of applying `border-bottom`
- Styles are now applied on all screen sizes, not just desktop

### Improved documentation and readability
 
- Added/updated comments to clearly indicate the purpose of each section of code
- Update Sass variables to be more descriptive and consistent with existing variable names

## Why

Using the same approach to style button states makes it easier to maintain the code by ensuring styles are applied in a more predictable way.

[Trello card](https://trello.com/c/VVOnzVdk/3420-ensure-nav-menu-button-toggle-states-are-consistent)

### Mobile and Tablet - `:hover` state

The hover state will no long appear on most touchscreen devices such as mobile and tablet, this [fixes an issue](https://trello.com/c/6YteIa3f/687-top-nav-hover-state-doesnt-disappear-on-touch-screens) where the hover styles would stick when toggling a menu button in the super navigation menu.

https://github.com/user-attachments/assets/da712cad-d6b6-44b5-a8b0-114abd18ee04

The hover state will still appear on touch screen devices if the primary input mechanism allows the user to hover elements and is an accurate pointing device as shown in the screenshots below.

| variation | before | after |
| --- | --- | --- |
| default | mobile-default-before | mobile-default-after |
| homepage | mobile-homepage-before | mobile-homepage-after |

Further info on this solution:
- https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/
- https://developer.mozilla.org/en-US/docs/Web/CSS/:hover
- https://developer.mozilla.org/en-US/docs/Web/CSS/@media/pointer
- [Example of a similar approach in the design system](https://github.com/alphagov/govuk-frontend/blob/38c319b1224c34695baebfe669d0b3fac82a0866/packages/govuk-frontend/src/govuk/components/radios/_index.scss#L310-L324)

### All screen sizes - `:hover` state

| before | after |
| --- | --- |
| default-focus-hover-before | default-focus-hover-after |

### High contrast

| variation | before | after |
| --- | --- | --- |
| default | high-contrast-before | high-contrast-after |
| homepage | high-contrast-homepage-before | high-contrast-homepage-after | 

### Grade A
- [x] Chrome 135
- [x] Safari 18.4
- [x] Safari 18.4 iOS 
- [x] Firefox 137

### Grade B
- [x] Chrome 131
- [x] Chrome 133 for Android
- [x] Safari 18.0 iOS

### Grade C
- [x] Chrome 70
- [x] Safari v14.0 - iPhone
- [x] Safari v11.1

### Grade X
- [x] IE11 - JS not supported

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -627,18 +627,21 @@ $after-button-padding-left: govuk-spacing(4);
     border-color: $govuk-focus-colour;
     box-shadow: none;
     z-index: 11;
+
+    &:hover {
+      &::after {
+        background: none;
+      }
+    }
   }
 
-  &:focus:not(:focus-visible) {
+  @include focus-not-focus-visible {
     background: none;
-    border-color: govuk-colour("white");
     box-shadow: none;
     color: govuk-colour("white");
   }
 
   &.gem-c-layout-super-navigation-header__search-toggle-button--blue-background {
-    background: $govuk-brand-colour;
-
     &:hover {
       color: govuk-colour("white");
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -58,6 +58,9 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+// Using `:focus-visible` means that in supporting browsers the focus state won't
+// be visible when a user clicks on the element, but the focus state will still be
+// useful for those who use the keyboard to navigate around the page.
 @mixin focus-and-focus-visible {
   &:focus {
     @content;
@@ -68,6 +71,11 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+// For browsers that don't support `:focus-visible`, this defaults to using
+// `:focus` with a CSS-only fallback strategy.
+//
+// Undoes the :focus styles *only* for browsers that support :focus-visible.
+// See https://www.tpgi.com/focus-visible-and-backwards-compatibility/
 @mixin focus-not-focus-visible {
   & {
     @content;
@@ -465,8 +473,6 @@ $after-button-padding-left: govuk-spacing(4);
     // stylelint-enable max-nesting-depth
   }
 
-  // Undoes the :focus styles *only* for browsers that support :focus-visible.
-  // See https://www.tpgi.com/focus-visible-and-backwards-compatibility/
   @include focus-not-focus-visible {
     background: none;
     box-shadow: none;
@@ -505,6 +511,8 @@ $after-button-padding-left: govuk-spacing(4);
   }
 }
 
+// JS available - targets the "Menu" toggle button
+// Styles the "Menu" open state
 .gem-c-layout-super-navigation-header__navigation-top-toggle-button.gem-c-layout-super-navigation-header__navigation-top-toggle-button--blue-background,
 .gem-c-layout-super-navigation-header__navigation-top-toggle-button {
   // Open button modifier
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -609,6 +609,18 @@ $after-button-padding-left: govuk-spacing(4);
   position: relative;
   width: 51px;
 
+  &::after {
+    @include pseudo-underline($left: 0, $right: 0);
```
jon-kirwan, line 614: One quick note - I noticed that on mobile, when the search button is hovered, the pseudo hover underline stretches all the way to the left and right edges. This looks a bit different from how the Menu button behaves. Might be worth bringing up in a design review, but it seems like the left and right spacing could use a slight increase.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -623,6 +623,26 @@ $after-button-padding-left: govuk-spacing(4);
...
+      &::after {
+        background: none;
+      }
+
+      &:hover {
+        &::after {
+          background: govuk-colour("white");
+        }
+      }
+    }
```
jon-kirwan, line 635: I think this mixin might be generating some duplicate styles:
```
.gem-c-layout-super-navigation-header__search-toggle-button
.gem-c-layout-super-navigation-header__search-toggle-button--blue-background:hover::after {
  background: #fff;
}

.gem-c-layout-super-navigation-header__search-toggle-button
.gem-c-layout-super-navigation-header__search-toggle-button--blue-background:hover::after {
  background: #fff;
}
```

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
MartinJJones: Good spot, there appears to be several ways address this at the moment, including:

- Moving the hover state for the search toggle button back into the `$from: desktop` media query, this would avoid the issue but not really fix the root cause and would make the code more difficult to maintain in the future
- Updating the `$left` and `$right` values passed to the `pseudo-underline` mixin from `0`, easy to update, but overall quite tricky when deciding what the values should be without impacting the design/alignment with the rest of the page, I think we should revisit this in the future however once the variations are consolidated
- Fixing the sticky hover states on touch devices, this would avoid the hover state appearing on touch devices, unless you are using an input, mouse for example:
  - Blog post on how to fix the issue - https://css-tricks.com/solving-sticky-hover-states-with-media-hover-hover/
  - This would also fix the existing Trello card in the backlog - https://trello.com/c/6YteIa3f/687-top-nav-hover-state-doesnt-disappear-on-touch-screens

From the options available, I think it makes sense to go with fixing the sticky hover state as part of this PR, I've added a new commit to fix this issue, what do you think of this approach?

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
jon-kirwan: I agree - the duplicate styles aren’t causing any issues and have a negligible impact on file size, so it makes sense to add this to our backlog and address it at a later stage

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
jon-kirwan: Thanks, @MartinJJones. Just adding a quick note to say I agree with your last option for fixing the sticky hover state on touch devices - plus it’s great that this also addresses a backlog issue as part of the work.

jon-kirwan (approved):
Thanks @MartinJJones. All looks good to me. Nice work 👍

PR #4774, opened 2025-04-25
Move tracker

patrickpatrickpatrick:


andysellick (commented):
I appreciate that you're trying to preserve the history of these files but that will be preserved regardless and this PR is ultimately quite confusing (for example one commit updates a polyfill path, the next commit removes that line entirely).

I'd recommend simply copying the files as they currently are from `finder-frontend` and linking to the original commit history (see [example here](https://github.com/alphagov/frontend/pull/4689/commits/c95c188f410601688b7b3be0ffa459b8231d3ff3)) and then doing additional commits for any changes needed now that the code is in a different app.

patrickpatrickpatrick:
The commits are split for readability and can be squashed when this is merged.

patrickpatrickpatrick:
The support for specifying additional elements could probably reside in the applications that would be using it (as I could override the `getElementInfo` function using `Ga4FinderTracker.getElementInfo = ...`). In that case the only change I would need to make would be that `getElementInfo` needs to have `event` added to it's parameters.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
```diff
@@ -46,76 +46,9 @@
...
-          } else {
-            // Do not track partially filled in fields
-            return null
-          }
-          break
-        }
-      }
-
-      return { elementValue: elementValue, wasFilterRemoved: wasFilterRemoved }
+      return supportedElements[elementType] ? supportedElements[elementType](event.target, event) : { elementValue: '', wasFilterRemoved: false }
```
AshGDS, line 51: Does `event` need to be passed here? Is this just for future elements to have access to it if they need it?

AshGDS (commented):
Generally looks good to me, added one question in the code.

Could you also confirm you've tested this in `finder-frontend` and everything is working as expected? Thanks 👍

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-finder-tracker.js
patrickpatrickpatrick: Yeah, this is for when an `getElementType` function needs to access some information in the event. If you refer to `creates the correct GA4 object for an additionally specified element` in the spec, you can see an example of this.

patrickpatrickpatrick:
> Generally looks good to me, added one question in the code.
> 
> Could you also confirm you've tested this in `finder-frontend` and everything is working as expected? Thanks 👍

Tried it on this branch https://github.com/alphagov/finder-frontend/pull/3734 (which can be merged when this is released) and everything seems to work still.

AshGDS (approved):
LGTM 👍

AshGDS:
Make sure to add a changelog entry!

PR #4779, opened 2025-04-30
Update search component 'on blue' button background colour

jon-kirwan:
## What

Update search component 'on blue' button background colour

**Review URL**: https://components-gem-pr-4766.herokuapp.com/component-guide/feedback/default/preview

## Why

https://trello.com/c/uSBCFXJc/3452-update-search-component-blue-background-variation


MartinJJones (commented):
The change looks good to me, just wanted to check if you think we should do some further refactoring here now or in the future, for example:

- It appears we could update the default background-color for `.gem-c-search--on-govuk-blue` here - https://github.com/alphagov/govuk_publishing_components/blob/1354b9d7a8945033c514552a62ca32f353d21f45/app/assets/stylesheets/govuk_publishing_components/components/_search.scss#L185

This would then likely make the `homepage` option for the search component redundant - https://components.publishing.service.gov.uk/component-guide/search/homepage as it is currently only used to update the submit button color.

jon-kirwan:
> The change looks good to me, just wanted to check if you think we should do some further refactoring here now or in the future, for example:

Good suggestion - just updated it, thanks!

PR #4780, opened 2025-04-30
Allow dependamerger to auto-merge some PRs in the gem

KludgeKML:
- We exclude rails, and most of the JS packages (especially govuk-frontend).

### DIFF
--- a/.govuk_dependabot_merger.yml
+++ b/.govuk_dependabot_merger.yml
@@ -1,0 +1,18 @@
+api_version: 2
+defaults:
+  update_external_dependencies: true
+  auto_merge: true
+  allowed_semver_bumps:
+    - patch
+    - minor
+overrides:
+  - dependency: rails
+    auto_merge: false
+  - dependency: accessible-autocomplete
+    auto_merge: false
+  - dependency: axe-core
+    auto_merge: false
+  - dependency: govuk-frontend
+    auto_merge: false
+  - dependency: govuk-single-consent
+    auto_merge: false


MartinJJones:
Looks good, just a couple of thoughts/questions:

- From looking at the [developer docs](https://docs.publishing.service.gov.uk/manual/test-and-build-a-project-with-github-actions.html#name-of-ci-workflow-and-jobs), do we need to update `name`, from `Continuous integration` to `CI` in [ci.yml](https://github.com/alphagov/govuk_publishing_components/blob/main/.github/workflows/ci.yml#L1)?
 > In contrast, the ci.yml must have a name property called CI, as this is [relied upon by our automated tooling](https://github.com/alphagov/govuk-dependabot-merger/pull/30).

- Do we meet the [Conditions required for automatic patching](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-167-auto-patch-dependencies.md#no-manual-deployment-step) in the gem? for example:

> "MUST only be applied where there is [no manual deployment step](https://github.com/alphagov/govuk-rfcs/blob/main/rfc-167-auto-patch-dependencies.md#no-manual-deployment-step)"

KludgeKML:
@MartinJJones Interesting!

There certainly seems to be precedent for gems: https://github.com/alphagov/govuk-dependabot-merger/blob/main/config/repos_opted_in.yml#L22C3-L22C19 ...gds-api-adapters among others are included in the list. I think I'd argue it's safe because in this case they are themselves part of apps that are manually deployed (so if government frontend auto-deploys with a broken version of this gem, although the downstream question of "which dependency in publishing components caused this" is muddied, the question of "which dependency in government-frontend cause this is clear" (ie it's publishing components). I will seek guidance.

PR #4786, opened 2025-05-06
`populateEcommerceSchema` can use `data-ga4-ecommerce-item-name`

patrickpatrickpatrick:
## What

`populateEcommerceSchema` can populate the `item_name` of the fired event data with a data attribute (`data-ga4-ecommerce-item-name`) assigned to the clicked target or the text of the clicked target if the data attribute is not defined.

## Why

There are instances in which the link being clicked might not have text which describes the item it is associated with. In whitehall and other publishing applications, the search results are in a table and the link can be one of several to be displayed depending on the user and the content. This is also in line with the other attributes of the event data that can be set using data attributes.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
@@ -280,6 +280,7 @@ describe('Google Analytics 4 ecommerce tracking', function () {
       }
 
       resultToBeClicked = document.querySelector("[data-ga4-ecommerce-path='/foreign-travel-advice']")
+      resultToBeClicked.removeAttribute('data-ga4-ecommerce-item-name')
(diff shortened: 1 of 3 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-ecommerce-tracker.spec.js
```diff
@@ -321,6 +322,14 @@ describe('Google Analytics 4 ecommerce tracking', function () {
       resultToBeClicked.click()
       expect(window.dataLayer[3].event).toBe(onSearchResultClickExpected.event)
       expect(window.dataLayer[3].search_results).toEqual(onSearchResultClickExpected.search_results)
+
+      var differentListItemName = 'different list item name'
```
andysellick, line 326: I think it might be worth splitting this out into a separate test, to highlight what this functionality is.

patrickpatrickpatrick: That makes sense, have made this change!

andysellick (approved):
Seems good 👍

PR #4791, opened 2025-05-07
Update logo assets

jon-kirwan:
## What

Update logo assets

## Why

https://trello.com/c/TdJyAPgj/3367-update-gem-assets-containing-a-logo-which-will-not-be-included-in-the-design-system-brand-release


MartinJJones (commented):
The new icons looks great, just one small thing to point out on the [favicon-example.png](https://github.com/alphagov/govuk_publishing_components/pull/4791/files#diff-ce35933c29001c6ef81cf608ffaaa5c1e6f047c302e72f660d6b38fb412f0a6f), the size of the favicons appear to be 32px x 32px, the new icon has a width of 33px and appears to have a transparent background to the right of the icon.

PR #4798, opened 2025-05-09
Stop YouTube links being enhanced when they shouldn't

AshGDS:
## What

- A user reported that a YouTube link they had added to their `govspeak` content was replacing a whole paragraph of their content with our JS YouTube embed, even though they added `data-youtube-player="off"` to the link.
- This has surfaced an issue where paragraphs on pages are being replaced entirely when cookies are disabled.
- There are two reasons for this:
   - Firstly, in general we don't have a check in place to stop paragraphs being replaced when a YouTube link is in a paragraph containing other content. The YouTube link enhancement works by replacing the entire paragraph it's in with the enhancement, but it should only do this if the YouTube link is the only thing in the paragraph. Therefore, there's a risk that bugs (like this one) end up replacing page content.
   - More specifically, currently our cookies rejected "How to watch this video" box is ignoring the `data-youtube-player="off"` attribute and replacing HTML when it shouldn't.
   - These two things combined has led to this bug.

- Therefore, as the user's YouTube link was disabled AND within a paragraph containing other content, our `youtube-link-enhancement` code replaced a whole paragraph of their content with the "How to watch this video" box, hence the user complaint.
- To fix this, I've changed the code so that `data-youtube-player="off"` is checked before rendering the cookies disabled message.
- I've also added a check that ensures our YouTube link enhancement only replaces HTML when the YouTube link is the only thing inside the paragraph.
    - As an aside, you may wonder why we don't just replace the YouTube `<a>` tag with the embed, instead of replacing the parent paragraph. I considered this, but realised that would break content. Editors write sentences like `We have produced an explainer video about the [new project we are working on.]`. If `[new project we are working on]` was a YouTube link and became an embed, the sentence would no longer make sense as it would read as `We have produced an explainer video about the` and then have a YouTube embed.

- A less impactful example of the disabled cookies message showing up when it shouldn't is here: https://www.gov.uk/government/publications/code-compliance-officer-contact-details see how the YouTube video link appears when cookies are accepted vs on incognito mode/rejected cookies.
- Another example is here - they have social media links towards the bottom, which are entirely replaced by the cookies disabled message if your cookies are disabled: https://www.gov.uk/government/news/applications-open-for-fully-funded-opportunities-to-study-in-the-uk

## Why

- Trello card: https://trello.com/c/FRxTPCkl/92-govspeak-youtube-embed-replaces-content-paragraph

## Visual Changes

"How to watch this video" boxes shouldn't exist when `data-youtube-player-off` exists or a YouTube link is within other content. Therefore some content on GOVUK may resurface as it'll no longer be replaced by our YouTube embed code.

New examples have been added to the component guide too.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Stop YT links being enhanced when they shouldn't ([PR #4798](https://github.com/alphagov/govuk_publishing_components/pull/4798))
+
 ## 56.3.0
 
 * Update to LUX 4.1.1 ([PR #4778](https://github.com/alphagov/govuk_publishing_components/pull/4778))
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -23,6 +23,12 @@
     for (var i = 0; i < $youtubeLinks.length; ++i) {
       var $link = $youtubeLinks[i]
       var $linkParent = $link.closest('p')
+
+      // Only replace the <p> with a YT embed if the YT link is the only thing in the <p>.
+      // This prevents other content in the <p> being lost.
+      if ($linkParent.innerHTML !== $link.outerHTML) {
+        continue
+      }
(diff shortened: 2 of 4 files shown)


AshGDS:
@maxgds Thanks for suggesting to add the examples to the component guide, it surfaced another state where the bug occurred that I hadn't accounted for.

maxgds (approved):
Nice work!

PR #4804, opened 2025-05-13
Extra meta tags for `PageViewTracker`

patrickpatrickpatrick:
## What

Add additional meta tags for tracking in `PageViewTracker`.

## Why

These meta tags are used in `page_view` events in publishing applications.


AshGDS (approved):
Looks good, I would add a changelog entry for consistency.

PR #4808, opened 2025-05-14
Fix edge cases with YouTube link enhancement JS

AshGDS:
## What/Why

- There are a couple of edge cases in our YouTube embed code which this PR addresses.
- The first one was the result of something I introduced in https://github.com/alphagov/govuk_publishing_components/pull/4798. In this PR, I made it so that a YouTube link would only enhance into an embed if the YouTube link was the only thing within a `<p>` tag. This was because of a bug on production that was replacing content with YouTube embeds, so this was added as a safety check.
    -  However, there's a use case where editors have a `<p>` that contains a YouTube link and punctuation only. Therefore the existence of the punctuation meant that the safety code was triggerring, and therefore the YouTube link was not turning into an embed. Therefore, I've fixed this by removing punctuation entirely when the string equality check takes place. 
    - Example of the issue in production: 
image

- The second thing I discovered recently was that YouTube playlist links enhance to the "How to watch this YouTube video" message when cookies aren't accepted. This shouldn't be the case, as YouTube playlist links never enhance into an embed. See an example on production here if you disable cookies/go into incognito mode: https://www.gov.uk/government/publications/code-compliance-officer-contact-details - that cookies disabled message ideally shouldn't exist. Therefore I've added a simple check to stop this happening. 
- Trello card: https://trello.com/c/nwnUVPao/618-allow-youtube-embeds-to-also-work-when-theres-punctuation-in-the-same-paragraph-but-outside-the-link
- More context on Slack: https://gds.slack.com/archives/C06TXUDS5G8/p1747230836274189
## Visual Changes

- Pages like this one, where a YT link exists with punctuation, should turn into an embed again: https://www.gov.uk/guidance/download-the-hmrc-app
- I've added another example to the govspeak component, just so it's clear exactly that our YT enhancement works with punctuation.
- Pages like this one shouldn't show a cookies disabled message for YouTube playlist links: https://www.gov.uk/government/publications/code-compliance-officer-contact-details

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -27,8 +27,7 @@
     var $youtubeLinks = this.$element.querySelectorAll('a[href*="youtube.com"], a[href*="youtu.be"]')
     for (var i = 0; i < $youtubeLinks.length; ++i) {
       var $link = $youtubeLinks[i]
-
-      if (this.hasDisabledEmbed($link)) {
+      if (this.hasDisabledEmbed($link) || $link.getAttribute('href').includes('/playlist')) {
         continue
       }
 
--- a/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
@@ -6,6 +6,7 @@
   var YoutubeLinkEnhancement = function ($element, $classOverride) {
     this.$element = $element
     this.$classOverride = typeof $classOverride !== 'undefined' ? $classOverride : 'gem-c-govspeak__youtube-video'
+    this.punctuationRegex = /[\.!\?"']/g // eslint-disable-line no-useless-escape
   }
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
```diff
@@ -102,7 +104,8 @@
 
     // Only replace the <p> with a YT embed if the YT link is the only thing in the <p>.
     // This prevents other content in the <p> being lost.
-    if (parentPara.innerHTML !== $link.outerHTML) {
+    // However, if a <p> exists with only a YT link and punctuation, we do allow the punctuation characters to be lost to the YT embed.
+    if (parentPara.innerHTML.replaceAll(this.punctuationRegex, '') !== $link.outerHTML.replaceAll(this.punctuationRegex, '')) {
```
andysellick, line 107: If this line is duplicated maybe worth splitting off into a separate function?

andysellick (approved):
Looks good, well done for spotting all these complexities. One non-blocking comment.

app/assets/javascripts/govuk_publishing_components/lib/govspeak/youtube-link-enhancement.js
AshGDS: Good point, will move it to a function - thanks!

PR #4813, opened 2025-05-19
Add middleware to optimise GA4 Attributes by reducing unnecessary escaping

KludgeKML:
## What

Optimise GA4 Attributes by removing unnecessary escaping (useful when not using Slimmer). 

## Why

Because the process by which Slimmer stitches together templates from static with views from the apps involves a complete parse of both sets of HTML into a nokogiri tree, then writing out that tree as HTML after manipulation, it optimises GA4 data attributes slightly - replacing the double-quotes around the attribute with single quotes, so that the double quotes in the JSON inside the attribute don't need to be escaped. Without this, the generated HTML can be up to 2 kB larger when Slimmer is not involved.

For apps without slimmer, we can replicate this with a lighter-weight middleware that only needs to be included on non-slimmer apps, using regex and gsub/sub to perform a similar optimisation in place in the response.

https://trello.com/c/0o6io1aA/569-remove-slimmer-feedback

## Visual Changes

No visual changes expected.

### DIFF
--- a/lib/govuk_publishing_components/middleware/ga4_optimise.rb
+++ b/lib/govuk_publishing_components/middleware/ga4_optimise.rb
@@ -1,0 +1,33 @@
+module GovukPublishingComponents
+  module Middleware
+    class Ga4Optimise
+      DATA_GA4_REGEX = /data-ga4-(?:link|event|form|focus-loss)=("[^"]*")/
+
+      def initialize(app)
+        @app = app
+      end
+
+      def call(env)
+        status, headers, response = @app.call(env)
+        return [status, headers, safe_body_from_response(response)] unless headers["content-type"]&.start_with?("text/html")
+
+        body = response.each.first
+        matched_sections = body.scan(DATA_GA4_REGEX)
+
+        if matched_sections.any?
+          matched_sections.each do |section|
+            body.sub!(section.first, section.first.gsub('"', "'").gsub("&quot\;", '"'))
+          end
+        end
+
+        [status, headers, [body]]
+      end
+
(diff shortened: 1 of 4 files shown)


lib/govuk_publishing_components/middleware/ga4_optimise.rb
```diff
@@ -0,0 +1,33 @@
+module GovukPublishingComponents
+  module Middleware
+    class Ga4Optimise
+      DATA_GA4_REGEX = /data-ga4-(?:link|event|form)=("[^"]*")/
```
AshGDS, line 4: We also have `data-ga4-focus-loss` which takes a JSON. See https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/trackers/ga4-focus-loss-tracker.md

lib/govuk_publishing_components/middleware/ga4_optimise.rb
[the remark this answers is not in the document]
KludgeKML: Good question! It might, although that section in the docs is already translated, so I think would emerge unchanged.

More importantly, the middleware only gets invoked if you explicitly add it, so it shouldn't have any effect on existing applications.

lib/govuk_publishing_components/middleware/ga4_optimise.rb
KludgeKML: Thanks! Have added focus-loss to the options in the regex.

AshGDS (approved):
LGTM but probably needs a backend dev to look at it too.

deborahchua (approved):

PR #4824, opened 2025-05-23
Update super nav button styles

MartinJJones:
## What

- Update the search toggle button styles
  - left and right padding increased from `15px` to `20px`
  - the color of the "X" is now govuk-blue the same as the "menu" button in the open state
  - includes the pseudo-underline in the open state
- Fix the `:focus` style in browsers that do not support `:focus-visible`
- Fix styles when JS is not available
  -  only the `:focus` pseudo-class is used on the super navigation menu links
  - `:focus-visible` is now only used when JS is available on the super navigation buttons
- Some refactoring to make it easier to maintain and updates the super navigation menu styles in the future
- The pseudo-underline that appears beneath the menu buttons/links are have been updated to have consistent left and right values 

## Why

The search toggle button styles are now consistent with the "Menu" button and scale with the new menu height

[Trello card](https://trello.com/c/aIixDpBb)

### Search toggle button

| Before | After |
| --- | --- |
| search-button-open-before | search-button-open-after |

### `:focus` and `:hover` state
This styling appears in browsers that do not support `:focus-visible`

| Before | After |
| --- | --- |
| focus-hover-before | focus-hover-after |

### Menu underline

| Before | After |
| --- | --- |
| menu-underline-before | menu-underline-after |

### Search underline

| Before | After |
| --- | --- |
| search-underline-before |  search-underline-after|

### Grade A
- [x] Chrome 137
- [x] Safari 18.5
- [x] Safari 18.5 iOS 
- [x] Firefox 139

### Grade B
- [x] Chrome 132
- [x] Chrome 133 for Android
- [x] Safari 18.3 iOS

### Grade C
- [x] Chrome 70
- [x] Safari v14.0 - iPhone
- [x] Safari v11.1

### Grade X
- [x] IE11 - JS not supported

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -49,7 +49,7 @@ $search-icon-height: 20px;
   top: 0;
 }
 
-@mixin pseudo-underline($left: govuk-spacing(4), $right: govuk-spacing(4), $width: false) {
+@mixin pseudo-underline($left: govuk-spacing(2), $right: govuk-spacing(2), $width: false) {
   background: none;
   content: "";
   height: $pseudo-underline-height;
@@ -199,7 +199,7 @@ $search-icon-height: 20px;
 
     &::after {
       @include make-selectable-area-bigger;
-      @include pseudo-underline($left: $after-link-padding, $right: $after-link-padding);
+      @include pseudo-underline;
     }
 
     &::before {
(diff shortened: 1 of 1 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -16,7 +16,7 @@ $after-link-padding: govuk-spacing(4);
 $after-button-padding-right: govuk-spacing(4);
 $after-button-padding-left: govuk-spacing(4);
 
-$search-button-padding-top-bottom: govuk-spacing(3);
+$search-button-padding-top-bottom: govuk-spacing(4);
```
jon-kirwan, line 19: The padding now seems to be the same on all sides - top, right, bottom, and left. Should we update the variable name to `search-button-padding`?

MartinJJones: Good spot, I've updated the variable name to `search-button-padding` as suggested, thanks 👍

jon-kirwan (approved):
Nice work! I've reviewed the commits and tested across various browsers and devices and everything looks good and works well.

PR #4825, opened 2025-05-23
Replace use of tint mixins with hex variables

jon-kirwan:
## What

Replace use of tint mixins with hex variables

## Why

Dart Sass 1.79.0 introduced support for CSS Color Level 4 color spaces. Consequently, the `govuk-tint` mixin now outputs rgb values with floating-point precision, leading to unsupported property value warnings in some browsers, such as Safari 11.1.1. To maintain compatibility, we're defining equivalent hex colour values as variables instead.

Screenshot 2025-05-14 at 14 08 19

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
@@ -26,7 +26,7 @@
 }
 
 .gem-c-feedback__prompt {
-  background-color: govuk-tint(govuk-colour("blue"), 95%);
+  background-color: $govuk-rebrand-template-background-colour;
   color: govuk-colour("black");
   border-top: 1px solid govuk-tint(govuk-colour("blue"), 50%);
   outline: 0;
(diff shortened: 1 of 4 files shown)


MartinJJones (approved):
Nice work on finding this issue and investigating it further, the fix looks really good to me 👍 

Side note, should we create an issue in the govuk_publishing_components gem relating to the Dart Sass compilation?, similar to https://github.com/alphagov/govuk-frontend/issues/5988

jon-kirwan:
> Nice work on finding this issue and investigating it further, the fix looks really good to me 👍
> 
> Side note, should we create an issue in the govuk_publishing_components gem relating to the Dart Sass compilation?, similar to [alphagov/govuk-frontend#5988](https://github.com/alphagov/govuk-frontend/issues/5988)

Thanks, @MartinJJones. Good idea and I'll link it with the issue in GOV.UK Frontend.

jon-kirwan:
https://github.com/alphagov/govuk_publishing_components/issues/4830

PR #4840, opened 2025-06-03
Add phone numbers to GA4 PII redaction

AshGDS:
## What / Why

- Adds phone numbers to our GA4 PII redaction. 
- This includes four different regexes:
  - UK Mobile numbers `e.g. 07...`
  - UK Mobile numbers with the international code `+44`, which causes the first `0` in the mobile number to be dropped (`e.g. +447...`)
  - UK Landline numbers  `e.g. 020...`
  - UK Landline numbers with the international code `+44`, which causes the first `0` in the phone number to be dropped (`e.g. +4420...`)
  - I have these separate as I found combining the UK number/international regexes together led to an incorrect scenario where `+440...` is valid when it shouldn't be, so it was easier to separate them out.
  - Trello card: https://trello.com/c/V7q79XLH

## Visual Changes

None.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
@@ -320,6 +320,114 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
     })
   })
 
+  function generatePhoneNumber (prefix, amount, suffix = '') {
+    var phoneNumber = prefix
+    for (var i = 0; i < amount; i++) {
+      phoneNumber += Math.floor(Math.random() * 10)
+    }
+    return `${phoneNumber}${suffix}`
+  }
+
+  describe('phone numbers', function () {
+    it('are stripped if they are valid UK mobile numbers', function () {
+      // Test 50 random phone numbers so we aren't storing any numbers in code.
+      for (let i = 0; i < 50; i++) {
+        const phoneNumber = generatePhoneNumber('07', 9)
+        const redactedNumber = pii.stripPIIWithOverride(phoneNumber)
+        expect(redactedNumber).toEqual('[phone number]')
+      }
+    })
+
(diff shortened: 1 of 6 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -320,6 +320,110 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
     })
   })
 
+  function generatePhoneNumber (prefix, amount, suffix = '') {
+    var phoneNumber = prefix
+    for (var i = 0; i < amount; i++) {
+      phoneNumber += Math.floor(Math.random() * 10)
+    }
+    return phoneNumber + suffix
```
andysellick, line 328: I have a slight nervousness around using `+` here. You're obviously passing all strings in these tests so the result gets cast to a string, which is fine, but maybe it would be worth explicitly making sure this is a string to help future tests from hitting an issue? Something like `return "${phoneNumber}${suffix}"`?

spec/javascripts/govuk_publishing_components/analytics-ga4/pii-remover.spec.js
```diff
@@ -320,6 +320,110 @@ describe('GOVUK.analyticsGa4.PIIRemover', function () {
...
+        for (const number of numbers) {
+          const redactedNumber = pii.stripPIIWithOverride(number)
+          expect(redactedNumber).toEqual('[phone number]')
+        }
+      }
+    })
+
+    it('are not stripped if they are invalid UK numbers', function () {
+      const numbers = [
+        generatePhoneNumber('07', 7),
```
andysellick, line 384: Might be worth documenting the details of these numbers like you have with the other tests. Presumably any number starting with `07` followed by 7 digits is invalid?

Side thought - is it possible the rules for phone numbers could change in future?

andysellick (approved):

PR #4842, opened 2025-06-04
Ensure logo and menu items do not overlap on small screens

MartinJJones:
## What

Ensure logo and menu items do not overlap on small screens:
- By default the nav menu buttons will have 15px of padding to the left and right, this will increase to 20px on screens with a min width of 360px, the same width as when the chevron is displayed
- Added new variables for nav button padding
- Removed unused `$after-button-padding-` variables

## Why

The menu should not overlap with the GOVUK logo - [Trello card](https://trello.com/c/qZhAzpax)


jon-kirwan (approved):
Everything looks good to me and I’ve approved it. I added a minor suggestion, but feel free to skip it if you don’t think it's needed. Also, there might be an opportunity to use a mixin to simplify things further, but that can be looked at later on. Thanks!

PR #4847, opened 2025-06-09
Add extra functionality to `trackFormSubmit` of `Ga4FormTracker`

patrickpatrickpatrick:
## What

New functionality and adjustments to the `Ga4FormTracker` submit tracking

### More details
- Change the `Ga4FormTracker` spec to be more DRY and easier to add tests to
- Support for adding labels to the `text` of the event fired (if `data-ga4-form-record-json` set), updated docs to reflect change
- Option to use character count instead of `[REDACTED]` for text fields (if `data-ga4-form-use-text-count` set), updated docs to reflect change
- Takes into account conditional fields (if conditional checkbox not checked, don't include conditional fields)

## Why

As part of adding tracking to Publishing Applications. Forms are more complex than those on GOV.UK and Smart Answers, so we need additional information included on a form submit to be able to get useful insights.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
@@ -9,6 +9,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.trackingTrigger = 'data-ga4-form' // elements with this attribute get tracked
     this.includeTextInputValues = this.module.hasAttribute('data-ga4-form-include-text')
     this.recordJson = this.module.hasAttribute('data-ga4-form-record-json')
+    this.useTextCount = this.module.hasAttribute('data-ga4-form-use-text-count')
     this.redacted = false
     this.useFallbackValue = this.module.hasAttribute('data-ga4-form-no-answer-undefined') ? undefined : 'No answer given'
   }
@@ -87,6 +88,8 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
 
       input.section = labelText
 
+      var isTextField = inputTypes.indexOf(inputType) !== -1 || inputNodename === 'TEXTAREA'
+
       if (elem.checked) {
         input.answer = labelText
 
@@ -101,10 +104,14 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       } else if (inputNodename === 'SELECT' && elem.options[elem.selectedIndex] && elem.options[elem.selectedIndex].value) {
         input.answer = elem.options[elem.selectedIndex].text
-      } else if (inputTypes.indexOf(inputType) !== -1 && elem.value) {
+      } else if (isTextField && elem.value) {
         if (this.includeTextInputValues || elem.hasAttribute('data-ga4-form-include-input')) {
-          var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
-          input.answer = PIIRemover.stripPIIWithOverride(elem.value, true, true)
+          if (this.useTextCount) {
+            input.answer = elem.value.length
+          } else {
+            var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
+            input.answer = PIIRemover.stripPIIWithOverride(elem.value, true, true)
+          }
         } else {
           // if recording JSON and text input not allowed
           // set the specific answer to '[REDACTED]'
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -84,19 +85,33 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+        }
       } else if (inputNodename === 'SELECT' && elem.options[elem.selectedIndex] && elem.options[elem.selectedIndex].value) {
         input.answer = elem.options[elem.selectedIndex].text
       } else if (inputTypes.indexOf(inputType) !== -1 && elem.value) {
         if (this.includeTextInputValues || elem.hasAttribute('data-ga4-form-include-input')) {
           var PIIRemover = new window.GOVUK.analyticsGa4.PIIRemover()
           input.answer = PIIRemover.stripPIIWithOverride(elem.value, true, true)
         } else {
-          this.redacted = true
+          if (this.recordJson) {
```
patrickpatrickpatrick, line 141: potentially add comment explaning

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -89,9 +89,34 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
+            input.section = conditionalCheckboxLabel.innerText + ' ' + labelText
+          }
+        }
+      }
 
       if (elem.checked) {
-        input.answer = labelText
+        if (elem.hasAttribute('aria-controls')) {
+          inputs.splice(i, 1)
+          continue
```
andysellick, line 116: What does `continue` do here? I can't see a loop.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -8,6 +8,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
     this.module = module
     this.trackingTrigger = 'data-ga4-form' // elements with this attribute get tracked
     this.includeTextInputValues = this.module.hasAttribute('data-ga4-form-include-text')
+    this.recordJson = this.module.hasAttribute('data-ga4-form-record-json')
```
andysellick, line 11: I'd like to understand more about this change - specifically how the new output differs from the old and what the existing code lacked. Also if this is useful is it not worth making it the default behaviour, was this discussed with the other PAs?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
```diff
@@ -89,9 +89,34 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
       input.section = labelText
 
       var isTextField = inputTypes.indexOf(inputType) !== -1 || inputNodename === 'TEXTAREA'
+      var conditionalField = elem.closest('.govuk-checkboxes__conditional')
```
andysellick, line 92: Just checking, are you aware of the `data-ga4-form-include-input` option on the form tracker, that can be applied to inputs (like the conditional ones) that should be ignored?

patrickpatrickpatrick: I thought that this marked an element to be exempt from redaction rather than it being ignored?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
andysellick: That's a really helpful summary, thanks. Could you add that into the docs?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
andysellick: Ah yes you're right.

patrickpatrickpatrick: I'm pretty keen to rewrite this tracker using more contemporary JS, as myself and other frontend developers that I talked through this PR with on a call found the existing code difficult to parse (such as it using `for` and not `forEach`). Since this PR is just to add new functionality, I thought that changing the style was out of scope.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
patrickpatrickpatrick: I've added that to docs in the commit that adds this change 🙂

andysellick (approved):

PR #4848, opened 2025-06-09
Use helpers in Jasmine specs

patrickpatrickpatrick:
## What

Move repeatedly defined helper functions and variables to `SpecHelper.js`

### More details

- Assign `agreeToCookies` to `this` in the Jasmine context, remove all definitions of `agreeToCookies` in individual specs
- Assign `denyCookies` to `this` in the Jasmine context, remove all definitions of `denyCookies` in individual specs 
- Use the `gem_version` that's assigned in the `beforeAll` of `SpecHelper.js`
  - Set `window.GOVUK.analyticsGa4.vars.gem_version` to `aGemVersion` (which is used in all specs) in `beforeAll` 
  - Update `scroll-tracker.spec.js` to use `aGemVersion` (only test that uses `gem-version` for `gem_version`)
  - Remove stubs or assignments to `window.GOVUK.analyticsGa4.vars.gem_version` in specs
    - Not needed because `window.GOVUK.analyticsGa4.vars.gem_version` is set before each test

## Why

Easy to maintain, write new tests and read if we use shared functions and variables.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-auto-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-auto-tracker.spec.js
@@ -5,25 +5,16 @@ describe('Google Analytics auto tracker', function () {
   var element
   var expected
 
-  function agreeToCookies () {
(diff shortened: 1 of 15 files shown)


andysellick (approved):
Nice improvement 👏

PR #4855, opened 2025-06-11
Make links relative instead of absolute

catalinailie:
Because the CSV preview used to be served from the asset domain, a few links on that page were made absolute in order to work (otherwise it would link to assets.publishing.service.gov.uk/help/cookies for example, which is incorrect).

Now the preview url is served from the [gov.uk](http://gov.uk/) domain, so we can make those links relative again and be consistent with the rest of the website.

Trello card: https://trello.com/c/7NZ39TdM/3525-remove-absolute-links-from-the-csv-preview-pages

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased changes
 
+* Make global links relative instead of absolute ([PR #4855](https://github.com/alphagov/govuk_publishing_components/pull/4855))
 * Add Kyrgyz translations ([PR #4875](https://github.com/alphagov/govuk_publishing_components/pull/4875))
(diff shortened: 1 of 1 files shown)


jon-kirwan (commented):
Thanks Catalina, the changes look good and work as expected for me!

Just one small thing. Could you please update the Changelog to add in a line about the PR like so https://github.com/alphagov/govuk_publishing_components/commit/1fb9ebafeff016f014570fc231ec26cced7132be#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4ed

Or

> ## Unreleased
* Make links relative instead of absolute ([PR #4855](https://github.com/alphagov/govuk_publishing_components/pull/4855))

catalinailie:
Yes, absolutely. I forgot to do that.

PR #4888, opened 2025-06-23
Add extra options to `AddAnother` component tracking

patrickpatrickpatrick:
## What

Adds `index-start` and `index-section-count` options to the `AddAnother` tracking.

## Why

Allows overriding the default indexing behaviour, if the indexing should be a continuation of surrounding content of the page.


AshGDS (approved):
LGTM - small question, could `index_section_count` ever be set to 0? If so I'm guessing it wouldn't work here as 0 would be falsy.

patrickpatrickpatrick:
> LGTM - small question, could `index_section_count` ever be set to 0? If so I'm guessing it wouldn't work here as 0 would be falsy.

If it was 0 then yeah it would be falsy but your comment made me realise, as I wasn't casting `this.module.dataset.ga4IndexSectionCount` to a string you could use an arbitrary string as an index section count. So I've made that change to prevent this happening!

PR #4893, opened 2025-06-25
Add extra attributes to `Schema`

patrickpatrickpatrick:
## What

Adds `content_id` and `user_id` to `Schema`

## Why

Requested changes from the publishing PAs, track `content_id` when content is being edited (URL is too consistent to use for this) and `user_id`.


patrickpatrickpatrick:
> @patrickpatrickpatrick not sure if you know but we have some analytics documentation on the dev docs, which we're supposed to keep up to date. Would you be able to add this change into that please? The schema stuff is on this page: https://docs.publishing.service.gov.uk/analytics/schema.html and happy to explain how the rest of it fits together if you want help.

I saw this and also opened this PR https://github.com/alphagov/govuk-developer-docs/pull/5147, was going to wait before merging it

PR #4906, opened 2025-07-02
Reduce specificity of `.email-url-number` rules

pezholio:
At the moment, Content Blocks that render individual telephone objects are still showing with the label and phone number adjacent to each other as they’re not wrapped in a `.content` div. This reduces the specificity of the selector, so the phone numbers appear correctly.

There should be no difference in the Govspeak output as it stands, but just for visibility, I've added some screenshots:

## Before

image

## After

image


MartinJJones (commented):
I think I understand the changes being made here, but just to help me review it, as Percy has not picked up an visual regressions that I can look at, please can you include a before and after image of the visual changes, and if a possible a page example, thanks

pezholio:
@MartinJJones Cool - have added some screenshots - there should be no difference in the out of the box rendering. This is just allowing us to apply contact styling without having to have multiple divs. Here's an example from Content Block Manager:

## Before

image

## After

image

PR #4910, opened 2025-07-08
Allow skip_account to act based on boolean and string

andysellick:
## What / why
- currently skip_account is only true if a string value of "true" is passed to the component, rather than a boolean
- the expected situation would be that a boolean should be accepted by the component rather than a string
- since this is already happening in a few places, modify the code to temporarily accept either, until we can unpick the instances where a string is currently being passed
- also add a spec test for this helper, as there wasn't one

## Visual Changes
None.

Trello card: https://trello.com/c/LTS5YNJS/725-fix-skipaccount-handling-in-components-gem

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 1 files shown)


AshGDS (approved):

KludgeKML (approved):
LGTM!

Issue #4611, opened 2025-02-05
Component guide has tag with poor contrast


andysellick:
There's a tag in the header of https://components.publishing.service.gov.uk/component-guide using black text on a purple background...

![Image](https://github.com/user-attachments/assets/b65c83e7-ae4b-4ece-bb78-f17bf20e2c86)

Haven't checked but fairly sure this fails contrast requirements. Might be worth checking https://components.publishing.service.gov.uk/component-guide/layout_header to see if this example should be included in that page.

Might also be worth checking the [other component guides](https://docs.publishing.service.gov.uk/manual/components.html#component-guides) for this problem too.

jon-kirwan:
It might also be a good opportunity to update the tag styles to align with those in GOV.UK Frontend https://github.com/alphagov/govuk-frontend/pull/3502 and https://github.com/alphagov/govuk-frontend/releases/tag/v5.0.0.

andysellick:
Thanks @jon-kirwan that's a good point. Also I've wondered for a while whether we should implement the tag component from the Design System, but we've never seemed to use it enough to justify it. Maybe we should?

jon-kirwan:
Yep, I had the same thought. It appears to be used in several publishing apps but isn't very common elsewhere. Given that the code is relatively small, the benefits of turning it into a component might be small e.g.
```
<strong class="govuk-tag govuk-tag--grey">
  Inactive
</strong>
```
However, it could make future updates easier if any part of the markup needs to change.

PR #4914, opened 2025-07-10
**BREAKING** Introduce tag component

AshGDS:
## What / Why

- Introduces a new `Tag` component based on [the Design System component](https://design-system.service.gov.uk/components/tag/)
- Creating a tag component gives us a few benefits:
  - The default tag component can now be used elsewhere - the previous one was specifically styled for the header
  - Individual component CSS will now be used, so the tag CSS won't be downloaded where it isn't needed
  - The component has validation now and uses the component wrapper
## Breaking change
- I've put this as a breaking change as:
  - apps may need to import `@import 'govuk_publishing_components/components/tag';`
  -  the app `content-data-admin` seems to have some custom override styles for `.gem-c-environment-tag`, which no longer exists as that was a class tied to the old tag.
- Trello card: https://trello.com/c/PfnvvqX0/727-fix-our-component-guide-using-a-tag-with-poor-colour-contrast

## The tag in our header
This PR replaces the existing "environment" tag in our `layout_header` with the new component, which will fix https://github.com/alphagov/govuk_publishing_components/issues/4611 

The tag in the header now sits outside of the header `<a>` link. This is because the design system guidance states that tags should not be used as a link/an interactive thing.

As well as this, our header logo div uses `whitespace: no-wrap` which is a problem. This is because it causes the environment tag to not be calculated as part of the element's width. Therefore the environment tag doesn't wrap onto a new line currently across the `layout_header`, and leads to this visual bug that currently exists on production:

**This is now outdated - see PR discussion**
image

Therefore to solve this, I've added a modifier class that makes the header logo section use `display: flex` if an environment tag exists. This means that the environment tag will sit on a new line on desktop now and won't ever overlap a menu.

**This is now outdated - see PR discussion**
image

I think to fix this using other means would require refactoring the `layout_header` component, because:
- The service app name can be a varying length, and the menu items can be of a varying length, so it's hard to come up with a fix that adheres to those requirements without a large refactor
- the header is using `float: left` and `float: right` for layout, which is making things more difficult to fix (as opposed to it using `flex` or `grid`). It's probably worth raising a new issue to refactor `layout_header` and update it to use `flex` or `grid` instead.

## Visual Changes (OUTDATED - See PR discussion)

See https://components-gem-pr-4914.herokuapp.com/component-guide/tag for all examples

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image | 
| Production new red example image | Production new red example image |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/_all_components.scss
@@ -91,6 +91,7 @@
 @import "components/summary-list";
 @import "components/tabs";
 @import "components/table";
+@import "components/tag";
 @import "components/textarea";
 @import "components/translation-nav";
 @import "components/warning-text";
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
@@ -2,24 +2,6 @@
 @import "govuk_publishing_components/components/search";
 @import "govuk_publishing_components/components/skip-link";
 @import "govuk/components/header/header";
-@import "govuk/components/tag/tag";
-
-.gem-c-layout-header--production .govuk-header__container {
-  border-bottom-color: govuk-colour("red");
-}
-
-.gem-c-layout-header--integration .govuk-header__container,
-.gem-c-layout-header--staging .govuk-header__container {
-  border-bottom-color: govuk-colour("yellow");
-}
-
-.gem-c-layout-header--example .govuk-header__container {
-  border-bottom-color: govuk-colour("bright-purple");
-}
-
-.gem-c-layout-header--development .govuk-header__container {
-  border-bottom-color: govuk-colour("dark-grey");
-}
 
 .gem-c-layout-header--no-bottom-border,
 .gem-c-layout-header--no-bottom-border .govuk-header__container {
@@ -76,6 +58,11 @@
   }
 }
 
+.gem-c-header__environment {
+  padding-top: 2px;
+  padding-bottom: 3px;
+}
+
 .gem-c-header__logo {
   @include govuk-media-query($from: desktop) {
     white-space: nowrap;
@@ -96,34 +83,6 @@
   @include govuk-media-query($from: tablet) {
     display: inline-block;
   }
-}
-
-.gem-c-environment-tag {
-  padding: 2px 5px 0;
-  margin-left: 3px;
-  margin-top: govuk-spacing(-2);
-  vertical-align: middle;
-  @include govuk-font($size: 16, $weight: "bold");
-}
-
-.gem-c-environment-tag--production {
-  background-color: govuk-colour("red");
-  color: govuk-colour("white");
-}
-
-.gem-c-environment-tag--example {
-  background-color: govuk-colour("bright-purple");
-}
-
-.gem-c-environment-tag--development {
-  background-color: govuk-colour("dark-grey");
-  color: govuk-colour("white");
-}
-
-.gem-c-environment-tag--staging,
-.gem-c-environment-tag--integration {
-  background-color: govuk-colour("yellow");
-  color: govuk-colour("black");
 }
 
 .govuk-header__navigation-item--collapsed-menu-only {
@@ -155,8 +114,7 @@
       padding: 0 2mm 2mm 0;
     }
 
-    .gem-c-header__product-name,
-    .gem-c-environment-tag {
+    .gem-c-header__product-name {
       color: $govuk-print-text-colour;
       background: none;
     }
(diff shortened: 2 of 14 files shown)


AshGDS:
Thanks for reviewing @andysellick 👍  Good idea with using `absolute` positioning, I can look into it tomorrow but I'm assuming it would be an issue in situations like this: 
image

image

andysellick:
@AshGDS yeah that could be tricky. Well, just a thought.

AshGDS:
@andysellick Did a bit of investigation and I can get your suggestion working on desktop/mobile by adding extra margin bottom to the whole header so that there's space for the tag to sit. Having some issues with tablet though.

## Desktop
image

## Mobile

image


## Tablet

Tablet is potentially a problem though because the height of the header becomes dependent on the length of the "product name". Therefore we can't confidently position the tag as every app is going to have a different length product name and therefore there will be differences in header height. Unless we move the tag HTML outside of the header logo area? 

image

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
andysellick, line 1: Random thing I've just noticed... this file already imports the tag scss from `govuk-frontend` for a legacy reason (line 5) so the styles are duplicated now.

app/assets/stylesheets/govuk_publishing_components/_all_components.scss
```diff
@@ -91,6 +91,7 @@
 @import "components/summary-list";
 @import "components/tabs";
 @import "components/table";
+@import "components/tag";
```
andysellick, line 94: I don't know why, but tags have a maximum width of 160px set, which is noticeable if you display a tag in Firefox and use 'zoom text only' (or just increase the default font size in your browser). Fixable with `max-width: none` in the gem's tag scss.

Screenshot 2025-07-15 at 11 29 38

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -82,6 +82,13 @@
   }
 }
 
+.gem-c-header__logo--environment {
```
andysellick, line 85: Feel like I've opened a can of worms on this aspect, apologies. Here's a suggestion (that still needs refinement, unfortunately)

```
.gem-c-header__logo--environment {
  padding-top: 3em !important;

  .govuk-tag {
    position: absolute;
    top: 0;
    left: 0;
  }
}
```

Looks like this:

Screenshot 2025-07-15 at 11 32 48

andysellick (commented):
Sorry, just spotted a few things.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
andysellick: Your suggestion of moving the tag outside of the header makes sense the more I think about it. We could just put it in a separate div above the rest of the header, and then wouldn't need to worry about spacing or layout because it would set its own height and not collide with anything else. It also simplifies the code (don't need that extra class). Layout header partial looks something like this:

```
<% if environment_exists %>
  <%= render "govuk_publishing_components/components/tag", { text: environment, header: true, colour: colour_map[environment.downcase] } %>
<% end %>
<div class="govuk-header__logo gem-c-header__logo <%= "gem-c-header__logo--environment" if environment_exists %>">
  <a href="<%= logo_link %>" class="govuk-header__link govuk-header__link--homepage">
    <%= render "govuk_publishing_components/components/govuk_logo/govuk_logo" %>
    <% if product_name %>
      <span class="govuk-header__product-name gem-c-header__product-name">
        <%= product_name %>
      </span>
    <% end %>
  </a>
</div>
```

Visual:

Screenshot 2025-07-15 at 11 38 39

I get a slightly weird overlap with the element below when zoomed, but this is probably easily fixed:

Screenshot 2025-07-15 at 11 40 26

andysellick: That overlap ☝️ might be because the tag styles have a margin top of -2px and a margin-bottom of -3px. I'm sure there's a reason for this but it's obviously causing a slight problem here.

app/assets/stylesheets/govuk_publishing_components/components/_tag.scss
```diff
@@ -0,0 +1,9 @@
+@import "govuk_publishing_components/individual_component_support";
+@import "govuk/components/tag/tag";
+
+@include govuk-media-query($media-type: print) {
```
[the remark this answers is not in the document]
AshGDS: Thanks, done 👍

AshGDS:
Thanks @andysellick - should be ready for another review 👍

PR #4921, opened 2025-07-15
[WHIT-2297] [WIP] Fix redundant first option in `select-with-search` component

matthillco:
## What
When using the `select-with-search` component, hide the redundant first option in the drop-down.

## Why
The `select-with-search` component is used to enhance `<select>` components with a search field and customised visuals. This is enabled by a third party JavaScript library: `choices.js`. This library interrogates a designated `<select>` and then hides it. It then uses information from the hidden `<select>` to build a new progressively enhanced version, using new markup and new behaviour. 

This component previously used a custom `placeholder` element to display a message to the user:
- "Select one" was displayed for the 'single' variant (the user can select only one option)
- "Select all that apply" was displayed for the 'multiple' variant (the user can select many options)

image
image

The `select-with-search` component accepts a parameter, `include_blank: true` which adds an empty first option to the original `<select>`. This is copied through into the progressively enhanced version. 

Why would we want a blank first option for a `<select>`? 

Well, the `<select>` element will automatically default to selecting the first option in the list unless you tell it otherwise. But sometimes we don't want that behaviour: we want the user to actively choose an option before they can continue. In these cases, by sending a blank item as the first item, the blank item is selected and the other options -- the actual user choices -- remain unselected. This forces the user to choose one, and not rely on a default pre-selected option.

But there was a problem: the first blank option was inheriting the instruction text for the placeholder. This meant there was a placeholder AND a select option both with the text content "Select One". It was confusing:

image

Not only was this confusing visually, it created a poor screen-reader experience where the announcer included the phrase "Select one, select one' in the list of interactive elements.

This empty element therefore must be hidden from the visual and assistive element tree. It proved unworkable to remove it using JS due to various constraints of the `choices.js` library. After a lot of experimentation, I found a solution using a few JS changes and a CSS tweak that seems robust (see the next section).

## How

This has now been fixed by:
- Deleting the previous code that assigned the "Select One / Select all that apply" labels to the empty option in the `<select>`.
- The placeholder label is now applied by choices.js using the `placeholderValue` option.
- This now means the empty first `<option>` remains empty and when choices.js replicates the functionality in the new markup, the new first option also remains empty. But now we have a visible empty option which looks strange and makes no sense to the user.
- So we then hide the visible empty `<option>` with CSS. We don't want to completely remove it otherwise we'll get back the old behaviour of the first option in the list being selected by default. 
- And as screen-readers don't announce hidden elements, we also solve the long standing accessibility issue of "Select One / All" being read out twice.

## Visual Changes

Note that the before/after for the _mutiple_ variant is the same: this variant always highlights the first option as it's waiting for the user to select one or many options.

| Before | After |
| --- | ----------- |
| image | image |
| image | image |
| image | image |


matthillco:
This fix was rejected in favour of https://github.com/alphagov/govuk_publishing_components/pull/5251.

PR #4927, opened 2025-07-17
Change skip_account to only accept a boolean

andysellick:
## What / why
- following on from the work in https://github.com/alphagov/govuk_publishing_components/pull/4910 to allow the single page notification button component to allow a boolean true as well as a string true, now remove the ability for it to recognise a string true for the skip_account option
- skip_account should only be a boolean true, not a string "true"

## Visual Changes
None.

Trello card: https://trello.com/c/LTS5YNJS/725-fix-skipaccount-handling-in-components-gem

### DIFF
--- a/spec/lib/govuk_publishing_components/presenters/single_page_notification_button_helper_spec.rb
+++ b/spec/lib/govuk_publishing_components/presenters/single_page_notification_button_helper_spec.rb
@@ -52,11 +52,6 @@ RSpec.describe GovukPublishingComponents::Presenters::SinglePageNotificationButt
       expect(single_helper.skip_account_param).to eq("single_page_subscription")
     end
 
-    it "sets skip_account when skip_account is a string value of true" do
-      single_helper = described_class.new(skip_account: "true")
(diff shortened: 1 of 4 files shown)


spec/lib/govuk_publishing_components/presenters/single_page_notification_button_helper_spec.rb
```diff
@@ -52,9 +52,9 @@
       expect(single_helper.skip_account_param).to eq("single_page_subscription")
     end
 
-    it "sets skip_account when skip_account is a string value of true" do
+    it "does not set skip_account when skip_account is a string value of true" do
```
KludgeKML, line 55: I feel like you can probably take this test out? I'm not sure it adds anything to the coverage or the understanding of the component's behaviour.

andysellick: Fair enough, have removed 👍

KludgeKML (approved):
LGTM!

PR #4935, opened 2025-07-21
Add `user-id` to `page_view` tracked event

patrickpatrickpatrick:
## What

Add `user-id` to `page_view` tracked event.

## Why

Requested change from Publishing PA team. Addresses WHIT-2306.


matthillco (approved):
Looks good, ship it :)

PR #4936, opened 2025-07-21
[WHIT-2318] Fix opacity issue on first selected item in choices.js dropdown

matthillco:
## What
Fix opacity issue on first selected item in choices.js dropdown.

Note that there are some accessibility issues with the repetition of 'Select One' which we will be addressing in a future update. A previous PR attempted an unsuccessful fix: https://github.com/alphagov/govuk_publishing_components/pull/4921

## Why
The first item in a "select one" variant of the `select-with-search` component is ghosted out due to inheriting the 50% opacity of the placeholder item.

## Visual Changes
| Before | After |
| --- | ----------- |
|  |  |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Change skip_account to only accept a boolean ([PR #](https://github.com/alphagov/govuk_publishing_components/pull/4927))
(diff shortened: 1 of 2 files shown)


patrickpatrickpatrick (approved):
LGTM! Might be worth adding a changelog entry too!

PR #4937, opened 2025-07-22
Add spaces to phone number PII redaction

AshGDS:
## What / Why

- Adds optional spaces to the regex that is detecting phone numbers in the GA4 PII
- On the [trello card](https://trello.com/c/V7q79XLH) the initial examples provided did not imply they wanted spaces redacted. However when they tested the change they were using spaces in the phone numbers. Therefore we need to add spaces to meet their requirements.
- It was easier to create new regex lines as trying to combine these extra cases into a fancy one liner caused the tests to fail and caused false positives in a way that was hard to diagnose.
- I had to make the phone number generation in the tests a bit more complex as well to account for spaces being in the test cases.

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add spaces to phone number PII redaction ([PR #4937](https://github.com/alphagov/govuk_publishing_components/pull/4937))
+
 ## 59.1.0
 
 * Change skip_account to only accept a boolean ([PR #4927](https://github.com/alphagov/govuk_publishing_components/pull/4927))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@
 * Change skip_account to only accept a boolean ([PR #](https://github.com/alphagov/govuk_publishing_components/pull/4927))
 * Fix opacity issue on first selected item in choices.js dropdown ([PR #4936](https://github.com/alphagov/govuk_publishing_components/pull/4936))
 * Add `user-id` to `page_view` tracked event ([PR #4935](https://github.com/alphagov/govuk_publishing_components/pull/4935))
+* Add spaces to phone number PII redaction ([PR #4937](https://github.com/alphagov/govuk_publishing_components/pull/4937))
```
andysellick, line 15: I think the changelog needs a rebase?

andysellick (approved):
Looks good 👍

PR #4950, opened 2025-07-28
Split long `text` for `Ga4FormTracker` submit

patrickpatrickpatrick:
## What

Add optional text splitting functionality to `Ga4FormTracker` if `data-ga4-form-split-response-text` present on `module`
  - Add `this.splitText`, `this.splitFormResponseText` to `Ga4FormTracker`
  - `Ga4FormTracker` splits `text` into up to five attributes `text_n` if `text` is over 500 characters
  - Add `text_2`, `text_3`, `text_4`, `text_5` to `Schemas`

## Why

Requested change from Publishing PAs. Form responses can be significantly longer in Whitehall, so this is to ensure that the entire response is captured.

### DIFF
--- a/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.spec.js
+++ b/spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.spec.js
@@ -390,6 +390,67 @@ describe('Google Analytics form tracking', function () {
     })
   })
 
+  describe('when tracking a form with text splitting enabled', function () {
+    var longText = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam pulvinar quam ac elit pulvinar, quis iaculis lorem tristique. Nunc hendrerit urna et metus iaculis lobortis. Aenean tempor, lectus at blandit aliquet, neque felis aliquet leo, ornare tristique nibh augue sit amet tellus. Curabitur sit amet maximus lorem. Nullam volutpat, dolor id mollis ultrices, dui libero venenatis nibh, vel mollis nulla diam vel leo. Vestibulum tempus quam sit amet nisi cursus, et cursus ipsum semper. Vivamus eu.'
+    var overriddenText = JSON.parse(JSON.stringify(attributes))
+
+    beforeEach(function () {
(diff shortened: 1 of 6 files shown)


spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.spec.js
```diff
@@ -390,6 +390,30 @@ describe('Google Analytics form tracking', function () {
...
+      var longText = 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nullam pulvinar quam ac elit pulvinar, quis iaculis lorem tristique. Nunc hendrerit urna et metus iaculis lobortis. Aenean tempor, lectus at blandit aliquet, neque felis aliquet leo, ornare tristique nibh augue sit amet tellus. Curabitur sit amet maximus lorem. Nullam volutpat, dolor id mollis ultrices, dui libero venenatis nibh, vel mollis nulla diam vel leo. Vestibulum tempus quam sit amet nisi cursus, et cursus ipsum semper. Vivamus eu.'
+      var extraText = 'Extra characters after 500 characters.'
+      overriddenText.text = longText + extraText
+      element.setAttribute('data-ga4-form', JSON.stringify(overriddenText))
+      window.GOVUK.triggerEvent(element, 'submit')
+      expected.event_data.text = longText
+      expected.event_data.text_2 = extraText
+      expect(window.dataLayer[0]).toEqual(expected)
+    })
+  })
```
AshGDS, line 415: Could there maybe be some more tests here? For example maybe:

- Exactly 500 characters
- A for loop to create a 2,500 character long string so that all 5 text params are tested? 
- Possibly boundaries (499 / 501 characters)
- Possibly below 500 chars (but I guess this is implied working by other tests)

AshGDS (commented):
Looks good, I think the tests could maybe cover a bit more cases though.

Worth noting that around 15KB GA4 will start rejecting events, and as this adds potentially 2.5KB it's worth keeping in mind - see this PR: https://github.com/alphagov/govuk_publishing_components/pull/3994

AshGDS (approved):

PR #4951, opened 2025-07-29
Fix miscellaneous sass compilation warnings

andysellick:
## What
- updates some outdated syntax for colour adjustment
- declarations like `@include` have to come after other lines, otherwise the compiler complains

Moving some of these rules around caused a little bit of breakage when for some elements when focussed and hovered, should be correct now, but worth checking those states when reviewing.

## Why
Issues found as part of https://github.com/alphagov/govuk_publishing_components/pull/4931 but splitting them out into their own PR to get them done quicker.

## Visual Changes
Hopefully none.

Trello card: https://trello.com/c/8vKVSmVz/764-switch-from-import-to-use-in-sass-compilation

### DIFF
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -1,5 +1,5 @@
 // This file contains the styles for the Component Guide.
-
+@use "sass:color";
 @import "govuk_publishing_components/govuk_frontend_support";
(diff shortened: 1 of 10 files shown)


MartinJJones (approved):
Nice work!, I tested the changes in BrowserStack as well and it all looks good to me, the PR description in https://github.com/alphagov/govuk_publishing_components/pull/4931 was also super helpful

PR #4957, opened 2025-07-30
Use a template element to build new fieldsets for the add another component

ryanb-gds:
We want to make sure that if the content added to the DOM by the add another component has any JS modules associated with it, that these are initialised when the content is added. To achieve this, we can call `GOVUK.modules.start` after the fieldset has been added.

However, for our select with search module, this didn't work when we were removing the empty fieldset from the DOM and then cloning it and reinstating it to the DOM when the "Add another" button was clicked. It failed because ChoicesJS believed it was already initialised for that fieldset.

To work around this, we use a template element containing the empty fieldset. The template is not part of the DOM, so Choices is not initialised on page load as it is for the empty fieldset in the DOM. Now when the user clicks the add another button, we construct the new fieldset DOM from the template instead of the removed empty fieldset.

The existing empty fieldset rendered to the DOM is retained for cases where JS is disabled.

## Visual Changes

No visual changes, but there is a preview with the select with search module active inside the add another component. The CSS is missing for the select with search component, but the searching and selection does work.

Screenshot 2025-07-30 at 17 40 22

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Add special characters to GA4 phone number PII redaction ([PR #4955](https://github.com/alphagov/govuk_publishing_components/pull/4955))
+* Initialise JS modules for newly added fieldset in add another component ([PR #4957](https://github.com/alphagov/govuk_publishing_components/pull/4957))
 
 ## 59.2.0
 
--- a/app/assets/javascripts/govuk_publishing_components/components/add-another.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/add-another.js
@@ -5,7 +5,6 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   function AddAnother (module) {
     this.module = module
     this.disableGa4 = this.module.dataset.disableGa4
-    this.emptyFieldset = undefined
     this.addAnotherButton = undefined
     this.startIndex = Number(this.module.dataset.ga4StartIndex) || 1
     this.indexSectionCount = Number(this.module.dataset.ga4IndexSectionCount)
@@ -94,8 +93,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
   }
 
   AddAnother.prototype.removeEmptyFieldset = function () {
-    this.emptyFieldset = this.module.querySelector('.js-add-another__empty')
-    this.emptyFieldset.remove()
+    this.module.querySelector('.js-add-another__empty').remove()
   }
 
   AddAnother.prototype.updateFieldsetsAndButtons = function () {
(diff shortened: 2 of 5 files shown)


app/assets/javascripts/govuk_publishing_components/components/add-another.js
```diff
@@ -129,15 +127,18 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
...
     newFieldset.classList.add('js-add-another__fieldset')
     this.createRemoveButton(newFieldset, this.removeNewFieldset.bind(this))
     button.before(newFieldset)
 
-    this.incrementAttributes(this.emptyFieldset)
+    this.incrementAttributes(newFieldsetTemplate.content)
     this.updateFieldsetsAndButtons()
 
+    // Initialise any modules included in new fieldset
+    window.GOVUK.modules.start(newFieldset)
```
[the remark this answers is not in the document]
patrickpatrickpatrick: Is this not testable? This should run in Jasmine and then the component within the new fieldset should have a `data-name-of-component-module-started` data attribute added to it.

app/views/govuk_publishing_components/components/docs/add_another.yml
```diff
@@ -83,3 +83,37 @@ examples:
...
+
+  with_select_with_search:
+    description: Add another with a select with search input
+    data:
+      fieldset_legend: "Organisation"
+      add_button_text: "Add another organisation"
+      items:
+        - fields: >
+            <div class="govuk-form-group">
+            <select name="organisations[0][id]" id="organisations_1" class="govuk-select" data-module="select-with-search" tabindex="-1">
```
patrickpatrickpatrick, line 95: If you could add a label to this select then we won't get an accessibility failure warnings in the Component Guide I think.

app/assets/javascripts/govuk_publishing_components/components/add-another.js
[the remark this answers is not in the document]
ryanb-gds: It'd need to be a real module, and I'm loathe to couple the success of these tests to another module. However, I might be able to create a fake module in the test - let me have a think about it.

spec/javascripts/components/add-another-spec.js
```diff
@@ -477,4 +529,74 @@ describe('GOVUK.Modules.AddAnother', function () {
...
+
+    afterEach(function () {
+      document.body.removeChild(fixture)
+      delete GOVUK.Modules.TestModule
+    })
+
+    it('should start modules in new fields', function () {
+      window.GOVUK.triggerEvent(addButton, 'click')
+      var newFieldset = document.querySelectorAll('.js-add-another__fieldset')[1]
+      expect(!!newFieldset.dataset.testModuleModuleStarted).toBeTrue()
```
patrickpatrickpatrick, line 599: I think you could say 

```
expect(newFieldset.dataset.testModuleModuleStarted).toBeDefined()
```

as I don't think it'll ever be anything other than `true` if it is present and that reads a bit nicer 🙂

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Initialise JS modules for newly added fieldset in add another component
```
patrickpatrickpatrick, line 12: Could you link the PR in the same way as the other entries in the CHANGELOG?

patrickpatrickpatrick (approved):
Looks good to me!

PR #4958, opened 2025-07-30
Make H3's on the feedback component into H2's to avoid them appearing subordinate to unrelated headings

hannalaakso:
## What
DAC has reported an issue where the H3 headings "Help us improve GOV.UK" on the feedback component might appear subordinate to previous, unrelated H2 headings on the page. 

The component does have its own H2 "Is this page useful?" but this is hidden when one of the H3 headings on the feedback component is unhidden.

The change in this PR doesn't result in any visual changes. 

I've tested the change on NVDA + Chrome and JAWS + Chrome, as well as Mac Voiceover and have not seen any issues. 

It should also be safe to assume that this change will not cause any issues with the existing heading hierarchy on pages where the feedback component appears since the default heading on the component has always been a H2, we've ran this past our content designer who agrees with this.

## Why
Our accessibility specialist has recommended that the H3's are changed to H2's to convey the correct heading hierarchy to screen reader users. 

Fixes https://trello.com/c/6WtHvnis/3594-dacinformationandrelationships01

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove conditional comments and styles for Internet Explorer ([PR #4966](https://github.com/alphagov/govuk_publishing_components/pull/4966))
+* Make H3's on the feedback component into H2's to avoid them appearing subordinate to unrelated headings ([PR #4958](https://github.com/alphagov/govuk_publishing_components/pull/4958))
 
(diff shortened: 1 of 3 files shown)


MartinJJones (commented):
The change looks good to me, looks like this PR just needs a rebase following the recent release (v60) to resolve the merge conflict in the changelog, then happy to approve 👍

PR #4973, opened 2025-08-11
Add content block styles

pezholio:
This adds some rudimentary styling to to the Govspeak component for the Contact Content Block generated via the Content Block Manager. We originally intended for these styles to be delivered via the [content_block_tools gem](https://github.com/alphagov/content_block_tools).

However, this resulted in a more weighty CSS output and also meant we had to pull in a dependency on `govuk-frontend` in another place, which meant there was a risk of dependencies getting out of whack (see
https://github.com/alphagov/frontend/pull/4943).

After some discussion with @maxgds and @richardTowers, we decided that the least worse option would be to distribute these styles in the Components gem. This is a minor misuse of tools, as content blocks are not _strictly_ a component, but this is already repeating a pattern that we’ve already established with the packaging of `govspeak` as a
“component but not a component”.

Thoughts welcome!

## Screenshot

image

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Remove unneeded Sass mixins ([PR #4969](https://github.com/alphagov/govuk_publishing_components/pull/4969))
 * Upgrade to LUX version 4.3.2 ([PR #5018](https://github.com/alphagov/govuk_publishing_components/pull/5018))
+* Add content block styles ([PR #4973](https://github.com/alphagov/govuk_publishing_components/pull/4973))
 
 ## 60.2.1
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
@@ -1,0 +1,27 @@
+@import "govuk/core/lists";
+@import "govuk/core/typography";
+
+.content-block,
+.content-block__contact-key,
+.content-block__contact-value,
+.content-block__contact-list {
+  margin: 0;
+  padding: 0;
+}
+
+.content-block__contact-list {
+  .content-block__contact-key {
+    @extend %govuk-heading-m;
+  }
+}
+
+.content-block__contact-list--nested {
+  .content-block__contact-key {
+    @extend %govuk-heading-s;
+  }
+}
+
+.content-block .content-block__list {
+  @extend %govuk-list;
+  margin-left: 0;
+}
--- a/app/views/govuk_publishing_components/components/docs/govspeak.yml
+++ b/app/views/govuk_publishing_components/components/docs/govspeak.yml
@@ -544,6 +544,75 @@ examples:
             </div>
           </div>
         </div>
+  contact_content_block:
+    data:
+      block: |
+        <div class="content-block content-block--contact">
+          <dl class="vcard content-block__contact-list">
+            <dt class="content-block__contact-key fn org">Get help with your application</dt>
+            <dd class="content-block__contact-value">
+              <dl class="content-block__contact-list--nested">
+                <dt class="content-block__contact-key">Address</dt>
+                <dd class="content-block__contact-value">
+                  <p class="adr">
+                    <span class="street-address"> 49 to 53 Cherry Street</span>,<br><span class="locality">London</span>,<br><span class="postal-code"> AB1 2DC </span>
+                  </p>
+                </dd>
+                <dt class="content-block__contact-key">Email</dt>
+                <dd class="content-block__contact-value">
+                  <ul class="content-block__list">
+                    <li>
+                      <a href="mailto:name@example.com" class="email">
+                        name@example.com
+                      </a>
+                    </li>
+                    <li>
+                      <p>We aim to respond within 2 working days</p>
+                    </li>
+                  </ul>
+                </dd>
+                <dt class="content-block__contact-key">Phone</dt>
+                <dd class="content-block__contact-value">
+                  <p>If you have a unique reference number, have it with you when you call.</p>
+                  <ul class="content-block__list">
+                    <li>
+                      <span>Phone: </span>
+                      <span class="tel">020 7946 0101</span>
+                    </li>
+                    <li>
+                      <span>Textphone: </span>
+                      <span class="tel">020 7946 0102</span>
+                    </li>
+                    <li>
+                      <span>Welsh language: </span>
+                      <span class="tel">020 7946 0103</span>
+                    </li>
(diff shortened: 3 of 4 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/content_blocks/_contact.scss
andysellick, line 1: Is there a reason this bit is in a separate file? If not I'd suggest merging it with the main one.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
```diff
@@ -0,0 +1,32 @@
+@import "govuk/core/links";
+@import "govuk/core/lists";
+@import "govuk/core/typography";
```
andysellick, line 3: I think you can replace all these specific lines with 

```
@import "govuk_publishing_components/individual_component_support";
```

...which ultimately imports `govuk/base`, which I think includes all of this stuff? Worth trying out.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
```diff
@@ -0,0 +1,32 @@
...
+
+@import "content_blocks/contact";
+
+.govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
+.gem-c-govspeak {
+  .content-block {
+    .content-block__body {
+      @extend %govuk-body-m;
```
andysellick, line 11: Curious about the logic behind these class definitions (this one and `.content-block__link` on line 28). If you can control the class names in the code and they're only extending existing `govuk-frontend` classes, why not use those classes directly in the code i.e. `govuk-body` and `govuk-link`?

andysellick: (same for contact block title and subtitle)

app/assets/stylesheets/govuk_publishing_components/components/govspeak/content_blocks/_contact.scss
```diff
@@ -0,0 +1,14 @@
+@import "govuk/core/typography";
+
+.govspeak, // Legacy class name that's still used in some content items - needs to be kept until `.govspeak` is removed from the content items.
```
andysellick, line 3: It might be that you don't need to nest your new classes inside the `.govspeak` class - this is a (possibly overused) convention to prevent buttons that are links from being styled incorrectly.

CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add styling for a contact content block ([PR #4973](https://github.com/alphagov/govuk_publishing_components/pull/4973))
```
andysellick, line 12: Nitpick: can this match the PR title please?

andysellick (commented):
I've given this a quick glance but not an in-depth test, some suggestions, some questions.

app/assets/stylesheets/govuk_publishing_components/components/govspeak/content_blocks/_contact.scss
pezholio: That seems to work! 👍

app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -674,6 +674,70 @@ examples:
             It may even span multiple paragraphs.
           </p>
         </div>
+  contact_content_block:
+    data:
+      block: |
+        <div class="content-block content-block--contact" data-content-block="" data-document-type="contact" data-content-id="eafc7eb0-778f-4b48-ba23-10690a1c90d1" data-embed-code="{{embed:content_block_contact:get-help-with-your-application--2}}">
+          <div class="vcard">
+            <p class="fn org content-block__title">Get help with your application</p>
```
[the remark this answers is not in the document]
pezholio: Yep, I agree.  It's in our backlog for this spring to change the markup to be a Definition List (https://gov-uk.atlassian.net/jira/software/c/projects/CM/boards/134?selectedIssue=CM-334), but I wanted to get agreement on the direction of travel before we made any changes.

app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -674,6 +674,70 @@ examples:
             It may even span multiple paragraphs.
           </p>
         </div>
+  contact_content_block:
+    data:
+      block: |
+        <div class="content-block content-block--contact" data-content-block="" data-document-type="contact" data-content-id="eafc7eb0-778f-4b48-ba23-10690a1c90d1" data-embed-code="{{embed:content_block_contact:get-help-with-your-application--2}}">
```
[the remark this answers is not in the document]
pezholio: Probably not for the example, no

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
```diff
@@ -0,0 +1,37 @@
...
+
+  // stylelint-disable selector-no-qualifying-type
+  // This qualifying type allows us to override the default styling for `ul` elements within
+  // `.govspeak` elements
+  ul.content-block__list {
+    @extend %govuk-list;
+    margin-left: 0;
+
+    // stylelint-disable max-nesting-depth
+    & > li {
```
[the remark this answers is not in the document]
pezholio: Cool - nice spot 👍

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_content-block.scss
```diff
@@ -0,0 +1,23 @@
+@import "govuk/core/links";
+@import "govuk/core/lists";
+@import "govuk/core/typography";
+
+.content-block {
+  // stylelint-disable selector-no-qualifying-type
+  // This qualifying type allows us to override the default styling for `ul` elements within
+  // `.govspeak` elements
+  ul.content-block__list {
```
andysellick, line 9: Is `ul` needed here? I've tried it without and it seems to render the same.

app/views/govuk_publishing_components/components/docs/govspeak.yml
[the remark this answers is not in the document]
pezholio: Yeah, the problem is, we can't guarantee where this block will be embedded, so having (say) a `h2` could be risky if this is placed after (say) a `h3`, because it could mess with the reading order of the page.

PR #4992, opened 2025-08-18
Style govspeak footnote targets

andysellick:
## What
Make the chosen footnote visually different from the others.

- add styling using the [target](https://developer.mozilla.org/en-US/docs/Web/CSS/:target) pseudo class to govspeak footnotes, so that the currently selected footnote is visually distinct from any others
- this should distinguish the chosen footnote particularly in situations where it is not at the top of the scrolled viewport
- considered using the govuk-frontend focussed state for this but can't because footnotes may contain links already, so could result in a focus state within a focus state

I've also slightly expanded the govspeak footnote documentation just to make it clearer.

Preview app link: https://components-gem-pr-4992.herokuapp.com/component-guide/govspeak/footnotes

Example of a page with footnotes for testing: https://www.gov.uk/government/consultations/environmental-assessment-levels-eals-used-in-air-emissions-risk-assessments/public-feedback/appendix-c-summary-of-toxicological-evidence-for-mea-and-ndma

Checked with @nnagewad on the design of this.

## Why
Going to blame Jake Archibald for making a [really strong case against footnotes](https://jakearchibald.com/2025/give-footnotes-the-boot/) where I learned of the target pseudo class. We can't get rid of footnotes overnight, but maybe we can make them a bit better.

## Visual Changes
Highlighted footnote:

Screenshot 2025-08-19 at 13 19 54

Highlighted footnote containing a focussed link:

Screenshot 2025-08-19 at 13 20 04

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Style govspeak footnote targets ([PR #4992](https://github.com/alphagov/govuk_publishing_components/pull/4992))
 * Add Kurdish Sorani and Tigrinyan translations ([PR #4987](https://github.com/alphagov/govuk_publishing_components/pull/4987))
 * Remove unused imports in layout-header.scss ([PR #4985](https://github.com/alphagov/govuk_publishing_components/pull/4985))
 * Remove aria-expanded from feedback component Cancel buttons ([PR #4974](https://github.com/alphagov/govuk_publishing_components/pull/4974))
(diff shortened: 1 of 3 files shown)


maxgds:
I'm not sure they're suitable for this purpose Nik. We need something to say "this item is selected" while minimising noise and distraction. While we can't use colour alone those things look like they'd bring extra issues if applied to something that's primarily there to be read. Happy to be proven wrong, but my instinct says borders/underlines/dot etc would be better things to apply to a chunk of text than a background. We can use the background Andy started out with as a complementary thing, it just can't be the only thing.

andysellick:
@maxgds we've had another go at this and added a border, so now the active state slightly resembles the focus state. See attached screenshots. Thoughts?

PR #4994, opened 2025-08-19
Adjustments to how `Ga4FormTracker` sets `text` on `submit`

patrickpatrickpatrick:
## What

- `Ga4FormTracker` conditional field adjustments
- `Ga4FormTracker` improve date component handling
- `Ga4FormTracker` report single selected option
- `Ga4FormTracker` stop including placeholders

### `Ga4FormTracker` conditional field adjustments

The fields within a conditional checkbox were always being included even if the checkbox was not checked. This was because the query for the conditional checkbox input was being executed from the wrong node. Moving the query to `this.module` has fixed this. This was not identified before as the markup within the tests was not accurate to how conditional checkboxes are rendered. In addition, the structure of conditional fields within the JSON assigned to `text` (if JSON recording enabled) has been flattened.

###  `Ga4FormTracker` improve date component handling

Improve how dates are represented within `text`. Given a form with a date component of values:

Screenshot 2025-08-19 at 12 24 40

with `data-ga4-form-use-text-count` this would be the response:

```
{
  "Day": "2,
  "Month": "2",
  "Year": "4"
}
```

which would make it difficult to identify specific fields if a form had more than one date. Additionally, as the date fields are `input[type=text]` the character count of these fields would be returned instead of the date values. The adjustments introduced in this PR mean the date component is now represented as the following:
 
```
{
  "First published date - Day": "23",
  "First published date - Month": "08",
  "First published date - Year": "2025"
}
```

For a date component that isn't within a conditional fieldset such as:

Screenshot 2025-08-19 at 12 26 26

it will be represented as the following:

```
{
  "Date (required) - Day": "23",
  "Date (required) - Month": "08",
  "Date (required) - Year": "2025"
}
```

If the Date is within another legend (for selecting the date and time for instance) then it will be represented as the following:

Screenshot 2025-09-24 at 12 10 46

```
{
  "Opening Date - Date (required) - Day": "23",
  "Opening Date - Date (required) - Month": "08",
  "Opening Date - Date (required) - Year": "2025"
}
```

### `Ga4FormTracker` report single selected option

Instead of reporting `1` for the value of a `select` with a single selected `option` with `useSelectCount` enabled, the text of the selected `option` will be used instead.

### `Ga4FormTracker` stop including placeholders

For `select` elements that have a blank placeholder option (that will be selected so that it acts as a placeholder), `Ga4FormTracker` will no longer include the text of that blank placeholder `option` in the reported submit event.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Adjustments to how Ga4FormTracker sets text on submit ([PR #4994](https://github.com/alphagov/govuk_publishing_components/pull/4994))
+
 ## 61.1.0
 
 * Revert GA4 phone number PII changes ([PR #5039](https://github.com/alphagov/govuk_publishing_components/pull/5039))
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-form-tracker.js
@@ -135,7 +135,7 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         }
       } else if (inputNodename === 'SELECT' && elem.querySelectorAll('option:checked')) {
         var selectedOptions = Array.from(elem.querySelectorAll('option:checked')).map(function (element) { return element.text })
-        input.answer = this.useSelectCount ? selectedOptions.length : selectedOptions.join(',')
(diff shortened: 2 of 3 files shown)


patrickpatrickpatrick:
> Looks good - is a performance analyst on `www.gov.uk` aware of these changes? I suspect this will change how data is collected on certain forms for them as well (mainly smart answers I think), so just wanted to check.

Ah, how do you think it will change for them out of interest? I tried to have all changes behind flags so I think in theory this should not impact the resulting data collection.

AshGDS (approved):
@patrickpatrickpatrick Ah OK, no worries - yeah I was just thinking if it would affect pages like this https://www.gov.uk/calculate-inheritance-tax-interest but I'm not sure conditional checkboxes are used on smart answers anyway.

PR #4996, opened 2025-08-20
(NAV-2611) Remove `aria-expanded` attribute from feedback component

MartinJJones:
## What

- Remove the `aria-expanded` attribute from the yes/no buttons in the feedback component view template and JavaScript file
- Updated the tests for the feedback component
  - Update the fixture to reflect the latest version of the feedback component HTML
  - Remove tests relating to `aria-expanded` on the yes/no buttons as this attribute was removed from the HTML

## Why

DAC have reported an issue where the two "Cancel" buttons on the feedback component aren't using aria-expanded as expected. The buttons do collapse the region, but behave more like a 'Close' button than an expandable control.

This change ensures the aria-expanded attribute is not set in the view template or via JavaScript for the feedback component.

### Further info

- Jira: https://gov-uk.atlassian.net/browse/NAV-2611
- Relates to https://github.com/alphagov/govuk_publishing_components/pull/4974


MartinJJones:
> I think you can remove this.pageIsNotUsefulButton since it's no longer used anywhere.

@jon-kirwan good spot, I've removed as suggested, thanks

Issue #4929, opened 2025-07-17
Use `erb_lint` in the gem


AshGDS:
Other apps on GOVUK are using [erb_lint](https://github.com/Shopify/erb_lint) to ensure `.html.erb` files are linted properly.

We can probably implement this in the gem. I've run `bundle add erb_lint` and then run `erb_lint` on `app/` and `spec/` to see how many changes would be needed. It looks like there are 160 errors currently in the gem, but most of these seem fairly straightforward to fix as they are things like fixing spacing or adding new lines to the end of files.

Here's what it output:

```bash
No space detected where there should be a single space.
In file: app/views/govuk_publishing_components/audit/_component_contents.html.erb:6

No space detected where there should be a single space.
In file: app/views/govuk_publishing_components/audit/_component_contents.html.erb:64

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:65

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:117

Extra whitespace detected at end of line.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:73

Extra whitespace detected at end of line.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:85

Extra whitespace detected at end of line.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:89

Extra whitespace detected at end of line.
In file: app/views/govuk_publishing_components/audit/_components.html.erb:92

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/audit/show.html.erb:13

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/component_guide/example.html.erb:14

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/component_guide/index.html.erb:6

Remove newline before `%>` to match start of tag.
In file: app/views/govuk_publishing_components/component_guide/show.html.erb:14

No space detected where there should be a single space.
In file: app/views/govuk_publishing_components/components/_back_to_top_link.html.erb:16

Bad ERB comment syntax. Should be <%# without a space between.
Leaving a space between ERB tags and the Ruby comment character can cause parser errors.
In file: app/views/govuk_publishing_components/components/_big_number.html.erb:31

Avoid using "application/ld+json" as type for `<script>` tag. Must be one of: text/javascript (or no type attribute).
[...]

PR #5003, opened 2025-08-28
Introduce erb_lint linter

AshGDS:
## What / Why

- Introduce `erb_lint`, a `.html.erb` linter, copying the config from `frontend`. Adding this will improve the quality of code written in the repo.
- I've added the command to our linting rake task, so it should run when `bundle exec rake` is used.
- In one of the commits I ran `--autocorrect` which resulted in quite a lot of files changed, but it's mostly spacing, quotation marks, and commas that were changed. In a couple instances it added `rel: noopener` to `target=_blank` URLs. In subsequent commits I manually fixed the linting errors it couldn't autocorrect.
- Trello card: https://trello.com/c/t4lTuvtr/799-add-erblint-to-govukpublishingcomponents
- Closes https://github.com/alphagov/govuk_publishing_components/issues/4929

## Visual Changes

None


Gemfile
```diff
@@ -16,3 +16,5 @@ gemspec
 
 # To use a debugger
 # gem 'byebug', group: [:development, :test]
+
+gem "erb_lint", "~> 0.9.0"
```
MartinJJones, line 20: Just wanted to check if `erb_lint` should/could be included in the gemspec file as a development dependency - https://github.com/alphagov/govuk_publishing_components/blob/main/govuk_publishing_components.gemspec#L29-L46

MartinJJones (commented):
Nice work, the change looks good to me overall, I just have one question on where `erb_lint` should be included

Gemfile
AshGDS: Thanks @MartinJJones - have added this now 👍

MartinJJones (approved):
Nice work, the change looks good to me 👍

PR #5007, opened 2025-09-02
Upgrade to LUX 4.2.1

andysellick:
## What
Upgrade to LUX version 4.2.1.

A lot of changes in this one, although they're mostly changing instances of `var` to `const` or `let`. One or two other interesting things, see my comment below.

## Why
New version. Have tested it according to the instructions at https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md and it seems to be working correctly.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
+++ b/app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
@@ -22,9 +22,9 @@
 (function () {
   'use strict';
 
-  var max = Math.max;
-  var floor = Math.floor;
-  var round = Math.round;
+  const max = Math.max;
+  const floor = Math.floor;
+  const round = Math.round;
   /**
   * Clamp a number so that it is never less than 0
   */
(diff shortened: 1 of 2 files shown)


app/assets/javascripts/govuk_publishing_components/vendor/lux/lux-reporter.js
```diff
@@ -468,7 +466,7 @@
       maxBeaconUrlLength: getProperty(obj, "maxBeaconUrlLength", 8190),
       maxBeaconUTEntries: getProperty(obj, "maxBeaconUTEntries", 20),
       maxErrors: getProperty(obj, "maxErrors", 5),
-      maxMeasureTime: getProperty(obj, "maxMeasureTime", 60000),
+      maxMeasureTime: getProperty(obj, "maxMeasureTime", 60_000),
```
andysellick, line 469: This is interesting... haven't seen a variable of the form `60_000` before, but it's in the original code.

PR #5008, opened 2025-09-02
Update feedback component tests

MartinJJones:
## What

Update feedback component tests:
- Update the fixture HTML to reflect the latest version of the feedback component in use
- Remove redundant tests, the "page is not useful" form no longer contains an input to capture an email address, instead it directly links to a separate survey, tests relating the input field to capture the email address have been removed

## Why

Ensure the feedback component tests are accurate, this issue was originally spotted when working on the the [PR to remove aria-expanded attribute from feedback component](https://github.com/alphagov/govuk_publishing_components/pull/4996)

Jira: https://gov-uk.atlassian.net/browse/NAV-2611

### DIFF
--- a/spec/javascripts/components/feedback-spec.js
+++ b/spec/javascripts/components/feedback-spec.js
@@ -188,15 +188,6 @@ describe('Feedback component', function () {
       expect($('.js-something-is-wrong').attr('aria-expanded')).toBe('false')
     })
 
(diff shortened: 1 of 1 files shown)


hannalaakso (approved):
Nice tidy up, thanks Martin 🧹  As discussed IRL, we'll review with the wider community. how these tests get updated in the future.

PR #5009, opened 2025-09-03
Add fallback for missing GA4 canonical URL

AshGDS:
## What/Why
- Some pages on GOV.UK have a `<link rel="canonical` tag which contains the canonical url of the page. We grab this value for GA4 tracking purposes.
- The pages that have the canonical url are ones that render the `machine_readable_metadata` component.
- Not all pages are rendering this component though, so we need an alternative way to get the canonical URL.
- Looking at the code, the canonical_url is usually a combination of the domain name and the base_path.:
https://github.com/alphagov/govuk_publishing_components/blob/63cfc71c53293949b584963556a5f21e598d9b68/lib/govuk_publishing_components/presenters/machine_readable/page.rb#L14-L16
- This only seems to be overruled in [one case by travel advice](https://github.com/search?q=org%3Aalphagov+govuk_publishing_components%2Fcomponents%2Fmachine_readable_metadata+AND+canonical_url&type=code), so I think we can be confident most canonical urls are domain + base_path. 
- Therefore, we can recreate this code in our GA4 tracking where the `<link rel="canonical"` tag doesn't exist in the DOM.
- Trello card: https://trello.com/c/LbnGWY6Y

## Visual Changes

None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 * Remove date PII from GA4 page view tracking ([PR #5004](https://github.com/alphagov/govuk_publishing_components/pull/5004))
 * Introduce erb_lint linter ([PR #5003](https://github.com/alphagov/govuk_publishing_components/pull/5003))
 * Remove aria-expanded attribute from feedback component ([PR #4996](https://github.com/alphagov/govuk_publishing_components/pull/4996))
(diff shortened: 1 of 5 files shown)


andysellick (approved):
Looks good to me 👍

PR #5011, opened 2025-09-04
Set row classes to be an array

pezholio:
In #4999 we attempted to address the deprecation errors in the logs. This is mainly fixed, but there’s still one place where we’re pushing to a string using `<<`. 

In other components, we push to arrays, which is supported by the tag helpers, so for consistency, let’s do it here. This should remove the deprecation errors in newer Rubies.


CHANGELOG.md
```diff
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Set `row_classes` in the Summary Card component to be an array ([PR #5011](https://github.com/alphagov/govuk_publishing_components/pull/5011))
```
andysellick, line 12: Nitpick: can this match the PR name please?

PR #5021, opened 2025-09-10
[DO NOT MERGE] Add fix for menu height when zoomed

andysellick:
## What / why
Adds a fix for the problem of making the background of the dropdown menus (light blue) full width.

Note `100000px` is a deliberately huge value to ensure the elements always reach the edge of the page. Might be safer to make them even wider?

cc @MartinJJones 

## Visual Changes
None.


MartinJJones:
Nice, thanks for this @andysellick 👍 

I've created a new card to further [improve the super navigation menu component](https://gov-uk.atlassian.net/browse/NAV-18511) and added it to our backlog, I've included a link to this PR as well so we can look at making the change and complete browser testing before implementing

PR #5023, opened 2025-09-16
GOV.UK Frontend v5.11.0 and v5.11.1 updates

jon-kirwan:
## What

https://gov-uk.atlassian.net/browse/NAV-18205

GOV.UK Frontend v5.11.0 and v5.11.1 updates
- The Service navigation component no longer uses a menu on mobile for a single link
- Add inverse styling to Service navigation
- Reapply colour mixins instead of hex values (used for the brand refresh)

**Review URL**: https://components-gem-pr-5023.herokuapp.com/component-guide/service_navigation

## Why

Although GOV.UK Publishing Components was previously updated to versions GOV.UK Frontend 5.11.0 - 5.11.2, not all fixes and improvements from those releases were applied. This pull request ensures that the remaining updates are now implemented.

## Anything else

If testing from the 'preview all' page in the component guide, only the first service navigation menu toggle responds. This happens because each toggle is tied to a unique navigation container id `id="navigation"`. This matches the https://design-system.service.gov.uk/components/service-navigation/ examples.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* GOV.UK Frontend v5.11.0 and v5.11.1 updates ([PR #5023](https://github.com/alphagov/govuk_publishing_components/pull/5023))
+
 ## 61.0.0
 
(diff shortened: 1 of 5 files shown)


MartinJJones (approved):
Nice work, the updates look good to me 👍

PR #5062, opened 2025-10-08
Move draft watermark from static to layout_for_public

AshGDS:
## What/Why

- The draft watermark is missing from `frontend` now that it has been moved away from `static`
- In `static` it was on the `main` element, but it can go on the `body` element which `layout_for_public` handles
- Therefore the draft watermark can just go in `layout_for_public`, so this PR moves it to there
- We should probably raise an issue to see if we can get this into Percy in the future.

## Visual Changes

| Before | After |
|--------|--------|
|image|image|


andysellick (approved):
👍

PR #5075, opened 2025-10-14
(CM-540) Fix unlabelled listbox

pezholio:
During an accessibility review, we were made aware that the `listbox` element that is created by choices.js did not have an associated label (required for WCAG 2.2 AA).

This uses the `callbackOnInit` function to add an `aria-labelledby` attribute to the `listbox` element that points to the form component’s label, which seems to fix the issue.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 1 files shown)


ChrisBAshton (approved):
LGTM!

PR #5098, opened 2025-10-24
Standardise success alert text size

andysellick:
## What / why
Change the text size in the success alert component.

- size of the initial text in the component (bold) varied from 19px if no description provided to 24px if it was
- can't see any reason for this behaviour and not consistent with the original [Design System component](https://design-system.service.gov.uk/components/notification-banner/), so making them the same size

## Visual Changes
With this change the bold text is always 24px, regardless of whether a description is included or not.

Before | After
------ | ------
Screenshot 2025-10-24 at 16 06 06 | Screenshot 2025-10-24 at 16 06 14

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Standardise success alert text size ([PR #5098](https://github.com/alphagov/govuk_publishing_components/pull/5098))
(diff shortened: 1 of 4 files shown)


AshGDS (approved):
LGTM, is there a reason these don't use the heading component?

andysellick:
@AshGDS good question. I think probably because the headings use classes from the Design System component that have extra attributes (e.g. max-width) beyond the standard govuk-heading classes.

PR #5111, opened 2025-10-30
Add applications status page

andysellick:
## What
Adds a new page to the component guide to show some of the information previously shown in the applications consistency spreadsheet.

This is an initial version of this page. It doesn't include everything from the spreadsheet, because not everything is still useful, and some things are harder to get. There may also be other things not in the spreadsheet that we can add here. I'm hoping we'll expand this over time as we need to.

This works by first looking locally to find copies of the repos and the Gemfile.lock within them. If it can't find them locally, it'll send requests to github, again for the Gemfile.lock. These requests are cached by Rails for 5 minutes, to reduce server load.

## Why
The spreadsheet is hard to maintain, somewhat out of date, and likely to disappear in the next while.

I've made this separate from the existing component auditing pages because while the information is related the code there is structured in a very specific way and it's hard to extend to include this new functionality. It also may be that some of the features in the auditing will become less useful in the future so I didn't want to mingle the two too much.

## Visual Changes
Adds a new page at `/component-guide/applications` that shows the following information for our applications:

- name and link to github repo
- source, showing where the information was obtained from (this is either local, remote or not found - the page tries to find local repos to examine first, but if they're not there will ask github)
- gem version (the version of govuk_publishing_components in use by that app, highlighting in red if it's not the latest version)
- slimmer version (indicating whether that app is still reliant on static)
- sass version (this one is also readily available in Gemfile.lock, and may be useful soon)

Screenshot 2025-11-05 at 13 02 42

### DIFF
--- a/app/controllers/govuk_publishing_components/applications_page_controller.rb
+++ b/app/controllers/govuk_publishing_components/applications_page_controller.rb
@@ -62,7 +62,7 @@ module GovukPublishingComponents
           gem_version: app.gem_version,
           gem_version_status: app.gem_version == GovukPublishingComponents::VERSION ? "green" : "red",
           slimmer_version: app.slimmer_version,
-          dartsass_version: app.dartsass_version,
+          sass_version: app.sass_version,
         }
       end
 
--- a/app/models/govuk_publishing_components/applications_page.rb
+++ b/app/models/govuk_publishing_components/applications_page.rb
@@ -20,8 +20,8 @@ module GovukPublishingComponents
       parse_file(@gemfilelock, /slimmer \(([^)>=~ ]+)\)/)
     end
 
-    def dartsass_version
-      parse_file(@gemfilelock, /dartsass-rails \(([^)>=~ ]+)\)/)
+    def sass_version
+      parse_file(@gemfilelock, /sass-embedded \(([^)>=~ ]+)\)/)
     end
 
   private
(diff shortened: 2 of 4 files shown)


app/controllers/govuk_publishing_components/applications_page_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+          sass_version: app.sass_version,
+        }
+      end
+
+      output
+    end
+
+    # def cache_key_with_version
+    #   (0...8).map { (65 + rand(26)).chr }.join # random string
+    # end
```
AshGDS: Commented out code

app/controllers/govuk_publishing_components/applications_page_controller.rb
```diff
@@ -0,0 +1,77 @@
...
+        service-manual-publisher
+        short-url-manager
+        signon
+        smart-answers
+        specialist-publisher
+        static
+        support
+        transition
+        travel-advice-publisher
+        whitehall
```
jon-kirwan: Should we include GOV.UK Chat in the list? https://github.com/alphagov/govuk-chat

app/controllers/govuk_publishing_components/applications_page_controller.rb
andysellick: Whoops, fixed, thanks.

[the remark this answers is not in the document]
andysellick: Removed, thanks for spotting 👍

jon-kirwan (approved):
Nice work - looks good to me, and it's great to have all of this information in the gem rather than in the spreadsheet. For a future change, it might also be good to separate the frontend, backend, and utility apps similar to how they’re grouped in the spreadsheet.

Issue #5099, opened 2025-10-27
Option select component JS tests fail when stylesheets are included


andysellick:
In https://github.com/alphagov/govuk_publishing_components/pull/5087 we're trying to fix the component guide to render properly in applications that don't use the individual CSS model. The decision made was to turn off that model in the component guide universally, regardless of the state of an application.

This had an interesting side effect - the option select component JavaScript tests suddenly started failing. This was happening because Jasmine is configured to include the component guide stylesheet, and this now includes the component styles - previously it did not.

The difference is a bit clearer if we run one of the failing tests and disable the automatic cleanup of the component to the DOM for testing purposes:

Without component CSS (original) | With component CSS (new)
------------- | -----------
<img width="1345" height="490" alt="Image" src="https://github.com/user-attachments/assets/67afc3fb-d580-41c0-91eb-871f2c4ef09f" /> | <img width="1351" height="473" alt="Image" src="https://github.com/user-attachments/assets/01506991-5d11-4707-9f0c-778bd8555b0d" />

The tests for the option select component are based around the height of an element within the component. I think the problem is happening because when the stylesheets are included it changes the height of all the elements involved, thus breaking the tests. **However** it's not clear whether the tests are now failing because this is wrong, or whether the tests themselves are flawed and need fixing (hence this issue).

Currently the fix has been to remove the stylesheets from the Jasmine config, by removing the lines below, but we should put them back in and check to see why the tests are failing and whether they need rewriting.

```
  "cssFiles": [
    "component_guide/application-*.css"
  ],
```

## History

The addition of the component guide stylesheet to the Jasmine config happened in this commit in 2022 - https://github.com/alphagov/govuk_publishing_components/commit/1a42953c9c124dbc07447ee7d952fe756758e02d - prior to the [migration of the option select component to the gem](https://github.com/alphagov/govuk_publishing_components/pull/3623) in 2023 (which is why this problem didn't show up until now).

## Details of the test failures

```
1) An option select component getvisibleCheckboxes returns all the checkboxes if the container doesn't overflow
  Message:
    Expected 12 to be 4.
  Stack:
        at <Jasmine>
        at UserContext.<anonymous> (http://localhost:37255/__spec__/components/option-select-spec.js:270:50)
        at <Jasmine>

2) An option select component setupHeight expands the checkbox-container to fit checkbox list if the list is < 50px larger than the container
  Message:
    Expected 200 to be 202.
  Stack:
        at <Jasmine>
[...]

PR #5117, opened 2025-10-30
Restore option select tests

andysellick:
## What
- restores tests that were disabled as part of a recent change that caused them to break
- also fixes a small issue where max-height was overriding height for the component calculation

## Why
Relates to issue https://github.com/alphagov/govuk_publishing_components/issues/5099

## Visual changes
Should make the automatic calculation of the option select component to always show that there are more checkboxes if you scroll down more reliable.


andysellick:
@MartinJJones interestingly from reading the tests (see 2nd commit) it looks like the component never expected its own CSS to load - there's places where we set the height of the component with JS setting CSS values.

Issue #3281, opened 2023-03-01
Can we extend GOV.UK Publishing Component 


nnagewad:
# Issue

- Noticed the Previous and Next navigation component isn't up to date with GOV.UK Design System's Pagination component
    - [Previous and Next navigation - GOV.UK Component Guide](https://components.publishing.service.gov.uk/component-guide/previous_and_next_navigation)
    - [Pagination - for larger number of pages - GOV.UK Design System](https://design-system.service.gov.uk/components/pagination/#for-larger-numbers-of-pages)
- Relevant to the following PR: https://github.com/alphagov/whitehall/pull/7376

MartinJJones:
After further discussion, we should:

- Align with the design pagination component - https://design-system.service.gov.uk/components/pagination/#for-larger-numbers-of-pages
- Rename the component to "Pagination", the same as the design system

PR #5121, opened 2025-11-03
BREAKING: Replace `Previous and Next Navigation` with `Pagination` [WHIT-2660]

patrickpatrickpatrick:
## Breaking change

Applications will need to update:
- `previous_and_next_navigation` to `pagination`
- `url` of `previous_page` and `next_page` to `href`

The applications that use Previous and Next (from [this search](https://github.com/search?q=org%3Aalphagov+govuk_publishing_components%2Fcomponents%2Fprevious_and_next_navigation&type=Code)) are:

- government-frontend
- frontend
- finder-frontend
- govuk-chat
- specialist-publisher

Most of these use a helper function within a presenter to generate the arguments passed into the component. Therefore updating these applications will be straightforward.

## What

- Rename `Previous and Next Navigation` to `Pagination`
- Rename `url` parameter to `href` (for consistency in the parameter naming and to reflect that it sets `href` of an `a`) 
- Add pagination functionality to `Pagination`
- Add `PaginationHelper`

### Further technical details

- Adds new `items` parameter. Each `item` is rendered as a page by the Pagination component. The same parameter shape is used as the Pagination component in GOV.UK Frontend.
- `PaginationHelper` generates the previous and next links in the pagination component via `prev_link` and `next_link`, this moves duplicated mark-up and logic out of the template
- If `items` are defined in the `pagination` component then `data-ga4-link` will be updated to use `pagination` instead of `previous and next` for the `type` attribute. This won't change any uses of the current Previous and Navigation component that don't use pages.

Resolves https://github.com/alphagov/govuk_publishing_components/issues/3281

## Why

This adds the pagination functionality from GOV.UK Design System Pagination component (which the Previous and Next Navigation component was based on). It also renames the component to Pagination for parity with the Design System.

The pagination component is used throughout publishing and currently [each publishing application that requires it has it's own implementation of it](https://github.com/search?q=org%3Aalphagov%20pagination.html.erb&type=code). After [a discussion in the GOV.UK frontend community](https://github.com/alphagov/govuk_publishing_components/issues/3281#issuecomment-2858116072), we determined the path forward was to add the missing functionality to `Previous and Next Navigation` and rename it to `Pagination`.

The parameters for an item are the same as those in the GOV.UK Design System Pagination component.

## Visual Changes

The existing `Previous and Next Navigation` appearance will not change if `items` have not been specified.

Screenshot 2025-11-04 at 11 53 06

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* **BREAKING** Replace Previous and Next Navigation with Pagination ([PR #5121](https://github.com/alphagov/govuk_publishing_components/pull/5121))
+
 ## 62.3.0
 
 * Add an option for title language to public layout ([PR #5169](https://github.com/alphagov/govuk_publishing_components/pull/5169))


lauraghiorghisor-tw (commented):
Was just reading through - LGTM, but will defer to FE community.

MartinJJones (approved):
The changes look great, nice work on this 🙌 

I also tested out the changes in `finder-frontend` by following the upgrade instructions in the PR description and everything worked as expected. Just needs a `BREAKING:` changelog entry when you are ready to merge

PR #5127, opened 2025-11-10
LUX v4.4.2

andysellick:
## What
Upgrade to latest version of LUX. Have tested and passes in the debug tool as documented.

Also including an update to make the updating instructions easier to follow.

## Why
Based on https://github.com/alphagov/govuk_publishing_components/blob/main/docs/real-user-metrics.md#how-to-update-to-a-new-version

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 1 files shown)


AshGDS (approved):
LGTM, thanks for doing the update 👍

PR #5131, opened 2025-11-13
Do not add crest classes for organisation logos with no visual identity or a custom logo [WHIT-1799]

ryanb-gds:
## What

The crest class causes some extra padding and a border to appear around the organisation logo (or name, if there is no logo). This is not desirable in cases where a custom logo is being used or the organisation has no visual identity.

## Why

Previously this has not been a problem, because Whitehall has excluded the crest attribute from the Publishing payload for organisations if the crest has no visual identity or is a custom logo.

However, because we are migrating organisations to Search API v1's govuk index, and [collections frontend depends on the "custom" and "no-identity" values in Search API responses](https://github.com/alphagov/collections/blob/5cf9c38bba6051bce4a71b976688218669364109/app/models/search_api_organisation.rb#L23) to group organisations on taxon pages, we need to start including these crest values in the Publishing API payload.

This means that the organisation logo component must not apply the crest classes when a custom logo is in use or the crest has no identity.

#### Custom Crest
Screenshot 2025-11-13 at 14 13 49

#### No Identity
Screenshot 2025-11-13 at 15 32 50

#### Custom Crest
Screenshot 2025-11-13 at 14 13 57

#### No Identity
Screenshot 2025-11-13 at 15 33 19

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/organisation_logo.yml
+++ b/app/views/govuk_publishing_components/components/docs/organisation_logo.yml
@@ -28,6 +28,12 @@ examples:
         name: 'Single Identity'
         url: 'http://www.gov.uk'
         crest: 'single-identity'
+  no_identity:
+    data:
+      organisation:
+        name: 'No Identity'
+        url: 'http://www.gov.uk'
+        crest: 'no-identity'
   attorney_generals_office:
     data:
       organisation:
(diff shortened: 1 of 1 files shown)


AshGDS (approved):
No worries, thanks 👍

PR #5132, opened 2025-11-14
[DO NOT MERGE] Add options details to component wrapper helper

andysellick:
## What
Improve debugging for components using the component wrapper helper.

- adds a data attribute to components using the component wrapper helper that contains a JSON string of all of the options passed to that component, and their (truncated to 50 characters) values
- includes a suppress option (so we can disable it in components that don't need this, and so we don't have to rewrite all of the existing tests to include the new data)
- won't be included on staging or production, as adds a bit to the page weight

## Why
It would be good to have better transparency around which options are being passed to instances of components. Sometimes it's easy to see (they're written in the .erb) but sometimes they're passed an object from a presenter method, and it's a bit trickier to follow the code and see what the component is actually being passed.

With this addition, all of the options passed to the component are recorded as a data attribute on the parent element of the component. We can add something into the GOV.UK browser extension to run a bit of JS (something like the below) to read this information and append it to the page.

```JS
const components = document.querySelectorAll('[data-options]')
for (var i = 0; i < components.length; i++) {
  const thisel = components[i]
  const data = JSON.parse(thisel.getAttribute('data-options'))

  const match_name = /\bgem-c-[a-z0-9]+(?:-[a-z0-9]+)*(?=\s|$)/;
  let name = thisel.getAttribute('class')
  name = name.match(match_name)[0].replace('gem-c-', '').replaceAll('-', ' ')

  let formatted_data = `<strong>${name} component options</strong>`
  for (var key in data) {
    let value = data[key]
    if (value.length == 50) {
      value = `${value}..`
    }
    formatted_data = `${formatted_data}<br/>${key}: ${value}`
  }

  const div = document.createElement('div')
  div.style.border = 'solid 1px black'
  div.style.padding = '5px'
  div.innerHTML = formatted_data

  const parent = thisel.parentNode
  parent.insertBefore(div, thisel)
}
```

## Visual Changes
No visual change, but once we've hooked up the browser extension it could look something like the below.

Screenshot 2025-11-14 at 15 33 25

### DIFF
--- a/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
+++ b/app/views/govuk_publishing_components/components/_breadcrumbs.html.erb
@@ -13,6 +13,7 @@
   classes << "govuk-breadcrumbs--inverse" if inverse
   classes << "gem-c-breadcrumbs--border-bottom" if border == "bottom"
 
+  local_assigns[:suppress_output] = true
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.add_class(classes.join(" "))
   component_helper.add_aria_attribute({ label: "Breadcrumb" })
--- a/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
@@ -24,6 +24,7 @@
                                 )))
   services_cookies ||= nil
 
+  local_assigns[:suppress_output] = true
   component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
   component_helper.set_id(id)
   component_helper.add_class("gem-c-cookie-banner govuk-clearfix govuk-cookie-banner js-banner-wrapper")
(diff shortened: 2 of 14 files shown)


AshGDS (changes requested):
Seems like a good idea, but on certain pages with a lot of components I assume it will end up adding a lot of unnecessary weight to the page for no benefit to our users - and on every page I assume it would duplicate data like menu/footer links into a data-attribute?

Instead, could you maybe hide this functionality behind a query parameter? Or are you able to prove this wouldn't affect page performance?

EDIT: (Something else I thought of, some components receive large blocks of HTML govspeak so I assume it would duplicate those too?)

andysellick:
@AshGDS good point. I've done a couple of things to fix this...

- the output is now much more streamlined, and only includes the option names, not the values. The values aren't really that important, this is mainly so we can see which options are in use
- I've added the suppress flag to common options (like header, feedback) where we don't really care what options have been passed to them
- tested on a page with a lot of components, description updated to show result (TL;DR it only added about 3kB)

For components that take large blocks e.g. govspeak this wouldn't have been a problem because they use `yield` so the content isn't passed through the component wrapper helper, but the changes above will now mitigate this for any components where the content is passed as an option.

andysellick:
@AshGDS I've reworked this a bit, happy to discuss further though. It should be disabled on production and staging, although I'm not quite sure how to test that fully. PR description has been updated accordingly.

AshGDS (dismissed):
Cool, sounds good, thanks 👍

PR #5138, opened 2025-11-21
Change figure component design

andysellick:
## What
Change the design of the [figure component ](https://components.publishing.service.gov.uk/component-guide/figure), and update govspeak images to appear the same.

- move caption and credit beneath the image, previously to the side
- allow the image to be as wide as itself or the container
- make the caption always be the same width as the image
- use govuk secondary text colour on caption
- remove top border
- update examples

The component uses the CSS `display: table` for layout and `display: table-caption` for the caption, which forces it to match the width of the image. This means that the caption will always match the width of the image, even if the image is narrower than the column (I tried using flexbox and grid but couldn't make it work - added benefit is these are old CSS rules that are widely supported).

## Why
We're moving towards using larger images. The figure component was designed to display the image and caption side by side, which restricts the size of the image to half of a two-thirds column (which it normally occurs within).

## Visual Changes
Currently most of the images output by the figure component are approximately the same dimensions as the size they're displayed, so most of the examples below show an image now centred in a column. The intent in the future is to allow larger images in these spaces, which will fill the column.

Worked with @nnagewad on the design and build.

**Component guide**

 Before | After
--------|--------
Screenshot 2025-11-21 at 15 20 56 | Screenshot 2025-12-17 at 10 30 07

The figure component is currently only used in `frontend`, on case studies, fatality notice, news articles and speeches.

**[Case study](https://www.gov.uk/government/case-studies/the-rana-plaza-disaster)**

 Before | After
--------|--------
Screenshot 2025-11-21 at 15 23 19 | Screenshot 2025-12-17 at 10 33 39

Here's an [interesting example](https://www.gov.uk/government/case-studies/2013-elections-in-swaziland) - with this change the image on this page gets smaller, because the image is slightly less than half the width of the column but was set to width: 50%, making it appear slightly larger than it was.

**[Fatality notice](https://www.gov.uk/government/fatalities/corporal-mark-anthony-palin-killed-in-afghanistan)**

 Before | After
--------|--------
Screenshot 2025-11-21 at 15 26 15 | Screenshot 2025-12-17 at 10 34 31

**[News story](https://www.gov.uk/government/news/derailment-of-a-passenger-train-at-shap)**

 Before | After
--------|--------
Screenshot 2025-11-21 at 15 27 10 | Screenshot 2025-12-17 at 10 35 36

**[Speech](https://www.gov.uk/government/speeches/the-uk-calls-on-member-states-who-have-not-yet-done-so-to-join-the-majority-and-recognise-kosovos-independence-uk-statement-at-the-un-security-counc)**

 Before | After
--------|--------
Screenshot 2025-11-21 at 15 28 14 | Screenshot 2025-12-17 at 10 36 27

This one is interesting because the image is larger than it was being displayed. It now fills the 630px column (actual image width is 960px).

https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?selectedIssue=PNP-9713

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Change figure component design ([PR #5138](https://github.com/alphagov/govuk_publishing_components/pull/5138))
 * Adjust pagination component ([PR #5176](https://github.com/alphagov/govuk_publishing_components/pull/5176))
 * Guess `world_location_news` paths by `document_type` ([PR #5177](https://github.com/alphagov/govuk_publishing_components/pull/5177))
 * Add image card helper spec and refactor ([PR #5179](https://github.com/alphagov/govuk_publishing_components/pull/5179))


AshGDS (commented):
The code looks good 👍 

I was wondering if this has been checked from an accessibility perspective? Generally centering text is recommended against - here it might be OK because captions are short, but it might become an issue depending on the caption length. See: https://www.google.com/search?q=centred+text+wcag - also the caption loses some of its contrast when it goes from black to grey (it's still within AA compliance - but it slightly reduces the accessibility compared to black according to https://www.whocanuse.com/?bg=ffffff&fg=505a5f&fs=16&fw=) 

Also, when I test [this page](https://www.gov.uk/government/fatalities/corporal-mark-anthony-palin-killed-in-afghanistan) the figcaption is not being centred. It looks like because it's govspeak it doesn't get the `gem-c-figure` class - am I doing something wrong when testing?

andysellick:
@AshGDS we've reworked the design to avoid the accessibility issues raised:

- the caption is now left aligned
- the caption is the width of the image
- we're still using `govuk-secondary-text-colour` for the caption, but have increased the font size to match the body

I've also updated govspeak images to mimic the same styles, so we should have more consistency.

Regarding [the page you highlighted](https://www.gov.uk/government/fatalities/corporal-mark-anthony-palin-killed-in-afghanistan) the image is being centred because it has whitespace embedded either side of the image, see: https://assets.publishing.service.gov.uk/media/5a6218dced915d266017a84d/CorporalMarkPalin.jpg

app/views/govuk_publishing_components/components/docs/figure.yml
```diff
@@ -15,29 +15,30 @@ uses_component_wrapper_helper: true
...
+      src: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg
+      credit: Lewis Crankled
   figure_with_caption_and_credit:
     data:
-      src: 'https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg'
-      caption: The Lords Chamber
-      credit: Wallis Crankled
+      src: https://assets.publishing.service.gov.uk/government/uploads/system/uploads/feature/image/54374/s630_the_lords_chamber.jpg
+      caption: The Lords Chamber on the 14th of August. A large number of people are in attendance, filling the seats in the lower and upper areas. Several people are wearing hats.
+      credit: Lewis Crankled
```
AshGDS: Just curious on why the credit was changed? 😄 

(Looks like the real photographer is Roger Harris https://www.gov.uk/government/news/queens-speech-explainer)

AshGDS (approved):
LGTM 👍

PR #5153, opened 2025-12-02
Make govspeak chart's turquoise colour AAA compliant

AshGDS:
## What / Why

- Change the turquoise colour in the govspeak chart component to a lighter shade with black text so that it meets AAA compliance
- The old turquoise failed to meet the 4.5:1 contrast ratio against white text, so we needed to fix this 
- Initially we were going to update the shade to just be AA compliant, but we decided to go with AAA compliance instead as it's more inclusive as shown by https://www.whocanuse.com/?bg=1bbbaf&fg=000000&fs=16&fw=
- This change has been approved by our accessibility specialist
- This colour also keeps the contrast against the adjacent colours higher than 3:1 (thanks @jon-kirwan for spotting this initially so I could fix it)
- Preview app link: https://components-gem-pr-5153.herokuapp.com/component-guide/govspeak/chart_with_colours/preview
- Jira card: https://gov-uk.atlassian.net/browse/NAV-18538

## Visual Changes

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Update related navigation toggle texts for improved accessibility ([PR #5145](https://github.com/alphagov/govuk_publishing_components/pull/5145))
+* Make govspeak chart's turquoise colour AAA compliant ([PR #5153](https://github.com/alphagov/govuk_publishing_components/pull/5153))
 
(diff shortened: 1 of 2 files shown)


AshGDS:
Hi @jon-kirwan - this should be ready for review now

PR #5166, opened 2025-12-09
Add a new app promo banner component

MartinJJones:
## What
Add a new app promo banner component to help promote the GOV.UK app:
- the app promo banner will only be displayed on Android devices
- includes a link to the GOV.UK app in the Google Play Store
- add `app_promo_banner` as a `settings` cookie which is used to stop the banner reappearing once it has been closed

## Why
To help promote the GOV.UK app

Jira card: https://gov-uk.atlassian.net/browse/NAV-18626

## Visual Changes

Preview page: https://components-gem-pr-5166.herokuapp.com/public

#### 320px
mobile-320px

#### 412px
mobile-412px

#### 641px
tablet-641px


app/views/govuk_publishing_components/components/docs/app_promo_banner.yml
```diff
@@ -0,0 +1,30 @@
+name: App Banner Promo
```
jon-kirwan: Small detail, but I think we're using the name 'app promo banner' elsewhere, so we should probably keep the same naming convention here.

app/views/govuk_publishing_components/components/_app_promo_banner.html.erb
```diff
@@ -0,0 +1,64 @@
...
+        class: "gem-c-app-promo-banner__icon",
+        width: "64",
+        height: "64",
+      ) %>
+      <%= tag.div(class: "gem-c-app-promo-banner__content") do %>
+        <%= tag.div(class: "gem-c-app-promo-banner__text") do %>
+          <%= tag.h2 t("components.app_promo_banner.title_text"), class: "govuk-heading-s govuk-!-margin-bottom-0" %>
+          <%= tag.p t("components.app_promo_banner.body_text"), class: "govuk-body-s govuk-!-margin-bottom-0" %>
+        <% end %>
+        <a href="https://play.google.com/store/apps/details?id=uk.gov.govuk&hl=en_GB"
```
jon-kirwan: We could consider replacing this with a `link_to` helper, if only since we’re mostly using tag helpers elsewhere in the template.

config/locales/en.yml
```diff
@@ -13,6 +13,12 @@ en:
       hide_all: Hide all sections
       show_all: Show all sections
       this_section_visually_hidden: " this section"
+    app_promo_banner:
+      aria_label: GOVUK app promotion banner label text
```
jon-kirwan: Do we need to confirm this label text with content design?

jon-kirwan (commented):
Looks really good and working as expected. I’m still doing some testing but have added a few initial comments.

jon-kirwan (commented):
I noticed that when the close button is clicked, focus moves to the `<body>` element. I’m wondering if it might be unexpected. For TalkBack users, the skip link will already have been passed, so once the app promo banner component is closed, the next interactive elements should be in the header - either the header itself (if it’s focusable) or the logo.

app/views/govuk_publishing_components/components/docs/app_promo_banner.yml
MartinJJones: Good spot, I've updated this to be consistent

app/assets/images/govuk_publishing_components/app-promo-banner/govuk-app-icon.png
andysellick: This PNG is pretty small, but I wonder if it could be smaller - we know the background colour, so we could remove the alpha transparency and set the transparent corners to match that background.

Alternatively could it be an SVG?

app/assets/javascripts/govuk_publishing_components/components/app-promo-banner.js
```diff
@@ -0,0 +1,143 @@
...
+    startModule () {
+      if (!this.$closeButton) return
+
+      this.$closeButton.addEventListener('click', (event) => this.handleClick(event))
+
+      this.showBanner()
+    }
+
+    /**
+     * Checks if the current device is an Android device
```
andysellick: I'm a bit conflicted on the JSDoc - it feels like a useful addition, and I like that we're trying something new, but it makes the file longer to navigate and for functions like this it feels a bit unnecessary. Maybe we need to decide how to use it best, but this is a good first step.

app/views/govuk_publishing_components/components/_app_promo_banner.html.erb
```diff
@@ -0,0 +1,71 @@
...
+          <title><%= t("components.app_promo_banner.close") %></title>
+          <circle cx="12" cy="12" r="11.5" stroke="currentColor" />
+          <path stroke="currentColor" stroke-linecap="round" stroke-width="1.5" d="m7 7 10 10m0-10L7 17" />
+        </svg>
+      <% end %>
+      <%= image_tag((asset_path "govuk_publishing_components/app-promo-banner/govuk-app-icon.png"),
+        alt: "",
+        class: "gem-c-app-promo-banner__icon",
+        width: "64",
+        height: "64",
```
andysellick: If you're setting the width and height of this image here you don't need to also set it in the CSS. Debatable where it's better to put it - probably the CSS, so all the styling is in one place.

spec/components/app_promo_banner_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+describe "App promo banner", type: :view do
+  def component_name
+    "app_promo_banner"
+  end
+
+  it "renders the component" do
+    render_component({})
+
+    assert_select ".gem-c-app-promo-banner[aria-label='GOV.UK app']"
+    assert_select ".gem-c-app-promo-banner__close-button[data-module='ga4-event-tracker']", true
```
andysellick: Just interested - what's the difference between this line ending in `true` and the others in this test that don't have that?

andysellick (commented):
Looks good. I have a few comments, but they're mostly nitpicks.

I assume the last commit needs to be removed, and there should be a changelog entry 👍

app/assets/javascripts/govuk_publishing_components/components/app-promo-banner.js
```diff
@@ -0,0 +1,143 @@
+window.GOVUK = window.GOVUK || {}
+window.GOVUK.Modules = window.GOVUK.Modules || {};
+
+(function (Modules) {
+  class AppPromoBanner {
+    constructor ($module) {
+      this.$module = $module
+      this.$closeButton = this.$module.querySelector('.js-close-app-promo-banner')
+      this.APP_PROMO_BANNER = 'app_promo_banner'
```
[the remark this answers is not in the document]
MartinJJones: Good point, I've updated this to be more explicit by updating to `APP_PROMO_BANNER_COOKIE, this is also consistent with the naming used in [globar-bar.js](https://github.com/alphagov/govuk_publishing_components/blob/8438b112884b23418fb15b85e3fc8608024770dd/app/assets/javascripts/govuk_publishing_components/components/global-banner.js#L8)

app/assets/javascripts/govuk_publishing_components/components/app-promo-banner.js
```diff
@@ -0,0 +1,143 @@
...
+      this.startModule()
+    }
+
+    /**
+     * Starts the app banner module by adding an event listener to the
+     * close button and showing the banner
+     *
+     */
+    startModule () {
+      if (!this.$closeButton) return
```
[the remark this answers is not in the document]
MartinJJones: This is to help avoid any JavaScript errors when trying to attach the event listener to the close button, using the querySelector will either return the Element or null, for example:

Screenshot 2026-01-13 at 11 25 29

app/assets/javascripts/govuk_publishing_components/components/app-promo-banner.js
MartinJJones: Yeah, I definitely agree on this, I think it will need further discussion with the frontend community on the best approach

app/assets/javascripts/govuk_publishing_components/components/app-promo-banner.js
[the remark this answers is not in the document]
andysellick: I think the scenario you're talking about should be mitigated by tests, not by conditions within the code. If the button element is present but the code can't find it, a mismatch between the markup and the JS has been introduced, which should be highlighted to the developer by a test (although we have to write our own markup for the JS tests, it should be possible to get that checked in the ruby spec).

If you want to protect against this scenario you'd probably want to make the close button invisible until the JS executes and successfully finds it to attach the listener. Or attach the close button with JS. I guess the underlying question is which is worse: that the banner appears and the user can't close it, or that the banner doesn't appear?

app/views/govuk_publishing_components/components/_app_promo_banner.html.erb
```diff
@@ -0,0 +1,72 @@
...
+        aria: {
+          label: t("components.app_promo_banner.close"),
+        },
+        data: {
+          module: disable_ga4 ? nil : "ga4-event-tracker",
+          ga4_event: ga4_event_attributes,
+        },
+      }) do %>
+        <svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" fill="none" focusable="false">
+          <title><%= t("components.app_promo_banner.close") %></title>
```
jon-kirwan: Does the SVG need a title? Since it’s the same as the button text, should we consider adding `aria-hidden` to the SVG and removing its title?

app/assets/stylesheets/govuk_publishing_components/components/_app-promo-banner.scss
```diff
@@ -0,0 +1,58 @@
+@import "govuk_publishing_components/individual_component_support";
+
+.gem-c-app-promo-banner {
+  display: none;
+  @include govuk-text-colour;
```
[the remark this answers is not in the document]
MartinJJones: Good spot, I've moved this below the other declarations, it was causing sass compilation warnings as suspsected.

app/assets/stylesheets/govuk_publishing_components/components/_app-promo-banner.scss
```diff
@@ -0,0 +1,58 @@
...
+}
+
+.gem-c-app-promo-banner__container {
+  display: flex;
+  gap: govuk-spacing(3);
+  align-items: center;
+}
+
+.gem-c-app-promo-banner__icon {
+  display: inline-block;
```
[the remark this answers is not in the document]
MartinJJones: Good spot, I've removed this

jon-kirwan (approved):
All looks good to me - great work on this, @MartinJJones 👏

jon-kirwan (approved):

PR #5344, opened 2026-03-18
Reset `window.LUX` when user rejects cookies

hannalaakso:
## What

When usage tracking is disabled or withdrawn for cookies, we should not only stop sending analytics data but also clear any data stored in the LUX object. The implementation currently assumes that LUX is available at `window.GOVUK.LUX`, but this property does not exist. It's possible that something in our implementation or the API has changed.

 Reset `window.LUX` instead to attempt to fix the issue. However, this is a preliminary attempt at fixing this as it doesn't appear to be documented in [the RUM API](https://support.speedcurve.com/docs/rum-js-api#lux-properties) as a way to reset the usage data, and we're still gathering information on how to best handle this.

## Why

Jira card: https://gov-uk.atlassian.net/jira/software/c/projects/NAV/boards/1422?selectedIssue=NAV-18727

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@
 ## Unreleased
 
 * Remove reliance on CGI.parse ([PR #5424](https://github.com/alphagov/govuk_publishing_components/pull/5424))
+* Reset `window.LUX` when user rejects cookies ([PR #5344](https://github.com/alphagov/govuk_publishing_components/pull/5344))
 
 ## 66.0.0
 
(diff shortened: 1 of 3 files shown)


jon-kirwan:
Closing this issue, as unfortunately resetting `window.LUX` does not prevent data from being sent to SpeedCurve.

Instead, I’ll open a new PR to remove the existing code, and log a separate issue to investigate a separate fix - likely by refreshing the page after the cookie settings form is submitted, which should clear LUX data when consent is withdrawn.

PR #5356, opened 2026-03-24
Fix signup link having data='false' attribute

AshGDS:
## What/Why

- This component had `data ||= false` for setting data attributes. This meant that when no data attributes existed it would add `data='false'` to the HTML. Changing the value from `false` to `nil` should fix this.

## Visual Changes

A new example added to the signup link component.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


jon-kirwan (approved):
Looks good to me. Do we need a changelog entry too?

AshGDS:
@jon-kirwan Thanks, will do 👍

PR #5382, opened 2026-04-07
Change feedback component styles

andysellick:
## What
Restyle the feedback component. Specifically:

- adds the light blue background to the 'no' and 'report a problem with this page' forms (previously white) to make the whole component visually consistent (the unopened state and success state for 'yes' had the blue background already)
- use the button component for certain buttons, which removes the 'hidden' attribute from the 'cancel' button in the 'no' option, but this doesn't appear to do anything (?)
- use the secondary quiet style for the existing secondary buttons, as this improves their contrast against the new background
- adjust spacing around the component to fit in these changes, including removing margin bottom on the 'no' 'cancel' button, and setting margins on the 'send' and 'cancel' buttons to keep the spacing consistent with the current look

## Why
Better aligns with the rest of the page, and makes it more consistent with itself.

Been meaning to fix this for a while.

Also fixes https://gov-uk.atlassian.net/jira/software/c/projects/PNP/boards/1356?quickFilter=1456&selectedIssue=PNP-9898&useStoredSettings=true

## Visual Changes
Changes below shown for mobile and desktop. The only slight wrinkle is that the newly applied secondary quiet button style on 'cancel' lacks the bottom edge of the green button style, so side-by-side they're not the same height (on desktop). It's a very small difference though, and one that should probably be addressed separately on the button component rather than here.

Before | After
------ | -----
Screenshot 2026-04-08 at 17 13 26 | Screenshot 2026-04-08 at 17 13 39
Screenshot 2026-04-08 at 17 13 51 | Screenshot 2026-04-08 at 17 14 02

For context here it is on an actual GOV.UK page. First showing the 'report a problem' form:

Screenshot 2026-04-08 at 17 19 37

And now showing the 'no' form:

Screenshot 2026-04-08 at 17 19 47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Change feedback component styles ([PR #5382](https://github.com/alphagov/govuk_publishing_components/pull/5382))
+
 ## 65.2.2
 
 * Add missing Welsh translations for devolved nations content types ([PR #5371](https://github.com/alphagov/govuk_publishing_components/pull/5371))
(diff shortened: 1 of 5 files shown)


AshGDS (approved):
LGTM 👍

PR #5389, opened 2026-04-09
Add support for aggregate cookie tracking

MartinJJones:
## What

Add support for aggregate cookie tracking:

- Update the default cookie consent to include the new aggregate cookie and set to true by default
- Update `cookie-settings.js` used by the [cookie settings page](https://www.gov.uk/help/cookies) to support the new aggregate cookie option
- Update `init-ga4.js` [used to load analytics](https://github.com/alphagov/govuk_publishing_components/blob/9a17b4b4cf13dd989e99187f223f54ec39cf664e/docs/analytics-ga4/consent.md#ga4-and-cookie-consent) to set the default consent cookie if it isn't already present
- Update [javascript modules](https://github.com/alphagov/govuk_publishing_components/blob/9a17b4b4cf13dd989e99187f223f54ec39cf664e/docs/javascript-modules.md#modules-and-cookie-consent) to start when aggregate is true
- Update default cookie banner content

## Why

- To support the new aggregate cookie
- Updated `init-ga4.js` to set the default consent cookie if it isn't already present, as `aggregate: true` is included in the updated default consent cookie. This ensures analytics loads as expected for visitors who do not already have the `cookies_policy` cookie set

Jira cards:
- https://gov-uk.atlassian.net/browse/NAV-18724
- https://gov-uk.atlassian.net/browse/NAV-18754

### Preview links

Existing cookie banner: https://components.publishing.service.gov.uk/component-guide/cookie_banner/default
New cookie banner: https://components-gem-pr-5389.herokuapp.com/component-guide/cookie_banner/default

### Default Banner - Mobile

| Before | After |
| --- | --- |
| cookie-banner-mobile-before | cookie-banner-mobile-before |

### Default Banner - Desktop

| Before | After |
| --- | --- |
| cookie-banner-desktop-before | cookie-banner-desktop-after |

### Keep Current Settings - Mobile

| Before | After |
| --- | --- |
| mobile-reject-before | mobile-keep-after |

### Keep Current Settings - Desktop

| Before | After |
| --- | --- |
| desktop-reject-before | desktop-keep-after |


spec/javascripts/govuk_publishing_components/lib/cookie-functions-spec.js
```diff
@@ -120,6 +120,7 @@ describe('Cookie helper functions', function () {
       GOVUK.setDefaultConsentCookie()
 
       expect(GOVUK.getConsentCookie().usage).toBe(false)
+      expect(GOVUK.getConsentCookie().aggregate).toBe(true)
       expect(GOVUK.getCookie('JS-Detection')).toBeFalsy()
```
jon-kirwan, line 124: Just mentioning that for this test, which checks cookie deletion, we’ll probably need to update it to check the removal of another usage cookie once `'JS-Detection'` is removed and I’ve updated the card https://gov-uk.atlassian.net/browse/NAV-18747 for updating the cookie categories in the cookie functions to reflect this.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -3,14 +3,20 @@ window.GOVUK.analyticsGa4 = window.GOVUK.analyticsGa4 || {}
...
   var consentCookie = window.GOVUK.getConsentCookie()
 
-  if (consentCookie && consentCookie.usage) {
+  if (consentCookie && (consentCookie.usage || consentCookie.aggregate)) {
     window.GOVUK.analyticsGa4.vars.internalDomains = []
     window.GOVUK.analyticsGa4.vars.internalDomains.push(window.GOVUK.analyticsGa4.core.trackFunctions.getHostname())
     window.GOVUK.analyticsGa4.core.trackFunctions.appendDomainsWithoutWWW(window.GOVUK.analyticsGa4.vars.internalDomains)
     window.GOVUK.analyticsGa4.core.load()
 
+    // Initialise analytics modules that start on page load
```
jon-kirwan, line 19: You could also refer to the docs here: [/analytics-ga4/analytics.md#code-structure](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/analytics.md#code-structure), which provides a bit more detail on these types of modules.

app/assets/javascripts/govuk_publishing_components/lib/cookie-settings.js
```diff
@@ -28,10 +28,11 @@ window.GOVUK.Modules = window.GOVUK.Modules || {};
         delete currentConsentCookieJSON.essential
         for (var cookieType in currentConsentCookieJSON) {
           var radioButton
-          if (currentConsentCookieJSON[cookieType]) {
-            radioButton = document.querySelector('input[name=cookies-' + cookieType + '][value=on]')
+          if (cookieType === 'aggregate' && (currentConsentCookieJSON.aggregate === true && currentConsentCookieJSON.usage !== true)) {
```
jon-kirwan, line 31: One very small refactoring suggestion: there are a few conditions here, and I wonder if it would be worth extracting them into their own function.

Something like:
```js
if (CookieSettings.prototype.isAggregate()) {
  ...
}

CookieSettings.prototype.isAggregate {
  return cookieType === 'aggregate' && 
  (currentConsentCookieJSON.aggregate === true && currentConsentCookieJSON.usage !== true)
}
```

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-auto-tracker.spec.js
```diff
@@ -31,6 +31,15 @@ describe('Google Analytics auto tracker', function () {
       expect(tracker.startModule).toHaveBeenCalled()
     })
 
+    it('starts the module if aggregate consent has already been given', function () {
```
jon-kirwan, line 34: Small suggestion on naming consistency: Since the new test is very specific ('starts the module if aggregate consent...'), we should probably update the previous one to match. Maybe we could change 'when the user has a cookie consent choice' to 'starts the module if usage consent has been given. This might apply to the other tracker spec changes too.

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
```diff
@@ -21,14 +25,47 @@ describe('Initialising GA4', function () {
...
+
+    it('does not update the `cookie_policy` cookie to the default values if previously set', function () {
+      GOVUK.cookie('cookies_policy', ALL_COOKIE_CONSENT)
+      spyOn(window.GOVUK, 'setDefaultConsentCookie').and.callThrough()
+
+      GOVUK.analyticsGa4.init()
+
+      expect(window.GOVUK.setDefaultConsentCookie).not.toHaveBeenCalled()
+      expect(GOVUK.getCookie('cookies_policy')).not.toEqual(DEFAULT_COOKIE_CONSENT)
+      expect(GOVUK.getCookie('cookies_policy')).toEqual(ALL_COOKIE_CONSENT)
```
jon-kirwan, line 50: I think the expect on line 49 (`not.toEqual(DEFAULT_COOKIE_CONSENT)`) can possibly be removed. If we’re already asserting that the cookie equals `ALL_COOKIE_CONSENT` on line 50, I think that guarantees it’s not the default value.

spec/javascripts/govuk_publishing_components/lib/cookie-settings-spec.js
```diff
@@ -40,43 +55,77 @@ describe('cookieSettings', function () {
...
+
+    it('does not update the `cookie_policy` cookie to the default values if previously set', function () {
+      GOVUK.cookie('cookies_policy', ALL_COOKIE_CONSENT)
+      spyOn(window.GOVUK, 'setDefaultConsentCookie').and.callThrough()
+
+      new GOVUK.Modules.CookieSettings(element).init()
+
+      expect(window.GOVUK.setDefaultConsentCookie).not.toHaveBeenCalled()
+      expect(GOVUK.getCookie('cookies_policy')).not.toEqual(DEFAULT_COOKIE_CONSENT)
+      expect(GOVUK.getCookie('cookies_policy')).toEqual(ALL_COOKIE_CONSENT)
```
jon-kirwan, line 80: Same as my comment on the init-ga4 spec - I think the expect on line 79 (not.toEqual(DEFAULT_COOKIE_CONSENT)) can possibly be removed. If we’re already asserting that the cookie equals ALL_COOKIE_CONSENT on line 80, I think that guarantees it’s not the default value.

spec/javascripts/govuk_publishing_components/lib/cookie-settings-spec.js
```diff
@@ -93,6 +142,19 @@ describe('cookieSettings', function () {
         }
       }
     })
+
+    it('does not error if aggregate is not included in the cookies_policy', function () {
```
jon-kirwan, line 146: Maybe the test name should be updated to something like: 'sets usage radio options correctly when the aggregate key is missing from the policy'?

jon-kirwan (commented):
Great work - everything looks good and works as expected. I’ve only added a few minor suggestions.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
MartinJJones: Thanks for the suggestion, I've added a link to the docs

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
MartinJJones: Good spot, I've removed this from the test

spec/javascripts/govuk_publishing_components/lib/cookie-settings-spec.js
MartinJJones: Agreed, I've removed this from the test as well

spec/javascripts/govuk_publishing_components/lib/cookie-settings-spec.js
MartinJJones: That sounds good to me, I've made this change, thanks

spec/javascripts/govuk_publishing_components/analytics-ga4/ga4-auto-tracker.spec.js
MartinJJones: Yeah this is a nice improvement, I've made the change as suggested, I had a look at the other tests as well and the descriptions look OK for the others

app/assets/javascripts/govuk_publishing_components/lib/cookie-settings.js
MartinJJones: Yeah that makes sense to me, I've added a new function as suggested and added some new tests to check the output is as expected as well

jon-kirwan (commented):
Thanks for the changes @MartinJJones - I’ve taken another look and everything looks good to me 👍

(I'll hold off on approving for now in case there are further changes to the banner, or until we’re closer to when we plan to merge).

spec/javascripts/govuk_publishing_components/lib/cookie-functions-spec.js
MartinJJones: The `JS-detection` usage cookie was removed in https://github.com/alphagov/govuk_publishing_components/pull/5432, I've rebased this PR to get the latest changes as well

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -41,7 +43,23 @@
...
+            <% else %>
+              <p class="govuk-body"><%= t("components.cookie_banner.text_intro") %></p>
+              <ul class="govuk-list govuk-list--bullet">
+              <% t("components.cookie_banner.text_list").each do |item| %>
+                <li><%= item %></li>
+              <% end %>
+              </ul>
+              <p class="govuk-body"><%= t("components.cookie_banner.text_additional_cookies") %></p>
+            <% end %>
+            <p class="govuk-body govuk-!-margin-bottom-6">
```
deberny: I've checked the link spacing (bottom-margin) in the button group and it is always 20px on both mobile and desktop. So the `govuk-!-margin-bottom-6` class, applies a responsive margin 30px on desktop and 20px on mobile.

We could use a static class instead to keep the 20px spacing on any screen size `govuk-!-static-margin-bottom-4`

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -1,17 +1,19 @@
...
+      text = newtext
+    else
+      text = "<p class='govuk-body'>#{text}</p>"
     end
-    text = newtext
-  else
-    text = "<p class='govuk-body'>#{text}</p>"
+    text = raw(text)
   end
-  text = raw(text)
```
jon-kirwan, line 14: Thanks Martin - everything is looking great and working well.

Just one thought on this: as we discussed, it would probably make sense to move this into a helper as a longer-term home for the logic.

In the meantime, I wondered whether it might help to remove the `text` logic from the header (lines 5 - 16) and move it entirely into the main part of the view instead. Something along the lines of:

```rb
text ||= ...
```

```erb
<% if text.present? %>
  <% Array(text).each do |content| %>
    <p class="govuk-body"><%= content %></p>
  <% end %>
<% else %>
  <p class="govuk-body"><%= t("components.cookie_banner.text_intro") %></p>
  <ul class="govuk-list govuk-list--bullet">
    <% t("components.cookie_banner.text_list").each do |item| %>
      <li><%= item %></li>
    <% end %>
  </ul>
  <p class="govuk-body"><%= t("components.cookie_banner.text_additional_cookies") %></p>
<% end %>
```

This seems to work and handles single-line text, arrays of text, and custom content with embedded HTML links.

MartinJJones: Thanks, this looks much cleaner to me, I've updated the template with the suggested changes

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
MartinJJones: I've updated the spacing to use the static class and keep the spacing at 20px on any screen size.

I've updated the before and after screenshots included in the PR description, but you can also preview the changes at - https://components-gem-pr-5389.herokuapp.com/component-guide/cookie_banner/default

Just let me know if this looks OK or if any further changes are required, thanks

deberny: Looks great! 🚀

PR #5390, opened 2026-04-10
Update install doc

### LINKED ISSUE #4134: Remove `all_components.js` and `all_components.scss`
## What
Remove `all_components.js` and `all_components.scss`

## Why
To avoid excessive code from being imported in the application.

Both [all_components.js](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/install-and-use.md#import-all-sass-deprecated-will-be-removed-in-a-later-version) and [all_components.scss](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/install-and-use.md#import-all-javascript-deprecated-will-be-removed-in-a-later-version) files are marked as "(deprecated, will be removed in a later version)" in the install and use docs.

### Further info
- `all_components.js` marked as deprecated - https://github.com/alphagov/govuk_publishing_components/pull/1472
- `all_components.scss` marked as deprecated - https://github.com/alphagov/govuk_publishing_components/pull/1321
- [RFC108 - Include specific component assets in applications](https://github.com/alphagov/govuk-rfcs/blob/bc1ad0677abcbfdf48ba6bc0b2e0f24557e48e9a/rfc-108-including-gem-component-assets.md)

andysellick:
## What
- split 'Include the assets' into two sections, one for Sass and one for JS
- move the 'import all JavaScript' section to immediately follow import for individual components, to show that it is the alternative
- change heading structure a bit for the following JS sections on bundles and ES6
- remove the 'deprecated' notice on the 'import all JS' section

## Why
Partly to improve readability.

Partly to fix https://github.com/alphagov/govuk_publishing_components/issues/4134

We were going to remove the deprecated `all_components` Sass and JS files but I don't think we should.

I think they're still useful, and since they're still in place for a few applications (including the gem itself, which uses the Sass one for the component guide) removing them would be an unnecessary breaking change to fix. 

## Visual Changes
None.

### DIFF
--- a/docs/install-and-use.md
+++ b/docs/install-and-use.md
@@ -45,9 +45,7 @@ GovukPublishingComponents.configure do |c|
 end
 ```
 
-## 4. Include the assets
-
-To make use of the components, you'll have to include the JS and CSS in your application.
+## 4. Include Sass
 
 ### Import Sass for specific components (recommended)
 
@@ -86,7 +84,9 @@ Alternatively, you can simply import the Sass for all of the components. This is
 @import "govuk_publishing_components/all_components";
 ```
 
-### Import JavaScript for individual components
+## 5. Include and configure JavaScript
+
+### Import JavaScript for individual components (recommended)
 
 ```ruby
 # application.js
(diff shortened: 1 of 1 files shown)


MartinJJones (approved):
Nice update, the changes look good and make sense to me 👍

PR #5392, opened 2026-04-13
Expand audit table

andysellick:
## What
- use CSS container queries to allow the component files table in the auditing pages to expand to fill the page width, when the page is above 1040px wide

## Why
I wanted to use container queries to experiment with tables. This table has been quite cramped for a while and I wanted to see if something like this could be possible. I don't fully understand the maths (I've had to add an extra pixel to the calculation to get it exactly right) and below a screen size of 1040px it doesn't work properly, but overall it seems like an improvement.

On browsers that don't support container queries it should just look as it did before.

## Visual Changes
Table now expands outside of the width of the page wrapper to meet the edge of the browser viewport. I've added some pseudo elements to style a 'box' around the whole thing.

Screenshot 2026-04-13 at 13 39 36

### DIFF
--- a/app/assets/stylesheets/component_guide/application.scss
+++ b/app/assets/stylesheets/component_guide/application.scss
@@ -477,6 +477,57 @@ $code-delete-bg: #fadddd;
   }
 }
 
+.table-container {
+  container-type: inline-size;
+  container-name: page;
+}
+
+@container page (width > 1040px) {
+  $w: 960px;
+
+  .table-wrapper {
+    position: relative;
+    z-index: 0;
+    box-sizing: border-box;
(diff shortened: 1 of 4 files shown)


AshGDS (commented):
I haven't looked at the code yet but it renders like this for me. I think it's because I always have scrollbars visible on the browser, which I think adds some extra pixels to the page width https://www.smashingmagazine.com/2023/12/new-css-viewport-units-not-solve-classic-scrollbar-problem/#the-two-types-of-scrollbars

Image

andysellick:
Thanks @AshGDS I'll take another look.

andysellick:
@AshGDS should be fixed now.

AshGDS (approved):
LGTM 👍

Issue #5122, opened 2025-11-05
Some tables in govspeak are hard to navigate


andysellick:
We're still investigating this issue but here are some initial thoughts.

## Problem

Content publishers are using govspeak tables and in some cases the data is too tall or too wide (or both) for the page. This is hard to navigate and causes some specific further problems:

- scrollbars on the bottom of tables aren't always visible or easy to get hold of (because they're so thin)
- if the table is too wide and too tall, users can't easily find the scrollbar along the bottom of the table, having to scroll a lot to get to it
- there's a further orientation issue with this - if you have to scroll many pages to move the table across, then back up again to read

Some example pages (thanks to @selfthinker for supplying many of these):

- [page with a long table](https://www.gov.uk/government/publications/dbs-privacy-policies/cookies-policy)
- [page with a reeeally long table](https://www.gov.uk/government/publications/skilled-worker-visa-eligible-occupations/skilled-worker-visa-eligible-occupations-and-codes)
- [page with a wide (and tall) table](https://www.gov.uk/government/publications/skilled-worker-visa-going-rates-for-eligible-occupations/skilled-worker-visa-going-rates-for-eligible-occupation-codes)
- [table used for layout, a calculation formula](https://www.gov.uk/government/publications/16-to-19-funding-allocations-supporting-documents-for-2022-to-2023/16-to-19-further-education-allocation-calculation-toolkit-2022-to-2023--2)
- [table without any headers](https://www.gov.uk/government/publications/agri-tech-catalyst-innovation-funding-in-developing-countries/competition-guidance-agri-tech-catalyst-round-6-industrial-research-awards)
- [tables that include an empty row](https://www.gov.uk/guidance/manual-of-patent-practice-mopp/section-1-patentability)
- [single column tables used for paragraphs](https://www.gov.uk/guidance/ofqual-handbook/section-h-from-marking-to-issuing-results)
- [table used as a list](https://www.gov.uk/government/publications/an-in-depth-review-of-fire-related-fatalities-and-severe-casualties-in-england-2010-to-201819/an-in-depth-review-of-fire-related-fatalities-and-severe-casualties-in-england-201011-to-201819#appendix-b)
- [table used as layout](https://www.gov.uk/government/consultations/arrangements-for-the-assessment-and-awarding-of-vocational-technical-and-other-general-qualifications-2021-to-2022/vocational-and-technical-qualifications-contingency-regulatory-framework-covid-19-guidance-draft-for-consultation)
[...]

PR #5397, opened 2026-04-14
[DO NOT MERGE] Limit govspeak table heights

andysellick:
## What
Limits the height of tables within govspeak to within the screen height.

On some browsers/hardware combinations scrollbars do not appear by default, notably iOS touch devices. This is already a problem for some existing tables where the width exceeds the page width. 

So we set overflow to scroll to force some browsers to display the scrollbar (downside - it will always be visible even if scroll is not needed). We also style the scrollbar using `webkit-scrollbar` pseudo selectors. This ensures the scrollbars appear for a few more browsers (including Chrome on iOS desktop).

You can also use [scrollbar-width](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-width) and [scrollbar-color](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/scrollbar-color) but these seem to both override the webkit selectors, and then be ignored by those devices (so the scrollbars don't appear anyway). Better to go for the option that displays the scrollbars on the most browsers.

The webkit scrollbars are styled to match scrollbars on other browsers although we take the opportunity to increase their width. Oddly this only seems to apply to the vertical scrollbar, but it makes it easier to use.

Note that this PR includes consideration for print styles, which would be effected by this change.

## Why
Many tables are too wide to fit on a screen. If they are also too tall, users cannot see the scrollbar as it is at the bottom of the table. This change forces the table to be no taller than the screen, so users can more easily see the scrollbar along the bottom, if it exists.

Doing some thinking around https://github.com/alphagov/govuk_publishing_components/issues/5122

## Visual Changes
Firefox.

Before | After
---- | -----
Screenshot 2026-04-22 at 13 41 09 | Screenshot 2026-04-22 at 13 41 19

Chrome (desktop)
Before | After
---- | -----
Screenshot 2026-04-22 at 13 42 25 | Screenshot 2026-04-22 at 13 42 33

iPhone (still doesn't work, but we have the scroll right/width problem already)

Screenshot 2026-04-22 at 13 45 35

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_tables.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_tables.scss
@@ -12,9 +12,24 @@
     border-spacing: 0;
     display: block;
     margin: 0 0 govuk-spacing(6) 0;
-    overflow-x: auto;
+    overflow-x: scroll;
+    overflow-y: scroll;
     width: 100%;
+    max-height: 75vh;
     @include govuk-font($size: 19);
+
+    &::-webkit-scrollbar {
+      -webkit-appearance: none;
+      width: 20px;
+      background-color: #fafafa;
+      border: solid 1px #e7e7e7;
+    }
+
+    &::-webkit-scrollbar-thumb {
+      border-radius: 4px;
+      background-color: #c2c2c2;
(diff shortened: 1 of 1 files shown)


andysellick:
Going to close this as I think https://github.com/alphagov/govuk_publishing_components/pull/5396 is a better option.

PR #5417, opened 2026-04-21
Improve applications page layout

andysellick:
## What / why
Small improvements to the applications status page:

- add `govuk_publishing_components` to the list of applications, so we can also see what versions of Ruby etc. the gem itself is using
- use the expanding table styles from the auditing pages here as well, and improve them

## Visual Changes
Before

Screenshot 2026-04-21 at 10 50 21

After

Screenshot 2026-04-21 at 10 50 31


MartinJJones (approved):

PR #5420, opened 2026-04-21
Add default cooldown to dependabot config

sihugh:
I haven't specified exclusions for internal dependencies, as I expect these will generally not be immediately urgent on this repo. It's easily tweaked at a later date if deemed necessary.

I haven't changed the [version update schedule to the suggested monthly](https://docs.publishing.service.gov.uk/manual/manage-dependencies.html#schedule-and-cooldown) because I am not usually involved with maintaining this repo. It might be worth picking up though (or I can change this PR if it would be useful)

NB security updates are not affected by dependabot cooldown configuration.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -4,11 +4,17 @@ updates:
     directory: "/"
     schedule:
       interval: daily
+    cooldown:
+      default-days: 3
   - package-ecosystem: npm
     directory: "/"
     schedule:
       interval: daily
+    cooldown:
+      default-days: 3
   - package-ecosystem: "github-actions"
     directory: "/"
     schedule:
       interval: daily
+    cooldown:
+      default-days: 3


maxgds:
I think govspeak could be something we'd need excluded, not sure about other internal things, but suggest we add that exclusion.

sihugh:
> I think govspeak could be something we'd need excluded, not sure about other internal things, but suggest we add that exclusion.

Looks like we no longer have that as a dependency: https://github.com/alphagov/govuk_publishing_components/pull/827

maxgds:
> > I think govspeak could be something we'd need excluded, not sure about other internal things, but suggest we add that exclusion.
> 
> Looks like we no longer have that as a dependency: #827

Yup, far less entangled than I thought we still were.

MartinJJones (approved):
Thanks for adding this in, we discussed further at one our dev catchups today and agreed to review and update the interval as needed in the future

PR #5424, opened 2026-04-22
Remove reliance on CGI.parse

### LINKED ISSUE #5423: Step by Step nav helper incompatible with Ruby 4
Ruby 4 appears to remove the CGI.parse method used here: https://github.com/alphagov/govuk_publishing_components/blob/fb83c74c9a47f55109da18488ddc11a5591245bd/lib/govuk_publishing_components/presenters/step_by_step_nav_helper.rb#L110

...which means pages trying to render Step by Step nav components crash.

andysellick:
## What
Remove code relying on `CGI.parse` as it has been removed from Ruby 4.

It was used in the step nav helper. The helper outputs links in the step nav component and appends the ID of the current step nav onto those links as URL params, e.g. `gov.uk/page?step-by-step-nav=ehc6wjf7fgsi`

To do this correctly the method first checks for existing URL params, so that it doesn't omit or break them when appending the new one. It's not clear where this occurred (very few GOV.UK pages include URL params outside of step by step navs) but it's possible they could occur so this seems like valid checking (maybe step navs could contain links to search pages, which could have URL params).

There wasn't a test in place for this so I've added one, and it turns out that when using `CGI.parse` it didn't quite work correctly - the resulting object contained attributes of single strings in arrays, instead of single strings, which when output included `%5B%5D` (which is `[]` when encoded for URLs) e.g. `href=\"/link1?another%5B%5D=one&amp;existing%5B%5D=url-param`.

This behaviour has now been corrected. Note that the link converts `&` characters into `&amp;` but browsers correctly interpret this as `&`.

## Why
We want to upgrade to Ruby 4 at some point and this was preventing that.

Fixes https://github.com/alphagov/govuk_publishing_components/issues/5423

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Remove reliance on CGI.parse ([PR #5424](https://github.com/alphagov/govuk_publishing_components/pull/5424))
+
 ## 66.0.0
 
 * Change feedback component styles ([PR #5382](https://github.com/alphagov/govuk_publishing_components/pull/5382))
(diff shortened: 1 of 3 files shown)


KludgeKML (approved):
LGTM

PR #5425, opened 2026-04-23
Fix error in `Ga4FormTracker`

patrickpatrickpatrick:
## What

Prevent a check of `dataset` of `fieldset` if `fieldset` is `null`.

## Why

Was resulting in an error preventing form submit being tracked on specific pages without `fieldset` in a tracked `form`.


davidtrussler (approved):
LGTM 👍

PR #5426, opened 2026-04-27
Expand output in rspec failures

andysellick:
## What
Modify spec helper to output the full diff when an RSpec test fails.

## Why
If the outputted value for `expected` or `got` exceeds a certain number of characters it is automatically truncated. This means that for large diffs it is sometimes impossible to see what is actually different. This change removes that restriction so that the full diff is always shown.

## Visual Changes
Before (note how the expected/got at the bottom of the screenshot are shorter and the middle contracts to `..`)

Screenshot 2026-04-27 at 08 40 26

After - full expected/got are output

Screenshot 2026-04-27 at 08 44 29

### DIFF
--- a/spec/spec_helper.rb
+++ b/spec/spec_helper.rb
@@ -20,6 +20,10 @@ end
 WebMock.disable_net_connect!(allow_localhost: true)
 
 RSpec.configure do |config|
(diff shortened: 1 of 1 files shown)


KludgeKML (approved):
LGTM!

PR #5430, opened 2026-04-27
Remove data-ga4-track-links-only

### LINKED ISSUE #5350: Can we remove data-ga4-track-links-only from our GA4 link tracker?
After updating the docs in https://github.com/alphagov/govuk_publishing_components/pull/5348, I realised that it might be possible to remove `data-ga4-track-links-only` entirely. I believe we built the link tracker under the assumption that the link tracker could be used on elements that aren't `a` tags -[ preserving Universal Analytics functionality.](https://github.com/alphagov/govuk_publishing_components/pull/2263/changes/33d5ea3836b8e9e7a0ac2d6c25de3a5ce3f850ca)  This is true in one instance, where we use it on a `button` element. 

Therefore, we added `data-ga4-track-links-only` when we needed to specifically limit tracking to `a` tags, which only happens for us when we put the tracker on a parent element (like a parent `govspeak` div) so that when the user clicks on the element, only `a` tags trigger the tracking all the way through. 

However, I'm pretty sure we only ever use this tracker when clicking `a` tags across GOV.UK, and in one specific instance of a `button` element. We also actually [look specifically for an `a` tag later in the code](https://github.com/alphagov/govuk_publishing_components/blob/5478dfdfba1bd4670c9c58b21ae280a2c687bc02/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js#L128), and [cancel tracking entirely if no `href` exists on the element](https://github.com/alphagov/govuk_publishing_components/blob/5478dfdfba1bd4670c9c58b21ae280a2c687bc02/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js#L100). 

As mentioned, you _can_ add the tracking to a non-link element, but you have to specify a `url` in the GA4 JSON, which I think we only added for [one case with the `<button>` element](https://github.com/alphagov/govuk_publishing_components/commit/c50aaa667ec8902754af189fca466e008ca78099) - though I can't seem to find where we are using this via GitHub search so those pages may have been removed since. 

This supports the idea that 99% of the time this tracker is used on `a` tags.  Therefore we could probably just add a conditional early on in the tracker code to abort tracking if the clicked element isn't an `a` or `button` tag, which would remove the need for `data-ga4-track-links-only` entirely. If other element types are needed to be tracked in the future they could just be added to the conditional. 

Therefore, it might be worth removing this functionality from the code which will simplify things wherever we add the link tracker?

It could be a breaking change if we just removed it, so we could do a two phased approach (e.g. make the code work without needing `data-ga4-links-only`, then remove the attribute from our views, and then remove the functionality once the attribute is no longer in our views)

AshGDS:
## What/Why

- All that `data-track-links-only` seems to do in our code is check that the thing being clicked on is an `<a>` tag (see [here](https://github.com/alphagov/govuk_publishing_components/blob/f2e56d77a95446030adeae939d85d02d2f2dd77f/app/assets/javascripts/govuk_publishing_components/analytics-ga4/ga4-link-tracker.js#L44)). This isn't really necessary as I believe the link tracker's always used on `<a>` elements.
- Therefore we can remove this attribute - all we need to do is check that the event target is an `<a>` tag as the tracker's default behaviour, and everything should work as normal.
  - (I have kept the link tracker working if a `button` is clicked that also has `data-ga4-link` - mainly because there was one instance that I can no longer find where we had to force a `<button>` to work with the link tracker.)
- Some extra context is in https://github.com/alphagov/govuk_publishing_components/issues/5350
- Closes https://github.com/alphagov/govuk_publishing_components/issues/5350
- Jira card: https://gov-uk.atlassian.net/browse/PNP-9954
- Once this PR is merged, there would be some work to do to remove the attribute from all the frontend repos, as it would no longer be doing anything useful.

## Pages tested

I've tested the following pages locally that all have `data-ga4-track-links-only`. They all seem to be working OK:

### Frontend
- Featured links on the homepage
- Administration and finance section on /government/collections/statutory-guidance-schools
- Get emails alert for all countries on /foreign-travel-advice
- Get email alerts on /foreign-travel-advice/azerbaijan
- The links on /help
- Recently opened on /government/get-involved
- Summary box on /government/consultations/post-16-l2-and-l3-qualifications-2027-proposed-subject-content
- Applies to England, Scotland, Wales on government/publications/car-show-me-tell-me-vehicle-safety-questions/car-show-me-tell-me-vehicle-safety-questions
- Contents links on /guidance/the-highway-code
- Useful websites on /guidance/the-highway-code/other-information
- What worked well /guidance/systems-change-learning/1-securing-strategic-buy-in-and-alignment
- Accordion on guidance/systems-change-learning/1-securing-strategic-buy-in-and-alignment#fn:1
- Explore the topic on /government/news/the-personal-independence-payment-amendment-regulations-2017-statement-by-paul-gray
- Get notifications on /service-manual/service-standard
- Get notifications on /service-manual/helping-people-to-use-your-service

### Collections

- Org list on  /government/organisations
- Image cards on  /government/organisations/the-adjudicator-s-office
- Most viewed links on  /browse/driving
- Get emails link on /government/ministers/secretary-of-state-for-education
- Org links on /government/organisations/companies-house
- Driving and motorcycle tests on  /transport/driving-and-motorcycle-tests
- People page on /government/people/matthew-hancock
- Topical Event on /government/topical-events/autumn-statement-2022
- World location news on /world/ukraine/news
- Survey banner on /world/ukraine

### Finder-frontend

- Get emails and Subscribe to feed on /administrative-appeals-tribunal-decisions
- Survey on /tax-and-chancery-tribunal-decisions

## Visual Changes
None.


andysellick (approved):
Looks good, very thorough 👍

PR #5437, opened 2026-04-30
Fix details underline issue

### LINKED ISSUE #5436: Details element has intermittent underline
Under certain conditions, an open details component has an underline beneath it that shouldn't be there (although the reproducing the issue is a little erratic). It looks like this:

Image

You can achieve this by clicking to open the element (triggering the focus state) then clicking elsewhere to remove the focus state.

Image

It seems to be dependent on the content of the details element - I couldn't reproduce it in the component guide, only in `frontend` in the examples shown.

It seems to be related to the `box-shadow` on the `focus` state of the `.govuk-details__summary` element itself, which is the usual Design System focus style. If I focus the element and modify the styles to have an orange `box-shadow` and increase the height, when I blur the element and reproduce the problem it looks like this:

Image

...so it appears that the mystery line is an artefact of the focus state, which suggests a browser bug (this occurs in Chrome, can't reproduce in Firefox or Safari).

I can fix it (about to raise a PR) by adding a transparent `box-shadow` to the non-focus state of the element, although it feels a bit clunky. Thought it worth documenting in an issue.

andysellick:
## What
Apply a transparent box shadow to the default state of the details summary element.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/5436

## Visual Changes
See linked issue.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_details.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_details.scss
@@ -33,3 +33,8 @@
     @include govuk-link-style-inverse;
   }
 }
+
+// fix for issue https://github.com/alphagov/govuk_publishing_components/issues/5436
+.govuk-details__summary {
+  box-shadow: 0 4px transparent;
+}


AshGDS (approved):

NickColley:
@andysellick do you know if this is something that makes sense to upstream into GOV.UK Frontend?

andysellick:
@NickColley possibly, although I'm not proud of the solution - seems like if it's a browser bug we should figure out a way of consistently reproducing it then report it to Google.

PR #5448, opened 2026-05-05
Migrate 'main navigation' component from frontend

AshGDS:
## What/Why
- Migrate the "main navigation" component from `frontend` into the gem
- I've called it the "Menu" component here - I would have called it "Secondary navigation menu" but we have a "Secondary navigation" component already, so was trying to avoid confusion. I'm open to suggestions on a new name. Some ideas are `Sub menu`, `Child navigation`, `Section menu`, `Topic menu`
  - There's a similar naming issue to resolve as the accessible `<h2>` for this component says `Secondary navigation menu`, which again is probably confusing in combination with the Secondary navigation component
- See https://components-gem-pr-5448.herokuapp.com/component-guide/menu
- Jira card: https://gov-uk.atlassian.net/browse/PNP-9971

## Visual Changes

New component added:

image


AshGDS:
@andysellick I think the code is ready to review, we probably need to agree on a name for the component though.

app/assets/javascripts/govuk_publishing_components/components/menu.js
```diff
@@ -0,0 +1,29 @@
+/* istanbul ignore next */
+window.GOVUK = window.GOVUK || {}
+/* istanbul ignore next */
+window.GOVUK.Modules = window.GOVUK.Modules || {};
```
andysellick: Can these lines be safely removed now?

Also I think the `istanbul` lines should also be removed, they're to let the test coverage ignore the next line, don't think it should be in front of the entire function (line 5).

app/assets/javascripts/govuk_publishing_components/components/menu.js
```diff
@@ -0,0 +1,29 @@
...
+/* istanbul ignore next */
+(function (Modules) {
+  'use strict'
+
+  function Menu (module) {
```
andysellick: We've not had a general conversation about adopting a more modern JS style for these modules (and I'm not an expert) but I think the structure can now be more like the below.

```JavaScript
(function (Modules) {
  class Menu {
    constructor ($module) {
    }

    init () {
    }

    anotherFunction(variable) {
    }
  Modules.Menu = Menu
})(window.GOVUK.Modules)

```

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
```diff
@@ -0,0 +1,148 @@
+.govuk-block.govuk-block__main_navigation {
```
andysellick: Is this class still needed? I can't see it anywhere else in the component.

If it's come from `frontend`, does it need to stay there?

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
```diff
@@ -0,0 +1,148 @@
...
+  background-color: govuk-colour("light-grey");
+  border-bottom: 1px solid govuk-colour("mid-grey");
+}
+
+.gem-c-menu__button-container--collapsed {
+  border-bottom: 0;
+}
+
+.gem-c-menu__button {
+  height: 70px;
```
andysellick: I'd generally avoid a hard coded pixel height for elements when considering zooming (particularly zoom text only in Firefox, where this causes a small visual issue, although admittedly only at very high zooms). Is it possible to either switch this to use an em value, or use padding top and bottom of the element instead?

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
```diff
@@ -0,0 +1,70 @@
...
+      - text: "Example Heading"
+        links:
+          - text: "Hello world"
+            href: "/hello-world"
+      - text: "Another Heading"
+        links:
+          - text: "Example Link"
+            href: "/hello-world"
+
+  negative_margin_left_on_button:
```
andysellick, line 55: Is this a new addition? I don't remember it being in use anywhere. I don't know the background, but for such a small design change, should it either be the default or not an option?

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
```diff
@@ -0,0 +1,148 @@
...
+  border-right: 2px solid govuk-colour("blue");
+  content: "";
+  display: inline-block;
+  height: 8px;
+  margin: 0 10px 0 2px;
+  vertical-align: middle;
+  width: 8px;
+}
+
+.gem-c-menu__button:focus {
```
andysellick: Presumably just including `@include govuk-focused-text;` here isn't enough?

app/assets/stylesheets/govuk_publishing_components/components/_sub-navigation-menu.scss
```diff
@@ -0,0 +1,148 @@
...
+  &.gem-c-menu__button::before {
+    border-color: govuk-colour("black");
+  }
+}
+
+.gem-c-menu__button[aria-expanded="false"]::before {
+  // Collapsed arrow icon
+  -webkit-transform: translateY(-35%) rotate(45deg) scale(1);
+  -ms-transform: translateY(-35%) rotate(45deg) scale(1);
+  transform: translateY(-35%) rotate(45deg) scale(1);
```
andysellick, line 55: [Caniuse](https://caniuse.com/transforms2d) seems to suggest that `transform` is widely supported - are the fallbacks needed here?

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
```diff
@@ -0,0 +1,70 @@
+name: Menu
+description: The menu component helps users to navigate pages within a section of GOV.UK
+
+accessibility_criteria: |
```
andysellick, line 4: You could include the shared criteria for links here as a shortcut:

```
  - link
```

app/views/govuk_publishing_components/components/_menu.html.erb
```diff
@@ -0,0 +1,65 @@
...
+  button_classes = %w[gem-c-menu__button gem-c-menu__button--no-js govuk-link govuk-link--no-underline]
+  button_classes << "gem-c-menu__button--negative-margin-left" if negative_margin_left_on_button
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-menu")
+  component_helper.add_data_attribute({ module: "menu" })
+%>
+<% unless links.empty? %>
+  <%= tag.div(**component_helper.all_attributes) do %>
+    <h2 class="govuk-visually-hidden" id="gem-c-menu-heading">Secondary navigation menu</h2>
```
andysellick: I think this was already accessibility tested but presumably this H2 would occur before the H1 on a normal page - is that okay? If it's visually hidden anyway, would it be better as an `aria-label` on the element? (I can see it has a labelledby pointing to the h2 though)

spec/components/menu_spec.rb
```diff
@@ -0,0 +1,148 @@
...
+
+  it "does not render anything if nothing is in the items array" do
+    test_data = { links: [] }
+
+    assert_empty render_component(test_data)
+  end
+
+  it "renders unordered lists with headings" do
+    render_component({
+      links: [
```
andysellick: This data repeats a few times - could probably use a global variable for some of it?

andysellick (commented):
Nothing really wrong, mainly just thoughts and questions. I am also aware that this is kind of a temporary measure and that this component is subject to change. Given that, is it worth marking as experimental?

No good ideas on the name I'm afraid - sub navigation menu is alright, maybe in page navigation?

app/assets/javascripts/govuk_publishing_components/components/menu.js
AshGDS: Good spot - will remove, thanks

app/assets/javascripts/govuk_publishing_components/components/menu.js
AshGDS: Good idea, I'll try and convert it 👍

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
AshGDS: Good spot, I did remove it but then I think it got reintroduced as I was squashing/cleaning up commits 😄

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
AshGDS: On the missions pages we added this negative margin so that the chevron aligned with the content around it (since it's a `:before` psuedo element). I thought I'd separate it out into an option as I assume it may not be needed in every situation. (Sorry, should've documented that somewhere when committing the change)

| Without negative margin | With negative margin |
|--------|--------|
| image | image |

app/assets/stylesheets/govuk_publishing_components/components/_sub-navigation-menu.scss
AshGDS: You're probably right - I likely took this code from one of our existing components when building this, so it might be worth looking into that for all our chevrons

app/views/govuk_publishing_components/components/_menu.html.erb
AshGDS: Yeah, I think it is intentional, and meant to replicate how our main navigation is built. This is the missions page heading structure:

image

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
AshGDS: Thanks, switched to em 👍

app/assets/stylesheets/govuk_publishing_components/components/_menu.scss
AshGDS: Good spot, done

app/assets/stylesheets/govuk_publishing_components/components/_sub-navigation-menu.scss
AshGDS: Removed 👍

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
AshGDS: Thanks, done 👍

spec/components/menu_spec.rb
AshGDS: Done, thanks

AshGDS:
@andysellick This should be ready for review again - I went for `Sub navigation menu` for the name, `In page navigation` sounded like it would only apply to navigating around the same page. I've marked it as experimental too 👍

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
andysellick: That's fine, can you clarify where it's used? Is it not used on the main page but used on the sub pages?

app/views/govuk_publishing_components/components/_sub_navigation_menu.html.erb
```diff
@@ -0,0 +1,65 @@
...
+  button_classes = %w[gem-c-sub-navigation-menu__button gem-c-sub-navigation-menu__button--no-js govuk-link govuk-link--no-underline]
+  button_classes << "gem-c-sub-navigation-menu__button--negative-margin-left" if negative_margin_left_on_button
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-sub-navigation-menu")
+  component_helper.add_data_attribute({ module: "sub-navigation-menu" })
+%>
+<% unless links.empty? %>
+  <%= tag.div(**component_helper.all_attributes) do %>
+    <h2 class="govuk-visually-hidden" id="gem-c-sub-navigation-menu-heading">Secondary navigation menu</h2>
```
andysellick: It's a bit more work and unlikely to happen in the wild, but currently this hard coded id is causing an accessibility warning in the component guide for duplicate IDs. You can use SecureRandom in this template to generate a unique id, then intercept the method in the test to check the ID is present.

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
```diff
@@ -0,0 +1,65 @@
+name: Sub navigation menu (experimental)
+description: The sub navigation menu component helps users to navigate pages within a section of GOV.UK
+
+accessibility_criteria: |
+  The menu component must:
+
+    * be recognisable as navigation
+    * have correctly associated labels
```
andysellick: Wasn't immediately obvious to me what this criteria means - can it be a bit more specific? What labels does it refer to?

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
AshGDS: It's used on all the `/missions` pages - [it's part of the original design](https://github.com/alphagov/frontend/blob/563413c9532bc8ee3a60c0b62853cd5532399828/app/assets/stylesheets/views/landing_page/main-navigation.scss#L26), i had just separated out for the component version as if we use this component on another page we might not want that margin adjustment.

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
```diff
@@ -0,0 +1,65 @@
+name: Sub navigation menu (experimental)
+description: The sub navigation menu component helps users to navigate pages within a section of GOV.UK
+
+accessibility_criteria: |
+  The menu component must:
```
andysellick, line 5: Do we need any criteria around the expand/collapse control?

app/assets/stylesheets/govuk_publishing_components/components/_sub-navigation-menu.scss
```diff
@@ -0,0 +1,137 @@
...
+  cursor: pointer;
+  position: relative;
+  @include govuk-font(19);
+}
+
+.gem-c-sub-navigation-menu__button--negative-margin-left {
+  margin-left: govuk-spacing(-3); // Ensures the button looks perfectly left-aligned with content.
+}
+
+// Shared styles for the expanded and collapsed arrow icon. Based off of the super navigation menu button.
```
andysellick, line 34: Nitpick: what does 'shared styles' mean here? It implies they're used by something else, but I think it just means they're partly duplicated?

app/views/govuk_publishing_components/components/_menu.html.erb
andysellick: I guess that's okay then?

andysellick (commented):
Looks good, just a few more nitpicks.

app/views/govuk_publishing_components/components/_sub_navigation_menu.html.erb
AshGDS: Good idea, will do this 👍

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
AshGDS: Thanks, I think I can remove it, looks like it's meant to be used with form inputs.

app/assets/stylesheets/govuk_publishing_components/components/_sub-navigation-menu.scss
AshGDS: There's separate classes that handle the chevron looking collapsed or expanded - this class refers to the shared styles between the two states. E.g. `.gem-c-sub-navigation-menu__button[aria-expanded="false"]::before` only needs to have one line in it because of this shared class.

app/views/govuk_publishing_components/components/docs/sub_navigation_menu.yml
AshGDS: Good idea, will add

AshGDS:
Thanks @andysellick - should be ready for another review (the two latest commits) 👍

andysellick (approved):
👏

andysellick:
@AshGDS changelog entry?

PR #5455, opened 2026-05-12
Add body class option to layout for public

andysellick:
## What
- allows a string of classes to be passed to append to the body element in the component

## Why
Needed for this change: https://github.com/alphagov/frontend/pull/5533

The CSS only works if the container is the body element, as this takes into account the scrollbar for width.

## Visual Changes
None.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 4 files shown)


AshGDS (approved):

MartinJJones (approved):
The change looks good to me, and it appears to align with the approach used in the design system page template in Nunjucks - https://design-system.service.gov.uk/styles/page-template/#changing-template-content

PR #5466, opened 2026-05-19
Reintroduce z-index on focus on sub navigation menu

AshGDS:
## What/Why

- Reintroduces a `z-index` to the sub navigation menu button on focus
- This was in the original component styles but was removed when moved to this gem (we thought it wasn't needed)
- Turns out it's needed so that the focus outline appears above surrounding elements on `/missions` pages
- Jira card: https://gov-uk.atlassian.net/browse/PNP-9971

## Visual Changes

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
The changes looks good to me and is the [same approach used previously in frontend](https://github.com/alphagov/frontend/blob/aae328692a6ebde2379b83885bafa869553a2c05/app/assets/stylesheets/views/landing_page/main-navigation.scss#L51)

I think this change will need a changelog entry as well 🤔

PR #5474, opened 2026-05-26
Sanitise HTML strings in toggle.js

davidsauntson:
## What
This change ensures that the value of the `data-toggled-text` attribute conforms to a list of allowed tags and attributes, by adding some sanitisation logic to the `toggle.js` module used by the `metadata_block` and `published_dates` components.

This is limited to only allow spans (and text type nodes) at the moment, based on a check of [where `data-toggled-text` is used in `alphagov`](https://github.com/search?q=org%3Aalphagov%20data-toggled-text&type=code), this should be ok.

I've also added some config to codeql that should mean it will consider any return value from `sanitiseHTML` as safe from xss attacks.  I'm hoping this will dismiss security alert #48 when it is merged, but we might have to do that manually still.

## Why
CodeQL flagged the interpretation of DOM text (in this case in `data-toggled-text` attribute) as a possible XSS injection vector.

Fixes security alert #48, but this could be refactored out into its own module or helper if needed.  I'll look at some of the other HTML injection warnings next and see if this can be refactored to fix those, if people are happy with this approach and this gets merged.

## Visual Changes

None expected.

### DIFF
--- a/.github/codeql/extensions/qlpack.yml
+++ b/.github/codeql/extensions/qlpack.yml
@@ -1,0 +1,8 @@
+# See https://docs.github.com/en/code-security/tutorials/customize-code-scanning/creating-and-working-with-codeql-packs#creating-a-codeql-model-pack
+name: govuk_publishing_components/javascript-sanitizers
+version: 1.0.0
+library: true
+extensionTargets:
+  codeql/javascript-all: '*'
+dataExtensions:
+  - sanitizers.yml
--- a/.github/codeql/extensions/sanitizers.yml
+++ b/.github/codeql/extensions/sanitizers.yml
@@ -1,0 +1,8 @@
+# See https://codeql.github.com/docs/codeql-language-guides/customizing-library-models-for-javascript/#example-adding-flow-through-decodeuricomponent
+extensions:
+  - addsTo:
+      pack: codeql/javascript-all
+      extensible: barrierModel
+    data:
+      # Format: [ type, path, kind ]
+      - [ "global", "Member[sanitiseHTML].ReturnValue", "html-injection" ]
--- a/app/assets/javascripts/govuk_publishing_components/lib/toggle.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/toggle.js
@@ -124,13 +124,59 @@
         }
       }
 
-      var toggledText = this.getAttribute('data-toggled-text')
+      var rawToggledText = this.getAttribute('data-toggled-text')
+      var sanitisedText = sanitiseHTML(rawToggledText)
 
-      if (typeof toggledText === 'string') {
+      if (typeof rawToggledText === 'string') {
         this.setAttribute('data-toggled-text', this.innerHTML)
-        this.innerHTML = toggledText
+        this.innerHTML = sanitisedText
       }
     })
+  }
+
+  /*
+   * Sanitises a string representation of HTML so it can be safely rendered on the page
+   * @param {string} htmlString - HTML string, eg <span>hello</span>
+   * @returns {string} - the sanitised version of the string
+   */
+  const sanitiseHTML = function (htmlString) {
+    /* global DOMParser, Node */
+
+    const ALLOWED_TAGS = ['SPAN']
+    const ALLOWED_ATTRS = [/class/, /aria/]
+
+    const parser = new DOMParser()
+
+    // Suppressing alert here because this is the HTML that will be sanitised
+    // codeql[js/xss-through-dom]
+    var htmlDoc = parser.parseFromString(htmlString, 'text/html')
+
+    sanitiseNode(htmlDoc)
+    return htmlDoc.body.innerHTML
+
+    function sanitiseNode (node) {
+      const childNodes = Array.from(node.childNodes)
+      childNodes.forEach(childNode => {
+        // only sanitize element type nodes
+        if (childNode.nodeType === Node.ELEMENT_NODE) {
+          // forbidden nodes are replaced with their children (which might be nothing, ie they are removed)
+          if (!ALLOWED_TAGS.includes(childNode.tagName)) {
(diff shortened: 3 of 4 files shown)


AshGDS (commented):
The code looks good - I'm cautious about approving this though because even though GitHub says it's a security issue, it doesn't feel like one considering we're in control of everything that gets passed through to this JS (via translation strings)

There doesn't seem to be an avenue where a malicious user or editor can write HTML that ends up on this component, and we also have a content security policy which would limit what domains could have scripts executed https://github.com/alphagov/govuk_app_config/blob/00c683ef0eca0da03904e8899488fcce4f8aee5e/lib/govuk_app_config/govuk_content_security_policy.rb#L55

Also, could we sanitise the strings at the Rails level with a built in Rails function (e.g.  `sanitize`) instead of needing custom JS code?

Therefore adding this for the sake of dismissing a GitHub alert feels a bit wrong to me? Others may disagree though, so might be worth someone else looking at it.

davidsauntson:
Thanks @AshGDS!  I've responded to some of the points below.  LMK if any of it changes how you think, happy to chat about it more here or somewhere else (or not lol) 🙂 

I guess if we don't do any sanitisation then we will just need to manually dismiss all the CodeQL alerts that relate to this type of vuln, which feels a bit wrong too 😬 Unless there is another way? 🤔 

> The code looks good - I'm cautious about approving this though because even though GitHub says it's a security issue, it doesn't feel like one considering we're in control of everything that gets passed through to this JS (via translation strings)
> 
> There doesn't seem to be an avenue where a malicious user or editor can write HTML that ends up on this component, and we also have a content security policy which would limit what domains could have scripts executed https://github.com/alphagov/govuk_app_config/blob/00c683ef0eca0da03904e8899488fcce4f8aee5e/lib/govuk_app_config/govuk_content_security_policy.rb#L55

I think that is true to an extent, but I did some research and I think that there are some types of attacks where a malicious user could still exploit how the `toggle.js` code works at the moment.  There are some DOM based XSS attacks that don't rely on injecting bad data and getting it rendered by the server (ie an upstream injection), and that can be executed just on the client side.  [OWASP talk about it a bit here](https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html)

Re the CSP - you are totally right and I agree that the risk is pretty low.  But I think defence in depth is best practice, so I still think it would be good to have multiple ways of protecting against XSS. 

I didn't realise there was a central CSP before today - thanks for showing me that.  I checked whitehall in production and I actually can't see a CSP header in the response 🫨 😬, so I'll check about that with the team.  All that to say I think this gem should protect against these vulns as much as it can, as there isn't a way of enforcing what other measures the consuming apps might take.

> Also, could we sanitise the strings at the Rails level with a built in Rails function (e.g. `sanitize`) instead of needing custom JS code?

Yeah we should probably do that as well, as that would prevent upstream HTML injection (I'm guessing some apps will render user based input at some point).  I can see that in the examples we expect users to sanitise the HTML strings they add into the arrays, but maybe we should do that in the components too?

MartinJJones:
Nice work on this.

After looking at the `toggle.js` functionality further, I'm wondering if we can "simplify" the approach here and instead of using `this.innerHTML` use `this.textContent` instead, further info on this approach - https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-6-populate-the-dom-using-safe-javascript-functions-or-properties

`GemToggle` appears to be used in a few places from [searching on GitHub](https://github.com/search?q=org%3Aalphagov+data-toggled-text++NOT+is%3Aarchived&type=code), for example:

**collections frontend app**

Page example
https://www.gov.uk/entering-staying-uk/refugees-asylum-human-rights

Template
https://github.com/alphagov/collections/blob/d706307be5c19018adc0c656e0242652f15d3859/app/views/taxons/_organisations.html.erb#L41

The organisations section includes a 'Show more organisations' toggle link, using `textContent` here should be fine as no HTML is required

**published dates component**

Component docs
https://components.publishing.service.gov.uk/component-guide/published_dates

Page example
https://www.gov.uk/government/collections/drug-driving#full-publication-update-history

Component template
https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_published_dates.html.erb#L30-L33

This does include HTML, so using `textContent` here would require further work. It appears the HTML is only used to add a `+` or `-` prefix to the toggle link and nothing else, for example - `<span aria-hidden="true">+ </span>`.

To overcome this we could update the component to include a new option, for example, `data-include-prefix` (better name likely needed), then in the JS we can use `innerHTML` as before but this is not coming from the DOM to avoid the issue reported.

This approach would essentially be a refactor to use `textContent` and not allow custom HTML to be passed in, the only HTML set is what we set in the JavaScript to add the prefix

AshGDS:
@davidsauntson Interesting, thanks David. Happy to revisit things after you've considered Martin's comment 👍

MartinJJones:
Further thought on this, I think in the near future we should just be able to go with the `textContent` approach, it looks like we are removing the `+` and `-` prefixes as part of this Pull Request - https://github.com/alphagov/govuk_publishing_components/pull/5475

If we decide to go with the `textContent` approach, it may be best to wait until this PR is merged, to avoid refactoring the published dates component as I originally suggested 😁

davidsauntson:
Ah ok - thanks both!  Yes if we are getting rid of the +/- spans in the published_dates component then I think all of this is redundant and can go, and we can use `textContext` as you suggest.  I'll close this in the meantime and revisit once #5475 is merged 👍🏻

MartinJJones:
@davidsauntson no rush, just wanted to let you know that the +/- spans in the published_dates component have now been removed as part of this PR https://github.com/alphagov/govuk_publishing_components/pull/5475 🥳

PR #5481, opened 2026-06-01
Move any phase banner components to the header element

luke-booker:
## What
Adds support for rendering the phase banner component inside the semantic header element, aligning with updated GOV.UK Design System patterns.

Updated layout_for_public and layout_super_navigation_header to optionally accept phase banner parameters. Introduced an inner wrapping element around the primary navigation inside the header template.

Moved the ::before pseudo-element background inside _layout-super-navigation-header.scss away from the root header and attached it to the new nav wrapper. This prevents the absolute-positioned backdrop from expanding over and covering the phase banner when it is appended inside the header container.

Introduced .govuk-phase-banner-wrapper to colour a clean full-width white bar beneath the navigation wrapper, which contains a centred govuk-width-container to maintain text alignment.

## Why
This change fulfills the accessibility requirement raised in https://github.com/alphagov/govuk-design-system/issues/5123. Placing the phase banner within the header element ensures assistive technologies read the banner within the proper page landmark rather than leaving it isolated in the outer body space.

Jira card: https://gov-uk.atlassian.net/browse/[NAV-18918]

## Visual Changes
The phase banner now sits inside the bottom of the global header frame. Verified that expanding the desktop mega-menu dropdowns or search bars does not result in the menu layer clipping or being hidden behind the banner.

Mobile Breakpoints: Confirmed that the banner stacks gracefully beneath the mobile menu navigation triggers without overlapping.

| Before | After |
|--------|--------|
| Screenshot 2026-06-02 at 10 35 25 | Screenshot 2026-06-02 at 10 35 42 | 
Screenshot 2026-06-02 at 10 37 54 | Screenshot 2026-06-02 at 10 38 48 |

[NAV-18918]: https://gov-uk.atlassian.net/browse/NAV-18918?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
@@ -553,7 +553,9 @@ $search-icon-height: 20px;
 
 // Ensure the dropdown menu background is offset from the menu height and
 // takes up the full screen width
-.govuk-frontend-supported .gem-c-layout-super-navigation-header {
+.govuk-frontend-supported .gem-c-layout-super-navigation-header__nav-wrapper {
+  position: relative;
+
   &::before {
     content: "";
     position: absolute;
@@ -563,6 +565,10 @@ $search-icon-height: 20px;
     background: $govuk-rebrand-template-background-colour;
     box-shadow: 0 1px govuk-colour("mid-grey");
   }
+}
+
+.gem-c-phase-banner-wrapper {
+  background: govuk-colour("white");
 }
 
 // JS available - styles the links in the dropdown menu
--- a/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
+++ b/app/views/govuk_publishing_components/components/_layout_for_public.html.erb
@@ -1,6 +1,7 @@
 <%
   emergency_banner ||= nil
   global_banner ||= nil
+  phase_banner ||= nil
   html_lang ||= "en"
   homepage ||= false
   layout_helper = GovukPublishingComponents::Presenters::PublicLayoutHelper.new(local_assigns)
@@ -90,7 +91,8 @@
         <% I18n.with_locale(:en) do %>
           <%= render "govuk_publishing_components/components/layout_super_navigation_header", {
             homepage:,
-            logo_link: logo_link,
+            logo_link:,
+            phase_banner:,
           } %>
         <% end %>
       <% end %>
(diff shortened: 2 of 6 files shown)


app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -91,6 +92,7 @@
           <%= render "govuk_publishing_components/components/layout_super_navigation_header", {
             homepage:,
             logo_link: logo_link,
+            phase_banner: phase_banner,
```
jon-kirwan: Pretty sure in Rails, you don't need to pass in the value name if it's the same as the key so could just change this to `phase_banner:,` (logo_link could also be updated at the same time under a separate commit, but it's not essential).

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
```diff
@@ -565,6 +567,10 @@ $search-icon-height: 20px;
   }
 }
 
+.govuk-phase-banner-wrapper {
+  background: govuk-colour("white");
+}
+
```
jon-kirwan, line 573: Just a minor change request: `govuk-` is the naming convention used for GOV.UK Frontend styles, whereas GOV.UK Publishing Components use the `gem-c-` prefix.

https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component_conventions.md#namespacing.

So, this could be renamed from `govuk-phase-banner-wrapper` to `gem-c-phase-banner-wrapper`.

jon-kirwan (commented):
Thanks @luke-booker, looking really good. I've just added a couple of minor comments.

We should also update the component docs to include the phase banner as an example https://components.publishing.service.gov.uk/component-guide/layout_for_public/with_global_banner. We should also add a test for it too here https://github.com/alphagov/govuk_publishing_components/blob/main/spec/components/layout_for_public_spec.rb.

app/assets/stylesheets/govuk_publishing_components/components/_layout-super-navigation-header.scss
luke-booker: Thanks @jon-kirwan. Suggestions should now be reflected.

jon-kirwan (approved):
All looks good to me! Thanks @luke-booker.

PR #5484, opened 2026-06-02
Reintroduce experimental chart component

AshGDS:
## What/Why
- Reintroduces the chart component removed in #4518
- Before doing this, research was done into ONS' chart offering, and they are using a paid for service (Highcharts), which was out of scope for this experiment. Therefore I think ChartJS is appropriate for our current use case. It's worth noting that ONS mention in their accessibility statement that they're currently reviewing how they create accessible visualisations.
- Initially I was going to start from scratch and have one basic component with one chart type, but as it went on it felt more appropriate to reintroduce the previous work done, as it has been accessibility tested and already included useful things like fallback data tables and fixes for various quirks. 
- This also adds a brand colours, subtitle, and source option, as recommended by brand guidelines https://brand.design-system.service.gov.uk/data/charts/
- It's worth noting that we've identified that the recommended brand colours do not meet WCAG's colour contrast for adjacent colours, so I've kept it as a separate option rather than making it the component default. The appropriate team are looking into this but have told me it might be a while before this is resolved.

Jira card: https://gov-uk.atlassian.net/browse/PNP-10028

## Visual Changes

The chart component is back, with three new examples.


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+            - 740
+            - 820
+            - 270
+            - 450
+            - 110
+            - 210
+            - 670
+            - 430
+  multiple rows of data:
+    description: Where more than one series is shown on a chart, do not hide the legend.
```
andysellick: Later feature suggestion - it should be possible to auto detect if there is more than one series, and override any hide legend option at that point?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+            - 740
+            - 820
+            - 270
+            - 450
+            - 110
+            - 210
+            - 670
+            - 430
+  with_line_colours_and_styles:
+    description: Note that line styles and colours currently only work on line charts.
```
andysellick: It might be good to expand this description a little to explain what is allowed. It looks like any colour is allowed (as a valid CSS colour name or hex code) and the option for line styles are solid or dotted?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+    data:
+      chart_heading: Page views chart
```
andysellick: Wondering why the lines on this chart don't start and end at the extreme edges of the axes, like the others? Might be worth investigating later.

Image

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+            - 21
+            - 42
+            - 18
+  bar_chart:
+    data:
+      chart_heading: Hats owned
+      h_axis_title: Day
+      v_axis_title: Views
+      chart_overview: This chart shows page views for January 2015.
+      chart_type: "bar"
```
andysellick: I've got a new approach to writing YML examples, which is to put the thing that's being demonstrated in the current example at the top where possible, to make it obvious that this is the option being shown - just a suggestion.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+        - James
+        - Phil
+      rows:
+        - label: Hats owned
+          values:
+            - 3
+            - 1
+            - 1
+            - 2
+  format_axis_labels:
```
andysellick, line 209: No change needed here, but this example definitely falls into the category of something we probably don't need right now but looks really, really useful.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+      rows:
+        - label: Cost
+          values:
+            - 14
+            - 29
+            - 45
+            - 63
+            - 54
+            - 22
+  vertical table:
```
andysellick: I seem to remember some discussion around this option first time around, where the suggestion was that a vertical table should be the default, as it could be the best way of showing the data in most situations (we're more likely to have a higher number of data points than there are different lines on a chart). I'd definitely keep the option, as horizontal may be appropriate in some situations, but looking at the examples here it seems like vertical would be a better default.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  with_margin_bottom:
```
andysellick: Do we need this example given that this is a component wrapper helper option?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,575 @@
...
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  minimal_version:
```
andysellick: I think we should consider removing the minimal option. It was added for a very specific requirement (to make tiny graphs that showed a trend, rather than actual data) and I think is better done using a different approach (like describing it text, or having a big number components or similar).

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,658 @@
...
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  with_a_different_point_size:
```
andysellick: Two thoughts on this - do we need this option, or should we just set a sensible default (maybe one that isn't quite as large as currently)?

If we decide to keep it, we should document and limit the numbers - obviously zero is the minimum, but is there a maximum?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,658 @@
...
+        - label: January 2016
+          values:
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  with_brand_colours:
```
andysellick: Given that the brand colours currently don't meet contrast requirements I'm hesitant to have this as an explicit option. I haven't looked at the colours in detail, but would be possible to choose a selection that do meet contrast requirements, use them as the default, and potentially document/limit what colours are allowed to be passed? Currently it looks like anything can be passed, which could lead to contrast issues anyway.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,658 @@
...
+            - 1
+            - 1
+            - 2
+        - label: Shoes owned
+          values:
+            - 2
+            - 3
+            - 4
+            - 5
+  with_subtitle:
```
andysellick: I'd say this is more of a description than a subtitle? Also I'd be tempted to include it in the default example, given that it is recommended.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,658 @@
...
+            - 2
+            - 1
+            - 10
+            - 4
+        - label: Helmets owned
+          values:
+            - 1
+            - 1
+            - 1
+            - 3
```
andysellick: Clearly Iron Man owns more than 3 helmets...

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,658 @@
...
+  with_source:
```
andysellick: Again if the source is recommended we could fold it into the default example?

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,140 @@
...
+  @external_script ||= OpenStruct.new(loaded: 0)
+  @external_script[:loaded] += 1
+
+  chart_helper = GovukPublishingComponents::Presenters::ChartHelper.new(local_assigns)
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class("gem-c-chart")
+  component_helper.add_class("gem-c-chart--minimal") if minimal
+
+  require "chartkick"
+  Chartkick.options[:html] = '<div id="%{id}"><noscript><p class="govuk-body">Our charts are built using JavaScript but all the data is also available in the table below.</p></noscript></div>'
```
andysellick: Might be worth asking Kati for some better content here. For reference the similar message for maps is `This map uses JavaScript to work. To view the map, turn on JavaScript in your browser's settings.`

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,140 @@
...
+
+  if !minimal && !chart_heading
+    raise ArgumentError, "A chart heading must be provided for accessibility purposes."
+  end
+%>
+<% if rows.length > 0 && keys.length > 0 %>
+  <%= javascript_include_tag "https://www.gstatic.com/charts/loader.js" if @external_script[:loaded] == 1 %>
```
andysellick: I think this could be clearer. Again taking inspiration from the map component, you could do something like:

```Ruby
@include_script ||= OpenStruct.new(script_included: false)

...

<% unless @include_script[:script_included] %>
   # script tag here
  <% @include_script[:script_included] = true %>
<% end %>
```

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,140 @@
...
+  <%= tag.div(**component_helper.all_attributes) do %>
+    <% if chart_heading && !hide_heading %>
+      <div class="govuk-grid-row">
+        <div class="govuk-grid-column-two-thirds">
```
andysellick: I've realised a possible problem with this approach. We wrap the heading in a two thirds column to prevent the line length being too long, but this assumes the chart is being shown in a full width column. If it is put in a two thirds column, this heading wrapper shrinks to two thirds of the two thirds column, which looks a bit narrow.

I can't immediately think of an elegant solution to this (we should be using the chart in a full width column as that's the best use of space, but it might not always be the case) so maybe worth recording in an issue?

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,140 @@
...
+      <div class="gem-c-chart__table" id="<%= table_id %>">
+        <%= render("govuk_publishing_components/components/details",
+          title: t("components.chart.table_dropdown"),
+        ) do %>
+          <div tabindex="0" class="gem-c-chart__table-wrapper">
+            <table class="govuk-table">
+              <caption class="govuk-visually-hidden" id="<%= "data-table-caption-#{SecureRandom.hex(4)}" %>">
+                <%= t("components.chart.accessibility_heading", chart_heading: chart_heading) %>
+              </caption>
+              <% if table_direction == "horizontal" %>
```
andysellick, line 94: This table output is quite verbose, given the two different layout options. I wonder if it would be possible/cleaner to transform the data in the helper first depending on the option, for easier output here?

andysellick (commented):
I've not finished my review yet but github seems to have gone a bit weird so wanted to post my comments in case they were lost. Apologies - some of the comments seem to not be attached to the lines I commented on, happy to discuss if you need clarification.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: I think ChartJS evenly spaces out depending on how many labels there are, but maybe there's an option to change that

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Good idea, i'll do that - thanks

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Unfortunately any combination of the "categorical palette" doesn't meet a 3:1 Adjacent Contrast ratio required by WCAG: https://brand.design-system.service.gov.uk/colour/charts/#categorical-palette. Should I remove this option?

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Thanks, I've added it to the default example. It's called a subtitle in the brand guidelines: https://brand.design-system.service.gov.uk/data/charts/ - should we still rename it to description?

app/views/govuk_publishing_components/components/_chart.html.erb
AshGDS: I looked into this but after starting to write code I think I'm missing what you mean here - I can't see much data being transformed here, to me it looks like it's being rendered as is?

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Maybe I should use our current colours instead? https://www.gov.uk/government/publications/examples-of-visual-content-to-use-on-govuk/examples-of-visual-content-to-use-on-govuk#bar-charts

andysellick: That's an excellent idea - and we can update once the new colours are sorted out.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Up to you, but description feels more accurate to me.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Sorry, clarifying. So the if statement basically repeats the table output (with variation) depending on whether it's vertical or horizontal. What I was wondering was if it's possible to configure the data in a helper depending on that option, then just have it output once, reducing the duplication of markup in this template. But I haven't checked if that's possible or not.

AshGDS: Cool, I've tried a couple ways but run into things that might make the code more complicated to understand than if it was just verbose (mainly handling the differences in classes and accessing the cell data). Feel free to take a look at it as I may not be approaching it in the way you're thinking of!

andysellick: Yeah, this might be a bit tricky to implement, sorry. One thing I think we could do though is trim some of the repeated lines in the template. For example, the following lines are duplicated within both of the `if` conditions:

```
              <thead class="govuk-table__head">
                <tr class="govuk-table__row">
                  <td class="govuk-table__cell"></td>
```

AshGDS: Thanks @andysellick - this should be done now, it's the latest commit.

andysellick: That's good, but I think it could be a little cleaner still, how about this?

```Ruby
<table class="govuk-table">
  <caption class="govuk-visually-hidden" id="<%= "data-table-caption-#{SecureRandom.hex(4)}" %>">
    <%= t("components.chart.accessibility_heading", chart_heading: chart_heading) %>
  </caption>
  <thead class="govuk-table__head">
    <tr class="govuk-table__row">
      <td class="govuk-table__cell"></td>

      <% if table_direction == "horizontal" %>
        <% keys.each do |key| %>
          <th class="govuk-table__header" scope="col">
            <%= key %>
          </th>
        <% end %>
      <% else %>
        <% rows.each do |row| %>
          <th class="govuk-table__header govuk-table__header--stacked" scope="row">
            <%= row[:label] %>
          </th>
        <% end %>
      <% end %>
    </tr>
  </thead>
  <tbody class="govuk-table__body">
    <% if table_direction == "horizontal" %>
      <% rows.each do |row| %>
        <tr class="govuk-table__row">
          <th class="govuk-table__header" scope="row"><%= row[:label] %></th>
          <% row[:values].each do |value| %>
            <td class="govuk-table__cell govuk-table__cell--numeric">
              <%= number_with_delimiter value %>
            </td>
          <% end %>
        </tr>
      <% end %>
    <% else %>
      <% keys.each_with_index do |key, index| %>
        <tr>
          <th class="govuk-table__header" scope="row">
            <%= key %>
          </th>
          <% rows.each do |row| %>
            <td class="govuk-table__cell govuk-table__cell--numeric">
              <%= number_with_delimiter row[:values][index] %>
            </td>
          <% end %>
        </tr>
      <% end %>
    <% end %>
  </tbody>
</table>
```

AshGDS: Good idea - done, thanks

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+            - 1190
+            - 740
+            - 820
+            - 270
+            - 450
+            - 110
+            - 210
+            - 670
+            - 430
+  multiple rows of data:
```
andysellick: This example name is missing `_` characters, which results in a URL of `component-guide/chart/multiple%20rows%20of%20data` (there's a few others like this as well).

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Did you decide to not do this? I can't see a change.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Did you get anywhere with improving this text?

app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
```diff
@@ -0,0 +1,55 @@
+.gem-c-chart {
+  position: relative;
```
andysellick: Is this necessary? I can't think what it would be for, especially since the JS appears to add a child element with position relative. I did a quick visual check without it and couldn't see anything broken, but maybe you know why it's here?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
+name: Chart (experimental)
+description: The chart component presents a data chart and a tabular version of the same data
+part_of_admin_layout: true
```
andysellick: What does this do?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+  This component takes a set of data and presents it as a line graph and a table with one or more rows and lines.
+  The chart relies upon chartkick and renders using JavaScript, so the table is provided as a fallback for a lack of JavaScript,
+  an accessible view on the data for screenreaders, and a simple view of the raw data for all users.
+
+  The `chart_overview` option can be used to provide an explanation for screen reader users of what the chart shows.
+accessibility_criteria: |
+  Charts must:
+
+    - use line colours with a contrast ratio higher than 4.5:1 against the background colour to meet WCAG AA
+    - contain a `chart_heading` option if `minimal` is not set to true, so that skip links and visually hidden headings are accurate
```
andysellick: Reference to `minimal` that should be reworded.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
+name: Chart (experimental)
+description: The chart component presents a data chart and a tabular version of the same data
+part_of_admin_layout: true
+body: |
+  This component takes a set of data and presents it as a line graph and a table with one or more rows and lines.
+  The chart relies upon chartkick and renders using JavaScript, so the table is provided as a fallback for a lack of JavaScript,
+  an accessible view on the data for screenreaders, and a simple view of the raw data for all users.
+
+  The `chart_overview` option can be used to provide an explanation for screen reader users of what the chart shows.
+accessibility_criteria: |
```
andysellick, line 7: I wonder if there's any more accessibility criteria we should be including here? There's probably something about interactions on the graph or text size of graph elements. This could be quite big though so maybe worth raising an issue instead of holding up the PR.

andysellick, line 7: I'm trying to remember if we had this component accessibility tested the first time around - we probably did, but it wouldn't hurt to get it done again, particularly given that an audit is happening. Again, this shouldn't block this PR, we can look at any improvements as a follow up.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+            - 3
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  with_hidden_heading:
+    description: |
+      The heading element is optional, allowing a heading to be set outside the component. Hiding the heading doesn't remove the need to populate `chart_heading` with text, as `chart_heading` is needed for accessibility.
```
andysellick: I think the start of this wording is a little confusing, as from looking at the code (and reading the whole sentence) heading is not optional. 

I was going to suggest some alternative wording but I think this was another option on the component that we needed for a very specific use case (linked to the now removed minimal option) so maybe it would be better to remove this option entirely?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+            - 8
+            - 37
+            - 82
+            - 118
+            - 85
+            - 80
+  with_description:
+    description: The chart overview can be rendered underneath the heading as a description. Showing a description is recommended in the brand guidelines.
+    data:
+      description: true
```
andysellick: Is there a particular reason for having a true/false flag for the description, in addition to the actual text content? Could `description: true` be removed and the logic be based on the existence of a string passed for `chart_overview`?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+      brand_colours: true
+      chart_heading: Comic book character stats
```
andysellick: Nitpick: some of the options are a little verbose and very specific to this component, when we have an open issue about trying to standardise our API more. Can I suggest `chart_heading` be something simpler, like `heading`, `chart_overview` could be description, etc.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -0,0 +1,521 @@
...
+            - 1
+            - 1
+            - 1
+            - 3
+  with_source:
+    description: The chart can be rendered with the data source. Showing a source is recommended in the brand guidelines.
+    data:
+      source: Marvel and D.C. comics
+      description: true
+      brand_colours: true
```
andysellick: Is this option still used? It also doesn't seem relevant to this example?

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
+<%
+  chart_heading_level ||= 2
+  table_direction ||= "vertical"
+  rows ||= []
+  keys ||= []
+  chart_overview ||= nil
+  chart_type ||= "line"
+  hide_heading ||= false
+  link ||= false
```
andysellick: Ironically after suggesting simplifying some of these option names I'm wondering if `link` should be `download_link` for clarity. What do you think?

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
...
+  chart_id = "chart-id-#{SecureRandom.hex(4)}"
+  table_id = "table-id-#{SecureRandom.hex(4)}"
+
+  @include_script ||= OpenStruct.new(script_included: false)
+
+  chart_helper = GovukPublishingComponents::Presenters::ChartHelper.new(local_assigns)
+
+  description ||= false
+
+  source ||= false
```
andysellick: Nitpick: I would move all of the lines that accept/declare an option passed to the component to the top of this file (if possible) and group similar other lines together, like the declaration of helpers.

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
...
+    raise ArgumentError, "A chart heading must be provided for accessibility purposes."
+  end
+%>
+<% if rows.length > 0 && keys.length > 0 %>
+  <% unless @include_script[:script_included] %>
+    <%= javascript_include_tag "https://www.gstatic.com/charts/loader.js" %>
+    <% @include_script[:script_included] = true %>
+  <% end %>
+  <%= tag.div(**component_helper.all_attributes) do %>
+    <% if chart_heading && !hide_heading %>
```
andysellick: Nitpick: probably don't need to check for the existence of `chart_heading` here given that an exception would already have been raised on line 30.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: I've thought of a solution. Instead of wrapping this in a grid column, it could go in a gem style set to have a max-width of whatever the width is of a two thirds column.

app/views/govuk_publishing_components/components/docs/chart.yml
andysellick: Oh I see - it looks like this controls whether the chart description is visually hidden or not? Okay that makes sense. I think given that the description sounds like it's important, maybe we just make it always visible? I think this is one of those situations where everyone would benefit from it.

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
...
+        <%= line_chart(chart_helper.chart_format_data, colors: chart_helper.brand_colours, library: chart_helper.chart_options) %>
+      <% end %>
+    <% end %>
+
+    <%= content_tag :p, t("components.chart.source", source: source), class: "gem-c-chart__source govuk-body" if source %>
+    <div class="gem-c-chart__table" id="<%= table_id %>">
+      <%= render("govuk_publishing_components/components/details",
+        title: t("components.chart.table_dropdown"),
+      ) do %>
+        <div tabindex="0" class="gem-c-chart__table-wrapper">
```
andysellick: Do you know why there's a tabindex here? It makes the table focusable (which might be useful) but it has no styled focus state (see below).

Image

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -0,0 +1,139 @@
...
+          <table class="govuk-table">
+            <caption class="govuk-visually-hidden" id="<%= "data-table-caption-#{SecureRandom.hex(4)}" %>">
```
andysellick: What's this ID for? Does it need to be referenced by something?

andysellick (commented):
I'm so sorry Ash - I went to do a last minute glance over the code and thought of a whole load of new things. And this is particularly unfair since I wrote most of this in the first place. I'm more than happy to pair on this if you want - seems only fair.

AshGDS:
@andysellick No worries - I'll take a look at the new comments 👍

app/views/govuk_publishing_components/components/_chart.html.erb
AshGDS: Ah yeah, I'll try and follow up on this - thanks.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Good spot, I had merged the subtitle in but not the source.

app/views/govuk_publishing_components/components/_chart.html.erb
AshGDS: Good idea, i'll do that 👍

app/assets/stylesheets/govuk_publishing_components/components/_chart.scss
AshGDS: I looked at the original chart PR and this was part of the `minimal` code so will remove it - thanks

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Looks like it's a legacy thing that got copy/pasted over into this doc file. We can probably remove it across our component examples:

https://github.com/alphagov/govuk_publishing_components/blob/72458b5b57a53092a74d55ec1236f6c306c85cea/app/views/layouts/govuk_publishing_components/application.html.erb#L20

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Yeah it's best to remove it, especially considering best practice for charts it to have all the key information contained near the chart, so the heading should ideally always be there.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Good idea, will make the description a required field instead of having `chart_overview`.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Will do 👍

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Thanks, will remove

app/views/govuk_publishing_components/components/_chart.html.erb
AshGDS: I think it might have just been for determinism in testing: https://github.com/alphagov/govuk_publishing_components/commit/2bc33535787af1b3c95cf5a7446dc98420dcc541 - should I remove it?

app/views/govuk_publishing_components/components/_chart.html.erb
AshGDS: Looks like it was included in the initial chart commit: https://github.com/alphagov/govuk_publishing_components/commit/3ffdb83f8a6503540468fa5a6005fa1f30973229 - should we add some styles?

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Thanks, have fixed this across the examples.

AshGDS:
Thanks @andysellick - I think I've done everything requested now, apart from raising the accessibility criteria issue, and asking for this component to be accessibility tested (I'll probably do that once this is merged)

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: Does the table need to receive focus? The Design System table component doesn't have this. I think maybe just remove the tabindex.

app/views/govuk_publishing_components/components/_chart.html.erb
andysellick: If it's just for testing and there's another way of doing it, let's remove it.

andysellick:
Thanks @AshGDS have gone through the comments and resolved comments where appropriate. It's looking good.

AshGDS:
Thanks @andysellick - I've done both of those changes now 👍

andysellick (approved):
It's looking good, well done. Definitely interested in the outcome of the accessibility testing, and we might need to iterate some of the options once we start using it, but this feels like a good and simpler improvement over the original component.

andysellick (commented):
Changelog entry?

PR #5488, opened 2026-06-04
Remove unused 'services_cookies' option from cookie banner

jon-kirwan:
## What

Remove unused `services_cookies` option from cookie banner

## Why

No longer used anywhere https://github.com/search?q=org%3Aalphagov%20services_cookies&type=code.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,11 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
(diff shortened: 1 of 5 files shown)


MartinJJones (approved):

Looks like this variation was originally added in https://github.com/alphagov/govuk_publishing_components/pull/1438, I had a further look to see if any specific styles could be removed as well, but couldn't find anything

PR #5492, opened 2026-06-05
Create new datetime fields component

davidtrussler:
## What

Add a new "Datetime fields" component to the gem based on the [existing component](https://whitehall-admin.integration.publishing.service.gov.uk/component-guide/datetime_fields) on Whitehall.
The newly created component differs from the Whitehall component in a few ways in order to make it more generic: 
- the date and time headings are not set to a specific level or size: this is done via a "With heading level and size" option
- it introduces a "With separate date time error items" option which is required on Publisher 

In order to do this there are also some changes introduced to the existing "Form date input" component which is used in the new component. These are 
- add a new "With legend level and size" option
- add a new "With data attributes" option

## Why

Jira card: https://gov-uk.atlassian.net/browse/MAIN-1311

This is a piece of maintenance work on the Mainstream Publisher app. We added a component on Publisher that is similar to the Whitehall component but this was mainly copied and pasted from Whitehall as the full component was not available on the gem. 

This task is to reproduce the component on the gem and then introduce that back to Whitehall in place of the local component and also onto Publisher. 

## Visual Changes

| Option | Component on Whitehall | Component on Gem |
|--------|--------|--------|
| Default | Screenshot 2026-06-05 at 12 14 39 | Screenshot 2026-06-05 at 12 14 55 |
|With heading level and size | Screenshot 2026-06-05 at 12 17 08 | Screenshot 2026-06-05 at 12 17 19 |
| With custom heading text | New option | Screenshot 2026-06-05 at 12 19 15 |
| With hint text | Screenshot 2026-06-05 at 12 20 21 | Screenshot 2026-06-05 at 12 21 13 |
| With values | Screenshot 2026-06-05 at 12 22 40 | Screenshot 2026-06-05 at 12 22 51 |
| With a prepopulated time | Screenshot 2026-06-05 at 12 23 58 | Screenshot 2026-06-05 at 12 24 09 |
| With error items | Screenshot 2026-06-05 at 12 25 37 | Screenshot 2026-06-05 at 12 25 48 |
| With separate date time error items | New option | Screenshot 2026-06-05 at 12 27 21 |
| With date only option | Screenshot 2026-06-05 at 12 28 24 | Screenshot 2026-06-05 at 12 28 33 |

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
@@ -1,9 +1,11 @@
 .gem-c-datetime-fields__date-time-wrapper {
-  display: inline-flex;
+  margin-top: govuk-spacing(-2);
 }
 
 .gem-c-datetime-fields__date-time {
+  display: inline-block;
   margin-right: govuk-spacing(2);
+  margin-top: govuk-spacing(2);
 }
 
 .govuk-select {
@@ -19,6 +21,7 @@
 }
 
 .gem-c-datetime-fields__time-separator {
+  display: inline-block;
   margin: govuk-spacing(6) + 5 govuk-spacing(2) 0 0;
   margin-bottom: 0;
 }
--- a/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
@@ -1,0 +1,24 @@
+.gem-c-datetime-fields__date-time-wrapper {
+  display: inline-flex;
+}
+
+.gem-c-datetime-fields__date-time {
+  margin-right: govuk-spacing(2);
+}
+
+.govuk-select {
+  &.gem-c-datetime-fields__date-time-input {
+    max-width: 4.5em;
+    min-width: 4.5em;
+  }
+
+  &.gem-c-datetime-fields__date-time-input--large {
+    max-width: 7.5em;
+    min-width: 7.5em;
+  }
+}
+
+.gem-c-datetime-fields__time-separator {
+  margin: govuk-spacing(6) + 5 govuk-spacing(2) 0 0;
+  margin-bottom: 0;
+}
--- a/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
@@ -6,22 +6,39 @@
   display: inline-block;
   margin-right: govuk-spacing(2);
   margin-top: govuk-spacing(2);
+
+  // For using the select component
+  .govuk-select {
+    max-width: 4.5em;
+    min-width: 4.5em;
+  }
 }
 
+// This is not needed when using the select component
 .govuk-select {
   &.gem-c-datetime-fields__date-time-input {
     max-width: 4.5em;
     min-width: 4.5em;
   }
 
+  // Is this ever used? Pretty sure it isn't
   &.gem-c-datetime-fields__date-time-input--large {
     max-width: 7.5em;
     min-width: 7.5em;
   }
 }
 
+// Probably better to do this as generated text rather than a <p> element
 .gem-c-datetime-fields__time-separator {
   display: inline-block;
   margin: govuk-spacing(6) + 5 govuk-spacing(2) 0 0;
   margin-bottom: 0;
 }
+
+// Don't think this is ever used but if it is will need to be restored
+// .gem-c-datetime-fields__date-time-input--large {
+//   .govuk-select {
+//     max-width: 7.5em;
+//     min-width: 7.5em;
+//   }
+// }
--- a/app/views/govuk_publishing_components/components/_datetime_fields.html.erb
+++ b/app/views/govuk_publishing_components/components/_datetime_fields.html.erb
@@ -1,0 +1,144 @@
+<%
+  prefix ||= nil
+  field_name ||= nil
+  id = id
+  date_id = "#{id}_date"
+  date_only ||= false
+  date_heading ||= "Date (required)"
+  time_heading ||= "Time"
+
+  error_items = error_items ||= nil
+  error_items_date = error_items_date ||= nil
+  error_items_time = error_items_time ||= nil
+  error_id_time = "error-#{SecureRandom.hex(4)}"
+
+  heading_level = heading_level ||= nil
+  heading_size = heading_size ||= nil
+
+  date_hint = date_hint ||= nil
+  time_hint = time_hint ||= nil
+  time_hint_id = "time-hint-#{SecureRandom.hex(4)}"
+
+  year ||= nil
+  month ||= nil
+  day ||= nil
+  date_input_items = []
+  date_input_items << day if day
+  date_input_items << month if month
+  date_input_items << year if year
+  date_input_items = nil unless date_input_items.any?
+
+  hour ||= {}
+  hour_value = hour[:value]
+  hour_select_id = hour[:id] || "select-hour-#{SecureRandom.hex(4)}"
+  hour_label_id = "hour-#{SecureRandom.hex(4)}"
+
+  minute ||= {}
+  minute_value = minute[:value]
+  minute_select_id = minute[:id] || "select-minute-#{SecureRandom.hex(4)}"
+  minute_label_id = "minute-#{SecureRandom.hex(4)}"
+
+  date_ga4_form_section = date_heading
+  time_ga4_form_section = time_heading
+
+  root_classes = %w[gem-c-datetime-fields govuk-form-group]
+  root_classes << "govuk-form-group--error" if error_items.present?
+
+  time_group_classes = %w[govuk-form-group]
+  time_group_classes << "govuk-form-group--error" if error_items_time.present?
+
+  time_select_classes = %w[govuk-select gem-c-datetime-fields__date-time-input]
+  time_select_classes << "govuk-select--error" if (error_items.present? || error_items_time.present?)
+
+  data_attributes ||= {}
+%>
+
+<%= tag.div id: do %>
+  <% if prefix && field_name %>
+    <%= tag.div class: root_classes, data: data_attributes do %>
+      <%= render "govuk_publishing_components/components/date_input", {
+        id: date_id,
+        hint: date_hint,
+        error_items: (error_items.presence || error_items_date),
+        items: date_input_items,
+        legend_text: date_heading,
+        legend_level: heading_level,
+        legend_size: heading_size,
+        data_attributes: {
+          ga4_form_section: date_ga4_form_section,
+        },
+    } %>
+      <% unless date_only %>
+        <%= tag.div class: time_group_classes do %>
+          <%= render "govuk_publishing_components/components/fieldset", {
+            legend_text: time_heading,
+            heading_level: heading_level,
+            heading_size: heading_size,
+            classes: "govuk-date-input",
+            data_attributes: {
+              ga4_form_section: time_ga4_form_section,
+            },
+          } do %>
+            <% if time_hint %>
+              <%= render "govuk_publishing_components/components/hint", {
+                text: time_hint,
+                id: time_hint_id,
+              } %>
+            <% end %>
+
+            <% if error_items_time %>
+              <%= render "govuk_publishing_components/components/error_message", {
+                id:  error_id_time,
+                items: error_items_time,
+              } %>
+            <% end %>
+
+            <div class="gem-c-datetime-fields__date-time-wrapper">
+              <div class="gem-c-datetime-fields__date-time">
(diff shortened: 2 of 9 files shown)


app/views/govuk_publishing_components/components/docs/datetime_fields.yml
```diff
@@ -0,0 +1,132 @@
+name: Datetime fields (experimental)
+description: Use the datetime fields component to help users enter a datetime for a specific event.
+body: |
+  This differs from the [Date fields][] component in a few notable ways:
+    1. In addition, to day, month and year, it also takes hour and minute
+    2. It generates select fields in place of inputs
+    3. As this uses the `ActionView::Helpers::DateHelper` helpers you must
+    provide a prefix and field_name to construct the fields
+     - the prefix must be the downcased and underscored model name
+     - the field_name is the attribute for the model
```
davidsauntson: This renders a bit awkwardly in the summary text but I think it would behave if you followed the format in the accessibility criteria list below

Image

davidsauntson (dismissed):
LGTM! 🎉 

And thanks for raising the example PR in Whitehall 🙏🏻 

Out of interest, how different is the `date_only` option compared with the existing date component?  Do you think we would need to keep both in the long run?

davidtrussler:
> LGTM! 🎉
> 
> And thanks for raising the example PR in Whitehall 🙏🏻
> 
> Out of interest, how different is the `date_only` option compared with the existing date component? Do you think we would need to keep both in the long run?

Thanks! Yes, good question. This component uses the `date_input` component and then adds new fields for the time inputs so it's basically the same thing with that option. In which case it would make sense for this to be the only component and use the `date_only` option if that's what is required. 

It would be quite a bit of work to replace all instances of the `date_input` component retrospectively but I wonder if it's worth adding a comment in the description for that component to recommend any future requirements for a date input use this one with that option instead. WDYT?

davidtrussler:
> > LGTM! 🎉
> > And thanks for raising the example PR in Whitehall 🙏🏻
> > Out of interest, how different is the `date_only` option compared with the existing date component? Do you think we would need to keep both in the long run?
> 
> Thanks! Yes, good question. This component uses the `date_input` component and then adds new fields for the time inputs so it's basically the same thing with that option. In which case it would make sense for this to be the only component and use the `date_only` option if that's what is required.
> 
> It would be quite a bit of work to replace all instances of the `date_input` component retrospectively but I wonder if it's worth adding a comment in the description for that component to recommend any future requirements for a date input use this one with that option instead. WDYT?

Actually thinking this through a bit more it's raised a couple of other thoughts: 
- by default the `date_input` component has no legend text where this one assumes that is present. So to use this instead of the `date_input` one would require this one to have an option to not display that. I'm not sure that makes much sense though with both date and time fields.
- I wonder if the existing default headers ("Date (required)" & "Time") are a bit too Whitehall specific and should be fully customisable (we will use totally different headers on Publisher).

Maybe we just leave this one as it is for now and have these as points for consideration in the future.

davidsauntson:
Yep I agree that we should merge this now and maybe think  about whether we should maintain both in the long run.  I was just curious - I'm sure how we handle deprecations or really even if we'd want to do that anyway, especially in light of the differences you have mentioned.  Probably one to discuss as a group!

app/views/govuk_publishing_components/components/_datetime_fields.html.erb
```diff
@@ -0,0 +1,144 @@
...
+  date_heading ||= "Date (required)"
+  time_heading ||= "Time"
+
+  error_items = error_items ||= nil
+  error_items_date = error_items_date ||= nil
+  error_items_time = error_items_time ||= nil
+  error_id_time = "error-#{SecureRandom.hex(4)}"
+
+  heading_level = heading_level ||= nil
+  heading_size = heading_size ||= nil
```
andysellick: I think for these kind of lines you can just do `heading_level ||= nil` ?

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -36,6 +36,11 @@ examples:
   with_legend:
     data:
       legend_text: "What is your date of birth?"
+  with_legend_level_and_size:
```
andysellick, line 39: It's a little confusing that this option is added to the documentation in one commit, but the code behind it is added in subsequent commits. It would be cleaner to have a single commit that includes only this change, including the documentation, code and test changes.

spec/components/datetime_fields_spec.rb
```diff
@@ -0,0 +1,188 @@
+require "rails_helper"
+
+describe "Date input", type: :view do
+  def component_name
+    "datetime_fields"
+  end
+
+  it "renders the basic component" do
```
andysellick, line 8: Does this need a test for if no data is passed to the component? What would be expected to render in that situation?

spec/components/datetime_fields_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    assert_select ".govuk-fieldset__legend", text: "Date (required)"
+    assert_select ".govuk-input[name='day']"
+    assert_select ".govuk-input[name='month']"
+    assert_select ".govuk-input[name='year']"
+    assert_select ".govuk-fieldset__legend", text: "Time"
+    assert_select ".govuk-label", text: "Hour"
+    assert_select ".govuk-label", text: "Minute"
+  end
+
+  it "allows heading level and size" do
```
andysellick, line 37: This test doesn't appear to check the heading size, only the heading level? Does it need to do both?

spec/components/datetime_fields_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      field_name: "field_name",
+      heading_level: 1,
+      heading_size: "l",
+    })
+
+    assert_select "h1.govuk-fieldset__heading", text: "Date (required)"
+    assert_select "h1.govuk-fieldset__heading", text: "Time"
+  end
+
+  it "allows custom test for the date and time headings" do
```
andysellick: Typo? `custom test` should be custom text?

spec/components/datetime_fields_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+    render_component({
+      prefix: "prefix",
+      field_name: "field_name",
+      date_heading: "The date",
+      time_heading: "The time",
+    })
+
+    assert_select "legend.govuk-fieldset__legend", text: "The date"
+    assert_select "legend.govuk-fieldset__legend", text: "The time"
```
andysellick: There are a few tests like this one here that check two things. Is it worth expanding the test to check the correct text is appearing in the correct element? In theory this test could pass but the texts could be output in the wrong place. You can do something like '2nd occurrence of' - can't remember the syntax off the top of my head but there'll be examples elsewhere in this repo.

spec/components/datetime_fields_spec.rb
```diff
@@ -0,0 +1,188 @@
...
+      field_name: "field_name",
+      date_hint: "For example, 01 August 2022",
+      time_hint: "For example, 09:30 or 19:30",
+    })
+
+    assert_select ".govuk-hint", text: "For example, 01 August 2022"
+    assert_select ".govuk-hint", text: "For example, 09:30 or 19:30"
+  end
+
+  it "accepts an id" do
```
andysellick, line 69: This option doesn't appear to be documented in the YML?

spec/components/date_input_spec.rb
```diff
@@ -46,6 +46,19 @@ def component_name
...
+    render_component(
+      legend_text: "What is your date of birth?",
+      legend_level: 1,
+      legend_size: "l",
+    )
+
+    assert_select ".govuk-fieldset[role=group] .govuk-fieldset__legend.govuk-fieldset__legend--l"
+    assert_select ".govuk-fieldset[role=group] .govuk-fieldset__legend h1", text: "What is your date of birth?"
+    assert_select ".govuk-fieldset[role=group] .govuk-date-input", 1
+    assert_select ".govuk-fieldset[role=group] .govuk-date-input__item", 3
```
andysellick: Nitpick: the last two lines in this text don't appear to be related to the name/intent of the test - looks like they're checking that a certain number of elements are appearing. I think these would be better placed in another test that intentionally counts the outputted elements, if that is required.

app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
```diff
@@ -0,0 +1,24 @@
+.gem-c-datetime-fields__date-time-wrapper {
+  display: inline-flex;
```
andysellick: This might need some checking on mobile as it doesn't provide any option for wrapping, if the user has specified a larger font size or zoom. Date fields look okay though.

Image

app/views/govuk_publishing_components/components/_datetime_fields.html.erb
```diff
@@ -0,0 +1,144 @@
...
+
+            <div class="gem-c-datetime-fields__date-time-wrapper">
+              <div class="gem-c-datetime-fields__date-time">
+                <%= render "govuk_publishing_components/components/label", {
+                  text: "Hour",
+                  html_for: hour_select_id,
+                  id: hour_label_id,
+                } %>
+
+                <%= select_hour hour_value,
```
andysellick: I haven't checked so this is a genuine question - should these be select elements? Design System guidance for date fields now is that users should type in their own values rather than choosing from long lists of options. Does that not apply for times as well?

app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
davidtrussler: Yes, the Time fields are layed out with flexbox and the Date fields are not. I've udpated the Time fields to match the Date fields in this respect. Doesn't make much sense to do them differently.

app/views/govuk_publishing_components/components/_datetime_fields.html.erb
davidtrussler: Good question - it would make sense for them both to be the same. This component is based on the one that Whitehall have been using for a while now and I understand the request for dropdown time selectors has come from content designers and may be related to the requirement for them to enter times in 24 hour clock format. Do you think we should chase this up with the Design System team?

spec/components/date_input_spec.rb
davidtrussler: Yeah, I don't see the point of those assertions. That test was just copied from the previous one ("renders with legend and fieldset") and modified. I have deleted those assertions in both tests.

app/views/govuk_publishing_components/components/_datetime_fields.html.erb
davidtrussler: FYI I've raised that [here](https://gds.slack.com/archives/CAF8JA25U/p1781089740611949), though the general subject of date and time selections is something we are aware of and have as a future improvement so it may be we just want to go with something like this for now and come back to it later.

spec/components/datetime_fields_spec.rb
davidtrussler: Yes, that's a good point. As it stands the component uses the Rails `select_hour` and `select_minute` helpers meaning that we **must** give values for `prefix` and `field_name`. If either one of these is not present the component does not render at all. This seems to be not ideal to me, as is the fact that it enforces a specific format for the `name` parameter of the input. (This will require further work on Publisher when we use it there). I'll give some thought to the best way of dealing with this which might also be affected by the question of whether we use `select` elements for the time inputs or not.

spec/components/datetime_fields_spec.rb
davidtrussler: Good point, I've added tests for size as well.

spec/components/datetime_fields_spec.rb
davidtrussler: I've tried to do this in three places in the file using `css_select` to get an array of the elements we need to test and `assert_equal` to test each of those against the expected text. I am running into trouble though with whitespace making this fail. So it may be there's a better way to do this along the lines you are suggesting - though I can't find what it would be.

app/assets/stylesheets/govuk_publishing_components/components/_datetime-fields.scss
andysellick: Much better, thanks.

spec/components/datetime_fields_spec.rb
andysellick: I think you could use the GA4 attributes to help here, something like...

```
assert_select "[data-ga4-form-section='The date'] .govuk-fieldset__legend", text: "The date"
```

spec/components/datetime_fields_spec.rb
davidtrussler: True. I have added it. I'm not sure how useful it is TBH but Whitehall have used it on their local component so I guess it needs to be there.

spec/components/datetime_fields_spec.rb
davidtrussler: Ah nice, that's better. I've updated the three instances of the tests where this is a problem along those lines.

PR #5499, opened 2026-06-10
Add auto layout option to cards component

andysellick:
## What
Adds an option to the cards component to make it adapt to its container size.

- automatically adjusts the number of columns used by the component based on the size of its parent container, as well as the screen size
- columns are based on the govuk-grid system
- so if the component is in a full width column, it will show three columns of cards
- in a two thirds column, two columns
- in a one third column, one column
- it still respects the normal screen size rules, so on mobile it collapses to a single column as usual

## Why
We want to try out this kind of behaviour in flexible pages, where components could be positioned anywhere in a page and it would be good to be able to do that without having to specifically configure each instance.

## Visual Changes
Example in the component guide shows the behaviour in a two thirds column. You can see how it behaves by modifying the width of a container in the dev tools.

Screenshot 2026-06-10 at 14 11 44

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Add auto layout option to cards component ([PR #5499](https://github.com/alphagov/govuk_publishing_components/pull/5499))
+
 ## 66.5.0
 
 * Remove unused 'services_cookies' option from cookie banner ([PR #5488](https://github.com/alphagov/govuk_publishing_components/pull/5488))
--- a/spec/components/cards_spec.rb
+++ b/spec/components/cards_spec.rb
@@ -9,6 +9,26 @@ describe "Cards", type: :view do
     assert_empty render_component({})
   end
 
+  it "renders the default component with the correct default classes" do
+    test_data = {
+      heading: "Topics",
+      items: [
+        {
+          link: {
+            text: "Benefits",
+            path: "http://www.gov.uk",
+          },
+        },
+      ],
+    }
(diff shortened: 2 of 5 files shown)


spec/components/cards_spec.rb
```diff
@@ -121,6 +121,22 @@ def component_name
...
+          link: {
+            text: "Benefits",
+            path: "http://www.gov.uk",
+          },
+        },
+      ],
+    }
+    render_component(test_data)
+    assert_select ".gem-c-cards.gem-c-cards--auto-layout", count: 1
+  end
```
AshGDS, line 158: Ideally these tests should also be accompanied by an opposing test like "It doesn't render this class when the default options are passed", otherwise we're only really testing one side of the boolean. But I don't think we're that good at doing that in general in our tests, so this is a bit of a nitpick - feel free to add that test if you think it makes sense though.

CHANGELOG.md
```diff
@@ -9,10 +9,9 @@
 
 ## Unreleased
 
+* Add auto layout option to cards component ([PR #5499](https://github.com/alphagov/govuk_publishing_components/pull/5499))
```
AshGDS, line 12: You might need to rebase this, otherwise this could end up in the previous version release notes.

AshGDS (approved):
LGTM, smart solution 👍

CHANGELOG.md
andysellick: Good spot, thanks, have rebased 👍

spec/components/cards_spec.rb
andysellick: Have added an extra test at the start of the file, see what you think.

PR #5501, opened 2026-06-12
Remove extra breadcrumbs padding on mobile

### LINKED ISSUE #5022: Align breadcrumb styles with the Design System
## What

Align the breadcrumb component styles with the Design System, the approach here will mostly likely involve reviewing existing styles and removing any overrides where possible, for example:

- Remove `--inverse` gem styles and add support for the [`--inverse` styles from the design system](https://design-system.service.gov.uk/components/breadcrumbs/#breadcrumbs-on-dark-backgrounds) in the breadcrumb view template
- Remove [`govuk-breadcrumbs--collapse-on-mobile` overrides](https://github.com/alphagov/govuk_publishing_components/blob/f00373af0e3b36df253e1addb6f626236256bfa5/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss#L36-L56)

## Why

We currently have quite a few style overrides that are likely no longer needed following updates in the design system over the last few years. By  using styles from the design system directly, the codebase will be easier to maintain and the breadcrumb styling will be consistent across services.

### Example visual change

The breadcrumbs styling in the gem increases in height across smaller screen sizes, by using the design system styles we will have consistent spacing across mobile, tablet and desktop

| Before | After |
| --- | --- |
| Image | Image |

### Further info

- https://design-system.service.gov.uk/components/breadcrumbs/
- https://components.publishing.service.gov.uk/component-guide/breadcrumbs

andysellick:
## What
Remove some custom styles from the breadcrumb component, to better align it with the Design System and reduce inconsistency.

Also update some of the examples in the component guide.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/5022

## Visual Changes
This change removes some extra padding we added around the breadcrumb when the `collapse_on_mobile` option is enabled (which seems to be the case for most instances of the component).

Before | After
----- | -------
Screenshot 2026-06-12 at 10 39 10 | Screenshot 2026-06-12 at 10 39 20

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Remove extra breadcrumbs padding on mobile ([PR #5501](https://github.com/alphagov/govuk_publishing_components/pull/5501))
 * Add auto layout option to cards component ([PR #5499](https://github.com/alphagov/govuk_publishing_components/pull/5499))
 
 ## 66.5.0
--- a/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_breadcrumbs.scss
@@ -11,28 +11,6 @@
   }
 }
 
-.govuk-breadcrumbs--collapse-on-mobile {
-  @include govuk-media-query($until: tablet) {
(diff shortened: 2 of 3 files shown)


MartinJJones (approved):
Nice work on fixing this, great to align with the Design System styles 🥳

PR #5507, opened 2026-06-16
Remove navigation functionality from the layout header component

luke-booker:
## What

1. Removed `layout-header.js` entirely, as the header component no longer relies on JavaScript to handle navigation functionality.
2. Updated `_layout_header.html.erb` and `_header_logo.html.erb` to remove deprecated navigation branches. Extracted the sign-out link away from the legacy navigation wrapper class which has been removed.
3. Cleared multiple redundant selectors used for the old mobile navigation and menu toggle buttons from `_layout-header.scss`. Adjusted the layout to flex on the logo for desktop only, allowing the sign-out link to align and scale cleanly across breakpoints.
4. govuk-header__link classes were removed [here](https://github.com/alphagov/govuk-frontend/pull/6423/changes#diff-d8691ea3f2132ce5fe2a9453168f515568ec9ebb4851757fba62f73702427880L197) and have now been replaced with gem-c-layout-header__signout-link
5. Removed obsolete tests targeting the deprecated navigation code paths and added a check for sign-out link functionality.
6. Removed the deprecated navigation from the component guide documentation. 
7. Checked and updates the template and tests for the [Layout super navigation header](https://components.publishing.service.gov.uk/component-guide/layout_super_navigation_header) to use govuk-header__homepage-link class.

## Why

These changes align the gem with the design system and remove the deprecated layout header navigation functionality. Cleaning out this this dead code and isolating the sign-out link styles, we eliminate technical debt, ease maintainability and safeguard against layout regressions.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18927

## Visual changes

For the majority of pages, users won't see any visual changes at all because the deprecated navigation wasn't actively being used by any applications. 

## Testing

Confirmed the navigation functionality has been removed from the header component. Checked in browsers (Chrome, Edge, Safari) that the header appears and behaves as expected and tests are passing. 

## Visual Changes - Mobile (unchanged*)

Layout header (Before)

Screenshot 2026-06-26 at 14 08 58

Layout super navigation header (Before)

Screenshot 2026-06-25 at 16 26 40

Layout header (After)

Screenshot 2026-06-26 at 14 09 06

Layout super navigation header (After)

Screenshot 2026-06-25 at 16 29 38

## Visual Changes - Tablet (unchanged*)

Layout header (Before)

Screenshot 2026-06-26 at 14 08 24

Layout super navigation header (Before)

Screenshot 2026-06-25 at 16 26 46

Layout header (After)

Screenshot 2026-06-26 at 14 08 38

Layout super navigation header (After)

Screenshot 2026-06-25 at 16 29 43

## Visual Changes - Desktop (unchanged*)

Layout header (Before)

Screenshot 2026-06-26 at 14 07 34

Layout super navigation header (Before)

Screenshot 2026-06-25 at 16 26 51

Layout header (After)

Screenshot 2026-06-26 at 14 07 45

Layout super navigation header (After)

Screenshot 2026-06-25 at 16 29 48

## Anything else?

* There is a 1px height difference with the new banner. 

Before

Screenshot 2026-06-26 at 11 23 33

After

Screenshot 2026-06-26 at 11 23 44

The reason for this is because the original nav and sign out padding was dictating the header height. We've now removed this responsibility to use flex and aligned the height to be consistent with other headers (60px).

Non-sign out header at 60px height 

Screenshot 2026-06-26 at 16 10 47


app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -8,16 +8,9 @@
...
 %>
 
-<% if environment_exists && deprecated_navigation %>
-  <div class="gem-c-header__environment">
-    <%= render "govuk_publishing_components/components/tag", { text: environment, colour: colour_map[environment.downcase] } %>
-  </div>
-<% end %>
-
 <div class="govuk-header__logo gem-c-header__logo">
   <a href="<%= logo_link %>" class="govuk-header__link govuk-header__link--homepage">
```
jon-kirwan: It looks like these styles have been replaced in v6 so we can swap `govuk-header__link govuk-header__link--homepage` for `govuk-header__homepage-link` https://github.com/alphagov/govuk-frontend/pull/6423#:~:text=tests%20(for%20JavaScript).-,Refactored,-.govuk%2Dheader__link.govuk.

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
```diff
@@ -27,17 +20,15 @@
...
   <% end %>
 </div>
 
 <% if signout_link %>
-  <div class="govuk-header__content gem-c-header__content">
-    <nav class="gem-c-header__nav govuk-header__navigation">
-      <a class="govuk-header__link govuk-!-font-weight-bold" href="/auth/gds/sign_out">Sign out</a>
-    </nav>
+  <div class="gem-c-layout-header__signout">
+    <a class="govuk-header__link govuk-!-font-weight-bold gem-c-layout-header__signout-link" href="/auth/gds/sign_out">Sign out</a>
```
jon-kirwan: I'm not sure whether we need to keep the nav element, even if there's only one link. It might be worth checking with Calum to confirm.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
```diff
@@ -60,24 +60,29 @@
...
-    padding: govuk-spacing(3) 0 govuk-spacing(2);
+  &:has(+ .gem-c-layout-header__signout) {
+    @include govuk-media-query($from: desktop) {
+      width: 75%;
+      float: left;
+    }
   }
 }
 
-.gem-c-header__content.govuk-header__content {
```
jon-kirwan, line 69: It looks like there's another occurence of `govuk-header__content` which can be removed here
https://github.com/alphagov/govuk_publishing_components/blob/fd2e3151cd8b2d756ed13c02e5161cb274dd91d6/app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss#L39-L43

app/views/govuk_publishing_components/components/_layout_header.html.erb
```diff
@@ -6,50 +6,28 @@
...
-          <%= render "govuk_publishing_components/components/layout_header/header_logo", {
-            environment:,
-            logo_link:,
-            product_name:,
-            deprecated_navigation: true,
-          } %>
-        </div>
-        <% if navigation_items.any? %>
-          <div class="govuk-header__content gem-c-header__content govuk-grid-column-full govuk-!-display-none-print">
-            <%= render "govuk_publishing_components/components/layout_header/navigation_items", navigation_items: navigation_items, navigation_aria_label: navigation_aria_label %>
```
jon-kirwan, line 49: I'm pretty sure the `/navigation_items` view template can now be removed entirely. I don't think anything is using it anymore.

There may also be some related styles that can be cleaned up once it's deleted.

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
jon-kirwan: There are a few specs that will also need updating to reflect this change, here:

https://github.com/alphagov/govuk_publishing_components/blob/fd2e3151cd8b2d756ed13c02e5161cb274dd91d6/spec/components/layout_header_spec.rb#L142-L152

jon-kirwan (commented):
Thanks Luke, this is looking really good. I've added a few minor comments.

One thing I did notice is that the sign-out link variant appears to have some display issues when tested against the v6 from https://github.com/alphagov/govuk_publishing_components/tree/v6-upgrade-feature-branch.

I suspect it may be related to the floated columns used in this layout, but it might need a bit more investigation to confirm.

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
luke-booker: > I'm not sure whether we need to keep the nav element, even if there's only one link. It might be worth checking with Calum to confirm.

Thanks Jon. Good shout on this. My thinking removing the nav element is because it's a single link (sign out) rather than a major navigation block. A single link alone in a nav creates an extra landmark for screen reader users which can clutter menus and cause confusion. What do you think @calumryan ?

luke-booker: @jon-kirwan Confirmed with Calum it should be ok but will need a nav element should more links be added again in the future

luke-booker:
> Thanks Luke, this is looking really good. I've added a few minor comments.
> 
> One thing I did notice is that the sign-out link variant appears to have some display issues when tested against the v6 from https://github.com/alphagov/govuk_publishing_components/tree/v6-upgrade-feature-branch.
> 
> I suspect it may be related to the floated columns used in this layout, but it might need a bit more investigation to confirm.

Thanks for reviewing @jon-kirwan. I **think** `govuk-header__link` has been retired. I've replaced it using `govuk-link govuk-link--inverse` to achieve the same.

app/assets/stylesheets/govuk_publishing_components/components/_layout-header.scss
luke-booker: Done, removed that

app/views/govuk_publishing_components/components/_layout_header.html.erb
luke-booker: Removed /navigation_items

MartinJJones (approved):
Nice work, PR description and commits are very helpful

app/views/govuk_publishing_components/components/layout_header/_header_logo.html.erb
jon-kirwan: > @jon-kirwan Confirmed with Calum it should be ok but will need a nav element should more links be added again in the future

Great, thanks for confirming, Luke. Sounds good 👍

MartinJJones (approved):

PR #5509, opened 2026-06-18
Remove innerHTML from filter-list.js

MartinJJones:
## What
Refactor  the `appendFilterInput` method in `/lib/filter-list.js` to remove the use of innerHTML and replace with textContent

## Why
Improve code quality and security, as recommended in https://cheatsheetseries.owasp.org/cheatsheets/DOM_based_XSS_Prevention_Cheat_Sheet.html#rule-6-populate-the-dom-using-safe-javascript-functions-or-properties

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/lib/filter-list.js
+++ b/app/assets/javascripts/govuk_publishing_components/lib/filter-list.js
@@ -16,10 +16,21 @@
   FilterList.prototype.appendFilterInput = function () {
     const form = document.createElement('div')
     form.classList.add('govuk-form-group')
-    form.innerHTML = `
-      <label for="filterInput" class="gem-c-label govuk-label">${this.labelText}</label>
-      <input class="gem-c-input govuk-input" id="filterInput" name="name" spellcheck="false" type="text">
-    `
+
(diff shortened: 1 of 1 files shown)


AshGDS (approved):
LGTM 👍 

Is it possible that `this.labelText` ever contained HTML (e.g. a screen reader only hidden span) and this would break that? (IIRC this JS is used internally only, so this probably isn't relevant here.)

maxgds:
Can you edit the description - you've got a quite "what-ish" why, explain it's for improved security

MartinJJones:
> LGTM 👍
> 
> Is it possible that this.labelText ever contained HTML (e.g. a screen reader only hidden span) and this would break that? (IIRC this JS is used internally only, so this probably isn't relevant here.)

@AshGDS good spot, that certainly could happen in the future and we would likely need to decide with level of customisation we support. At the moment though it is only used internally as you say, added fairly recently in this PR - https://github.com/alphagov/govuk_publishing_components/pull/5126

MartinJJones:
> Can you edit the description - you've got a quite "what-ish" why, explain it's for improved security

@maxgds good point, I've updated the why section and also referenced the guidance followed, thanks

PR #5511, opened 2026-06-18
[IA-2883] Accept and provide link decoration to/from specific links

JonathanHallam:
## What
Accept and provide link decoration to/from specific links for cookie consent. This works using a JS script that iterates over the links on each page and adds the query string required.

## Why
This will allow users to come from a government site like example.gov.uk accept or decline cookies and not have to see our cookie banner again. This will also work in the other direction. This is part of wider work on end to end journeys. The only goal of this specific work is to create a better journey for users. It will mean that they only have to accept cookies once

## Exceptions, caveats, missing things and tips for review
There's a few things that need to be assumed when reviewing this PR. The first and biggest is that this is legal and something that we want to do. That's not currently known, but it's outside the scope of this work and something that our team is actively working on. 

Currently this also takes a really naive approach to cookie management, it can only deal with a binary yes/no choice and not the detailed cookie settings we offer, nor does it care about anything but a binary coming into GOV.UK. That will all be dealt with later when we know more about how other departments/partners will use this feature. 

There's an issue with people sharing links after they've landed somewhere and deciding someone else's cookie choices for them, again I'm hoping to deal with that later. 

If you don't accept or decline cookies it still decorates links to the negative and won't give you another chance to accept them later on on other sites. Not sure if this is mitigable but it's a risk we're happy with at the moment

As it stands this points to 3 sites that we over in insights and analytics have control over. This allows us to test how it works with dummy sites and mitigates the risk of this escaping into the wild. If we end up using this in anger further down the road we'll update these sites to be outside departments who know how to use it. 

When it comes to review, basically assume that this functions as intended and review the code itself, please and thank you!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Accept and provide link decoration to/from government links ([PR #5511](https://github.com/alphagov/govuk_publishing_components/pull/5511))
+
 ## 67.0.0
 
 * Add options to the details component ([PR #5594](https://github.com/alphagov/govuk_publishing_components/pull/5594))
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -29,7 +29,7 @@ var initFunction = function () {
         const url = new URL(link.href, window.location.origin)
 
         if (allowedDomains.includes(url.hostname)) {
-          url.searchParams.set('consent', consent)
+          url.searchParams.set('cookies[analytics]', consentValue)
           link.href = url.toString()
         }
       } catch (e) {
@@ -38,35 +38,37 @@ var initFunction = function () {
     })
   }
 
+  window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration(window.location)
+
   var consentCookie = window.GOVUK.getConsentCookie()
-  if (consentCookie && consentCookie.usage) {
-    window.GOVUK.analyticsGa4.decorateLinks(Object.values(consentCookie).includes(false).toString())
 
-    window.GOVUK.analyticsGa4.vars.internalDomains = []
-    window.GOVUK.analyticsGa4.vars.internalDomains.push(window.GOVUK.analyticsGa4.core.trackFunctions.getHostname())
-    window.GOVUK.analyticsGa4.core.trackFunctions.appendDomainsWithoutWWW(window.GOVUK.analyticsGa4.vars.internalDomains)
-    window.GOVUK.analyticsGa4.core.load()
+  if (consentCookie) {
+    var consentValue = consentCookie.usage ? 'yes' : 'no'
+    window.GOVUK.analyticsGa4.decorateLinks(consentValue)
 
-    if (!window.GOVUK.analyticsGa4.analyticsModulesStarted) {
-      // Initialise analytics modules that start on page load
-      // https://github.com/alphagov/govuk_publishing_components/blob/main/docs/analytics-ga4/analytics.md#code-structure
-      var analyticsModules = window.GOVUK.analyticsGa4.analyticsModules
-      for (var property in analyticsModules) {
-        var module = analyticsModules[property]
-        if (typeof module.init === 'function') {
-          try {
-            module.init()
-          } catch (e) {
-            // if there's a problem with the module, catch the error to allow other modules to start
-            console.warn('Error starting analytics module ' + property + ': ' + e.message, window.location)
+    if (consentCookie.usage) {
+      window.GOVUK.analyticsGa4.vars.internalDomains = []
+      window.GOVUK.analyticsGa4.vars.internalDomains.push(window.GOVUK.analyticsGa4.core.trackFunctions.getHostname())
+      window.GOVUK.analyticsGa4.core.trackFunctions.appendDomainsWithoutWWW(window.GOVUK.analyticsGa4.vars.internalDomains)
+      window.GOVUK.analyticsGa4.core.load()
+
+      if (!window.GOVUK.analyticsGa4.analyticsModulesStarted) {
+        var analyticsModules = window.GOVUK.analyticsGa4.analyticsModules
+        for (var property in analyticsModules) {
+          var module = analyticsModules[property]
+          if (typeof module.init === 'function') {
+            try {
+              module.init()
+            } catch (e) {
+              console.warn('Error starting analytics module ' + property + ': ' + e.message, window.location)
+            }
           }
         }
+        window.GOVUK.analyticsGa4.analyticsModulesStarted = true
       }
-      window.GOVUK.analyticsGa4.analyticsModulesStarted = true
     }
   } else {
     window.addEventListener('cookie-consent', window.GOVUK.analyticsGa4.init)
-    window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration(window.location)
   }
 }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -1,5 +1,23 @@
 var initFunction = function () {
   window.removeEventListener('cookie-consent', window.GOVUK.analyticsGa4.init)
+
+  window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration = function (location) {
+    if (!location || !location.search) return
+    // this checks for the presence of cookie consent query parameter and updates our consent cookie accordingly
+    // This is so that users don't see multiple cookie banners when they move between different domains
+    var cookieConsent = /([?&]cookie_consent=)(accept|reject)/.exec(location.search)
+    if (cookieConsent) {
+      if (cookieConsent[2] === 'accept') {
+        window.GOVUK.setConsentCookie({ usage: true })
+        // set cookies_preferences_set to true to prevent cookie banner showing
+        window.GOVUK.cookie('cookies_preferences_set', 'true')
+      } else if (cookieConsent[2] === 'reject') {
+        window.GOVUK.setConsentCookie({ usage: false })
+        window.GOVUK.cookie('cookies_preferences_set', 'true')
+      }
+    }
+  }
+
   var consentCookie = window.GOVUK.getConsentCookie()
   if (consentCookie && consentCookie.usage) {
     window.GOVUK.analyticsGa4.vars.internalDomains = []
@@ -26,6 +44,7 @@ var initFunction = function () {
     }
   } else {
     window.addEventListener('cookie-consent', window.GOVUK.analyticsGa4.init)
+    window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration(window.location)
   }
 }
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -18,8 +18,30 @@ var initFunction = function () {
     }
   }
 
+  window.GOVUK.analyticsGa4.decorateLinks = function (consent) {
+    // Select all anchor tags with an href attribute
+    const links = document.querySelectorAll('a[href]')
+    const allowedDomains = ['example.service.gov.uk', 'micropigs.campaign.gov.uk']
+
+    links.forEach(link => {
+      try {
+        // Use the URL constructor to safely parse absolute or relative URLs
+        const url = new URL(link.href, window.location.origin)
+
+        if (allowedDomains.includes(url.hostname)) {
+          url.searchParams.set('consent', consent)
+          link.href = url.toString()
+        }
+      } catch (e) {
+        // Silently skip invalid URLs
+      }
+    })
+  }
+
   var consentCookie = window.GOVUK.getConsentCookie()
   if (consentCookie && consentCookie.usage) {
+    window.GOVUK.analyticsGa4.decorateLinks(Object.values(consentCookie).includes(false).toString())
+
     window.GOVUK.analyticsGa4.vars.internalDomains = []
     window.GOVUK.analyticsGa4.vars.internalDomains.push(window.GOVUK.analyticsGa4.core.trackFunctions.getHostname())
     window.GOVUK.analyticsGa4.core.trackFunctions.appendDomainsWithoutWWW(window.GOVUK.analyticsGa4.vars.internalDomains)
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -72,4 +72,11 @@ var initFunction = function () {
   }
 }
 
+document.querySelectorAll('a').forEach(function(link) {
+  if (link.href) {
+    // Check if the link already has a query string to decide between '?' and '&'
+    link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
+  }
+});
+
 window.GOVUK.analyticsGa4.init = initFunction
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -75,7 +75,7 @@ var initFunction = function () {
 document.querySelectorAll('a').forEach(function(link) {
   if (link.href) {
     // Check if the link already has a query string to decide between '?' and '&'
-    link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
+    link.href += (link.href.includes('?') ? '&' : '?') + 'b=b';
   }
 });
 
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -1,4 +1,16 @@
+document.addEventListener("DOMContentLoaded", function() {
+  console.log("DOM is ready. Decorating cross-domain links...");
+  
+  document.querySelectorAll('a').forEach(function(link) {
+    // Only decorate links that have an href AND don't contain a hash/anchor
+    if (link.href && !link.href.includes('#')) {
+      link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
+    }
+  });
+});
+
 var initFunction = function () {
+  
   window.removeEventListener('cookie-consent', window.GOVUK.analyticsGa4.init)
 
   window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration = function (location) {
@@ -72,25 +84,5 @@ var initFunction = function () {
   }
 }
 
-document.querySelectorAll('a').forEach(function(link) {
-  if (link.href) {
-    // Check if the link already has a query string to decide between '?' and '&'
-    link.href += (link.href.includes('?') ? '&' : '?') + 'b=b';
-  }
-});
-
 window.GOVUK.analyticsGa4.init = initFunction
 
-// 1. A simple log to prove the file is executing
-console.log("init-ga4.js has successfully loaded!");
-
-// 2. Wait for the HTML body to finish loading before looking for links
-document.addEventListener("DOMContentLoaded", function() {
-  console.log("DOM is ready. Decorating links now...");
-  
-  document.querySelectorAll('a').forEach(function(link) {
-    if (link.href) {
-      link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
-    }
-  });
-});
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -80,3 +80,17 @@ document.querySelectorAll('a').forEach(function(link) {
 });
 
 window.GOVUK.analyticsGa4.init = initFunction
+
+// 1. A simple log to prove the file is executing
+console.log("init-ga4.js has successfully loaded!");
+
+// 2. Wait for the HTML body to finish loading before looking for links
+document.addEventListener("DOMContentLoaded", function() {
+  console.log("DOM is ready. Decorating links now...");
+  
+  document.querySelectorAll('a').forEach(function(link) {
+    if (link.href) {
+      link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
+    }
+  });
+});
--- a/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
+++ b/app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
@@ -1,12 +1,7 @@
 document.addEventListener("DOMContentLoaded", function() {
-  console.log("DOM is ready. Decorating cross-domain links...");
-  
-  document.querySelectorAll('a').forEach(function(link) {
-    // Only decorate links that have an href AND don't contain a hash/anchor
-    if (link.href && !link.href.includes('#')) {
-      link.href += (link.href.includes('?') ? '&' : '?') + 'a=a';
-    }
-  });
+  if (typeof window.GOVUK.analyticsGa4.init === 'function') {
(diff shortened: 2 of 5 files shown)


AshGDS (commented):
Would you be able to provide some more context for this change? I haven't looked at the code fully but on first glance some things I can think of are:

- If users end up clicking on a link shared with them that has `cookie_consent=accept` in it, they're going to be accepting cookie consent without knowing, which we probably don't want?
- The navigation team are currently rewriting the cookie consent code - have you been working with them on this change?
- This will modify how links are tracked for every `.gov.uk` domain as the path will change to include this search query - are the analytics team OK with this?

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
```diff
@@ -104,6 +105,10 @@
       <%= raw(global_banner) %>
     <% end %>
 
+    <% if test_links.present? %>
```
andysellick: Is this just for testing purposes, or is it used by the live code somehow?

spec/components/layout_for_public_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+require "byebug"
```
andysellick: Is this still needed?

spec/components/layout_for_public_spec.rb
```diff
@@ -187,6 +188,40 @@ def component_name
...
+    })
+
+    visit "/public"
+
+    click_button "Accept additional cookies"
+
+    visit "/public"
+
+    assert_select("#link_1", href: "www.gov.uk")
+    assert_select("#link_2", href: "www.example.gov.uk?cookie_consent=true")
```
andysellick: Is this an example of the kind of link decoration that this change will introduce? The cookie consent is a bit more complex as there are three categories of cookie types that can be accepted/declined - would this change allow for that?

CHANGELOG.md
```diff
@@ -7,6 +7,9 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+* Accept and provide link decoration to/from government links ([PR #5511](https://github.com/alphagov/govuk_publishing_components/pull/5511))
```
andysellick: Can you add some documentation explaining this functionality and how it should work?

andysellick (commented):
I've only had a very quick glance at this PR but it would benefit from more detail in the description explaining the context of this change - attempts like this have been tried before and it would be good to understand the wider context for this piece of work.

AshGDS:
Another thing I thought of - if different sites have different types of cookies you accept, how will this work from a policy perspective? Accepting cookies on `www.gov.uk` probably means something different to accepting them on a `service.gov.uk` domain and vice versa, as we use cookies for different things.

JonathanHallam:
I've moved this back into draft as it is less legislatively sound than I first thought!

JonathanHallam:
> Would you be able to provide some more context for this change? I haven't looked at the code fully but on first glance some things I can think of are:
> 
> * If users end up clicking on a link shared with them that has `cookie_consent=accept` in it, they're going to be accepting cookie consent without knowing, which we probably don't want?
> * The navigation team are currently rewriting the cookie consent code - have you been working with them on this change?
> * This will modify how links are tracked for every `.gov.uk` domain as the path will change to include this search query - are the analytics team OK with this?

Hi Ash, thanks for this, I'll go through it point by point;

* That is a bit of a risk yes, but I think minimal because the link decoration should be visible to the person it's being shared with
* I've spoken briefly to Jon and I don't think I'm stepping on any toes here but I'll make sure to keep in touch
* This is coming from the analytics team, I absolutely do need to give a bit more context here!

app/views/govuk_publishing_components/components/_layout_for_public.html.erb
JonathanHallam: it was just for testing purposes, it honestly hadn't occurred to me that it shouldn't be there, but obviously it shouldn't. I was throwing a lot of stuff at the wall to try and get the tests passing. I've removed it now

spec/components/layout_for_public_spec.rb
JonathanHallam: nope! Sorry, gone!

spec/components/layout_for_public_spec.rb
JonathanHallam: In this iteration I'm treating everything but blanket acceptance as refusal. It won't affect how cookies work on gov.uk but will on other sites as and when they do anything with the link decoration. I've got a ticket on my backlog to allow for more nuanced acceptance

JonathanHallam:
> Another thing I thought of - if different sites have different types of cookies you accept, how will this work from a policy perspective? Accepting cookies on `www.gov.uk` probably means something different to accepting them on a `service.gov.uk` domain and vice versa, as we use cookies for different things.

We're only talking to a very small number of departments about this at the moment, this is an excellent thought but should come out in the wash. This comment kind of applies to your earlier question about link sharing as well, I'm only being responsible for the technical changes here, the risk management and legal side of this are being dealt with by the leads

AshGDS (commented):
Hi @JonathanHallam  - just curious as to why this sits within `window.GOVUK.analyticsGA4` when it applies cookies in general? Shouldn't it be in somewhere else like `cookie-settings.js`, or in it's own module that can then be applied to certain routes?

JonathanHallam:
> Hi @JonathanHallam - just curious as to why this sits within `window.GOVUK.analyticsGA4` when it applies cookies in general? Shouldn't it be in somewhere else like `cookie-settings.js`, or in it's own module that can then be applied to certain routes?

Heya, it's largely because I don't know what I'm doing. I basically need it to run whenever any page is loaded and I figured that would happen within the GA4 code. If it'll happen in `cookie-settings.js` as well I'm happy to move it?

CHANGELOG.md
```diff
@@ -9,8 +9,8 @@
 
 ## Unreleased
 
-* Improve filter Javascript ([PR #5646](https://github.com/alphagov/govuk_publishing_components/pull/5646))
 * Add new brand colour for BIST ([PR #5652](https://github.com/alphagov/govuk_publishing_components/pull/5652))
+* Accept and provide link decoration to/from government links ([PR #5511](https://github.com/alphagov/govuk_publishing_components/pull/5511))
```
andysellick: Changelog entry should match the PR name.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
+document.addEventListener('DOMContentLoaded', function () {
+  if (typeof window.GOVUK.analyticsGa4.init === 'function') {
+    window.GOVUK.analyticsGa4.init()
```
andysellick: I'm guessing you're initialising the analytics code here in order to allow for users arriving with consent set to `yes` but I think this is presumptuous. We already have a mechanism in place for initialising this code when users choose an option from the cookie banner, or if cookies are already set, so we should be using that.

You can see how the cookie banner does it here: https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/components/cookie-banner.js#L67-L76

We also shouldn't initialise GA4 automatically regardless, because we don't want to track users until they consent (that's a big part of ensuring users aren't tracked unless they want to be).

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
...
+  }
+})
+
 var initFunction = function () {
   window.removeEventListener('cookie-consent', window.GOVUK.analyticsGa4.init)
+
+  window.GOVUK.analyticsGa4.checkCookieConsentLinkDecoration = function (location) {
```
andysellick: Just to echo @AshGDS's comment, all of this new code should be in a separate place. I'm not that familiar with it yet but if it could be a new file I think that would be the clearest thing.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
...
+    if (!location || !location.search) return
+
+    var params = new URLSearchParams(location.search)
+
+    var cookieConsent = params.get('cookies[analytics]')
+
+    if (cookieConsent) {
+      if (cookieConsent === 'yes') {
+        window.GOVUK.setConsentCookie({ usage: true })
```
andysellick: I'm interested why this only sets the usage cookie - is that because that's the only one we're interested in, or is that the only common cookie category with the other sites?

AshGDS:
Ah OK, I think it should be in a different file then. Maybe make it similar to how `load-analytics` is called in `_layout_for_public.html.erb`?
https://github.com/alphagov/govuk_publishing_components/blob/031b3d05be98ce2667c51a321fc0272c757cf4c6/app/views/govuk_publishing_components/components/_layout_for_public.html.erb#L52

So something like

`<%= javascript_include_tag "govuk_publishing_components/cookie-consent-link-decoration", type: "module" %>`

Your JS file `app/assets/javascripts/govuk_publishing_components/cookie-consent-link-decoration.js` will need to be inside an immediately invoked function expression, for example :

```
(function () {
  // Define & execute stuff you want to run on load here

  window.GOVUK.checkCookieConsentLinkDecoration = function () { // etc. } 
  window.GOVUK.checkCookieConsentLinkDecoration(window.location) // etc.
}())
```

Alternatively, if you're only testing this on a few pages on GOV.UK, it can be a module, that way we can make the code run only on pages we choose.

EDIT: Removed `.analyticsGa4` from the example JS.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
...
+        window.GOVUK.cookie('cookies_preferences_set', 'true')
+      } else if (cookieConsent === 'no') {
```
andysellick: I don't think this code accounts for a conflict i.e. if you decline cookies on another site, then visit GOV.UK with the URL decoration, having previously accepted cookies. In that situation the URL decoration overrides whatever has previously been set. Is that intentional?

Given that we have the potential issue of people sharing URLs that could include the URL decoration it might be sensible to first check for previously set cookies and let them have priority over the URL decoration?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
...
+
+    links.forEach(link => {
+      try {
+        const url = new URL(link.href, window.location.origin)
+
+        if (allowedDomains.includes(url.hostname)) {
+          url.searchParams.set('cookies[analytics]', consent)
+          link.href = url.toString()
+        }
+      } catch (e) {
```
andysellick: No catch outcome in this block?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
```diff
@@ -1,32 +1,87 @@
...
-      for (var property in analyticsModules) {
-        var module = analyticsModules[property]
-        if (typeof module.init === 'function') {
-          try {
-            module.init()
-          } catch (e) {
-            // if there's a problem with the module, catch the error to allow other modules to start
-            console.warn('Error starting analytics module ' + property + ': ' + e.message, window.location)
+
+  if (consentCookie) {
```
andysellick: I think the changes from this line down to the end of the file shouldn't be necessary if the approach is modified to use the existing consent/analytics init mechanisms.

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
```diff
@@ -80,4 +80,75 @@ describe('Initialising GA4', function () {
...
+      }
+
+      GOVUK.analyticsGa4.checkCookieConsentLinkDecoration(location)
+
+      expect(GOVUK.getCookie('cookies_preferences_set')).toBe('true')
+      expect(GOVUK.getConsentCookie().usage).toBe(true)
+    })
+  })
+
+  describe('Link decoration on the DOM', function () {
```
andysellick: Another suggestion - I think the link decoration and the handling of incoming URL decoration should be separate modules.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: It is intentional I'm afraid. Given that any sites we're working with will have to update the information they give out about cookies to include the fact that they share with GOV.UK we have to assume that if they decline cookies somewhere else they don't want them here either.

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: It's because I was thieving from elsewhere. I should probably update it

https://github.com/alphagov/govuk_publishing_components/commit/510201ab7ebc172e10cd70e5b3126ad689e3b340#diff-72447199be3fbd810689803d9be76395cb480bd97ab87622be60ae168c51d235R18-R23

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: Done

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: It's actually because it's where it was in the code that I borrowed from DI. I've made a new module now and I'll change how it's called

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: Have now updated this to count the number of consents on the cookie object. 

https://github.com/alphagov/govuk_publishing_components/pull/5511/changes/70861c1e431cfcd7fc81e5763e4dc568a4382575

I'm not sure it's the most elegant solution. But it does what I want it to

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: I did this so that it'd silently fail. I didn't want it to show errors to the user if link decoration failed. Do you think I should remove it or make a new error object?

spec/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.spec.js
JonathanHallam: I've put it in one for now and it's working in the way that I'd hoped. Let me know what you think

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
```diff
@@ -0,0 +1,61 @@
...
+      var link = links[i]
+      try {
+        var url = new URL(link.href, window.location.origin)
+
+        if (allowedDomains.includes(url.hostname)) {
+          url.searchParams.set('cookies', consentValue)
+          link.href = url.toString()
+        }
+      } catch (e) {
+      }
```
AshGDS, line 49: Empty catch statement, maybe instead do `console.error` here?

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
```diff
@@ -0,0 +1,61 @@
...
+
+  Modules.CookieConsentLinkDecoration = CookieConsentLinkDecoration
+})(window.GOVUK.Modules)
+
+document.addEventListener('DOMContentLoaded', function () {
+  if (window.GOVUK && window.GOVUK.Modules && window.GOVUK.Modules.CookieConsentLinkDecoration) {
+    var linkDecoration = new window.GOVUK.Modules.CookieConsentLinkDecoration(document)
+    linkDecoration.init()
+  }
+})
```
AshGDS, line 61: This isn't how we usually use these modules - instead you could add `data-module="cookie-consent-link-decoration"` to the page HTML (e.g. on the `body`) and then our GOVUK Modules code will handle executing the code.

spec/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration-spec.js
```diff
@@ -0,0 +1,80 @@
...
+
+    it('sets usage consent cookie to false when cookies[analytics] query string parameter has a value of "no"', function () {
+      window.history.replaceState(null, null, '?cookies=no')
+
+      var consentModule = new window.GOVUK.Modules.CookieConsentLinkDecoration()
+      consentModule.init()
+
+      expect(GOVUK.getCookie('cookies_preferences_set')).toBe('true')
+      expect(GOVUK.getConsentCookie().usage).toBe(false)
+    })
```
AshGDS, line 25: Can these tests also test the cookie state before your code has run otherwise this could result in false positives?

spec/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration-spec.js
```diff
@@ -0,0 +1,80 @@
...
+
+  it('appends positive cookie decoration to relevant links after cookies are accepted', function () {
+    GOVUK.setCookie('cookies_policy', '{"essential":true,"settings":true,"usage":true,"campaigns":true}')
+
+    var consentModule = new window.GOVUK.Modules.CookieConsentLinkDecoration()
+    consentModule.decorateLinks()
+
+    expect(link1.href).toContain('cookies=yes')
+    expect(link2.href).toEqual('https://www.example.gov.uk/')
+  })
```
AshGDS, line 79: This is intending to check links after cookies are accepted, but your module isn't running at the point in which cookies are accepted. Shouldn't this test be ensuring all the links update if the cookie consent changes midway through the user browsing the page?

AshGDS (commented):
Looking good 👍 Haven't run the code locally yet but had some comments based off of looking at the code

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
andysellick: Okay, thanks for the clarification. Given that there are nuances here, would it be worth adding some documentation to explain this code, including subtle aspects like this?

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
andysellick: I think it'd be okay to add a `console.error` as that still amounts to failing silently.

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
```diff
@@ -0,0 +1,61 @@
...
+        window.GOVUK.declineNonEssentialCookieTypes()
+        window.GOVUK.cookie('cookies_preferences_set', 'true')
+      }
+    }
+  }
+
+  CookieConsentLinkDecoration.prototype.decorateLinks = function () {
+    var consentCookie = window.GOVUK.getConsentCookie ? window.GOVUK.getConsentCookie() : null
+    var consentCount = Object.values(consentCookie || {}).filter(val => val === true).length
+    var consentValue = consentCount === 4 ? 'yes' : 'no'
```
andysellick, line 24: Currently it looks like if the user hasn't chosen anything in the cookie banner the link decoration will apply `?cookies=no` to relevant outgoing links, which I think is wrong - that would mean other sites visited will reject cookies and hide the cookie banner, but the user has yet to make a choice.

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
```diff
@@ -0,0 +1,61 @@
...
+    var links = document.querySelectorAll('a[href]')
```
andysellick: I think you could save yourself some iterations here. Instead of getting all of the links on the page then iterating through them to check if they match the domains, you could only get the links that are external i.e. that start with `https`.

```Javascript
var links = document.querySelectorAll("[href^='https']")
```

It's not perfect but on the homepage for example it cuts the number of links to iterate from 111 to 5.

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
```diff
@@ -0,0 +1,61 @@
...
+    var allowedDomains = [
+      'end-to-end-journeys-545890405086.europe-west2.run.app',
+      'x-domain-prototype-2-545890405086.europe-west2.run.app',
+      'x-domain-prototype-3-545890405086.europe-west2.run.app'
+    ]
+
+    for (var i = 0; i < links.length; i++) {
+      var link = links[i]
```
andysellick: You can save yourself a variable declaration here by using `forEach` e.g.

```JavaScript
links.forEach((link) => {
  try {
    var url = new URL(link.href, window.location.origin)
    // etc.
  }
})
```

app/assets/javascripts/govuk_publishing_components/analytics-ga4/init-ga4.js
JonathanHallam: cool, done :)

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: done! :)

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
andysellick: @AshGDS is right although we probably don't want to add a data attribute to every page. It might be better to initialise this code not as a module (it doesn't use `this.$module` for example) but maybe more like `cookie-functions.js`, which I think starts automatically?

andysellick (commented):
I was part way through a review and another commit has been pushed, so I'm going to pause here and check again later 👍

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: Thanks, done

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: Done :)

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: Good shout, have done :)

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: You're absolutely right that was happening, I think I've fixed it but addressing ash's comments on the testing strategy will make it a touch more certain

CHANGELOG.md
andysellick: Bumping this comment to make sure it doesn't get missed. You could put a short file in `docs` including an overview of what it does and why, and the exact details of the link decoration.

spec/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration-spec.js
JonathanHallam: Nice shout. For this suggestion and your other below I've added a line to each test that checks the state before the cookies are set

spec/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration-spec.js
JonathanHallam: see above

app/assets/javascripts/govuk_publishing_components/lib/cookie-consent-link-decoration.js
JonathanHallam: have fixed this by adding code to the dependencies file

CHANGELOG.md
JonathanHallam: Will do, I've pretty heavily edited the PR description since you made this comment but I'll add something into the docs section too

Issue #5103, opened 2025-10-27
Presenter test coverage could be improved


andysellick:
Looking through our presenter code it seems we could write more tests - a number of presenters have no test file.

Some presenters are either very simple or rely heavily on private methods, so presumably are mostly tested by component tests and may not require a specific test file.

Presenters that might be in need of a test file are:

- [x] big_number_helper.rb
- [x] button_helper.rb
- [x] checkboxes_helper.rb
- [x] image_card_helper.rb
- [x] organisation_logo_helper.rb
- [ ] select_helper.rb
- [ ] subscription_links_helper.rb
- [ ] translation_nav_helper.rb

Below is a complete list, including where presenters already have a test file or where they probably don't need one.

```
# lib/govuk_publishing_components/presenters

accordion_helper.rb (not sure much to test)
attachment_helper.rb - attachment_helper_spec.rb
big_number_helper.rb
breadcrumb_selector.rb - breadcrumb_selector_spec.rb
breadcrumbs_helper.rb (not sure much to test)
button_helper.rb
checkboxes_helper.rb
component_wrapper_helper.rb - component_wrapper_helper_spec.rb
content_breadcrumbs_based_on_ancestors.rb - content_breadcrumbs_based_on_ancestors_spec.rb
content_breadcrumbs_based_on_organisations.rb - content_breadcrumbs_based_on_organisations_spec.rb
content_breadcrumbs_based_on_taxons.rb - content_breadcrumbs_based_on_taxons_spec.rb
content_item.rb - content_item_spec.rb
contents_list_helper.rb - contents_list_helper_spec.rb
contextual_navigation.rb - contextual_navigation_spec.rb
curated_taxonomy_sidebar_links.rb (not sure much to test)
devolved_nations_helper.rb - devolved_nations_helper_spec.rb
emergency_banner_helper.rb (not sure much to test)
heading_helper.rb (not sure much to test)
image_card_helper.rb
intervention_helper.rb - intervention_helper_spec.rb
layout_footer_helper.rb - layout_footer_helper_spec.rb
meta_tags.rb (not sure much to test)
organisation_logo_helper.rb
page_with_step_by_step_navigation.rb - page_with_step_by_step_navigation_spec.rb
public_layout_helper.rb - public_layout_helper_spec.rb
related_navigation_helper.rb - related_navigation_helper_spec.rb
schema_org.rb - schema_org_spec.rb
select_helper.rb
[...]

PR #5513, opened 2026-06-19
Add helper tests to component auditing

andysellick:
## What
- include spec files for component presenters/helpers in the auditing
- adds a new column to the 'Component files' section of the 'Components' tab in the auditing, showing if a component has a spec/test file for any helpers, or not

## Why
Was not included in the auditing, and should be - initial checks show not all helpers have an associated test file.

## Visual Changes
New column 'Helper test' added to the right.

Screenshot 2026-06-19 at 16 34 27


andysellick:
This PR relates to (but doesn't fix) this issue: https://github.com/alphagov/govuk_publishing_components/issues/5103

MartinJJones (approved):
Nice work, the changes look good to me, I tried out the new feature in the `frontend` app as well and looks really useful :)

PR #5525, opened 2026-06-24
Fix tabs documentation

andysellick:
## What
- the code example for the tabs component was showing a load of data attributes in the code snippet that weren't being passed to the component from the YML
- these attributes weren't needed for the example shown and shouldn't have been there
- on investigation, it turns out that when the component was rendered the default data attributes for tracking were being created but then appended into the data passed to the the component, hence the corruption of the code example in the guide page
- replacing this with a simple variable fixes the issues

## Why
The documentation was inaccurate, suggesting that data attributes needed to be passed when they did not.

## Visual Changes
Before
Screenshot 2026-06-24 at 15 57 00

After
Screenshot 2026-06-24 at 15 57 07


AshGDS (approved):
LGTM 👍

PR #5527, opened 2026-06-25
Attachment component changes

jon-kirwan:
## What

- Standardise some aspects of the attachment components (the Whitehall block attachment used in Govspeak and the attachment component).
- Use multiple box shadows to achieve the same visual effect as having both a box shadow and an outline.
- Make the high-contrast outline consistent throughout, following current Firefox behaviour

## Why

This relates to investigating issues encountered during the v6 upgrade, where we've found that the use of `color.adjust` and, in some cases, `darken` can cause compilation failures. Following a design review, we'd like to standardise these colours using GOV.UK Frontend colours where possible https://docs.google.com/document/d/10bvnIGAgSv4BE1tzP-c2FTEtmB4cp_Qr9e59H7Jn3jA/edit?disco=AAAB9F6l1fQ.

As these colours also use semi-transparency, some reworking is needed to ensure the box shadow remains visible; I've used multiple box shadows to achieve the same visual effect (which has good browser support https://caniuse.com/?search=multiple+box+shadows).

I've also removed some older border styles that appear to have been intended for IE9/IE10 support and are no longer needed (see attached screenshots).

### IE9 (Before)

bs_win7_IE_9 0 (1)

### IE9 (After, no difference)

bs_win7_IE_9 0

### IE10 (Before)

bs_win8_IE_10 0

### IE10 (After, no difference)

bs_win8_IE_10 0 (1)

### Chrome, forced colours (Before)

Screenshot 2026-06-24 at 14 52 40

### Chrome, forced colours (After)

Screenshot 2026-06-24 at 14 52 50

#### Edge (Before)

Screenshot 2026-06-24 at 11 27 58

#### Edge (After)

Screenshot 2026-06-24 at 11 27 09

#### Firefox, custom settings (Before)

Screenshot 2026-06-24 at 11 29 03

#### Firefox, custom settings After, no difference)

Screenshot 2026-06-24 at 11 28 32

#### Firefox automatic (system settings) (Before)

Screenshot 2026-06-24 at 11 29 14

#### Firefox automatic (system settings) (After, no difference)

Screenshot 2026-06-24 at 11 28 22

## Anything else

- I’m not sure why earlier versions of IE add a blue border around the image (as it still appears to do after the change), but since it’s an existing issue and not causing any problems, I haven’t tried to fix it.
- It was also noticed that the current high-contrast styling adds an outline, but this only showed in Firefox. The recent changes removed that outline, so I’ve made another update to bring the behaviour back in line across all tested browsers. That said, this still might not be the best way to handle high-contrast mode in these components, and it probably needs a bit more design input to figure out the right approach.
- I spent some time trying to get this approach to work in Chrome with Auto Dark Mode enabled, but ultimately had to abandon it due to a range of inconsistent behaviours. Since this is an experimental feature on Android and is otherwise only available through DevTools, I've not supported it.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@
 
 * Remove extra breadcrumbs padding on mobile ([PR #5501](https://github.com/alphagov/govuk_publishing_components/pull/5501))
 * Add auto layout option to cards component ([PR #5499](https://github.com/alphagov/govuk_publishing_components/pull/5499))
+* Use multiple box shadows in attachment components in place of outline and box shadow ([PR #5503](https://github.com/alphagov/govuk_publishing_components/pull/5503))
 
 ## 66.5.0
 
--- a/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_attachment.scss
@@ -1,7 +1,6 @@
 $thumbnail-width: 99px;
 $thumbnail-height: 140px;
 $thumbnail-border-width: 5px;
-$thumbnail-background: govuk-colour("white");
 $thumbnail-shadow-colour: rgba(11, 12, 12, .4);
 $thumbnail-shadow-width: 0 2px 2px;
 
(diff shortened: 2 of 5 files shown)


MartinJJones (approved):
Nice work on this, the changes look great!

luke-booker:
Changes all look good, Jon 👍

PR #5531, opened 2026-06-25
Add translation navigation helper spec

andysellick:
## What
Adds a missing test file for the translation nav helper.

## Why
There wasn't one, see issue https://github.com/alphagov/govuk_publishing_components/issues/5103

## Visual Changes
None.


AshGDS (approved):
LGTM 👍 

Non-blocking, but maybe this component should have locale validation to ensure the locale is correct, similar to the component wrapper helper's `check_lang_is_valid` method

andysellick:
> Non-blocking, but maybe this component should have locale validation to ensure the locale is correct, similar to the component wrapper helper's `check_lang_is_valid` method

Sounds like a good idea. I'll raise an issue.

PR #5539, opened 2026-06-29
Stop using size 14 from the type scale

luke-booker:
## What
We are replacing all instances of 14px typography across GOV.UK Publishing Components and frontend apps. This involves removing or updating Sass mixins passing $size: 14, eliminating utility classes like govuk-body-xs and govuk-!-font-size-14, and clearing out any raw 14px/rem equivalents.

## Why
Size 14 is being officially removed from the type scale. We want to update these references to an approved font size to avoid Sass triggering a compilation error and fail to build.

Jira card: https://gov-uk.atlassian.net/browse/NAV-18915

## Visual Changes
| Before | After |
|--------|--------|
| Screenshot 2026-06-29 at 15 37 50 | Screenshot 2026-06-29 at 15 39 28 |

### DIFF
--- a/app/views/govuk_publishing_components/components/_glance_metric.html.erb
+++ b/app/views/govuk_publishing_components/components/_glance_metric.html.erb
@@ -30,6 +30,6 @@
         <% end %>
     <% end %>
     </span>
-    <p class="gem-c-glance-metric__context govuk-body-xs"><%= context %></p>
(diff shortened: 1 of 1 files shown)


MartinJJones (approved):
Nice work, the change looks good to me, I cannot see any other references to the `-xs` or `14`px font size

deberny:
All looks good to me 🚀

PR #5543, opened 2026-07-03
Add GA4 tracking to cookie banner buttons and link

jon-kirwan:
## What

Add GA4 event tracking to the cookie banner's **Accept**, **Keep**, and **View cookie settings** buttons and link.

## Why

Aggregate tracking is now enabled by default, so we want to track clicks on the **Accept** and **Keep** buttons, as well as the **View cookie settings** link.

### DIFF
--- a/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
+++ b/app/views/govuk_publishing_components/components/_cookie_banner.html.erb
@@ -43,7 +43,13 @@
               </ul>
               <p class="govuk-body"><%= t("components.cookie_banner.text_additional_cookies") %></p>
             <% end %>
-            <p class="govuk-body govuk-!-static-margin-bottom-4">
+            <p class="govuk-body govuk-!-static-margin-bottom-4"
+            <% unless disable_ga4 %>
+                data-ga4-cookie-banner
+                data-module="ga4-link-tracker"
+                data-ga4-set-indexes
+                data-ga4-link="<%= { event_name: "navigation", type: "cookie banner", section: t("components.cookie_banner.title", locale: :en) }.to_json %>"
+              <% end %>>
               <a class="govuk-link" href="<%= cookie_preferences_href %>"><%= t("components.cookie_banner.buttons.view_cookies") %></a>
             </p>
           </div>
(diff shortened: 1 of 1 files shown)


app/views/govuk_publishing_components/components/_cookie_banner.html.erb
MartinJJones: From looking at the [card](https://gov-uk.atlassian.net/browse/NAV-18957), it appears that tracking is also required here in the scenario where user has selected "Keep current settings" and aggregate is set to true

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
```diff
@@ -59,19 +65,41 @@
...
+        }.to_json unless disable_ga4
+      %>
       <%= render "govuk_publishing_components/components/button", {
         name: "cookies",
         text: t("components.cookie_banner.buttons.accept_cookies"),
         data_attributes: {
           "accept-cookies": "true",
           "cookie-types": "all",
+          "module": ("ga4-event-tracker" unless disable_ga4),
+          "data-ga4-event": ga4_event_accept,
```
MartinJJones: The ga4 events do not appearing to be working at the moment when [testing](https://github.com/alphagov/govuk_publishing_components/blob/54edb0d4535163478793a894a0f3da31c234d823/docs/analytics-ga4/developer-guide.md?plain=1#L12) in the browser, I think this is likely because of the duplicate `data-` value in the `data-data-ga4-event=`, with the current approach, I think `data-ga4-event` should change to `ga4-event`.

Although it might be possible to reduce the duplication a bit further by attaching the `data-module` attribute to the parent div, and adding the values required to each button using the `ga4-event` attribute, further info here - https://github.com/alphagov/govuk_publishing_components/blob/54edb0d4535163478793a894a0f3da31c234d823/docs/analytics-ga4/trackers/ga4-event-tracker.md#basic-use

MartinJJones (commented):
The changes are looking good, just a couple of changes/suggestions after testing locally, happy to catch up on this further anytime, thanks

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
jon-kirwan: Oh yes, good spot. I think I've added that correctly now.

app/views/govuk_publishing_components/components/_cookie_banner.html.erb
jon-kirwan: Great, thanks - I’ve fixed the invalid data attribute and moved it to the parent container. The events now seem to be firing correctly and updating the dataLayer, so it looks like it’s working as expected.

MartinJJones (approved):

PR #5549, opened 2026-07-03
Add option to change Service navigation toggle text

andysellick:
## What
- adds an option to the service navigation component to change the text of the menu toggle on mobile, that shows/hides the rest of the menu items
- previously was only 'menu'

## Why
We're going to need this option for some upcoming work.

## Visual Changes

Screenshot 2026-07-03 at 15 32 12

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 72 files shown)


AshGDS (approved):
LGTM!

PR #5558, opened 2026-07-07
Add chartkick to the gemspec

AshGDS:
## What / Why
- Chartkick was added to the gem as a dev dependency, but it needs to be in the gemspec to work properly when used in other applications, so this adds it to the gemspec. Without this, rendering the component in a rendering app crashes.
- Related to https://github.com/alphagov/govuk_publishing_components/pull/5484
## Visual Changes

None.


andysellick (approved):
I think I missed this myself the first time 😁

PR #5561, opened 2026-07-08
Set chart component's chartArea to auto

AshGDS:
## What/Why

- Google Charts has a `chartArea` value, which sets the size and positioning of the chart within its container
- Previously, all charts were hardcoded to take up 80% of the width of its container, and 60% of the height
- In practice this reduces the responsiveness of the charts, because the chart was always forced to take up 80% of the container width, even if that constrained the rendering of the labels. This was the case when I tried to use this component on real data on GOV.UK
- Therefore, we set the `width` and `height` to `auto` by default, allowing the chart to decide how it can render most optimally.

## Visual Changes (default rendering)

| Before | After |
|--------|--------|
| image | image |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
+* Set chart component's chartArea to auto ([PR #5561](https://github.com/alphagov/govuk_publishing_components/pull/5561))
 * Add chartkick to the gemspec ([PR #5558](https://github.com/alphagov/govuk_publishing_components/pull/5558))
 * Do not render hidden "share" text elements if no hidden text is provided ([PR #5554](https://github.com/alphagov/govuk_publishing_components/pull/5554))
 
--- a/lib/govuk_publishing_components/presenters/chart_helper.rb
+++ b/lib/govuk_publishing_components/presenters/chart_helper.rb
@@ -30,7 +30,7 @@ module GovukPublishingComponents
       # config options are here: https://developers.google.com/chart/interactive/docs/gallery/linechart
       def chart_options
         {
-          chartArea: { width: "80%", height: "60%" },
+          chartArea: { width: "auto", height: "auto" },
           crosshair: { orientation: "vertical", trigger: "both", color: "#ccc" },
           curveType: "none",
           enableInteractivity: @enable_interactivity,
--- a/spec/lib/govuk_publishing_components/presenters/chart_helper_spec.rb
+++ b/spec/lib/govuk_publishing_components/presenters/chart_helper_spec.rb
@@ -10,7 +10,7 @@ RSpec.describe GovukPublishingComponents::Presenters::ChartHelper do
       }
 
       expected = {
-        chartArea: { width: "80%", height: "60%" },
+        chartArea: { width: "auto", height: "auto" },
         crosshair: { orientation: "vertical", trigger: "both", color: "#ccc" },
         curveType: "none",
         enableInteractivity: true,


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -423,3 +423,34 @@ examples:
...
+      Configures the placement and size of the chart itself within its parent container. By default, this is set to `{ width: "auto", height: "auto" }`.
+
+      You can also pass through `left`, `right`, `top`, and `bottom` values for alignment on the horizontal and vertical axis of the parent container.
+    data:
+      chart_area:
+        width: 10%
+        height: 30%
+        left: 10%
+        bottom: 50%
+        top: 0
```
andysellick: Is the value for `top` needed here? There's a value for `left` but not for `right`, so does it work like CSS absolute positioning where you only need one of each pair, i.e. either `top` or `bottom`, either `left` or `right`, or is it more complicated? I think it would be good to clarify the documentation on this point.

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -423,3 +423,34 @@ examples:
             - 118
             - 85
             - 80
+  with_a_different_chart_area:
+    description: |
+      Configures the placement and size of the chart itself within its parent container. By default, this is set to `{ width: "auto", height: "auto" }`.
+
+      You can also pass through `left`, `right`, `top`, and `bottom` values for alignment on the horizontal and vertical axis of the parent container.
```
andysellick: Nitpick: I think the word `also` is unnecessary here?

app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -423,3 +423,34 @@ examples:
             - 118
             - 85
             - 80
+  with_a_different_chart_area:
+    description: |
+      Configures the placement and size of the chart itself within its parent container. By default, this is set to `{ width: "auto", height: "auto" }`.
```
andysellick: Good to have this as a fallback option if the auto doesn't work, but I'm wondering whether we'd recommend auto and describe this option as only for advanced use? In what circumstances would auto not work properly?

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Thanks - yeah it probably only needs one of top or bottom.

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Cool, can remove 👍

app/views/govuk_publishing_components/components/docs/chart.yml
AshGDS: Yeah, we might not need this - In the original PR the fixed widths were needed for something on the missions pages, looks like originally it was something to do with if there are over 99999 rows? I'm guessing we'd never want to support that anyway? https://github.com/alphagov/govuk_publishing_components/commit/3ffdb83f8a6503540468fa5a6005fa1f30973229#diff-a22187b2bd3eb0c9613717c69d527eba260f559aba83d2e22c057b2a6aafb72cR41

AshGDS:
Thanks @andysellick - looks like we can probably remove setting the chartArea entirely and keep it as auto, what do you think?

andysellick:
> Thanks @andysellick - looks like we can probably remove setting the chartArea entirely and keep it as auto, what do you think?

Yeah, the more I think about it the more that seems like the best approach 👍

AshGDS:
Thanks @andysellick - I've done that now 👍

andysellick (approved):
👏

PR #5562, opened 2026-07-08
Add options to the service navigation component

andysellick:
## What
Adds two options to the service navigation component.

## Why
Trying out some options to allow us to fit the service navigation component to a possible design.

## Visual Changes
Columns layout is relatively easy to demonstrate. The links sit to one side of the service name. Appearance on mobile is unchanged.

Screenshot 2026-07-08 at 12 16 23

The links on mobile option is a little more complicated. We want to keep the service name as distinct from the rest of the links on desktop, but avoid duplication of that text/link on mobile. So we add a duplicate link to the service name that is only shown on mobile, because the service name link is hidden on mobile.

Essentially, we're allowing the `My service home` link to be shown on desktop and mobile, but using different elements.

On desktop:

Screenshot 2026-07-08 at 12 19 31

On mobile:

Screenshot 2026-07-08 at 12 19 42

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/service_navigation.yml
+++ b/app/views/govuk_publishing_components/components/docs/service_navigation.yml
@@ -70,6 +70,25 @@ examples:
         href: "#"
       - text: Navigation item 3
         href: "#"
+  columns_layout:
+    description: Separates the service name and the navigation items into two separate columns on desktop. Has no effect on mobile or when the service name is not present.
+    data:
+      service_name: My service home
+      service_name_url: "/home"
+      columns_layout: true
+      navigation_items:
+      - text: Introduction
+        href: "#"
+      - text: Help and guidance
+        href: "#"
+      - text: The application process
+        href: "#"
+      - text: Your first application
+        href: "#"
+      - text: Contact us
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/service_navigation.yml
```diff
@@ -52,6 +52,42 @@ examples:
...
+      navigation_items:
+      - text: My service home
+        href: "/home"
+      - text: Navigation item 1
+        href: "#"
+      - text: Navigation item 2
+        href: "#"
+      - text: Navigation item 3
+        href: "#"
+  columns_layout:
```
AshGDS, line 73: Should this have a description which describes what it does?

AshGDS (commented):
LGTM - will approve once there's a Changelog entry 👍

app/views/govuk_publishing_components/components/docs/service_navigation.yml
andysellick: Added (and added a changelog entry, thanks).

AshGDS (approved):

PR #5567, opened 2026-07-10
Add header_three_quarters option to the chart component

AshGDS:
## What / Why

- The chart component had a `max-width: 66%` restriction on the heading and description
- Instead use `max-width: 630px` otherwise when the heading is in a two-thirds column, it will only take up 66% of the width of the column
- Also add a `header_max_width` option that allows the value to be changed to accommodate for different column widths

## Visual Changes

| Before | After |
|--------|--------|
| image  | image |

### DIFF
--- a/app/views/govuk_publishing_components/components/_chart.html.erb
+++ b/app/views/govuk_publishing_components/components/_chart.html.erb
@@ -7,6 +7,9 @@
   download_link ||= false
   heading ||= nil
   description ||= nil
+  header_three_quarters ||= false
+  header_class = "gem-c-chart__header"
+  header_class << "--three-quarters" if header_three_quarters
   source ||= false
 
   if !heading || !description
@@ -33,7 +36,7 @@
     <% @include_script[:script_included] = true %>
   <% end %>
   <%= tag.div(**component_helper.all_attributes) do %>
-    <div class="gem-c-chart__heading">
+    <div class="<%= header_class %>">
         <%= render "govuk_publishing_components/components/heading", {
           text: heading,
           heading_level: heading_level,
--- a/app/views/govuk_publishing_components/components/docs/chart.yml
+++ b/app/views/govuk_publishing_components/components/docs/chart.yml
@@ -423,3 +423,42 @@ examples:
             - 118
             - 85
             - 80
+  with_a_different_header_max_width:
+    description: |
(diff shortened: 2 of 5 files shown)


app/views/govuk_publishing_components/components/docs/chart.yml
```diff
@@ -423,3 +423,42 @@ examples:
             - 118
             - 85
             - 80
+  with_a_different_header_max_width:
+    description: |
+      By default, the `max-width` of the header (heading and description) is set to `630px` which is the width of a standard GOV.UK two-thirds column. This prevents the header being too wide for legibility purposes if it is rendered in a full width container.
+
+      In some cases you might want the `max-width` to match a different column width. Therefore you can set `header_max_width` to another value like `three-quarters` (`two-thirds` is the default setting).
```
andysellick: Nitpick: the wording here suggests there are other options? I'd just say `You can set header_max_width to three-quarters` (also it already says two-thirds is the default).

spec/components/chart_spec.rb
```diff
@@ -144,4 +144,24 @@ def component_name
...
     }.to raise_error("A heading and description must be provided for accessibility purposes.")
   end
+
+  it "has a two-thirds max width header by default" do
+    render_component(data)
+
+    assert_select ".gem-c-chart__header--two-thirds"
+  end
+
+  it "can have a different header max width by default" do
```
andysellick: Nitpick: maybe just `can have a different header max width`?

app/views/govuk_publishing_components/components/_chart.html.erb
```diff
@@ -7,6 +7,8 @@
   download_link ||= false
   heading ||= nil
   description ||= nil
+  header_max_width = header_max_width.in?(%w[two-thirds three-quarters]) ? header_max_width : "two-thirds"
+  header_class = "gem-c-chart__header--#{header_max_width}"
```
andysellick: I think this option would be cleaner as a true/false flag, used to add a modifier class (setting the width to three quarters) to the original class (which would set the element to the two thirds width) e.g.

```
header_three_quarters ||= false
header_class = %w[gem-c-chart__heading]
header_class << "gem-c-chart__heading--three-quarters" if header_three_quarters
```

AshGDS:
Thanks @andysellick - I've refactored it to use your proposed solution 👍

andysellick (approved):
Nice work 👍

PR #5572, opened 2026-07-10
Merge published dates component into metadata component

### LINKED ISSUE #5560: Metadata block and Published dates components could be merged
The [metadata block](https://components.publishing.service.gov.uk/component-guide/metadata) and [published dates](https://components.publishing.service.gov.uk/component-guide/published_dates) components are visually similar and perform very similar tasks and could be merged into a single component. 

The difference is that published dates can display a specifically formatted page history of details and dates (using `time` elements) in an expandable element. Metadata has a similar option, but the 'see all updates' text is a link to something else, rather than an expanding element. However, it seems like this could be an option on the metadata component without too much difficulty.

Visual similarities:

Metadata | Published
------- | --------
Image | Image

The published dates component also has a line along the top when it shows expandable information:

Image

andysellick:
## What
Copy the functionality from the published dates component into the metadata component.

Doesn't delete the published dates component - this PR is deliberately non-breaking, so we can remove the published dates component when it is safe to do so.

Notes for switching from the published dates to the metadata component:

- the `history` option needs to change to `page_history`
- the `published` option needs to change to `first_published`

It's worth noting the behaviour of this component on certain pages like [this call to action page](https://www.gov.uk/government/case-studies/reception-quality-webinar-series). There's a link at the top (in the metadata component) to `See all updates`, when clicked jumps to the published dates component at the bottom, and then auto-expands the `Show all updates` link. This is handled by some JS in the metadata component. The change now will be that the published dates component at the bottom of the page will be replaced with the metadata component, but it should all still function as before.

I've opened [this draft PR](https://github.com/alphagov/frontend/pull/5661) to check that switching to the metadata component with these new options works correctly, and all seems okay.

Examples:

- [case study](https://www.gov.uk/government/case-studies/hm-government-licensing-cyber-security-tech-for-a-global-market)
- [document collection](https://www.gov.uk/government/collections/visa-processing-times)
- [fatality notice](https://www.gov.uk/government/fatalities/fourteen-personnel-in-afghanistan-nimrod-crash-named)
- [news article](https://www.gov.uk/government/news/ukhsa-updates-heat-health-alerts-across-england)
- [service manual guide](https://www.gov.uk/service-manual/helping-people-to-use-your-service/understanding-wcag)
- [help page](https://www.gov.uk/help/beta)
- [call for evidence](https://www.gov.uk/government/calls-for-evidence/carers-allowance-call-for-evidence)
- [consultation](https://www.gov.uk/government/consultations/growing-up-in-the-online-world-a-national-consultation)
- [detailed guide](https://www.gov.uk/guidance/verify-your-identity-for-companies-house)
- [publication](https://www.gov.uk/government/publications/budget-2016-documents)
- [specialist document](https://www.gov.uk/cma-cases/veterinary-services-market-for-pets-review)
- [speech](https://www.gov.uk/government/speeches/we-are-appalled-by-the-continued-restrictions-imposed-on-the-women-and-girls-of-afghanistan-uk-statement-at-the-un-security-council)
- [statistical data set](https://www.gov.uk/government/statistical-data-sets/unclaimed-estates-list)

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/5560

## Visual Changes
The new part of the metadata component should look exactly the same as the published dates component, although there is more line spacing between items in the metadata component.

Published dates | Metadata
------- | -------
Screenshot 2026-07-13 at 09 15 21 | Screenshot 2026-07-13 at 09 15 42
Screenshot 2026-07-13 at 09 17 34 | Screenshot 2026-07-13 at 09 17 40


AshGDS (commented):
Code looks OK to me, but on the live site the page history shows when JS disabled, but it doesn't on your Heroku example, so this might need to be fixed first? It's probably worth adding the Changelog entry too so the approval will be smoother.

(As an aside, a `Cancel` button is rendering on our Feedback component when JS is disabled, which we should fix)

AshGDS:
This is the page I tested https://www.gov.uk/government/news/ukhsa-updates-heat-health-alerts-across-england#full-publication-update-history

AshGDS:
Ah actually, it might be my fault - I was testing using Heroku rather than a local version of `frontend` 😄  Give me a minute.

AshGDS (commented):
Looks all OK to me - will need a changelog entry though 👍 

Possibly something to improve - the "Show all updates" link is acting like a button, but it doesn't expand or collapse when the Spacebar is used to interact with it like other buttons do, which I think might be something to fix?

andysellick:
> Possibly something to improve - the "Show all updates" link is acting like a button, but it doesn't expand or collapse when the Spacebar is used to interact with it like other buttons do, which I think might be something to fix?

Thanks @AshGDS. Is it okay if I raise that separately? The expand/collapse for this component is now handled by the `toggle.js` script, so it would be something to fix there. Also I agree Spacebar should work, but worth noting that the current published dates component doesn't support that key either.

Have added a changelog entry 👍

AshGDS (approved):
Yeah no worries, LGTM 👍

PR #5585, opened 2026-07-15
Move to using `govuk-functional-colour` for borders and secondary text

hannalaakso:
## What
Replace instances where`govuk-colour("black", $variant: "tint-80")` is being used as border colour with `govuk-functional-colour("border")`. Also replace some instances of it in `box-shadow` where it is clearly being used as a border; if usage is more ambigious, do not replace with `govuk-functional-colour("border")`.

Replace instances where `govuk-colour("black", $variant: "tint-25")` is being used as secondary text colour with `govuk-functional-colour("secondary-text")`.

## Why
https://gov-uk.atlassian.net/browse/NAV-18929

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
@@ -39,7 +39,7 @@
         width: $thumbnail-width;
         height: 140px;
         background: govuk-colour("white");
-        box-shadow: 0 2px 2px govuk-colour("black", $variant: "tint-50"), 0 0 0 5px govuk-colour("black", $variant: "tint-80");
+        box-shadow: 0 2px 2px govuk-colour("black", $variant: "tint-50"), 0 0 0 5px govuk-functional-colour("border");
 
         @include high-contrast-outline;
       }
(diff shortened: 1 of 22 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
```diff
@@ -39,7 +39,7 @@
         width: $thumbnail-width;
         height: 140px;
         background: govuk-colour("white");
-        box-shadow: 0 2px 2px govuk-colour("black", $variant: "tint-50"), 0 0 0 5px govuk-colour("black", $variant: "tint-80");
+        box-shadow: 0 2px 2px govuk-colour("black", $variant: "tint-50"), 0 0 0 5px govuk-functional-colour("border");
```
MartinJJones, line 42: I think this is a nice change, I was a bit unsure at first, but I now understand that the border-box style here is used to add a border, so I find this more descriptive

MartinJJones (approved):
Nice work, the changes look good to me

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss
hannalaakso: Same, I considered leaving it as is but seeing that the box-shadow actually renders a border I think it does help the purpose.

Issue #5580, opened 2026-07-14
toggle.js doesn't work with spacebar


andysellick:
We have a generic script called [toggle.js](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/javascripts/govuk_publishing_components/lib/toggle.js) that is used to show/hide elements in a variety of places. 

The elements that get converted into toggles behave like buttons but don't respect the use of Spacebar for their toggling, as buttons should.

Noted in https://github.com/alphagov/govuk_publishing_components/pull/5572

maxelliotmingail-afk:
I prepared a focused fix and regression coverage for this keyboard gap.

The patch adds Space/legacy `Spacebar` handling to the toggle trigger, prevents the page-scroll default and delegates activation to the existing click path. The tests verify that Space changes `aria-expanded`, reveals the controlled target and is cancelled, while an unrelated key does nothing.

Verification completed:

- StandardX on both changed JavaScript files: passed;
- `git diff --check`: passed.

The full Jasmine command requires Ruby asset compilation, but Ruby is unavailable on this workstation. GitHub's fork endpoint is returning HTTP 403 for my authenticated account, so the complete `git am`-compatible patch is published here instead of being described as a PR: https://gist.github.com/maxelliotmingail-afk/eed57b132976d0d9650166eae7661ed0

PR #5592, opened 2026-07-17
Use the details component for toggling element visibility in the metadata component

andysellick:
## What
Change the metadata component so that show/hide of elements is handled by the details component instead of the `gem-toggle` script.

Depends upon changes in https://github.com/alphagov/govuk_publishing_components/pull/5594

## Why
Using a details element prevents cumulative layout shift, as the display of the contents of the details element (visible or hidden) is handled by the HTML, which is rendered without needing to wait for CSS or JS to load. 

Using native toggle functionality in the browser also feels better than writing a script to do the same, as it is accessible and [keyboard ready out of the box](https://github.com/alphagov/govuk_publishing_components/issues/5580).

The major difference (apart from visual, below) is that the text no longer changes when the visibility of elements is toggled, but my understanding is that this is acceptable from an accessibility standpoint.

Also, in theory this would reduce the JS, but the `gem-toggle` script is used in other places still.

## Visual Changes

Changes to the metadata component, desktop.

Before | After
------ | ------
Screenshot 2026-07-17 at 11 56 12 | Screenshot 2026-07-17 at 11 56 21
Screenshot 2026-07-17 at 11 56 31 | Screenshot 2026-07-17 at 11 56 40

Changes to the metadata component, mobile.

Before | After
------ | ------
Screenshot 2026-07-17 at 12 00 16 | Screenshot 2026-07-17 at 12 00 23
Screenshot 2026-07-17 at 12 00 32 | Screenshot 2026-07-17 at 12 00 40

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Use the details component for toggling element visibility in the metadata component ([PR #5592](https://github.com/alphagov/govuk_publishing_components/pull/5592))
+
 ## 67.0.0
 
 * Add options to the details component ([PR #5594](https://github.com/alphagov/govuk_publishing_components/pull/5594))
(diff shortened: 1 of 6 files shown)


AshGDS (approved):
LGTM 👍

andysellick:
Thanks @AshGDS !

PR #5594, opened 2026-07-17
Add options to the details component

andysellick:
## What
Adds a `small` option and right to left support to the details component.

PR note: I had to write some slightly complex CSS to get the RTL option to play nicely with the `small` option. It's worth double checking that this works as expected, both when the `direction-rtl` class is applied to the component, as well as separately applied to a parent element.

## Why
Needed for this PR: https://github.com/alphagov/govuk_publishing_components/pull/5592 (was originally part of, but decided to split the relevant commits out).

## Visual Changes
New `small` option on the details component:

Screenshot 2026-07-17 at 11 55 36

New right to left option on the details component:

Screenshot 2026-07-17 at 12 55 29

### DIFF
--- a/spec/components/details_spec.rb
+++ b/spec/components/details_spec.rb
@@ -11,6 +11,8 @@ describe "Details", type: :view do
     end
 
     assert_select "details.gem-c-details[data-module='ga4-event-tracker']"
+    assert_select ".gem-c-details--black", false
+    assert_select ".gem-c-details--inverse", false
     assert_select ".gem-c-details--small", false
     assert_select ".govuk-details__summary-text[data-ga4-expandable]", text: "Some title"
     assert_select ".govuk-details__text", text: "This is more info"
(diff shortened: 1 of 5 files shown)


spec/components/details_spec.rb
```diff
@@ -88,4 +88,12 @@ def component_name
     assert_select ".govuk-details__summary[data-ga4-event]", false
     assert_select ".govuk-details__summary-text[data-ga4-expandable]", false
   end
+
+  it "renders small" do
+    render_component(title: "Some title", small: true) do
+      "This is more info"
+    end
+
+    assert_select ".gem-c-details.gem-c-details--small"
```
AshGDS, line 116: Thanks for adding the test 👍  possibly a nitpick, but the other classes aren't being tested, and we aren't testing that this class doesn't exist by default

AshGDS (dismissed):
LGTM 👍

spec/components/details_spec.rb
andysellick: No problem to add. Which other classes are you referring to?

AshGDS: @andysellick Thanks - `gem-c-details--inverse` and `gem-c-details--black`

andysellick: @AshGDS thanks, have added in a separate commit 👍

AshGDS (approved):
LGTM, thanks 👍

PR #5617, opened 2026-07-27
Hide breadcrumbs when printed

### LINKED ISSUE #5601: Breadcrumb components have inconsistent print behaviour
## What

We currently hide contextual breadcrumbs when printing, but standard breadcrumbs are still shown.

Contextual breadcrumbs:
https://github.com/alphagov/govuk_publishing_components/blob/main/app/views/govuk_publishing_components/components/_contextual_breadcrumbs.html.erb#L16

## Why

Is there a reason for this difference, or should we standardise the print behaviour across both breadcrumb components? If so, should they both be visible when printed, or should they both be hidden?

jon-kirwan:
## What

Currently, contextual breadcrumbs are hidden when printed, whereas the breadcrumbs component is not. This change ensures both components are hidden when printing.

## Why

https://github.com/alphagov/govuk_publishing_components/issues/5601


AshGDS (approved):
LGTM 👍

PR #5629, opened 2026-08-03
Remove govuk-govspeak class

andysellick:
## What
Removes the `govuk-govspeak` class from the govspeak component.

## Why
This class doesn't appear to be needed. 

I thought it was the original class for govspeak styling, and when that was turned into the govspeak component we added the `gem-c-govspeak` class to conform to component conventions. However on investigation that class turns out to be `govspeak` (which is still present [throughout the govspeak component styles](https://github.com/alphagov/govuk_publishing_components/blob/main/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_attachment.scss#L11)), so I'm not sure why this class was here. There are no uses of it across alphagov apart from in prototypes, tests, archived repos and places where the govspeak component is being manually reproduced.

## Visual Changes
There's one visual change in the contents list with body component, because the YML file had the `govuk-govspeak` class in it (which didn't style the contents of the component) and I've replaced that with the correct `gem-c-govspeak` class (which does style the contents of the component). So this looks different, but technically better, as it's now being styled correctly.

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_notice.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_notice.scss
@@ -1,7 +1,7 @@
 @import "govuk/components/notification-banner/notification-banner";
 
 .gem-c-notice {
-  .govuk-govspeak {
+  .gem-c-govspeak {
     p:last-child {
       margin-bottom: 0;
     }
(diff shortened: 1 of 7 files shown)


MartinJJones (approved):
Nice work, the changes look good to me, nice improvement in the component guide as well

kevindew:
Just a quick flag that this change caused us a bit of confusion due to the lack of change note - though I imagine it was just an accident so no biggie. It was just a little surprising when we experienced Ruby test failures when it looked otherwise like all 67.0.1 did was introduce a CSS change: https://github.com/alphagov/govuk_publishing_components/pull/5637

andysellick:
@kevindew many apologies, we also had similar, I should have put a changelog entry in.

kevindew:
@andysellick not to worry 😄

PR #2099, opened 2021-05-26
Apply new link styles to our components

owenatgov:
## What
Applies the new link styles from the latest release of govuk frontend.

## Why
Part of work by the govuk frontend community to ensure that we are using the new link styles from govuk frontend.

This PR is the amalgamation of several smaller PRs to get the new link styles working across all our components. You can see these individual PRs including visual changes in the changelog updates in this PR.

This PR additionally includes a minor reversion of focus/hover states introduced in https://github.com/alphagov/govuk_publishing_components/pull/2092. These changes were added too hastily and undermine standard design system behaviour of not triggering hover states when an item is focussed.

PR #5636, opened 2026-08-05
Restore underline to subscription links feed link

andysellick:
## What / why
- removes the no underline class from the 'subscribe to feed' link in the subscription links component
- there are two links in this component, both acting as links, yet this one was not underlined but the other is
- this seems to have been added deliberately five years ago without a clear explanation why: https://github.com/alphagov/govuk_publishing_components/pull/2099/changes/192639a5661948fa3047737f3c8f1219f7913014
- links should look consistent

## Visual Changes

Before | After
----- | ------
Screenshot 2026-08-05 at 08 46 44 | Screenshot 2026-08-05 at 08 48 10
Screenshot 2026-08-05 at 08 48 35 | Screenshot 2026-08-05 at 08 48 47

### DIFF
--- a/app/views/govuk_publishing_components/components/_subscription_links.html.erb
+++ b/app/views/govuk_publishing_components/components/_subscription_links.html.erb
@@ -46,7 +46,7 @@
             lang: feed_link_text_locale,
           ) if sl_helper.feed_link_box_value %>
           <%= link_to(feed_link_text, sl_helper.feed_link,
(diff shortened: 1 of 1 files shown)


luke-booker (approved):
All looks good to me 👍

andysellick:
Brief update - some digging revealed that originally neither of these links had an underline, but the underline showed up on 'Get emails' sometime around 12th Jan 2023. Since both of these things behave similarly and behave like links, it seems appropriate that they have an underline, and be consistent with each other.

PR #5637, opened 2026-08-05
Fix text wrapping in govspeak footnotes

andysellick:
## What / why
- move the text wrap property from being only applied to A tags within footnotes to anything within a footnote
- an example has occurred where a URL is in a footnote but not in link, and the text is failing to wrap on mobile and breaking the page layout
- considered a wider solution to this (put break-word on all of govspeak?) but I'm not sure of the consequences, and this seems like a sensible fix in the circumstances
- example problem found in footnote 27 on this page: https://www.gov.uk/government/publications/disability-foresight-research-future-trends-for-disabled-people/disability-foresight-research-uk-final-report#executive-summary

## Visual Changes
Before | After
----- | ------
Screenshot 2026-08-05 at 15 23 22 | Screenshot 2026-08-05 at 15 23 48

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
+* Fix text wrapping in govspeak footnotes ([PR #5637](https://github.com/alphagov/govuk_publishing_components/pull/5637))
+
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):
Nice work, the changes look good to me, works nicely for both links as plain text and as an anchor element

PR #5644, opened 2026-08-11
Update form date input component

davidtrussler:
## What
Makes updates to the "Form date input" component: 
- Add new option "With legend level and size"
- Add new option "With data attributes"
- Add new option "With error id but no message"
  - similar to the option in the [Select](https://components.publishing.service.gov.uk/component-guide/select#with_error_id_but_no_message) component
- Update how "aria-describedby" attributes are applied: 
  - By default there is no such attribute
  - If hint text is present `input` elements take the `hint-id` as the value for `aria-describedby`
    - similar to the option in the [Select](https://components.publishing.service.gov.uk/component-guide/select#with_hint) component
  - If error messages are present `input` elements take the `error-id` as the value for `aria-describedby`
    - similar to the option in the [Select](https://components.publishing.service.gov.uk/component-guide/select#with_error) component
  - If both hint and error text are present `input` elements take the `error-id` as the value for `aria-describedby

## Why

Jira card: https://gov-uk.atlassian.net/browse/MAIN-1311

This is part of the bigger task to introduce a new "Datetime fields" component. The work to do this is partially done and exists in [another PR](https://github.com/alphagov/govuk_publishing_components/pull/5492). The "Datetime fields" component uses the "Form date input" component to render date input fields but requires some changes for it to function correctly so I am splitting out the required changes to that component in this PR to avoid that becoming overly complicated. These include:
- making the legend level and text size configurable
- allowing data-attributes to be passed through to the component
- allowing the component to display the input fields styled as errors without also displaying error messages and for these to have an aria-describedby attribute that matches the error id
- updating how the aria-describedby attributes are applied, which is incorrect in the component as it currently exists. 

## Visual Changes

| Option | Screenshot | Markup (partial) generated by component |
|--------|--------|--------|
|With legend level and size|Screenshot 2026-08-11 at 14 36 29|\<legend class="govuk-fieldset__legend govuk-fieldset__legend--l"\>\<h2 class="govuk-fieldset__heading">What is your date of birth?\</h2>\</legend>|
|With data attributes|Screenshot 2026-08-11 at 14 38 30|\<div class="govuk-form-group" data-module="some-module" data-other-attribute="something-else"\> ... \</div\>|
|With error id but no message|Screenshot 2026-08-11 at 14 40 17|\<div class="govuk-form-group--error"\> ... \<input aria-describedby="error_id"\> ... \<input aria-describedby="error_id"\> ... \<input aria-describedby="error_id"\> ... \</div\>|
|If hint text is present `input` elements take the `hint-id` as the value for `aria-describedby`|Screenshot 2026-08-11 at 14 42 48|\<div id="hint-0342f3e9"\>For example, 31 3 1980\</div\> ... \<input aria-describedby="hint-0342f3e9"\> ... \<input aria-describedby="hint-0342f3e9"\> ... \<input aria-describedby="hint-0342f3e9"\>|
|If error messages are present `input` elements take the `error-id` as the value for `aria-describedby`|Screenshot 2026-08-11 at 14 44 45|\<div id="hint-6fded4c1"\>For example, 31 3 1980\</div\> ... \<p id="error-cd5fedb6"\>Error message goes here\</p\> ... \<input aria-describedby="error-cd5fedb6"\> ... \<input aria-describedby="error-cd5fedb6"\> ... \<input aria-describedby="error-cd5fedb6"\>|

### DIFF
--- a/app/views/govuk_publishing_components/components/_date_input.html.erb
+++ b/app/views/govuk_publishing_components/components/_date_input.html.erb
@@ -8,30 +8,34 @@
   ]
   autocomplete_date_of_birth ||= nil
   legend_text ||= nil
+  legend_heading_level ||= nil
+  legend_size ||= nil
   hint ||= nil
   error_message ||= nil
   error_items ||= []
-  describedby ||= nil
-  has_error ||= error_message || error_items.any?
+  error_id ||= nil
+  has_error ||= error_message || error_items.any? || error_id
 
   css_classes = %w(gem-c-date-input govuk-date-input)
   form_group_css_classes = %w(govuk-form-group)
   form_group_css_classes << "govuk-form-group--error" if has_error
 
   hint_id = "hint-#{SecureRandom.hex(4)}"
-  error_id = "error-#{SecureRandom.hex(4)}"
+  error_id = error_id || "error-#{SecureRandom.hex(4)}"
 
-  aria_described_by ||= nil
-  if hint || has_error || describedby
+  aria_described_by = nil
+  if hint || has_error
     aria_described_by = []
-    aria_described_by << hint_id if hint
+    aria_described_by << hint_id if hint && !has_error
     aria_described_by << error_id if has_error
-    aria_described_by << describedby if describedby
     aria_described_by = aria_described_by.join(" ")
   end
+
+  component_helper = GovukPublishingComponents::Presenters::ComponentWrapperHelper.new(local_assigns)
+  component_helper.add_class(form_group_css_classes.join(" "))
 %>
 
-<%= content_tag :div, class: form_group_css_classes do %>
+<%= tag.div(**component_helper.all_attributes) do %>
   <% fieldset_content = capture do %>
     <% if hint %>
       <%= render "govuk_publishing_components/components/hint", {
@@ -41,7 +45,7 @@
       } %>
     <% end %>
 
-    <% if has_error %>
+    <% if error_message || error_items.any? %>
       <%= render "govuk_publishing_components/components/error_message", {
         id:  error_id,
         text: error_message,
@@ -62,6 +66,7 @@
             value: item[:value],
             width: item[:width],
             id: item[:id],
+            describedby: aria_described_by,
             type: "number",
             data: item[:data],
             autocomplete: autocomplete_date_of_birth ? "bday-#{item[:name]}" : nil,
@@ -73,8 +78,9 @@
 
   <% if legend_text %>
     <%= render "govuk_publishing_components/components/fieldset", {
-      describedby: aria_described_by,
       legend_text: legend_text,
+      heading_level: legend_heading_level,
+      heading_size: legend_size,
       text: fieldset_content,
       role: "group",
     } %>
--- a/app/views/govuk_publishing_components/components/docs/date_input.yml
+++ b/app/views/govuk_publishing_components/components/docs/date_input.yml
@@ -1,5 +1,6 @@
 name: Form date input
 description: Use the date input component to help users enter a memorable date or one they can easily look up.
+uses_component_wrapper_helper: true
 accessibility_criteria: |
   Inputs in the component must:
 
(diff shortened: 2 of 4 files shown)


app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -36,6 +36,11 @@ examples:
   with_legend:
     data:
       legend_text: "What is your date of birth?"
+  with_legend_level_and_size:
+    data:
```
andysellick, line 42: Might be worth a description here to summarise the options - presumably `xl l m s`?

app/views/govuk_publishing_components/components/docs/date_input.yml
```diff
@@ -36,6 +36,11 @@ examples:
...
+      legend_text: "What is your date of birth?"
+      legend_level: 2
```
andysellick: Nitpick: maybe this option would be clearer if it was called `legend_heading_level`?

spec/components/date_input_spec.rb
```diff
@@ -42,8 +42,17 @@ def component_name
     )
 
     assert_select ".govuk-fieldset[role=group] .govuk-fieldset__legend", text: "What is your date of birth?"
-    assert_select ".govuk-fieldset[role=group] .govuk-date-input", 1
-    assert_select ".govuk-fieldset[role=group] .govuk-date-input__item", 3
+  end
+
+  it "renders with custom legend and fieldset level and size" do
```
andysellick: Might be worth adding a test if there isn't one already to check that the expected legend heading level and size are used by default?

andysellick: Nitpick: test name is a little confusing - should it just be `renders with custom legend heading level and size`?

app/views/govuk_publishing_components/components/_date_input.html.erb
```diff
@@ -31,9 +31,11 @@
     aria_described_by << describedby if describedby
     aria_described_by = aria_described_by.join(" ")
   end
+
+  data_attributes ||= {}
```
andysellick: Have you considered adding the [component wrapper helper](https://github.com/alphagov/govuk_publishing_components/blob/main/docs/component-wrapper-helper.md) to this component, which would add this functionality automatically and not require any additional tests or documentation? 

I just tried it on `main` and it seems to work correctly, I added the following:

```
component_helper.add_class(form_group_css_classes.join(" "))
%>
<%= tag.div(**component_helper.all_attributes) do %>
```

spec/components/date_input_spec.rb
```diff
@@ -61,7 +61,12 @@ def component_name
       hint: "For example, 31 3 1980",
     )
 
+    hint_id = css_select(".govuk-hint").attr("id")
```
andysellick: I think you might be able to make use of this in these tests?

```
allow(SecureRandom).to receive(:hex).and_return("1234")
```

spec/components/date_input_spec.rb
davidtrussler: I *think* the preceding test covers this: there is no heading at all, just a <legend> element and that is not modified with any `govuk-fieldset__legend--l` type of class. 

It isn't perfect I guess since it doesn't test that either of those things are not present but I can't see how to easily do that without testing that *every* heading level and size variation are *not* present, which seems a bit overkill. Or I may be missing something.

app/views/govuk_publishing_components/components/_date_input.html.erb
davidtrussler: I have added this though I'm not sure if it should incorporate more than just the css classes.

spec/components/date_input_spec.rb
davidtrussler: Ah nice, that does the job more elegantly than what I had. That does exist elsewhere incidentally but I guess it isn't worth changing that right now.

spec/components/date_input_spec.rb
andysellick: Ah yes I see, it would be a bit onerous to check that every heading level wasn't present. Fair enough.

app/views/govuk_publishing_components/components/_date_input.html.erb
andysellick: Nice. Two small things - you shouldn't need line 34 anymore (`data_attributes ||= {}`) as the component wrapper helper manages this itself. And you should add `uses_component_wrapper_helper: true` to the YML file so that the component wrapper helper options are included automatically.

davidtrussler: Cheers, I have attended to that. 

Additionally I updated the value for `error_id` elsewhere to prevent accessibility warnings triggered by multiple elements on the page with the same `id` value.

andysellick: Nice 👍

andysellick (approved):

PR #5646, opened 2026-08-11
Improve filter Javascript

andysellick:
## What
Improve the generic filter JavaScript. This is currently used in the component guide main page and on the component guide applications page.

- adds a result count
- announces the result count for screen reader users
- prevents itself from initialising more than once on a page
- allow the script to be used anywhere on a page, not tied to the element containing the things that need filtering

## Why
These improvements will make this a more viable option for public use.

## Visual Changes
Example from its use on the component guide main page:

Screenshot 2026-08-11 at 17 02 01

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@
 
 ## Unreleased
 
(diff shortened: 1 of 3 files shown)


MartinJJones (approved):
Nice work, the change looks good to me 👍

PR #5655, opened 2026-08-14
Fix step by step navigation JS security warning

andysellick:
## What
Slightly refactor the JS for the step by step navigation component to remove a potential security risk.

## Why
See https://github.com/alphagov/govuk_publishing_components/security/code-scanning/47

However... I'm starting to wonder if this security warning is a false positive. The thing it seems to be worried about is `titleText` being inserted into a chunk of markup, which we get around here by creating the chunk of markup then inserting `titleText` afterwards using `textContent`, which sanitises it. But `titleText` is itself initially created by using `textContent`, so surely it's already been sanitised? Regardless, if this change silences a security warning, that still seems worth doing.

## Visual Changes
None.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/step-by-step-nav.js
@@ -168,7 +168,7 @@
       var thisel = this.$module.steps[i]
       var title = thisel.querySelectorAll('.js-step-title')[0]
       var contentId = thisel.querySelectorAll('.js-panel')[0].getAttribute('id')
-      var titleText = title.textContent || title.innerText // IE8 fallback
+      var titleText = title.textContent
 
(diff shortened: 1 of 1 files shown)


MartinJJones (commented):
As you mention in the PR description, it does seem like this could be false positive which textContent should avoid.

I think it could be worth removing the innerText IE8 fallback and switching to template literals as you have, but then removing `var text` as it is only referenced once and updating `<span class="gem-c-step-nav__title-text js-step-title-text"></span>` to `<span class="gem-c-step-nav__title-text js-step-title-text">${title.textContent}</span>` and removing the need for the querySelector, we can then merge this and see if this resolves the issue and if we need to do anything further?

MartinJJones (approved):
Nice work, the fix looks good to me, thanks for testing out my other (unsuccessful) suggestions as well

PR #5658, opened 2026-08-14
Remove grid helper and update Cards to native CSS Grid

luke-booker:
## What
- The cards component was using grid-helper.scss to manage multiple card columns. It used the fractions() function, however, browsers without CSS grid support still displayed a single column. 
- The same can be achieved simply with using standard CSS Grid syntax.
- The graceful fallback will still be a single card layout. 

## Why

With modern browsers natively supporting standard CSS grid and legacy browsers automatically falling back to a single block layout, the use of grid-helper.scss in cards.scss is no longer necessary. This change helps remove technical debt while bringing the component in line with current design system standards.

Jira card: https://gov-uk.atlassian.net/browse/NAV-19065

## Visual Changes
There should be no visual changes. This is an update to the approach in the codebase only. 

Live three column cards can be seen in use on the the service manual page https://www.gov.uk/service-manual

Screenshot 2026-08-20 at 08 36 08

## Testing
- Verified the 1 column, 2 column, 3 column, and Auto layout examples render properly aligned grids without issues
- Spot checked the cards component across modern browsers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,7 @@
 
 * Review and fix image card spacing ([PR #5661](https://github.com/alphagov/govuk_publishing_components/pull/5661))  
 * Upgrade Stylelint to v17 ([PR #5654](https://github.com/alphagov/govuk_publishing_components/pull/5654))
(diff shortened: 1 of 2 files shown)


MartinJJones (approved):

PR #5659, opened 2026-08-14
remove percyci from deps

patrickpatrickpatrick:
## Why

Jira card: https://gov-uk.atlassian.net/browse/[CARD-ID]

## Visual Changes

| Before | After |
|--------|--------|
| (Insert screenshot here) | (Insert screenshot here) |


.github/workflows/visual-regression-tests.yml
```diff
@@ -50,6 +50,7 @@ jobs:
           cache: yarn
       - run: yarn install --frozen-lockfile
       - run: bundle exec rake app:dartsass:build
+      - run: yarn add @percy/cli
```
AgaDufrat, line 53: Sorry, if this PR is still a work in progress. My immediate concern with installing the latest version is that we're opening ourselves to the supply chain vulnerabilities. 
Pinning it to a specific version `yarn add @percy/cli@1.31.13` would address this but we couldn't benefit from the automatic updates by Dependabot. 

I couldn't find any official GitHub action we could use (and pin it to commit sha) so the current approach seems safer to me.

PR #5660, opened 2026-08-17
Option select vuln

andysellick:
## What
Attempting to remove some potential vulnerabilities highlighted by the code scanning on this repo.

## Why
Should address:

- https://github.com/alphagov/govuk_publishing_components/security/code-scanning/44
- https://github.com/alphagov/govuk_publishing_components/security/code-scanning/45

## Visual Changes
None, hopefully.

### DIFF
--- a/app/assets/javascripts/govuk_publishing_components/components/option-select.js
+++ b/app/assets/javascripts/govuk_publishing_components/components/option-select.js
@@ -9,7 +9,7 @@
     this.$optionsContainer = this.$optionSelect.querySelector('.js-options-container')
     this.$optionList = this.$optionsContainer.querySelector('.js-auto-height-inner')
     this.$allCheckboxes = this.$optionsContainer.querySelectorAll('.govuk-checkboxes__item')
-    this.hasFilter = this.$optionSelect.getAttribute('data-filter-element') || ''
+    this.filterAttributes = this.$optionSelect.getAttribute('data-filter-attributes')
 
     this.checkedCheckboxes = []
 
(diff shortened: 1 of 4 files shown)


app/assets/javascripts/govuk_publishing_components/components/option-select.js
```diff
@@ -73,6 +67,25 @@
...
   }
 
+  OptionSelect.prototype.createFilter = function () {
+    const filter = document.createElement('div')
+    filter.classList.add('gem-c-option-select__filter')
+    filter.innerHTML = `
+      <label class="govuk-label govuk-visually-hidden"></label>
+      <input type="text" class="gem-c-option-select__filter-input govuk-input"></input>
+    `
+    const label = filter.querySelector('label')
```
MartinJJones: non-blocking: rather than using the querySelector to get the label element, I'm wondering if it would be more efficient to create the label element and add the class to it here 🤔 

If we used the same approach for the input element, we could update the code similar to that below:

```js
  OptionSelect.prototype.createFilter = function () {
    const filter = document.createElement('div')
    filter.classList.add('gem-c-option-select__filter')

    const label = document.createElement('label')
    label.classList.add('govuk-label', 'govuk-visually-hidden')
    ...

    const input = document.createElement('input')
    input.classList.add('gem-c-option-select__filter-input', 'govuk-input')
    input.type = 'text'
    ...

    filter.append(label, input)
    return filter
}
```

app/assets/javascripts/govuk_publishing_components/components/option-select.js
```diff
@@ -73,6 +67,25 @@
...
+    label.setAttribute('for', this.filterAttributes.id)
+    label.textContent = this.filterAttributes.title
+
+    const input = filter.querySelector('input')
+    input.name = 'option-select-filter'
+    input.setAttribute('aria-describedby', this.filterAttributes.describedby)
```
MartinJJones, line 79: From testing locally it looks like we need to add `input.id = this.filterAttributes.id` to hook it up to the `for` attribute on the label

MartinJJones (commented):
Nice work on this, the changes look good to me 👍 

I spotted one small thing with an missing id on the input element, and have one non-blocking suggestion, happy to catch up on it as well

app/assets/javascripts/govuk_publishing_components/components/option-select.js
andysellick: Good suggestion, have amended. Feels cleaner.

app/assets/javascripts/govuk_publishing_components/components/option-select.js
andysellick: Disappointed that I overlooked that! Thanks for pointing out, fixed.

MartinJJones (approved):

Not sure if this one needs a change log entry? seems OK as it keeps the functionality the same

PR #5665, opened 2026-08-18
Adjust cards content elements widths

andysellick:
## What
- change the cards component heading and description elements to use a fixed pixel max-width instead of a percentage
- percentage works correctly when the component is used in a full width column, but when it is constrained to a 2/3rds column (as on the homepage) this means these text elements were reduced in width further, necessitating a component override in `frontend` (which breaks component isolation)
- instead, use a pixel value for the max-width, which does the same job but means this problem doesn't occur, is equivalent to the percentage at full width and means the width isn't unnecessarily reduced when the component is in a 2/3rds column
- switching to a pixel value doesn't mean the element can overlap its container, it's still constrained by the normal behaviour of a block level element
- this means we can remove the override in `frontend`
- also add text-wrap: pretty, to prevent single words from dangling onto a second line by themselves (as was occurring in some of the guide examples), note that this will cause a small visual diff

## Why
This change will allow us to remove this [component style override in frontend](https://github.com/alphagov/frontend/blob/main/app/assets/stylesheets/views/_homepage.scss#L294-L300).

## Visual Changes
The width change doesn't make any visual change, but the text-wrap change ensures any second lines won't be a single word. Before:

Screenshot 2026-08-18 at 08 23 33

After:

Screenshot 2026-08-18 at 08 23 13

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
@@ -158,10 +158,11 @@
 
 .gem-c-cards__sub-heading {
   margin: 0 govuk-spacing(6) govuk-spacing(2) 0;
+  text-wrap: pretty;
 
   // Ensure card content width is constrained to two thirds on desktop
   @include govuk-media-query($from: "desktop") {
-    max-width: 66.6667%;
+    max-width: 640px;
   }
 }
 
@@ -208,10 +209,11 @@
 
 .gem-c-cards__description {
   margin: 0 govuk-spacing(6) 0 0;
+  text-wrap: pretty;
 
   // Ensure card content width is constrained to two thirds on desktop
   @include govuk-media-query($from: "desktop") {
-    max-width: 66.6667%;
+    max-width: 640px;
   }
 }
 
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/_cards.scss
```diff
@@ -158,10 +158,11 @@
 
 .gem-c-cards__sub-heading {
   margin: 0 govuk-spacing(6) govuk-spacing(2) 0;
+  text-wrap: pretty;
```
luke-booker, line 161: Might be worth mentioning [text-wrap: pretty is not yet available on all browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/text-wrap#:~:text=17.4-,pretty,-117)  

Looks like Firefox gracefully handles this though and will just text-wrap: wrap anyway. Couldn't find any immediate plans for devs to implement on Firefox so there will be that small difference going forwards.

andysellick: Yes, it's not a comprehensive improvement, but it does impact a lot of browsers, so I thought worth including.

luke-booker (approved):
Nice work, LGTM!

PR #5669, opened 2026-08-19
Allow exemptions from `data-ga4-form-use-*-count` [WHIT-3832]

davidsauntson:
## What
Creates a new `data` attribute for the `Ga4FormTracker` module that allows consumers to exempt `select` and `input` fields from the `data-ga4-form-use-select-count` and `data-ga4-form-use-text-count` attributes added to a form.

This means we can add `data-ga4-force-report-value-in-form-tracker` to a field and get report the values submitted for those fields to GA, whilst still only submitting text and/or selected option counts for other fields in the same form.

Also make a small refactor to how `PIIRemover` is used in the form tracker module, since we only need one per form and not one per field.  I can pull this out into a separate PR is people would prefer.

## Why

In Whitehall, we need to maintain the text character counts and select counts for most fields, but we do need to know the actual values submitted by other fields.  This is due to access limiting work, where we don't want to report the contents of an edition's body, but do want some information about how it is access limited from other fields.

Addresses some of the points in https://gov-uk.atlassian.net/browse/WHIT-3832

## Visual Changes
No visual changes.


davidtrussler (approved):
All looks good to me - nice one! Also tested it locally. 👍

PR #5672, opened 2026-08-21
Prevent metadata component rendering when no data

andysellick:
## What
- stops anything from being rendered when nothing is passed to the metadata component
- previously would render an empty wrapping element, crucially with a margin bottom, which can disrupt layouts
- add explicit variables that could be passed to the component but not defined i.e. history and first_published

## Why
We're using it in a few places in `frontend` where it can have a variety of things passed to it, and it would be easier if it just didn't render if nothing was passed to it, than having to wrap it in multiple `if` statements.

## Visual Changes
Shouldn't be any.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,9 +7,13 @@
   useful summary for people upgrading their application, not a replication
   of the commit log.
 
+## Unreleased
+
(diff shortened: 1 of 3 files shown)


AshGDS (approved):
Nice, LGTM!

PR #5685, opened 2026-09-03
Update chart colours to align with the brand update

jon-kirwan:
## What

Update chart colours to align with the brand update, based on the new colours added in https://gov-uk.atlassian.net/browse/NAV-18942?focusedCommentId=204506.

## Why

https://gov-uk.atlassian.net/browse/NAV-18938

## Visual Changes

Screenshot 2026-09-03 at 11 29 57

Screenshot 2026-09-03 at 11 31 18

Screenshot 2026-09-03 at 11 32 01

Screenshot 2026-09-03 at 11 32 43

### DIFF
--- a/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
+++ b/app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
@@ -30,18 +30,18 @@
   // For charts, the palette should be used instead of the GOV.UK colour palette - https://design-system.service.gov.uk/styles/colour/
   // https://gss.civilservice.gov.uk/policy-store/data-visualisation-colours-in-charts/#section-5
 
-  $gss-colour-dark-blue: #12436d;
-  $gss-colour-turquoise: #1bbbaf;
-  $gss-colour-dark-pink: #801650;
-  $gss-colour-orange: #f46a25;
-  $gss-colour-dark-grey: #3d3d3d;
-  $gss-colour-plum: #a285d1;
+  $chart-blue-shade-50: govuk-colour("blue", $variant: "shade-50");
+  $chart-teal-tint-25: govuk-colour("teal", $variant: "tint-25");
+  $chart-magenta-shade-25: govuk-colour("magenta", $variant: "shade-25");
+  $chart-primary-orange: govuk-colour("orange");
+  $chart-green-shade-50: govuk-colour("green", $variant: "shade-50");
+  $chart-accent-purple: #ba4aff; // There is currently no equivalent colour in the GOV.UK palette, but this can be updated in future if one is added.
 
   // CHART COLOUR SCHEME
 
   $chart-border: govuk-colour("white"); // Chart border colour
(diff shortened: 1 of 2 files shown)


app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
```diff
@@ -24,24 +24,23 @@
...
   // For charts, the palette should be used instead of the GOV.UK colour palette - https://design-system.service.gov.uk/styles/colour/
   // https://gss.civilservice.gov.uk/policy-store/data-visualisation-colours-in-charts/#section-5
 
-  $gss-colour-dark-blue: #12436d;
-  $gss-colour-turquoise: #1bbbaf;
-  $gss-colour-dark-pink: #801650;
-  $gss-colour-orange: #f46a25;
-  $gss-colour-dark-grey: #3d3d3d;
-  $gss-colour-plum: #a285d1;
+  $chart-blue-shade-50: #0f385c;
```
MartinJJones: Just wanted to check, now that we are using v6, can/should these use the new functional colours approach?

For example, `$chart-blue-shade-50: #0f385c;` becomes `$chart-blue: govuk-colour("blue", $variant: "shade-50");`, I updated the variable name slightly as the shade is already mentioned when setting the variant.

Hopefully it will be possible to update all of them using this page as a reference - https://design-system.service.gov.uk/styles/colour/

MartinJJones (commented):
The changes look good to me, I have just one small suggestion on switching to using the functional colours approach now we are on v6

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
jon-kirwan: Ah yes, good spot. I've added a second temporary commit to make those changes. The only one I can't change is `$chart-accent-purple`, as that doesn't seem to have an equivalent Design System colour.

If that looks okay, I'll squash this commit together with the other colour updates.

MartinJJones: Thanks, that looks good to me, the purple colour chosen matches the [intended design](https://gov-uk.atlassian.net/browse/NAV-18938) 👍 

**non-blocking:** I wonder if we should add a comment for `$chart-accent-purple` for anyone looking at this in the future, just to mention that the colour is not available in the design system, and it was added to ensure sufficient contrast?

jon-kirwan: Yep, that'll be helpful. I've added that now and squashed everything together.

MartinJJones (approved):

app/assets/stylesheets/govuk_publishing_components/components/govspeak/_charts.scss
MartinJJones: Nice, thanks for adding this in, the changes look good to me 😃

PR #5686, opened 2026-09-03
Make 'No data' more logical on bar charts

AshGDS:
## What / Why

- A user reported that we recommend adding "No data" to charts when they have some missing data for a given key. However "No data" (or anything that isn't a number) renders as a full width bar on our govspeak bar charts.
- This is an attempt to fix that by showing them with a 2px coloured border instead. The coloured border matches the key's colour.  
  - The only downside to this is the 2px border can end up being the same width as a small value on a chart (e.g. if a chart had a bar that was 500 and a bar that was 1,  500 would be full width, and 1 would appear as 2px wide, similar to "No data".) Though I'd argue that you would still able to understand the difference between 1 and "No data" in this case.
- I prevented "No data" from rendering on [stacked bar charts](https://components-gem-pr-5686.herokuapp.com/component-guide/govspeak/chart_stacked/preview) too. Rendering "No data" here would make a chart not make sense. For example, a bar could be wider than another due to it having a 'No data' on it, despite it having a lower total than other bars, which would mean the width of the stacked bar no longer conveyed its size relative to other bars.  (Very crude unstyled example in the screenshot)
  - image

  - Maybe we shouldn't promote the ability to add "No data" to a stacked chart anyway as it breaks the 3:1 adjacent colour contrast ratio. Each bar's colour is only meant to sit next to the specific ones next to it in our code, but a "No data" value here means a colour is skipped over, so bars end up next to colours they are not meant to be. 
- This PR also adjusts the spacing of the chart slightly, as requested by our designer.
- The design has been approved by our designer.
- Component link: https://components-gem-pr-5686.herokuapp.com/component-guide/govspeak
- Jira card: https://gov-uk.atlassian.net/browse/PNP-10137

## Visual Changes

| Before | After |
|--------|--------|
| image | image |
| image | image |
| image | image |
|image | image |

### DIFF
--- a/app/views/govuk_publishing_components/components/docs/govspeak.yml
+++ b/app/views/govuk_publishing_components/components/docs/govspeak.yml
@@ -634,7 +634,7 @@ examples:
               <td>row 3</td><td>2</td>
             </tr>
             <tr>
-              <td>row 4</td><td>48</td>
+              <td>row 4</td><td>No data</td>
             </tr>
           </tbody>
         </table>
@@ -686,9 +686,9 @@ examples:
             <tr><th>Some Data</th><th>YES</th><th>NO</th><th>MAYBE</th></tr>
           </thead>
           <tbody>
-            <tr><th>Testing One</th><td>5</td><td>6</td><td>11</td></tr>
-            <tr><th>Testing Two</th><td>6</td><td>2</td><td>8</td></tr>
-            <tr><th>Testing Three</th><td>3</td><td>9</td><td>12</td></tr>
+            <tr><th>Testing One</th><td>No data</td><td>7</td><td>6</td></tr>
+            <tr><th>Testing Two</th><td>6</td><td>2</td><td>No data</td></tr>
+            <tr><th>Testing Three</th><td>3</td><td>No data</td><td>9</td></tr>
           </tbody>
         </table>
   chart_stacked:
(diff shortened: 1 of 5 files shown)


app/views/govuk_publishing_components/components/docs/govspeak.yml
```diff
@@ -634,7 +634,32 @@ examples:
...
+              <td>Hulk</td><td>6000</td>
+            </tr>
+            <tr>
+              <td>Thor</td><td>5000</td>
+            </tr>
+            <tr>
+              <td>Spiderman</td><td>-2000</td>
+            </tr>
+            <tr>
+              <td>Iron Man</td><td>No data</td>
```
andysellick, line 659: Just want to check that if there's no data in the table then the value is actually `No data`. Testing this by removing that text from the table cell results in nothing being displayed. Is that expected behaviour?

andysellick (commented):
Changes look sensible and good, happy to approve but one small question.

Also this feels like it might need a changelog entry?

app/views/govuk_publishing_components/components/docs/govspeak.yml
AshGDS: @andysellick I think that might be OK, as the guidance is to not add any empty cells: https://guidance.publishing.service.gov.uk/formatting-content/text-formatting/tables/#make-cells-readable

andysellick: Great, thanks. And having read that, does it work on both `No data` and `Not applicable`, as well as case variants? (e.g. `no data`)

AshGDS: Yeah it should do, it will render whatever the editor puts in 👍

andysellick (dismissed):

andysellick (approved):

PR #5688, opened 2026-09-03
BREAKING: Remove grid-helper mixin

luke-booker:
## What

- This PR deletes grid-helper.scss from the gem and removes all referencing import statements.
- It removes the grid-helper.scss export entirely. Any downstream applications using this mixin will require updates to use standard CSS Grid.
- Replace legacy columns(...) and fractions(...) functions with standard CSS Grid declarations.

## Why

- Technical debt: grid-helper.scss was originally introduced to generate manual grid coordinates for legacy Grade X browsers that lacked standard CSS Grid support.
- Modern Grade C and above browsers fully support standard CSS Grid, rendering custom calculations redundant.
- Grade X browsers can fall back to an acceptable single block column layout.
- Aligns styling across repositories around native CSS standards rather than maintenance heavy mixins.

Jira card: https://gov-uk.atlassian.net/browse/NAV-19070

## Visual changes

None, grid-helper.scss was not being used in publishing components gem


MartinJJones (dismissed):

MartinJJones (approved):

MartinJJones (approved):

PR #5695, opened 2026-09-07
Upgrade jasmine-browser-runner dependencies

MartinJJones:
## What
Upgrade jasmine-browser-runner dependencies

## Why
This change ensures we are using the latest versions that include security fixes, resolving an issue with `brace-expansion`

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -837,10 +837,10 @@ brace-expansion@^2.0.2:
   dependencies:
     balanced-match "^1.0.0"
 
-brace-expansion@^5.0.2:
-  version "5.0.4"
-  resolved "https://registry.yarnpkg.com/brace-expansion/-/brace-expansion-5.0.4.tgz#614daaecd0a688f660bbbc909a8748c3d80d4336"
-  integrity sha512-h+DEnpVvxmfVefa4jFbCf5HdH5YMDXRsmKflpf1pILZWRFlTbJpxeU55nJl4Smt5HQaGzg1o6RHFPJaOqnmBDg==
+brace-expansion@^5.0.8:
+  version "5.0.9"
+  resolved "https://registry.yarnpkg.com/brace-expansion/-/brace-expansion-5.0.9.tgz#7c72438809b5fa5babf54199a1f1c281a6984fcf"
+  integrity sha512-ScQ4IuvIEF1TMlP7Zt+vjJ//9zlPb2SDcxWxM3bk8s6t6GGdJ7KO1dCcTidOPJKePW30LE/2cT7wCyPho9/Wxg==
   dependencies:
     balanced-match "^4.0.2"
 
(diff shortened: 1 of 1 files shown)


yarn.lock
```diff
@@ -2410,7 +2410,7 @@ isarray@^2.0.5:
 isarray@~1.0.0:
   version "1.0.0"
   resolved "https://registry.yarnpkg.com/isarray/-/isarray-1.0.0.tgz#bb935d48582cba168c06834957a54a3e07124f11"
-  integrity sha1-u5NdSFgsuhaMBoNJV6VKPgcSTxE=
+  integrity sha512-VLghIWNM6ELQzo7zwmcg0NmTVyWKYjvIeM83yjp0wRDTmUnrM678fQbcKBo6n2CJEF0szoG//ytg+TKla89ALQ==
```
andysellick, line 2413: How has this one changed the `integrity` but the version number remains the same?

MartinJJones: Good spot, I've checked the entry for the isarray package by visiting the resolved link - https://registry.yarnpkg.com/isarray - 

```
 "1.0.0": {
...
      "dist": {
        "shasum": "bb935d48582cba168c06834957a54a3e07124f11",
        "tarball": "https://registry.npmjs.org/isarray/-/isarray-1.0.0.tgz",
        "integrity": "**sha512-VLghIWNM6ELQzo7zwmcg0NmTVyWKYjvIeM83yjp0wRDTmUnrM678fQbcKBo6n2CJEF0szoG//ytg+TKla89ALQ==**",
        "signatures": [
          {
            "keyid": "SHA256:jl3bwswu80PjjokCgh0o2w5c2U4LhQAE57gj9cz1kzA",
            "sig": "MEYCIQDAUFha3cw0zrMmy/gRN+em7pQTK/Th9NvZu629sSQJEQIhAOA18G6lEgwD6Vu7/NsqFDjSDBx0WIS23RbPOBWbtgx+"
          }
        ]
      },
...
```

The `integrity` value in the registry matches the new `sha512-` value, this is also [used in several of our repos](https://github.com/search?q=org%3Aalphagov+VLghIWNM6ELQzo7zwmcg0NmTVyWKYjvIeM83yjp0wRDTmUnrM678fQbcKBo6n2CJEF0szoG&type=code), although I'm not 100% sure why this has changed now.

My assumption is that when running `yarn upgrade jasmine-browser-runner`, yarn checks all dependencies and can see that `isarray` is required (`yarn why isarray`)

```
=> Found "readable-stream#isarray@1.0.0"
info This module exists because "jasmine-browser-runner#selenium-webdriver#jszip#readable-stream" depends on it.
```

There does appear to be some history behind migrating from `sha1-` to `sha-512` that I would need to read further into:

- https://blog.npmjs.org/post/172999548390/new-pgp-machinery
- https://classic.yarnpkg.com/lang/en/docs/yarnrc/#toc-unsafe-disable-integrity-migration
- [migrate from npm to yarn - govuk_publishing_components gem PR](https://github.com/alphagov/govuk_publishing_components/commit/7c9580f9d2cce108f7c8e9f9704cb90fb4519d6d) - lock file shows 119 dependencies using sha1, this number has also reduced over time

Also worth noting that applications using Yarn Modern (yarn v2+) have a different approach in the lock file, isarray example in frontend - https://github.com/alphagov/frontend/blob/main/yarn.lock#L3254-L3259. I imagine this is unlikely to be an issue in the future if we migrate to Yarn Modern, related GitHub issue - https://github.com/alphagov/govuk_publishing_components/issues/5244

I'm also happy to revert the change for isarray if preferred here?

andysellick: That was thorough, thanks. I think you answered my question though.

andysellick (approved):

PR #457, opened 2018-07-31
Add prioritise_taxon_breadcrumbs flag to contextual breadcrumbs

oscarwyatt:
Contextual breadcrumbs will show taxon based breadcrumbs if prioritise_taxon_breadcrumbs is true (defaults to false if not passed) 

Remember to update the CHANGELOG if your change needs to be listed in the next version of the gem.

---

Component guide for this PR:
https://govuk-publishing-compon-pr-457.herokuapp.com/component-guide/
https://govuk-publishing-compon-pr-457.herokuapp.com/component-guide/contextual_breadcrumbs

Trello card
https://trello.com/c/yZfYuiJp/84-use-taxons-breadcrumb-in-b-sample

PR #5708, opened 2026-09-11
Remove prioritise_taxon_breadcrumbs

### LINKED ISSUE #5698: prioritise_taxon_breadcrumbs isn't used currently anywhere in alphagov
This option on contextual breadcrumbs alters the order of priority on which breadcrumb is chosen. But we don't actually pass it anywhere, so it might be a bit of zombie code.

AshGDS:
## What/ Why
- Remove `prioritise_taxon_breadcrumbs` from the gem as it isn't used anywhere anywhere across `alphagov`
- It was added in 2018 https://github.com/alphagov/govuk_publishing_components/pull/457 with the description "Rendering taxonomy breadcrumbs because the page is tagged to live taxons and we want to prioritise them over all other breadcrumbs"
- I locally checked out January 31st 2019 versions of `frontend`, `publishing-api`, `finder-frontend`, `collections`, `smart-answers`, `feedback`, `static`, `govuk_publishing_components`, and `government-frontend`, but I couldn't see any references to it at that date either, so I'm not sure it was ever used (i.e. No reference to the word `prioritise` in any of these codebases).
- Closes https://github.com/alphagov/govuk_publishing_components/issues/5698

## Visual Changes

None

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,8 +9,9 @@
 
 ## Unreleased
 
-* **BREAKING:** Replace prefixed-transform usages with standard CSS transform ([PR #5686](https://github.com/alphagov/govuk_publishing_components/pull/5704))  
-* Replace and remove scale mixin ([PR #5686](https://github.com/alphagov/govuk_publishing_components/pull/5709))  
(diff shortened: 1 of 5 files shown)


MartinJJones (dismissed):
Great work, the changes look good to me 👍

MartinJJones (approved):

PR #5710, opened 2026-09-11
Update feedback component view template

MartinJJones:
## What
Change `disable_ga4: ` to `disable_ga4: disable_ga4` in the feedback component view template

## Why
Updated to be consistent with the approach used in the all other view templates - https://github.com/search?q=repo%3Aalphagov%2Fgovuk_publishing_components+disable_ga4%3A&type=code


AshGDS (approved):
LGTM 👍

PR #5712, opened 2026-09-11
Fix invalid HTML in image card component

### LINKED ISSUE #5650: Image card component includes invalid HTML
## What

Review and fix the invalid HTML in the [image card component](https://components.publishing.service.gov.uk/component-guide/image_card#with_lazy_loading).

### Example from the 

Error: Bad value auto for attribute loading on element [img](https://html.spec.whatwg.org/#the-img-element).

```HTML
   <img class="gem-c-image-card__image" loading="auto" height="200" width="300" src="" alt="" />
```

W3C Validator: https://validator.w3.org/nu/?doc=https%3A%2F%2Fwww.gov.uk%2Fgovernment%2Forganisations%2Fhm-revenue-customs

## How to fix

A simple fix would be to change the default loading value from "auto" to "eager" in ImageCardHelper - https://github.com/alphagov/govuk_publishing_components/blob/71c15d6001bd8ef1dce1282c6a616890332fa616/lib/govuk_publishing_components/presenters/image_card_helper.rb#L16

Further info: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/img#loading

jon-kirwan:
## What

Fix invalid HTML in image card component

## Why

https://github.com/alphagov/govuk_publishing_components/issues/5650

### DIFF
--- a/lib/govuk_publishing_components/presenters/image_card_helper.rb
+++ b/lib/govuk_publishing_components/presenters/image_card_helper.rb
@@ -13,7 +13,7 @@ module GovukPublishingComponents
         @image_alt = local_assigns[:image_alt] || ""
         @srcset = local_assigns[:srcset] || nil
         @sizes = local_assigns[:sizes] || nil
-        @image_loading = local_assigns[:image_loading] || "auto"
+        @image_loading = local_assigns[:image_loading] || "eager"
         @context = local_assigns[:context]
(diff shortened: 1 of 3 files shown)


luke-booker (approved):
Looks good. Confirmed this has resolved the error in the W3C Validator 👍

PR #5718, opened 2026-09-15
Remove full stop from details component in attachment component

andysellick:
## What
Remove the full stop from the end of the sentence 'Request an accessible format' in the details component that sometimes appears in the attachment component.

## Why
Based on the guidance that link text should not include a full stop, even if it is the end of a sentence. In this case the 'link' is actually a details component summary element, but the same rule should probably apply.

It looks like the full stop was added a while back as part of [this PR to standardise the text with whitehall](https://github.com/alphagov/govuk_publishing_components/pull/1306) but it looks as though the point of the PR was to change the text - from `different format` to `accessible format` - rather than add the full stop. The full stop seems to have then been propagated throughout the language files by automatic updates to add missing keys.

Raised as a support ticket: https://gov-uk.atlassian.net/browse/PNP-10153?issueKey=PNP-10153&subProduct=jira-software

## Visual Changes
Before | After
------ | ------
Screenshot 2026-09-15 at 08 45 15 | Screenshot 2026-09-15 at 08 45 00


KludgeKML (approved):
LGTM!

AshGDS (approved):

PR #5719, opened 2026-09-15
Allow GemToggle to be activated with the space key

### LINKED ISSUE #5580: toggle.js doesn't work with spacebar

jon-kirwan:
## What

Allow GemToggle to be activated with the space key

## Why

https://github.com/alphagov/govuk_publishing_components/issues/5580


MartinJJones (commented):
Nice work, the changes look good to me, hopefully we can refactor this component in the future and make further improvements

Just wanted to check if this needs a Changelog entry as well given the new functionality?

MartinJJones (approved):

PR #5722, opened 2026-09-15
Update govuk-frontend install docs to ensure correct version is installed

hannalaakso:
## What
Update the docs for installing a new version of govuk-frontend to avoid situations where the latest (caused by the caret added to the version number in `package.json` under certain circumstances when running `npm install` / `yarn upgrade`), not the intended, version could get added to `yarn.lock`. 

Also prompt double checking the version number in `yarn.lock` after running the command to avoid unexpected edge cases (for instance, running `yarn upgrade govuk-frontend@6.5 --exact` will result in 6.5.1, which is the latest available version at the time of writing, being installed).

Yarn docs mention that:
> By default, the existing range specifier in package.json will be reused if it is one of: ^, ~, <=, >, or an exact version. Otherwise, it will be changed to a caret (^). One of the flags --caret, --tilde or --exact can be used to explicitly specify a range.

## Why
We tend to be a couple of versions behind the latest release of govuk-frontend so don't always want to install the latest version available. Whilst the minor and fix versions shouldn't contain any breaking changes, we often prefer depending on what it's in the release to do some manual testing or to update code that has become deprecated.


jon-kirwan (approved):
Thanks Hanna, nice update to the docs 👍

PR #5725, opened 2026-09-16
Hide feedback cancel button when JS is disabled

### LINKED ISSUE #5717: Feedback component when JS disabled has a non-functional 'cancel' button
If JavaScript is disabled much of the functionality of the [feedback component](https://components.publishing.service.gov.uk/component-guide/feedback) won't work, so it gets replaced with a link to a feedback survey.

It also now includes a 'cancel' button that doesn't do anything.

Image

The same cancel button also appears when you click the 'no' button when JS is enabled, so I guess it's been accidentally overlooked when hiding elements when JS is disabled.

It's this button:

```
<button class="gem-c-button govuk-button gem-c-button--secondary-quiet govuk-!-margin-bottom-6 js-close-form" type="submit" aria-controls="page-is-not-useful">
Cancel
</button>
```

andysellick:
## What
Adjust the feedback component so that the 'cancel' button (shown when the user chooses 'no') is only visible when JS is enabled.

Also removes some redundant code.

## Why
Fixes https://github.com/alphagov/govuk_publishing_components/issues/5717

## Visual Changes
Should be no visual change when JS is enabled, but the cancel button should now be hidden when JS is not enabled.

Before | After
------ | -----
Screenshot 2026-09-16 at 09 57 09 | Screenshot 2026-09-16 at 09 56 57


app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
```diff
@@ -228,9 +228,10 @@ $feedback-prompt-border-top-colour: $govuk-blue-tint-50;
 }
 
 .gem-c-feedback__close {
-  margin: 0 govuk-spacing(2);
-  @include govuk-media-query($until: tablet) {
-    margin: govuk-spacing(4) 0 0;
+  display: none;
+
+  .js-enabled & {
```
MartinJJones: Something I remember from the v5 upgrade, I think we may need to change `js-enabled` to [govuk-frontend-supported](https://github.com/alphagov/govuk_publishing_components/blob/82ab3bae5f8752dfd49e648f9124fcc422805af6/app/views/layouts/govuk_publishing_components/application.html.erb#L49)

This should ensure the button does not show in [Safari 10.1](https://frontend.design-system.service.gov.uk/v5/browser-support/#grade-c) as well 

Example screenshot from BrowserStack

Image

MartinJJones (commented):
Nice find on this issue and the fix, the changes look good to me overall, just one small comment relating to very specific older browser support/behaviour

app/assets/stylesheets/govuk_publishing_components/components/_feedback.scss
andysellick: Thanks, have updated 👍

MartinJJones: Nice, that's fixed it, thanks :)

MartinJJones (approved):
Nice work, the changes look to me

PR #5726, opened 2026-09-16
Add support for custom search terms in filter-list.js

### LINKED ISSUE #4764: Make the search work with alternative search terms
Example: I found code referring to the `layout_for_admin` component. So I visited <https://components.publishing.service.gov.uk/component-guide> and:

1. Searched for "layout_for_admin" - no results.
2. Removed underscores to search for "layout for admin" instead - no results.
3. Searched for "admin" instead and had an "[Admin layout](https://components.publishing.service.gov.uk/component-guide/layout_for_admin)" result.

I think the human-readable name _and_ the 'partial' name should be usable in search.

MartinJJones:
## What
Add support for custom search terms in `filter-list.js`

## Why
To help improve the search results, you can now pass custom search terms to `data-filter-item-search-terms`, this is used in the component guide to pass in the component template name, `layout_for_admin` for example

Fixes: https://github.com/alphagov/govuk_publishing_components/issues/4764


app/views/govuk_publishing_components/component_guide/_component_list.html.erb
```diff
@@ -4,7 +4,7 @@
 <%= local_assigns[:details] if local_assigns[:details] %>
 <ul class="govuk-list govuk-list--bullet govuk-list--spaced">
   <% local_assigns[:components].each do |component_doc| %>
-    <li data-filter-item>
+    <li data-filter-item data-filter-item-search-terms="<%= component_doc.id %>">
```
andysellick, line 7: Is this specific addition useful? Do people search for components using underscores? Or are you just putting it somewhere so it's being used?

Now that this is getting a bit more feature rich we should probably have documentation for it?

andysellick (approved):
Nice addition 👏

app/views/govuk_publishing_components/component_guide/_component_list.html.erb
MartinJJones: > Is this specific addition useful? Do people search for components using underscores? Or are you just putting it somewhere so it's being used?

There is an example of underscores being used in the issue raised here - https://github.com/alphagov/govuk_publishing_components/issues/4764, I think I've tried this in the past a few times as well

> Now that this is getting a bit more feature rich we should probably have documentation for it?

Yes definitely, I'll take a look at this

PR #5730, opened 2026-09-17
Fix search label clipping

### LINKED ISSUE #5703: Longer search labels unexpectedly clipped
The search component label .gem-c-search__label (search.scss) is constrained to max-width: 68%.

Image

While intended for inline labels to prevent text overlapping the search button, these styles are also applied to search labels above the search input. When a search component uses a longer label (e.g. instructions containing "for example X"), the text hits the 68% limit and is visually clipped out of view. The first line "Search keywords" is clipped and has an ellipsis added. 

Search containing "for example X" second line: https://www.gov.uk/find-licences
Inline label example: https://www.gov.uk/hmrc-internal-manuals/partnership-manual/134300
PR fixing initial bug: https://github.com/alphagov/govuk_publishing_components/pull/4204
Component guide: https://components.publishing.service.gov.uk/component-guide/search/preview

**Proposed fix**
Scope the styles to the inline label option only.
Add a multi line variant to search in the component guide for visibility and testing.

luke-booker:
## What / Why
Currently .gem-c-search__label applies max-width: 68% globally to prevent absolute positioned inline label text from overlapping the search button. As a result, labels block positioned above the search bar with longer (or multiline) text get cut at 68% width with an ellipsis, despite having full width space available. 

This PR adds CSS constraints so labels can expand naturally, across multiple lines if needed. It also includes a long label example to the component guide.

Fixes: #5703

Related issue around HTML label injection: https://github.com/alphagov/finder-frontend/issues/4179

## Testing / Visual Changes

https://www.gov.uk/find-licences (only use case)

### Mobile (no change as text not clipped)
Screenshot 2026-09-17 at 12 09 31

### Tablet 

Before

Screenshot 2026-09-17 at 17 16 08

After

Screenshot 2026-09-17 at 12 09 52

### Desktop (no change as text not clipped)
Screenshot 2026-09-17 at 12 10 10

### Device / Browser Spot Checks

Device / browser | Outcome
-- | --
macOS Tahoe - Chrome 153 | No change, as expected
macOS Tahoe - Safari 26.4 | No change, as expected
macOS Sonoma - Firefox 156 | No change, as expected
macOS Catalina - Edge 128 | No change, as expected
Android 13 - Samsung Galaxy S23 (Chrome) | No change, as expected
Android 11 - Samsung Galaxy S21 (Chrome) | No change, as expected
Android 10 - Samsung Galaxy S20 (Samsung Internet) | No change, as expected
iOS 11 - Safari | No change, as expected
iOS 17 - Chrome | No change, as expected
iOS 13 Pro - Safari | No change, as expected
Windows 10 - Edge 100 | No change, as expected
Windows 10 - Edge 15 | No change, as expected
Windows 10 - IE11 | No change, as expected


app/views/govuk_publishing_components/components/docs/search.yml
```diff
@@ -98,6 +98,14 @@ examples:
...
       label_margin_bottom: 9
       inline_label: false
+  with_long_label_text:
+    description: |
+      Demonstrates how long label text wraps across multiple lines without being clipped.
+
+      Requires `inline_label` to be false.
+    data:
+      inline_label: false
+      label_text: Search for a licence, permit, or regulation by entering keywords, document title, or reference number.
```
MartinJJones: non-blocking nitpick: I would be tempted to remove the full stop from the label text here to be consistent with the other labels

MartinJJones (dismissed):
Nice work, the changes look good to me, I have one small non-blocking suggestion as well

app/views/govuk_publishing_components/components/docs/search.yml
luke-booker: Good spot. Updated that

MartinJJones (approved):

PR #5735, opened 2026-09-18
Fix details component when metadata is inverse

andysellick:
## What
Passes the `inverse` flag given to the metadata component to the details components used as child components, so that when the metadata component is on a dark background the details component also uses white text

## Why
It was unreadable. Thanks to @AshGDS for spotting this.

## Visual Changes
Before | After
------ | -----
Screenshot 2026-09-18 at 12 01 07 | Screenshot 2026-09-18 at 12 01 20
Screenshot 2026-09-18 at 12 01 41 | Screenshot 2026-09-18 at 12 01 48


AshGDS (approved):
LGTM 👍

